PR #14, opened 2021-09-05
add Dockerfile & docker-compose for pd

hawkw:
This branch adds an initial setup for running a Penumbra node using
`docker-compose`, and a `Dockerfile` for building a `pd` docker image.

There's some room for additional work here. In particular:

- the `docker-compose` file could also include Prometheus and
  Grafana containers (and Prometheus can be configured to scrape metrics
  from both Tendermint *and* pd)
- the correct way to volume mount the Tendermint dir so that it can be
  effected by running tendermint CLI commands locally on the host (for
  dev purposes) remains to be figured out
- `docker-compose` configuration should probably make a private network
  for tendermint -> pd RPCs so that nothing else on the host is allowed
  to talk to the pd process' ABCI RPCs
- Tendermint has a nifty [docker-compose config][1] for running a local
  testnet of several nodes in Docker, we might want something like that
  as well for dev purposes?
- I should actually write down how to test this.
- Docker image Rust build times could be improved using [experimental
  Docker features][2]; I punted on this for now because I didn't want
  to introduce potential complexity for anyone trying to run this...
- It would be cool to use a `scratch` base image rather than Debian,
  but then we would need a statically linked libc (read: musl), which
  is its own can of worms...

We may want to eventually have separate `docker-compose` configs for
actually distributing to run in production and for spinning up dev
testnets. But, for now, this basically Just Works.

[1]: https://docs.tendermint.com/master/networks/docker-compose.html
[2]: https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/experimental.md#run---mounttypecache

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,10 @@
+notes/**
+.github
+.firebaserc
+theme
+book.toml
+*.md
+Dockerfile
+docker-compose.yml
+*.json
+target/**
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,23 @@
+FROM rust:1.54.0 as build
+
+WORKDIR /usr/src
+COPY . .
+
+# Fetch dependencies in a separate layer, so that they can be cached.
+RUN cargo fetch --locked
+
+RUN cargo build --bin pd --release --frozen && \
+    mkdir -p /out && \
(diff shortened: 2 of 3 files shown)


hdevalence:
Awesome, thanks!!

PR #20, opened 2021-09-06
docker-compose: add grafana & prometheus, misc improvements

hawkw:
This branch adds a `docker-compose.override.yml` containing a local dev configuration for testing the Penumbra daemon. This override file adds a local Prometheus and Grafana instance configured to scrape the Tendermint node and Penumbra daemon (note that the penumbra daemon doesn't actually expose metrics yet). This is factored out into an [override file][1] so that the main `docker-compose.yml` can be used as a canonical deployment configuration without also including a per-node Grafana & Prometheus instance (prod users will probably want to bring their own monitoring, and e.g. deploy a central Prometheus that collects metrics from a large number of Penumbra nodes; include authorization for viewing Grafana, etc).  The Grafana dashboard configuration is a read-write bind mount in the container, so we can develop dashboards using the web UI and save them to the git repo. Right now, though, there aren't any dashboards yet. It would be nice to find some grafana dashboards for general-purpose tendermint monitoring and drop them in, so we can at least look at what the Tendermint node is doing...

I also fixed up a couple things in the existing config. In particular, I switched the indentation to 2-space tabs, which is how you're supposed to indent yaml, and I changed how Tendermint data is stored. It's now in a docker volume mount rather than a bind mount, so it can be managed using Docker and (potentially) persisted, rather than just going straight to `/tmp`. 

[1]: https://docs.docker.com/compose/extends/
image
(prometheus considers the penumbra daemon "down" because it isn't actually responding to metrics scrapes, as it...doesn't have metrics yet...)


hdevalence (approved):
Awesome, thanks so much!!

Issue #24, opened 2021-09-07
Specify key hierarchy for Penumbra's shielded transaction model


hdevalence:
The [protocol specification](https://protocol.penumbra.zone/main/index.html) should specify the key hierarchy for Penumbra's shielded transaction model. While this is based on Sapling, there will be some changes (e.g., `decaf377` provides infallible hash-to-group, so there are no invalid diversifiers, addresses should have detection keys, etc).

PR #61, opened 2021-09-17
spec: addresses

redshiftzero:
Rest of initial spec for #24


hdevalence (approved):
LGTM.  We should probably have somewhere that specifies the address encodings.

PR #62, opened 2021-09-20
proto: add skeleton of penumbra-proto crate

hdevalence:
This adds a skeleton of a crate that holds Protobuf definitions of Penumbra
datatypes, and a `transaction.proto` file with some scratch work on transaction
structure definitions.

### DIFF
--- a/proto/proto/transaction.proto
+++ b/proto/proto/transaction.proto
@@ -1,0 +1,81 @@
+syntax = "proto3";
+package penumbra.transaction;
+
+// A Penumbra transaction.
+message Transaction {
+  TransactionBody body = 1;
+  // The binding signature is stored separately from the transaction body that it signs.
+  bytes binding_sig = 2;
+}
+
+// The body of a transaction.
+message TransactionBody {
+  // A list of actions (state changes) performed by this transaction.
+  repeated Action actions = 1;
+  // The root of some previous state of the note commitment tree.
+  bytes anchor = 2;
+  // The maximum height that this transaction can be included in the chain.
+  uint32 expiry_height = 3;
+  // The chain this transaction is intended for.  Including this prevents
+  // replaying a transaction on one chain onto a different chain.
+  string chain_id = 4;
+}
+
+// A state change performed by a transaction.
+message Action {
+  oneof action {
+    Fee fee = 1;
+    Spend spend = 2;
+    Output output = 3;
+  }
+}
+
+// Specifies fees paid by a transaction.
+message Fee {
+    uint64 amount = 1;
+}
+
+// Spends a shielded note.
+message Spend {
+  SpendBody body = 1;
+  // The spend authorization signature is stored separately from the spend body it authorizes.
+  bytes auth_sig = 2;
+}
+
+// The body of a spend description, stored separately from the signatures that authorize it.
+message SpendBody {
+  // A commitment to the value of the input note.
+  bytes cv = 1;
+  // The nullifier of the input note.
+  bytes nullifier = 3;
+  // The randomized validating key for the spend authorization signature.
+  bytes rk = 4;
+  // The spend proof.
+  bytes zkproof = 5;
+}
+
+// Creates a new shielded note.
+message Output {
+  OutputBody body = 1;
+  // An encrypted memo. 528 bytes.
+  bytes encrypted_memo = 2;
+  // The key material used for note encryption, wrapped in encryption to the
+  // sender's outgoing viewing key. 80 bytes.
+  bytes ovk_wrapped_key = 3;
+}
+
+// The body of an output description, not including a memo or ovk wrapping.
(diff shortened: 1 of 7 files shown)


proto/src/lib.rs
```diff
@@ -0,0 +1,23 @@
...
+//! from them.  These types only handle parsing the wire format; validation
+//! should be performed by converting them into an appropriate domain type, as
+//! in the following diagram:
+//!
+//! ```ascii
+//! ┌───────┐          ┌──────────────┐               ┌──────────────┐
+//! │encoded│ protobuf │penumbra_proto│ TryFrom/Into  │ domain types │
+//! │ bytes │<──wire ─>│    types     │<─validation ─>│(other crates)│
+//! └───────┘  format  └──────────────┘   boundary    └──────────────┘
+//! ```
```
hdevalence, line 13: Eventually, we should have domain types that model the proto types, but for now (not sure for how long?), I think it's easier to just iterate on the proto design than on both protos + a set of domain types + conversions.

proto/proto/transaction.proto
```diff
@@ -0,0 +1,81 @@
...
+  // An encrypted memo. 528 bytes.
+  bytes encrypted_memo = 2;
+  // The key material used for note encryption, wrapped in encryption to the
+  // sender's outgoing viewing key. 80 bytes.
+  bytes ovk_wrapped_key = 3;
+}
+
+// The body of an output description, not including a memo or ovk wrapping.
+// Splitting this data out allows its use where the memo / ovk wrapping is not
+// required (e.g., IBC).
```
hdevalence, line 69: The plan is for IBC transfers into Penumbra to work by having the IBC Fungible Token Transfer packet on the counterparty chain specify a base64-encoded output description in the destination address field; in that case, it's better to keep the description as small as possible, and there's no clear need to have a memo field or ovk support (since the transaction is in the clear on the counterparty chain).

proto/proto/transaction.proto
```diff
@@ -0,0 +1,81 @@
...
+  // The randomized validating key for the spend authorization signature.
+  bytes rk = 4;
+  // The spend proof.
+  bytes zkproof = 5;
+}
+
+// Creates a new shielded note.
+message Output {
+  OutputBody body = 1;
+  // An encrypted memo. 528 bytes.
```
hdevalence, line 60: 528 = 512 + 16-byte MAC

proto/proto/transaction.proto
```diff
@@ -0,0 +1,81 @@
...
+// Splitting this data out allows its use where the memo / ovk wrapping is not
+// required (e.g., IBC).
+message OutputBody {
+  // A commitment to the value of the output note. 32 bytes.
+  bytes cv = 1;
+  // The u-coordinate of the note commitment for the output note. 32 bytes.
+  bytes cmu = 2;
+  // The encoding of an ephemeral public key. 32 bytes.
+  bytes ephemeral_key = 3;
+  // An encryption of the newly created note. 80 bytes.
```
hdevalence, line 77: 80 = 64 + 16-byte MAC; this will grow by 32 bytes probably with multi-asset support.

proto/proto/transaction.proto
```diff
@@ -0,0 +1,81 @@
...
+  // The binding signature is stored separately from the transaction body that it signs.
+  bytes binding_sig = 2;
+}
+
+// The body of a transaction.
+message TransactionBody {
+  // A list of actions (state changes) performed by this transaction.
+  repeated Action actions = 1;
+  // The root of some previous state of the Sapling note commitment tree.
+  bytes anchor = 2;
```
hdevalence, line 16: This anchor is common to all spends, unlike Sapling and like Orchard.

proto/proto/transaction.proto
```diff
@@ -0,0 +1,81 @@
+syntax = "proto3";
+package penumbra.transaction;
+
+// A Penumbra transaction.
+message Transaction {
+  TransactionBody body = 1;
+  // The binding signature is stored separately from the transaction body that it signs.
```
hdevalence, line 7: Here and elsewhere the idea is to separate signatures on some structure from the structure itself, so that we can more easily define what data the signature is signing, without having to go field by field.

proto/proto/transaction.proto
```diff
@@ -0,0 +1,81 @@
...
+  bytes binding_sig = 2;
+}
+
+// The body of a transaction.
+message TransactionBody {
+  // A list of actions (state changes) performed by this transaction.
+  repeated Action actions = 1;
```
hdevalence, line 14: The idea is that the value balance over all actions performed by a transaction should be 0.

hdevalence:
Self-review comments have some notes on the stub proto file for discussion, but they should be non-blocking for this PR.

redshiftzero (approved):
LGTM!

PR #68, opened 2021-10-05
spec: flesh out `decaf377` section

### LINKED ISSUE #3: Specify `decaf377`, Decaf-for-Edwards-over-BLS12-377.
Unless there's a dramatic change to the analysis in #2, very likely that Penumbra will use BLS12-377. Penumbra needs a cryptographic group that can be used inside of an arithmetic circuit. This group would play the role Jubjub plays in Sapling, and be used for other purposes. 

The Zexe paper, which defined BLS12-377, also defined (but did not name) a cofactor-4 Edwards curve defined over the BLS12-377 scalar field.  It would be possible to use the Edwards curve directly, but doing so would provide a leaky abstraction, forcing all of the downstream constructions to pay attention to cofactors.  Instead, it would be much better to specify (and name) Decaf for this curve, so that all protocols using BLS12-377 (including Penumbra) can have a clean abstraction to work with.

In the "machine" cost model for elliptic curve implementations that execute machine instructions, Decaf (and Ristretto) impose negligible additional cost compared to the underlying Edwards curve, because the dominant cost of encoding and decoding is an inverse square root operation.  However, in the "circuit" cost model for an elliptic curve inside of an arithmetic circuit, this is no longer the case, because arithmetic circuits certify computation rather than perform it, and certifying an inverse is exactly as expensive as certifying a multiplication.

This means that unlike the machine case, using Decaf in a circuit does impose additional costs.  These costs should be quantified precisely, but they are almost certainly worthwhile for general-purpose applications.  Decaf provides a single, unified abstraction that works the same way inside and outside of a circuit and the same way for all applications, making it a good default choice.

Some special cases, such as Pedersen hashes, may benefit from using the underlying curve directly, but this functionality can be encapsulated as, e.g., a hash function.

The notes should include:

- [x] the above problem description;
- [x] a name for Decaf-for-Edwards-over-BLS12-377 (and Edwards-over-BLS12-377, which is missing one);
- [x] a specification of the encoding and decoding functions with test vectors;
- [x] a specification of hash-to-group -- this could usefully have two, one single-width and one double-width (hash twice and add), depending on the cost of the hash-to-group method;
- [x] a survey of other methods for handling cofactors inside a circuit and a comparison of the costs.

### LINKED ISSUE #67: deviation between spec and impl for point compression
For decaf377 [encoding on step 1 in the spec](https://penumbra.zone/main/crypto/decaf377/encoding.html) we're doing `u1​<-(X+T)(Y−T)`, but in the [code](https://github.com/penumbra-zone/decaf377/blob/main/src/group.rs#L75) we're doing `let u_1 = (p.x + p.t) * (p.x - p.t);`

Spec is wrong, code is right?

redshiftzero:
Closes #3. Closes #67.

Remaining TODOs I'll put in a new issue on MVP2 when this is merged:
- Optimize inverse square root algorithm
- Optimize hash-to-group


spec/crypto/decaf377/test_vectors.md
```diff
@@ -21,4 +21,19 @@ This table has hex-encodings of small multiples of the generator $B$:
...
+| `1502379126429822955521756759528876454108853047288874182661923263559139887582` | `7074060208122316523843780248565740332109149189893811936352820920606931717751` |
+| `2943006201157313879823661217587757631000260143892726691725524748591717287835` | `4988568968545687084099497807398918406354768651099165603393269329811556860241` |
+| `2893226299356126359042735859950249532894422276065676168505232431940642875576` | `5540423804567408742733533031617546054084724133604190833318816134173899774745` |
+| `2950911977149336430054248283274523588551527495862004038190631992225597951816` | `4487595759841081228081250163499667279979722963517149877172642608282938805393` |
+| `3318574188155535806336376903248065799756521242795466350457330678746659358665` | `7706453242502782485686954136003233626318476373744684895503194201695334921001` |
+| `3753408652523927772367064460787503971543824818235418436841486337042861871179` | `2820605049615187268236268737743168629279853653807906481532750947771625104256` |
+| `7803875556376973796629423752730968724982795310878526731231718944925551226171` | `7033839813997913565841973681083930410776455889380940679209912201081069572111` |
```
hdevalence, line 39: Here it would probably be good to specify the s-coordinate, but it could be a later change than this PR.

hdevalence (approved):

PR #71, opened 2021-10-06
feat: keys implementation

### LINKED ISSUE #24: Specify key hierarchy for Penumbra's shielded transaction model

redshiftzero:
Closes #24

### DIFF
--- a/crypto/src/keys.rs
+++ b/crypto/src/keys.rs
@@ -1,13 +1,136 @@
-use once_cell::sync::Lazy;
+use std::convert::TryInto;
 
 use ark_ff::PrimeField;
 use decaf377;
+use once_cell::sync::Lazy;
+use rand_core::{CryptoRng, RngCore};
 
-use crate::Fq;
+use crate::{Fq, Fr};
 
-pub struct OutgoingViewingKey {}
+pub const DIVERSIFIER_LEN_BYTES: usize = 11;
+pub const SPEND_LEN_BYTES: usize = 32;
+pub const OVK_LEN_BYTES: usize = 32;
 
-pub struct Diversifier(pub [u8; 11]);
+pub struct SpendingKey(pub [u8; SPEND_LEN_BYTES]);
+
+impl SpendingKey {
+    pub fn generate<R: RngCore + CryptoRng>(mut rng: R) -> Self {
+        let mut key = [0u8; SPEND_LEN_BYTES];
+        // Better to use Rng trait here?
+        rng.fill_bytes(&mut key);
+        SpendingKey(key)
+    }
+}
+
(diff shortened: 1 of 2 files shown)


crypto/src/keys.rs
```diff
@@ -1,20 +1,144 @@
...
+}
+
+pub struct EphemeralPublicKey(pub decaf377::Element);
+
+pub struct ProofAuthorizationKey {
+    pub ak: AuthorizationKey,
+    pub nsk: NullifierPrivateKey,
+}
+
+pub struct AuthorizationKey(pub decaf377::Element);
```
hdevalence, line 100: this might be a decaf377-rdsa signing key instead?

redshiftzero: ah yeah, looks like `SpendAuthorizationKey` should be `decaf377-rdsa::SigningKey` and `AuthorizationKey` should be `decaf377_rdsa::VerificationKey` (related: the spend auth keys could be named a bit better imho, maybe `SpendAuthorizationSigningKey` and `AuthorizationVerificationKey` might be clearer in the future, not intending to do that right now but food for thought)

hdevalence: one thing to note is that since they have a type parameter, they end up as `SigningKey<SpendAuth>` or `VerificationKey<Binding>`, so the domain is still included in the name

crypto/src/keys.rs
```diff
@@ -1,20 +1,144 @@
...
+
+pub struct NullifierPrivateKey(pub Fr);
+
+impl NullifierPrivateKey {
+    pub fn derive(key: &SpendingKey) -> Self {
+        let mut input = Vec::<u8>::new();
+        input.extend_from_slice(b"Penumbra_ExpandSeed");
+        input.extend_from_slice(&key.0);
+        input.extend_from_slice(&[0; 1]);
+        let hash_input = blake2b_simd::blake2b(&input).as_bytes();
```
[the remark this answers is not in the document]
redshiftzero: ah sweet, it also looks like there is an incremental hashing interface in blake2b_simb so i switched to that, but should we use the `blake2` crate over the `blake2b_simb` crate?

redshiftzero:
note: this is pinned to a branch over in https://github.com/penumbra-zone/decaf377-rdsa/pull/1 such that CI passes - will snip that out before merge

Issue #26, opened 2021-09-07
Specify note contents


hdevalence:
Fill in this section of the protocol spec; settle on (and write up) a choice of `leadByte` method.

hdevalence:
This is done now that we use the `rseed` mechanism.

PR #73, opened 2021-10-06
feat: progress towards OutputDescription, SpendDescription

redshiftzero:
progress towards #26

### DIFF
--- a/crypto/Cargo.toml
+++ b/crypto/Cargo.toml
@@ -7,7 +7,7 @@ edition = "2018"
 
 [dependencies]
 decaf377 = { git = "https://github.com/penumbra-zone/decaf377" }
-decaf377-rdsa = { git = "https://github.com/penumbra-zone/decaf377-rdsa", branch = "more-stuff-for-pen-keys" }
+decaf377-rdsa = { git = "https://github.com/penumbra-zone/decaf377-rdsa" }
 poseidon377 = { git = "https://github.com/penumbra-zone/poseidon377" }
 hex = "0.4"
 blake2b_simd = "0.5"


crypto/Cargo.toml
```diff
@@ -7,8 +7,8 @@ edition = "2018"
 
 [dependencies]
 decaf377 = { git = "https://github.com/penumbra-zone/decaf377" }
-poseidon377 = { git = "https://github.com/penumbra-zone/poseidon377" }
 decaf377-rdsa = { git = "https://github.com/penumbra-zone/decaf377-rdsa" }
+poseidon377 = { git = "https://github.com/penumbra-zone/poseidon377", branch = "rate5" }
```
hdevalence, line 11: merged the rate5 branch just now 👍

crypto/src/actions.rs
```diff
@@ -0,0 +1,71 @@
...
+use decaf377::{Fq, Fr};
+use decaf377_rdsa;
+
+use crate::addresses::PaymentAddress;
+use crate::keys::{EphemeralPublicKey, SpendAuthorizationKey};
+use crate::note::{Note, NoteCommitment};
+use crate::nullifier::Nullifier;
+use crate::value::{Commitment, Value};
+
+pub struct OutputDescriptionBody {
```
hdevalence, line 13: On naming / code organization, since we'll eventually have a bunch of actions, I think it would be good to have this be `action::output::Body` (i.e., `actions.rs -> action.rs`, with `output` and `spend` submodules).  This way we can keep the names short and tidy and keep the code for the different actions in different submodules.

redshiftzero: i dig it

crypto/src/keys.rs
```diff
@@ -57,6 +57,15 @@ impl SpendAuthorizationKey {
...
     }
+
+    pub fn randomize(
+        &self,
+        randomizer: Fr,
+    ) -> decaf377_rdsa::VerificationKey<decaf377_rdsa::SpendAuth> {
+        let rk: decaf377_rdsa::VerificationKey<decaf377_rdsa::SpendAuth> =
+            self.0.randomize(&randomizer).into();
+        rk
+    }
```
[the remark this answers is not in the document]
redshiftzero: Makes sense, I'll also do the same for the `AuthorizationKey` (can be `VerificationKey<SpendAuth>` directly) since the same logic applies

PR #74, opened 2021-10-07
feat: add key agreement crate for decaf377

redshiftzero:


decaf377-ka/Cargo.toml
```diff
@@ -0,0 +1,11 @@
...
+name = "decaf377-ka"
+version = "0.1.0"
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+ark-ff = { git = "https://github.com/penumbra-zone/algebra", branch = "ours" }
+decaf377 = { git = "https://github.com/penumbra-zone/decaf377" }
+rand_core = { version = "0.6.3", features = ["getrandom"] }
```
hdevalence, line 11: what's the `getrandom` feature for? it doesn't seem to be used in the code, am i missing something?

redshiftzero: ah, it's used in the tests (via `OsRng`) - I can move this into a dev dependencies block though

decaf377-ka/src/lib.rs
```diff
@@ -0,0 +1,52 @@
...
+
+use decaf377;
+use rand_core::{CryptoRng, RngCore};
+
+/// A `SharedSecret` derived at the end of the key agreement protocol.
+#[derive(Debug, PartialEq)]
+pub struct SharedSecret(pub(crate) decaf377::Element);
+
+impl SharedSecret {
+    pub fn derive(other_public_key: &EphemeralPublicKey, secret_key: EphemeralSecretKey) -> Self {
```
hdevalence, line 11: I think the API would be more ergonomic if this were

```rust
impl EphemeralSecretKey {
    pub fn key_agreement_with(self, other: &EphemeralPublicKey) -> SharedSecret { ... }
}
```

or some other name, we could bikeshed it a little bit, but the important thing is that by making the consumed secret be a `self` parameter, there's a clearer distinction between the two parameters: I use my secret key with the counterparty's public key as `my_secret.key_agreement_with(other_pub)` or something, rather than having to remember which parameter comes first (is it public/secret or secret/public?). 

It's also more common to have APIs that consume `self` as a way to enforce state transitions, so I think that's also a bit more familiar.

decaf377-ka/src/lib.rs
```diff
@@ -0,0 +1,52 @@
+use ark_ff::UniformRand;
+
+use decaf377;
+use rand_core::{CryptoRng, RngCore};
+
+/// A `SharedSecret` derived at the end of the key agreement protocol.
+#[derive(Debug, PartialEq)]
+pub struct SharedSecret(pub(crate) decaf377::Element);
```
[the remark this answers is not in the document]
hdevalence: Also, whichever of those options are picked, I don't think there's a downside to making the inner type `pub`; it's a bit easier for people to access the shared secret data that way, and there aren't any operations on the `SharedSecret` (it's the terminal point of the computation) so there are no invariants to maintain.

decaf377-ka/Cargo.toml
hdevalence: (expansion of what I mean by "not worth it": it causes problems sufficient that there are entire tools whose purpose is to re-unify feature selections, like @sunshowers ' `cargo-hakari`: https://docs.rs/cargo-hakari/0.9.5/cargo_hakari/about/index.html#what-are-workspace-hack-crates )

decaf377-ka/src/lib.rs
```diff
@@ -0,0 +1,52 @@
...
+
+impl SharedSecret {
+    pub fn derive(other_public_key: &EphemeralPublicKey, secret_key: EphemeralSecretKey) -> Self {
+        Self(other_public_key.0 * secret_key.0)
+    }
+}
+
+/// An `EphemeralSecretKey` is used once and consumed when forming a `SharedSecret`.
+pub struct EphemeralSecretKey(pub(crate) decaf377::Fr);
```
hdevalence, line 17: Non-blocking for this PR, but we should think about how the incoming viewing key is supposed to interact with this API -- we probably also want to have a non-ephemeral secret key type, and have the IVK be an instance of that type.

hdevalence (approved):

PR #75, opened 2021-10-09
feat: incremental 4-arity merkle tree

redshiftzero:
This is a module with an incremental 4-arity merkle tree (where we are including the layer with the domain separator)

needed next:
* Serialization/deserialization
* proofs of inclusion


crypto/src/error.rs
```diff
@@ -0,0 +1,11 @@
...
+
+#[derive(Error, Debug)]
+pub enum MerkleTreeError {
+    #[error("Merkle tree too large")]
+    TooLarge,
+    #[error("Already at root")]
+    RootHasNoParent,
+    #[error("Already at leaf")]
+    LeafHasNoChild,
+}
```
hdevalence, line 11: I wonder whether we should try to keep all of the errors in one place (here, `crate::error::MerkleTreeError`), or to keep the errors near the functionality they relate to (in this case, `crate::merkle::Error`). I kind of lean towards the latter, since it means that the errors are together with the code that creates them, and the names can be factored into the module path (as in the stdlib, where you have `std::io::Error`, etc).

crypto/src/keys.rs
```diff
@@ -121,15 +125,44 @@ pub struct ProofAuthorizationKey {
...
     pub ak: VerificationKey<SpendAuth>,
     pub nk: NullifierDerivingKey,
     pub ovk: OutgoingViewingKey,
 }
 
+impl FullViewingKey {
+    pub fn derive(
+        ak: &VerificationKey<SpendAuth>,
+        nsk: &NullifierPrivateKey,
+        ovk: &OutgoingViewingKey,
```
hdevalence, line 148: similar question as above about references/ownership.

Is there a misuse risk from taking input parameters like this -- what happens if the inputs that are passed in aren't all derived from the same root key material? Could we alternately have the FVK derived from another structure that would ensure all the key material is consistent?

crypto/src/keys.rs
```diff
@@ -139,6 +172,13 @@ static DIVERSIFY_GENERATOR_DOMAIN_SEP: Lazy<Fq> = Lazy::new(|| {
 });
 
 impl Diversifier {
+    /// Generate a new random diversifier.
+    pub fn generate<R: RngCore + CryptoRng>(mut rng: R) -> Self {
+        let mut diversifier = [0u8; DIVERSIFIER_LEN_BYTES];
+        rng.fill_bytes(&mut diversifier);
+        Diversifier(diversifier)
+    }
```
hdevalence, line 180: This is good for now, but in the future I think we should probably shift the design to use a Poseidon-based diversifier design as in Orchard.

crypto/src/keys.rs
```diff
@@ -150,11 +190,34 @@ impl Diversifier {
     }
 
     #[allow(non_snake_case)]
-    pub fn derive_transmission_key(&self, ivk: IncomingViewingKey) -> TransmissionKey {
+    pub fn derive_transmission_key(&self, ivk: &IncomingViewingKey) -> TransmissionKey {
```
hdevalence, line 193: should this be a method on the `Diversifier` or on the `IncomingViewingKey`? since there are multiple diversifiers for the same IVK, maybe it would make more sense for it to be a method on the IVK rather than the diversifier?

crypto/src/keys.rs
```diff
@@ -150,11 +190,34 @@ impl Diversifier {
...
+    use rand_core::OsRng;
+
+    use crate::addresses::PaymentAddress;
+
+    #[test]
+    fn complete_key_generation_happy_path() {
+        let mut rng = OsRng;
+        let diversifier = Diversifier::generate(&mut rng);
+        let sk = SpendingKey::generate(&mut rng);
+        let expanded_sk = ExpandedSpendingKey::derive(&sk);
```
hdevalence, line 216: general comment: I wonder if we want to have separate types for the spending key and expanded spending key, or whether we could use an approach like the one I implemented for `decaf377-rdsa`, where the signing key type contains both the secret key as well as a cached copy of the public key: https://github.com/penumbra-zone/decaf377-rdsa/blob/main/src/signing_key.rs#L18-L21

crypto/src/merkle/mod.rs
```diff
@@ -0,0 +1,5 @@
+pub mod constants;
+pub mod hash;
+pub mod node;
+pub mod proof;
+pub mod tree;
```
hdevalence, line 5: I think it probably makes sense for these modules to be private, and have their contents re-exported out of the `merkle` module (e.g., `merkle::Tree`, `merkle::Proof`, etc) as needed.

crypto/src/keys.rs
```diff
@@ -121,15 +125,44 @@ pub struct ProofAuthorizationKey {
     pub nsk: NullifierPrivateKey,
 }
 
+impl ProofAuthorizationKey {
+    pub fn derive(ask: &SigningKey<SpendAuth>, nsk: &NullifierPrivateKey) -> Self {
```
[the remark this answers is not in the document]
redshiftzero: Ah I see, this should be an impl on `ExpandedSpendingKey` :+1:

crypto/src/keys.rs
redshiftzero: I like that idea, especially given how closely related the two keys are - I didn't make the change in this PR but filed #76 for this so we don't lose this feedback

redshiftzero:
Recapping plan of record from Discord: Gonna merge this in given the size / key changes / code organization changes, then I'll work on dropping in [our fork](https://github.com/penumbra-zone/incrementalmerkletree) of `incrementalmerkletree` in a followup PR

PR #78, opened 2021-10-13
Refactor spending key

### LINKED ISSUE #76: replace `SpendingKey` and `ExpandedSpendingKey` with a single type
[Followup](https://github.com/penumbra-zone/penumbra/pull/75/files#r725563096) from #75:

>general comment: I wonder if we want to have separate types for the spending key and expanded spending key, or whether we could use an approach like the one I implemented for decaf377-rdsa, where the signing key type contains both the secret key as well as a cached copy of the public key: https://github.com/penumbra-zone/decaf377-rdsa/blob/main/src/signing_key.rs#L18-L21

hdevalence:
Closes #76


spec/protocol/addresses_keys/spend_key.md
```diff
@@ -0,0 +1,32 @@
...
+The outgoing viewing key is derived by truncating the PRF output:
+```
+ovk = truncate_32_bytes(prf_expand(2))
+```
+
+[^1]: Note that it is technically possible for the derived $ask$ or $nsk$ to be
+$0$, but this happens with probability approximately $2^{-252}$, so we ignore
+this case, as, borrowing phrasing from [Adam Langley][agl_elligator], it happens
+significantly less often than malfunctions in the CPU instructions we'd use to
+check it.
```
hdevalence, line 30: This replaces the previous language about discarding the keys; we don't do that, and it doesn't make sense to fit the API around the possibility.

spec/protocol/addresses_keys/spend_key.md
```diff
@@ -0,0 +1,32 @@
...
+* $ask$, the *spend authorization key*, a scalar value
+* $nsk$, the *nullifier private key*, a scalar value
+* $ovk$, the *outgoing viewing key*, a 32 byte number
+
+The scalars are derived by using the spend seed as the key for a PRF,
+expanding different labels, and converting the output to a `decaf377` scalar.
+In pseudocode this becomes:
+
+```
+prf_expand(x) = BLAKE2b_512("Penumbra_ExpandSeed", ss || x )
```
hdevalence, line 14: I split this out because it's clearer to think of it as a PRF, separate from the details of the construction.

crypto/src/merkle/tree.rs
```diff
@@ -203,10 +203,9 @@ mod tests {
         };
 
         let diversifier = Diversifier::generate(&mut rng);
-        let sk = SpendingKey::generate(&mut rng);
-        let expanded_sk = ExpandedSpendingKey::derive(&sk);
-        let proof_auth_key = expanded_sk.derive_proof_authorization_key();
-        let fvk = expanded_sk.derive_full_viewing_key();
+        let expanded_sk = SpendKey::generate(&mut rng);
+        let proof_auth_key = expanded_sk.proof_authorization_key();
+        let fvk = expanded_sk.full_viewing_key();
```
hdevalence, line 208: didn't update this beyond the compile fix since it's queued for alteration

crypto/src/keys/spend.rs
```diff
@@ -0,0 +1,123 @@
...
+    /// Loads a [`SpendKey`] from the provided [`SpendSeed`].
+    pub fn from_seed(seed: SpendSeed) -> Self {
+        let ask = {
+            let mut hasher = blake2b_simd::State::new();
+            hasher.update(b"Penumbra_ExpandSeed");
+            hasher.update(&seed.0);
+            hasher.update(&[0; 1]);
+            let hash_result = hasher.finalize();
+
+            // XXX: add Fr constructor to SigningKey
```
hdevalence, line 45: We should add a constructor to the `SigningKey` API in `decaf377_rdsa` that takes a field element as input.

redshiftzero (approved):
LGTM

PR #80, opened 2021-10-13
feat: transaction builder

redshiftzero:
This PR:
* adds a transaction builder (and a sketch of how Transactions would be created)
* adds more logic for the individual actions
* moves the fee in the tx proto from the actions enum to the tx
* reorganizes the `action` module 
* replaces `NoteCommitment::new()` with `Note::commit()` to match the API with Value commitments

(there are still some TODOs inline but am at a good pause point to merge this in)


crypto/Cargo.toml
```diff
@@ -17,5 +17,6 @@ ark-ff = { git = "https://github.com/penumbra-zone/algebra", branch = "ours" }
 ark-serialize = { git = "https://github.com/penumbra-zone/algebra", branch = "ours" }
 once_cell = "1.8"
 rand_core = { version = "0.6.3", features = ["getrandom"] }
+rand = "0.6.3"
```
hdevalence, line 20: Could this be a dev-dependency?

crypto/src/note.rs
```diff
@@ -24,36 +25,34 @@ static NOTECOMMIT_DOMAIN_SEP: Lazy<Fq> = Lazy::new(|| {
...
+        self.dest.diversifier()
+    }
 
-impl Commitment {
-    pub fn new(
-        dest: &PaymentAddress,
-        value: &Value,
-        note_blinding: &Fq,
-    ) -> Result<Self, ka::Error> {
+    pub fn commit(&self) -> Result<Commitment, ka::Error> {
```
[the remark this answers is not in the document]
hdevalence: (to expand a little: the issue was that a `ka::Public` doesn't necessarily represent a public key with a valid `s` coordinate, so I moved the validation into the `PaymentAddress` structure).

crypto/src/action/spend.rs
```diff
@@ -0,0 +1,80 @@
...
+        spend_key: SpendKey,
+        _merkle_path: MerklePath,
+        note: Note,
+    ) -> Self {
+        // TODO: Derive nullifier from note commitment, note position, and
+        // nullifier deriving key
+        // See p.55 ZCash spec
+        let nullifier = Nullifier::new();
+
+        let v_blinding = Fr::rand(&mut rng);
```
hdevalence, line 31: I think we're going to need to hold on to this value to be able to construct binding signatures.

crypto/src/constants.rs
```diff
@@ -0,0 +1,5 @@
+pub const MEMO_CIPHERTEXT_LEN_BYTES: usize = 528;
+pub const OVK_WRAPPED_LEN_BYTES: usize = 80;
+
+// This is the `MEMO_CIPHERTEXT_LEN_BYTES` - MAC size (16 bytes).
+pub const MEMO_LEN_BYTES: usize = 512;
```
hdevalence, line 5: do we want to have a dedicated, catch-all constants module, or should we define constants locally in the modules they relate to? currently we've been doing the latter, should we switch to the former?

crypto/src/action/spend.rs
```diff
@@ -0,0 +1,80 @@
...
+        let a = Fr::rand(&mut rng);
+        let rk = ask.randomize(&a).into();
+        let proof = SpendProof::new();
+        Body {
+            value_commitment,
+            nullifier,
+            rk,
+            proof,
+        }
+    }
```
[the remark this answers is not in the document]
hdevalence: I'm not sure that there's a use case for creating `Spend`s or `Output`s outside of the transaction builder?

crypto/src/tx.rs
```diff
@@ -0,0 +1,126 @@
...
+impl TransactionBuilder {
+    pub fn new() -> Self {
+        Self {
+            spends: Vec::<Spend>::new(),
+            outputs: Vec::<Output>::new(),
+            fee: 0,
+            // xx Per asset?
+            balance: 0,
+        }
+    }
```
hdevalence, line 35: I think this would be better as the body of an `impl Default`; we might not want `TransactionBuilder::new()` at all, and instead have a `Transaction::builder() -> TransactionBuilder { Default::default() }`, so that people don't need to import the builder type separately.

redshiftzero:
Discussed this and we'll merge this as is and then will address these valuable comments in a followup

crypto/src/constants.rs
redshiftzero: ah word, I'll remove the constants module and we can still with defining constants in the relevant module

crypto/Cargo.toml
hdevalence: Yeah, makes sense, just wondering!

PR #86, opened 2021-10-19
feat: domain types <-> proto conversions - part 1

redshiftzero:
~Will mark this ready once I've finished the other proto conversions~ Gonna do a part 2 since this is getting big

Immediate next steps to complete this part of the codebase (these have todos noted inline):
- [ ] Note encryption
- [ ] Nullifier derivation
- [ ] Memo encryption implemented
- [ ] Ovk wrapped key implemented

### DIFF
--- a/crypto/src/builder.rs
+++ b/crypto/src/builder.rs
@@ -11,7 +11,7 @@ use crate::{
     memo::MemoPlaintext,
     merkle,
     nullifier::Nullifier,
-    transaction::{Transaction, TransactionBody},
+    transaction::{Fee, Transaction, TransactionBody},
     Fr, Note, Output, Spend, Value,
 };
 
@@ -28,7 +28,7 @@ pub struct TransactionBuilder {
     // Actions we'll perform in this transaction.
     pub actions: Vec<Action>,
     // Transaction fee. None if unset.
-    pub fee: Option<u64>,
+    pub fee: Option<Fee>,
     // Sum of blinding factors for each value commitment.
     pub synthetic_blinding_factor: Fr,
     // The root of the note commitment merkle tree.
@@ -112,7 +112,7 @@ impl TransactionBuilder {
 
     /// Set the transaction fee in PEN.
     pub fn set_fee(mut self, fee: u64) -> Self {
-        self.fee = Some(fee);
+        self.fee = Some(Fee(fee));
         self
     }
 
--- a/crypto/src/transaction.rs
+++ b/crypto/src/transaction.rs
@@ -1,5 +1,7 @@
 use ark_ff::Zero;
 
+use penumbra_proto::{transaction, Protobuf};
+
 use crate::{
     action::Action,
     builder::TransactionBuilder,
@@ -13,7 +15,7 @@ pub struct TransactionBody {
     pub merkle_root: merkle::Root,
     pub expiry_height: u32,
     pub chain_id: String,
-    pub fee: u64,
+    pub fee: Fee,
 }
 
 impl TransactionBody {
@@ -21,6 +23,8 @@ impl TransactionBody {
         todo!()
     }
 }
+
+pub struct Fee(pub u64);
 
 // temp: remove dead code when Transaction fields are read
 #[allow(dead_code)]
@@ -40,6 +44,20 @@ impl Transaction {
             expiry_height: None,
             chain_id: None,
         }
+    }
+}
+
+impl Protobuf<transaction::Fee> for Fee {}
+
+impl From<Fee> for transaction::Fee {
+    fn from(fee: Fee) -> Self {
+        transaction::Fee { amount: fee.0 }
+    }
+}
+
+impl From<transaction::Fee> for Fee {
+    fn from(proto: transaction::Fee) -> Self {
+        Fee(proto.amount)
     }
 }
 


crypto/src/action/error.rs
```diff
@@ -0,0 +1,27 @@
+use thiserror;
+
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
```
hdevalence, line 4: Maybe this could be `ProtoError` or something, so that it's scoped to errors that can occur while deserializing?

crypto/src/action/output.rs
```diff
@@ -1,26 +1,80 @@
...
 
-use crate::{addresses::PaymentAddress, ka, memo::MemoPlaintext, note, value, Fq, Fr, Note, Value};
+use penumbra_proto::{transaction, Protobuf};
 
-pub const OVK_WRAPPED_LEN_BYTES: usize = 80;
+use crate::{
+    action::{
+        constants::{NOTE_ENCRYPTION_BYTES, OVK_WRAPPED_LEN_BYTES},
+        error::Error,
+    },
```
hdevalence, line 12: nit: I think it's slightly preferable to import stuff in the same module with (at most one level of) `super`, rather from the crate root (i.e., relative to where we are, rather than relative to the crate), since it means that if the module tree gets moved around, it's less likely that the imports will need changing.

crypto/src/action/output.rs
```diff
@@ -1,26 +1,80 @@
...
+
+impl TryFrom<transaction::Output> for Output {
+    type Error = Error;
+
+    fn try_from(proto: transaction::Output) -> Result<Self, Self::Error> {
+        if proto.body.is_none() {
+            return Err(Error::ProtobufMissingFieldError);
+        }
+
+        let body = Body::try_from(proto.body.unwrap());
```
hdevalence, line 47: Rather than unwrapping here, it could be merged with the preceding

```
let body = proto.body.ok_or(Error::OutputBodyMalformed)?
    .try_into()
    .map_err(Error::OutputBodyMalformed)?;
```

crypto/src/action/output.rs
```diff
@@ -38,14 +92,69 @@ impl Body {
...
+            zkproof: Bytes::copy_from_slice(&proof_bytes),
+        }
+    }
+}
+
+impl TryFrom<transaction::OutputBody> for Body {
+    type Error = Error;
+
+    fn try_from(proto: transaction::OutputBody) -> Result<Self, Self::Error> {
+        let value_commitment: value::Commitment = (proto.cv[..])
```
hdevalence, line 134: since these types are specified by the fields of `Body`, the type ascription isn't necessary to constrain the `try_into()` call, and they could even be inlined into the `Body { ... }` expression

crypto/src/builder.rs
```diff
@@ -13,18 +15,30 @@ use crate::{
     Fr, Note, Output, Spend, Value,
 };
 
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error("Chain ID not set")]
+    NoChainID,
+    #[error("Expiry height not set")]
+    ExpiryHeightNotSet,
```
hdevalence, line 23: The expiry height should probably actually be optional, it's fine for people not to set it.

crypto/src/builder.rs
```diff
@@ -107,13 +118,40 @@ impl TransactionBuilder {
...
+            return Err(Error::NoChainID);
+        }
+
+        if self.expiry_height.is_none() {
+            return Err(Error::ExpiryHeightNotSet);
+        }
+
+        if self.fee.is_none() {
+            return Err(Error::FeeNotSet);
+        }
```
hdevalence, line 147: We should do this later, but I think it would probably be good to assert that the transaction balances as constructed, rather than waiting for its verification to fail.

We could do this by keeping a vec of `Value`s in the builder, then using a `BTreeMap` or something to sum up per-id amounts and checking balances.

crypto/src/transaction.rs
```diff
@@ -15,17 +22,23 @@ impl TransactionBody {
...
+    binding_sig: Signature<Binding>,
+}
 
 impl Transaction {
     /// Start building a transaction relative to a given [`merkle::Root`].
     pub fn build_with_root(merkle_root: merkle::Root) -> TransactionBuilder {
         TransactionBuilder {
-            spends: Vec::<Spend>::new(),
-            outputs: Vec::<Output>::new(),
+            actions: Vec::<Action>::new(),
```
hdevalence, line 36: This can just be `Vec::new()`; the inner type is already constrained by the `TransactionBuilder` definition.

decaf377-ka/src/lib.rs
```diff
@@ -27,6 +28,8 @@ pub struct SharedSecret(pub [u8; 32]);
 pub enum Error {
     #[error("Invalid public key")]
     InvalidPublic(Public),
+    #[error("Cannot deserialize public key")]
+    PublicKeyDeserializationError,
```
hdevalence, line 32: Could this just be `SliceLenError`, since that's the only failure case of `TryFrom`?

hdevalence:
Looking at the proto validations, there's a lot of work creating custom error types with `thiserror`.  Enum errors that completely model the domain are useful if the caller might want to do control flow based on the error variant.  But in the case of proto validation, it seems mostly useful to just error or not, and to then report the cause of the failure.  So I think it would be a lot less work and less code to maintain to use `anyhow::Error` for the proto conversions, rather than defining a lot of proto-specific errors.

crypto/src/action/error.rs
hdevalence: superseded by https://github.com/penumbra-zone/penumbra/pull/86#issuecomment-947255729

PR #90, opened 2021-10-21
doc: add more crates to GH Actions

hdevalence:


hdevalence:
(merging immediately because there's currently no other way to test this part of the CI)

PR #91, opened 2021-10-21
Begin rework of key hierarchy design

hdevalence:
This begins reworking the key hierarchy to incorporate some ideas from Orchard.
Updating the design documents is still TODO, but:

- the proof authorizing key is removed, as in Orchard;
- diversifiers are generated with a diversifier key, as in Orchard;
- the OVK is derived from the FVK, as in Orchard;
- the IVK is derived as the output of a PRF keyed by nk on ak (unlike Orchard, we
  are ok with relying on Poseidon as a PRF);
- detection keys are derived as the output of a PRF keyed by ivk on the diversifier;
- addresses include clue keys.

This PR is structured as follows:

- preamble commits that fix issues required to do the reworking;
- the reworking itself;
- some opportunistic cleanup I spotted while looking at the rendered docs post-reworking;

so it's probably most clear going commit-by-commit.


crypto/src/address.rs
```diff
@@ -0,0 +1,71 @@
...
+    /// of an [`Fq`] `s` value.
+    pub(crate) fn from_components(
+        d: Diversifier,
+        g_d: decaf377::Element,
+        pk_d: ka::Public,
+        ck_d: fmd::ClueKey,
+    ) -> Option<Self> {
+        // XXX ugly -- better way to get our hands on the s value?
+        // add to decaf377::Encoding? there's compress_to_field already...
+        if let Ok(cached_s) = Fq::deserialize(&pk_d.0[..]) {
```
redshiftzero, line 37: oh next time we edit this part of the codebase we can use the `FieldExt` trait here now that we have it implemented for `Fq`

PR #93, opened 2021-10-21
feat: derive nullifiers for positioned notes

redshiftzero:


crypto/src/note.rs
```diff
@@ -60,6 +66,13 @@ impl Note {
 
         Commitment(commit)
     }
+
+    pub fn nf(&self, pos: Position, nk: &keys::NullifierKey) -> Nullifier {
+        Nullifier(hash_3(
+            &NULLIFIER_DOMAIN_SEP,
+            (nk.0, self.commit().0, (u64::from(pos)).into()),
+        ))
+    }
```
hdevalence, line 75: this might be cleaner as a method on `NullifierKey` (and possibly a convenience method on `FullViewingKey`)

PR #103, opened 2021-10-27
feat: client state

### LINKED ISSUE #34: Implement client state
As explained in the first part of #28, the state on a shielded blockchain like Penumbra is shaped differently than on a public one:

> On a shielded blockchain, however, the state is fragmented across all users of the application, as each user has a view only of their "local" portion of the application state. Transactions update a user's state privately, and use a zero-knowledge proof to prove to all other participants that the update was allowed by the application rules.

This means that the client library has to maintain client state. This probably looks like:

* a set of viewing keys
* a record of how much of the chain has been scanned with each viewing key
* a set of live notes under the user's control
* a record of transactions they have visibility into

Some questions:

- what should back this data store? ideally, it'd be something that's storage-agnostic, so it could just as easily work for `pcli` as any other environment that allows blob storage.

redshiftzero:
Some initial work for client state storage & adds a separate crate for pcli/the wallet software

Closes #34


hdevalence (approved):
LGTM, happy to merge this now and keep iterating on it

PR #106, opened 2021-10-28
note serialization/deserialization

redshiftzero:
and fixes to spec/code comments re: size of fields

### DIFF
--- a/crypto/src/keys/diversifier.rs
+++ b/crypto/src/keys/diversifier.rs
@@ -1,7 +1,9 @@
 use aes::Aes256;
+use anyhow::anyhow;
 use ark_ff::PrimeField;
 use fpe::ff1;
 use once_cell::sync::Lazy;
+use std::convert::TryFrom;
 
 use crate::Fq;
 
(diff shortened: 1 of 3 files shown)


crypto/src/keys/diversifier.rs
```diff
@@ -25,12 +27,29 @@ impl Diversifier {
...
     }
 }
 
+impl TryFrom<&[u8]> for Diversifier {
+    type Error = anyhow::Error;
+
+    fn try_from(slice: &[u8]) -> Result<Diversifier, Self::Error> {
+        if slice.len() != DIVERSIFIER_LEN_BYTES {
+            return Err(anyhow!(
+                "diversifier must be 11 characters, got {:?}",
```
hdevalence, line 42:
```suggestion
                "diversifier must be 11 bytes, got {:?}",
```

hdevalence (approved):
LGTM

crypto/src/keys/diversifier.rs
redshiftzero: ahh good catch

PR #107, opened 2021-10-29
feat: Note, memo, ovk encryption/decryption

### LINKED ISSUE #99: Implement note, memo, and OVK encryption
There are three fields in our transactions that still need to be implemented:
1. The encrypted note in the output body
2. The (wrapped) outgoing viewing key in the output
3. The encrypted memo in the output

redshiftzero:
Closes #99 

This is using `AEAD_CHACHA20_POLY1305` as in Sapling/Orchard note encryption (ref: p.64-65 spec). One difference for Penumbra is since our memos are separated from the note, the memo is encrypted separately, but using the same method (i.e. same key, symmetric cipher). I'm using a unique nonce from the Note encryption (notes use `[0u8; 12]`, memos are using `[1u8; 12]`) but let me know if this is not the best choice. Another option could be to make further modifications to the Memo encryption/decryption.

### DIFF
--- a/crypto/src/memo.rs
+++ b/crypto/src/memo.rs
@@ -34,10 +34,13 @@ impl MemoPlaintext {
             .key_agreement_with(&transmission_key)
             .expect("key agreement success");
 
-        let mut kdf_input = Vec::new();
-        kdf_input.copy_from_slice(&shared_secret.0);
-        kdf_input.copy_from_slice(&epk.0);
-        let kdf_output = blake2b_simd::blake2b(&kdf_input);
+        // Use Blake2b-256 to derive encryption key.
+        let mut kdf_params = blake2b_simd::Params::new();
+        kdf_params.hash_length(32);
+        let mut kdf = kdf_params.to_state();
+        kdf.update(&shared_secret.0);
+        kdf.update(&epk.0);
+        let kdf_output = kdf.finalize();
         let key = Key::from_slice(kdf_output.as_bytes());
 
         let cipher = ChaCha20Poly1305::new(key);
@@ -57,3 +60,32 @@ impl MemoPlaintext {
 
 #[derive(Clone)]
 pub struct MemoCiphertext(pub [u8; MEMO_CIPHERTEXT_LEN_BYTES]);
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    use crate::keys::SpendKey;
+    use rand_core::OsRng;
+
+    #[test]
+    fn test_memo_encryption_and_decryption() {
+        let mut rng = OsRng;
+
+        let sk = SpendKey::generate(&mut rng);
+        let fvk = sk.full_viewing_key();
+        let ivk = fvk.incoming();
+        let (dest, _dtk_d) = ivk.payment_address(0u64.into());
+
+        let mut memo_bytes = [0u8; MEMO_LEN_BYTES];
+        memo_bytes[0..2].copy_from_slice(b"Hi");
+
+        let esk = ka::Secret::new(&mut rng);
+
+        let memo = MemoPlaintext(memo_bytes);
+
+        let _ciphertext = memo.encrypt(&esk, dest.transmission_key(), dest.diversified_generator());
+
+        // TODO: Decryption
+    }
+}
--- a/crypto/src/memo.rs
+++ b/crypto/src/memo.rs
@@ -3,6 +3,7 @@ use chacha20poly1305::{
     aead::{Aead, NewAead},
     ChaCha20Poly1305, Key, Nonce,
 };
+use once_cell::sync::Lazy;
 use std::convert::TryInto;
 
 use crate::{ka, keys::IncomingViewingKey};
@@ -11,6 +12,12 @@ pub const MEMO_CIPHERTEXT_LEN_BYTES: usize = 528;
 
 // This is the `MEMO_CIPHERTEXT_LEN_BYTES` - MAC size (16 bytes).
 pub const MEMO_LEN_BYTES: usize = 512;
+
+/// The nonce used for memo nonce encryption.
+pub static MEMO_ENCRYPTION_NONCE: Lazy<[u8; 12]> = Lazy::new(|| {
+    let nonce_bytes = 1u128.to_le_bytes();
+    nonce_bytes[0..12].try_into().expect("nonce fits in array")
+});
 
 // The memo is stored separately from the `Note`.
 #[derive(Clone, Debug, PartialEq, Eq)]
@@ -45,7 +52,7 @@ impl MemoPlaintext {
         let key = Key::from_slice(kdf_output.as_bytes());
 
         let cipher = ChaCha20Poly1305::new(key);
-        let nonce = Nonce::from_slice(&[1u8; 12]);
+        let nonce = Nonce::from_slice(&*MEMO_ENCRYPTION_NONCE);
 
         let encryption_result = cipher
             .encrypt(nonce, self.0.as_ref())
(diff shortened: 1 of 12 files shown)


hdevalence:
I think using `[1,0,...]` (can be instantiated as `1u128.to_le_bytes()`) might be better than all 1s, just because it's the sequentially next value, but it doesn't matter a great deal.

crypto/src/memo.rs
```diff
@@ -16,12 +23,107 @@ impl Default for MemoPlaintext {
...
+    pub fn encrypt(
+        &self,
+        esk: &ka::Secret,
+        transmission_key: &ka::Public,
+        diversified_generator: &decaf377::Element,
+    ) -> Result<MemoCiphertext, anyhow::Error> {
+        let epk = esk.diversified_public(diversified_generator);
+        let shared_secret = esk
+            .key_agreement_with(&transmission_key)
+            .map_err(|_| anyhow!("could not perform key agreement"))?;
```
hdevalence, line 36: Could this take an address type or something, so that we know the transmission key is valid, and we can treat the key agreement as infallible?

crypto/src/memo.rs
```diff
@@ -16,12 +23,107 @@ impl Default for MemoPlaintext {
...
+        let cipher = ChaCha20Poly1305::new(key);
+        let nonce = Nonce::from_slice(&[1u8; 12]);
+
+        let encryption_result = cipher
+            .encrypt(nonce, self.0.as_ref())
+            .map_err(|_| anyhow!("encryption error!"))?;
+
+        let ciphertext: [u8; MEMO_CIPHERTEXT_LEN_BYTES] = encryption_result
+            .try_into()
+            .map_err(|_| anyhow!("memo encryption result does not fit in ciphertext len"))?;
```
hdevalence, line 56: When would this happen? We know the size of the memo plaintext, so we know the size of the ciphertext, right?

Could this be an `.expect` instead?

crypto/src/memo.rs
```diff
@@ -16,12 +23,107 @@ impl Default for MemoPlaintext {
...
+        kdf.update(&epk.0);
+        let kdf_output = kdf.finalize();
+        let key = Key::from_slice(kdf_output.as_bytes());
+
+        let cipher = ChaCha20Poly1305::new(key);
+        let nonce = Nonce::from_slice(&[1u8; 12]);
+
+        let encryption_result = cipher
+            .encrypt(nonce, self.0.as_ref())
+            .map_err(|_| anyhow!("encryption error!"))?;
```
hdevalence, line 52: Could this also be infallible? I looked at the docs and it seems like it's a generic method, maybe the error is there only generically? https://docs.rs/aead/0.4.3/aead/trait.Aead.html#tymethod.encrypt

hdevalence: It looks like the generic implementation here https://docs.rs/aead/0.4.3/src/aead/lib.rs.html#360-370 creates a new buffer and then calls `encrypt_in_place` https://docs.rs/aead/0.4.3/aead/trait.AeadInPlace.html#method.encrypt_in_place which errors only if the buffer is wrongly sized.

So the method is infallible, but you can't tell from the types because there's always a generic `Error` type.  It'd be neat if there were a way for the generic impl to specify `Error = !`, but it looks like the crate hardcodes an opaque error type in an attempt to prevent error oracles.

hdevalence: So, we could safely `.unwrap()` or `.expect()` this away, and combined with the other two comments, we could make the whole function infallible.

crypto/src/memo.rs
```diff
@@ -16,12 +23,107 @@ impl Default for MemoPlaintext {
...
 
 impl MemoPlaintext {
-    pub fn encrypt(&self, _key: &Address) -> MemoCiphertext {
-        // TODO!
-        let memo_bytes = [0u8; MEMO_CIPHERTEXT_LEN_BYTES];
-        MemoCiphertext(memo_bytes)
+    /// Encrypt a memo, returning its ciphertext.
+    pub fn encrypt(
+        &self,
+        esk: &ka::Secret,
```
hdevalence, line 36: `esk` is supposed to be used ephemerally, but we need to use it twice, once for the memo encryption and once for the note encryption. I'm not sure if this would be better or worse, but would it make sense to try to join those together, so that the API is harder to misuse?

crypto/src/note.rs
```diff
@@ -47,6 +55,10 @@ pub enum Error {
     NoteTypeUnsupported,
     #[error("Note deserialization error")]
     NoteDeserializationError,
+    #[error("Encryption error")]
+    EncryptionError,
```
hdevalence, line 59: Ideally we could have all encryptions be infallible and avoid this error variant.

crypto/src/note.rs
```diff
@@ -90,6 +102,111 @@ impl Note {
         self.value
     }
 
+    /// Encrypt a note, returning its ciphertext.
+    pub fn encrypt(&self, esk: &ka::Secret) -> Result<[u8; NOTE_CIPHERTEXT_BYTES], Error> {
+        let epk = esk.diversified_public(&self.diversified_generator());
+        let shared_secret = esk
+            .key_agreement_with(&self.transmission_key())
+            .map_err(|_| Error::EncryptionError)?;
```
hdevalence, line 110: same comments as the memo encryption case in re: fallibility.

crypto/src/note.rs
```diff
@@ -90,6 +102,111 @@ impl Note {
...
+
+        // Use Blake2b-256 to derive encryption key.
+        let mut kdf_params = blake2b_simd::Params::new();
+        kdf_params.hash_length(32);
+        let mut kdf = kdf_params.to_state();
+        kdf.update(&shared_secret.0);
+        kdf.update(&epk.0);
+        let kdf_output = kdf.finalize();
+        let key = Key::from_slice(kdf_output.as_bytes());
```
hdevalence, line 119: if this is common with the note encryption, could it be extracted as a standalone function?

crypto/src/note.rs
```diff
@@ -90,6 +102,111 @@ impl Note {
...
+        let cipher = ChaCha20Poly1305::new(ock);
+        let nonce = Nonce::from_slice(&[0u8; 12]);
+
+        let encryption_result = cipher
+            .encrypt(nonce, op.as_ref())
+            .map_err(|_| Error::EncryptionError)?;
+
+        let wrapped_ovk: [u8; OVK_WRAPPED_LEN_BYTES] = encryption_result
+            .try_into()
+            .map_err(|_| Error::EncryptionError)?;
```
hdevalence, line 172: ditto fallibility

hdevalence (changes requested):
This looks great, but it would be good to see if we can avoid fallibility for encryptions.

crypto/src/note.rs
```diff
@@ -204,3 +340,42 @@ impl TryFrom<&[u8]> for Commitment {
...
+        .expect("can create note");
+        let esk = ka::Secret::new(&mut rng);
+
+        let ciphertext = note.encrypt(&esk).expect("can encrypt note");
+
+        let epk = esk.diversified_public(&dest.diversified_generator());
+        let plaintext = Note::decrypt(ciphertext, ivk, &epk).expect("can decrypt note");
+
+        assert_eq!(plaintext, note);
+    }
```
hdevalence, line 380: In this test I think it'd be useful to generate `ivk2` (from a totally unrelated spend key) and check that decrypting the note with the wrong IVK gives an error.

hdevalence: (motivation: then we know that we can use note decryption for scanning)

redshiftzero:
Thanks for these comments, I've added a commit making encryption infallible and addressing the other items (except the API change to combine memo, note encryption into a single method)

hdevalence:
Sounds good, that's definitely something that could be for later.

hdevalence (approved):

PR #111, opened 2021-11-02
Stub Postgres Scaffolding 

### LINKED ISSUE #109: Stub Postgres scaffolding
In #21, we tentatively settled on Postgres as a data store for the Penumbra daemon `pd`.

We should have some stub code that integrates Postgres and does some really basic load/store, so that we can start extending it and saving data. This stub code should:

- [x] add Postgres to the existing `docker-compose` setup;
- [x] have some way of attaching a persistent docker volume for the postgres state;
- [x] creates a single table recording note commitment tree anchors
  - This table should have two columns, a `u64` height and a `[u8; 32]` anchor.
  - On the Rust side, the height should correspond to a tendermint [`block::Height`](https://rustdoc.penumbra.zone/main/tendermint/block/struct.Height.html) and the anchor should correspond to a [`merkle::Root`](https://rustdoc.penumbra.zone/main/penumbra_crypto/merkle/struct.Root.html).  (The `merkle::Root` can be converted to and from bytes using the inner `Fq`'s `to_bytes` and `from_bytes` methods from the [`FieldExt`](https://rustdoc.penumbra.zone/main/decaf377/trait.FieldExt.html) trait).
- [x] add a stub function that gets called somewhere in the `pd` [`main` function](https://github.com/penumbra-zone/penumbra/blob/main/penumbra/src/bin/pd.rs#L20) that writes a hardcoded entry to the table;
- [x] add a stub function that gets called somewhere in the `main` function that does a hardcoded query and converts back into Rust types.
  - For these two functions it's not essential to organize the code nicely, since we'll be rewriting the entire binary anyways, and we'll need to reorganize the code as part of that work - the goal is just to get something going.

amanjeev:
This is the proof of concept for adding a database, bootstrapping it with one table, and doing dummy transactions.

This work is not ideal but is a quick PR to add the feature.
Also, there are no tests. 

To check this out - 

1. Checkout this branch
2. Kill all previous docker containers and delete docker volumes and images
3. Re-run `docker-compose up -d`
4. In another terminal see `docker-compose logs -f`
5. You will see log of the `println`

Alternatively, 

1. Checkout this branch
2. Kill all previous docker containers and delete docker volumes and images
3. Re-run `docker-compose up -d`
4. `exec` into docker container `db` - `docker exec -it db bash`
5. Run psql to see database and single row  
    - `psql -U postgres`
    - `\c penumbra`
    - `select * from note_commitment_tree_anchors;`

Closes #109

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,7 @@ Cargo.lock
 
 # MSVC Windows builds of rustc generate these, which store debugging information
 *.pdb
+
+# editor
+.swp
+.idea
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,7 @@
-FROM rust:1.54.0 as build
+FROM rust:1.56 as build
+
+RUN curl https://sh.rustup.rs -sSf | sh -s -- --no-modify-path --default-toolchain none -y
+RUN rustup component add rustfmt
 
 WORKDIR /usr/src
 COPY . .
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,6 +20,8 @@ RUN cargo build --bin pd --release --frozen && \
 # with a statically linked libc (read: musl), and musl's malloc exhibits
 # pathologically poor performance for Tokio applications...
 FROM debian:buster-slim as runtime
+ARG DATABASE_URL
+ENV DATABASE_URL=$DATABASE_URL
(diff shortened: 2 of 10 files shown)


hdevalence:
This is great, thanks!!

Issue #81, opened 2021-10-16
Update protocol spec to match `decaf377-fmd` work


hdevalence:
The `decaf377-fmd` implementation added in #79 should be reconciled with the existing protocol spec contents:

- the terminology in the spec should be updated to match the terminology in the code (clue, address, detection key);
- the section on compact keys should be cut out and moved to its own subsection;
- the section on diversified detection should be cut out and moved to a separate subsection, where it's clear that it's not actually used in Penumbra;
- some content should be added explaining the context that FMD will be used, hinting at how we can determine parameter choices (since a fair amount of existing protocol analysis doesn't apply to our setting).

Also:

- the `decaf377-fmd` docs should have a brief overview of the purpose of the crate.

hdevalence:
Added to MVP1 because it's easier to clean up the spec while it's small.

hdevalence:
This is done except for the last point, which could be its own issue

PR #114, opened 2021-11-02
feat: Bech32m encoding/decoding, set first testnet ID

### LINKED ISSUE #63: Specify address encodings
Followup from #61

Related reading:
* p. 112 of Zcash protocol spec
* ZIP-173 defining Bech32 encoding https://zips.z.cash/zip-0173
* ZIP-316 defining unified addresses https://zips.z.cash/zip-0316

redshiftzero:
Addresses now encode like: `penumbra_tn001_1kpgdhlzws6kyk2cf580wtt76t9nn2vf7em3pn05y3h8ym5a6aevdxshjgsnxecv94rzsxdhng6cjp8kgchqxud06p9xka0yxv99rty3njetqqnx2hrzz4tc03956e0`

Closes #63

### DIFF
--- a/crypto/src/address.rs
+++ b/crypto/src/address.rs
@@ -1,13 +1,16 @@
-use std::io::Write;
+use std::io::{Cursor, Read, Write};
 
+use anyhow::anyhow;
 use ark_serialize::CanonicalDeserialize;
-use bech32::{ToBase32, Variant};
+use bech32::{FromBase32, ToBase32, Variant};
 use decaf377;
 
 use crate::{fmd, ka, keys::Diversifier, Fq};
 
+pub const CURRENT_NETWORK_ID: &str = "penumbra_tn001_";
+
 /// A valid payment address.
-#[derive(Clone)]
+#[derive(Clone, Debug, PartialEq, Eq)]
 pub struct Address {
     d: Diversifier,
     /// cached copy of the diversified base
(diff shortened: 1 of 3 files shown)


crypto/src/address.rs
```diff
@@ -64,3 +67,44 @@ impl Address {
...
+            .expect("can write diversifier into vec");
+        addr_content
+            .write_all(&self.transmission_key().0)
+            .expect("can write transmission key into vec");
+        addr_content
+            .write_all(&self.clue_key().0)
+            .expect("can write clue key into vec");
+
+        // Mainnet "pm" prefix, testnet "pt" prefix
+        let human_part = "pt";
```
hdevalence, line 85: Cosmos uses `cosmos` which is a bit longer but I think reads more nicely -- WDYT of `penumbra` for the human-readable part? The addresses are long enough already that I don't think a few extra characters is going to make a big difference.

For the testnets, I think it would be useful to encode the testnet number in the address, like `penumbra-tn001-`, `penumbra-tn002-`, etc (not sure about the exact format, but 3 digits seems good, and having a trailing dash separates the number from the bech32 `1` separator).

redshiftzero: Yeah it seems like underscores have better double-click to select behavior so going with that (for mainnet we wont have the suffix so can just do `penumbra` and this won't be an issue)

spec/protocol/addresses_keys/addresses.md
```diff
@@ -19,3 +19,8 @@ $pk_d = [ivk]B_d$
 The *compact flag key* is derived from a (diversified) *compact detection key* $cdtk_d$ using a non-diversified `decaf377` basepoint [$B$](../primitives/decaf377/test_vectors.md):
 
 $cfk_d = [cdtk_d]B$
+
+The diversifier, transmission key, and compact flag key are [Bech32m](https://github.com/bitcoin/bips/blob/master/bip-0350.mediawiki) encoded with a human readable prefix that is:
```
redshiftzero, line 23: while this is not the nomenclature that we're currently using for FMD, I kept this as is to avoid inconsistency/confusion until #81

hdevalence (approved):
LGTM

PR #120, opened 2021-11-03
feat(pcli): add `generate` command, save/load wallet data in/from `penumbra_wallet.dat`

redshiftzero:
This PR:
* Adds a `generate` command for generating keys
* Stores key material once generated in `penumbra_wallet.dat` (or alternatively a custom location)
* Loads and saves key material using a minimal (minimal bc this is what we serialize) `Wallet` struct that has a spend key seed and the last diversifier index in use
* Integrates https://docs.rs/directories/4.0.1/directories/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-wallet.dat
+*wallet.dat
 
 book
 .firebase/
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
+wallet.dat
+
 book
 .firebase/
 # For some reason mdbook dumps this here because the main page is ../README.md
--- a/README.md
+++ b/README.md
@@ -41,12 +41,12 @@ cargo run --bin pcli -- --help
 Generate keys using:
 
 ```
-$ cargo run --bin pcli -- --wallet-location penumbra_wallet.dat generate
-Wallet generated, stored in penumbra_wallet.dat. WARNING: This file contains your private keys. BACK UP THIS FILE!
+$ cargo run --bin pcli -- generate
+Wallet generated, stored in /Users/Alice/Library/Application Support/zone.penumbra.pcli/penumbra_wallet.dat. WARNING: This file contains your private keys. BACK UP THIS FILE!
 Your first address is penumbra_tn001_10ftlcft6c8n95cc5lpwdupt2d86n0td2t55xjwkgnte9jgzlqfp7xlqfnssljygq6fxspvnj5xuc5j0qtd6j398elyhrqugs07r7s8v5n0dpkgweytjqm2gv6hmdef
 ```
 
-Keys will be stored in the location you provide via `--wallet-location <path>` or if you specify no location, in `pcli`'s data directory:
+Keys will be stored in `pcli`'s data directory:
 
 * Linux: `/home/alice/.config/pcli/penumbra_wallet.dat`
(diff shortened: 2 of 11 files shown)


wallet/src/storage.rs
```diff
@@ -0,0 +1,11 @@
+use serde::{Deserialize, Serialize};
+
+use penumbra_crypto::keys;
+
+/// The contents of the wallet file that share a spend authority.
+#[derive(Serialize, Deserialize)]
+pub struct Wallet {
+    pub spend_seed: keys::SpendSeed,
+    // Store the last diversifier index in use.
+    pub last_used_diversifier_index: keys::DiversifierIndex,
```
redshiftzero, line 10: I was on the fence about making this `Vec<DiversifierIndex>` but ended up going with storing only the max diversifier index in use since it's smaller and the point of the `DiversifierIndex` was to have a nice way to generate addresses iirc

.gitignore
```diff
@@ -1,3 +1,5 @@
+*wallet.dat
```
chc4, line 1: This should be changed with the wallet file rename too

chc4: ...Except it already has a star, woops!

wallet/src/storage.rs
hdevalence: We probably want a way to track addresses, so maybe we should instead store a `Vec<String>` of labels, with `address_label[0]` having index `0`, etc?

redshiftzero: By storing the max `DiversifierIndex` in use, we can compute all the addresses in use via the `address_by_index` method on `ClientState` (deleted in the most recent commit), but it's more straightforward to store the encoded addresses, so I've made this change in a commit. The only downside is the wallet is no longer constant space but space is cheap

README.md
```diff
@@ -38,6 +38,20 @@ Now you can interact with Penumbra using `pcli`:
...
+$ cargo run --bin pcli -- --wallet-location penumbra_wallet.dat generate
+Wallet generated, stored in penumbra_wallet.dat. WARNING: This file contains your private keys. BACK UP THIS FILE!
+Your first address is penumbra_tn001_10ftlcft6c8n95cc5lpwdupt2d86n0td2t55xjwkgnte9jgzlqfp7xlqfnssljygq6fxspvnj5xuc5j0qtd6j398elyhrqugs07r7s8v5n0dpkgweytjqm2gv6hmdef
+```
+
+Keys will be stored in the location you provide via `--wallet-location <path>` or if you specify no location, in `pcli`'s data directory:
+
+* Linux: `/home/alice/.config/pcli/penumbra_wallet.dat`
+* macOS: `/Users/Alice/Library/Application Support/zone.penumbra.pcli/penumbra_wallet.dat`
+* Windows: `C:\Users\Alice\AppData\Roaming\penumbra\pcli\penumbra_wallet.dat`
```
hdevalence, line 53: Looking at this I worry that someone thinks that they'll have to specify `--wallet-location`, set the wallet in a nonstandard location, and then have to specify it every time.  I'm wondering if we should either (a) not include it in the instructions, falling through to the default, or (b) not support multiple wallet locations at all.

redshiftzero: Hmm true, let me remove it from the instructions (I feel like the option is useful itself but if we later decide it isn't we can remove it too), will remove this in a commit

PR #122, opened 2021-11-05
fix pd start command

chc4:
The docker command will just loop fail, since the proper command is `pd start --host <ip>`. I also had to add a version entry to the override yaml in order to start it.

I personally had to change it to version 3.7, since that's the latest my Docker package supports (Ubuntu), but I wasn't sure if you cared about supporting that or were using 3.9 for a specific reason, so I didn't include that in this.

### DIFF
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -1,3 +1,4 @@
+version: "3.9"
 # docker-compose oerrides for local development.
 #
 # this modifies the docker-compose configuration for locally testing and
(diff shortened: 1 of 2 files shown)


docker-compose.override.yml
```diff
@@ -1,3 +1,4 @@
+version: "3.9"
```
hdevalence, line 1:
```suggestion
version: "3.7"
```

This is the change to the version you needed to get it to work right?

hdevalence (approved):
Thanks for the PR! No, we don't currently have any particular opinions about the Docker version, so happy to change it.

docker-compose.override.yml
chc4: The version in the non-override docker compose also needs to be changed, but yes.

hdevalence: To be fixed in #128

PR #123, opened 2021-11-07
Spec refactor (part 2)

hdevalence:
some more improvements and refactorings (see comments).

Also, redo the proposed governance mechanism to allow concurrent proposals.


spec/concepts/governance.md
```diff
@@ -1,14 +1,107 @@
...
+decrypt the tallies.  These intermediate tallies are revealed, because it is not
+possible to batch value flows over time intervals longer than one epoch.  In
+practice, this provides a similar dynamic as existing (transparent) on-chain
+governance schemes, where tallies are public while voting is ongoing.
+
+At the end of the voting period, the per-epoch tallies are summed.  For each
+validator $v$, the votes for each option are summed to determine the portion of
+the delegation pool that voted; the validator's vote acts as the default vote
+for the rest of the delegation pool.  Finally, these per-validator subtotals are
+multiplied by the voting power adjustment function $\theta_v(e)$ to obtain the
```
redshiftzero, line 98: clarifying question: is `$\theta_v(e)$` just a function that multiplies in the weights of each validator? or is more happening here?

redshiftzero (approved):
LGTM!

Also, looks like the image we link to the readme is getting removed here, but it's already broken on `main` for me so mergin

spec/concepts/governance.md
hdevalence: fixed in #124

PR #128, opened 2021-11-09
Set Docker version to 3.7

hdevalence:
cf https://github.com/penumbra-zone/penumbra/pull/122#discussion_r744288524


redshiftzero (approved):
Ty!

PR #132, opened 2021-11-10
Rename the `pcli --addr` to `pcli --node`

### LINKED ISSUE #127: Wallet: Rename `--addr` to `--node`
Currently on `main` we have two `pcli` options which are too similar:
* `pcli addr` which is a subcommand that manages the diversified addresses
* `pcli --addr` which passes in the address of the tendermint node

Proposed fix is to rename `--addr` to `--node`

kwf:
Fixes #127.


redshiftzero (approved):
LGTM!

PR #136, opened 2021-11-10
Delete wallet command

### LINKED ISSUE #125: Wallet: command to delete wallet permanently
We want to default to a model where the wallet is frequently deleted and re-instantiated. To aid in testing this scenario, let's make a subcommand of the `wallet` noun that deletes the wallet. Note that this does not delete the key material, which can be used to re-instantiate the wallet.

Proposed syntax: `pcli wallet delete`.

kwf:
Fixes #125.

For reviewers: should any of the paths out of this command have a non-zero exit code?

### DIFF
--- a/penumbra/Cargo.toml
+++ b/penumbra/Cargo.toml
@@ -28,7 +28,6 @@ bincode = "1.3.3"
 blake2b_simd = "0.5"
 bytes = "1"
 comfy-table = "5"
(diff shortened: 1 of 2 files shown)


redshiftzero (approved):
looks great!

redshiftzero:
oh hmm, looks like there's a conflict

PR #146, opened 2021-11-13
Refactor state handling

hdevalence:
~~This is a slightly evolved version of the scratch work from this morning's session with @redshiftzero and @kwf; it's not complete but it's at least compileable.~~

This pulls the state handling out into a `State` structure that allows async access to the database state.  Also, it adds a `CompletionTracker` that allows us to correctly process requests asynchronously (i.e., avoiding `block_on`) while retaining the ability to specify that certain requests should fully complete before processing new ones.


README.md
```diff
@@ -87,9 +87,15 @@ export RUST_LOG=debug  # bash
...
+You'll need to set up a Postgres instance.  Here is one way:
+```bash
+# create a volume for pg data
+docker volume create tmp_db_data
+docker run --name tmp-postgres -e POSTGRES_PASSWORD=postgres -e POSTGRES_USER=postgres -e POSTGRES_DB=penumbra -p 5432:5432 -v tmp_db_data:/var/lib/postgresql/data -d postgres
+
 Start the Penumbra instance (you probably want to set `RUST_LOG` to `debug`):
 ```
-cargo run --bin pd
+cargo run --bin pd start -d postgres_uri
```
hdevalence, line 98: `pd` now takes the Postgres URI as a parameter instead of an environment variable -- I think we should try to avoid using any environment variables at all, since they end up acting as a kind of hidden shadow config system.

crypto/Cargo.toml
```diff
@@ -35,3 +36,5 @@ rand = "0.8"
 chacha20poly1305 = "0.9.0"
 # only needed because ark-ff doesn't display correctly
 num-bigint = "0.4"
+# this allows us to implement Decode for crypto types without orphan rule issues
+sqlx = { version = "0.5.9", features = ["postgres"], optional = true }
```
hdevalence, line 40: This is a bit awkward but seems like the least bad way to have sqlx impls for the crypto types that allow us to store them in the DB directly without having to write an extra translation layer.

crypto/src/note.rs
```diff
@@ -293,7 +294,8 @@ impl TryFrom<[u8; NOTE_LEN_BYTES]> for Note {
 }
 
 // Note commitment.
-#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
```
hdevalence, line 297: These serde impls are required to be able to use serde on the note commitment tree.

penumbra/src/app.rs
```diff
@@ -1,118 +1,89 @@
...
+    /// To do this, we add a mempool transaction's nullifiers to this set in
+    /// `CheckTx` and remove them when we see they've been committed to a block
+    /// (in `Commit`).  This means that if Tendermint pulls transactions from
+    /// the mempool as part of default block proposer logic, no conflicting
+    /// transactions can appear.
+    ///
+    /// However, it doesn't prevent a malicious validator from proposing
+    /// conflicting transactions, so we need to ensure (in `DeliverTx`) that we
+    /// ignore invalid transactions.
+    mempool_nullifiers: BTreeSet<Nullifier>,
```
hdevalence, line 53: I don't think this is actually used properly yet, but it should be fixed when we patch up the verification logic (untouched by this PR)

penumbra/src/db.rs
```diff
@@ -0,0 +1,58 @@
...
+    .execute(db)
+    .await?;
+
+    query(
+        r#"
+CREATE TABLE IF NOT EXISTS notes (
+    note_commitment bytea PRIMARY KEY,
+    ephemeral_key bytea NOT NULL,
+    encrypted_note bytea NOT NULL,
+    transaction_id bytea NOT NULL,
```
hdevalence, line 39: This just stores the transaction ID, not the full transaction, because Tendermint is going to be storing transactions already.

hdevalence: We should probably update the wallet protocol to return transaction IDs rather than transactions.

PR #148, opened 2021-11-15
Split `penumbra` crate into `pd`, `pcli` crates.

hdevalence:
These don't actually share code, so they should be separate.


hdevalence:
This doesn't clean up the dependencies in the two crates, but I thought it'd be easier to do a straight rename that's easy to merge, and then later work can be done without causing conflicts with any other work.

kwf:
LGTM!

Issue #138, opened 2021-11-11
`pcli`: `tx send` to create a new tx


redshiftzero:
I'm assuming the user should not really be exposed to the concept of notes, the user is just sending/spending, so the UX here TBD

hdevalence:
I think `send` could be a better word than `spend`, possibly something like
```
pcli tx send [value] [addr]
```

To do this we'll need to have an asset registry; I'll create a new issue for that.

Issue #135, opened 2021-11-10
pd: Implement `DeliverTx`


redshiftzero:
See related: #133 #134 #144 

Handle `DeliverTx` by replacing `App.deliver_tx` which should take a Penumbra transaction, verify it (see logic in `App.verify_transaction`), apply the state changes to the node's local state (e.g. Note commitment tree), and queue up changes to the database (which will be applied during the Tendermint `Commit` phase).  

As part of this, you'll need to add the nullifiers to `App.nullifier_set` and add a new table for `nullifer` storage. 

See: https://docs.tendermint.com/master/spec/abci/abci.html#delivertx-2

Issue #35, opened 2021-09-07
Implement client scanning / sync


hdevalence:
Using #33 and #34, implement synchronization logic that does client-side transaction scanning.

This functionality should be exposed with a new `pcli sync` command, and happen automatically on every client action that requires access to the chain state.

PR #157, opened 2021-11-16
feat: send dummy tx using pcli to pd, log for now

### LINKED ISSUE #133: pd, pcli: form penumbra `Transaction` and send to `broadcast_tx_sync` endpoint
(Probably should happen after #129 is merged)

Currently `pd` and `pcli` are using the example key, value store functionality for transaction submission.
Instead, we should have `pcli` form a Penumbra transaction, encode, and send to Tendermint's `BroadCastTxAsync` endpoint. 
If the `DeliverTx` handler is still using the key/value store functionality, then we can replace that by just logging the transaction for now.

redshiftzero:
This connects the transaction sending functionality of `pcli` with its receipt on the `pd` side (using hex-encoded protobuf serialization). The transaction is just a dummy transaction pending the functionality described in #138 (which in turn requires client sync in #35). When the dummy transaction is sent by the client via `pcli tx send`, one will see a `DeliverTx` request be handled by `pd`, and if deserialized correctly, a zero code in the `DeliverTx` response:

```
Nov 16 15:46:45.592  INFO abci{addr=127.0.0.1:52232}:consensus: pd::app: req=DeliverTx(DeliverTx { tx: b"0a3412208b406b738e5b126329c74ef246db66d4b8292032002e20eaa49c2f5c9b4f9a07220e70656e756d6272615f746e3030312a00124072f32a44611108830a035d4818cf1d4b5a920a70b376f064f601728a1ebb2f0e2ec3d0eadfecf0e023c4b39c1edb403bce22b1705337053d911eccf2e5159003" })
Nov 16 15:46:45.592  INFO abci{addr=127.0.0.1:52232}:consensus: pd::app: rsp=DeliverTx(DeliverTx { code: 0, data: b"", log: "", info: "", gas_wanted: 0, gas_used: 0, events: [], codespace: "" })
```

The other `DeliverTx` response fields are unpopulated (for #135). 

Closes #133

### DIFF
--- a/crypto/src/address.rs
+++ b/crypto/src/address.rs
@@ -6,6 +6,8 @@ use bech32::{FromBase32, ToBase32, Variant};
 
 use crate::{fmd, ka, keys::Diversifier, Fq};
 
+pub const CURRENT_CHAIN_ID: &str = "penumbra_tn001";
+/// Human-readable prefix in the address.
 pub const CURRENT_NETWORK_ID: &str = "penumbra_tn001_";
 
 /// A valid payment address.
--- a/crypto/src/address.rs
+++ b/crypto/src/address.rs
@@ -6,7 +6,7 @@ use bech32::{FromBase32, ToBase32, Variant};
 
 use crate::{fmd, ka, keys::Diversifier, Fq};
 
-pub const CURRENT_CHAIN_ID: &str = "penumbra_tn001";
+pub const CURRENT_CHAIN_ID: &str = "penumbra-tn001";
 /// Human-readable prefix in the address.
 pub const CURRENT_NETWORK_ID: &str = "penumbra_tn001_";
 
--- a/crypto/src/transaction.rs
+++ b/crypto/src/transaction.rs
@@ -184,6 +184,12 @@ impl From<Transaction> for transaction::Transaction {
     }
 }
 
+impl From<&Transaction> for transaction::Transaction {
+    fn from(msg: &Transaction) -> Self {
+        msg.into()
+    }
+}
+
 impl TryFrom<transaction::Transaction> for Transaction {
     type Error = ProtoError;
 
@@ -217,7 +223,22 @@ impl TryFrom<&[u8]> for Transaction {
     }
 }
 
+impl TryFrom<Vec<u8>> for Transaction {
+    type Error = ProtoError;
+
+    fn try_from(bytes: Vec<u8>) -> Result<Transaction, Self::Error> {
+        Ok(Self::try_from(&bytes[..])?)
+    }
+}
+
 impl Into<Vec<u8>> for Transaction {
+    fn into(self) -> Vec<u8> {
+        let protobuf_serialized: transaction::Transaction = self.into();
+        protobuf_serialized.encode_to_vec()
+    }
+}
+
+impl Into<Vec<u8>> for &Transaction {
     fn into(self) -> Vec<u8> {
         let protobuf_serialized: transaction::Transaction = self.into();
         protobuf_serialized.encode_to_vec()
--- a/crypto/src/transaction.rs
+++ b/crypto/src/transaction.rs
@@ -5,7 +5,10 @@ use bytes::Bytes;
 
 use decaf377::FieldExt;
 
-use penumbra_proto::{transaction, Message, Protobuf};
+use penumbra_proto::{
+    transaction::Fee as ProtoFee, transaction::Transaction as ProtoTransaction,
+    transaction::TransactionBody as ProtoTransactionBody, Message, Protobuf,
+};
 
 use crate::{
     action::{error::ProtoError, Action},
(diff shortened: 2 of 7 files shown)


crypto/src/address.rs
```diff
@@ -6,6 +6,8 @@ use bech32::{FromBase32, ToBase32, Variant};
 
 use crate::{fmd, ka, keys::Diversifier, Fq};
 
+pub const CURRENT_CHAIN_ID: &str = "penumbra_tn001";
```
hdevalence, line 9: nit: from what i've seen, most of the other chain IDs have `-` separators, maybe worth aligning on that?

crypto/src/transaction.rs
```diff
@@ -184,6 +184,12 @@ impl From<Transaction> for transaction::Transaction {
     }
 }
 
+impl From<&Transaction> for transaction::Transaction {
```
hdevalence, line 187: not really specific to this PR, but this naming is a little confusing, maybe we could do something like `use _ as RawTransaction` or `use _ as ProtoTransaction` or something?

pcli/src/main.rs
```diff
@@ -78,6 +78,12 @@ enum Addr {
     },
 }
 
+#[derive(Debug, StructOpt)]
+enum Tx {
+    /// Send transaction to the node.
+    Send,
```
hdevalence, line 84: This should probably eventually specify the amount, denom, and destination addresses, right?  (even if we're not currently using them for the dummy transactions?)

pd/src/app.rs
```diff
@@ -155,10 +155,30 @@ impl App {
     }
 
     #[instrument(skip(self))]
-    fn deliver_tx(&mut self, tx: Bytes) -> response::DeliverTx {
+    fn deliver_tx(&mut self, hex_encoded_bytes: Bytes) -> response::DeliverTx {
         // TODO: implement (#135)
 
-        // This should accumulate data from `tx` into `self.pending_block`
+        // Transactions that cannot be deserialized should return a non-zero `DeliverTx` code.
+        let txbytes = match hex::decode(hex_encoded_bytes.as_ref()) {
```
hdevalence, line 162: Can we have the `tx` be the proto-encoded bytes instead? I worry that this is going to mean that we have the blocks end up containing a bunch of hex-encoded protos.

wallet/src/state.rs
```diff
@@ -42,6 +46,21 @@ impl ClientState {
         self.wallet.new_address(label)
     }
 
+    /// Generate a new transaction.
+    pub fn new_transaction<R: RngCore + CryptoRng>(
```
hdevalence, line 50: Right now this is a stub function, but going forward, do we want to have a method like this, vs exposing a transaction builder? (Trying to imagine what this looks like if it had all of the arguments necessary to specify transaction details).

pcli/src/main.rs
hdevalence: maybe we want to allow multiple (amount, denom) pairs?

pcli/src/main.rs
```diff
@@ -101,13 +107,20 @@ async fn main() -> Result<()> {
...
+
+            // TODO: Set fee on CLI (part of issue #138)
+            let fee = 0;
+            let dummy_tx = local_storage.new_transaction(&mut OsRng, fee)?;
+            let serialized_tx: Vec<u8> = dummy_tx.into();
 
             let rsp = reqwest::get(format!(
-                r#"http://{}:{}/broadcast_tx_async?tx="{}={}""#,
-                opt.node, opt.abci_port, key, value
+                r#"http://{}:{}/broadcast_tx_async?tx="{}""#,
```
hdevalence, line 120: Per https://docs.tendermint.com/master/tendermint-core/using-tendermint.html#formatting , it looks like specifying the transaction with an `0x` prefix will cause Tendermint to interpret it as hex-encoded binary data, and pass it to the application as binary.

pd/src/app.rs
hdevalence: cf https://github.com/penumbra-zone/penumbra/pull/157/files#r750810376

crypto/src/transaction.rs
redshiftzero: good idea, took a pass through this file and replaced the imports (now `ProtoTransactionBody`, `ProtoTransaction`, `ProtoFee`)

pcli/src/main.rs
redshiftzero: I think initially supporting one spend makes sense (and then I guess under the hood we'd have the transaction builder create two spends in that case, one to the dest address, one with the change). The other thing we'd want as a required field is the fee (required field in the transaction builder).

wallet/src/state.rs
redshiftzero: I was thinking that we'd want the transaction building to happen in a method on the `ClientState` just because that's where we have knowledge of e.g. which notes we have (https://github.com/penumbra-zone/penumbra/blob/main/crypto/src/transaction/builder.rs#L40). So from the `pcli` side we'd pass in the data from the user's request "I want X of asset Y to go to address Z", and then the this method tries to build a transaction, bailing out if e.g. there are insufficient notes to send X of asset Y.

redshiftzero:
Thanks for these comments! Addressed in follow up commits

PR #160, opened 2021-11-17
fix: handle no genesis state

redshiftzero:
Currently if there is no Penumbra specific app state in the genesis
file, a panic occurs. This changes that behavior to warn instead, and
continue on.


redshiftzero:
Closing this - upon discussion the current behavior (panic if there is no Penumbra-specific state) is the desired behavior

PR #162, opened 2021-11-17
feat: #140 implement asset registry

### LINKED ISSUE #140: Create an asset registry
Our asset IDs are hashes of denomination traces, so we need to have a registry that inverts the map and allows looking up the denomination by the asset id.  This can be part of the public chain state, since conservation of value means that every denomination is either (a) one that was created as part of the geneses state or (b) one that was created as part of a (transparent) IBC transfer.

We could do this by (proposal sketch):

- [x] keeping a `BTreeMap<asset::Id, String>` in the node application;
- [x] initially populating it with the assets from genesis;
- [x] providing an ABCI query that returns a list of `(asset_id, denom)` pairs;
- [ ] ~having `pcli` pull the registry while syncing and cache it as part of its local state.~

zbuc:
This implements an Asset Registry containing mappings of asset IDs to human-readable asset denominations.

Also included are two ABCI methods (`AssetList` and `AssetLookup`) for obtaining registry data, which have also been implemented in `pcli`.

Closes #140

### DIFF
--- a/pcli/src/main.rs
+++ b/pcli/src/main.rs
@@ -7,7 +7,8 @@ use std::{fs, io, process};
 use structopt::StructOpt;
 
 use penumbra_proto::wallet::{
-    wallet_client::WalletClient, CompactBlockRangeRequest, TransactionByNoteRequest,
+    wallet_client::WalletClient, AssetLookupRequest, CompactBlockRangeRequest,
+    TransactionByNoteRequest,
 };
 use penumbra_wallet::{state, storage};
 
@@ -32,6 +33,16 @@ struct Opt {
     wallet_location: Option<String>,
 }
 
+// Note: can't use `Vec<u8>` directly, as structopt would instead look for
+// conversion function from `&str` to `u8`.
+type Bytes = Vec<u8>;
+
+fn parse_bytestring(s: &str) -> Result<Vec<u8>, String> {
+    let decoded = hex::decode(s).expect("Invalid bytestring");
+
+    Ok(decoded)
+}
+
 #[derive(Debug, StructOpt)]
 enum Command {
     /// Creates a transaction.
@@ -47,6 +58,11 @@ enum Command {
     FetchByNoteCommitment { note_commitment: String },
     /// Block request - TEMP (developer only, remove when sync implemented)
     BlockRequest { start_height: u32, end_height: u32 },
+    /// Asset Registry Lookup based on asset ID
+    AssetLookup {
+        #[structopt(parse(try_from_str = parse_bytestring))]
+        asset_id: Bytes,
+    },
 }
 
 #[derive(Debug, StructOpt)]
@@ -218,6 +234,18 @@ async fn main() -> Result<()> {
             while let Some(block) = stream.message().await? {
                 tracing::info!("got fragment: {:?}", block);
             }
+        }
+        Command::AssetLookup { asset_id } => {
+            // let spend_key = load_wallet(&wallet_path);
+            // let _local_storage = state::ClientState::new(spend_key);
+            // xx don't hardcode, need to pass multiple ports on the command line
+            let mut client =
+                WalletClient::connect(format!("http://{}:{}", opt.node, opt.wallet_port)).await?;
+            tracing::info!("requesting asset denom for asset id: {:?}", &asset_id,);
+            let request = tonic::Request::new(AssetLookupRequest { asset_id });
+            let asset = client.asset_lookup(request).await?.into_inner();
+
+            tracing::info!("got asset: {:?}", asset);
         }
         _ => todo!(),
     }
--- a/pcli/src/main.rs
+++ b/pcli/src/main.rs
@@ -203,7 +203,6 @@ async fn main() -> Result<()> {
         Command::FetchByNoteCommitment { note_commitment } => {
             let spend_key = load_wallet(&wallet_path);
             let _local_storage = state::ClientState::new(spend_key);
-            // xx don't hardcode, need to pass multiple ports on the command line
             let mut client =
                 WalletClient::connect(format!("http://{}:{}", opt.node, opt.wallet_port)).await?;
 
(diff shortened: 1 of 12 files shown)


pcli/src/main.rs
```diff
@@ -219,6 +237,29 @@ async fn main() -> Result<()> {
                 tracing::info!("got fragment: {:?}", block);
             }
         }
+        Command::AssetLookup { asset_id } => {
+            // xx don't hardcode, need to pass multiple ports on the command line
```
redshiftzero, line 241: This comment can be deleted here in `Command::AssetLookup` and below in `Command::AssetList` (leftover from a prior version before we were passing in the port of the wallet service on the command line)

pd/src/wallet.rs
```diff
@@ -66,4 +67,40 @@ impl Wallet for WalletApp {
...
+
+    async fn asset_list(
+        &self,
+        request: tonic::Request<AssetListRequest>,
+    ) -> Result<tonic::Response<Self::AssetListStream>, Status> {
+        let state = self.state.clone();
+
+        let (tx, rx) = mpsc::channel(100);
+
+        let state = self.state.clone();
```
redshiftzero, line 91: we have this also on line 87, looks like we can delete one

pd/src/genesis.rs
```diff
@@ -5,42 +5,34 @@ use rand_chacha::ChaCha20Rng;
...
-)]
-pub struct GenesisNotes {
-    notes: Vec<Note>,
-}
-
-impl GenesisNotes {
-    pub fn notes(&self) -> Vec<Note> {
-        self.notes.clone()
-    }
+#[derive(Clone, Debug, PartialEq, Eq)]
```
redshiftzero, line 31: OK I see that this `GenesisNotes`/`GenesisNote` refactor was done since we need to pass in the asset denomination in the genesis file instead of the asset ID to populate the asset registry :+1:. Do we want to preserve the `notes` key in the `app_state` field in the genesis file? Probably yes if we have other genesis application state, otherwise fine to remove. Do you know @hdevalence?

pd/src/genesis.rs
```diff
@@ -71,79 +63,61 @@ impl FromStr for GenesisAddr {
...
+            value: GenesisNoteValue,
+            note_blinding: Fq,
+        ) -> Result<Self, note::Error> {
+            Ok(GenesisNote {
+                diversifier: diversifier.0,
+                amount: value.amount,
+                note_blinding: note_blinding.to_bytes(),
+                asset_denom: value.asset_denom,
+                transmission_key: transmission_key.0,
+                // transmission_key_s: Fq::from_bytes(transmission_key.0)
```
redshiftzero, line 98: can delete, the s-component of the transmission key will be derived when the `Note` is created for realsies

pd/src/state.rs
```diff
@@ -181,4 +184,63 @@ INSERT INTO notes (
...
+
+        let mut assets = vec![];
+        for a in _assets {
+            assets.push(Asset {
+                asset_denom: a.0,
+                asset_id: a.1,
+            });
+        }
+
+        Ok(assets)
```
redshiftzero, line 220: (optional) Could do this in a map:

```rust
Ok(
    query_as::<_, (String, Vec<u8>)>("SELECT denom, asset_id FROM assets")
        .fetch_all(&mut conn)
        .await?
        .into_iter()
        .map(|(asset_denom, id)| Asset {
            asset_denom,
            asset_id: id.into(),
        })
        .collect(),
)
```

redshiftzero (commented):
Tried out this service, looks great! Dropped a few comments inline

zbuc:
@redshiftzero thanks! PR feedback addressed

pd/src/genesis.rs
hdevalence: Yeah, I think we should keep this filed under the `notes` key to leave open adding other data.

pd/src/genesis.rs
```diff
@@ -71,79 +63,59 @@ impl FromStr for GenesisAddr {
     }
 }
 
-mod helpers {
+pub mod helpers {
```
hdevalence, line 66: Rather than making this module public, I think it could be eliminated / have its contents folded into the main `genesis` module.  Originally, we were using serde on remote types, so it made sense to have a helpers module to contain the serde code, but now we can just expose the types directly.

pd/src/genesis.rs
```diff
@@ -5,42 +5,34 @@ use rand_chacha::ChaCha20Rng;
...
-pub struct GenesisNotes {
-    notes: Vec<Note>,
-}
-
-impl GenesisNotes {
-    pub fn notes(&self) -> Vec<Note> {
-        self.notes.clone()
-    }
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct GenesisNoteValue {
```
hdevalence, line 32: Maybe this could just be `GenesisValue`?

hdevalence: The doc comment could note that the difference between this type and the regular `Value` type is just that it stores the full denomination, rather than a hash output.

pd/src/state.rs
```diff
@@ -181,4 +184,61 @@ INSERT INTO notes (
...
+    ) VALUES ($1, $2)
+    "#,
+            )
+            .bind(&asset.0.to_bytes()[..])
+            .bind(asset.1)
+            .execute(&mut dbtx)
+            .await?;
+        }
+
+        dbtx.commit().await.map_err(Into::into)
```
hdevalence, line 242: Instead of having this logic be a standalone function, I think it would be better if it were part of `commit_block`, which is already going to be processing the `PendingBlock` -- that way, we can ensure that every database write happens as part of one big atomic transaction.

hdevalence (changes requested):
This looks great, just have two suggested changes:  (1) flatten the `genesis` module and (2) put the database writes into the main commit function rather than something standalone.

pd/src/state.rs
zbuc: Or should it accept a transaction, that way we can re-use it to save assets whenever new ones are found in other blocks?

hdevalence (approved):
LGTM!

PR #165, opened 2021-11-18
wip: implement `CheckTx` and `DeliverTx`

### LINKED ISSUE #134: pd: Implement `CheckTx`
See related: #133

Handle `CheckTx` in a new method on `App` which should take a Penumbra transaction and verify it (using the `App.verify_transaction` method as a guide). 

See https://docs.tendermint.com/master/spec/abci/abci.html#checktx-2

### LINKED ISSUE #135: pd: Implement `DeliverTx`

redshiftzero:
Work towards implementing `CheckTx` and `DeliverTx`, not ready yet but putting up as draft in case there are thoughts on the approach so far

There are two new transaction types used during transaction processing:
* `PendingTransaction` which is a transaction that has had its stateless checks performed
* `VerifiedTransaction` which is a transaction that has had all checks (stateless and stateful) performed

Closes #134 
Closes #135


pd/src/app.rs
```diff
@@ -107,17 +113,26 @@ impl App {
...
+        let pending_transaction = genesis_tx
+            .verify_stateless(Some(value_balance))
+            .expect("genesis tx must be valid");
+        let verified_transaction = pending_transaction
+            .verify_stateful(self.note_commitment_tree.root2())
+            .expect("genesis tx must be valid");
 
         // Now add the transaction and its note fragments to the pending state changes.
-        genesis_block.add_transaction(genesis_tx);
+        genesis_block.add_transaction(verified_transaction);
```
redshiftzero, line 132: we verify the genesis transaction to produce `VerifiedTransaction` which now `PendingBlock::add_transaction` required

pd/src/app.rs
```diff
@@ -160,33 +175,192 @@ impl App {
...
 
+    /// Perform checks before adding a transaction into the mempool via `CheckTx`.
+    ///
+    /// In the transaction validation performed before adding a transaction into the
+    /// mempool, we check that:
+    ///
+    /// * All binding and auth sigs signatures verify (stateless),
+    /// * All proofs verify (stateless and stateful),
+    /// * The transaction does not reveal nullifiers already revealed in another transaction
+    /// in the mempool or in the database,
```
redshiftzero, line 192: The checks we want to perform at the `CheckTx` could be very minimal since it's up to us, but for now we're performing the full validation to reduce the invalid transactions that make their way into blocks

pd/src/app.rs
```diff
@@ -160,33 +175,192 @@ impl App {
...
+            };
+
+            // Ensure we do not add any transactions with duplicate nullifiers into the mempool.
+            //
+            // Note that we only run this logic if this `CheckTx` request is from a new transaction
+            // (i.e. `CheckTxKind::New`). If this is a recheck of an existing entry in the mempool,
+            // then we don't need to add the nullifier again, as it's already in `self.mempool_nullifiers`.
+            // Rechecks occur whenever a block is committed if the Tendermint `mempool.recheck` option is
+            // true, which is the default option.
+            if request.kind == CheckTxKind::New {
```
redshiftzero, line 237: Another option here would be to additional store the transaction ID in addition to the nullifier in the `mempool_nullifiers` structure so we can verify that if there is a matching nullifier already in the mempool, it's from the same transaction ID

crypto/src/transaction.rs
```diff
@@ -173,6 +174,36 @@ impl Transaction {
...
+            (value_commitments - value_balance).compress().0.into();
+        let binding_verification_key: VerificationKey<Binding> = binding_verification_key_bytes
+            .try_into()
+            .expect("verification key is valid");
+
+        let transaction_body_serialized: Vec<u8> = self.transaction_body.clone().into();
+        binding_verification_key
+            .verify(&transaction_body_serialized, &self.binding_sig)
+            .is_ok()
+    }
```
hdevalence, line 206: Do we still need this function, if we're not going to be verifying the signatures on the genesis transactions, where it would be used? It doesn't look like it's in use anywhere as yet.

pd/src/verify.rs
```diff
@@ -0,0 +1,169 @@
...
+
+        Ok(VerifiedTransaction {
+            id: self.id,
+            new_notes: self.new_notes.clone(),
+            spent_nullifiers: self.spent_nullifiers.clone(),
+        })
+    }
+}
+
+/// One-off function used to generate a verified genesis transaction.
```
hdevalence, line 143: Maybe this could have a comment / name change to indicate that it doesn't actually do verification? Like `mark_genesis_as_verified` or something?

pd/src/app.rs
hdevalence: Yeah, I think until we do ABCI++ we can't influence the selection of transactions by the block proposer, so we need to rely on `CheckTx` doing full verification.

crypto/src/transaction.rs
redshiftzero: yeah this would be something we'd use if we have any other transactions with non-zero value balance, but I'll snip it out for now

redshiftzero:
thank you for these comments! Updated

PR #173, opened 2021-11-22
fix: Include fee in PEN in transaction value balance

### LINKED ISSUE #170: crypto: fix fee handling in transaction builder
Currently in the transaction builder we track the value balance and error if it is non-zero (meaning value is being created or destroyed, not just moved around): https://github.com/penumbra-zone/penumbra/blob/main/crypto/src/transaction/builder.rs#L29

We should edit the `set_fee` method to include the fee in the value balance computation: https://github.com/penumbra-zone/penumbra/blob/main/crypto/src/transaction/builder.rs#L130

For now the fee must be in set in `PEN`.

redshiftzero:
This includes the fee in the transaction builder's value balance computation

Closes #170

### DIFF
--- a/crypto/src/transaction/builder.rs
+++ b/crypto/src/transaction/builder.rs
@@ -7,7 +7,7 @@ use super::Error;
 use crate::rdsa::{Binding, Signature, SigningKey};
 use crate::{
     action::{output, spend, Action},
-    ka,
+    asset, ka,
     keys::{OutgoingViewingKey, SpendKey},
     memo::MemoPlaintext,
     merkle,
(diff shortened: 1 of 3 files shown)


crypto/src/transaction/builder.rs
```diff
@@ -126,7 +126,24 @@ impl Builder {
     }
 
     /// Set the transaction fee in PEN.
-    pub fn set_fee(mut self, fee: u64) -> Self {
+    pub fn set_fee<R: RngCore + CryptoRng>(mut self, rng: &mut R, fee: u64) -> Self {
+        let pen_trace = b"PEN";
```
redshiftzero, line 130: Did we want to standardize on `pen` or `PEN`?

hdevalence: let's use lower case in the code, and figure out a final name later

crypto/src/transaction/builder.rs
```diff
@@ -126,7 +126,24 @@ impl Builder {
     }
 
     /// Set the transaction fee in PEN.
-    pub fn set_fee(mut self, fee: u64) -> Self {
+    pub fn set_fee<R: RngCore + CryptoRng>(mut self, rng: &mut R, fee: u64) -> Self {
```
hdevalence, line 129: this could be rng: mut R instead, and then people can pass an owned or boorowed RNG (since &mut R impls RngCore if R does).

hdevalence (approved):
LGTM mod (tioli) comments

redshiftzero:
this has the wrong sign: 55cecbeb754e2ec169c4986730594b0663ca7c2c fixes

PR #176, opened 2021-11-22
feat: add pcli balance command 

### LINKED ISSUE #137: `pcli`: add `balance` command
This command would list the balance of the notes (by asset) that the user has.

This should be done after the sync (#33 / #35) is implemented.

redshiftzero:
Closes #137, adding a command that lists balance by asset. In that ticket another idea is proposed, adding balance by address (could be part of the address command perhaps?), I propose we do in a followup such that the `notes_by_asset_denomination` method and asset registry syncing can be used for #138 

Output looks like:
```
 Asset denomination  Balance
 pen                 100
```

### DIFF
--- a/crypto/src/asset.rs
+++ b/crypto/src/asset.rs
@@ -1,4 +1,6 @@
 //! Asset types and identifiers.
+use std::convert::{TryFrom, TryInto};
+
 use ark_ff::fields::PrimeField;
 use decaf377::FieldExt;
 use once_cell::sync::Lazy;
@@ -36,6 +38,7 @@ impl std::fmt::Debug for Id {
 
 // XXX define a DenomTrace structure ?
 
+// xx rename this derive?
 impl From<&[u8]> for Id {
     fn from(slice: &[u8]) -> Id {
         // Convert an asset name to an asset ID by hashing to a scalar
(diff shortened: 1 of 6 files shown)


hdevalence:
On the balance by address front, maybe it would make sense to have a `--by-address` `bool` flag for the `balance` command that would build a table like this?
```
Address            Asset        Balance
Default             pen            100
                    cubes         12
Dogs                shib            24
                    doge          9
```

hdevalence:
Another general thought: we should figure out how to keep track of what the base denomination of each asset is, since we'll want to be recording integral values of `upen` or whatever, but displaying as `9.23238921` or something.

crypto/src/asset.rs
```diff
@@ -36,6 +38,7 @@ impl std::fmt::Debug for Id {
 
 // XXX define a DenomTrace structure ?
 
+// xx rename this derive?
```
hdevalence, line 41: long-run I think we should probably eliminate it and have `From` for a new `Denom` type.

wallet/src/state.rs
```diff
@@ -76,6 +80,46 @@ impl ClientState {
...
+                Some(current_notes) => {
+                    let mut new_notes: Vec<Note> =
+                        current_notes.iter().map(|note| note.clone()).collect();
+                    new_notes.push(note.clone());
+                    new_notes.to_vec()
+                }
+                None => {
+                    vec![note.clone()]
+                }
+            };
```
hdevalence, line 105: non-blocking: this could be cleaner using the `Entry` [API](https://doc.rust-lang.org/stable/std/collections/hash_map/enum.Entry.html#), maybe sth like
```
notemap.entry(asset_denom).or_insert(Vec::new()).push(note.clone());
```

hdevalence (approved):
LGTM!

PR #180, opened 2021-11-23
Implement `pcli tx send`

### LINKED ISSUE #138: `pcli`: `tx send` to create a new tx

redshiftzero:
Closes #138

This draft PR implements the basic functionality for constructing a transaction in `pcli`. Other thoughts (also listed as comments inline) are:
* We could specify the memo in an optional arg on the command line for `tx send`
* It would be cleaner when we have no memo to let `Builder::add_output` take an `Option<MemoPlaintext>`
* It may be nice to let the user specify the change address - currently it's just the Default index 0 address

Not ready for merge as currently the spend auth sig is failing to verify in CheckTx: you will see this if you try to submit a transaction with debug level logging. I've added a regression test to display the problem in https://github.com/penumbra-zone/penumbra/pull/180/commits/5956d77ec5733956f77f92d79e1a70bd21827d3e (the test currently will fail during the spend auth verification as there is no fix yet)

### DIFF
--- a/pd/src/app.rs
+++ b/pd/src/app.rs
@@ -1,5 +1,5 @@
 use std::{
-    collections::BTreeSet,
+    collections::{BTreeSet, VecDeque},
     future::Future,
     pin::Pin,
     sync::{Arc, Mutex},
@@ -34,6 +34,8 @@ use crate::{
 
 const ABCI_INFO_VERSION: &str = env!("VERGEN_GIT_SEMVER");
 
+const NUM_RECENT_ANCHORS: usize = 64;
+
 /// The Penumbra ABCI application.
 #[derive(Debug)]
 pub struct App {
@@ -41,6 +43,11 @@ pub struct App {
 
     /// Written to the database after every block commit.
     note_commitment_tree: merkle::BridgeTree<note::Commitment, { merkle::DEPTH as u8 }>,
+
+    /// Recent anchors of the note commitment tree.
+    ///
+    /// TODO: load from database?
+    recent_anchors: VecDeque<merkle::Root>,
 
     /// We want to prevent two transactions from spending the same note in the
     /// same block.  Our only control over whether transactions will appear in a
@@ -77,6 +84,8 @@ impl App {
         Ok(Self {
             state,
             note_commitment_tree,
+            // TODO: load from DB
+            recent_anchors: Default::default(),
             mempool_nullifiers: Arc::new(Default::default()),
             pending_block: None,
             completion_tracker: Default::default(),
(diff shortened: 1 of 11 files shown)


pd/src/app.rs
```diff
@@ -195,27 +204,30 @@ impl App {
     ) -> impl Future<Output = Result<Response, BoxError>> {
         let state = self.state.clone();
         let mempool_nullifiers = self.mempool_nullifiers.clone();
-        let nct_root = self.note_commitment_tree.root2();
-        let finished_signal = self.completion_tracker.start();
+        let recent_anchors = self.recent_anchors.clone();
```
kwf, line 207: Why are we cloning the `VecDeque` of recent anchors here? Should there be a shared reference instead?

pd/src/app.rs
```diff
@@ -281,30 +295,32 @@ impl App {
     /// so it is not safe to assume all checks performed in `CheckTx` were done.
     #[instrument(skip(self))]
     fn deliver_tx(&mut self, txbytes: Bytes) -> impl Future<Output = Result<Response, BoxError>> {
-        let finished_signal = self.completion_tracker.start();
         let state = self.state.clone();
-        let nct_root = self.note_commitment_tree.root2();
+        let recent_anchors = self.recent_anchors.clone();
```
kwf, line 299: Similarly, why are we cloning the recent anchors here?

kwf (approved):
LGTM other than comments left about the handling of the recent NCT anchors.

pd/src/app.rs
hdevalence: It's cloning just because it's the simplest way to access the anchors from inside the future -- we could use a shared reference by wrapping it in an `Arc` or something, but then we'd have to think about how we update the data, etc.  Not claiming that this is the best solution, but it works just well enough while we're testing to avoid transactions getting stuck because the state rolled forward while they were being submitted.

pd/src/app.rs
hdevalence: cf https://github.com/penumbra-zone/penumbra/pull/180#discussion_r756450269 , the reason is the same so marking this arm as resolved and moving discussion there

pd/src/app.rs
kwf: That seems okay to me, then, provided we revisit later. I think an `Arc<RwLock<VecDeque<...>>>` should work well here; we don't need the futures-aware `RwLock` because we're not going to need to lock across an `.await`-point.

PR #182, opened 2021-11-24
Spend detection

### LINKED ISSUE #155: switch to `sqlx::query_as!` and `sqlx::query!` macros
We've recently hit two similar bugs (https://github.com/penumbra-zone/penumbra/issues/154 being the most recent) that @kwf pointed out we could be catching at build time if we were using the `sql::query!` and `sql::query_as!` macros instead of the corresponding functions.

This issue is to take a pass through the current codebase and switch to the macros. 

It does mean that a database connection will be required at build time, and in offline mode (i.e. CI) we'll need to check in a JSON file containing our database details: https://docs.rs/sqlx/0.5.9/sqlx/macro.query.html#offline-mode-requires-the-offline-feature

### LINKED ISSUE #175: Extend wallet protocol to detect spends
The current wallet protocol provides a `CompactBlock` proto containing a bunch of `StateFragment`s, which have information required to detect notes created by output descriptions.  The current scanning implementation only detects new inbound notes, and does not detect when notes controlled by the wallet are spent.

To do this, we should:

- [x] Extend the `CompactBlock` proto with a `repeated bytes nullifiers = 3;` field encoding a bunch of 32-byte nullifiers;
- [x] Extend the `ClientState::scan_block` method to check the spent nullifiers in the compact block against the nullifiers in the `unspent_set` and move them to the `spent_set` when the nullifier is observed.

Later, we could:

- Change the `TransactionByNoteRequest` to also allow specifying a nullifier and obtaining the transaction that spent that nullifier (this involves changing the database layout).  This isn't important to do now.

kwf:
Fixes #175 by extending the wallet protocol to send nullifier sets in `CompactBlock`s and using them to detect spends on the set of known unspent notes kept by the client.

Included in this PR are some improvements to the way we handle the database, which fix #155.

Additionally, if using VS Code, one can get `sqlx` query macros properly verified by rust-analyzer by opening the project directory as a folder; this will pick up the `.vscode/settings.json` file added here which sets the `DATABASE_URL` for the project appropriately.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "rust-analyzer.server.extraEnv": {"DATABASE_URL": "postgres://postgres:postgres@localhost/penumbra" },
+}
--- a/crypto/src/action/spend.rs
+++ b/crypto/src/action/spend.rs
@@ -1,6 +1,5 @@
 use std::convert::{TryFrom, TryInto};
 
-use ark_ff::UniformRand;
 use bytes::Bytes;
 use rand_core::{CryptoRng, RngCore};
 
@@ -76,7 +75,7 @@ pub struct Body {
 impl Body {
     #[allow(clippy::too_many_arguments)]
     pub fn new<R: RngCore + CryptoRng>(
-        rng: &mut R,
+        _rng: &mut R,
         value_commitment: value::Commitment,
         ask: SigningKey<SpendAuth>,
         spend_auth_randomizer: Fr,
--- a/crypto/src/note.rs
+++ b/crypto/src/note.rs
@@ -346,9 +346,9 @@ impl From<Commitment> for [u8; 32] {
     }
 }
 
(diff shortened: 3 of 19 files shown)


hdevalence (approved):
LGTM. One question about the `sqlx` changes -- what exactly is required to build the precompiled data? What are the implications for our dev env requirements?

kwf:
Regarding the `sqlx` changes:

You will need an instance of the database operational whenever you _change_ a `query!` statement. The `DATABASE_URL` environment variable must be set to point to the running database with the correct schema. Then, using the `sqlx-cli` tool, you can run `cargo sqlx prepare -- --lib --offline` to (re-)generate the `sqlx-data.json` file for `pd` (you have to run this inside the `pd` directory, since it won't work at the workspace root). Checking that file into version control means that other people who build the package, including CI, do not need to have the database running at build time.

PR #184, opened 2021-11-24
pcli: suppress warning message with undocumented env var

hdevalence:
This commit adds an (intentionally) undocumented environment variable that
suppresses the scary testnet warning text.  If someone can dig through the
source and/or git history to find it and then set it, they're good to go.


kwf (approved):
:rocket:

PR #187, opened 2021-11-25
WIP: serve compact blocks faster

hdevalence:
The current implementation processes each block sequentially. Instead, we could use the database asynchronously and query blocks concurrently.  This change makes the scanning process 2x faster on empty blocks (scanning 9000 blocks in 9s rather than 23s), but probably introduces a DoS risk (it spawns a task for each requested block, with no limit on the number of requested blocks).


hdevalence:
I'm not sure this is the right direction to go in now, so closing this for reevaluation in the future.

PR #194, opened 2021-11-29
168: Store backups of wallets in archive directory for debugging

### LINKED ISSUE #168: Build archive tooling for retroactive testnet debugging
We're planning to ship early testnets before we reasonably expect things to work, and iterate in public.  It's likely that we'll discover unexpected behavior or bugs, and we'd like the ability to inspect the chain state to figure out what happened.  It'd be useful to be able to work with early testers on debugging in case something goes wrong.

To enable this, a useful tool would be a local, append-only archive of the key material used at any point in any testnet. (We're not concerned with security at this point, since the testnets record no value). As we iterate on our wallet state, key derivation, crypto design, ..., we'll want to be unconstrained about the storage format, so it seems like the best approach would be to add a "sidecar" archive that stores copies of all of the root key material.  That way, we can delete or overwrite the main wallet state as much as we want during development, without losing the ability to introspect past data if we found out we needed to.

Currently we store the `pcli` wallet in a system-specific `pcli` directory.  In parallel, we could create a `testnet-archive` directory (`ProjectDirs::from("zone", "penumbra", "testnet-archive")`).  Then, each time we generate a new key, we would save the key material into `{ARCHIVE_DIR}/tn{XYZ}/{unique_id}`, where XYZ is the testnet number and unique_id is some kind of unique identifier (the point is just to be able to write in without collisions and without having to check the existing fs state, so maybe a hash of the key or something?).

It's not important to be able to read any of this at this time, instead we should just make sure that we're writing out data that's sufficient to leave the door open to debugging, if the need should arise.

zbuc:
Closes #168 

This only affects generation of new wallets, as a result existing wallets won't be archived (shouldn't be an issue, but worth pointing out).

Additionally I saw some logic in the wallet code for re-saving client state (e.g. `state.commit()`) however I don't think it's desirable to do that step for backups since we only really care about the seed data AFAIK -- please correct me if that's a bad assumption.

### DIFF
--- a/pcli/src/main.rs
+++ b/pcli/src/main.rs
@@ -3,9 +3,11 @@ use comfy_table::{presets, Table};
 use directories::ProjectDirs;
 use penumbra_crypto::keys::SpendSeed;
 use rand_core::OsRng;
+use sha2::{Digest, Sha256};
 use std::path::{Path, PathBuf};
 use structopt::StructOpt;
 
+use penumbra_crypto::CURRENT_CHAIN_ID;
 use penumbra_wallet::{ClientState, Wallet};
 
 pub mod opt;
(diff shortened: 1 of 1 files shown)


hdevalence (approved):
LGTM

PR #195, opened 2021-11-29
add asset denom type

### LINKED ISSUE #179: crypto: add asset `Denom` type
One odd part of the crypto crate's asset module is that we have an `impl From<&[u8]> for Id` intended for hashing denomination strings (e.g. `b'pen'`) to their asset IDs, and also an `impl TryFrom<Vec<u8>> for Id` for creating an asset ID from the inner Fq's bytes. 

The suggestion [here](https://github.com/penumbra-zone/penumbra/pull/176/files/aadf860318c9129f74719c0231f35922811729ce#r754720628) to resolve this is to add a `Denom` type in the asset module. This would mean that `impl From<&[u8]> for Id` would become `impl From<Denom> for Id`.

avahowell:
This PR adds the asset::Denom type, which wraps a `String` and adds a `From<Denom>` implementation for `asset::Id`.

Closes #179

### DIFF
--- a/crypto/src/asset.rs
+++ b/crypto/src/asset.rs
@@ -28,6 +28,8 @@ use crate::Fq;
 #[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
 pub struct Id(pub Fq);
 
+pub struct Denom(pub String);
+
 impl std::fmt::Debug for Id {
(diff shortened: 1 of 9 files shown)


kwf (approved):
This looks good to me!

PR #196, opened 2021-11-30
feat: integrate metrics

### LINKED ISSUE #12: Basic metrics support
Metrics are invaluable for debugging and observability. It's much easier to add them as the software is being developed, rather than to go back and add them into all of the existing code. So, at the outset:

- [ ] Add support for metrics using the `metrics` crate (at the outset, this could just be a Prometheus endpoint and a single metric);
- [ ] Write down a set of standards for metrics names, aiming for consistency in metrics as they're added.

redshiftzero:
Closes #12 

You can try this out for yourself by running and observing the counters on http://127.0.0.1:9000/ which is the endpoint for Prometheus to scrape

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,7 +152,11 @@ Now when you start `pd` and tendermint as described above, you will see a messag
 
 ### Metrics
 
-When adding new metrics, please following the [Prometheus metrics naming guidelines](https://prometheus.io/docs/practices/naming/).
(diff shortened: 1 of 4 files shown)


zbuc (approved):
LGTM

hdevalence:
When I tried running this with `docker-compose up --build` I ran into an error (waiting to rebuild / repro) about missing `glibc` versions, but I think that's unrelated to the content of the PR...

hdevalence (approved):

PR #198, opened 2021-11-30
feat: decouple address version from testnet ID

redshiftzero:
Proposed format for v0 Penumbra addresses

### DIFF
--- a/crypto/src/address.rs
+++ b/crypto/src/address.rs
@@ -8,7 +8,7 @@ use crate::{fmd, ka, keys::Diversifier, Fq};
 
 pub const CURRENT_CHAIN_ID: &str = "penumbra-tn001";
 /// Human-readable prefix in the address.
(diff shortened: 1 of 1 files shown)


crypto/src/address.rs
```diff
@@ -8,7 +8,7 @@ use crate::{fmd, ka, keys::Diversifier, Fq};
 
 pub const CURRENT_CHAIN_ID: &str = "penumbra-tn001";
 /// Human-readable prefix in the address.
-pub const CURRENT_NETWORK_ID: &str = "penumbra_tn001_";
+pub const CURRENT_NETWORK_ID: &str = "penumbra_testv0_";
```
hdevalence, line 11: what about `penumbratv0`? that way it's a bit shorter and there's no underscores. it'll run into the leading `1` but i don't think that's really a problem

redshiftzero: hmm true, better to have shorter addresses, will update

PR #202, opened 2021-11-30
pd: propagate errors from worker tasks

hdevalence:
We should error out if either service errors, rather than continuing to run.


kwf (approved):
I agree with this PR!

PR #204, opened 2021-12-01
Revert "feat: integrate metrics"

redshiftzero:
Reverts penumbra-zone/penumbra#196

Others hit the glibc issue reported in #196 , so reverting this until we diagnose


redshiftzero:
Closing, the glibc issue was unrelated

PR #207, opened 2021-12-01
test: de-duplicate transaction verification logic 

### LINKED ISSUE #189: Deduplicate `pd/src/verify.rs` and `crypto/src/transaction.rs` tests
We've ended up with two implementations of transaction verification checks, one in the unit tests of `crypto/src/transaction.rs` and another in the `StatelessTransactionExt` of the `pd` crate.  The unit tests are incomplete, and don't, e.g., check the binding signature. Instead, the unit tests should use the `verify_stateless` method to check that the transaction is valid.  This means that either (1) the `StatelessTransactionExt` logic should move to where the unit tests are (in `penumbra_crypto`), or (2) the unit tests should move to `pd` where the verification logic is.

I think (2) is cleaner, since the `StatelessTransactionExt` definition is coupled to the data model of `pd` (via `PendingTransaction`, which it produces as a verification result). Also, transaction verification doesn't need to be done outside of `pd`, since it will only be done by validator nodes.

This should be done after #188, which touches this logic.

redshiftzero:
Closes #189

### DIFF
--- a/crypto/src/transaction.rs
+++ b/crypto/src/transaction.rs
@@ -293,10 +293,8 @@ mod tests {
 
     use crate::keys::SpendKey;
     use crate::memo::MemoPlaintext;
-    use crate::merkle::{Tree, TreeExt};
     use crate::transaction::Error;
-    use crate::{note, Fq, Note, Value};
-    use incrementalmerkletree::Frontier;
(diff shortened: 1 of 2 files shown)


hdevalence (approved):
LGTM

PR #213, opened 2021-12-01
190: Use option types rather than booleans for verification return signatures

### LINKED ISSUE #190: Use `Result` rather than `bool` for verification checks.
Currently, we use a mix of `Result`-valued and `bool`-valued verification checks.  This is somewhat dangerous, because it opens the possibility of bugs arising from the mismatch between the two styles, and inconvenient, because the `false` bools can't be wrapped with additional context like a `Result` can.  We should ensure that verification checks return a `Result` (even a `Result<(),()>`, though preferably a `Result<(), SomeError>`) rather than a `bool`.

zbuc:
Closes #190 

There are some cases where the deeper verification error is discarded and a generic error is bubbled up, I'm not sure if we want to address these or not but I commented them.

### DIFF
--- a/crypto/src/proofs/transparent.rs
+++ b/crypto/src/proofs/transparent.rs
@@ -16,6 +16,24 @@ use crate::{
 pub enum Error {
     #[error("Invalid spend auth randomizer")]
     InvalidSpendAuthRandomizer,
+    #[error("Note commitment mismatch")]
+    NoteCommitmentMismatch,
+    #[error("Transmission key mismatch")]
+    TransmissionKeyMismatch,
+    #[error("Value commitment mismatch")]
+    ValueCommitmentMismatch,
+    #[error("Ephemeral public key mismatch")]
+    EphemeralPublicKeyMismatch,
+    #[error("Must not be an identity")]
+    IdentityUnexpected,
+    #[error("Unexpected depth for merkle path")]
+    MerklePathMismatch,
+    #[error("Merkle root mismatch")]
+    MerkleRootMismatch,
+    #[error("Invalid diversified address")]
+    InvalidDiversifiedAddress,
+    #[error("Bad nullifier")]
+    BadNullifier,
 }
 
 /// Transparent proof for spending existing notes.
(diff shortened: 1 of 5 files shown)


pd/src/verify.rs
```diff
@@ -64,11 +64,17 @@ impl StatelessTransactionExt for Transaction {
...
+                        .body
+                        .proof
+                        .verify(
+                            output.body.value_commitment,
+                            output.body.note_commitment,
+                            output.body.ephemeral_key,
+                        )
+                        .is_err()
+                    {
+                        // TODO should the verification error be bubbled up here?
```
redshiftzero, line 77: As soon as we see an invalid proof in any action, we can bubble up the error as the entire transaction has failed

redshiftzero (approved):
LGTM, but I'll let @hdevalence give his thoughts on the FMD parts before merging

decaf377-fmd/src/detection.rs
```diff
@@ -78,29 +78,29 @@ impl DetectionKey {
     /// This function executes in constant time with respect to the detection
     /// key material, but short-circuits to return early on a false detection.
     #[allow(non_snake_case)]
-    pub fn examine(&self, clue: &Clue) -> bool {
+    pub fn examine(&self, clue: &Clue) -> Result<(), Error> {
```
hdevalence, line 81: This method should return a `bool` rather than a `Result`, since (unlike, e.g., a signature verification), getting a `false` result doesn't indicate an error, but just that the clue was not intended for this detection key.

crypto/src/proofs/transparent.rs
```diff
@@ -16,6 +16,24 @@ use crate::{
...
+    #[error("Must not be an identity")]
+    IdentityUnexpected,
+    #[error("Unexpected depth for merkle path")]
+    MerklePathMismatch,
+    #[error("Merkle root mismatch")]
+    MerkleRootMismatch,
+    #[error("Invalid diversified address")]
+    InvalidDiversifiedAddress,
+    #[error("Bad nullifier")]
+    BadNullifier,
```
hdevalence, line 36: These are probably useful for debugging, but we could eventually replace this whole error enum with a single unit error type, since a snark verification will just succeed or fail, revealing nothing about the witness data.  (Maybe this replacement happens when we implement snark proofs, just noting it for later).

hdevalence (changes requested):
This looks good except the FMD changes, noted in comment that the `examine` method returns a `bool` intentionally, but happy to merge this once that part is rolled back, other comments are just informational.

PR #216, opened 2021-12-01
add tx memo support to pcli

### LINKED ISSUE #214: pcli: allow user to specify memo when sending a transaction
Users can create a transaction using `pcli tx send`.

Right now we have a [dummy memo](https://github.com/penumbra-zone/penumbra/blob/main/wallet/src/state.rs#L147) and users aren't able to specify the memo for their transaction.  

We should:
- Let the user specify the memo in an optional command line argument in pcli. This would require adding a new optional string argument to `TxCmd` for the memo.
- If the memo is set, then pass into [`ClientState::add_transaction`]( https://github.com/penumbra-zone/penumbra/blob/main/wallet/src/state.rs#L123). The memo should be associated with the non-change output.

avahowell:
This PR adds support for users to set their transaction's `memo` field using pcli.

Closes #214

### DIFF
--- a/crypto/src/memo.rs
+++ b/crypto/src/memo.rs
@@ -4,7 +4,7 @@ use chacha20poly1305::{
     ChaCha20Poly1305, Key, Nonce,
 };
 use once_cell::sync::Lazy;
-use std::convert::TryInto;
+use std::convert::{TryFrom, TryInto};
 
 use crate::{ka, keys::IncomingViewingKey, note::derive_symmetric_key, Address};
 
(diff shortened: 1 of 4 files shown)


crypto/src/memo.rs
```diff
@@ -29,6 +29,20 @@ impl Default for MemoPlaintext {
...
 
+impl TryFrom<String> for MemoPlaintext {
+    type Error = anyhow::Error;
+
+    fn try_from(input: String) -> Result<MemoPlaintext, Self::Error> {
+        if input.len() > MEMO_LEN_BYTES {
+            return Err(anyhow::anyhow!("provided memo exceeds maximum memo size"));
+        }
+        let mut mp = [0u8; MEMO_LEN_BYTES];
+        mp.copy_from_slice(input.as_bytes());
```
hdevalence, line 40: In the future, we might want to have some leading-byte code or something indicating whether the memo is text or binary data, but this isn't important to worry about now.

hdevalence (approved):
LGTM!

PR #218, opened 2021-12-02
Don't treat first row of genesis CSV as a header

kwf:
Previously we were parsing the genesis CSV file accidentally treating the first row as a header,
which meant that the first genesis allocation would always be dropped. Now, we don't do this.


zbuc:
LGTM!

PR #219, opened 2021-12-02
fix panic on setting tx memo field from pcli

avahowell:
`copy_from_slice` requires the source and dest to be the same size, or else it will panic at runtime as discovered during testnet testing.


hdevalence (approved):
LGTM

PR #222, opened 2021-12-03
pd: refactor check_tx, deliver_tx to have better instrumentation

hdevalence:
This refactors the check_tx and deliver_tx methods to return a `Result<(),
Error>` rather than an ABCI response, and then wraps those methods in an
invocation that transforms the result into an ABCI response, so that:

* we can return any errors that occur in the ABCI `log` field;
* we have one place where we send the completion signal, so we don't forget it;
* the implementation of the functions is simpler;
* we can instrument the processing with a span with the transaction hash.

### DIFF
--- a/pd/src/app.rs
+++ b/pd/src/app.rs
@@ -444,7 +444,7 @@ impl Service<Request> for App {
                 return async move {
                     tracing::debug!("starting check_tx");
                     let rsp = rsp.await;
-                    tracing::debug!(?rsp);
+                    tracing::info!(?rsp);
                     let _ = finished_signal.send(());
                     match rsp {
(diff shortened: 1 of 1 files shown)


hdevalence:
Example output (with `RUST_LOG="pd=debug,penumbra_crypto=debug"`):
```
Dec 02 18:00:36.327 DEBUG check_tx{kind=New txid="d0f544534e9dc074db8af2241bd591c64860301d443a2f0b3d35228d91225bd0"}: pd::app: starting check_tx
Dec 02 18:00:36.327  INFO check_tx{kind=New txid="d0f544534e9dc074db8af2241bd591c64860301d443a2f0b3d35228d91225bd0"}: pd::app: rsp=Err(Transaction malformed)
```

PR #224, opened 2021-12-03
verify that nullifiers are not double spent in the same block

avahowell:
we use `check_tx` and `app.mempool_nullifiers` to protect the mempool from duplicate nullifiers, however when running `deliver_tx` we didn't verify that the transaction that was put into the block didn't spend the same nullifier as a different transaction in the same block.

this PR adds a branch which ignores txns that spend nullifiers that were already spent in the pending block.


redshiftzero (approved):
Excellent catch, thank you for the fix!

PR #233, opened 2021-12-07
add `App.validators`, staking module, Validator type and track initial validator set

### LINKED ISSUE #36: Track validators in ABCI application state
To do staking and delegation, the Penumbra application needs to track the validator set and maintain records of who the validators are.

avahowell:
This PR adds the `validators` field to `App`, which is a `BTreeMap<tendermint::PublicKey, Validator>` that will track the validator set. Note that this just pulls the initial validator state from the tendermint genesis state, we will need to add additional logic to the ABCI app once undelegations, slashing/evidence handling, etc, are implemented.

Closes #36

### DIFF
--- a/pd/src/app.rs
+++ b/pd/src/app.rs
@@ -1,5 +1,5 @@
 use std::{
-    collections::{BTreeSet, VecDeque},
+    collections::{BTreeMap, BTreeSet, VecDeque},
     future::Future,
     pin::Pin,
     sync::{Arc, Mutex},
@@ -30,6 +30,7 @@ use penumbra_crypto::{
 use crate::{
     db::schema,
     genesis::GenesisAppState,
+    staking::Validator,
     verify::{mark_genesis_as_verified, StatefulTransactionExt, StatelessTransactionExt},
     PendingBlock, RequestExt, State,
 };
@@ -76,6 +77,10 @@ pub struct App {
 
     /// Epoch duration in blocks
     epoch_duration: u64,
+
+    /// Contains the validator set, with each validator uniquely identified by their tendermint
+    /// public key.
+    validators: Arc<Mutex<BTreeMap<tendermint::PublicKey, Validator>>>,
 }
 
 impl App {
@@ -90,6 +95,7 @@ impl App {
             note_commitment_tree,
             recent_anchors: recent_anchors,
             mempool_nullifiers: Arc::new(Default::default()),
+            validators: Arc::new(Default::default()),
             pending_block: None,
             completion_tracker: Default::default(),
             epoch_duration: genesis_config.epoch_duration,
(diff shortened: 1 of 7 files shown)


pd/src/staking.rs
```diff
@@ -0,0 +1,29 @@
+use tendermint::{vote, PublicKey};
+
+/// Validator tracks the Penumbra validator's long-term consensus key (tm_pubkey), as well as their
+/// voting power.
+#[derive(Debug, Eq, PartialOrd, Ord)]
+pub struct Validator {
+    tm_pubkey: PublicKey,
+    voting_power: vote::Power,
+}
```
hdevalence, line 9: non-blocking on this PR, but in the future we might want to merge this into types in https://rustdoc.penumbra.zone/main/tendermint/validator/index.html (and/or modify them to suit our purposes; the `tendermint-rs` API isn't totally fleshed out yet)

pd/src/app.rs
```diff
@@ -74,6 +75,10 @@ pub struct App {
 
     /// Used to allow asynchronous requests to be processed sequentially.
     completion_tracker: CompletionTracker,
+
+    /// Contains the validator set, with each validator uniquely identified by their tendermint
+    /// public key.
+    validators: Arc<Mutex<BTreeMap<tendermint::PublicKey, Validator>>>,
```
hdevalence, line 81: It looks like at the moment we don't need to share access to the `validators`, since it's set in the body of `init_genesis` where there's already a `&mut self` available. Are we expecting to need to read or write it in other places where we have to do asynchronous processing?

hdevalence:
Do we need to save the validator info to the application state?

avahowell:
Yes, I think so, working on that now

pd/src/staking.rs
avahowell: I wasn't sure if we wanted to have a separate type specifically for Penumbra validators: right now this type is the same as the Tendermint type, but in the future I was imagining that we may want to have more penumbra-specific data or methods in this struct

pd/src/app.rs
avahowell: I think we'll need to write to this in `end_block` (for validator updates) and `begin_block` (for handling evidence and rewards)

pd/src/staking.rs
hdevalence: Yeah, starting with our own type is a good call, just wanted to make the connection for the future.

pd/src/app.rs
hdevalence: Also non-blocking for this PR, but just a comment for the future: we might want to explore splitting out these fields into an `Inner` type, so that rather than having multiple chunks of synchronization, we can wrap all of the state that we need to access in one mutex.

avahowell:
added code that saves validators to the application state on init genesis and retrieves them in the constructor of `App`

pd/src/staking.rs
```diff
@@ -0,0 +1,33 @@
+use tendermint::{vote, PublicKey};
+
+const PENUMBRA_BECH32_VALIDATOR_PREFIX: &str = "penumbravalpub";
```
hdevalence, line 3: non-blocking, but I think cosmos uses `cosmosvaloper` or something, maybe we should match it exactly?

pd/src/state.rs
```diff
@@ -232,6 +233,58 @@ INSERT INTO notes (
...
+        }
+
+        Ok(validators)
+    }
+
+    /// set the initial validator set, inserting each validator in `validators` into the state.
+    pub async fn set_initial_validators(
+        &self,
+        validators: &BTreeMap<tendermint::PublicKey, Validator>,
+    ) -> Result<()> {
```
hdevalence, line 269: non-blocking on this PR: we have some centralization of our database writes into the `commit` method, maybe it would make sense to merge this setup code with the setup code in #234 into a single method that would be more clearly a call-once method?

hdevalence (approved):
This LGTM, only comments are non-blocking. Currently there's a merge conflict so I'll rebase and de-conflict before merging.

PR #234, opened 2021-12-07
Add support for epoch duration to genesis, EndBlock

### LINKED ISSUE #38: Add notion of epochs
Penumbra's validator set changes should only happen at epoch boundaries. Adding a notion of epochs involves:

- [ ] adding an epoch duration parameter (presumably this goes in the genesis file #31 #17; it's probably not necessary to change during a deployment, but it's useful to be able to adjust, so that, e.g., testnets can run with faster epochs)
- [ ] adding a spot in the block processing logic for validator changes to be propagated back to tendermint at the end of an epoch (this can be a no-op at first)

zbuc:
Fixes #38

### DIFF
--- a/pd/src/app.rs
+++ b/pd/src/app.rs
@@ -84,6 +84,7 @@ impl App {
     #[instrument(skip(state))]
     pub async fn new(state: State) -> Result<Self, anyhow::Error> {
         let note_commitment_tree = state.note_commitment_tree().await?;
+        let genesis_config = state.genesis_configuration().await?;
         let recent_anchors = state.recent_anchors(NUM_RECENT_ANCHORS).await?;
         Ok(Self {
             state,
@@ -92,8 +93,7 @@ impl App {
             mempool_nullifiers: Arc::new(Default::default()),
             pending_block: None,
             completion_tracker: Default::default(),
-            // 8640 blocks is 1 day at 10 second block intervals
-            epoch_duration: 8640,
+            epoch_duration: genesis_config.epoch_duration,
         })
     }
 
--- a/pd/src/app.rs
+++ b/pd/src/app.rs
@@ -30,7 +30,7 @@ use penumbra_crypto::{
 
 use crate::{
     db::schema,
-    genesis::GenesisNote,
+    genesis::GenesisAppState,
     verify::{mark_genesis_as_verified, StatefulTransactionExt, StatelessTransactionExt},
     PendingBlock, State,
 };
@@ -74,6 +74,9 @@ pub struct App {
 
     /// Used to allow asynchronous requests to be processed sequentially.
     completion_tracker: CompletionTracker,
+
+    /// Epoch duration in blocks
+    epoch_duration: Option<i64>,
 }
 
 impl App {
@@ -89,6 +92,7 @@ impl App {
             mempool_nullifiers: Arc::new(Default::default()),
             pending_block: None,
             completion_tracker: Default::default(),
+            epoch_duration: None,
         })
     }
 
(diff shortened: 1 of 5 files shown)


zbuc:
I left a comment -- I wasn't sure if we'd need a mutex around the epoch duration. I don't think so, since it should only ever be set once in `InitChain` and remain constant afterwards.

pd/src/app.rs
```diff
@@ -74,6 +74,9 @@ pub struct App {
 
     /// Used to allow asynchronous requests to be processed sequentially.
     completion_tracker: CompletionTracker,
+
+    /// Epoch duration in blocks
+    epoch_duration: Option<i64>,
```
hdevalence, line 79: What does it mean if this is `None`?

zbuc: If it's `None` it means that the App has been instantiated but hasn't received an `InitChain` message yet.

hdevalence: Is it important to ensure that we crash on app data that doesn't specify an epoch duration, or could we set a default one in the constructor and then overwrite it in the `InitChain` handler? If there's not a particular reason we don't want to have a default, it might be cleaner to know that the value is there so we don't have to constantly unwrap it.

hdevalence:
> I left a comment -- I wasn't sure if we'd need a mutex around the epoch duration. I don't think so, since it should only ever be set once in `InitChain` and remain constant afterwards.

Yeah, I think you're right, since the epoch duration is held exclusively by the `App`, which isn't shared across threads. But if it were required, the code would fail to compile.

pd/src/genesis.rs
```diff
@@ -67,6 +67,15 @@ impl FromStr for GenesisAddr {
     }
 }
 
+#[serde_as]
+#[derive(Deserialize, Serialize, Debug, PartialEq, Eq)]
+pub struct GenesisAppState {
+    /// Initial notes
+    pub notes: Vec<GenesisNote>,
+    /// Epoch duration in terms of blocks
+    pub epoch_duration: i64,
```
hdevalence, line 76: Could we make this a `u64` and figure out how to isolate the signedness of the block heights we get, rather than propagating the signedness outwards?

pd/src/app.rs
zbuc: > Is it important to ensure that we crash on app data that doesn't specify an epoch duration, or could we set a default one in the constructor and then overwrite it in the InitChain handler?

I think I like the idea of specifying a default, but none of the other data within the `InitChain` `app_state` currently has a default. Should we also supply a default for the genesis notes in code?

hdevalence: We can require that the field is present in the `InitChain` `app_state`, so we know we'll overwrite whatever default we set in the code (and the default in the code is just ensuring that we're not constantly option-unwrapping).  Or, we could let the other fields be optional and have default values. A reasonable default for the genesis notes would be an empty list, although this isn't useful for setting up a chain.

zbuc:
@hdevalence OK this is ready for re-review

pd/src/app.rs
```diff
@@ -102,14 +107,14 @@ impl App {
         genesis_block.set_height(0);
 
         // Note that errors cannot be handled in InitChain, the application must crash.
-        let genesis: Vec<GenesisNote> = serde_json::from_slice(&init_chain.app_state_bytes)
+        let genesis: GenesisAppState = serde_json::from_slice(&init_chain.app_state_bytes)
```
hdevalence, line 110: Non-blocking: I think it would be better to use bincode here, since we already use bincode for the note commitment tree, but it's not a big deal either way.

pd/src/app.rs
```diff
@@ -322,6 +336,16 @@ impl App {
...
 
+        if end.height < 0 {
+            panic!("block height should never be negative");
+        }
+
+        // TODO: if necessary, set the EndBlock response to add validators
+        // at the epoch boundary
+        if end.height.unsigned_abs() % self.epoch_duration == 0 {
+            // Epoch boundary -- add/remove validators if necessary
+            println!("New epoch")
```
hdevalence, line 347:
```suggestion
            tracing::info!("new epoch");
```

hdevalence (approved):
LGTM!

PR #235, opened 2021-12-07
Pcli cleanup

plaidfinch:
This PR does a few pieces of cleanup to `pcli` and implements one bug fix:
- sychronization is called in one place in the main function, eliminating multiple duplicate code blocks
- the determination of whether a command requires pre-synchronization is done in a separate function
- if starting from height zero, the INFO messages will use round 1000s rather than X999s
- refactor the `addr` subcommands to eliminate code duplication
- *bug fix*: wallets were not archived when imported, only when generated locally; this is now resolved

### DIFF
--- a/pcli/src/main.rs
+++ b/pcli/src/main.rs
@@ -51,14 +51,22 @@ async fn main() -> Result<()> {
         }
     }
 
+    // Synchronize the wallet if the command requires it to be synchronized before it is run.
+    let state = if opt.cmd.needs_sync() {
+        let mut state = ClientStateFile::load(wallet_path.clone())?;
+        sync(
+            &mut state,
+            format!("http://{}:{}", opt.node, opt.lightwallet_port),
+        )
+        .await?;
+        Some(state)
+    } else {
+        None
+    };
+
     match opt.cmd {
         Command::Sync => {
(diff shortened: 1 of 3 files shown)


pcli/src/opt.rs
```diff
@@ -61,6 +74,19 @@ pub enum WalletCmd {
     Delete,
 }
 
+impl WalletCmd {
+    /// Determine if this command requires a network sync before it executes.
+    pub fn needs_sync(&self) -> bool {
+        match self {
```
plaidfinch, line 80: This match statement is deliberately verbose because it means that the addition of another subcommand forces future programmers to consider whether or not it should require synchronization beforehand, rather than automatically defaulting it to `false`.

pcli/src/main.rs
```diff
@@ -30,35 +30,34 @@ async fn main() -> Result<()> {
...
+        let mut state = ClientStateFile::load(wallet_path.clone())?;
+        sync(
+            &mut state,
+            format!("http://{}:{}", opt.node, opt.lightwallet_port),
+        )
+        .await?;
+        Some(state)
+    } else {
+        None
+    };
```
hdevalence, line 56: this is a nice way to handle the synchronization requirements

hdevalence (approved):
LGTM!

PR #237, opened 2021-12-08
Track pending spends and change in `pcli`

plaidfinch:
This PR implements pending-spend and pending-change tracking in `pcli`, along the way cleaning up a couple APIs in the crypto crate. High level changes:

- `Note::new` is now `Note::from_parts`
- `Note::generate` is a new function to create a freshly random note given a destination address and a typed value (this should be preferred over `Note::from_parts` where possible, because it is infallible since addresses are always well-formed)
- A new method on the transaction builder allows adding an output while getting the internally-generated note back as a clone. This is useful for implementing change-tracking, because when we add change outputs, we need to keep track of those notes.
- Sets of pending spends and pending change have been added to the wallet state, with backwards-compatible serialization that fills in the empty set if they are not present
- Pending spends and change are pruned based on their timeout whenever they are read from disk, and additionally after every wallet sync:
  + pending spends return to the unspent set (the transaction failed, so we should be allowed to try again with those inputs)
  + pending change is dropped on the floor (the outputs of the transaction never happened, so we should forget them)
- Every pending spend or change is associated with an expiry time set 60 seconds in the future from when the transaction which generated those spends/change was created. The same timestamp for expiry is used for every pending piece of any given transaction, which means that both the pending spend and pending change from a transaction will drop from view atomically in the case of a transaction failure, preventing weird transient observations in `pcli balance`
- Internally, to implement the balance display functionality of this PR, a new `UnspentNote` enum has been added, which wraps a `&Note` with a descriptor of whether it is `Ready`, `PendingSpend`, or `PendingChange`. This type is used in the output of the various functions that ask for lists of notes in the wallet, and because it is only holding a reference to a note, this decreases cloning. Only `Ready` notes are used when forming transactions, because even though we can predict the change for a pending transaction, we can't predict its position in the note commitment tree, which means we can't spend change until the transaction clears.
- **`pcli balance` has been updated to display an additional "Pending" column (which only appears when there is anything to show) that gives pending transactions in the format "+N (change) / -M (spent)"**

### DIFF
--- a/crypto/src/note.rs
+++ b/crypto/src/note.rs
@@ -66,7 +66,7 @@ pub enum Error {
 }
 
 impl Note {
-    pub fn new(
+    pub fn from_parts(
         diversifier: Diversifier,
         transmission_key: ka::Public,
         value: Value,
@@ -84,11 +84,15 @@ impl Note {
 
     /// Generate a fresh note representing the given value for the given destination address, with a
     /// random blinding factor.
-    pub fn fresh(rng: &mut impl Rng, address: crate::Address, value: Value) -> Result<Self, Error> {
+    pub fn fresh(
+        rng: &mut impl Rng,
+        address: &crate::Address,
+        value: Value,
+    ) -> Result<Self, Error> {
         let diversifier = *address.diversifier();
         let transmission_key = *address.transmission_key();
         let note_blinding = Fq::rand(rng);
-        Note::new(diversifier, transmission_key, value, note_blinding)
+        Note::from_parts(diversifier, transmission_key, value, note_blinding)
     }
 
     pub fn diversified_generator(&self) -> decaf377::Element {
@@ -297,7 +301,7 @@ impl TryFrom<&[u8]> for Note {
             .try_into()
             .map_err(|_| Error::NoteDeserializationError)?;
 
-        Note::new(
+        Note::from_parts(
             bytes[1..12]
                 .try_into()
                 .map_err(|_| Error::NoteDeserializationError)?,
@@ -423,7 +427,6 @@ mod tests {
     use super::*;
 
     use crate::keys::SpendKey;
-    use ark_ff::UniformRand;
     use rand_core::OsRng;
 
     #[test]
@@ -439,13 +442,7 @@ mod tests {
             amount: 10,
             asset_id: asset::Denom::from("penumbra").into(),
         };
-        let note = Note::new(
-            *dest.diversifier(),
-            *dest.transmission_key(),
-            value,
-            Fq::rand(&mut rng),
-        )
-        .expect("can create note");
+        let note = Note::fresh(&mut rng, &dest, value).expect("can create note");
         let esk = ka::Secret::new(&mut rng);
 
         let ciphertext = note.encrypt(&esk);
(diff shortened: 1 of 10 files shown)


crypto/src/note.rs
```diff
@@ -81,6 +82,19 @@ impl Note {
         })
     }
 
+    /// Generate a fresh note representing the given value for the given destination address, with a
+    /// random blinding factor.
+    pub fn fresh(
```
hdevalence, line 87: Could we call this `generate` to match some existing constructors?

crypto/src/note.rs
```diff
@@ -81,6 +82,19 @@ impl Note {
...
+        rng: &mut impl Rng,
+        address: &crate::Address,
+        value: Value,
+    ) -> Result<Self, Error> {
```
hdevalence, line 91: Does this need to be fallible? If we're passing in an `Address`, can we be sure that the transmission key it contains is always valid?

hdevalence: Yeah, the `Address` is supposed to maintain the invariant that its fields are valid, so the `.expect` can be folded into the body of this function, and the function can be made infallible.

crypto/src/transaction/builder.rs
```diff
@@ -76,41 +76,39 @@ impl Builder {
         self
     }
 
-    /// Create a new `Output` to create a new note.
-    pub fn add_output<R: RngCore + CryptoRng>(
+    /// Add a new note to the output
```
hdevalence, line 79: Can the doc comment mention that this function returns the note that was added to the transaction?

crypto/src/transaction/builder.rs
```diff
@@ -119,7 +117,21 @@ impl Builder {
             ovk_wrapped_key,
         });
 
-        self
+        (note, self)
+    }
+
+    /// Create a new `Output`, implicitly creating a new note for it and encrypting the provided
+    /// [`MemoPlaintext`] with a fresh ephemeral secret key.
```
hdevalence, line 124: similarly to the above, maybe the doc comments should be unified, with this one noting that it is chainable?

wallet/src/state.rs
```diff
@@ -42,13 +52,46 @@ pub struct ClientState {
     wallet: Wallet,
 }
 
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub enum PendingCommitment {
```
hdevalence, line 56: Could this be `PendingNoteCommitment`?

plaidfinch:
Per discussion with @hdevalence, there's no reasonable performance justification for normalizing the timeouts by having a separate set for them, since there will at any given time be maybe 2 transactions in flight. This means we can inline the timeouts into the sets of pending notes and scan those whole sets during pruning, which means there are fewer invariants to maintain. This should be fixed before merging this PR.

plaidfinch:
> Per discussion with @hdevalence, there's no reasonable performance justification for normalizing the timeouts by having a separate set for them, since there will at any given time be maybe 2 transactions in flight. This means we can inline the timeouts into the sets of pending notes and scan those whole sets during pruning, which means there are fewer invariants to maintain. This should be fixed before merging this PR.

This has now been resolved, in 1fd9391.

hdevalence:
Did an optimistic rollup, let's see how well this works on a new "Euporie" testnet.

Issue #57, opened 2021-09-16
Choose a minimum denomination of the staking token


hdevalence:
This is the analogue of satoshi for BTC, wei for ETH, uATOM for ATOM, etc.

BTC: 1e-8 BTC
ETH: 1e-18 ETH
ATOM: 1e-6 ATOM

Choosing 1e-6 fits with the rest of the Cosmos ecosystem and is probably good enough. 1e-18 seems really excessive.

redshiftzero:
This was done in #283, the minimum denomination is 1e-6

PR #241, opened 2021-12-08
feat: support coin metadata including minimum denomination

### LINKED ISSUE #178: Figure out how to track base denomination per-asset
> We should figure out how to keep track of what the base denomination of each asset is, since we'll want to be recording integral values of `upen` or whatever, but displaying as `9.23238921` or something.

_Originally posted by @hdevalence in https://github.com/penumbra-zone/penumbra/issues/176#issuecomment-976016313_

redshiftzero:
Closes #178
Related to #57 

* Adds a new required key to the genesis file that specifies assets and their metadata, including alternative denominations:

```
"assets": [
      {
        "description": "The native staking token of the Penumbra zone.",
        "denom_units": [
          {
            "denom": "upenumbra",
            "exponent": 0,
            "aliases": [
              "micropenumbra"
            ]
          },
          {
            "denom": "mpenumbra",
            "exponent": 3,
            "aliases": [
              "millipenumbra"
            ]
          },
          {
            "denom": "penumbra",
            "exponent": 6,
            "aliases": []
          }
        ],
        "base": "upenumbra",
        "display": "penumbra"
      }
    ]
```

* Adds an `asset::Metadata` struct to the crypto crate, this is based on [ADR-024](https://docs.cosmos.network/master/architecture/adr-024-coin-metadata.html).
* Modifies the thin wallet protocol to return the full asset metadata instead of the `asset::Denom`.
* pcli now stores in its asset registry the full asset metadata (so heads up if testing you'll need to blow away the contents of the `asset_registry` key in the wallet state file, or the whole wallet state file if you prefer) 

Made two followups so far: #244 #243

### DIFF
--- a/crypto/src/asset.rs
+++ b/crypto/src/asset.rs
@@ -118,6 +118,7 @@ pub struct Metadata {
     /// The base unit of the asset, e.g. `upenumbra`. Calculations should be done with the base.
     pub base: String,
     /// The unit that should be used in displaying e.g. balances, e.g. `penumbra`.
+    /// The `Denom` (and thus the `Id`) for the asset is derived from the `display`.
     pub display: String,
 }
 
--- a/crypto/src/asset.rs
+++ b/crypto/src/asset.rs
@@ -4,6 +4,7 @@ use std::convert::{TryFrom, TryInto};
 use ark_ff::fields::PrimeField;
 use decaf377::FieldExt;
 use once_cell::sync::Lazy;
+use serde::{Deserialize, Serialize};
 
 use crate::Fq;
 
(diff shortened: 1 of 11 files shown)


crypto/src/asset.rs
```diff
@@ -89,6 +96,31 @@ impl Id {
...
+
+/// Metadata about each asset including the minimum denomination.
+///
+/// Based on [ADR-024](https://docs.cosmos.network/master/architecture/adr-024-coin-metadata.html).
+#[derive(Serialize, Deserialize, Clone, Debug, Eq, PartialEq)]
+pub struct Metadata {
+    /// The description of this asset, e.g. `The native token for the Penumbra zone.`
+    pub description: String,
+    /// A list of alternative denominations that can be used, e.g. `upenumbra`, `penumbra`.
+    pub denom_units: Vec<DenomUnit>,
```
zbuc, line 117: For validator staking token assets we'll probably only have a single denomination, right?

redshiftzero: For validators I think we'll want to mirror whatever we decide on for `penumbra`, e.g. so I could imagine there eventually being a `udPEN-{validatorkey}` if we have `upenumbra` etc. In terms of how to define the metadata for all the delegation assets, I think we'll want some template metadata that we can reuse for every delegation token, I'll file an issue about this

hdevalence:
How does the `Metadata` and `DenomUnit` relate to the `Denom` and `Id` types we already have? How do we determine which denomination is the primary one and get the corresponding asset id?

redshiftzero:
I think we should use the metadata's `base` as the `Denom`, I'll add a comment to this effect in the code

hdevalence:
I pushed a rebase of this branch onto main as `178-assetlists_r1`: https://github.com/penumbra-zone/penumbra/tree/178-assetlists_r1

redshiftzero:
This is on hold for now in light of further conversation in #178

redshiftzero:
Superseded by #283

PR #242, opened 2021-12-08
Add ValidatorAssetId type for creating validator staking token asset IDs based on validator pubkey

### LINKED ISSUE #37: Create asset identifier for each validator's staking token
Penumbra's staking design treats all delegations to a particular validator as fungible, and provides native liquid staking tokens associated to each validator.   To record these in the shielded pool (#27), we need a way to create an asset identifier linked to the validator's identity.

zbuc:
Closes #37

### DIFF
--- a/crypto/src/asset.rs
+++ b/crypto/src/asset.rs
@@ -1,12 +1,16 @@
 //! Asset types and identifiers.
+use once_cell::sync::Lazy;
 use std::convert::{TryFrom, TryInto};
+
+use tendermint::PublicKey;
 
 use ark_ff::fields::PrimeField;
 use decaf377::FieldExt;
-use once_cell::sync::Lazy;
 
 use crate::Fq;
 
+pub const NATIVE_ASSET_DENOM: &str = "penumbra";
+pub const PENUMBRA_BECH32_VALIDATOR_PREFIX: &str = "penumbravalpub";
 /// An identifier for an IBC asset type.
 ///
 /// This is similar to, but different from, the design in [ADR001].  As in
@@ -27,6 +31,21 @@ use crate::Fq;
 /// https://github.com/cosmos/ibc-go/blob/main/docs/architecture/adr-001-coin-source-tracing.md
 #[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
 pub struct Id(pub Fq);
+
+/// Validator-specific unique asset identifier, used for identifying staking tokens
+/// associated with a particular validator.
+pub struct ValidatorAssetId(pub Fq);
+
+impl From<&PublicKey> for ValidatorAssetId {
+    fn from(pk: &PublicKey) -> ValidatorAssetId {
+        let validator_asset_denom = format!(
+            "d{}-{}",
+            NATIVE_ASSET_DENOM,
+            pk.to_bech32(PENUMBRA_BECH32_VALIDATOR_PREFIX)
+        );
+        ValidatorAssetId::from(Denom(validator_asset_denom))
+    }
+}
 
 pub struct Denom(pub String);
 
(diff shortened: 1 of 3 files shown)


crypto/src/asset.rs
```diff
@@ -44,6 +63,19 @@ impl From<&str> for Denom {
     }
 }
 
+impl From<Denom> for ValidatorAssetId {
+    fn from(denom: Denom) -> ValidatorAssetId {
+        // Convert an asset name to an asset ID by hashing to a scalar
+        ValidatorAssetId(Fq::from_le_bytes_mod_order(
+            // XXX choice of hash function?
+            blake2b_simd::Params::default()
+                .personal(b"Penumbra_AssetID")
```
zbuc, line 72: should this be a different string? i copied this from the regular asset ID impl

crypto/src/asset.rs
```diff
@@ -28,6 +32,21 @@ use crate::Fq;
 #[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
 pub struct Id(pub Fq);
 
+/// Validator-specific unique asset identifier, used for identifying staking tokens
+/// associated with a particular validator.
+pub struct ValidatorAssetId(pub Fq);
```
hdevalence, line 37: In order to record the delegation tokens with the existing transaction model, we need to have an ordinary `asset::Id`, so rather than building a custom validator-specific ID type, I think it would be better to have a validator-specific `ValidatorDenom` type that would have an ordinary `asset::Id`.

The validator-specific denomination type could be a parsed representation, storing the validator's identity internally with methods to produce the corresponding `Id` and `Denom`, as well as the identity. It would probably also be useful to have some way to try to parse a `Denom` to find out whether or not it's a delegation token.

It might be worth thinking through this design with https://github.com/penumbra-zone/penumbra/pull/241#issuecomment-989558425 and https://github.com/penumbra-zone/penumbra/issues/244

crypto/src/asset.rs
hdevalence: I think this should probably be replaced by a call to the normal asset ID impl: https://github.com/penumbra-zone/penumbra/pull/242/files#r765482232

crypto/src/asset.rs
zbuc: I took a second pass at this by implementing a new `ValidatorAsset` type with `denom -> Denom` and `id -> Id` methods, and added a `is_validator_denom(Denom) -> bool` function as well.

crypto/src/asset.rs
```diff
@@ -28,8 +34,62 @@ use crate::Fq;
...
+    let caps = re.captures(&d.0);
+    if caps.is_none() {
+        // There was no bech32 address captured
+        return false;
+    }
+
+    let bech32_address = caps.unwrap().get(1).map_or("", |m| m.as_str());
+
+    re.is_match(&d.0) && bech32::decode(bech32_address.to_lowercase()).is_ok()
+}
```
hdevalence, line 83: This is doing the work of parsing the address but discarding it to return a bool, could it be refactored into an `impl FromStr` that would return the parsed data?  Then we could have `impl TryFrom<Denom>` that just calls `.from_str()` on the `Denom`'s inner string.

(I think it's better to put the parsing as a `FromStr` impl and have the `TryFrom<Denom>` forward to that rather than the other way around, since it means that we can use `.from_str()` to parse from any borrowed string, rather than just from an already-constructed `Denom`)

crypto/src/asset.rs
```diff
@@ -28,8 +34,62 @@ use crate::Fq;
...
+
 pub struct Denom(pub String);
 
+impl std::fmt::Display for Denom {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "{}", self.0)
```
hdevalence, line 89: Ideally `Display` and `FromStr` roundtrip, which I think would happen if we display the associated `Denom`:

```suggestion
        write!(f, "{}", self.denom().0)
```

crypto/src/asset.rs
```diff
@@ -28,8 +34,62 @@ use crate::Fq;
 #[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
 pub struct Id(pub Fq);
 
+/// Validator-specific unique assets, used for identifying staking tokens
+/// associated with a particular validator.
+pub struct ValidatorAsset {
```
hdevalence, line 39: Could we call it a `ValidatorDenom`, so the name is more clear that it's a validator-specific analog of the `Denom` type?

hdevalence (approved):
Had some suggestions but otherwise LGTM!

crypto/src/asset.rs
hdevalence: comment obsoleted by https://github.com/penumbra-zone/penumbra/issues/178#issuecomment-992214707

hdevalence:
#290 rebases this on top of the new structure in #283

PR #246, opened 2021-12-08
Remove zero outputs

plaidfinch:
Now, the transaction builder does not produce an output if that particular output would be equal to zero.

This prevents silliness like requiring you to have at least one penumbra-denominated note in hand in order to send a zero-fee transaction, because that note needs to be split into shares of (0% fee, 100% change) in order to create the (zero!) fee output.


hdevalence (approved):
LGTM

PR #248, opened 2021-12-09
track validator commission rate, commission address, and unclaimed reward

### LINKED ISSUE #40: Track validator-associated data
There are at least two important pieces of data to associate to a validator:

- a shielded address for the the validator's commission (#39);
- the validator's commission percentage;
- the validator's unclaimed reward balance.

To start, these can probably be defined at genesis, but eventually, validators should be able to update them. Determining how that should happen is out-of-scope for this issue.

avahowell:
This PR adds support for loading initial validator state from initial application state, and extends our `Validator` struct to include the commission address, commission rate, and unclaimed reward.

Closes #40

### DIFF
--- a/crypto/src/asset.rs
+++ b/crypto/src/asset.rs
@@ -38,6 +38,26 @@ impl std::fmt::Debug for Id {
     }
 }
 
+impl std::fmt::Display for Id {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        use ark_ff::BigInteger;
+        let bytes = self.0.into_repr().to_bytes_le();
+        f.write_fmt(format_args!("asset::Id({})", hex::encode(&bytes)))
+    }
+}
+
+impl std::str::FromStr for Id {
+    type Err = anyhow::Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let bytes: [u8; 32] = s.as_bytes().try_into().map_err(|_| {
+            anyhow::anyhow!("could not deserialize Asset ID: input vec is not 32 bytes")
+        })?;
+        let inner = Fq::from_bytes(bytes)?;
+        Ok(Id(inner))
+    }
+}
+
 impl From<&str> for Denom {
     fn from(strin: &str) -> Denom {
         Denom(strin.to_string())
--- a/pd/src/staking.rs
+++ b/pd/src/staking.rs
@@ -1,5 +1,3 @@
-use num::bigint;
-use num::rational;
 use penumbra_crypto::{Address, Value};
 use serde::{Deserialize, Serialize};
 use tendermint::{vote, PublicKey};
@@ -23,7 +21,7 @@ pub struct Validator {
 
     /// commission_rate is the portion of staking rewards that go to the validator (as opposed to
     /// the delegators).
-    pub commission_rate: rational::Ratio<bigint::BigInt>,
+    pub commission_rate_bps: u16,
 
     /// unclaimed_reward is the amount of commission that the validator has yet to claim.
     pub unclaimed_reward: Value,
@@ -40,14 +38,14 @@ impl Validator {
         pubkey: PublicKey,
         voting_power: vote::Power,
         commission_address: Address,
-        commission_rate: rational::Ratio<bigint::BigInt>,
+        commission_rate_bps: u16,
         unclaimed_reward: Value,
     ) -> Validator {
         Validator {
             tm_pubkey: pubkey,
             voting_power,
             commission_address,
-            commission_rate,
+            commission_rate_bps,
             unclaimed_reward,
         }
     }
--- a/pd/src/staking.rs
+++ b/pd/src/staking.rs
@@ -36,10 +36,19 @@ impl PartialEq for Validator {
 }
 
 impl Validator {
-    pub fn new(pubkey: PublicKey, voting_power: vote::Power) -> Validator {
+    pub fn new(
+        pubkey: PublicKey,
+        voting_power: vote::Power,
+        commission_address: Address,
+        commission_rate: rational::Ratio<bigint::BigInt>,
+        unclaimed_reward: Value,
+    ) -> Validator {
         Validator {
             tm_pubkey: pubkey,
             voting_power,
+            commission_address,
+            commission_rate,
+            unclaimed_reward,
         }
     }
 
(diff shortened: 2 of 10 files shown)


crypto/src/asset.rs
```diff
@@ -38,6 +38,26 @@ impl std::fmt::Debug for Id {
...
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let bytes: [u8; 32] = s.as_bytes().try_into().map_err(|_| {
+            anyhow::anyhow!("could not deserialize Asset ID: input vec is not 32 bytes")
+        })?;
+        let inner = Fq::from_bytes(bytes)?;
+        Ok(Id(inner))
+    }
+}
+
```
hdevalence, line 60: non-blocking for this PR: should we use bech32 for asset ids? off the top of my head:

- pros: error-correction, prefix can identify what the data is (e.g., `passet1aedfea...`)
- cons: slightly longer encodings (it's already 32 bytes, does this matter? doubt people will be typing manually)

pd/src/staking.rs
```diff
@@ -1,12 +1,32 @@
...
 /// Validator tracks the Penumbra validator's long-term consensus key (tm_pubkey), as well as their
 /// voting power.
-#[derive(Debug, Eq, PartialOrd, Ord, Clone, Copy)]
+#[derive(Deserialize, Serialize, Debug, Eq, Clone)]
 pub struct Validator {
+    /// tm_pubkey is the validator's long-term Tendermint public key, where the private component
+    /// is used to sign blocks.
     tm_pubkey: PublicKey,
+
+    /// voting_power is the validator's current voting power.
```
hdevalence, line 16: Style nit: since Rustdoc will always display the doc string in the context of the item it attaches to, here and on the other fields it'd be better to structure the doc without the name, like
```suggestion

    /// The validator's current voting power.
```

pd/src/staking.rs
```diff
@@ -1,12 +1,32 @@
...
+use num::rational;
+use penumbra_crypto::{Address, Value};
+use serde::{Deserialize, Serialize};
 use tendermint::{vote, PublicKey};
 
 const PENUMBRA_BECH32_VALIDATOR_PREFIX: &str = "penumbravalpub";
 /// Validator tracks the Penumbra validator's long-term consensus key (tm_pubkey), as well as their
 /// voting power.
-#[derive(Debug, Eq, PartialOrd, Ord, Clone, Copy)]
+#[derive(Deserialize, Serialize, Debug, Eq, Clone)]
```
hdevalence, line 10: What are we using this with Serde for? To load it from the `genesis.json` file, right?

crypto/src/asset.rs
avahowell: i think this is a great idea

pd/src/staking.rs
```diff
@@ -1,12 +1,32 @@
...
     pub voting_power: vote::Power,
+
+    /// commission_address is the validator's shielded commission address, where they receive their
+    /// portion of the staking rewards.
+    #[serde(with = "serde_with::rust::display_fromstr")]
+    pub commission_address: Address,
+
+    /// commission_rate is the portion of staking rewards that go to the validator (as opposed to
+    /// the delegators).
+    pub commission_rate: rational::Ratio<bigint::BigInt>,
```
hdevalence, line 26: Rather than having an arbitrary-precision rational, what if we specified the commission rate as a fixed-precision integer? 1 basis point is probably enough precision, and 100% is 10,000 bps, so this could be
```rust
pub commission_rate_bps: u16
```
(though there's no particular reason to prefer a `u16` over a bigger integer other than that it fits the entire range)

I think this will make ensuring that we do correct rounding much easier later on.

crypto/src/asset.rs
avahowell: I don't think that the slightly longer encodings matter, and having a checksum is nice

pd/src/staking.rs
```diff
@@ -1,12 +1,32 @@
...
+
+    /// unclaimed_reward is the amount of commission that the validator has yet to claim.
+    pub unclaimed_reward: Value,
```
hdevalence, line 29: Rather than tracking the unclaimed reward, could we just insert the appropriate notes rewarding the previous epoch into the NCT in the first block of the next epoch?  At the moment it doesn't look like the code in this PR is actually updating the `unclaimed_reward` anyways, so we could potentially just drop it now?

pd/src/staking.rs
```diff
@@ -1,12 +1,32 @@
...
+    /// tm_pubkey is the validator's long-term Tendermint public key, where the private component
+    /// is used to sign blocks.
     tm_pubkey: PublicKey,
+
+    /// voting_power is the validator's current voting power.
     pub voting_power: vote::Power,
+
+    /// commission_address is the validator's shielded commission address, where they receive their
+    /// portion of the staking rewards.
+    #[serde(with = "serde_with::rust::display_fromstr")]
```
hdevalence, line 21: (non-blocking on this PR) Should we implement `Serialize`, `Deserialize` for `Address` so that we don't need this annotation on any struct using them that wants to use Serde?

hdevalence (changes requested):
This looks good, I think it would be better to track the rate using fixed-precision integers (basis points seem good enough) and, if I'm understanding correctly, I think we could drop the `unclaimed_reward` field.

The other comments are non-blocking, and we could sort them out in future PRs

pd/src/staking.rs
avahowell: This sounds good, I implemented it this way due to (a) being very paranoid about future arithmetic rounding issues and (b) this seems to be how the Cosmos SDK does it (using Go's `big.Int`.) Arbitrary precision rationals are probably overkill here

hdevalence: Yeah, that makes sense, and definitely better than using a float or something, but I think that the rounding behavior will be easier to analyze with fixed-point arithmetic.

pd/src/staking.rs
avahowell: Yes, we're loading `Validator`s from the application genesis state in JSON format.

pd/src/staking.rs
avahowell: oh neat!

hdevalence (approved):

hdevalence:
Merged and moved the rest of the non-blocking comments into followup issues!

PR #251, opened 2021-12-09
Fixes #250 by adding --offline option to `pcli balance`

### LINKED ISSUE #250: Add offline mode for `pcli balance`
It should be possible to check your locally cached wallet state in the absence of a network connection; right now we unconditionally try to connect to a node, and fail if we can't.

plaidfinch:
Now we can check what the balance is in the currently cached wallet.


zbuc:
LGTM

PR #252, opened 2021-12-09
Compact block enhancement

### LINKED ISSUE #245: Record note positions in database and build additional indices for nullifier and note columns
Our chain scanning procedure supplies a stream of `CompactBlock`s to the client, which uses them to scan for new notes and to update their local copy of the note commitment tree.  Clients append each note commitment to the note commitment tree in sequence, and mark note commitments they're interested in (so that they can produce authentication paths later).  This imposes ordering constraints on the stream of blocks: the (compact) blocks must be in order, and the note data within each block must also be in order.

However, we currently only get this behavior by accident, because our `notes` table doesn't record the position of each note, so we can't explicitly order the notes by position.  To fix this, we should:

- [x] add a `position bigint` column to the `notes` table, corresponding to a `u64` in Rust (though we don't need to worry about signedness and can just `.unwrap()` conversions to the Postgres `i64` type, since our merkle trees are restricted to smaller sizes already);
- [x] create a `struct PositionedNoteData { pub position: u64, pub data: NoteData }` in `pd::verify` representing a positioned note;
- [x] change the `pub notes: BTreeMap<note::Commitment, NoteData>,` in `PendingBlock` to `pub notes: BTreeMap<note::Commitment, PositionedNoteData>`;
- [x] write the `position` into the database in `State::commit_block`;
- [x] change the `PendingBlock::add_transaction` method to compute the position immediately after appending the note commitment to the NCT, and saving it in the `notes` field, with something like the following:
```rust
        for (note_commitment, data) in transaction.new_notes {
            self.note_commitment_tree.append(&note_commitment);
            let (position, _) = self.note_commitment_tree
                    .authentication_path(&note_commitment)
                    .expect("we just appended this commitment");
            self.notes.insert(note_commitment, PositionedNoteData { position, data });
        }
```

We also want to be able to efficiently query by position and height, so we should:

- [x] add indices to the `position` and `height` columns of the `notes` table;
- [x] add an index to the `height` column of the `nullifiers` table.

aubrey-dyer:
Closes #245 , includes refactor of `compact_block` in `state.rs` to utilize a stream over a range of heights


pd/src/db.rs
```diff
@@ -14,6 +14,9 @@ pub async fn init_tables(db: &Pool<Postgres>) -> Result<()> {
         query_file!("src/db/notes.sql"),
         query_file!("src/db/nullifiers.sql"),
         query_file!("src/db/validators.sql"),
+        query_file!("src/db/indexes/notes_height_idx.sql"),
+        query_file!("src/db/indexes/notes_position_idx.sql"),
+        query_file!("src/db/indexes/nullifiers_height_idx.sql"),
```
hdevalence, line 19: what's the motivation to keep the index creation commands separate from the tables they're indexing?

aubrey-dyer: as far as I can tell, this is a limitation in the existing `query_file!` pattern - I would like to combine them, but `query_file!` is apparently generating prepared statements which forbid multiple commands in a single file. @plaidfinch do you happen to know an alternative way to do this that might let us consolidate the create table and index creation files?

hdevalence: No need to fix it, I was just curious!

plaidfinch: @aubrey-dyer The alternative way would be to use `sqlx`'s explicit migration infrastructure: https://docs.rs/sqlx/0.5.9/sqlx/macro.migrate.html. Perhaps we can look into that?

PR #254, opened 2021-12-09
245 data enhancements

### LINKED ISSUE #245: Record note positions in database and build additional indices for nullifier and note columns

aubrey-dyer:
Closes #245

### DIFF
--- a/pd/sqlx-data.json
+++ b/pd/sqlx-data.json
@@ -63,6 +63,16 @@
       ]
     }
   },
+  "25bd791f06a593a77395aa005a9abba0cfc3374be8c392cbf6b744100302b541": {
+    "query": "CREATE INDEX ON notes (position)",
+    "describe": {
+      "columns": [],
+      "parameters": {
+        "Left": []
+      },
+      "nullable": []
+    }
+  },
   "3cbb441b10f4a068c16dd3a40accd689be26f2a6619b363fd966ee10e740f07b": {
     "query": " INSERT INTO assets ( asset_id, denom) VALUES ($1, $2)",
     "describe": {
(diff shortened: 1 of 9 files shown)


hdevalence:
It looks like this doesn't change the query used for the `CompactBlock` construction, should we change it from
```
"SELECT note_commitment, ephemeral_key, encrypted_note FROM notes WHERE height = $1"
```
to something like
```
"SELECT note_commitment, ephemeral_key, encrypted_note FROM notes WHERE height = $1 ORDER BY position"
```
in order to ensure that the fragments are being returned in the order they are supposed to be inserted into the tree (least position to highest position)?

hdevalence (approved):
LGTM!

Issue #255, opened 2021-12-10
Implement `Serialize`, `Deserialize` for `Address`


hdevalence:
We should implement `Serialize`, `Deserialize` for `Address`, so that we don't need annotations on any struct using them that wants to use Serde.  The `Serialize` and `Deserialize` impls should convert to/from strings using `Display`/`FromStr`.

_Originally posted by @hdevalence in https://github.com/penumbra-zone/penumbra/pull/248#discussion_r766153933_

redshiftzero:
Closed by #258

PR #258, opened 2021-12-10
pd: rework and rename `create-genesis` to `create-genesis-template`

hdevalence:
Our create-genesis subcommand was serving two roles:

1. using command line parameters to generate JSON output that could be put into
the `app_state` field of the Tendermint `genesis.json` file;
2. constructing genesis notes given an input of genesis allocations.

The second role was only necessary because we made the `app_state` record
notes, rather than recording allocations directly, even though they have the
same information.  Instead, we can record the allocations as part of the
genesis state, and compute the notes as needed.  (Digging into this revealed
that the genesis transaction was being created non-deterministically, which
would cause an instant consensus divergence on a multi-node system; that's
fixed along the way).

The first role is not tenable as the scope of the Penumbra configuration
increases.  For instance, we now need to specify details about the validator
set.  Currently, it's not possible to feed the output of `create-genesis`
directly into the `genesis.json` file, because not all the fields are included.
This makes running nodes difficult.

Adding more and more command line options to the tool that generates the
config data, and then having people record those options in a script or
something, is much less ergonomic than just editing the config data directly.

To edit the config data directly, though, it's helpful to have a complete
example of the format.  So, the `create-genesis` command is renamed to
`create-genesis-template`, and it generates an editable example file.

Now that we specify validator info in the `app_state` together with the genesis
allocations, we'll need to change our setup for deployments so that the
`genesis.json` file is specified in advance.  This means that we'll need to
change the setup so that validator identity keys are chosen in advance (and
written into the genesis file).

### DIFF
--- a/crypto/src/transaction/genesis.rs
+++ b/crypto/src/transaction/genesis.rs
@@ -1,17 +1,14 @@
-use ark_ff::UniformRand;
-use rand_core::{CryptoRng, RngCore};
-use std::ops::Deref;
+use ark_ff::{One, Zero};
 
 use super::Error;
 use crate::note::OVK_WRAPPED_LEN_BYTES;
-use crate::rdsa::{Binding, Signature, SigningKey};
 use crate::{
     action::{output, Action},
     ka,
     memo::{MemoCiphertext, MEMO_CIPHERTEXT_LEN_BYTES},
     merkle,
     transaction::{Fee, Transaction, TransactionBody},
-    value, Fr, Note, Output,
+    Fr, Note, Output,
 };
 
 /// Used to construct a Penumbra transaction from genesis notes.
@@ -22,6 +19,10 @@ use crate::{
 ///
 /// The `GenesisBuilder` has no way to create spends, only outputs, and
 /// allows for a non-zero value balance.
+///
+/// While the genesis transaction has the same form as normal transactions,
+/// its outputs are not private, and the `GenesisBuilder` uses constant values
+/// instead of RNG outputs.
 pub struct GenesisBuilder {
     // Actions we'll perform in this transaction.
     pub actions: Vec<Action>,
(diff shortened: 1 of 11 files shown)


hdevalence:
Closes #255

hdevalence:
Output:
```
// Edit the following template according to your needs

{
  "allocations": [
    {
      "amount": 1000000,
      "denom": "penumbra",
      "address": "penumbrav0t1vg7v2keg023zc3zev052enwl2xd4n06sauusvlkx0xmctj07rfwgynfp8nf3td4dasysdpndmg8p4letzdff3wwz3t54a9w5aagrzujgx847st4xt2qpzncxr4h40n"
    },
    {
      "amount": 10000,
      "denom": "gm",
      "address": "penumbrav0t1m5nq5l5xlgdtm3kugwxvxf2ule62aezx2kusu5w95w6vxs445h706z4ltnzlvla3jq5qmahfqnfdnl80763xjsl03stygfkv48q28furhhyxyz8neqr4juss083vgg"
    },
    {
      "amount": 1000,
      "denom": "cubes",
      "address": "penumbrav0t1vrfyeusup7cd9xr3sq0fqtrp6p75ek6hkxy50yx7flfe77eyt3vtc5yckg0gkrpv7vlqxvn2e4pj7xnyv74ylqqz5kvyr3hp9v9hc0pdgdaxzc9jldaqc2g3ylnxm2"
    }
  ],
  "epoch_duration": 300,
  "validators": [
    {
      "tm_pubkey": {
        "type": "tendermint/PubKeyEd25519",
        "value": "DwVK4+6DWFTvbv+goBXpDCAc250SEJTp5RLgMw/PYbo="
      },
      "voting_power": "100",
      "commission_address": "penumbrav0t1sww0gfz0tcywdnphtt6v3htlqpdgnjwrwhfwgylvwpvrc00lpqt9437lp02g6uhals3q9gqlglu5cacf0xzwlnc92z5kfujffveyj47csvfwsx7necs4geg8a7a03z",
      "commission_rate_bps": 200
    }
  ]
}
```

crypto/src/transaction/genesis.rs
```diff
@@ -43,13 +44,17 @@ pub struct GenesisBuilder {
 
 impl GenesisBuilder {
     /// Create a new `Output` for the genesis note.
-    pub fn add_output<R: RngCore + CryptoRng>(&mut self, rng: &mut R, note: Note) {
```
redshiftzero, line 46: thanks for catching this

redshiftzero (approved):
This LGTM if it's workable with the deployment process (cc @zbuc since I'm not knowledgeable enough in that process)

zbuc:
We will need to slightly tweak the deploy process, I'll put out a PR later
today.

PR #261, opened 2021-12-10
change asset ID encoding to use bech32 with prefix `passet`

### LINKED ISSUE #257: Use bech32 for asset IDs
The display impl for asset IDs introduced in #248 should be changed to use bech32 with some prefix; we could use `passet` for now (it's a bit shorter than`penumbraasset`).

_Originally posted by @hdevalence in https://github.com/penumbra-zone/penumbra/pull/248#discussion_r766137332_

avahowell:
Closes #257

### DIFF
--- a/crypto/src/asset.rs
+++ b/crypto/src/asset.rs
@@ -2,10 +2,13 @@
 use std::convert::{TryFrom, TryInto};
 
 use ark_ff::fields::PrimeField;
+use bech32::{Variant, ToBase32};
 use decaf377::FieldExt;
 use once_cell::sync::Lazy;
 
 use crate::Fq;
+
+const PENUMBRA_BECH32_ASSET_PREFIX: &str = "passet";
 
 /// An identifier for an IBC asset type.
 ///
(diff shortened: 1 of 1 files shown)


crypto/src/asset.rs
```diff
@@ -30,19 +33,24 @@ pub struct Id(pub Fq);
...
+            .expect("bech32 hrp is valid")
+    }
+}
+
 impl std::fmt::Debug for Id {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        use ark_ff::BigInteger;
-        let bytes = self.0.into_repr().to_bytes_le();
-        f.write_fmt(format_args!("asset::Id({})", hex::encode(&bytes)))
+        f.write_fmt(format_args!("asset::ID({})", self.to_bech32()))
```
avahowell, line 47: Do we want debug to have the same output as Display? The only difference as of this PR is that debug will print `asset::ID(...bech32 addr...)` instead of just `bech32 addr`. This seems potentially helpful so we can distinguish in logging, but the HRP should provide that for us as well.

hdevalence: Hmm, I think the HRP probably tells us what the type is, and keeping the debug output compact seems nice.

hdevalence (approved):
LGTM either way on the `Debug` point

PR #262, opened 2021-12-10
pd: save genesis data in json

hdevalence:
Fixes a crash while restarting the node because of bincode/serde-features incompatibility.


redshiftzero (approved):
LGTM

PR #263, opened 2021-12-10
Add Euporie testnet genesis file

hdevalence:
This sets the validator key to one I generated locally (saved in 1Password), we should figure out how to inject it into the deployment somehow (could be as simple as just copying it into the VPS?)


zbuc:
LGTM. For loading into the testnet deployment it should be sufficient to just manually place the genesis file and keyfile in the Docker volume prior to starting the containers. We could improve it by using the Docker secrets mechanism we are using for the database connection string to handle the validator private key, and script the initialization/copying of the genesis file to the tendermint volume.

testnets/002-euporie/genesis.json
```diff
@@ -0,0 +1,206 @@
...
+        "ed25519"
+      ]
+    },
+    "version": {
+      "app_version": "0"
+    }
+  },
+  "validators": [
+    {
+      "address": "567604C1B639ED8C3055E0218C9D8674223C25B1",
```
avahowell, line 26: Shouldn't this be a bech32 string?

avahowell: oh, this is the tendermint validators field, not the app state validators field

PR #264, opened 2021-12-10
add validator FundingStreams to allow multiple commission payout destinations

### LINKED ISSUE #256: Allow multiple commission addresses for each validator
Although it would diverge from the current spec, it'd be useful to be able to specify multiple commission addresses (to allow validators to dedicate portions of their commission to particular funding streams).

To do that, we could change to having a `Vec` of (address, rate) pairs. It would probably be useful to have a separate struct for that data, maybe something like
```rust
pub struct FundingStream {
    pub address: Address,
    pub rate_bps: u16,
}
```
and then have the `Validator` have a `Vec<FundingStream>`.

(On naming, the reason to use "funding stream" over "commission" is that "commission" sounds like something that just benefits the validator, while "funding stream" is more generic).

Moving from one commission address to multiple funding streams requires some DB changes, and possibly a new table, so we could queue this up for some future testnet.

We should also update the protocol spec to describe multiple funding streams rather than a single commission.

_Originally posted by @hdevalence in https://github.com/penumbra-zone/penumbra/pull/248#discussion_r766151054_

avahowell:
Closes #256

### DIFF
--- a/pd/src/db/validators.sql
+++ b/pd/src/db/validators.sql
@@ -1,6 +1,5 @@
 CREATE TABLE IF NOT EXISTS validators (
     tm_pubkey bytea NOT NULL PRIMARY KEY,
     voting_power bigint NOT NULL,
-    commission_address varchar NOT NULL,
-    commission_rate bigint NOT NULL
+    funding_streams bytea NOT NULL
 )
--- a/pd/src/db/validators.sql
+++ b/pd/src/db/validators.sql
@@ -1,5 +1,4 @@
 CREATE TABLE IF NOT EXISTS validators (
     tm_pubkey bytea NOT NULL PRIMARY KEY,
-    voting_power bigint NOT NULL,
-    funding_streams bytea NOT NULL
+    voting_power bigint NOT NULL
 )
--- a/pd/src/staking.rs
+++ b/pd/src/staking.rs
@@ -5,7 +5,7 @@ use tendermint::{vote, PublicKey};
 const PENUMBRA_BECH32_VALIDATOR_PREFIX: &str = "penumbravalpub";
 
 /// A destination for a portion of a validator's commission of staking rewards.
-#[derive(Debug, Deserialize, Serialize, PartialEq, Eq, Clone)]
+#[derive(Debug, Deserialize, Serialize, PartialEq, Eq, Clone, Copy)]
 pub struct FundingStream {
     /// The destinatination address for the funding stream..
     pub address: Address,
--- a/pd/src/staking.rs
+++ b/pd/src/staking.rs
@@ -10,8 +10,8 @@ pub struct FundingStream {
     /// The destinatination address for the funding stream..
     pub address: Address,
 
-    /// The portion (in terms of basis points) of the validator's total commission that goes to
-    /// this funding stream.
+    /// The portion (in terms of [basis points](https://en.wikipedia.org/wiki/Basis_point)) of the
+    /// validator's total staking reward that goes to this funding stream.
     pub rate_bps: u16,
 }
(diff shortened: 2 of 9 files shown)


pd/src/db/validators.sql
```diff
@@ -1,6 +1,5 @@
 CREATE TABLE IF NOT EXISTS validators (
     tm_pubkey bytea NOT NULL PRIMARY KEY,
     voting_power bigint NOT NULL,
-    commission_address varchar NOT NULL,
-    commission_rate bigint NOT NULL
+    funding_streams bytea NOT NULL
```
hdevalence, line 4: Rather than storing the funding streams opaquely, could we keep a first-class table in the database, so we can query it later if we want to do other stuff with it?  I'm not sure what the right way to structure the table would be, since I guess we want a many-to-one relationship between funding streams and validators (cc @aubrey-dyer)

pd/src/staking.rs
```diff
@@ -3,6 +3,17 @@ use serde::{Deserialize, Serialize};
...
+
+/// A destination for a portion of a validator's commission of staking rewards.
+#[derive(Debug, Deserialize, Serialize, PartialEq, Eq, Clone)]
+pub struct FundingStream {
+    /// The destinatination address for the funding stream..
+    pub address: Address,
+
+    /// The portion (in terms of basis points) of the validator's total commission that goes to
+    /// this funding stream.
+    pub rate_bps: u16,
```
hdevalence, line 15: If we specify each funding stream as a percentage of the validator's total commission, we have an extra invariant that the sum of the `rate_bps` must be 10,000.

Could we instead specify the rate for each funding stream in absolute terms, and let the validator's total commission be specified implicitly as the sum of the commission for each funding stream?

hdevalence: (Alternatively, if that's what this is already supposed to be doing -- the total rate disappeared from the `Validator` structure, so maybe it is? -- I guess the comment should be changed to say that it's a portion of the total rewards for that validator's delegation pool.

pd/src/db/validators.sql
aubrey-dyer: simplest way might just be adding a funding streams table with a key or indexed column that points back to the associated validator

hdevalence:
Could this PR also update the spec sections that describe the commission as being a single rate to reflect having multiple funding streams?

avahowell:
Switched to a first-class table for `validator_fundingstreams`, updated the docstrings about how funding stream rates work, and updated the validator rewards part of the protocol spec to match.

spec/stake/validator-rewards.md
```diff
@@ -1,9 +1,15 @@
 # Validator Rewards and Fees
 
-Validators declare a commission percentage $c_{v,e} \in [0, 1]$, which
-determines the spread between the base reward rate $r_e$ and the reward rate for
-their delegators $r_{v,e} = (1 - c_{v,e})r_e$, or equivalently $r_e = r_{v,e} +
-c_{v,e}r_e$.
+Validators declare a set of funding streams that comprise the destination of
+all of their staking rewards. Each funding stream contains a rate $ri \in
+[0,1]$ and a destination address $ai$.  The validator's total commission rate
+is defined as $c_{v,e} = \sum ri$, the sum of the rate of each funding stream.
```
hdevalence, line 6: not sure i follow the notation, what's the `i`?

hdevalence (approved):
LGTM mod spec notation (nonblocking)

spec/stake/validator-rewards.md
avahowell: the `ith` funding stream

PR #268, opened 2021-12-13
Fix pending spend display in `pcli`

plaidfinch:
This fixes several bugs with displaying pending spends and change in `pcli`:
- State is committed after a transaction is created, thus saving pending spends/change which were
  previously implicitly dropped
- Better tracing messages make it easier to understand the management of the pending sets
- Logic error corrected where pending set was not cleared if a transaction was confirmed on-chain,
  leading to confusing display for 60 seconds
- Logic error corrected where notes were simultaneously present in the unspent and pending set after a spend
- Use `BTreeMap`s to store the array of assets to print, which results in alphabetically sorted assets

It also improves the balance display, so that pending transactions are easier to understand:

```
 Asset          Total    Available  Pending                     
 gm             20010                                           
 gn             19970                                           
 penumbra       1199897  999973     +199924 (held), -29 (spent) 
 pizza          10000                                           
 tungsten_cube  100
```

### DIFF
--- a/pcli/src/main.rs
+++ b/pcli/src/main.rs
@@ -301,7 +301,7 @@ async fn main() -> Result<()> {
 
                 // Set up headers for the table (a "Pending" column will be added if there are any
                 // pending transactions)
-                headers = vec!["Asset", "Unspent"];
+                headers = vec!["Asset", "Total"];
             }
 
             // Add an "Available" and "Pending" column if there are any pending transactions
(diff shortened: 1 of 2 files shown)


hdevalence:
As a general comment, the tracing messages definitely seem better, but do they need to be at INFO level rather than DEBUG? We tell people to set INFO-level logging to find out information about what the program is doing, so the detailed info about how notes are being moved around the client side state machine might be better at DEBUG level.

aubrey-dyer (approved):
LGTM!

hdevalence (approved):

Issue #267, opened 2021-12-13
Change Docker setup to support `genesis.json` and/or `priv_validator_key.json`


hdevalence:
Despite the instructions in the README, it's not currently possible to use the Docker setup to run a Penumbra node, since the workflow doesn't provide a way to specify the `genesis.json` & `priv_validator_key.json` that the node should use.

I think the cleanest way to fix this would be to specify a directory on the host machine that would be used as the `$TMHOME/config` directory for the Tendermint container. Then, the `genesis.json` and `priv_validator_key.json` files for a local testnet can be placed in that directory and be picked up by Docker.

hdevalence:
One remaining question is how to handle initialization. Our `create-genesis-template` produces something that's supposed to be edited into the `app_state` key of an existing `genesis.json` file, and for making a single-node testnet, the easiest thing is to use the auto-generated validator key. If we connect a host directory to `$TMHOME/config`, can we run something that will just start the Tendermint docker image and run `tendermint init validator`, so that those files are generated and written back to the host directory, without starting all of the services before they've been validly configured?

zbuc:
> can we run something that will just start the Tendermint docker image and run tendermint init validator, so that those files are generated and written back to the host directory, without starting all of the services before they've been validly configured?

 This is similar to what the Tendermint Makefile does for multi-node testnet setup -- though they use a special `tendermint/localnode` Docker image's `testnet` command to generate everything: https://github.com/tendermint/tendermint/blob/master/Makefile#L249

zbuc:
Here's the implementation of the `testnet` command in Tendermint, I think we can probably make use of it: https://github.com/tendermint/tendermint/blob/93eb940dcdc37e959f054479e9741c190bd4ca6c/cmd/tendermint/commands/testnet.go#L44

It looks like it will handle generating the keys and base config files. I'll dig in and see how to integrate it in our process.

zbuc:
Running the default `testnet` command results in the following directory structure (from an empty directory structure):

```
build
├── config
│   ├── config.toml
│   ├── genesis.json
│   ├── node_key.json
│   └── priv_validator_key.json
├── config-template.toml
├── data
│   └── priv_validator_state.json
├── node0
│   ├── config
│   │   ├── config.toml
│   │   ├── genesis.json
│   │   ├── node_key.json
│   │   └── priv_validator_key.json
│   └── data
│       └── priv_validator_state.json
├── node1
│   ├── config
│   │   ├── config.toml
[...]

PR #278, opened 2021-12-14
Integrate create-genesis-template into validator setup script, groundwork for multi-node

zbuc:
The fresh start script now integrates the `create-genesis-template` command, uses a persistent directory mount for tendermint configs, and uses the tendermint testnet setup command.

Closes: #267

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -19,20 +19,77 @@ services:
       - "26667:26667"
 
   # The Tendermint node
-  tendermint:
+  tendermint-node0:
     image: "tendermint/tendermint:latest"
-    container_name: tendermint
+    container_name: tendermint-node0
     ports:
       - "26656:26656"
       - "26657:26657"
     volumes:
-      - "tendermint_data:/tendermint"
+      - ~/scratch/testnet_build/node0:/tendermint:Z
     command: start --proxy-app=tcp://pd:26658
+    environment:
+      - ID=0
+      - LOG=${LOG:-tendermint.log}
     depends_on:
       - pd
     networks:
       localnet:
         ipv4_address: 192.167.10.2
+
+  # The Tendermint node
+  # tendermint-node1:
+  #   image: "tendermint/tendermint:latest"
+  #   container_name: tendermint-node1
+  #   ports:
+  #     - "26659-26660:26656-26657"
+  #   volumes:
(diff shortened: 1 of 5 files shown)


pd/src/main.rs
```diff
@@ -192,8 +192,10 @@ async fn main() -> anyhow::Result<()> {
                 )],
             };
 
-            println!("// Edit the following template according to your needs");
-            println!("\n{}\n", serde_json::to_string_pretty(&app_state)?);
+            // Print this comment to stderr so stdout can be redirected as
+            // syntactically valid JSON
+            eprintln!("// Edit the following template according to your needs\n");
```
hdevalence, line 197: 👍

docker-compose.yml
```diff
@@ -19,28 +19,82 @@ services:
...
+  #   command: start --proxy-app=tcp://pd:26658
+  #   environment:
+  #     - ID=3
+  #     - LOG=${LOG:-tendermint.log}
+  #   depends_on:
+  #     - pd
+  #   networks:
+  #     localnet:
+  #       ipv4_address: 192.167.10.5
+
```
hdevalence, line 93: The idea would be that these could be uncommented to run a multi-node testnet, right (mod having to run separate `pd` and `postgres` instances)?

zbuc: Exactly

hdevalence:
How much of the python script would be required if we didn't inject the `app_state` into the genesis file and instead had people copy it in manually? As is, they'll already have to edit the genesis data to match their situation, so we could potentially save on scripting. (I'm not totally sure what the first part is doing with Tendermint).

zbuc:
The first part is using the tendermint testnet setup script to create the
4-node configuration. The tendermint testnet image published on DockerHub
doesn't include the tendermint binary and expects it to be built and placed
on the host filesystem.

The Python script wouldn't really be necessary if we required manual
editing. It just makes it easier to get up and running.

hdevalence (approved):
LGTM!

One thing I wonder about is whether it might be better to tell people to edit the genesis file manually rather than having the script patch it automatically. If we patch it automatically, I think there might be an expectation that the template would work as-is, but in fact, it has to be manually edited anyways in order to be useful (e.g., changing the addresses for genesis allocations, and so on).  But that's non-blocking on this.

PR #281, opened 2021-12-14
Add Epoch type

### LINKED ISSUE #280: Add concept of Epoch to the codebase
Currently there's no actual `Epoch` type or method for converting from block height to epoch.

To enable epoch-based exchange rates an `Epoch` type should be added (`stake/src/epoch.rs`) with a method to instantiate from a block height.

zbuc:
Closes #280

### DIFF
--- a/stake/src/epoch.rs
+++ b/stake/src/epoch.rs
@@ -15,14 +15,16 @@ impl Epoch {
             return Err(anyhow::anyhow!("block height should never be negative"));
         }
 
-        Epoch::from_blockheight_unsigned(block_height.unsigned_abs())
+        Ok(Epoch::from_blockheight_unsigned(
+            block_height.unsigned_abs(),
+        ))
     }
 
     /// from_blockheight_unsigned instantiates a new `Epoch` from a given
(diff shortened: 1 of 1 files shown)


stake/src/epoch.rs
```diff
@@ -0,0 +1,30 @@
...
+        }
+
+        Epoch::from_blockheight_unsigned(block_height.unsigned_abs())
+    }
+
+    /// from_blockheight_unsigned instantiates a new `Epoch` from a given
+    /// unsigned block height. Due to the implementation in tendermint using
+    /// signed representation for block height, we provide this
+    /// as well as a signed implemention (`from_blockheight`)
+    pub fn from_blockheight_unsigned(block_height: u64, epoch_duration: u64) -> Result<Self> {
```
avahowell, line 25: should this just return a `Self`, since it's infallible?

avahowell:
LGTM apart from the one comment

stake/src/epoch.rs
zbuc: TBH I'm not sure how this is passing the compiler at all, I didn't wrap it in an `Ok` enum... good catch

PR #282, opened 2021-12-14
Clear the mempool nullifier set at the end of each block

### LINKED ISSUE #229: Pending nullifiers are not properly cleared when a block is committed
In order to prevent double-spends within a single block, we have to track the set of all nullifiers in any transaction in that block and make sure to only approve one of those transactions (this is necessary because ABCI does not give us the hooks to prevent transactions from making it into a block in the first place if they double-spend). However, this set currently hangs around forever, which means that if a transaction is rejected for any reason, its nullifier sits in the pending set forever and prevents the associated note from ever being spent until a server reboot.

plaidfinch:
Fixes #229.

### DIFF
--- a/pd/src/app.rs
+++ b/pd/src/app.rs
@@ -356,6 +356,9 @@ impl App {
             panic!("block height should never be negative");
         }
 
+        // Clear the mempool nullifier set at the end of each block
+        self.mempool_nullifiers.lock().unwrap().clear();
+
         // TODO: if necessary, set the EndBlock response to add validators
         // at the epoch boundary
         if end.height.unsigned_abs() % self.epoch_duration == 0 {


avahowell:
LGTM, per our offline discussion I think that this should be fine

redshiftzero:
does [this stanza](https://github.com/penumbra-zone/penumbra/blob/main/pd/src/app.rs#L254) need to change?

hdevalence:
I don't totally follow where we got to on this, what's the reasoning on why this gives us the behavior we want?

avahowell:
I think that the `if request.kind == CheckTxKind::New {` does need to change if we want to use tendermint to re-add all of the nullifiers from pending transactions in tendermint's memory pool from to `mempool_nullifiers`.

From what I recall the reasoning that got us here was:

- we already prevent double spending nullifiers in the same block in any case with [this check[(https://github.com/penumbra-zone/penumbra/blob/main/pd/src/app.rs#L321)
- mempool_nullifiers only purpose is to allow us to filter out invalid transactions before block proposal, i.e. reduce the amount of stuff committed that we have to just ignore
- tendermint will re-call CheckTx with the remainder of the memory pool after the block is committed, allowing us to re-populate the mempool_nullifiers db (and not needing to persist the cache across multiple blocks).

So, the conclusion that we came to was to clear the cache at the end of every block, and allow tendermint's recheck behavior to repopulate it. This should be safe and not have the bug where a nullifier can be stuck in the cache prevent a note from being spent until the server is restarted

hdevalence:
> * we already prevent double spending nullifiers in the same block in any case with this check
> * mempool_nullifiers only purpose is to allow us to filter out invalid transactions before block proposal, i.e. reduce the amount of stuff committed that we have to just ignore

The way I think of the check to prevent double-spending nullifiers is as a stopgap fallback, that's only there because ABCI doesn't let us control block proposal precisely, and eventually we shouldn't have to have it (because we should have the validators not vote for blocks with duplicate nullifiers). 

In the meantime, while we still have Honest Nodes(TM), it would be preferable to not have to worry about whether or not some transaction included in a block is actually included in the chain state, or have to implement some code that handles that -- for instance, I'm not sure our current client software or whatever will correctly handle ignored transactions, but I don't really want to spend effort on that edge case, so ideally we could use the `mempool_nullifiers` to prevent it from ever happening.

PR #283, opened 2021-12-14
`Denom` refactoring

hdevalence:
Follows discussion in #178 and https://github.com/penumbra-zone/penumbra/issues/178#issuecomment-992214707

This PR doesn't make use of the new `DisplayDenom`s, but just replaces everything with `BaseDenom`. Adding support for doing unit conversions is not included.

This code has a nice external API but is a bit tricky internally, and it would be good to test it.


crypto/src/asset/denom.rs
```diff
@@ -0,0 +1,187 @@
...
+        BaseDenom {
+            inner: self.inner.clone(),
+        }
+    }
+
+    // TODO: API for actually working with displayed denominations?
+    // - format-style printing of amounts of base denom in displaydenom units
+    // - conversion of numbers? (tricky? do you use floats? or what fractions?)
+    //
+    // if we have a good enough api, do we need to have these methods?
```
hdevalence, line 142: Conversions that are safe:

- `u64` in `DisplayDenom` to `u64` in `BaseDenom` (could error on failure of checked arithmetic if the original quantity is too big to represent, but this is probably `.expect`able)
- `u64` in `BaseDenom` to `String` in `DisplayDenom` (decimal formatted output)
- `String` in `DisplayDenom` (decimal representation) to `u64` in `BaseDenom`

I think we could skip the first one and just do the last two, since this steers arithmetic into being done with the base denomination at all times.

crypto/src/asset/denom.rs
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    pub fn raw_name(&self) -> String {
+        self.inner
+            .units
+            .get(self.unit_index as usize)
+            .expect("there must be an entry for unit_index")
+            .denom
+            .clone()
+    }
```
hdevalence, line 159: These could probably be removed if we had good `String` conversions? (unsure)

crypto/src/asset/registry.rs
```diff
@@ -0,0 +1,220 @@
...
+                    ],
+                )
+            }) as for<'r> fn(&'r str) -> _,
+        )
+        .add_asset(
+            // TODO: this doesn't restrict the length of the bech32 encoding
+            // TODO: maybe the following would be a nicer format:
+            // udelegation_penumbravaloper1grgelyng2v6v3t8z87wu3sxgt9m5s03xfytvz7
+            // - "delegation" in the denom
+            // - underscore means no line breaks
```
hdevalence, line 196: cc @zbuc WDYT of this delegation token denomination format?

redshiftzero (approved):

redshiftzero:
After discussion in Discord, merging this optimistically, please open followup issues if you find problems with this

PR #286, opened 2021-12-15
pcli: add asset denom to error output when syncing asset registry

redshiftzero:
After merge of #283, `penumbra` is no longer a valid base denomination (in favor of `upenumbra`). This causes an error when syncing (discovered by @plaidfinch):

```
Error: invalid asset denomination: penumbra
```

This PR just adds debug output for the future to display which asset denom is invalid


plaidfinch:
Thanks, LGTM!

PR #288, opened 2021-12-15
Commit the wallet state to disk when the `ClientStateFile` is dropped

### LINKED ISSUE #277: Commit wallet state on drop
Automatically committing the walker state when the `ClientStateFile` is dropped will likely prevent future bugs where we forget to commit the state before exiting the program.

plaidfinch:
Closes #277.


hdevalence (approved):
Looks like a trivial conflict?

plaidfinch:
> Looks like a trivial conflict?

Yup! Just wanted to make sure those two lines ended up in the right order (they happened to).

PR #291, opened 2021-12-16
pd: sqlx fixes, update README

hdevalence:
This commit attempts to solve some of the problems with building sqlx's offline
mode by:

* using the sqlx migrations feature, so that we can reproducibly create the
  database structure that sqlx is checking against;

* updating the README to describe the process involved in regenerating the
  `sqlx-data.json` from scratch;

* updating the README to describe the process involved in creating a genesis
  file and running the different `pd` components manually.

We should also update the Docker-based workflow to take a `genesis.json` file
as an input, since unfortunately, there are necessarily manual edits required
to configure a local testnet (like filling in different addresses that should
have genesis tokens).


zbuc:
> We should also update the Docker-based workflow to take a genesis.json file
as an input

I think it does what you want, by means of using a local disk volume. The workflow sans script could be:

```console
$ # Generate the testnet Tendermint genesis JSON files and private keys:
$ docker run -v build_dir:/tendermint:Z tendermint/localnode testnet --config /etc/tendermint/config-template.toml --populate-persistent-peers=false --o . --starting-ip-address 192.167.10.2
$
$ # Generate a default `app_state` from the hardcoded template:
$  cargo run --bin pd --create-genesis-template
$
$ # Now edit if desired and copy/paste that `app_state` into the various `genesis.json` files you'll find in the `node` subdirectories of `build_dir`
```

The `setup_validator.py` script does this. It should maybe be renamed to `setup_testnet.py`. Or just rewritten in Rust. At this point, the Docker containers haven't been started, so you can manually edit the `genesis.json` files yourself prior to running `docker-compose up --build -d`.

There's *also* the `[docker_compose_freshstart.sh](https://github.com/penumbra-zone/penumbra/blob/main/scripts/docker_compose_freshstart.sh)` script, which does more and will handle completely wiping all testnet Docker containers, *and* running `docker-compose up --build -d` , which will cause the genesis data to be loaded into Tendermint and the firing of the `InitChain` message.

pd/src/state.rs
```diff
@@ -33,8 +33,9 @@ impl State {
     pub async fn connect(uri: &str) -> Result<Self> {
         tracing::info!("connecting to postgres");
         let pool = PgPoolOptions::new().max_connections(4).connect(uri).await?;
-        tracing::info!("building tables");
-        db::init_tables(&pool).await?;
+        tracing::info!("running migrations");
+        sqlx::migrate!("./migrations").run(&pool).await?;
```
zbuc, line 37: 💯

zbuc (approved):

hdevalence:
> > We should also update the Docker-based workflow to take a genesis.json file as an input
>
> I think it does what you want, by means of using a local disk volume.

Yeah, the underlying Docker setup does that already, I just meant the workflow description in the README.

Getting the Tendermint container configured is annoying and it's useful to have a script that helps, but there's probably always going to be some manual editing of the testnet config (specifying various addresses, editing the genesis distribution, etc), so ideally the instructions would have something like "do this script / edit the genesis data / start the docker-compose command".

PR #296, opened 2021-12-16
Stream compact blocks to client from database

plaidfinch:
Implements streaming for compact block requests, reducing N database queries to 2 streaming queries.


pd/src/wallet.rs
```diff
@@ -59,8 +60,11 @@ impl LightWallet for State {
         let (tx, rx) = mpsc::channel(100);
         tokio::spawn(
             async move {
-                for height in start_height..=end_height {
-                    let block = state.compact_block(height.into()).await;
+                while let Some(block) = state
+                    .compact_blocks(start_height.into(), end_height.into())
+                    .next()
+                    .await
+                {
```
hdevalence, line 67: We have this buffered channel and a spawned task hanging around from the previous implementation, is it possible to turn the Rust `Stream` from `compact_blocks` into a gRPC stream directly?

hdevalence:
Could we delete the previous `compact_block` method that fetches a single block? We don't use it any more, and if we did need a single block, we can always get it with a range of size 1.

plaidfinch:
> Could we delete the previous `compact_block` method that fetches a single block? We don't use it any more, and if we did need a single block, we can always get it with a range of size 1.

@hdevalence I'm not sure where you see that function; it's not in this branch as far as I can tell.

pd/src/wallet.rs
hdevalence: This doesn't work, but I think something like this could work: https://github.com/penumbra-zone/penumbra/commit/9592d5f52040c831f25e478b14aaafff4e1b592e

hdevalence:
> > Could we delete the previous `compact_block` method that fetches a single block? We don't use it any more, and if we did need a single block, we can always get it with a range of size 1.
> 
> @hdevalence I'm not sure where you see that function; it's not in this branch as far as I can tell.

Sorry, I misread the diff! Please disregard.

hdevalence:
Testing this locally looks like about a 7x speedup on sync times. Nice!

PR #299, opened 2021-12-17
Support recording validator rate changes per epoch

zbuc:


pd/migrations/20211216025603_init.sql
```diff
@@ -33,12 +33,23 @@ CREATE INDEX ON notes (position);
 CREATE INDEX ON notes (height);
 
 CREATE TABLE IF NOT EXISTS validators (
-    tm_pubkey bytea NOT NULL PRIMARY KEY,
-    voting_power bigint NOT NULL
+    tm_pubkey bytea NOT NULL PRIMARY KEY
 );
```
hdevalence, line 37: non-blocking + non-action, just noting for later: we'll probably want to add in extra metadata fields here, like a display name, website, etc.

pd/migrations/20211216025603_init.sql
```diff
@@ -33,12 +33,23 @@ CREATE INDEX ON notes (position);
...
 
 CREATE TABLE IF NOT EXISTS validator_fundingstreams (
     tm_pubkey bytea NOT NULL,
     address varchar NOT NULL,
     rate_bps bigint NOT NULL
 );
+CREATE TABLE IF NOT EXISTS base_rates (
```
hdevalence, line 44: this is a nit, but it'd be nice to have an extra newline here so that there's one "paragraph" per table + indexes 
```suggestion

CREATE TABLE IF NOT EXISTS base_rates (
```

pd/src/app.rs
```diff
@@ -87,6 +87,11 @@ impl App {
         let note_commitment_tree = state.note_commitment_tree().await?;
         let genesis_config = state.genesis_configuration().await?;
         let recent_anchors = state.recent_anchors(NUM_RECENT_ANCHORS).await?;
+        // TODO: This is the only place `state.validators()` is called -- updates to
+        // the validator list won't be available! I think it needs to be called again in `commit`
+        // prior to updating validator rates. _Or_ maybe not depending on the database interaction design.
+        // But we need to be careful to ensure the `App.validators` doesn't come out of sync with what's
+        // in the database.
```
hdevalence, line 94: this is a good concern, and (non-blocking on this PR) it's probably worth removing the `validators` field from the `App` for now, to prevent inconsistency.  in the event that we need to access the validator data more frequently, we can add it back and ensure consistency then.

pd/src/app.rs
```diff
@@ -348,22 +355,31 @@ impl App {
...
-        }
+        let is_epoch_boundary = Epoch::is_epoch_boundary(end.height, self.epoch_duration)
+            .expect("unhandleable error determining epoch boundary");
+        // set the flag on the pending block
+        self.pending_block
+            .as_mut()
+            .expect("pending_block must be Some in EndBlock")
+            .lock()
+            .unwrap()
+            .set_epoch_boundary(is_epoch_boundary);
```
hdevalence, line 375: Here the epoch is set independently from the height (they are set at the same time, but from the point of view of `PendingBlock::set...` they're independent), and the epoch boundary is set separately from either.

I guess this is because the mechanics of computing the epoch and the boundary are done in `Epoch` methods that use the `epoch_duration`, which isn't available inside `PendingBlock`.  But maybe it would be cleaner if `epoch_duration` were passed to `PendingBlock::new`, so that instead of having `set_height_and_epoch` as well as `set_epoch_boundary`, there could just be `set_height`, which would compute the epoch info internally.

This way, there's no possibility of misuse in the API, either the height is set or it isn't, and when it is, the epoch info is guaranteed to be correctly filled in.

pd/src/app.rs
```diff
@@ -348,22 +355,31 @@ impl App {
...
             .expect("pending_block must be Some in EndBlock")
             .lock()
             .unwrap()
-            .set_height(end.height);
+            .set_height_and_epoch(end.height, epoch);
 
-        if end.height < 0 {
-            panic!("block height should never be negative");
-        }
+        let is_epoch_boundary = Epoch::is_epoch_boundary(end.height, self.epoch_duration)
```
hdevalence, line 367: non-blocking on this PR, but maybe it would be cleaner to have `Epoch::start_height(&self) -> block::Height` and `Epoch::end_height(&self) -> block::Height` instead of `Epoch::is_epoch_boundary()`?

That way, it's still possible to test if something is the boundary (`height == epoch.start_height()`, etc), but it's unambiguous which side of the boundary the test is checking, whereas with `is_epoch_boundary` it's not immediate whether that's checking for the first or last block.

pd/src/state.rs
```diff
@@ -112,6 +109,34 @@ ON CONFLICT (id) DO UPDATE SET data = $1
...
+        if block.is_epoch_boundary.unwrap() {
+            // validator rates need updating on epoch boundaries
+            let validators = self.validators().await?;
+            for validator in validators {
+                tracing::info!("updating validator rates for validator: {:?}", validator.0);
+                // TODO @ava insert calls here
+                let validator_rate = 0;
+                let voting_power = 0;
+
+                let pubkey_str = serde_json::to_string(&validator.0)?;
```
hdevalence, line 124: non-blocking on this PR: should we try to use bech32 for validator identities consistently? (we can't use it in the DB at the moment because there's no round trip parsing upstream...)

pd/src/state.rs
```diff
@@ -112,6 +109,34 @@ ON CONFLICT (id) DO UPDATE SET data = $1
...
+
+                if block.epoch.is_none() {
+                    return Err(anyhow::anyhow!("EndBlock must be called prior to Commit, `epoch` was not set on the pending block"));
+                }
+                query!(
+                r#" INSERT INTO validator_rates ( epoch, validator_pubkey, validator_rate, voting_power ) VALUES ($1, $2, $3, $4)"#,
+                block.epoch.as_ref().unwrap().0 as i64,
```
hdevalence, line 131: nit: there's multiple places where `block.epoch` has to be handled, maybe it would be cleaner to get the inner value upfront once, like
```rust
let epoch = block.epoch.ok_or_else(|| {
    anyhow::anyhow!("EndBlock must be called prior to Commit, `epoch` was not set on the pending block")
})?;
```

pd/src/state.rs
```diff
@@ -289,14 +314,15 @@ INSERT INTO blobs (id, data) VALUES ('gc', $1)
 
         let mut validators: BTreeMap<tendermint::PublicKey, Validator> = BTreeMap::new();
 
-        let stored_validators = query!(r#"SELECT tm_pubkey, voting_power FROM validators"#)
+        let stored_validators = query!(r#"select tm_pubkey, validator_rates.voting_power FROM validators LEFT JOIN validator_rates ON validator_rates.validator_pubkey = validators.tm_pubkey;"#)
```
hdevalence, line 317: non-blocking: later, we should set a validator limit in the genesis app state, and use it to limit the query here (selecting the top N validators by voting power).

pd/src/state.rs
```diff
@@ -310,6 +336,8 @@ INSERT INTO blobs (id, data) VALUES ('gc', $1)
             // `i64` in sqlx. try_into uses the `TryFrom<i64>` implementation for voting power from
             // Tendermint, so will return an error if voting power is negative (and not silently
             // overflow).
+            // TODO the voting power is actually stored on the `validator_rates` table now so
+            // we need to join against that table
```
hdevalence, line 340: non-blocking: later, could this method be refactored so that it can get all the data it needs out of a single database query, rather than having to do multiple queries? (cc @aubrey-dyer)

stake/src/epoch.rs
```diff
@@ -2,10 +2,8 @@ use anyhow::{anyhow, Error};
 
 /// Epoch represents a given epoch for Penumbra and is used
 /// for calculation of staking exchange rates.
-#[derive(Debug, Eq, PartialEq)]
-pub struct Epoch {
-    epoch: u64,
-}
+#[derive(Debug, Eq, PartialEq, Clone)]
+pub struct Epoch(pub u64);
```
hdevalence, line 6: non-blocking: I think it could be helpful to change the Epoch definition to
```rust
pub struct Epoch {
    index: u64,
    duration: u64,
}
```
so that the `Epoch` is self-contained enough to be able to compute data like its start and end heights, and check whether a particular height is contained within the epoch.  My guess is that this would make stuff like https://github.com/penumbra-zone/penumbra/pull/299/files#r771231900 easier to manage.

The only complication is that we might want to serialize into the database by `Epoch::index()`, so deserialization would require having the epoch duration on hand.

pd/src/app.rs
hdevalence: xref https://github.com/penumbra-zone/penumbra/pull/299/files#r771246246

hdevalence (approved):
This is a great start, almost all of the comments are non-blocking and about future design improvements, so happy to merge this now and keep iterating. The only real near-term improvement I'd suggest is this one: https://github.com/penumbra-zone/penumbra/pull/299/files#r771231900 , but happy to merge this without it and fold it into future changes.

pd/migrations/20211216025603_init.sql
```diff
@@ -33,12 +33,24 @@ CREATE INDEX ON notes (position);
...
 
 CREATE TABLE IF NOT EXISTS validator_fundingstreams (
     tm_pubkey bytea NOT NULL,
     address varchar NOT NULL,
     rate_bps bigint NOT NULL
 );
+
+CREATE TABLE IF NOT EXISTS base_rates (
+    epoch bigint PRIMARY KEY,
+    base_rate bigint NOT NULL
```
avahowell, line 47: passive note/non-blocking: I think postgres's `bigint` type is actually an i64, our rates are specified as being `u64`, so we'd have to do a type conversion here when we pull these out of the database and check for the overflow condition. This shouldn't be an issue since our rates are specified with an implicit 10^8 denominator, the u64 typing is so that we can safely compute products of rates in the same type without having to worry about overflow.

pd/src/app.rs
avahowell: 👍

PR #301, opened 2021-12-17
specify fixed-point arithmetic for rate calculations

### LINKED ISSUE #271: Specify fixed-precision arithmetic for computing base and validator rates and voting power

avahowell:
this PR specifies fixed-point arithmetic for computing:

* base exchange rate
* commission rate
* validator reward rate
* validator exchange rate
* validator voting power adjustment

by adding an extra section to the Staking and Delegation part of the spec called "arithmetic".

Closes #271

### DIFF
--- a/spec/stake/arithmetic.md
+++ b/spec/stake/arithmetic.md
@@ -9,21 +9,49 @@ and plenty of headroom for additions.
 
 All integer values should be interpreted as unsigned 64-bit integers, with the
 exception of the validator's commission rate, which is a `u16` specified in
-terms of basis points.
+terms of basis points. All integer values, with the exception of the
+validator's commission rate, have an implicit denominator of $10^8$. 
+
+Throughout this spec *representations* are referred to as $\mathtt x$, where
+$\mathtt x = x \cdot 10^{8}$, and the *value* represented by representations is
+$x = \mathtt x \cdot 10^{-8}$.
+
+As an example, let's take a starting value $\mathtt x$ represented in our scheme (so,
+$x \cdot 10^8$) and compute its product with $y$, also represented by our fixed point
+scheme, so $y \cdot 10^8$. The product $\mathtt x \mathtt y$ is computed as 
+
+$$\left\lfloor \frac{(x 10^8)(y 10^8)}{10^{8}} \right\rfloor = \left\lfloor \frac{\mathtt x \mathtt y}{10^{8}} \right\rfloor$$
+
+Since both $x$ and $y$ are both representations and both have a factor of
+$10^8$, computing their product includes an extra factor of 10^8 which we
+divide out. All representations are `u64` in our scheme, and any fixed-point
+number or product of fixed point numbers with 8 digits fits well within 64
+bits.
+
+
+## Table of Contents
+
+* $\mathtt r_{e}$: the fixed-point representation of the base rate at epoch $e$.
+* $\mathtt \psi(e)$: the fixed-point representation of the base exchange rate at epoch $e$.
+* $s_{i, e}$: the $10^4$ denominated fixed-point representation of the funding rate of the validator's funding stream at index $i$ and epoch $e$.
+* $c_{v,e}$: the sum of the validator's commission rates, which are specified as a fixed point integer with an implicit $10^4$ denominator.
+* $\mathtt r_{v,e}$: the fixed-point representation of the validator-specific reward rate for validator $v$ at epoch $e$.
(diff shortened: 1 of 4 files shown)


avahowell:
the math here is in the same notation as the main part of the spec, but I also have a python implementation that I wrote in parallel to test things out, happy to include that in the spec as well if you think it would be helpful

spec/stake/arithmetic.md
```diff
@@ -0,0 +1,59 @@
...
+base rate $r_{e}$ is determined is not yet decided. For now, we can assume it is
+derived from the block header. $r_{e}$ is a fixed-precision `u64` integer with
+8 digits of precision.
+
+
+## Base Exchange Rate
+
+The base exchange rate, $\psi(e)$, can be safely computed as follows:
+
+$$\psi(e) = \prod_{0 \leq i < e} \lfloor \frac{(1e8 + r_i)}{1e8} \rfloor $$
```
hdevalence, line 26: Should this be something like
```
$$\psi(e) = \left\lfloor \frac {\psi(e-1) (10^8 + r_e)} {10^8} \right\rfloor$$
```
instead? The existing formula doesn't seem quite right, because (1) if we apply the rounding to the rate term, we'll be rounding to 1 and then taking a product of 1s and (2) we know we have to compute the rate iteratively in order to have consistent rounding behavior.

spec/stake/arithmetic.md
```diff
@@ -0,0 +1,59 @@
...
+To compute base reward rate, base exchange rate, validator-specific exchange
+rates, and total validator voting power, we need to carefully perform
+arithmetic to avoid issues with precision and rounding. We use explicitly
+specified fixed-precision arithmetic for this, with a precision of 8 digits.
+This allows outputs to fit in a u64, with all products fitting in the output
+and plenty of headroom for additions.
+
+All integer values should be interpreted as unsigned 64-bit integers, with the
+exception of the validator's commission rate, which is a `u16` specified in
+terms of basis points.
```
hdevalence, line 12: Just saying that they're `u64`s isn't enough to specify the interpretation, since we also need to specify that they have an implicit denominator of $10^8$.

I think it'd be useful to add a paragraph or something describing the interpretation of a `u64` integer value, with an example of what happens during multiplication: the value $x$ is represented as the integer $x 10^8$, so the product $xy$ is represented as the integer $xy 10^8$, which we compute from the representatives $x 10^8$ and $y 10^8$ as $\lfloor (x 10^8) (y 10^8) \cdot 10^{-8} \rfloor$. 

Maybe there's a better way to explain this than the above, but it'd be useful to have an explanation of why we're dividing by 10^8, it's because when we multiply two representatives, we get an extra factor of 10^8 that we have to divide out.

spec/stake/arithmetic.md
```diff
@@ -0,0 +1,59 @@
...
+
+To compute the validator's commission rate from its set of funding streams, the following calculation can be performed:
+
+$$c_{v,e} = \sum_{0 \leq i < n} (s_{i,e})$$
+
+Where $s$ is the set of rates defined by the funding streams, $s_{i,e}$ being
+the ith rate at epoch $e$.  $s_i$ should be a fixed-precision `u16` integer
+type, specified in terms of [basis
+points](https://en.wikipedia.org/wiki/Basis_point). The resulting $c_{v,e}$ is
+defined in terms of basis points (i.e., four digits of precision).
```
hdevalence, line 39: nit: this phrasing is slightly mixing between the units of measurement and the bits of precision; in our case, we're using integer values, so these are connected, but in principle the two concepts are doing different things (you could have only a few digits of precision, but for some arbitrary base quantity)

spec/stake/arithmetic.md
```diff
@@ -0,0 +1,59 @@
...
+type, specified in terms of [basis
+points](https://en.wikipedia.org/wiki/Basis_point). The resulting $c_{v,e}$ is
+defined in terms of basis points (i.e., four digits of precision). 
+
+
+## Validator Reward Rate
+
+To compute the validator's base reward rate, we compute the following:
+
+$$r_{v,e} = \lfloor \frac{(1e8 - (c_{v,e}*1e4))r_e}{1e8} \rfloor$$
```
hdevalence, line 46: Here and elsewhere, could we do `10^8` rather than `1e8` in the math formulas?

spec/stake/arithmetic.md
avahowell: good catch, I actually ended up implementing it this way in python but the change didn't make it back into the notation here

avahowell:
updated the spec:

* fixed the formulas to be incremental
* added more information about the scheme, value representations, and a table of contents

PR #302, opened 2021-12-17
Docker Fixes

zbuc:
No longer automatically start testnet with freshstart.sh, ensure DB is present and configured prior to build


hdevalence (approved):
LGTM, having an explicit step for customizing the testnet configuration is great.

PR #304, opened 2021-12-20
feat: add `DisplayDenom` formatting and parsing, fix display parsing bug

### LINKED ISSUE #292: Implement value parsing methods on `DisplayDenom`
In #283 we created `BaseDenom` and `DisplayDenom` types. We'd like to be able to accept and display values in the display denominations as well as the base denomination.

Currently, all of our internal numerical representations are always integral values in the base denomination, preventing accidentally mixing units.  This seems like a useful property to maintain, which suggests using only formatted string representations for the display denomination values.

To do this, we should add two methods:
```rust
impl DisplayDenom {
    pub fn format_value(&self, value: u64) -> String {
        // e.g., with exponent 6, 1782000 formats to 1.782
   }
   pub fn parse_value(&self, value: &str) -> Result<u64, Error> {
        // e.g., with exponent 6, 1.782 parses as 1782000
        // and 1.7820001 errors (cannot be represented exactly)
   }
}
```
and remove the existing `exponent` (not needed once this exists) and `raw_name` (provided by `Display`) methods.

Both formatting and parsing should be done using exact arithmetic, for instance (sketch):

* To format, compute the quotient and remainder `v = v1 * 10^e + v2`, then format as `format!("{}.{}", v1, v2)`;
* To parse, split the input at `.` into `left` and `right` (with `right` possibly empty), parse `left` and `right` as integers `v1` and `v2`, multiply `v2` by `10^(e - right.len())` (effectively padding with 0s on the right, erroring if `right.len() > e` to detect values that can't be represented exactly), and then compute `v = v1 * 10^e + v2`.

The computations during parsing should probably be done with checked arithmetic to detect overflows.

redshiftzero:
Closes #292, now working on adding property test

### DIFF
--- a/crypto/src/asset.rs
+++ b/crypto/src/asset.rs
@@ -10,6 +10,8 @@ pub use registry::{Registry, REGISTRY};
 
 #[cfg(test)]
 mod tests {
+    use proptest::prelude::*;
+
     use super::*;
 
     #[test]
@@ -88,5 +90,22 @@ mod tests {
         // not included in the hardcoded registry.
         let base_denom = REGISTRY.parse_base("cube").unwrap();
         assert_eq!(format!("{}", base_denom), "cube".to_string());
+    }
+
+    proptest! {
+        #[test]
+        fn displaydenom_parsing_formatting_roundtrip(
+            v: u32
+        ) {
+            let penumbra_display_denom = REGISTRY.parse_display("penumbra").unwrap();
+            let formatted = penumbra_display_denom.format_value(v.into());
+            let parsed = penumbra_display_denom.parse_value(&formatted);
+            assert_eq!(v, parsed.unwrap() as u32);
+
+            let mpenumbra_display_denom = REGISTRY.parse_display("mpenumbra").unwrap();
+            let formatted = mpenumbra_display_denom.format_value(v.into());
+            let parsed = mpenumbra_display_denom.parse_value(&formatted);
+            assert_eq!(v, parsed.unwrap() as u32);
+        }
     }
 }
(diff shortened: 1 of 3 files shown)


crypto/src/asset/registry.rs
```diff
@@ -82,24 +82,36 @@ impl Registry {
...
+        if self.parse_base(raw_denom).is_some() {
+            let base_inner = self.parse_base(raw_denom).unwrap().inner;
+
+            // The display version of the base denomination is always the last in the
+            // `units`.
+            let unit_index = base_inner.units.len() - 1;
+            return Some(DisplayDenom {
+                inner: base_inner,
+                unit_index,
+            });
```
hdevalence, line 94: I think this code could be simpler, by doing something like
```
if let Some(base_denom) = self.parse_base(raw_denom) {
    return Some(base_denom.units().last().expect("base denom always has at least one display denom"));
}
```
since the current code is constructing the whole base denomination twice and then re-constructing the display denom that's already present in it.

crypto/src/asset.rs
```diff
@@ -42,11 +46,66 @@ mod tests {
...
+        assert_eq!(
+            penumbra_display_denom.parse_value("1.782").unwrap(),
+            1782000
+        );
+
+        let mpenumbra_display_denom = REGISTRY.parse_display("mpenumbra").unwrap();
+        assert_eq!(
+            mpenumbra_display_denom.parse_value("1782").unwrap(),
+            1782000
+        );
```
hdevalence, line 78: Could be nice to check that
```
mpenumbra_display_denom.parse_value("1782.0001").is_err()
```
just to exercise the too-much-precision codepath.

hdevalence (approved):
LGTM mod comments!

crypto/src/asset/registry.rs
redshiftzero: thanks! this is simpler, updated

PR #311, opened 2021-12-21
Allow the wallet to be reset even if it is only partially parseable

plaidfinch:
This requires that the `wallet` field of the top-level object be parseable as a `Wallet`, but does not require any other structure. As a result, this will still not allow migrations between wallet state formats where the serialization of a `Wallet` changes, but this logic could be added later.

This also adds a check to make sure that if the wallet cannot be parsed after being written to disk, it will not be overwritten.


hdevalence:
This is a really clean solution, nice work

hdevalence (approved):

PR #314, opened 2021-12-22
pd: temporarily increase buffer size to work around probable tower-abci bug

hdevalence:
The `tower-abci` `Server` state machine's event loop selects over readiness of
either the stream of incoming requests, or the stream of responses to those
requests.  Once it receives a request, it checks which service should process
the request, and *then* waits for that service to become ready.  While it
waits, it doesn't drive the response futures to completion.

Increasing the buffer size adds more slack to the system and seems to make the
problem go away for now.


hdevalence:
Using just #313 didn't make the testnet hang go away, but #313 + this change did.

PR #316, opened 2021-12-25
Feature: "Galileo" Discord bot to dispense tokens automatically

plaidfinch:


hdevalence:
As a general comment, I think we might want the discord bot to live in a dedicated repo, rather than living in this one -- it feels conceptually pretty different from the rest of the software here.

plaidfinch:
> As a general comment, I think we might want the discord bot to live in a dedicated repo, rather than living in this one -- it feels conceptually pretty different from the rest of the software here.

I'm not sure I agree: since the Discord bot is to be yet another consumer of the `penumbra-wallet` API, its dissimilarity to `pcli` is merely at the front-end. Underneath the hood, the same dev experience advantages to sharing a workspace apply to Galileo as they do to pcli: namely, as we iterate on galileo, it's less cumbersome to make adjustments to the APIs upon which it depends. A PR that modifies a penumbra-wallet API to make it easier to use in Galileo can atomically contain both the changed code in Galileo and the changed code in penumbra-wallet, as well as changes to adapt all other dependents on that API. Whereas, if Galileo were to live in a separate repository, such a change would require synchronizing separate PR workflows across two different repositories. Does that seem like a worthwhile reason to have it live here?

plaidfinch:
> > As a general comment, I think we might want the discord bot to live in a dedicated repo, rather than living in this one -- it feels conceptually pretty different from the rest of the software here.
> 
> I'm not sure I agree: since the Discord bot is to be yet another consumer of the `penumbra-wallet` API, its dissimilarity to `pcli` is merely at the front-end. Underneath the hood, the same dev experience advantages to sharing a workspace apply to Galileo as they do to pcli: namely, as we iterate on galileo, it's less cumbersome to make adjustments to the APIs upon which it depends. A PR that modifies a penumbra-wallet API to make it easier to use in Galileo can atomically contain both the changed code in Galileo and the changed code in penumbra-wallet, as well as changes to adapt all other dependents on that API. Whereas, if Galileo were to live in a separate repository, such a change would require synchronizing separate PR workflows across two different repositories. Does that seem like a worthwhile reason to have it live here?

In particular, one such improvement has already been made: moving the synchronization logic from the `pcli` crate into `penumbra-wallet`, and parameterizing by a checkpointing strategy that can be customized by caller of that `sync` function to use whatever persistence mechanism they want. See commits 
cdf3a3f  through f45c562 in this PR.

hdevalence:
> Underneath the hood, the same dev experience advantages to sharing a workspace apply to Galileo as they do to pcli: namely, as we iterate on galileo, it's less cumbersome to make adjustments to the APIs upon which it depends.

Right, but the meta-goal of making this discord bot isn't to have a discord bot that can auto-dispense testnet tokens (we're not near the point where the work we'd be automating is worth the work of the automation), it's to have a test case for whether it's viable for third parties to use the APIs we provide to write software that interacts with the chain.  If we can't effectively use those APIs without being in a Penumbra monorepo, it seems like we're not getting a great test case for third-party API use.

> its dissimilarity to `pcli` is merely at the front-end...

> moving the synchronization logic from the `pcli` crate into `penumbra-wallet`, and parameterizing by a checkpointing strategy that can be customized by caller of that `sync` function to use whatever persistence mechanism they want. See commits cdf3a3f through f45c562 in this PR.

I'm not sure this is true, because I don't think `pcli` and `galileo` actually do have essentially similar structures.  On the one hand, `pcli` is intended to be a command-line tool that's only run occasionally, does one thing per invocation, and can't maintain any state except via serialization to disk.  On the other hand, `galileo` is a long-running daemon, does multiple things over time, and can maintain in-memory data as well as persistent data (though of course, it should be resilient to shutdowns).

These are fairly different contexts with different concerns, and while it's possible to create some generic code that can be specialized into one or the other, I'm not sure it's better to do that than to just have different code for different contexts, and I suspect that writing different code for different contexts first, *then* trying to extract the common behavior, will produce better results than trying to create a generic implementation upfront.

Looking just at the `pcli` code, for instance, the synchronization logic becomes bigger and more complicated over those commits -- it doubles in size, and gets split over two different crates, with the entrypoint being the `sync` function on the `pcli::{...}::ClientStateFile` (*not* the `sync` function on the `penumbra_wallet::{...}::ClientState`), the behavior being determined by a new `SyncEvery` inner structure with a trait implementation with multiple lifetime bounds and a custom workaround for the file locking behavior that has some new invariants, before finally passing into the `penumbra_wallet` code that does the work.

The `galileo` code doesn't currently use this new interface, so I don't know what the intended strategy for it is, and I'd be happy to be corrected, but it doesn't seem like it cleanly fits there either.  The daemon is online, so ideally it would be notified whenever a new block occurs (perhaps using the [Tendermint RPC](https://docs.tendermint.com/master/rpc/#/Websocket/subscribe)), and use that notification to trigger a state update.  Unlike `pcli`, it's not necessary for it to have any kind of checkpointing while syncing, because each step is small (1 block, except in rare cases such as following a restart), so it doesn't need to make use of the checkpointing hooks at all.  And, since it already has to have a way to detect state updates (e.g., using RPC) and to communicate with the Penumbra network (e.g., to send transactions), it's not clear that encapsulating just the compact block requests inside the `penumbra_wallet` APIs reduces the conceptual complexity of the (future) code in the `galileo` crate.

Rather than trying to create a combined, generic synchronization implementation in the `penumbra_wallet` crate, and change its model from unopinionated data structures that are *driven by* the caller to opinionated (e.g., requires Tokio) ones that drive network and disk activity, I suspect that writing different code for different contexts first, then examining their (dis)commonality, will give us better results.  Otherwise, I think there's a real risk that the implementation strategy ends up being driven by the a priori generic interface, rather than the interface being created to capture the common needs of different implementations.

(There's also a third important use case that we'll want to support in the future, which is where this code is compiled to WASM and run in a browser, but until we get there, we won't know more than a vague sense of what those requirements would be, so it's hard to write generic interfaces that would support it.)

PR #321, opened 2022-01-02
Add protos for staking messages, set up code infrastructure for staking implementation

hdevalence:
This chunk of work was meant to be simple and turned into an unexpectedly deep
rabbithole, but going down the rabbithole revealed some lessons for our
development process and helped build out a bunch of basic infrastructure.

The start was the creation of `stake.proto`, meant to define some stub proto
messages that we could use to do staking and delegation actions:

- defining (or updating) a validator;
- delegating to a validator;
- undelegating from a validator.

Delegation and undelegation messages are fairly straightforward: they specify
the validator, the unbonded amount, and the epoch.  (Those details determine
the rest of the parameters; the implementation work is in the checks that
they're consistent with the rest of the transaction, rather than in the data in
the message).

To avoid a separate process for creating a new validator and updating its
configuration, the idea is to use the same message with create-or-update
semantics.  This configuration data should be signed by the validator, and, as
with the other transaction structures, we split out the data-to-be-signed into
a separate message (`Validator`), and have the signed message
(`ValidatorDescription`) consist of a message+signature pair.  A transaction
can include a `ValidatorDescription` message to (re)define a validator.

Because this message includes a signature from the validator it configures,
it's self-authenticating.  However, a third party could create a transaction
including an authentic but stale validator description to roll back a
validator's configuration to some previous state.  To prevent this, the
`Validator` message includes a sequence number, so that the chain can require
that redefinitions have increasing sequence numbers, forcing each update to
pass through a signing oracle.

To specify the contents of the `Validator` description itself, we look at
the contents of the Cosmos SDK `create-validator` transaction as a model [0].
Some fields are just basic metadata, like the name, website, and description,
and others we ignore:

- initial commission rate, since we use funding streams;

- maximum commission and commission max change rate, which add implementation
  complexity to prevent a scenario with low impact (a validator advertises 0%
commission, then updates it to 100% in the hope the delegators won't notice)
that could be addressed by other means (social pressure, easy redelegation);

- minimum self-delegation, which doesn't fit cleanly with our delegation token
  model (and has unclear impact in a system with native delegation tokens
anyways).

However, looking at the address and pubkey fields reveals a problem with our
existing data modeling: Cosmos uses separate keys for the validator's identity
(determining the address, etc), and Tendermint consensus.  This is important
for various reasons, but most importantly, it allows the validator's identity
key to be stored offline, unlike the consensus key, which must be used online
to sign blocks.

The new `Validator` message distinguishes between the validator's
identity key (a decaf377-rdsa spend authorization key) and the validator's
consensus key (a Tendermint Ed25519 pubkey).  The validator's identity key
authenticates the consensus key by signing the `Validator` message.

Our existing code didn't notice this distinction, and changing the code to
account for it required propagating a bunch of other changes to the state
handling, effectively rewriting it.  While rewriting code isn't the end of the
world (especially when going from a first-pass of mapping out functionality to
a second-pass, coherent organization), in this instance the process was less
than ideal, and it would be good to figure out how we could try to improve it
in the future.

One observation is that it's not that the existing code was bad, but that we
worked "upwards" from the Tendermint API towards some idea of higher-level
application requirements, rather than specifying the eventual requirements in
more detail first, and then working from those requirements to the
implementation.  Writing down the requirements (in the form of proto
definitions for the messages the system uses to communicate) revealed a bunch
of details that we didn't previously think of.  The identity/consensus
distinction is a big one, but there are other, more trivial ones, like the fact
that we should record a validator's name and website, or that we'll need to
record both the exchange rates and reward rates.

This observation points to a potentially improved process: for each chunk of
functionality,

1. Write protos that define the data structures it will use;
2. Write domain types corresponding to those protos, and write the data
validation code as part of the conversions bundled with our `Protobuf` trait;
3. Implement the functionality using the domain types.

This way, in step 1, we can see how the data structures will fit together in
code review and map out the functionality without having to worry about the
details of the implementation, and the different components of step 3 can be
more easily parallelized, because people can work on different chunks of
implementation work using already-defined data structures.

Finally, another side benefit of centering the proto structures is that it
provides a clear definition of the data format, a centralized pathway for
validation (the conversion between the proto types and the domain types), and a
single location to define serializations in other formats (e.g., JSON).  Rather
than using Serde to derive serialization implementations ad-hoc, making every
structure definition potentially load-bearing and littering the code with
custom format-shaping annotations, we can use Serde on the proto structures.
This commit sets up the infrastructure to do that, based on the approach used
in `tendermint-rs`:

- The `penumbra-proto` `build.rs` has machinery to inject Serde derives, or
  format-shaping annotations ("serialize as hex string", "serialize as base64
string", etc.) into the Rust types generated from the proto files;

- The domain type for some proto type has a `TryFrom` impl that performs all of
  the validation logic;

- The domain type `Foo` derives `Serialize, Deserialize` with
  `#[serde(try_from = "pb::Foo", into = "pb::Foo")]`, so that the domain type's
Serde implementations are determined by the proto type, and there's a common
data validation pathway for both Protobuf and Serde serialization.

[0]: https://hub.cosmos.network/main/validators/validator-faq.html#how-to-become-a-validator

### DIFF
--- a/pd/src/app.rs
+++ b/pd/src/app.rs
@@ -386,7 +386,7 @@ impl App {
                 };
 
                 let next_rates = current_rates
-                    .iter()
+                    .into_iter()
                     .map(|current_rate| {
                         // TODO (hdevalence) use funding streams here, this ignores funding streams
                         let validator_reward_rate = base_reward_rate; // (minus fs term)
--- a/pd/src/app.rs
+++ b/pd/src/app.rs
@@ -510,7 +510,7 @@ impl Service<Request> for App {
                     // nullifiers, until we can use ABCI++ and control block
                     // proposals, at which point check_tx can run concurrently.
                     let rsp = self.check_tx(check_tx);
-                    let rsp = self.sequencer.execute(rsp);
+                    let rsp = self.sequencer.execute(rsp.instrument(Span::current()));
                     return async move {
                         let rsp = rsp.await;
                         tracing::info!(?rsp);
@@ -530,7 +530,7 @@ impl Service<Request> for App {
                 Request::DeliverTx(deliver_tx) => {
                     // Process DeliverTx messages sequentially.
                     let rsp = self.deliver_tx(deliver_tx.tx);
-                    let rsp = self.sequencer.execute(rsp);
+                    let rsp = self.sequencer.execute(rsp.instrument(Span::current()));
                     return async move {
                         let rsp = rsp.await;
                         tracing::info!(?rsp);
@@ -550,17 +550,14 @@ impl Service<Request> for App {
                     let rsp = self.end_block(end);
                     return self
                         .sequencer
-                        .execute(rsp)
-                        // TODO: (@hdevalence): why doesn't this span attach to the inner future?
-                        .instrument(Span::current())
+                        .execute(rsp.instrument(Span::current()))
                         .boxed();
                 }
                 Request::Commit => {
                     let rsp = self.commit();
                     return self
                         .sequencer
-                        .execute(rsp)
-                        .instrument(Span::current())
+                        .execute(rsp.instrument(Span::current()))
                         .boxed();
                 }
 
@@ -569,8 +566,7 @@ impl Service<Request> for App {
                     let rsp = self.init_genesis(init_chain);
                     return self
                         .sequencer
-                        .execute(rsp)
-                        .instrument(Span::current())
+                        .execute(rsp.instrument(Span::current()))
                         .boxed();
                 }
 
--- a/pd/src/app.rs
+++ b/pd/src/app.rs
@@ -1,5 +1,5 @@
 use std::{
-    collections::{BTreeMap, BTreeSet, VecDeque},
+    collections::{BTreeSet, VecDeque},
     future::Future,
     pin::Pin,
     sync::{Arc, Mutex},
@@ -9,12 +9,12 @@ use std::{
 use anyhow::anyhow;
 use bytes::Bytes;
 use futures::future::FutureExt;
-use metrics::increment_counter;
 use penumbra_crypto::{
     asset,
     merkle::{self, NoteCommitmentTree, TreeExt},
     note, Nullifier,
 };
+use penumbra_stake::{BaseRateData, RateData};
 use penumbra_transaction::Transaction;
 use tendermint::abci::{
     request::{self, BeginBlock, CheckTxKind, EndBlock},
@@ -131,25 +131,22 @@ impl App {
         // Now add the transaction and its note fragments to the pending state changes.
         genesis_block.add_transaction(verified_transaction);
 
-        // load the validators from the genesis app state
+        // Extract the Tendermint validators from the genesis app state
         //
         // NOTE: we ignore the validators passed to InitChain.validators, and instead expect them
         // to be provided inside the initial app genesis state (`GenesisAppState`). Returning those
         // validators in InitChain::Response tells Tendermint that they are the initial validator
         // set. See https://docs.tendermint.com/master/spec/abci/abci.html#initchain
-        let genesis_validators: BTreeMap<_, _> = app_state
+        let validators = app_state
             .validators
             .iter()
-            .cloned()
-            .map(|v| (v.tm_pubkey().clone(), v))
-            .collect();
-        let mut tm_validators = Vec::new();
-        for (pubkey, val) in genesis_validators.iter() {
-            tm_validators.push(tendermint::abci::types::ValidatorUpdate {
-                pub_key: pubkey.clone(),
-                power: val.voting_power.clone(),
+            .map(|genesis::ValidatorPower { validator, power }| {
+                tendermint::abci::types::ValidatorUpdate {
+                    pub_key: validator.consensus_key,
+                    power: *power,
+                }
             })
-        }
+            .collect();
 
         self.epoch_duration = app_state.epoch_duration;
 
@@ -157,21 +154,22 @@ impl App {
         self.pending_block = Some(Arc::new(Mutex::new(genesis_block)));
         let commit = self.commit();
         let state = self.state.clone();
-        let gc = app_state.clone();
         async move {
-            commit.await?;
-
-            // Save the genesis config to the blobs table for future reference.
+            // Write genesis data before committing the genesis block.
             state
-                .set_genesis_configuration(&gc)
+                .commit_genesis(&app_state)
                 .await
                 .expect("able to save genesis config to blobs table");
 
-            state.set_initial_validators(&genesis_validators).await?;
+            // Now process the genesis block (note: this is a bit icky, since
+            // there's some processing that happens prior to creating the commit
+            // future, but its ordering doesn't matter at the moment)
+            commit.await?;
+
             let app_hash = state.app_hash().await.unwrap();
             Ok(Response::InitChain(response::InitChain {
                 consensus_params: Some(init_chain.consensus_params),
-                validators: tm_validators,
+                validators,
                 app_hash: app_hash.into(),
             }))
         }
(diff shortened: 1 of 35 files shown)


pd/src/app.rs
```diff
@@ -335,29 +333,106 @@ impl App {
                 .unwrap()
                 .add_transaction(verified_transaction);
 
-            increment_counter!("node_transactions_total");
+            metrics::increment_counter!("node_transactions_total");
```
hdevalence, line 336: Previously this was just `increment_counter!`; for tracing and metrics calls I think it's useful not to import the specific functions but just to use the crate name, so that it's immediately clear that it's some instrumentation rather than functional behavior.

pd/src/app.rs
```diff
@@ -335,29 +333,106 @@ impl App {
...
             Ok(())
         }
     }
 
-    fn end_block(&mut self, end: EndBlock) -> response::EndBlock {
-        let epoch = self
+    fn end_block(&mut self, end: EndBlock) -> impl Future<Output = Result<Response, BoxError>> {
```
hdevalence, line 341: The application logic for processing rate changes moved to the `EndBlock` handler; previously, it was included in the `State` code, but it's better for the state code to just handle writing state updates, rather than figuring out what the state updates should be.

Because processing rate changes requires querying the existing state, this method becomes async.

pd/src/app.rs
```diff
@@ -335,29 +333,106 @@ impl App {
...
+                let current_rates = state.rate_data(epoch.index).await?;
+                tracing::debug!(?current_base_rate);
+                for current_rate in &current_rates {
+                    tracing::debug!(?current_rate);
+                }
+
+                // here's where we should compute updated rates (moved from state code)
+                // TODO @ava insert calls here
+
+                // move below into calls to methods in stake crate?
```
hdevalence, line 375: @avahowell I moved your TODO along with moving the code out of the `State` impl; I filled in some really basic stub code below, but I guess the computations for computing new rates should move into somewhere in the `stake` crate (?)

pd/src/app.rs
```diff
@@ -335,29 +333,106 @@ impl App {
...
+                    base_exchange_rate,
+                    base_reward_rate,
+                    epoch_index: epoch.index + 1,
+                };
+
+                let next_rates = current_rates
+                    .iter()
+                    .map(|current_rate| {
+                        // TODO (hdevalence) use funding streams here, this ignores funding streams
+                        let validator_reward_rate = base_reward_rate; // (minus fs term)
```
hdevalence, line 392: This stub code ignores funding streams for no reason other than that it would involve doing another database query. 

It's a little annoying to have to constantly query the database for this stuff (why is this function async at all?), so we might want to consider keeping some data (like the consensus set) in memory in the `State` struct. I don't think this would impose a ton of cache coherence complexity, because we do all of our writes in one place.

pd/src/app.rs
```diff
@@ -335,29 +333,106 @@ impl App {
...
+                        let validator_exchange_rate = (current_rate.validator_exchange_rate
+                            * (validator_reward_rate + 1_0000_0000))
+                            / 1_0000_0000;
+
+                        // this is supposed to be multiplied by the number of delegation tokens,
+                        // how do we track that?
+                        let _voting_power_adjustment =
+                            (validator_exchange_rate * 1_0000_0000) / base_exchange_rate;
```
hdevalence, line 400: We can compute the voting power adjustment factor, but we can't compute the voting power itself without knowing the number of delegation tokens for that validator.  Rather than doing something special for tracking delegation tokens, it seems useful to track the supply of all tokens.

For now, this leaves the voting power unchanged (in the application level, it's also unchanged in the consensus level).

pd/src/app.rs
```diff
@@ -335,29 +333,106 @@ impl App {
...
+                    epoch_index: epoch.index + 1,
+                    base_reward_rate,
+                    base_exchange_rate,
+                });
+            }
+
+            // TODO: if necessary, set the EndBlock response to add validators
+            // at the epoch boundary
+            if end.height.unsigned_abs() == epoch.start_height().value() {
+                // TODO: does this go in the first block of the epoch or the last block of the epoch?
```
hdevalence, line 429: Should we be updating the consensus set in the first block of each epoch, or the last block of each epoch? I think we might want to be doing it in the last block of epoch e, so that the first block of epoch e+1 will be voted on by the new validators.

pd/src/app.rs
```diff
@@ -471,7 +546,15 @@ impl Service<Request> for App {
...
                 }
-                Request::EndBlock(end) => Response::EndBlock(self.end_block(end)),
+                Request::EndBlock(end) => {
+                    let rsp = self.end_block(end);
+                    return self
+                        .sequencer
+                        .execute(rsp)
+                        // TODO: (@hdevalence): why doesn't this span attach to the inner future?
+                        .instrument(Span::current())
+                        .boxed();
```
hdevalence, line 556: The sequencer doesn't propagate spans correctly.

I think the issue is that the sequencer future is instrumenting the future that waits on the joinhandle, but not instrumenting the future that gets spawned as a new task.  I'm not sure what the best way to propagate the span is.  Here are some options:

- we could use `Span::current()` in `Sequencer::execute`, which grabs the ambient span; in this case, that's the right one, but it might not be elsewhere;
- we could have the caller instrument the future passed to `execute`, like `.execute(rsp.instrument(Span::current()))`; if we wanted to instrument the outer (returned) future, we'd need two `instrument` calls, but since it's just waiting on a `JoinHandle` maybe we don't care (?);
- we could have the caller pass an explicit `Span` to `execute`, so they couldn't forget to instrument the future (but if they didn't want to instrument it, they'd have to make something up).

Doing #2 by moving the `.instrument` call to `rsp` rather than the result of `execute` everywhere in the `Service::call` impl seems the cleanest.

pd/src/genesis.rs
```diff
@@ -47,15 +47,21 @@ impl Allocation {
     }
 }
 
+#[derive(Deserialize, Serialize, Debug, Clone)]
+pub struct ValidatorPower {
+    pub validator: Validator,
+    pub power: tendermint::vote::Power,
+}
+
```
hdevalence, line 55: Now that the validator power isn't part of the `Validator` struct (since it's not an inherent attribute of the validator, but a contextual one), I needed a way to specify the validator power in the genesis file, but there's otherwise no principled reason for this to exist.

It might be more useful if the `stake` crate had a `ValidatorState` type that could record the validator's voting power, current state (is it in the consensus set? was it slashed? etc), which could then be used by the genesis data.

pd/src/genesis.rs
```diff
@@ -47,15 +47,21 @@ impl Allocation {
...
 pub struct AppState {
-    /// The initial token allocations.
-    pub allocations: Vec<Allocation>,
     /// The number of blocks in each epoch.
     pub epoch_duration: u64,
     /// The initial validator set.
-    pub validators: Vec<Validator>,
+    pub validators: Vec<ValidatorPower>,
+    /// The initial token allocations.
+    pub allocations: Vec<Allocation>,
```
hdevalence, line 64: Moving `allocations` to the end of the structure places it last in the JSON output, which is a bit easier to work with (the allocations might be much longer than other genesis data).

Maybe we should define the genesis structures also as proto files?

pd/src/pending_block.rs
```diff
@@ -24,6 +24,10 @@ pub struct PendingBlock {
     pub epoch: Option<Epoch>,
     /// Indicates the duration in blocks of each epoch.
     pub epoch_duration: u64,
+    /// If this is the last block of an epoch, base rates for the next epoch go here.
+    pub next_base_rate: Option<BaseRateData>,
+    /// If this is the last block of an epoch, validator rates for the next epoch go here.
+    pub next_rates: Option<Vec<RateData>>,
```
hdevalence, line 30: Adding this to the `PendingBlock` means that the `State` doesn't need to compute it, xref https://github.com/penumbra-zone/penumbra/pull/321/files#r777265332

pd/src/sequential.rs
```diff
@@ -35,6 +35,8 @@ impl Sequencer {
         self.waiting = true;
         self.completion.set(rx);
 
+        // TODO: (@hdevalence) how do we propagate spans here?
```
hdevalence, line 38: xref https://github.com/penumbra-zone/penumbra/pull/321/files#r777267153

pd/src/state.rs
```diff
@@ -27,6 +24,12 @@ pub struct State {
     pool: Pool<Postgres>,
 }
 
+// NOTE: because the state has a Postgres connection pool and supports shared
+// access, all methods (including write methods!) take &self, not &mut self.
+//
+// Writes to the database should *only* happen in `commit_genesis` (called once
+// on init) and `commit_block`.
```
hdevalence, line 31: The previous `State` code had multiple methods for writing data (as an artefact of those parts being written in parallel); along the way to these changes, those methods are now combined into two `commit` methods.  Going forward, we should add new DB writes only in those places, which correspond to the ABCI state changes.

pd/src/state.rs
```diff
@@ -39,6 +42,83 @@ impl State {
...
+            query!(
+                "INSERT INTO validators (
+                    identity_key, 
+                    consensus_key, 
+                    sequence_number, 
+                    validator_data
+                ) VALUES ($1, $2, $3, $4)",
+                validator.identity_key.to_bytes().to_vec(),
+                validator.consensus_key.to_bytes(),
+                validator.sequence_number as i64,
```
hdevalence, line 76: Here and elsewhere, I changed the state code to do the `i64`/`u64` conversion as close to the DB layer as possible.  It's annoying that Postgres doesn't support unsigned integers, but it seems better to avoid propagating that annoyance through all of our data structures.

This conversion is lossless (it's just changing the interpretation of the sign bit), so we'll always get the data back; the only risk is if we try to do arithmetic inside of Postgres on values in excess of 2^63. The only place that it seems likely we'd do arithmetic inside of Postgres is something like sorting, but values we'd be likely to sort on (like block heights) probably won't be brushing against the top of the value range anyways.

proto/build.rs
```diff
@@ -40,3 +50,40 @@ fn main() -> Result<()> {
...
+
+// Requires SERDE_AS on the container
+// :(
+// error: expected non-macro attribute, found attribute macro `::serde_with::serde_as`
+// use tendermint-rs approach instead?
+// https://github.com/penumbra-zone/tendermint-rs/blob/master/proto/src/serializers/bytes.rs#L4-L30
+// static AS_HEX: &str = r#"#[::serde_with::serde_as(as = "::serde_with::hex::Hex")]"#;
+
+static AS_HEX: &str = r#"#[serde(with = "crate::serializers::hexstr")]"#;
+static AS_BASE64: &str = r#"#[serde(with = "crate::serializers::base64str")]"#;
```
hdevalence, line 69: Some of this could be deleted; initially, I tried to use `serde_with`, but ran into problems adding the attribute macros to generated code output by prost. It seems kind of brittle anyways, and the `tendermint-rs` approach works pretty well.

proto/build.rs
```diff
@@ -40,3 +50,40 @@ fn main() -> Result<()> {
...
+
+static FIELD_ATTRIBUTES: &[(&str, &str)] = &[
+    // TODO: use bech32 here?
+    (".penumbra.stake.Validator.identity_key", AS_HEX),
+    // Using base64 for the validator's consensus key means that
+    // the format is the same as the Tendermint json config files.
+    (".penumbra.stake.Validator.consensus_key", AS_BASE64),
+    (".penumbra.stake.ValidatorDefinition.auth_sig", AS_HEX),
+    (".penumbra.stake.RateData.identity_key", AS_HEX),
+];
```
hdevalence, line 89: The upshot of this infrastructure is that all of our serialization formats can be determined entirely by the `.proto` files in the crate, together with these "shaping" annotations.  This way, we can confidently rearrange the internals of our domain types, without worrying that by doing so we've accidentally broken the JSON encoding, or something.

proto/proto/sighash.proto
```diff
@@ -26,5 +27,8 @@ message SigHashAction {
   oneof action {
     transaction.SpendBody spend = 1;
     transaction.Output output = 2;
+    stake.Delegate delegate = 3;
+    stake.Undelegate undelegate = 4;
+    stake.ValidatorDefinition validator_definition = 16;
```
hdevalence, line 32: Field numbers are encoded as varints, so the Proto style guide recommends that uncommon variants should be encoded using numbers that don't fit in a single byte.

proto/src/protobuf.rs
```diff
@@ -1,8 +1,20 @@
...
+    fn encode_to_vec(&self) -> Vec<u8> {
+        P::from(self.clone()).encode_to_vec()
+    }
+
+    /// Decode this domain type from a byte buffer, via proto type `P`.
+    fn decode<B: bytes::Buf>(buf: B) -> Result<Self, anyhow::Error> {
+        <P as prost::Message>::decode(buf)?
+            .try_into()
+            .map_err(Into::into)
+    }
```
hdevalence, line 19: This means that we don't have to manually implement a ton of `Vec<u8>` conversions for each domain type; I think there's some like that on `Transaction` that could now be cleaned up.

stake/src/epoch.rs
```diff
@@ -10,28 +9,10 @@ pub struct Epoch {
...
-            block_height.unsigned_abs(),
-            epoch_duration,
-        ))
-    }
-
-    /// from_blockheight_unsigned instantiates a new `Epoch` from a given
-    /// unsigned block height. Due to the implementation in tendermint using
-    /// signed representation for block height, we provide this
-    /// as well as a signed implemention (`from_blockheight`)
-    pub fn from_blockheight_unsigned(block_height: u64, epoch_duration: u64) -> Self {
```
hdevalence, line 32: Doing the `i64` conversion at the DB boundary means we don't propagate the complexity into the internal APIs, so this can be simplified to a single method.

stake/src/rate.rs
```diff
@@ -0,0 +1,81 @@
+use penumbra_crypto::rdsa::{SpendAuth, VerificationKey};
+use penumbra_proto::{stake as pb, Protobuf};
```
hdevalence, line 2: Importing the corresponding proto module `as pb` makes it easy to consistently name both the domain types `Foo` and the proto types `pb::Foo`; elsewhere in the code we're a bit less consistent.

proto/src/serializers/bech32str.rs
```diff
@@ -0,0 +1,96 @@
...
+    if hrp != expected_hrp {
+        return Err(anyhow::anyhow!(
+            "wrong bech32 human readable part {}, expected {}",
+            hrp,
+            expected_hrp
+        ));
+    }
+
+    Ok(Vec::from_base32(&data).expect("bech32 decoding produces valid base32"))
+}
```
hdevalence, line 37: The motivation for having separate, public `decode` and `encode` methods is to avoid having to do this dance every time someone wants to do manual Bech32 serialization.

The reason they'd want to do that is in order to implement `Display`/`FromStr`, the type needs to be able to pass to and from Bech32 directly, and I'm not sure that there's a way to reuse the Serde implementation to do so. (unless?)

stake/src/identity_key.rs
```diff
@@ -0,0 +1,55 @@
...
+
+impl std::fmt::Display for IdentityKey {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_str(&bech32str::encode(
+            &self.0.to_bytes(),
+            crate::VALIDATOR_IDENTITY_BECH32_PREFIX,
+            bech32str::Bech32m,
+        ))
+    }
+}
```
hdevalence, line 32: not 100% happy with this, it feels somewhat cumbersome & duplicative -- it'd be neat to come up with a way to reuse the Serialize/Deserialize implementations directly...

pd/src/sequential.rs
hdevalence: resolved in https://github.com/penumbra-zone/penumbra/pull/321/commits/8ccb8e13162b939e4d7d2fa524574b39cd977599

pd/src/app.rs
hdevalence: resolved in https://github.com/penumbra-zone/penumbra/pull/321/commits/8ccb8e13162b939e4d7d2fa524574b39cd977599

pd/src/app.rs
plaidfinch: That sounds like the least-confusing thing to me.

pd/src/genesis.rs
plaidfinch: Along the lines of your commentary above, it seems good to move away from ad-hoc serde derivations, so that makes sense to me.

pd/src/app.rs
avahowell: yes, I think it makes sense to have them in the stake crate. in a wip branch where i've implemented the rate calculations they're in `stake/src/rates.rs`

plaidfinch (approved):

redshiftzero (approved):

aubrika (approved):
LGTM!

PR #322, opened 2022-01-03
pcli: streamline state handling

hdevalence:
Some of the offline commands still need a state file, so it's easier to have
common code if we ensure the state is always available, rather than forcing
subcommands to load it on their own.


plaidfinch:
Makes sense to me!

plaidfinch (approved):

Issue #275, opened 2021-12-13
Add methods to the thin wallet API that allow clients to fetch all of these rates.


avahowell:

avahowell:
We now have this with  `rpc ValidatorRate(ValidatorRateRequest) returns (stake.RateData);`

Issue #49, opened 2021-09-08
Add queries for staking-related rates


hdevalence:
Extend the protocol created in #22 to allow querying the application state for:

- [ ] historical base reward rates
- [ ] historical commission percentages for each validator
- [ ] historical exchange rates for shielded staking tokens

PR #323, opened 2022-01-04
Stub staking support for pcli

hdevalence:
WIP code following #321 that tries to map out required internal functionality for staking by sketching an external interface via `pcli`.

### DIFF
--- a/pcli/src/command/stake.rs
+++ b/pcli/src/command/stake.rs
@@ -1,5 +1,7 @@
-use anyhow::Result;
-use penumbra_stake::IdentityKey;
+use anyhow::{anyhow, Result};
+use penumbra_crypto::Value;
+use penumbra_proto::thin_wallet::{thin_wallet_client::ThinWalletClient, ValidatorRateRequest};
+use penumbra_stake::{Epoch, IdentityKey, RateData, STAKING_TOKEN_ASSET_ID};
 use penumbra_wallet::ClientState;
 use structopt::StructOpt;
 
@@ -67,7 +69,7 @@ impl StakeCmd {
         true
     }
 
-    pub async fn exec(&self, _opt: &Opt, state: &ClientState) -> Result<()> {
+    pub async fn exec(&self, opt: &Opt, state: &ClientState) -> Result<()> {
         match self {
             StakeCmd::Delegate {
                 to,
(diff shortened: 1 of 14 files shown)


pcli/src/command/stake.rs
```diff
@@ -0,0 +1,145 @@
...
+        amount: String,
+        /// The transaction fee (paid in upenumbra).
+        #[structopt(long, default_value = "0")]
+        fee: u64,
+        /// Optional. Only spend funds originally received by the given address index.
+        #[structopt(long)]
+        source: Option<u64>,
+    },
+    /// Display this wallet's delegations and their value.
+    Show,
```
hdevalence, line 58: @avahowell and I thought it made sense to have this be separate from the `balance` command; the `balance` command will show the amounts of all of the delegation tokens, but doesn't know that they're anything other than normal tokens.

It'd be useful to have staking-specific information, like showing the equivalent amount in unbonded stake, bonded stake, showing more details on validators (like the human-readable name), etc., and putting all that into the normal `balance` command seems a bit messy.

pcli/src/command/stake.rs
```diff
@@ -0,0 +1,145 @@
...
+                };
+
+                let to = to.parse::<IdentityKey>()?;
+
+                // FIXME! need some kind of structure for recording chain
+                // parameters - connected with having protos for genesis data
+                // (though not all genesis data is chain parameters)
+                let epoch_duration = 10;
+                let current_epoch =
+                    Epoch::from_height(state.last_block_height().unwrap() as u64, epoch_duration);
```
hdevalence, line 95: We don't currently have any kind of structure of recording chain parameters that the client might want to know about. This has already been a problem when trying to use the chain ID to detect wrong-chain errors, since the client state doesn't have anywhere to store the chain ID.

It's connected to the idea of having proto-specified genesis data, since we'd probably want the genesis data to be a superset of the chain parameters (e.g., the chain params include the epoch duration, but not the genesis allocations).

proto/proto/stake.proto
```diff
@@ -59,6 +59,13 @@ message ValidatorDefinition {
   bytes auth_sig = 2;
 }
 
+// Describes the current state of a validator on-chain
+message ValidatorStatus {
+  Validator validator = 1;
+  uint64 voting_power = 2;
+  // ... ?
+}
+
```
hdevalence, line 68: This is a stub, but we should figure out what information we need to record about the state of a validator earlier rather than later.  The `voting_power` can move out of the `RateData` structure, since it's actually a different category of information (it's based on the delegations, not just the validator parameters).  We probably also want to record whether or not the validator is active, inactive, jailed, etc.

proto/proto/stake.proto
```diff
@@ -69,6 +76,12 @@ message Delegate {
   // The delegation amount, in units of unbonded stake.
   // TODO: use flow aggregation to hide this, replacing it with bytes amount_ciphertext;
   uint64 unbonded_amount = 3;
+  // The amount of delegation tokens produced by this action.
+  // 
+  // This is implied by the validator's exchange rate in the specified epoch
+  // (and should be checked in transaction validation!), but including it allows
+  // stateless verification that the transaction is internally consistent.
+  uint64 delegation_amount = 4;
```
hdevalence, line 84: To compute the action's contributions to the value balance, we need to compute the delegation amount, and this is only possible with knowledge of all of the validator exchange rates.  In principle, we could omit this data to save a few bytes, but doing so means that we wouldn't be able to statelessly verify the rest of the transaction data.  Instead, by adding it, we add a required consistency check to stateful verification, but allow stateless verification of the transaction data.

stake/src/lib.rs
```diff
@@ -20,3 +23,10 @@ pub use validator::{Validator, ValidatorDefinition};
 pub const VALIDATOR_CONSENSUS_BECH32_PREFIX: &str = "penumbravalconspub";
 
 pub use penumbra_proto::serializers::bech32str::validator_identity_key::BECH32_PREFIX as VALIDATOR_IDENTITY_BECH32_PREFIX;
+
+// TODO: go through the source tree and use these instead of hardcoding "upenumbra"
```
hdevalence, line 27: Noting that this should be broken out into another issue.

aubrika (approved):
LGTM!

pcli/src/command/stake.rs
hdevalence: This is now possible with #332

hdevalence:
I think it would make sense to go through this, create a list of unfinished chunks of work, and then merge it as a skeleton for filling in that work.

hdevalence:
Closes #275.

hdevalence:
Closes #49

PR #331, opened 2022-01-05
refactor rate calculations into stake crate

avahowell:
This PR:

- factors out the rate calculations from App.end_block and puts them in the stake crate
- adds the computation of the validator reward rate based on the commission rate calculated from the sum of their funding streams

### DIFF
--- a/stake/src/rate.rs
+++ b/stake/src/rate.rs
@@ -1,7 +1,14 @@
-use penumbra_proto::{stake as pb, Protobuf};
+use penumbra_proto::{
+    stake::{self as pb},
+    Protobuf,
+};
 use serde::{Deserialize, Serialize};
 
-use crate::IdentityKey;
+use crate::{Epoch, IdentityKey, FundingStream};
+
+/// FIXME: set this less arbitrarily, and allow this to be set per-epoch
+/// 3bps -> 11% return over 365 epochs, why not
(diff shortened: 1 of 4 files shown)


hdevalence (approved):
LGTM! It looks like the main missing piece left is computing the voting power, and doing that probably needs some machinery for tracking token balances.

stake/src/rate.rs
```diff
@@ -19,6 +26,51 @@ pub struct RateData {
...
+        //
+        // todo: consider specifying the voting power function as a pure function of current epoch
+        // state (delegation tokens, etc) instead of an adjustmenet function
+        let voting_power_adjustment =
+            (validator_exchange_rate * 1_0000_0000) / base_rate_data.base_exchange_rate;
+
+        RateData {
+            identity_key: self.identity_key.clone(),
+            epoch_index: self.epoch_index + 1,
+            voting_power: self.voting_power * voting_power_adjustment,
```
hdevalence, line 68: @avahowell ah, wait, this is not correct, the voting power adjustment should be multiplied by the supply of delegation tokens, not by the current voting power (sorry for missing this in review!).

The previous code left the voting power untouched since we didn't have token supply tracking (#335), which is also not the intended behavior, but I think that this line would cause the voting power to blow up (if we ever committed it back to tendermint).

Rather than fixing this in place, we might as well just fix it when doing the calculations properly in #337.

PR #332, opened 2022-01-05
326: Add ChainParams proto definition and penumbra-chain crate

### LINKED ISSUE #326: Create a `penumbra-chain` crate and `ChainParams` proto type
We ran into a problem in #323 where the client needs to know the current epoch index, and could in principle compute it from the height, but doesn't know the epoch duration. We've also had a problem where we would have liked to add the chain ID to the wallet protocol requests to detect chain mismatches, but didn't have a way to request or store it.

This points to a need to have some set of chain parameters that can be shared with clients.  To do this, we should:

- [x] create a new `params.proto` in the proto crate;
- [x] add a `ChainParams` message with (for now) `string chain_id` and `uint64 epoch_duration`;
- [x] add serialization annotations in the `build.rs`;
- [x] create a new `penumbra-chain` crate for common chain data structures that both `pcli` and `pd` can depend on;
- [x] add a `ChainParams` domain type to `penumbra-chain`.

zbuc:
Closes #326

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -21,10 +21,12 @@ COPY pd/build.rs ./pd
 # TODO If the protobuf definitions, crypto, or wallet implementations change,
 # there will be a complete rebuild. This is maybe possible to avoid.
 COPY proto ./proto
+COPY chain ./chain
 COPY crypto ./crypto
 COPY stake ./stake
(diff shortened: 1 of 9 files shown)


hdevalence (approved):
LGTM!

PR #334, opened 2022-01-05
Rename pending to submitted

### LINKED ISSUE #325: Change `pcli`'s "pending" to "submitted"
`pcli` tracks notes in transactions that have been submitted to the chain but not yet added to the chain state.  This is important because each transaction may spend more than the requested amount, returning the rest to the user as change, and those change notes can't be spent until after the first transaction has been accepted.

Currently, these are described (in the code and in the UI) as "pending" notes. However, in order to handle staking and unbonding, we need to handle cases where notes have been accepted by the chain, but cannot be spent until some point in the future (e.g., once the unbonding period is over).  It would be better to reserve the word "pending" for that state, and use a different word (I'd suggest "submitted") for `pcli`'s use case.

plaidfinch:
Fixes #325.

### DIFF
--- a/wallet/src/state.rs
+++ b/wallet/src/state.rs
@@ -41,9 +41,9 @@ pub struct ClientState {
     /// Notes that we have received.
     unspent_set: BTreeMap<note::Commitment, Note>,
     /// Notes that we have spent but which have not yet been confirmed on-chain.
-    pending_set: BTreeMap<note::Commitment, (SystemTime, Note)>,
+    submitted_spend_set: BTreeMap<note::Commitment, (SystemTime, Note)>,
     /// Notes that we anticipate receiving on-chain as change but which have not yet been confirmed.
-    pending_change_set: BTreeMap<note::Commitment, (SystemTime, Note)>,
+    submitted_change_set: BTreeMap<note::Commitment, (SystemTime, Note)>,
     /// Notes that we have spent.
     spent_set: BTreeMap<note::Commitment, Note>,
     /// Map of note commitment to full transaction data for transactions we have visibility into.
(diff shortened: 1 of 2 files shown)


wallet/src/state.rs
```diff
@@ -585,10 +592,10 @@ mod serde_helpers {
         note_commitment_tree: Vec<u8>,
         nullifier_map: Vec<(String, String)>,
         unspent_set: Vec<(String, String)>,
-        #[serde(default)]
-        pending_set: Vec<(String, SystemTime, String)>,
-        #[serde(default)]
-        pending_change_set: Vec<(String, SystemTime, String)>,
+        #[serde(default, alias = "pending_set")]
```
plaidfinch, line 595: This alias allows us to continue to parse wallets from before this change, without needing to run `pcli wallet reset`.

wallet/src/state.rs
```diff
@@ -585,10 +592,10 @@ mod serde_helpers {
...
+        submitted_spend_set: Vec<(String, SystemTime, String)>,
+        #[serde(default, alias = "pending_change_set")]
```
plaidfinch, line 597: This alias allows us to continue to parse wallets from before this change, without needing to run `pcli wallet reset`.

plaidfinch:
It's worth noting that the serde alias is only necessary _once_ per client; once the wallet has been read, when it is re-committed, it will be re-committed with the new field names.

We could remove these aliases at the next testnet release, because we will be asking people to do `pcli wallet reset` at that point anyhow.

hdevalence (approved):
LGTM!

PR #338, opened 2022-01-05
stake: avoid BASE_REWARD_RATE constant

hdevalence:
Since we want to eventually have variable reward rates, it seems better to
avoid having a reward rate constant, so that the rest of the API is built
around the idea that the base reward rate for the next epoch is an external
input.


avahowell:
LGTM

PR #342, opened 2022-01-07
Consistent sign handling for value commitments, and (un)delegate commitments

hdevalence:
Previously, we treated the value commitment in an output description as a
commitment to the (positive) value accruing to the output notes, rather than as
a commitment to the (negative) value contributed (i.e., withdrawn) from the
transaction's value balance.  This meant that when summing the value
commitments for each action, we needed to handle spends and outputs
differently, because they had inconsistent signs.

Instead, this commit changes the value commitment of the output description to
commit to the action's contribution to the transaction's value balance (a
negative number).  This means that the semantics of the action (whether it is
contributing or withdrawing) are encapsulated in the action logic, rather than
leaking out to the code that handles the actions.

### DIFF
--- a/transaction/src/transaction.rs
+++ b/transaction/src/transaction.rs
@@ -7,7 +7,7 @@ use penumbra_crypto::{
     asset,
     merkle::{self, NoteCommitmentTree, TreeExt},
     rdsa::{Binding, Signature, VerificationKey, VerificationKeyBytes},
-    Fr, Value,
+    value, Fr, Value,
 };
 use penumbra_proto::{
     transaction::{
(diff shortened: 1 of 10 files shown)


transaction/src/transaction.rs
```diff
@@ -106,16 +106,8 @@ impl Transaction {
...
-                Action::Output(inner) => {
-                    value_commitments -= inner.body.value_commitment.0;
-                }
-                Action::Spend(inner) => {
-                    value_commitments += inner.body.value_commitment.0;
-                }
-                _ => todo!("delegation binding signature contributions??"),
-            }
+        for action in &self.transaction_body.actions {
+            value_commitments += action.value_commitment().0;
```
redshiftzero, line 110: nice!

redshiftzero (approved):
LGTM

PR #343, opened 2022-01-08
update staking arithmetic spec for voting power calculation

avahowell:
this PR updates the arithmetic part of the staking spec to compute the voting power instead of just the voting power adjustment function.

### DIFF
--- a/spec/stake/arithmetic.md
+++ b/spec/stake/arithmetic.md
@@ -38,8 +38,8 @@ bits.
 * $c_{v,e}$: the sum of the validator's commission rates, in basis points.
 * $\mathtt r_{v,e}$: the fixed-point representation of the validator-specific reward rate for validator $v$ at epoch $e$.
 * $\mathtt {psi}_v(e)$: the fixed-point representation of the validator-specific exchange rate for validator $v$ at epoch $e$.
-* $\mathtt {theta}_v(e)$: the fixed-point representation of the validator's voting power adjustment function for validator $v$ at epoch $e$.
-
+* $\mathtt y_{v}$: the sum of the tokens in the validator's delegation pool.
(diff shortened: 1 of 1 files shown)


spec/stake/arithmetic.md
```diff
@@ -74,8 +74,8 @@ of the base reward rate:
...
 $$\mathtt {psi}_v(e) = \left\lfloor \frac {\mathtt {psi}_v(e-1) \cdot (10^8 + \mathtt r_{v,e})}{10^8} \right\rfloor$$
 
-## Validator Voting Power Adjustment
+## Validator Voting Power
 
-Finally, to compute the validator's voting power adjustment function, compute:
+Finally, to compute the validator's voting power, take:
 
-$$\mathtt {theta}_v(e) = \left\lfloor \frac{\mathtt {psi}_v(e) 10^8}{\mathtt {psi}(e)} \right\rfloor$$
+$$\mathtt {theta}_v(e) = \mathtt y_{v} \cdot \left\lfloor \frac{\mathtt {psi}_v(e) 10^8}{\mathtt {psi}(e)} \right\rfloor$$
```
hdevalence, line 81: I don't think this is quite right, since we don't want to redefine theta as the validator's voting power (rather than as the voting power adjustment function).  I think it would be helpful to reiterate the abstract quantity that's being computed, and then follow up with the expression we actually use to do the computation.

Also, since theta is a ratio, it won't generally be an integral value, so we need to do the rounding on the whole expression, rather than just on the ratio, and my suspicion is that there won't be factors of 10^8 floating around, so we'd get something like
```
\lfloor y_v \cdot ( psi_v(e) / psi(e) ) \rfloor
```

avahowell:
updated the arith spec to use iota instead of theta, and changed the calculation so that it is correct and reflects what is currently being done as of the latest changes to the staking code

hdevalence (approved):
LGTM!

Issue #335, opened 2022-01-05
Track the total supply of every asset recorded on-chain


hdevalence:
We need to be able to track the total supply of every asset recorded on-chain in order to maintain supply transparency, and in particular, we need to track the number of each validator's delegation tokens, so that we can compute the voting power for that validator.

We already need to maintain an on-chain registry of denominations, so that we can look up the denomination corresponding to an asset ID. I think it makes sense to extend/unify that functionality with the supply tracking.  This would mean that instead of just recording the denomination string corresponding to each asset ID, we would also track the current supply of that asset.

Currently, the thin wallet API contains
```proto
rpc AssetLookup(AssetLookupRequest) returns (Asset);
```
where
```
// Requests an asset denom given an asset ID
message AssetLookupRequest {
  // The asset ID
  bytes asset_id = 1;
}

message Asset {
  bytes asset_id = 1;
  string asset_denom = 2;
}
```
are defined in `thin_wallet.proto`.

Instead, we should return both the denomination and information on the token supply.  With a view to the direction suggested in the discussion around #321, I'd suggest the following as a path to get there that lays out some groundwork for future cleanup:

#### New data modeling

- [x] Add a new `crypto.proto` protobuf that should correspond to the `penumbra-crypto` crate;
- [x] Add the following proto types to `crypto.proto`, and change the existing `penumbra-crypto` types to be domain types for those proto types:
  - [x] `message AssetId { bytes asset_id = 1; }`
  - [x] `message Denom { string denom = 1; }`
- [x] Add an `AssetInfo` message to the new `params.proto` for the `penumbra-chain` crate added in #332 that has something like:
```proto
// open to suggestions on this, seems useful to include some info
// about when the total supply value was current
message AssetInfo {
  crypto.AssetId asset_id = 1;
  crypto.Denom denom = 2;
  uint64 total_supply = 3;
  uint64 as_of_block_height = 4;
}
```
- [x] Add the corresponding domain type to `penumbra-chain`;
- [x] Change the thin wallet RPC method to `rpc AssetLookup(crypto.AssetId) returns (params.AssetInfo)`;

#### Using the data modeling

To implement that RPC method, we need to be able to construct `AssetInfo` responses, which requires knowing the total supply of each token. As written above, while the *response* includes the block height at which the data was current (so that the client doesn't need to do anything special to be able to keep track of it, and can just ignore/discard it if they don't care), the request doesn't include a block height, so it's always asking for the current token supply.  

[...]

PR #346, opened 2022-01-11
track the total token supply, and update RPCs to use supply tracking

avahowell:
this PR implements the changes described in #335

### DIFF
--- a/crypto/src/asset/denom.rs
+++ b/crypto/src/asset/denom.rs
@@ -35,9 +35,9 @@ impl TryFrom<pb::Denom> for Denom {
     type Error = anyhow::Error;
 
     fn try_from(value: pb::Denom) -> Result<Self, Self::Error> {
-        Ok(Denom {
-            inner: Arc::new(Inner::new(value.denom, Vec::new())),
-        })
+        asset::REGISTRY
+            .parse_denom(&value.denom)
+            .ok_or_else(|| anyhow::anyhow!("invalid denomination {}", value.denom))
     }
 }
 
--- a/crypto/src/asset/denom.rs
+++ b/crypto/src/asset/denom.rs
@@ -5,17 +5,40 @@ use std::{
     sync::Arc,
 };
 
+use penumbra_proto::{crypto as pb, Protobuf};
+
 use ark_ff::fields::PrimeField;
 
 use crate::{asset, Fq, Value};
-
+use serde::{Deserialize, Serialize};
 /// An asset denomination.
 ///
 /// Each denomination has a unique [`asset::Id`] and base unit, and may also
 /// have other display units.
-#[derive(Clone)]
+#[derive(Serialize, Deserialize, Clone)]
+#[serde(try_from = "pb::Denom", into = "pb::Denom")]
 pub struct Denom {
     pub(super) inner: Arc<Inner>,
+}
+
+impl Protobuf<pb::Denom> for Denom {}
+
+impl From<Denom> for pb::Denom {
+    fn from(dn: Denom) -> Self {
+        pb::Denom {
+            denom: dn.inner.base_denom.clone(),
+        }
+    }
+}
+
+impl TryFrom<pb::Denom> for Denom {
+    type Error = anyhow::Error;
+
+    fn try_from(value: pb::Denom) -> Result<Self, Self::Error> {
+        Ok(Denom {
+            inner: Arc::new(Inner::new(value.denom, Vec::new())),
+        })
+    }
 }
 
 /// A unit of some asset denomination.
--- a/pd/src/state.rs
+++ b/pd/src/state.rs
@@ -4,11 +4,13 @@ use anyhow::{Context, Result};
 use async_stream::try_stream;
 use futures::stream::{Stream, StreamExt};
 use penumbra_crypto::{
+    asset,
     merkle::{self, NoteCommitmentTree, TreeExt},
     Address, Nullifier,
 };
 use penumbra_proto::{
     light_wallet::{CompactBlock, StateFragment},
+    chain,
     thin_wallet::{Asset, TransactionDetail},
     Protobuf,
 };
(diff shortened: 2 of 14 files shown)


crypto/src/asset/denom.rs
```diff
@@ -5,19 +5,42 @@ use std::{
...
+    }
+}
+
+impl TryFrom<pb::Denom> for Denom {
+    type Error = anyhow::Error;
+
+    fn try_from(value: pb::Denom) -> Result<Self, Self::Error> {
+        Ok(Denom {
+            inner: Arc::new(Inner::new(value.denom, Vec::new())),
+        })
```
hdevalence, line 40: I don't think this is safe, since we're not validating that the denomination is valid (and also missing the metadata) -- could we do
```rust
asset::REGISTRY.parse_denom(&value.denom)
    .ok_or_else(|| anyhow::anyhow!("invalid denomination {}", denom))
```
instead?

crypto/src/asset/id.rs
```diff
@@ -37,6 +39,24 @@ impl Id {
...
+    }
+}
+
+impl TryFrom<pb::AssetId> for Id {
+    type Error = anyhow::Error;
+
+    fn try_from(value: pb::AssetId) -> Result<Self, Self::Error> {
+        value.asset_id.try_into()
+    }
+}
```
hdevalence, line 58: It looks like this is making the Protobuf AssetId store the bytes of the bech32 encoding, but since the proto is already a binary format, I think it would be cleaner if it just did `asset_id: ai.0.to_bytes()` on the way out to put the raw bytes of the asset ID in the proto.

To make the proto be Bech32 encoded when using JSON or some other serde serializer, you'd use the `build.rs` to stick the right annotations on the generated proto type, there are some existing examples to copy there.

pd/src/state.rs
```diff
@@ -496,18 +498,27 @@ ON CONFLICT (id) DO UPDATE SET data = $1
...
+            "SELECT denom, asset_id, total_supply FROM assets WHERE asset_id = $1",
             asset_id
         )
         .fetch_one(&mut conn)
         .await?;
-        Ok(Asset {
-            asset_denom: asset.denom,
-            asset_id: asset.asset_id,
+
+        // TODO: should we be returning proto types from our state methods, or domain types?
```
hdevalence, line 511: Proto types are good when they're being fed into RPC responses since there's no point in validating data that's already in our database before we ship it to the client.  If we consumed it internally, we could just `.try_into().expect()` on the proto, and then we're only paying the validation cost when we need the refined type.

pd/src/state.rs
```diff
@@ -496,18 +498,27 @@ ON CONFLICT (id) DO UPDATE SET data = $1
...
+        Ok(params::AssetInfo {
+            denom: Some(
+                asset::REGISTRY
+                    .parse_denom(asset.denom.as_str())
+                    .unwrap()
+                    .into(),
+            ),
+            asset_id: Some(asset::Id::try_from(asset.asset_id)?.into()),
+            total_supply: asset.total_supply.try_into()?, // postgres only has i64....
```
hdevalence, line 520: I think it'd be better to do `as u64` here, so that in the (unlikely) event that we'd recorded a `total_supply > 2^63`, we'd get the correct value out (since the `u64 <> i64` `as` conversion is lossless, just not semantically-preserving, and we don't do arithmetic inside the database).

pd/src/state.rs
```diff
@@ -496,18 +498,27 @@ ON CONFLICT (id) DO UPDATE SET data = $1
...
+            as_of_block_height: 0, // TODO: currently having the caller do this, should we instead pull the latest block height here?
```
hdevalence, line 521: Yeah, I think that it'd be better to pull the latest block height here, for two reasons: (a) it means that there's only one round-trip to the database, (b) it means that the response data is consistent.

We do all of the writes involved in processing a block transactionally, so although we have no guarantees about what order a given read query will execute relative to some block commit, we do know that each individual query will report consistent state. But if we query the height earlier, then later query the total supply, there could have been an intervening write, causing us to serve inconsistent data.

pd/src/wallet.rs
```diff
@@ -90,14 +93,24 @@ impl ThinWallet for State {
...
             .await
             .map_err(|_| tonic::Status::not_found("asset not found"))?;
+
+        let current_height = self
+            .height()
+            .await
+            .map_err(|_| tonic::Status::unavailable("database error"))?
+            .value() as u32;
+
+        // TODO should asset_lookup be doing this?
```
hdevalence, line 111: 👍 , xref https://github.com/penumbra-zone/penumbra/pull/346/files#r781783721

pd/src/wallet.rs
```diff
@@ -118,7 +131,7 @@ impl ThinWallet for State {
                     .map_err(|_| tonic::Status::unavailable("database error"))
                     .unwrap();
                 for asset in &assets[..] {
-                    tracing::debug!(asset_id = ?hex::encode(&asset.asset_id), asset_denom = ?asset.asset_denom, "sending asset");
+tracing::debug!(asset_id = ?hex::encode(&asset.asset_id), asset_denom = ?asset.asset_denom, "sending asset");
```
hdevalence, line 134: Not sure if the unindenting was a rebase artifact or something (it doesn't look like this changed?) but, while looking at it, it could be nice to format the asset ID using its `Display` impl.  If the parsing of the asset ID (which I think is raw bytes here?) is done *inside* the `debug!` macro expression (where the `hex::encode` is now), we won't pay any cost unless the tracing filter would actually use the output.

avahowell: I think this was just from running cargo fmt, weird

avahowell:
updated this with all of the relevant changes and added the remaining task from the issue. should be ready for review

pd/src/wallet.rs
```diff
@@ -90,14 +93,16 @@ impl ThinWallet for State {
...
-    ) -> Result<tonic::Response<Asset>, Status> {
-        tracing::debug!(asset_id = ?hex::encode(&request.get_ref().asset_id));
+        request: tonic::Request<AssetId>,
+    ) -> Result<tonic::Response<AssetInfo>, Status> {
+        let aid = request.into_inner().inner;
+        tracing::debug!(asset_id = ?hex::encode(&aid));
         let state = self.clone();
-        let asset = state
-            .asset_lookup(request.into_inner().asset_id)
+        let mut asset = state
```
hdevalence, line 101:
```suggestion
        let asset = state
```

hdevalence (approved):
LGTM!

Issue #228, opened 2021-12-06
Move to multi-tendermint setup for testnets


hdevalence:
Our first testnet is just a single instance of `pd`+`tendermint`. For MVP2 we'll want to have multiple validators. In advance of that, it would be helpful to move to multi-node tendermint, to help us catch any non-determinism.

hdevalence:
This is done in #347

PR #347, opened 2022-01-12
Add generate-testnet command to pd

zbuc:
This adds a new `generate-testnet` command to `pd` that can be used to generate Tendermint configurations for running a local testnet.

The testnet will be generated with randomly generated keys. Currently the Penumbra validator signing key is saved to the config directory; it's not currently used by anything but it would otherwise be unrecoverable.

```
pd-generate-testnet 0.1.0
Generates a directory structure containing necessary files to run a testnet based on input configuration

USAGE:
    pd generate-testnet [OPTIONS] --allocations-input-file <allocations-input-file> --chain-id <chain-id> --epoch-duration <epoch-duration> --num-validator-nodes <num-validator-nodes> --output-dir <output-dir> --validators-input-file <validators-input-file>

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

OPTIONS:
    -a, --allocations-input-file <allocations-input-file>    Path to CSV file containing initial allocations
    -c, --chain-id <chain-id>                                Testnet name, e.g. `penumbra-euporie`
    -e, --epoch-duration <epoch-duration>                    Number of blocks per epoch
    -n, --num-validator-nodes <num-validator-nodes>          How many validator nodes to create configuration for
    -o, --output-dir <output-dir>                            Path to directory to store output in. Must not exist
    -s, --starting-ip <starting-ip>
            IP Address to start `tendermint` nodes on. Increments by three to make room for `pd` and `postgres` per node
            [default: 192.167.10.2]
    -v, --validators-input-file <validators-input-file>      Path to JSON file containing initial validator configs
```


pd/src/main.rs
```diff
@@ -48,9 +50,31 @@ enum Command {
...
+        /// Path to directory to store output in. Must not exist.
+        #[structopt(short, long, parse(from_os_str))]
+        output_dir: PathBuf,
+        /// Testnet name, e.g. `penumbra-euporie`
+        #[structopt(short, long)]
+        chain_id: String,
+        /// IP Address to start `tendermint` nodes on. Increments by three to make room for `pd` and `postgres` per node.
+        #[structopt(short, long, default_value = "192.167.10.2")]
+        starting_ip: Ipv4Addr,
+    },
```
hdevalence, line 77: Should we specify default values for these fields?  

My sense is that: (a) we'll probably have more parameters in the future, and people might not want to specify all of them, and that (b) if we do have default values, it's better for them to be in the Rust source than in a wrapper script.

For instance, down below in the script, it plugs in 60 as a default value, but if we keep the defaults here, they'll all be in one place and we just have one thing to edit, and the default is provided to all of the users of the program, not just users of the script.

pd/src/main.rs
```diff
@@ -153,66 +177,285 @@ async fn main() -> anyhow::Result<()> {
...
         }
-        Command::CreateGenesisTemplate => {
-            use penumbra_crypto::keys::SpendKey;
+        Command::GenerateTestnet {
+            num_validator_nodes,
+            // TODO this config is gated on a "populate persistent peers"
+            // setting in the Go tendermint binary. Populating the persistent
+            // peers will be useful in local setups until peer discovery via a seed
+            // works.
+            starting_ip,
```
hdevalence, line 186: is the unusedness of this variable related to the TODO comment?

pd/src/main.rs
```diff
@@ -153,66 +177,285 @@ async fn main() -> anyhow::Result<()> {
...
-            };
+                    // always empty in genesis json
+                    app_hash: vec![],
+                    app_state: app_state,
+                    // List of initial validators. Note this may be overridden entirely by
+                    // the application, and may be left empty to make explicit that the
+                    // application will initialize the validator set with ResponseInitChain.
+                    // - https://docs.tendermint.com/v0.32/tendermint-core/using-tendermint.html
+                    // For penumbra, we can leave this empty since the app_state also contains Validator
+                    // configs.
```
hdevalence, line 362: 👍

pd/src/main.rs
```diff
@@ -153,66 +177,285 @@ async fn main() -> anyhow::Result<()> {
...
+                );
+                let mut config_file = File::create(config_file_path)?;
+                config_file.write_all(tm_config.as_bytes())?;
+
+                // Write this node's node_key.json
+                // the underlying type doesn't implement Copy or Clone (for the best)
+                let priv_key: tendermint::PrivateKey;
+                unsafe {
+                    priv_key = mem::transmute_copy(&vk.node_key_sk);
+                }
```
hdevalence, line 396: I'm OK with this for now, but it would be nice not to use this long-term, maybe we could add a `Clone` impl upstream?

pd/src/main.rs
zbuc: I'll set the defaults to be the latest testnet configurations.

pd/src/main.rs
zbuc: Yes. It will be fixed by #228

testnets/003-eupheme/validators.json
```diff
@@ -0,0 +1,15 @@
...
+        "funding_streams": [
+            [
+                200,
+                "penumbrav0t14k6uhn8pnuvm869h3kgfywk03cn8nxv4fgu2ehdjdrgk2277lgsp7qrzduzyka2v3pfsy8py259ge9eq36nq8czsjgjsm5rv28s9mq54z3yw3veyz52lt0qsfp29w3"
+            ]
+        ],
+        "sequence_number": 0,
+        "voting_power": 100
+    }
+]
```
hdevalence, line 15: Since we already created eupheme, could we move these new files to a new `004-thelxinoe` directory?

hdevalence:
This looks great, I'm really excited to try it out!!

testnets/003-eupheme/validators.json
zbuc: @hdevalence we can generate and commit a `genesis.json` to the directory after deploying the testnet

pd/src/main.rs
```diff
@@ -48,9 +50,41 @@ enum Command {
...
     },
 
-    /// Prints a sample `app_data` JSON object that can act as a template for
-    /// editing genesis configuration.
-    CreateGenesisTemplate,
+    /// Generates a directory structure containing necessary files to run a
+    /// testnet based on input configuration.
+    GenerateTestnet {
+        /// How many validator nodes to create configuration for.
+        #[structopt(short, long, default_value = "4")]
```
plaidfinch, line 57: Why is the default 4 here? I suspect most people who want to run a testnet node will probably only want to run one.

pd/src/main.rs
```diff
@@ -153,66 +187,286 @@ async fn main() -> anyhow::Result<()> {
...
+                num_validator_nodes > 0,
+                "must have at least one validator node"
+            );
 
-            let validator_id_sk = SigningKey::<SpendAuth>::new(OsRng);
-            let validator_id_vk = VerificationKey::from(&validator_id_sk);
+            let genesis_time = Time::from_unix_timestamp(
+                SystemTime::now()
+                    .duration_since(UNIX_EPOCH)
+                    .expect("time travels linearly in a forward direction")
```
plaidfinch, line 230: lolol we can only hope :)

pd/src/testnet.rs
```diff
@@ -0,0 +1,671 @@
...
+use penumbra_crypto::Address;
+use tendermint::PrivateKey;
+
+use crate::genesis;
+
+/// Methods and types used for generating testnet configurations.
+
+pub fn parse_allocations_file(input_file: PathBuf) -> Result<Vec<TestnetAllocation>> {
+    let file = match File::open(&input_file) {
+        Err(why) => panic!("couldn't open {}: {}", input_file.display(), why),
```
plaidfinch, line 20: Instead of a `panic!`, I would recommend using the `.context()` method from the `anyhow::Context` trait to add additional context to the error, then propagating it as an `anyhow::Error`.

pd/src/testnet.rs
```diff
@@ -0,0 +1,671 @@
...
+        let record: TestnetAllocation = result?;
+        res.push(record);
+    }
+
+    Ok(res)
+}
+
+pub fn parse_validators_file(input_file: PathBuf) -> Result<Vec<TestnetValidator>> {
+    let file = match File::open(&input_file) {
+        Err(why) => panic!("couldn't open {}: {}", input_file.display(), why),
```
plaidfinch, line 36: Instead of a `panic!`, I would recommend using the `.context()` method from the `anyhow::Context` trait to add additional context to the error, then propagating it as an `anyhow::Error`.

pd/src/testnet.rs
```diff
@@ -0,0 +1,671 @@
...
+
+    deserializer.deserialize_any(U64StringVisitor)
+}
+
+/// Hardcoded Tendermint config template. Should produce tendermint config similar to
+/// https://github.com/tendermint/tendermint/blob/6291d22f46f4c4f9121375af700dbdafa51577e7/cmd/tendermint/commands/init.go#L45
+/// There exists https://github.com/informalsystems/tendermint-rs/blob/a12118978f2ffea4042d6d38ebfb290d12611314/config/src/config.rs#L23 but
+/// this seemed more straightforward as only the moniker is changed right now.
+pub fn generate_tm_config(node_name: &str) -> String {
+    format!(
```
plaidfinch, line 82: This large format string can be put in a separate file by doing `format!(include_str!("tm_config_template.toml"), node_name)`.

pd/src/testnet.rs
```diff
@@ -0,0 +1,671 @@
...
+    "round": 0,
+    "step": 0
+}
+"#
+    .to_string()
+}
+
+/// Expand tildes in a path.
+/// Modified from https://stackoverflow.com/a/68233480
+pub fn canonicalize_path(input: &str) -> PathBuf {
```
plaidfinch, line 653: Will this work on non-*nix operating systems?

plaidfinch (commented):
This is really great work. I'm excited to try it out!

pd/src/main.rs
zbuc: Good question. Right now I'm optimizing the defaults to run a self-contained testnet (i.e. for the Penumbra team doing local development) that doesn't connect to any peers over the internet.

However you're right that most users wanting to operate a node connecting to the existing testnet will only want to run one.

These are two different use cases and I don't think it makes sense for users attempting to add an additional node to the testnet to use this script, they should instead have an alternative method documented that will connect to the correct seed peers for the running testnet.

pd/src/testnet.rs
zbuc: I think it should. I can give it a test on my Windows machine later

pd/src/main.rs
hdevalence: I'm not sure that there's actually a tension here -- users who want to run a node for the main testnet aren't generating their own testnet, they're joining an existing one. So I don't think they'd be using this command at all.

PR #350, opened 2022-01-12
Add tests for roundtrip parsing of note commitments, and fix parse error

plaidfinch:
This fixes the error on latest `main` where `pcli wallet reset && pcli balance` would result in:

```
Error: Could not parse wallet data

Caused by:
    tag for enum is not valid, found 64
```

### DIFF
--- a/proto/build.rs
+++ b/proto/build.rs
@@ -93,5 +93,4 @@ static FIELD_ATTRIBUTES: &[(&str, &str)] = &[
     (".penumbra.stake.IdentityKey.ik", AS_BECH32_IDENTITY_KEY),
     (".penumbra.crypto.Address.inner", AS_BECH32_ADDRESS),
     (".penumbra.crypto.AssetId.inner", AS_BECH32_ASSET_ID),
-    (".penumbra.crypto.NoteCommitment.inner", AS_HEX),
 ];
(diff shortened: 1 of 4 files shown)


proto/build.rs
```diff
@@ -93,5 +93,4 @@ static FIELD_ATTRIBUTES: &[(&str, &str)] = &[
     (".penumbra.stake.IdentityKey.ik", AS_BECH32_IDENTITY_KEY),
     (".penumbra.crypto.Address.inner", AS_BECH32_ADDRESS),
     (".penumbra.crypto.AssetId.inner", AS_BECH32_ASSET_ID),
-    (".penumbra.crypto.NoteCommitment.inner", AS_HEX),
```
plaidfinch, line 96: Without the `AS_HEX`, this passes the roundtrip parse test, and `pcli` can sync the chain correctly, picking up where it left off by re-parsing the NCT appropriately. It's not immediately clear to me why adding `AS_HEX` here is causing a deserialization failure for bincode. Any ideas?

hdevalence, line 96: Hmm, what does this do to the JSON format used for the wallet.json file?

hdevalence: I'm really unsure why this would cause a deserialization failure.  From the bytes you posted, it looked like the encoding was 8 bytes of length encoding "64", then 64 bytes of UTF-8 text with the hex encoding.  The error we got on deserialization occurs when bincode is trying to decode an `Option`; it errors because it's expecting 0 (`None`) or 1 (`Some`), but sees 64, the length field, and treats it as an enum discriminant.

One problem with this fix is that I think it will screw up the JSON format used for wallet.json; maybe that's less important than having it work, but it's something to consider.

hdevalence (approved):

PR #353, opened 2022-01-12
Use an associated type for the Protobuf trait

plaidfinch:
This prevents anyone from making two impls of `Protobuf` for different domain types that are backed by the same protobuf type, ensuring a 1:1 mapping of domain types to protobuf types.

### DIFF
--- a/chain/src/params.rs
+++ b/chain/src/params.rs
@@ -9,7 +9,9 @@ pub struct AssetInfo {
     pub total_supply: u64,
 }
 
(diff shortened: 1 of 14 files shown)


plaidfinch:
Fixing any in-flight branches to use the new trait is a quick search-replace (which is how i generated this PR):

`s/impl Protobuf<(.*)> for (.*) {}/impl Protobuf for \2 { type Protobuf = \1; }/`

hdevalence:
Closing this for now; since it's mechanical, we could redo it at some future point.

PR #354, opened 2022-01-13
add validator status, compute voting power correctly at epoch boundary

### LINKED ISSUE #336: Extract voting power from validator rates and add into validator status
Prior to #321, we recorded the voting power in the `Validator` type.  In #321, this was removed, as the `Validator` type was changed to include only the configuration data inherent to the validator, so that it could be used in a `ValidatorDefinition` transaction action. 

However, this left a vacuum for where we should record information about the on-chain state of the validator.  The voting power was moved into the `RateData` struct that describes a validator's exchange rates, but it doesn't really fit there either, because the rest of the `RateData` fields are also in some sense inherent to the validator's config, while the voting power depends also on the supply of delegation tokens (#335).

We need to track the state of a validator for other reasons, too: a validator can be part of the consensus set (active), recently bumped out of the consensus set (unbonding), removed and slashed for some offensive behavior (jailed) or simply  not part of the consensus set at all (inactive).

In #323, we added a stub `ValidatorStatus` proto (with no corresponding domain type yet). We should change the proto to something like:
```proto
message ValidatorStatus {
  Validator validator = 1;
  RateData rate_data = 2;
  uint64 delegation_token_supply = 3;
  // In the future, add info on validator state
}
```
and add a domain type for it.

The thin wallet RPC has a
```proto
  rpc ValidatorInfo(stake.IdentityKey) returns (stake.ValidatorDefinition);
```
which should probably be changed to
```proto
  rpc ValidatorStatus(stake.IdentityKey) returns (stake.ValidatorStatus);
```

### LINKED ISSUE #337: Refactor spec+code to avoid computation of the voting power adjustment function
Currently, we calculate the voting power adjustment function when computing the validator rates modeled by the `RateData` structure in order to compute the voting power.  Per the reasoning in #336, it would be better for the voting power not to be recorded as part of the `RateData`.

We also currently specify the fixed-point arithmetic for computing the voting power adjustment function. But, this value is only used for computing the voting power, so per discussion with @avahowell, it seems better to specify the entire voting power calculation as one function.  This computation as input the total supply of delegation tokens (tracked in #335), the validator exchange rate, and the base exchange rate.

Thinking about where would be a good place to put it: the primary case where we want to compute it is when we're processing changes at epoch boundaries, where we've already got a single `BaseRateData`, and for each validator we can query the current rates and delegation token supply.  So one convenient option for use in `pd` would be
```rust
RateData::voting_power(&self, base_rate_data: &BaseRateData, delegation_token_supply: u64) -> u64
```
We could also add a convenience method on `ValidatorStatus` that uses the delegation token supply it records:
```
ValidatorStatus::voting_power(&self, base_rate_data: &BaseRateData) -> u64
```
but I'm not sure if we have a good use case for that at the moment, so we could skip it for now.

avahowell:
this PR implements correct calculation of validator's voting power at every epoch boundary, moves the voting power field to the validators table, and introduces a new `ValidatorStatus` struct which contains the voting power and identity key, currently.

Closes #336 
Closes #337


proto/proto/thin_wallet.proto
```diff
@@ -14,7 +14,7 @@ service ThinWallet {
   rpc AssetLookup(crypto.AssetId) returns (chain.AssetInfo);
   rpc AssetList(AssetListRequest) returns (stream Asset);
   // TODO: return ValidatorStatus?
-  rpc ValidatorInfo(stake.IdentityKey) returns (stake.ValidatorDefinition);
+  rpc ValidatorStatus(stake.IdentityKey) returns (stake.ValidatorStatus);
```
hdevalence, line 17: previously, the validator definition included the actual validator data, but now that the validator status just includes the identity key, we might want a second RPC...

hdevalence (approved):
LGTM! I tacked on a commit at the end that restores some debug output we removed while refactoring.

PR #355, opened 2022-01-13
pd: make next epoch's rates available

hdevalence:
Staking actions are supposed to make use of the rates in the epoch in which
they take effect (the next epoch), so we need to compute them in advance. This
commit changes the end-epoch processing to more cleanly model the epoch
transition (it happens after the end of the last block in the epoch) and
precompute the next epoch's rates.

### DIFF
--- a/pd/src/app.rs
+++ b/pd/src/app.rs
@@ -355,22 +355,32 @@ impl App {
         let state = self.state.clone();
         async move {
             if epoch.end_height().value() == height {
+                // We've finished processing the last block of `epoch`, so we've
+                // crossed the epoch boundary, and (prev | current | next) are:
+                let prev_epoch = epoch;
+                let current_epoch = prev_epoch.next();
+                let next_epoch = current_epoch.next();
+
                 tracing::info!(
                     ?height,
-                    ?epoch,
-                    "last block of epoch, processing rate updates"
+                    ?prev_epoch,
+                    ?current_epoch,
+                    ?next_epoch,
+                    "crossed epoch boundary, processing rate updates"
                 );
                 metrics::increment_counter!("epoch");
 
                 // TODO (optimization): batch these queries
-                let current_base_rate = state.base_rate_data(epoch.index).await?;
-                let current_rates = state.rate_data(epoch.index).await?;
+                let current_base_rate = state.base_rate_data(current_epoch.index).await?;
+                let current_rates = state.rate_data(current_epoch.index).await?;
 
                 // steps (foreach validator):
                 // - get the total token supply for the validator's delegation tokens
-                // - process the updates to the token supply
-                // - call next_rates.voting_power()
-                // - persist both the voting power and the updated supply
+                // - process the updates to the token supply:
+                //   - collect all delegations occurring in previous epoch and apply them (adds to supply);
+                //   - collect all undelegations started in previous epoch and apply them (reduces supply);
+                // - feed the updated (current) token supply into current_rates.voting_power()
+                // - persist both the current voting power and the current supply
 
                 /// FIXME: set this less arbitrarily, and allow this to be set per-epoch
                 /// 3bps -> 11% return over 365 epochs, why not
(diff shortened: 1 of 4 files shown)


hdevalence:
```
Jan 13 00:34:57.666  INFO abci:BeginBlock{height=block::Height(689) hash="fd61fb0b2edeb9e8b815d61fe7b2660357570aa440fd75dd6570d955429d6680"}: pd::app: rsp=BeginBlock(BeginBlock { events: [] })
Jan 13 00:34:57.667  INFO abci:EndBlock{height=689}: pd::app: crossed epoch boundary, processing rate updates height=689 prev_epoch=Epoch { index: 68, duration: 10 } current_epoch=Epoch { index: 69, duration: 10 } next_epoch=Epoch { index: 70, duration: 10 }
Jan 13 00:34:57.686 DEBUG abci:EndBlock{height=689}: pd::app: curr_base_rate=BaseRateData { epoch_index: 69, base_reward_rate: 30000, base_exchange_rate: 102060611 }
Jan 13 00:34:57.686 DEBUG abci:EndBlock{height=689}: pd::app: next_base_rate=BaseRateData { epoch_index: 70, base_reward_rate: 30000, base_exchange_rate: 102091229 }
Jan 13 00:34:57.700 DEBUG abci:EndBlock{height=689}: pd::app: curr_rate=RateData { identity_key: penumbravalid1f2zfhqe8ctc3vx4qgs37a4j9x2vytk7sejew3dkh53qxjeu0lcgq3etsdt, epoch_index: 69, validator_reward_rate: 29400, validator_exchange_rate: 101988998 }
Jan 13 00:34:57.700 DEBUG abci:EndBlock{height=689}: pd::app: next_rate=RateData { identity_key: penumbravalid1f2zfhqe8ctc3vx4qgs37a4j9x2vytk7sejew3dkh53qxjeu0lcgq3etsdt, epoch_index: 70, validator_reward_rate: 29400, validator_exchange_rate: 102018982 }
Jan 13 00:34:57.700 DEBUG abci:EndBlock{height=689}: pd::app: next_status=ValidatorStatus { identity_key: penumbravalid1f2zfhqe8ctc3vx4qgs37a4j9x2vytk7sejew3dkh53qxjeu0lcgq3etsdt, voting_power: 0 }
Jan 13 00:34:58.800  INFO abci:BeginBlock{height=block::Height(690) hash="5093fd21234a1b8466062b96d867afe67bd0a2c55f69a9ede4e61799fa7e4868"}: pd::app: rsp=BeginBlock(BeginBlock { events: [] })
Jan 13 00:34:59.941  INFO abci:BeginBlock{height=block::Height(691) hash="bfc49b5d512f84b332f14d373f9ac05f7c3c3faaf21b185859cbe8a995eae764"}: pd::app: rsp=BeginBlock(BeginBlock { events: [] })
```

pd/src/app.rs
```diff
@@ -355,22 +355,32 @@ impl App {
...
-                // - call next_rates.voting_power()
-                // - persist both the voting power and the updated supply
+                // - process the updates to the token supply:
+                //   - collect all delegations occurring in previous epoch and apply them (adds to supply);
+                //   - collect all undelegations started in previous epoch and apply them (reduces supply);
+                // - feed the updated (current) token supply into current_rates.voting_power()
+                // - persist both the current voting power and the current supply
 
                 /// FIXME: set this less arbitrarily, and allow this to be set per-epoch
                 /// 3bps -> 11% return over 365 epochs, why not
```
plaidfinch, line 386: Is this meant to be 30,000? `3_0000` is a kind of misleading way to write that, since it looks like 3,000 at a glance.

pd/src/state.rs
```diff
@@ -41,6 +41,14 @@ impl State {
         tracing::info!("running migrations");
         sqlx::migrate!("./migrations").run(&pool).await?;
         tracing::info!("finished initializing state");
+
+        let start = std::time::Instant::now();
+        let mut conn = pool.acquire().await?;
+        query!("SELECT true").fetch_all(&mut conn).await?;
+        let end = start.elapsed();
+        let ms = (end.as_micros() as f64) / 1000.;
+        tracing::info!("no-op query took {}ms ({:.0} per second)", ms, 1000. / ms);
```
plaidfinch, line 50: Non-blocking: Instead of inlining the info into the string, should we report these as fields in the tracing message so they could be mechanically consumed?

pd/src/state.rs
```diff
@@ -59,11 +67,23 @@ INSERT INTO blobs (id, data) VALUES ('gc', $1)
...
+        for epoch in [0, 1] {
+            query!(
+                "INSERT INTO base_rates (
+                epoch, 
+                base_reward_rate, 
+                base_exchange_rate
+            ) VALUES ($1, $2, $3)",
+                epoch,
+                0,
+                1_0000_0000
```
plaidfinch, line 82: This is written somewhat confusingly: `1_0000_0000` looks at a glance like "1 million" but it's actually "100 million". Is this intentional?

plaidfinch (commented):
A couple places where it looks like some literals might be incorrect, or written in a way that is confusing. If those places check out, LGTM.

pd/src/app.rs
hdevalence: I's not meant to be 30,000, it's meant to be 3 basis points, and our rates are expressed in basis points of basis points (i.e., a factor of 10^(4+4)), so we want a value of 3 * 10^4 (which is 30,000, but only incidentally).

pd/src/state.rs
hdevalence: No objection, but the message isn't intended to be mechanically consumed -- I think that metrics would be better for that than tracing messages -- it's just meant to give a ballpark value that primes our intuition about the cost model for making queries.

pd/src/state.rs
hdevalence: Yep, the value is in basis points of basis points, so the digits are grouped in 10^4-sized chunks, and this is supposed to be read as "1.0000.0000"

PR #356, opened 2022-01-14
Track delegations per epoch.

hdevalence:
Work with @plaidfinch on tracking (un)delegations during an epoch.

There are two parts to the database changes:

- a `delegation_changes` table used for tracking the changes to delegation token supplies;
- some `undelegation_*` tables which are not currently used, but will be used for doing unbonding.

In this PR, only the delegation changes are used. We validate the (un)delegate descriptions statefully against the rate data for the next epoch.


hdevalence:
Still to do in future work:

- Use the unbonding tables to implement an unbonding period
- Using the data this PR saves to the database to update the token supply at the epoch boundaries
- Creating Delegate and Undelegate descriptions in a transaction

PR #360, opened 2022-01-14
Add tracking of ChainParams to pcli wallet

### LINKED ISSUE #329: Track chain parameters in `pcli`
Following #326 #328, we should have `pcli` track the chain parameters:

- [x] add an RPC method to the light wallet proto that takes no input data and returns the `ChainParams` (this can go in the light wallet, not the thin wallet, since it doesn't have any info about the client data that wouldn't already be revealed by syncing);
- [x] add the `ChainParams` structure to the client state in `penumbra-wallet`, and require it to be passed into the constructor (*);
- [x] change the sync logic so that, before starting sync on an empty state, the client fetches and saves the chain params (*);

(*): I don't recall exactly how the default / empty case is handled in the client state, if it's filled in with default values instead of being missing, we might have to remodel the data a bit ?

zbuc:
This will automatically sync the ChainParams to `pcli` when initially instantiated. There's currently no support for updating chain parameters afterwards. This shouldn't be an issue for now as they're currently static for the lifetime of the chain, but that may not always be the case.

Fixes #329

### DIFF
--- a/chain/src/params.rs
+++ b/chain/src/params.rs
@@ -1,3 +1,5 @@
+use serde::{Deserialize, Serialize};
+
 use penumbra_crypto::asset;
 use penumbra_proto::{chain as pb, crypto as pbc, Protobuf};
 
@@ -35,7 +37,7 @@ impl From<AssetInfo> for pb::AssetInfo {
     }
 }
 
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct ChainParams {
     pub chain_id: String,
     pub epoch_duration: u64,
--- a/pcli/src/sync.rs
+++ b/pcli/src/sync.rs
@@ -1,6 +1,6 @@
 use anyhow::Result;
 use penumbra_proto::light_wallet::{
-    light_wallet_client::LightWalletClient, CompactBlockRangeRequest,
+    light_wallet_client::LightWalletClient, ChainParamsRequest, CompactBlockRangeRequest,
 };
 use tracing::instrument;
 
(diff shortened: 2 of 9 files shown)


chain/src/params.rs
```diff
@@ -35,7 +37,7 @@ impl From<AssetInfo> for pb::AssetInfo {
     }
 }
 
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Serialize, Deserialize)]
```
zbuc, line 40: This is necessary because of `ClientState`

hdevalence: Since the `ChainParams` is the domain type for a proto type, we want to add Serde annotations so that the Serde format is derived from the proto formats, not from our internal representation:
 
```suggestion
#[derive(Clone, Debug, Serialize, Deserialize)]
#[serde(try_from = pb::ChainParams, into = pb::ChainParams)]
```

This change won't work on its own, because currently the proto-derived `pb::ChainParams` type doesn't implement the Serde traits. To do that, we need to edit the `build.rs` in the `proto` crate to add
```
(".penumbra.chain.ChainParams", SERIALIZE)
```
to `TYPE_ATTRIBUTES` so that Prost will add the serde attributes when it generates the proto types.

pcli/src/sync.rs
```diff
@@ -36,3 +36,21 @@ pub async fn sync(state: &mut ClientStateFile, wallet_uri: String) -> Result<()>
...
+        .chain_params(tonic::Request::new(ChainParamsRequest {}))
+        .await?
+        .into_inner();
+
+    let cp_mut = state.chain_params_mut();
+    *cp_mut = Some(params.into());
+    state.commit()?;
+    tracing::info!(end_height = ?state.last_block_height().unwrap(), "finished sync");
+    Ok(())
+}
```
hdevalence, line 56: I think this would fit better in the `fetch` module as `fetch::chain_params`, and rather than editing the state itself, it could just fetch the params, so the caller can do whatever with them (either set them, or something else).

wallet/src/state.rs
```diff
@@ -85,7 +88,7 @@ impl AsRef<Note> for UnspentNote<'_> {
 }
 
 impl ClientState {
-    pub fn new(wallet: Wallet) -> Self {
+    pub fn new(wallet: Wallet, chain_params: Option<ChainParams>) -> Self {
```
hdevalence, line 91: The `ClientState` is only ever constructed with `chain_params = None`, so maybe it would be simpler to skip passing it as an argument, set it to `None` by default, and use the existing logic to set it if it doesn't exist?

hdevalence (approved):
This looks good, I left some review comments but those suggestions are all non-blocking, while they'd be good to address at some point in the future I think it would be easier to do so after merging this together with other staking changes.

pcli/src/sync.rs
hdevalence: Did this while resolving merge conflicts with #361

Issue #48, opened 2021-09-08
Implement undelegations


hdevalence:
Implement #47:

- add implementations of undelegation transaction creation to client libraries and expose in `pcli`;
- add implementation of undelegation queue to `pd`;
- add implementation of undelegation checks to `pd`;

hdevalence:
This is done now.

PR #361, opened 2022-01-17
Implement delegations and undelegations

### LINKED ISSUE #45: Implement delegation transactions
Implement #44: add implementations of transaction creation to underlying client libraries and expose them through `pcli`.

hdevalence:
Closes #45 #48, and does cleanup to `pcli` and the transaction builder code along the way; more details on the steps are in the commit messages.

### DIFF
--- a/pcli/src/command/stake.rs
+++ b/pcli/src/command/stake.rs
@@ -1,9 +1,6 @@
 use anyhow::{anyhow, Context, Result};
 use penumbra_crypto::Value;
-use penumbra_proto::{
-    thin_wallet::{thin_wallet_client::ThinWalletClient, ValidatorRateRequest},
-    Protobuf,
-};
+use penumbra_proto::thin_wallet::ValidatorRateRequest;
 use penumbra_stake::{DelegationToken, Epoch, IdentityKey, RateData, STAKING_TOKEN_ASSET_ID};
 use rand_core::OsRng;
 use structopt::StructOpt;
@@ -101,11 +98,7 @@ impl StakeCmd {
                     Epoch::from_height(state.last_block_height().unwrap() as u64, epoch_duration);
                 let next_epoch = current_epoch.next();
 
-                let mut client = ThinWalletClient::connect(format!(
-                    "http://{}:{}",
-                    opt.node, opt.thin_wallet_port
-                ))
-                .await?;
+                let mut client = opt.thin_wallet_client().await?;
 
                 let rate_data: RateData = client
                     .validator_rate(tonic::Request::new(ValidatorRateRequest {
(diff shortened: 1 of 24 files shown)


pcli/src/command/stake.rs
```diff
@@ -25,10 +25,7 @@ pub enum StakeCmd {
     },
     /// Withdraw stake from a validator's delegation pool.
     Undelegate {
-        /// The identity key of the validator to withdraw delegation from.
-        #[structopt(long)]
-        from: String,
```
hdevalence, line 30: I removed this since it turns out it's already implied by the delegation token denomination.  Alternatively, we could bring it back as an optional parameter, and allow the amount to be specified in either unit (i.e., i can just give the amount of delegation tokens to cash in, or, i can name a validator and an amount of unbonded stake, and have it compute the amount of delegation tokens).

hdevalence:
Going to rebase this onto main now that #360 is merged

wallet/src/state.rs
```diff
@@ -188,12 +227,168 @@ impl ClientState {
...
+        let (_label, self_address) = self
+            .wallet()
+            .address_by_index(source_address.unwrap_or(0) as usize)?;
+
+        let mut tx_builder = Transaction::build_with_root(self.note_commitment_tree.root2());
+
+        tx_builder
+            .set_fee(fee)
+            // XXX populate chain id from params?
+            .set_chain_id(CURRENT_CHAIN_ID.to_string())
```
hdevalence, line 251: Now that we have `ChainParams`, we should look through all calls to `.set_chain_id` and have them pass a string from the chain params, rather than a hardcoded constant.

wallet/src/state.rs
```diff
@@ -188,12 +227,168 @@ impl ClientState {
...
+                asset_id: rate_data.identity_key.delegation_token().id(),
+            },
+            memo::MemoPlaintext([0u8; memo::MEMO_LEN_BYTES]),
+            self.wallet.outgoing_viewing_key(),
+        );
+
+        let change_amount = spent_amount - spend_amount;
+        // TODO: support dummy notes, and produce a change output unconditionally.
+        // let change_note = if change_amount > 0 { ... } else { /* dummy note */}
+        if change_amount > 0 {
```
hdevalence, line 281: This will be important since we reveal the actions in a transaction, and we'd ideally not reveal whether or not the delegation spent an exact note amount, since that leaks a fair amount of informaiton.

wallet/src/state.rs
```diff
@@ -188,12 +227,168 @@ impl ClientState {
...
+                fee
+            )
+        })?;
+
+        let delegation_denom = rate_data.identity_key.delegation_token().denom();
+
+        // XXX if the undelegation is for less than their total amount of delegation tokens,
+        // all of their remaining delegation tokens will also be quarantined.
+        // this sucks lmao
+        let mut spent_amount = 0;
```
hdevalence, line 342: Per discussion on discord, the solution to this is to pre-prepare a note with the exact number of delegation tokens.  Doing this will require some other changes to be able to track transaction completion.

hdevalence:
Merging so we can take stock of some follow-on tasks.

PR #362, opened 2022-01-18
Unbonding period

plaidfinch:


pd/src/app.rs
```diff
@@ -39,6 +40,9 @@ const NUM_RECENT_ANCHORS: usize = 64;
 /// The Penumbra ABCI application.
 #[derive(Debug)]
 pub struct App {
+    /// Parameters for the chain (determined at genesis and thenceforth by governance).
+    pub chain_params: ChainParams,
```
hdevalence, line 44: Should this be `pub` ?  Thoughts:

(1) Currently we don't ever access the `App` externally, we wrap it in a bunch of Tower stuff, so I'm not sure what would be accessing it;
(2) IME it's useful to have POD structures with public fields, or structures with invariants and private fields, but I haven't had great success with mixing the two.

pd/src/genesis.rs
```diff
@@ -47,29 +48,19 @@ impl Allocation {
     }
 }
 
-#[derive(Deserialize, Serialize, Debug, Clone)]
+#[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct ValidatorPower {
     pub validator: Validator,
     pub power: tendermint::vote::Power,
 }
```
hdevalence, line 55: Non-blocking on this PR, but we should add docs to this struct that explains its purpose, relative to `Validator` and `ValidatorStatus`.  The `Validator` domain type has all of the config data for a validator, and the `ValidatorStatus` has the voting power, (in the future) the validator state (inactive/active/slashed/unbonding), etc.

As we expand the `ValidatorStatus` structure, we'll need to replicate its fields here, or change this structure to be a `Validator`, `ValidatorStatus` pair.

The only reason not to change the structure to be a pair would be that the `ValidatorStatus` references a validator by identity key, so if we had the validators specified that way in the genesis config, it would be necessary to copy the identity key in two different places. Maybe that's better than maintaining a custom, genesis-only type?

pd/src/genesis.rs
```diff
@@ -47,29 +48,19 @@ impl Allocation {
...
 
 /// The application state at genesis.
-#[derive(Deserialize, Serialize, Debug, Clone)]
+#[derive(Debug, Clone, Serialize, Deserialize, Default)]
 pub struct AppState {
     /// The number of blocks in each epoch.
-    pub epoch_duration: u64,
+    pub chain_params: ChainParams,
```
hdevalence, line 61: outdated doc comment?

pd/src/pending_block.rs
```diff
@@ -16,14 +17,21 @@ pub struct PendingBlock {
     pub note_commitment_tree: NoteCommitmentTree,
     /// Stores note commitments for convienience when updating the NCT.
     pub notes: BTreeMap<note::Commitment, PositionedNoteData>,
+    /// Notes that are undelegation outputs in this block (not to be committed to the NCT yet).
+    pub undelegation_notes: BTreeMap<note::Commitment, QuarantinedNoteData>,
```
hdevalence, line 21: Should this be called `quarantined_notes` to match `quarantined_nullifiers`?

pd/src/pending_block.rs
```diff
@@ -16,14 +17,21 @@ pub struct PendingBlock {
...
+    /// with their unbonding epoch and validator identity key.
+    pub quarantined_nullifiers: BTreeMap<Nullifier, u64>,
     /// Records any updates to the token supply of some asset that happened in this block.
     pub supply_updates: BTreeMap<asset::Id, (asset::Denom, u64)>,
     /// Indicates the epoch the block belongs to.
     pub epoch: Option<Epoch>,
     /// Indicates the duration in blocks of each epoch.
     pub epoch_duration: u64,
+    /// Indicates the number of epochs before undelegated stake is unbonded.
+    pub unbonding_epochs: u64,
```
hdevalence, line 34: At this point we've replicated two fields from the chain parameters into the pending block -- should we just have it store the whole set of parameters?

pd/src/state.rs
```diff
@@ -152,15 +152,18 @@ INSERT INTO blobs (id, data) VALUES ('gc', $1)
         let app_hash = nct_anchor.to_bytes();
         let height = block.height.expect("height must be set");
 
+        // Any outputs in this block marked as quarantined will be delayed until this epoch index:
+        let unbonding_epoch_index =
+            block.epoch.as_ref().expect("block has epoch").index + block.unbonding_epochs;
+
```
hdevalence, line 158: Not sure if I'm missing something, but is this currently being used?

pd/src/state.rs
```diff
@@ -198,6 +201,15 @@ ON CONFLICT (id) DO UPDATE SET data = $1
...
 
+        // Add undelegated note outputs into the quarantine queue.
+        for (pre_position, (note_commitment, quarantined_note)) in
+            block.undelegation_notes.into_iter().enumerate()
+        {
+            // TODO: quarantine undelegated note outputs
+        }
+
+        // TODO: quarantine nullifiers
+
```
hdevalence, line 212: Should we fill this in before merging -- are there open questions on how we'd do that, or would it be straightforward?

Also, there should be a third step, which would be to add previously quarantined notes into the note commitment tree as they exit the unbonding period, right?

pd/src/verify.rs
```diff
@@ -29,8 +29,12 @@ pub struct VerifiedTransaction {
     pub new_notes: BTreeMap<note::Commitment, NoteData>,
     /// List of spent nullifiers from spends in this transaction.
     pub spent_nullifiers: BTreeSet<Nullifier>,
-    /// Net delegations performed in this transaction.
+    /// Net delegations performed in this transaction. An identity key mapped to zero is different
+    /// from an identity key that is absent; the former indicates that a validator's net change in
+    /// delegation in this transaction was zero *but it experienced some (un)delegations*.
```
hdevalence, line 34: style nit: could "An identity key..." move down a paragraph, so it's easier to parse the "summary / details" format while scanning the code?

(good observation though)

pd/src/verify.rs
```diff
@@ -269,6 +290,12 @@ impl StatefulTransactionExt for PendingTransaction {
             new_notes: self.new_notes.clone(),
             spent_nullifiers: self.spent_nullifiers.clone(),
             delegation_changes,
+            undelegation_validators: self
+                .undelegations
+                .iter()
+                .cloned()
+                .map(|u| u.validator_identity)
```
hdevalence, line 297: nit: I think this would be epsilon-more-efficient if these lines were swapped, so that we're just cloning the identities and not the whole undelegations.

proto/build.rs
```diff
@@ -84,6 +84,7 @@ static TYPE_ATTRIBUTES: &[(&str, &str)] = &[
     (".penumbra.crypto.Value", SERIALIZE),
     (".penumbra.crypto.Denom", SERIALIZE),
     (".penumbra.crypto.Denom", SERDE_TRANSPARENT),
+    (".penumbra.chain.ChainState", SERIALIZE),
```
hdevalence, line 87: Is this supposed to be `ChainParams`? I don't think we have a `ChainState`.  If it is supposed to be `ChainParams`, actually using the serialize impl requires an extra annotation, as in #363

pd/src/pending_block.rs
```diff
@@ -16,14 +17,21 @@ pub struct PendingBlock {
     pub note_commitment_tree: NoteCommitmentTree,
     /// Stores note commitments for convienience when updating the NCT.
     pub notes: BTreeMap<note::Commitment, PositionedNoteData>,
+    /// Notes that are undelegation outputs in this block (not to be committed to the NCT yet).
+    pub undelegation_notes: BTreeMap<note::Commitment, QuarantinedNoteData>,
     /// Nullifiers that were spent in this block.
     pub spent_nullifiers: BTreeSet<Nullifier>,
+    /// Nullifiers that were associated with undelegation transactions in this block, associated
+    /// with their unbonding epoch and validator identity key.
+    pub quarantined_nullifiers: BTreeMap<Nullifier, u64>,
```
hdevalence, line 26: What's the `u64` here, is that the unbonding epoch? If so, how are the nullifiers associated with the validator identity key?

pd/src/pending_block.rs
```diff
@@ -61,19 +72,37 @@ impl PendingBlock {
 
     /// Adds the state changes from a verified transaction.
     pub fn add_transaction(&mut self, transaction: VerifiedTransaction) {
+        let no_undelegations = transaction.undelegation_validators.is_empty();
```
hdevalence, line 75: Currently this is just used in processing the notes, but it should also be used in processing nullifiers below, something like

```rust
for nullifier in transaction.spent_nullifiers {
    self.spent_nullifiers.insert(nullifier);
    if undelegations {
        self.quarantined_nullifiers.do_stuff();
    }
}
```

Since the nullifier processing path is "do A unconditionally, if undelegations, do B", while the note processing path is "do A if no undelegations, do B if undelegations", I think it'd be clearer if this bool were changed to
```rust
        let undelegations = !transaction.undelegation_validators.is_empty();
```

pd/src/genesis.rs
hdevalence: Another option: perhaps we shouldn't specify a power at all, and instead compute the power from the genesis allocations, which can include delegation tokens?

pd/src/app.rs
plaidfinch: Makes sense to make it private to me -- done.

pd/src/genesis.rs
plaidfinch: yup, looks like an automerge caused it to get out of date, fixed now

pd/src/pending_block.rs
plaidfinch: I ended up using a slightly different architecture, which avoids duplicating data by grouping nullifiers and notes by transaction, obviating this (valid) suggeestion.

pd/src/pending_block.rs
plaidfinch: Replaced with `QuarantineGroup`, which stores all the important information.

pd/src/pending_block.rs
plaidfinch: done!

pd/src/state.rs
plaidfinch: It will be shortly :)

pd/src/state.rs
plaidfinch: This is now implemented.

pd/src/verify.rs
plaidfinch: Thanks, fixed.

pd/src/verify.rs
plaidfinch: agreed, fixed

proto/build.rs
plaidfinch: yes, it should be, but now it appears to be unused? deleting for now.

plaidfinch:
Closing in favor of #427.

PR #369, opened 2022-01-19
process changes to delegation and staking token supplies at epoch boundary

### LINKED ISSUE #357: Process (un)delegations at epoch boundaries.
With #356 in place, the last remaining part of #274 is to use the data it records to update the supply of delegation and staking tokens at the epoch boundary:

- for each validator, aggregate the changes to its token supply over the epoch, and update the supply accordingly;
- compute the change to the supply of the staking token, and update its supply.

avahowell:
Closes #357

### DIFF
--- a/pd/src/app.rs
+++ b/pd/src/app.rs
@@ -458,12 +458,9 @@ impl App {
                 // the delegations in pending_block with the ones already committed to the
                 // state. otherwise the delegations committed in the epoch threshold block
                 // would be lost.
-                let mut delegation_changes =
-                    state.delegation_changes(prev_epoch.index).await?;
+                let mut delegation_changes = state.delegation_changes(prev_epoch.index).await?;
                 for (id_key, delta) in &pending_block.lock().unwrap().delegation_changes {
-                    *delegation_changes
-                        .entry(id_key.clone())
-                        .or_insert(0) += delta;
+                    *delegation_changes.entry(id_key.clone()).or_insert(0) += delta;
                 }
 
                 for current_rate in &current_rates {
@@ -475,20 +472,10 @@ impl App {
 
                     // TODO: if a validator isn't part of the consensus set, should we ignore them
                     // and not update their rates?
-                    let delegation_delta = delegation_changes
-                        .get(&identity_key)
-                        .unwrap_or(&0i64);
+                    let delegation_delta = delegation_changes.get(&identity_key).unwrap_or(&0i64);
 
-                    // update staking token supply
-                    let unbonded_amount =
-                        current_rate.unbonded_amount(delegation_delta.abs() as u64);
-                    if *delegation_delta > 0 {
-                        // delegation: subtract the unbonded amount from the staking token supply
-                        staking_token_supply -= unbonded_amount;
-                    } else {
-                        // undelegation: add the unbonded amount to the staking token supply
-                        staking_token_supply += unbonded_amount;
-                    }
+                    let delegation_amount = delegation_delta.abs() as u64;
+                    let unbonded_amount = current_rate.unbonded_amount(delegation_amount);
 
                     let mut delegation_token_supply = state
                         .asset_lookup(identity_key.delegation_token().id().encode_to_vec())
@@ -496,11 +483,23 @@ impl App {
                         .map(|info| info.total_supply)
                         .unwrap_or(0);
 
-                    delegation_token_supply =
-                        (delegation_token_supply as i64 + delegation_delta) as u64;
+                    if *delegation_delta > 0 {
+                        // net delegation: subtract the unbonded amount from the staking token supply
+                        staking_token_supply =
+                            staking_token_supply.checked_sub(unbonded_amount).unwrap();
+                        delegation_token_supply = delegation_token_supply
+                            .checked_add(delegation_amount)
+                            .unwrap();
+                    } else {
+                        // net undelegation: add the unbonded amount to the staking token supply
+                        staking_token_supply =
+                            staking_token_supply.checked_add(unbonded_amount).unwrap();
+                        delegation_token_supply = delegation_token_supply
+                            .checked_sub(delegation_amount)
+                            .unwrap();
+                    }
 
                     // update the delegation token supply
-                    // TODO: should we use a method which panics on integer overflow here?
                     pending_block.lock().unwrap().supply_updates.insert(
                         identity_key.delegation_token().id(),
                         (
--- a/pd/src/app.rs
+++ b/pd/src/app.rs
@@ -460,14 +460,15 @@ impl App {
                     // TODO: if a validator isn't part of the consensus set, should we ignore them
                     // and not update their rates?
                     //
-                    //
                     // this is a bit complicated: because we're in the EndBlock phase, and the
                     // delegations in this block have not yet been committed, we have to combine
                     // the delegations in pending_block with the ones already committed to the
                     // state. otherwise the delegations committed in the epoch threshold block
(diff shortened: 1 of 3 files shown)


pd/src/app.rs
```diff
@@ -443,24 +452,63 @@ impl App {
...
-                    let delegation_token_supply = state
+                    // TODO: if a validator isn't part of the consensus set, should we ignore them
+                    // and not update their rates?
+                    let delegation_delta = committed_delegation_changes
+                        .get(&identity_key)
+                        .unwrap_or(&0i64);
+
+                    // update staking token supply
+                    let unbonded_amount =
+                        current_rate.unbonded_amount(delegation_delta.abs() as u64);
```
avahowell, line 485: is this correct? should we be updating delegation token and staking token supply based on current_rate, or should we be using next_rate?

pd/src/app.rs
```diff
@@ -423,13 +425,20 @@ impl App {
                 let current_base_rate = state.base_rate_data(current_epoch.index).await?;
                 let current_rates = state.rate_data(current_epoch.index).await?;
 
+                let mut staking_token_supply = state
+                    .asset_lookup(STAKING_TOKEN_ASSET_ID.encode_to_vec())
+                    .await?
+                    .map(|info| info.total_supply)
+                    .unwrap();
```
hdevalence, line 432: non-blocking: in a case like this where the code is going to unwrap the option in any case, I think it's slightly cleaner to do so earlier and then work with the inner value directly, rather than mapping on the inner value and then unwrapping.

pd/src/app.rs
```diff
@@ -443,24 +452,63 @@ impl App {
...
+                // the delegations in pending_block with the ones already committed to the
+                // state. otherwise the delegations committed in the epoch threshold block
+                // would be lost.
+                let mut committed_delegation_changes =
+                    state.delegation_changes(prev_epoch.index).await?;
+                for (id_key, delta) in &pending_block.lock().unwrap().delegation_changes {
+                    let _ = *committed_delegation_changes
+                        .entry(id_key.clone())
+                        .and_modify(|change| *change += delta)
+                        .or_insert(*delta);
```
hdevalence, line 467: I think this might be a bit clearer if it were written as

```suggestion
                    *committed_delegation_changes
                        .entry(id_key.clone())
                        .or_insert(0) += delta;
```

(assuming I'm reading the logic right) -- this way, there's only one code path to consider (adding delta), and it's more explicit that the default value is 0.

pd/src/app.rs
```diff
@@ -443,24 +452,63 @@ impl App {
 
                 let mut next_rates = Vec::new();
                 let mut next_validator_statuses = Vec::new();
+
+                // this is a bit complicated: because we're in the EndBlock phase, and the
+                // delegations in this block have not yet been committed, we have to combine
+                // the delegations in pending_block with the ones already committed to the
+                // state. otherwise the delegations committed in the epoch threshold block
+                // would be lost.
```
hdevalence, line 460: 👍

pd/src/app.rs
```diff
@@ -443,24 +452,63 @@ impl App {
...
                 for current_rate in &current_rates {
                     let identity_key = current_rate.identity_key.clone();
 
                     let funding_streams = state.funding_streams(identity_key.clone()).await?;
 
                     let next_rate = current_rate.next(&next_base_rate, funding_streams);
 
-                    let delegation_token_supply = state
+                    // TODO: if a validator isn't part of the consensus set, should we ignore them
+                    // and not update their rates?
```
hdevalence, line 478: Yeah, currently we don't have a notion of validator state, but when we do, we want to split into two paths:

- for validators in the consensus set, use the base rate data and funding streams to compute rate updates (the current code);
- for validators not in the consensus set, roll forward their existing rates, changing only the epoch index.

Maybe `RateData::next` should be split into `RateData::next_active` and `RateData::next_inactive`?

pd/src/app.rs
hdevalence: Yes, exactly, using current_rate here is correct, we crossed the epoch boundary at lines 409-413 above:
```rust
                // We've finished processing the last block of `epoch`, so we've
                // crossed the epoch boundary, and (prev | current | next) are:
                let prev_epoch = epoch;
                let current_epoch = prev_epoch.next();
                let next_epoch = current_epoch.next();
```
All the (un)delegations we're processing were prepared in the (now-previous) epoch using that epoch's `next_rate`, which is now `current_rate`.

pd/src/app.rs
```diff
@@ -443,24 +452,63 @@ impl App {
 
                 let mut next_rates = Vec::new();
                 let mut next_validator_statuses = Vec::new();
+
+                // this is a bit complicated: because we're in the EndBlock phase, and the
+                // delegations in this block have not yet been committed, we have to combine
+                // the delegations in pending_block with the ones already committed to the
+                // state. otherwise the delegations committed in the epoch threshold block
+                // would be lost.
+                let mut committed_delegation_changes =
```
hdevalence, line 461: Should this maybe be called `delegation_changes` rather than `committed_delegation_changes`, if we're adding in the delegation changes from the pending block that haven't been committed yet?

pd/src/app.rs
```diff
@@ -443,24 +452,63 @@ impl App {
...
-                    // TODO: we should process all of the delegations and undelegations in the
-                    // epoch here
-                    //
-                    // TODO: if a validator isn't part of the consensus set, should we ignore them
-                    // and not update their rates?
+                    delegation_token_supply =
+                        (delegation_token_supply as i64 + delegation_delta) as u64;
+
+                    // update the delegation token supply
+                    // TODO: should we use a method which panics on integer overflow here?
```
hdevalence, line 504: Is this comment supposed to attach to the line above?  I checked if there were any helpful methods, and there's going to be a `checked_add_signed` [in the future](https://doc.rust-lang.org/stable/std/primitive.u64.html#method.checked_add_signed) but it's not stable now.

Since there's already a branch on the sign of delta above, maybe it would be cleanest to merge the delegation token supply and staking token supply changes together, like
```rust
let delegation_amount = delegation_delta.abs() as u64;
let unbonded_amount = current_rate.unbonded_amount(delegation_amount);

let mut delegation_token_supply = /* ... */

if *delegation_delta > 0 {
    // net delegation: ...
    staking_token_supply = staking_token_supply.checked_sub(unbonded_amount).unwrap();
    delegation_token_supply = delegation_token_supply.checked_add(delegation_amount).unwrap();
} else {
    // net undelegation: ...
    staking_token_supply = staking_token_supply.checked_add(unbonded_amount).unwrap();
    delegation_token_supply = delegation_token_supply.checked_sub(delegation_amount).unwrap();
}
```

pd/src/state.rs
```diff
@@ -595,4 +595,23 @@ ON CONFLICT (id) DO UPDATE SET data = $1
             })
             .collect())
     }
+
+    /// Retrieve the delegation changes for the supplied epoch
+    /// TODO: should we have a DelegationChanges struct instead of just returning a BTreeMap?
```
hdevalence, line 600: Possibly -- do we need it to do anything special or maintain any invariants, or would a type alias work?

hdevalence (approved):
This looks good, I worked through the arithmetic and it seems right to me -- I left some suggestions about ways it could be streamlined, but none of them are blocking, happy to merge this now and keep iterating!

avahowell:
Updated the PR with some of the suggested changes

hdevalence (approved):
awesome!

PR #370, opened 2022-01-19
Implement Genesis data as protobufs

### LINKED ISSUE #327: Change genesis configuration to be proto-based
Currently we specify genesis configuration using data structures defined in the `pd::genesis` module.  It's important for these to have a well-defined encoding, so it would be better to change them to be proto-based:

- [x] create a `genesis.proto` file in the proto crate and fill it in;
- [x] add serialization annotations to the `build.rs`;
- [x] change the existing types in `pd::genesis` to be domain types for those proto types.

zbuc:
Fixes #327

### DIFF
--- a/pd/src/genesis.rs
+++ b/pd/src/genesis.rs
@@ -1,15 +1,46 @@
 use ark_ff::Zero;
 use decaf377::Fq;
 use penumbra_crypto::{asset, Address, Note, Value};
+use penumbra_proto::{genesis as pb, Protobuf};
 use penumbra_stake::Validator;
+
 use serde::{Deserialize, Serialize};
 
 /// A (transparent) genesis allocation.
 #[derive(Clone, Serialize, Deserialize)]
+#[serde(
+    try_from = "pb::genesis_app_state::Allocation",
+    into = "pb::genesis_app_state::Allocation"
+)]
 pub struct Allocation {
     pub amount: u64,
     pub denom: String,
     pub address: Address,
+}
+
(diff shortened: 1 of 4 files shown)


proto/proto/genesis.proto
```diff
@@ -0,0 +1,23 @@
...
+        uint64 amount = 1;
+        string denom = 2;
+        crypto.Address address = 3;
+    }
+
+    // This should maybe be moved to stake.proto
+    message ValidatorPower {
+        stake.Validator validator = 1;
+        uint64 power = 2;
+    }
```
hdevalence, line 18: xref https://github.com/penumbra-zone/penumbra/pull/362/files#r787160824

pd/src/genesis.rs
```diff
@@ -64,6 +127,50 @@ pub struct AppState {
...
+    fn from(a: AppState) -> Self {
+        pb::GenesisAppState {
+            epoch_duration: a.epoch_duration,
+            // validators: a.validators.iter().map(|v| v.into()).collect(),
+            validators: a
+                .validators
+                .into_iter()
+                .map(TryInto::try_into)
+                .collect::<Result<_, _>>()
+                .expect("able to serialize validators"),
```
hdevalence, line 140: Hmm, does it not work to do `.map(Into::into).collect()`? I think the domain types are supposed to be infallibly convertible into the proto types, does that not work here for some reason?

hdevalence (approved):
I think the `try_into` calls in the domain -> proto conversions should be able to be replaced by `into` calls, but otherwise this looks good to me!

PR #371, opened 2022-01-19
Use ChainParams in Genesis

### LINKED ISSUE #328: Use `ChainParams` in genesis data
Following #327 #326, the genesis configuration should be changed so that, instead of specifying the epoch directly, its proto includes a `ChainParams` proto (and its domain type includes the corresponding domain type).

zbuc:
Fixes #328

### DIFF
--- a/pcli/src/command/wallet.rs
+++ b/pcli/src/command/wallet.rs
@@ -123,6 +123,9 @@ impl WalletCmd {
             let spend_key_hash = Sha256::digest(&state.wallet().spend_key().seed().0);
             let wallet_archive_dir = archive_dir
                 .data_dir()
+                // TODO the chain ID should be synced from the server if
+                // `chain_params` is `None` (meaning a new wallet file),
+                // as it could have changed via consensus.
                 .join(CURRENT_CHAIN_ID)
                 .join(hex::encode(&spend_key_hash[0..8]));
             std::fs::create_dir_all(&wallet_archive_dir)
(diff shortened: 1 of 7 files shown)


wallet/src/state.rs
```diff
@@ -227,6 +227,15 @@ impl ClientState {
         }
     }
 
+    fn chain_id(&self) -> Result<String, anyhow::Error> {
+        let chain_params = self.chain_params();
+        if chain_params.is_none() {
+            return Err(anyhow::anyhow!("chain_params not set on state"));
+        }
+
+        Ok(chain_params.unwrap().chain_id.clone())
```
hdevalence, line 236: tioli: this could be simplified to `Ok(self.chain_params().ok_or_else(|| anyhow!(...))?.chain_id.clone())`, but it doesn't really matter here.

pcli/src/command/wallet.rs
```diff
@@ -123,6 +123,9 @@ impl WalletCmd {
             let spend_key_hash = Sha256::digest(&state.wallet().spend_key().seed().0);
             let wallet_archive_dir = archive_dir
                 .data_dir()
+                // TODO the chain ID should be synced from the server if
+                // `chain_params` is `None` (meaning a new wallet file),
+                // as it could have changed via consensus.
                 .join(CURRENT_CHAIN_ID)
```
hdevalence, line 129: Maybe we should have an issue to get rid of the `CURRENT_CHAIN_ID` constants?

hdevalence (approved):
LGTM! I rebased this onto `main`.

PR #380, opened 2022-01-20
implement validator commission rewards

### LINKED ISSUE #42: Implement rewards for funding streams
Following #41 and #40, the application should use the formulas in the staking description to implement payouts for validator's funding streams.

Along the lines of https://github.com/penumbra-zone/penumbra/issues/39#issuecomment-987369226 , this should be done by creating additional notes while processing the epoch transition and adding them into the pending block.  These notes should be controlled by the addresses specified in the funding streams.

avahowell:
closes #42 

this PR adds commission rewards processing to the EndBlock phase of consensus, by inserting notes for each validator's funding stream. the reward is calculated using the formula from the spec, that is 

`yv*cv,e*re*ψ(e−1)`

where yv is the total number of delegation tokens, cv,e is the rate for the funding stream, re is the base reward rate for the epoch, and psi(e-1) is the base exchange rate for the previous epoch. 

these notes are constructed using a zero ephemeral key and are assigned a 0 transaction ID in their `NoteData`.

### DIFF
--- a/pd/src/app.rs
+++ b/pd/src/app.rs
@@ -12,7 +12,7 @@ use futures::future::FutureExt;
 use penumbra_crypto::{
     asset,
     merkle::{self, NoteCommitmentTree, TreeExt},
-    note, Nullifier,
+    note, Nullifier, Value,
 };
 use penumbra_stake::{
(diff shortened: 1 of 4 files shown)


hdevalence (approved):
LGTM

PR #384, opened 2022-01-20
Allow clippy::clone_on_copy, fix various clippy lint warnings

zbuc:


crypto/src/asset/denom.rs
```diff
@@ -259,10 +259,11 @@ impl Unit {
...
                 .checked_mul(v1_power_of_ten)
                 .and_then(|x| x.checked_add(v2));
-            if v.is_none() {
-                return Err(anyhow::anyhow!("overflow!"));
+
+            if let Some(value) = v {
+                Ok(value)
             } else {
-                return Ok(v.unwrap());
+                Err(anyhow::anyhow!("overflow!"))
```
plaidfinch, line 266: can we make this more specific, like "integer overflow in `Unit::parse_value`"?

plaidfinch (approved):
LGTM, just a single non-blocking comment about an error message that could be improved.

hdevalence (approved):

PR #386, opened 2022-01-20
scripts: make docker_compose_freshstart work without DATABASE_URL

hdevalence:


zbuc:
I think this is ok for local development but it wouldn't work on a production testnet. Probably fine.

zbuc (approved):

hdevalence:
> I think this is ok for local development but it wouldn't work on a production testnet. Probably fine, the script's purpose is for wiping all state from an existing testnet deployment prior to redeploying anyways, which is more of a local dev process than production process.

Hmm, maybe we should change the script to set the env variable only if it's missing? It seems useful to also use this for setting up all the docker volumes etc. for a production deployment.

zbuc:
That sounds good to me

hdevalence:
Ran into this this morning with @aubrika, so I'm going to merge this now and then we can change it to set only if missing in a later PR.

PR #388, opened 2022-01-20
Use `u64` to represent block heights in light wallet `CompactBlock`

### LINKED ISSUE #365: Change `CompactBlock` height to be a `u64`
Everywhere else, we use a `u64` to represent block heights, so the different type in the `CompactBlock` causes a bunch of annoying conversions.  The original motivation was to have something slightly smaller on the wire, but protos already use a variable-length encoding, so there's no real savings to be had.

plaidfinch:
Fixes #365.

### DIFF
--- a/pd/src/state/reader.rs
+++ b/pd/src/state/reader.rs
@@ -425,7 +425,7 @@ impl Reader {
 
             for height in start_height..=end_height {
                 let mut compact_block = CompactBlock {
-                    height: height as u32,
+                    height: height as u64,
                     fragments: vec![],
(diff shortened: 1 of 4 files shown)


hdevalence:
This looks fine to me, but I think we should wait for the end of the week, and slip it into the next testnet, since it's a breaking change to the protos (editing fields rather than just adding them).

redshiftzero:
this looks good and seems like a good candidate to rebase and merge in advance of the next testnet?

hdevalence:
Rebased and resolved conflicts

Issue #375, opened 2022-01-19
Model validator states and consensus participation


hdevalence:
Our current staking implementation gets the basic economics mostly right, but doesn't actually model whether or not validators are part of the consensus set, or what the validator life cycle is.

We should map out the state transitions validators can make, and work out the interactions with the staking system.  Here is one sketch:
```ascii                     
                                       ┌ ─ ─ ─ ─ ┐             
                                         Genesis               
                                       │Validator│             
                                        ─ ─ ─ ─ ─              
                                            │                  
                                            │                  
┌ ─ ─ ─ ─ ─ ─ ─ ─                           ▼                  
    Validator    │      ┏━━━━━━━━┓      ╔══════╗      ┏━━━━━━━┓
│   Definition    ─────▶┃Inactive┃─────▶║Active║─────▶┃Slashed┃
 (in transaction)│      ┗━━━━━━━━┛      ╚══════╝      ┗━━━━━━━┛
└ ─ ─ ─ ─ ─ ─ ─ ─            ▲              ▲             ▲    
                             │              │             │    
                             │              ▼             │    
                             │         ╔═════════╗        │    
                             └─────────║Unbonding║────────┘    
                                       ╚═════════╝             
```
Single lines represent unbonded stake, and double lines represent bonded stake.

Validators become known to the chain either at genesis, or by means of a transaction with a `ValidatorDefinition` action in them. Validators transition through four states:

* **Inactive**, where the validator is not part of the consensus set, the stake in the validator's delegation pool is not bonded;
* **Active**, where the validator is part of the consensus set, and the stake in the validator's delegation pool is bonded;
* **Unbonding**, where the validator is not part of the consensus set, but the stake in the validator's delegation pool is still bonded;
* **Slashed**, where the validator is not part of the consensus set, and the stake in the validator's delegation pool is not bonded.

Validators specified in the genesis config begin in the active state, with whatever stake was allocated to their delegation pool at genesis. Otherwise, validators begin in the inactive state, with no stake in their delegation pool.  At this point, the validator is known to the chain, and stake can be contributed to its delegation pool.  Stake contributed to an inactive validator's delegation pool does not earn rewards (the validator's rates are held constant), but it is also not bonded, so undelegations are effective immediately, with no unbonding period and no output quarantine.

[...]

PR #389, opened 2022-01-21
Add validator_state/unbonding_epoch to the validators table and domain types

zbuc:
Remaining work around validator state transitions is to be completed in #375 but this helps unblock @plaidfinch

### DIFF
--- a/pd/src/main.rs
+++ b/pd/src/main.rs
@@ -12,7 +12,7 @@ use penumbra_proto::{
     light_wallet::light_wallet_server::LightWalletServer,
     thin_wallet::thin_wallet_server::ThinWalletServer,
 };
-use penumbra_stake::{FundingStream, Validator};
+use penumbra_stake::{FundingStream, FundingStreams, Validator};
 use rand_core::OsRng;
 use structopt::StructOpt;
 use tonic::transport::Server;
@@ -306,16 +306,22 @@ async fn main() -> anyhow::Result<()> {
                                     name: v.name.clone(),
                                     website: v.website.clone(),
                                     description: v.description.clone(),
-                                    funding_streams: v
-                                        .funding_streams
-                                        .iter()
-                                        .map(|fs| FundingStream {
+                                    funding_streams: FundingStreams::try_from(
+                                        v.funding_streams
+                                            .iter()
+                                            .map(|fs| {
+                                                FundingStream {
                                             address: Address::from_str(&fs.address).expect(
                                                 "invalid funding stream address in validators.json",
                                             ),
                                             rate_bps: fs.rate_bps,
-                                        })
-                                        .collect(),
+                                        }
+                                            })
+                                            .collect::<Vec<FundingStream>>(),
+                                    )
+                                    .expect(
+                                        "unable to construct funding streams from validators.json",
+                                    ),
                                     sequence_number: v.sequence_number,
                                 },
                                 power: v.voting_power.into(),
--- a/pd/src/main.rs
+++ b/pd/src/main.rs
@@ -296,7 +296,7 @@ async fn main() -> anyhow::Result<()> {
                     validators: validators
                         .iter()
                         .map(|v| {
-                            ValidatorPower {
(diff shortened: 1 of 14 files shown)


pd/src/main.rs
```diff
@@ -306,16 +306,22 @@ async fn main() -> anyhow::Result<()> {
...
-                                    funding_streams: v
-                                        .funding_streams
-                                        .iter()
-                                        .map(|fs| FundingStream {
+                                    funding_streams: FundingStreams::try_from(
+                                        v.funding_streams
+                                            .iter()
+                                            .map(|fs| {
+                                                FundingStream {
                                             address: Address::from_str(&fs.address).expect(
```
plaidfinch, line 314: It would be better to throw an error here rather than panicking, if possible. Can we replace with `anyhow::anyhow!(...)?`?

pd/src/main.rs
```diff
@@ -306,16 +306,22 @@ async fn main() -> anyhow::Result<()> {
...
                                                 "invalid funding stream address in validators.json",
                                             ),
                                             rate_bps: fs.rate_bps,
-                                        })
-                                        .collect(),
+                                        }
+                                            })
+                                            .collect::<Vec<FundingStream>>(),
+                                    )
+                                    .expect(
```
plaidfinch, line 322: Likewise here, it would be better (in general) to use `anyhow` to throw errors rather than `expect`.

pd/src/state.rs
```diff
@@ -96,19 +98,29 @@ INSERT INTO blobs (id, data) VALUES ('gc', $1)
...
+                    validator_state,
+                    unbonding_epoch
+                ) VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9)",
                 validator.identity_key.encode_to_vec(),
                 validator.consensus_key.to_bytes(),
                 validator.sequence_number as i64,
-                validator.encode_to_vec(),
+                validator.name.clone(),
+                validator.website.clone(),
+                validator.description.clone(),
```
plaidfinch, line 113: You don't need to clone any of these, you can refer to them by reference and it will work just fine.

pd/src/state.rs
```diff
@@ -403,6 +415,8 @@ ON CONFLICT (id) DO UPDATE SET data = $1
 
     pub async fn rate_data(&self, epoch_index: u64) -> Result<Vec<RateData>> {
         let mut conn = self.pool.acquire().await?;
+        // TODO: This query needs to be updated to select the *most recent* rate data
+        // to the given epoch
```
plaidfinch, line 419: Can we make an issue for this TODO?

pd/src/state.rs
```diff
@@ -480,11 +500,27 @@ ON CONFLICT (id) DO UPDATE SET data = $1
...
+                    tendermint::PublicKey::from_raw_ed25519(row.consensus_key.as_slice())
+                        .ok_or_else(|| anyhow::anyhow!("invalid ed25519 consensus pubkey"))?;
                 Ok(ValidatorInfo {
-                    validator: Validator::decode(row.validator_data.as_slice())?,
+                    validator: Validator {
+                        identity_key: identity_key.clone(),
+                        consensus_key: consensus_key.clone(),
+                        name: row.name.clone(),
+                        website: row.website.clone(),
+                        description: row.description.clone(),
```
plaidfinch, line 512: No need to clone `name`, `website`, `description`: they can be passed by reference here.

plaidfinch (approved):
This looks great :) Just a few non-blocking nits about errors vs. panics, and some clones that I think are unnecessary.

pd/src/main.rs
zbuc: I think this is OK because an error here would be unrecoverable (this is in the testnet generation command). I'm also unsure how to bubble up the `Err` response to the scope outside of the `map`.

pd/src/state.rs
zbuc: https://github.com/penumbra-zone/penumbra/issues/402

PR #391, opened 2022-01-22
Add jmt key enum

aubrika:
Adding enum to be able to use `jellyfish::Key::T.hash()` for generating unique `HashValue`s for relevant types, utilizing the `Hasher` generated by `jmt`'s 'define_hasher!` macro.

Depends on changes to `jmt`: https://github.com/penumbra-zone/jmt/pull/9

### DIFF
--- a/pd/src/state/jellyfish.rs
+++ b/pd/src/state/jellyfish.rs
@@ -1,6 +1,7 @@
 use anyhow::Result;
 use futures::future::BoxFuture;
 use jmt::{
+    hash::HashValue,
     node_type::{LeafNode, Node, NodeKey},
     NodeBatch, TreeReaderAsync, TreeWriterAsync, Value,
 };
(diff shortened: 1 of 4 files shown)


pd/src/state/jellyfish.rs
```diff
@@ -1,14 +1,42 @@
...
+pub enum Key {
+    NoteCommitmentAnchor,
+}
+
+impl Key {
+    pub fn hash(self) -> HashValue {
+        match self {
+            Key::NoteCommitmentAnchor => {
+                let mut state = NoteCommitmentAnchorHasher::default();
+                state.update(b"");
```
plaidfinch, line 24: why are you doing this update with the empty bytestring here?

aubrika: this is unclear from the function itself, but that string literal serves as a salt for the hash function, which I don't think we need here since each `Key` type will have a corresponding hasher type - we may want to change this as we add more to the `Key` enum though

ex.:

`define_hasher! {
    /// The hasher used to compute the hash of an internal node in the Sparse Merkle Tree.
    (
        SparseMerkleInternalHasher,
        SPARSE_MERKLE_INTERNAL_HASHER,
        SPARSE_MERKLE_INTERNAL_SEED,
        b"SparseMerkleInternal"
    )
}

define_hasher! {
    /// The hasher used to compute the hash of an internal node in the transaction accumulator.
    (
        VoteProposalHasher,
        VOTE_PROPOSAL_HASHER,
        VOTE_PROPOSAL_SEED,
        b"VoteProposalHasher"
    )
}

define_hasher! {
    /// The hasher used only for testing. It doesn't have a salt.
    (TestOnlyHasher, TEST_ONLY_HASHER, TEST_ONLY_SEED, b"")
}
`

PR #399, opened 2022-01-24
WIP: Chain id checking

### LINKED ISSUE #394: Confirm chain id in `pcli` requests
Currently, it's very possible to accidentally connect to the wrong chain with `pcli`, resulting in confusion (where are my assets? why has time gone backwards?). This should instead result in an error.

I can see 3 possible ways this could be implemented, depending on where the check occurs:

1. `pcli` transmits the id of the chain it expects to be connecting to in each request, and `pd` returns an error if that does not match the chain id of the chain `pd` is managing.
2. `pd` returns, along with any other result data, its chain id, and `pcli` checks this and returns an error instead of processing the rest of the result if it does not match the expected chain id.
3. `pcli` precedes each request other than `ChainParams` itself with a `ChainParams` request, and checks to make sure the chain id matches before proceeding.

**I think that (3) is the best solution, because:**

- unlike (1) it means that `pcli` doesn't ever disclose to `pd` what chain id it is expecting to see (which could leak local development information to a public server in the case of misconfiguration), and
- unlike (1) and (2) it means that this can be a transparent upgrade, because it does not change the proto types.

Aside from the changes mentioned in (3), implementing this will require a change to the wallet format to store the expected chain id. Considerations here:

- This field should be optional: when it is not yet populated (as in a wallet that pre-existed prior to this update), it should be filled in the first time a chain is contacted.
- When the wallet is initialized, it should also be left blank, and then only when the client connects to the chain should it be filled in, at this point effectively "locking" the client to that chain.
- `pcli wallet reset` should remove this field, so that keys can be reused across multiple chains by resetting all the rest of the state.

plaidfinch:
Fixes #394 (not yet finished, need to actually check on `pd` side).

### DIFF
--- a/pcli/src/command/stake.rs
+++ b/pcli/src/command/stake.rs
@@ -106,6 +106,9 @@ impl StakeCmd {
                     .validator_rate(tonic::Request::new(ValidatorRateRequest {
                         identity_key: Some(to.into()),
                         epoch_index: next_epoch.index,
+                        chain_id: state
+                            .chain_id()
+                            .ok_or_else(|| anyhow!("missing chain_id"))?,
                     }))
                     .await?
                     .into_inner()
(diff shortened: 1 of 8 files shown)


hdevalence:
Rebased following the `pd` refactor that makes this data available

PR #406, opened 2022-01-27
wallet: remove authentication paths for spent notes

hdevalence:
Since they're spent, we don't need to ever produce an auth path for them, so we
can prune them to reduce the size of the NCT.


plaidfinch (approved):
This looks good to me!

Issue #402, opened 2022-01-25
rate_data needs to return **most recent** rate data per given epoch


zbuc:
Currently the `pub async fn rate_data(&self, epoch_index: u64) -> Result<Vec<RateData>>` function will retrieve rate data **only** for the given epoch. If no rate data for a validator exists in the given epoch, it won't be returned.

This needs to be updated to return the **most recent** rate data to the given epoch, so that currently inactive validators are represented in the response.

This can likely be accomplished by modifying the existing query to join two subqueries: one for the most recent rate data and one for the specified epoch.

hdevalence:
Closed as stale

PR #415, opened 2022-02-02
Update rate data query to select most recent rate data

zbuc:
Fixes the issue described in #402 where inactive validators without rate data for the given epoch don't return their last rate data.

### DIFF
--- a/pd/src/consensus/worker.rs
+++ b/pd/src/consensus/worker.rs
@@ -313,6 +313,7 @@ impl Worker {
                 let identity_key = current_rate.identity_key.clone();
 
                 let funding_streams = reader.funding_streams(identity_key.clone()).await?;
+                // TODO: if this rate is for an inactive validator, do we need to do anything special here regarding the next rate value?
                 let next_rate = current_rate.next(&next_base_rate, funding_streams.as_ref());
 
                 // TODO: if a validator isn't part of the consensus set, should we ignore them
--- a/pd/sqlx-data.json
+++ b/pd/sqlx-data.json
@@ -730,8 +730,8 @@
       "nullable": []
     }
   },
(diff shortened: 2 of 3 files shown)


hdevalence:
does this need another `sqlx prepare`?

pd/src/consensus/worker.rs
```diff
@@ -313,6 +313,7 @@ impl Worker {
                 let identity_key = current_rate.identity_key.clone();
 
                 let funding_streams = reader.funding_streams(identity_key.clone()).await?;
+                // TODO: if this rate is for an inactive validator, do we need to do anything special here regarding the next rate value?
```
zbuc, line 316: @hdevalence can you weigh in?

zbuc: Looking at the code, it looks like inactive validators will incorrectly continue having their exchange rates increase with this change. I think we need to check the status here and decide whether to update the rate or not.

hdevalence: Yeah, we want the rate updates to only be done for active validators, for exactly that reason -- but I didn't think we'd implemented validator states yet?

zbuc: @hdevalence correct the state transition PR is still in progress. The state representation is available as a struct property though.

What about validators that went inactive during the block being ended? Their state hasn't yet been committed to the DB (only happens in commit_block) -- should we use the "next validator state" (which is a property of the pending block) or the "current committed validator state" (from the database)?

hdevalence: Hmm, maybe we should merge this code as-is, and then in a separate chunk, refactor it to be iterating over the validators rather than iterating over the rates?

PR #416, opened 2022-02-02
Update mneme allocations

hdevalence:
In addition to the CSV file, this also adds a single file with discord addresses that we can update as we go.


plaidfinch:
Do we have written documentation of the process by which the list of discord addresses gets translated to the funding streams JSON and initial allocations? Might be a good opportunity to write that down if not.

hdevalence:
There's not much to write down, I just read through the backscroll and spend a few minutes copy-pasting from the most recently recorded addresses, then pasted the new ones into the allocations CSV.

PR #420, opened 2022-02-04
fix: send transaction in POST request body

### LINKED ISSUE #404: Internal panic in `reqwest` when transaction body is large
There is an [open bug in `reqwest` (#668)](https://github.com/seanmonstar/reqwest/issues/668): it is an internal expectation that any valid `Url` is also a valid `Uri`. Per the current logic for parsing `Uri`s and `Url`s, this is not always the case. The corner case that affects us is that when the length of the `Url` exceeds a certain threshold (unclear of the exact limit at time of writing), an internal `.expect()` fails in `request`, resulting in:

```
thread 'main' panicked at 'a parsed Url should always be a valid Uri: InvalidUri(TooLong)', /Users/$USER/.cargo/registry/src/github.com-1ecc6299db9ec823/reqwest-0.11.9/src/into_url.rs:70:14
```

We encounter this issue whenever we try to submit a transaction to the Tendermint RPC API which is sufficiently long to trigger this condition when included as hex-encoding in the URI, because we are using the URI-over-HTTP RPC API from Tendermint:

https://github.com/penumbra-zone/penumbra/blob/1996412d5ea10e4eb03a1170ae909bdb2bb21820/pcli/src/network.rs#L18-L26

When the hex-encoding of the transaction is too long, the direct inclusion of that encoding in the URI of the GET request causes `reqwest` to panic.

This issue points to two things that need to be solved:

1. Transactions are limited in size by the arbitrary bound of the size of a URI, and
2. Transactions are often very very large.

This issue addresses (1); a follow-up issue will address (2).

## Proposed solution to (1)

We should use the JSONRPC-over-HTTP endpoint for Tendermint instead of the URI-over-HTTP endpoint. This encodes the transaction body in the body of a POST, which can be arbitrarily large. See the documentation for the Tendermint RPC: https://docs.tendermint.com/master/rpc/.

redshiftzero:
Closes #404

### DIFF
--- a/README.md
+++ b/README.md
@@ -253,15 +253,15 @@ cargo run --bin pcli -- -w testnet_wallet.json addr new "Test Address 2"
 
 Next, produce a template with
 ```
-cargo run --bin pd -- create-genesis-template
+cargo run --bin pd -- generate-testnet
 ```
-and copy the output into the `genesis.json` as the `app_state` field.
+and copy the `app_state` field of one the genesis files. You'll need to
 edit it to match the key material you'll be using, which includes:
 
 * changing the validator public keys to match the one Tendermint generated;
 * editing the genesis allocations to use your testing addresses, or have other asset types, etc.
 
-You may wish to edit other parts of the testnet config.  Example `genesis.json`
+You may wish to edit other parts of the testnet config. Example `genesis.json`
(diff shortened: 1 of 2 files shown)


hdevalence (approved):
LGTM!

PR #423, opened 2022-02-05
pcli: add sweep implementation

hdevalence:
This code is done outside of the client state as a test case for whether it's possible to use that interface to implement bespoke note handling -- some changes were required, because we'd accidentally closed off some required methods.

For each (account, denom) pair, we do a parallel, SWEEP_COUNT-to-1 sweep of as many of that denom's notes as possible, sending them to the same account.  This command can be run multiple times; if there's fewer than SWEEP_COUNT notes remaining, it's a no-op.  The original implementation counted an estimate of the number of notes following the sweep and used it to decide whether to recurse, but doing so requires async recursion and extra bookkeeping, and this is simpler and works now.


wallet/src/state.rs
```diff
@@ -154,6 +169,23 @@ impl ClientState {
...
+    /// Register a note as spent.
+    ///
+    /// This marks the note as having been spent (pending confirmation) by the
+    /// chain.  Tracking these notes allows the wallet to not accidentally
+    /// attempt to double-spend a note, just because the first transaction that
+    /// spent it hasn't been finalized yet.
+    ///
+    /// This registration is temporary; if the spend is not observed on-chain
+    /// before some timeout, it will be forgotten, and the note marked as unspent again.
+    pub fn register_spend(&mut self, note: &Note) {
```
plaidfinch, line 181: This is a nice consolidation!

plaidfinch (approved):
This looks good, and works fine for me when tested!

Non-blocking: I wonder whether it's worth accelerating the parallel construction of the transactions by actually parallelizing this.

plaidfinch:
Perhaps a nice touch would be to exit with a non-zero exit code when `sweep` is complete, so that it's easy to write a shell script that loops this command until finished?

PR #424, opened 2022-02-06
corrected deadlinks in spec

ejmg:
Walking through the spec, I found a few deadlinks and corrected them. From what I could tell, they were all relative links that became outdated, so my fixes will fall victim to the same problem later if there is another shuffle of the protocol's documentation outline.


zbuc:
Thanks for fixing those!

Issue #211, opened 2021-12-01
Add additional metrics


hdevalence:
Following #12, which just set up the basic metrics infra, it would be useful to add a few extra metrics.  Some off the top of my head:

- counter of compact blocks sent out to wallet clients (allows creating rates in grafana)
- ~counter of notes created~ in #432
- ~counter of nullifiers revealed~ in  #426
- counter of transactions
- gauge of block heights

PR #426, opened 2022-02-08
feat(metrics): track counter of spent nullifiers

redshiftzero:
Implements one of the bullets in #211

### DIFF
--- a/pd/src/state/reader.rs
+++ b/pd/src/state/reader.rs
@@ -112,6 +112,19 @@ impl Reader {
         Ok(note_commitment_tree)
     }
 
+    /// Returns the number of revealed nullifiers (for updating metrics dashboard).
+    pub async fn nullifier_count(&self) -> Result<u64> {
+        let mut conn = self.pool.acquire().await?;
+
+        // Special column name "count!" is used to return i64 instead of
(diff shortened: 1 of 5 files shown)


pd/src/state/reader.rs
```diff
@@ -112,6 +112,19 @@ impl Reader {
...
+
+        // Special column name "count!" is used to return i64 instead of
+        // Option<i64> due to https://github.com/launchbadge/sqlx/issues/864
+        let nullifiers = query!(r#"SELECT COUNT(*) as "count!" FROM nullifiers"#)
+            .fetch_one(&mut conn)
+            .await?;
+
+        Ok(nullifiers.count.try_into().unwrap())
+    }
+
```
hdevalence, line 127: Since we'd never want the nullifier count for any reason other than metrics/statistics, I think it might be better if the method were a "fetch chain stats / metrics / etc" that returned a bundle of data, so that as we want to add more metrics, we can add fields to the returned data and still only have to do one query.

hdevalence (approved):
LGTM, in the future I think it'd be cleaner to rework the new method into a general metrics-retriever one, but that's non-blocking on this PR

redshiftzero:
yep that's a good point, I'll do that in a followup PR adding another metric

PR #427, opened 2022-02-08
Unbonding (new architecture)

plaidfinch:
Implements unbonding period and reversion of quarantined outputs in the event of slashing (pending slashing being implemented separately to this PR).

### DIFF
--- a/chain/src/params.rs
+++ b/chain/src/params.rs
@@ -41,6 +41,7 @@ impl From<AssetInfo> for pb::AssetInfo {
 pub struct ChainParams {
     pub chain_id: String,
     pub epoch_duration: u64,
+    pub unbonding_epochs: u64,
 }
 
 impl Protobuf<pb::ChainParams> for ChainParams {}
@@ -50,6 +51,7 @@ impl From<pb::ChainParams> for ChainParams {
         ChainParams {
             chain_id: msg.chain_id,
             epoch_duration: msg.epoch_duration,
+            unbonding_epochs: msg.unbonding_epochs,
         }
     }
 }
@@ -59,6 +61,7 @@ impl From<ChainParams> for pb::ChainParams {
         pb::ChainParams {
             chain_id: params.chain_id,
             epoch_duration: params.epoch_duration,
+            unbonding_epochs: params.unbonding_epochs,
         }
     }
 }
@@ -68,6 +71,7 @@ impl Default for ChainParams {
         Self {
             chain_id: String::new(),
             epoch_duration: 8640,
+            unbonding_epochs: 30,
         }
     }
 }
--- a/crypto/src/nullifier.rs
+++ b/crypto/src/nullifier.rs
@@ -7,7 +7,7 @@ use once_cell::sync::Lazy;
 
 use crate::Fq;
 
-#[derive(PartialEq, Eq, Derivative, Clone, Hash, PartialOrd, Ord)]
+#[derive(PartialEq, Eq, Derivative, Clone, Copy, Hash, PartialOrd, Ord)]
 #[derivative(Debug)]
 pub struct Nullifier(#[derivative(Debug(format_with = "crate::fmt_fq"))] pub Fq);
 
--- a/crypto/src/value.rs
+++ b/crypto/src/value.rs
@@ -45,14 +45,8 @@ impl TryFrom<pb::Value> for Value {
     }
 }
 
-#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Default)]
 pub struct Commitment(pub decaf377::Element);
-
-impl Default for Commitment {
-    fn default() -> Self {
-        Self(decaf377::Element::default())
-    }
-}
 
 pub static VALUE_BLINDING_GENERATOR: Lazy<decaf377::Element> = Lazy::new(|| {
     let s = Fq::from_le_bytes_mod_order(blake2b_simd::blake2b(b"decaf377-rdsa-binding").as_bytes());
--- a/pcli/src/command/addr.rs
+++ b/pcli/src/command/addr.rs
@@ -50,7 +50,7 @@ impl AddrCmd {
                 let (label, address) = state.wallet().address_by_index(*index as usize)?;
 
                 if *addr_only {
-                    println!("{}", address.to_string());
+                    println!("{}", address);
                     return Ok(()); // don't print the label
                 } else {
                     table.add_row(vec![index.to_string(), label, address.to_string()]);
--- a/pd/migrations/20211216025603_init.sql
+++ b/pd/migrations/20211216025603_init.sql
@@ -1,42 +1,55 @@
+-- Miscellaneous key-value storage for the application
 CREATE TABLE IF NOT EXISTS blobs (
     id varchar(64) PRIMARY KEY,
     data bytea NOT NULL
 );
 
+-- Backing storage for the public Jellyfish Merkle Tree
 CREATE TABLE IF NOT EXISTS jmt (
     key bytea PRIMARY KEY,
     value bytea NOT NULL
 );
 
+-- All known assets
 CREATE TABLE IF NOT EXISTS assets (
     asset_id bytea PRIMARY KEY NOT NULL,
     denom varchar NOT NULL,
-    total_supply bigint NOT NULL
+    total_supply bigint NOT NULL,
+    -- total supply can't be negative
+    CONSTRAINT positive_total_supply CHECK (total_supply >= 0)
 );
 
+-- Blocks, indexed by height
 CREATE TABLE IF NOT EXISTS blocks (
     height bigint PRIMARY KEY,
     nct_anchor bytea NOT NULL,
-    app_hash bytea NOT NULL
+    app_hash bytea NOT NULL,
+    -- height can't be negative
+    CONSTRAINT positive_height CHECK (height >= 0)
 );
 
+-- Nullifiers, indexed by height
 CREATE TABLE IF NOT EXISTS nullifiers (
     nullifier bytea PRIMARY KEY,
     height bigint NOT NULL REFERENCES blocks (height)
 );
 CREATE INDEX ON nullifiers (height);
 
+-- Notes that appear in each block, indexed by height and position
 CREATE TABLE IF NOT EXISTS notes (
     note_commitment bytea PRIMARY KEY,
     ephemeral_key bytea NOT NULL,
     encrypted_note bytea NOT NULL,
     transaction_id bytea NOT NULL,
     position bigint NOT NULL,
-    height bigint NOT NULL REFERENCES blocks (height)
+    height bigint NOT NULL REFERENCES blocks (height),
+    -- position can't be negative
+    CONSTRAINT positive_position CHECK (position >= 0)
 );
 CREATE INDEX ON notes (position);
 CREATE INDEX ON notes (height);
 
+-- All validators who have ever been declared
 CREATE TABLE IF NOT EXISTS validators (
     identity_key bytea NOT NULL PRIMARY KEY,
     consensus_key bytea NOT NULL,
@@ -45,38 +58,57 @@ CREATE TABLE IF NOT EXISTS validators (
     website varchar NOT NULL,
     description varchar NOT NULL,
     voting_power bigint NOT NULL,
-    -- TODO: validator_state should be limited to valid state strings
     validator_state varchar NOT NULL,
-    -- TODO: unbonding epoch can only be set when validator_state is unbonding
-    unbonding_epoch bigint
+    unbonding_epoch bigint,
+    -- sequence_number can't be negative
+    CONSTRAINT positive_sequence_number CHECK (sequence_number >= 0),
+    -- voting power can't be negative
+    CONSTRAINT positive_voting_power CHECK (voting_power >= 0),
+    -- validator state can only be one of the valid strings
+    CONSTRAINT valid_state_name
+        CHECK (validator_state IN ('INACTIVE', 'ACTIVE', 'UNBONDING', 'SLASHED')),
+    -- the unbonding epoch is not null precisely when the validator is unbonding
+    CONSTRAINT unbonding_epoch_exactly_when_unbonding
+        CHECK ((validator_state =  'UNBONDING' AND unbonding_epoch IS NOT NULL) OR
+               (validator_state != 'UNBONDING' AND unbonding_epoch IS NULL))
 );
 
--- TODO: the sum of funding streams for any validator should never exceed 10,000.
--- this is enforced in code, but could also be enforced here.
+-- The funding streams for all validators who have ever been declared
 CREATE TABLE IF NOT EXISTS validator_fundingstreams (
     identity_key bytea NOT NULL REFERENCES validators (identity_key),
     address varchar NOT NULL,
-    rate_bps bigint NOT NULL
+    rate_bps bigint NOT NULL,
+    -- rate_bps must range between [0, 10000] inclusive for it to represent a percentage rate
+    CONSTRAINT valid_bps CHECK (rate_bps >= 0 AND rate_bps <= 10000)
 );
 
+-- The base reward rate for each epoch
 CREATE TABLE IF NOT EXISTS base_rates (
     epoch bigint PRIMARY KEY,
     base_reward_rate bigint NOT NULL,
-    base_exchange_rate bigint NOT NULL
+    base_exchange_rate bigint NOT NULL,
+    -- epoch can't be negative
+    CONSTRAINT positive_epoch CHECK (epoch >= 0)
 );
 
+-- The validator rates for each epoch (some may be skipped if the validator was not active)
 CREATE TABLE IF NOT EXISTS validator_rates (
     identity_key bytea NOT NULL REFERENCES validators (identity_key),
     epoch bigint NOT NULL,
     validator_reward_rate bigint NOT NULL,
     validator_exchange_rate bigint NOT NULL,
-    PRIMARY KEY(epoch, identity_key)
+    PRIMARY KEY(epoch, identity_key),
+    -- epoch can't be negative'
+    CONSTRAINT positive_epoch CHECK (epoch >= 0)
 );
 
+-- Changes to delegations that occurred in each block
 CREATE TABLE IF NOT EXISTS delegation_changes (
     validator_identity_key bytea NOT NULL REFERENCES validators (identity_key),
     epoch bigint NOT NULL,
-    delegation_change bigint NOT NULL
+    delegation_change bigint NOT NULL,
+    -- epoch can't be negative
+    CONSTRAINT positive_epoch CHECK (epoch >= 0)
 );
 CREATE INDEX ON delegation_changes (epoch);
 CREATE INDEX ON delegation_changes (validator_identity_key);
--- a/pd/migrations/20211216025603_init.sql
+++ b/pd/migrations/20211216025603_init.sql
@@ -120,10 +120,12 @@ CREATE TABLE IF NOT EXISTS quarantined_notes (
     ephemeral_key bytea NOT NULL,
     encrypted_note bytea NOT NULL,
     transaction_id bytea NOT NULL,
-    height bigint NOT NULL REFERENCES blocks (height),
-    validator_identity_key bytea NOT NULL REFERENCES validators (identity_key)
+    unbonding_height bigint NOT NULL, -- height at which to make the note available
+    validator_identity_key bytea NOT NULL REFERENCES validators (identity_key),
+    -- unbonding_height can't be negative
+    CONSTRAINT positive_unbonding_height CHECK (unbonding_height >= 0)
 );
-CREATE INDEX ON quarantined_notes (height);
+CREATE INDEX ON quarantined_notes (unbonding_height);
 CREATE INDEX ON quarantined_notes (validator_identity_key);
 
 -- Set of quarantined nullifiers: once the epoch is large enough, all old-enough quarantined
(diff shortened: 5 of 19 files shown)


chain/src/params.rs
```diff
@@ -41,6 +41,7 @@ impl From<AssetInfo> for pb::AssetInfo {
 pub struct ChainParams {
     pub chain_id: String,
     pub epoch_duration: u64,
+    pub unbonding_epochs: u64,
```
plaidfinch, line 44: `ChainParams` now contains an `unbonding_epochs` field, which will be accessible throughout the application.

crypto/src/nullifier.rs
```diff
@@ -7,7 +7,7 @@ use once_cell::sync::Lazy;
 
 use crate::Fq;
 
-#[derive(PartialEq, Eq, Derivative, Clone, Hash, PartialOrd, Ord)]
+#[derive(PartialEq, Eq, Derivative, Clone, Copy, Hash, PartialOrd, Ord)]
```
plaidfinch, line 10: It's useful for `Nullifier` to be `Copy` and since other `Fq` newtype already are, I've done this here.

crypto/src/value.rs
```diff
@@ -45,15 +45,9 @@ impl TryFrom<pb::Value> for Value {
     }
 }
 
-#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Default)]
```
plaidfinch, line 48: The existing `Default` implementation was identical to the derived implementation.

pcli/src/command/addr.rs
```diff
@@ -50,7 +50,7 @@ impl AddrCmd {
                 let (label, address) = state.wallet().address_by_index(*index as usize)?;
 
                 if *addr_only {
-                    println!("{}", address.to_string());
```
plaidfinch, line 53: This `.to_string()` is unnecessary.

pd/migrations/20211216025603_init.sql
```diff
@@ -1,42 +1,55 @@
+-- Miscellaneous key-value storage for the application
```
plaidfinch, line 1: These changes to the schema are of three varieties:

1. extra comments, since this file is growing large enough that comments make it easier to quickly comprehend,
2. additional constraints as safety checks on the well-formedness of data we insert (some of these replace `TODO` comments calling for their creation), and
3. modifications to the quarantine tables to update them for the schema used in this work.

I wouldn't be opposed to splitting out (1) and (2) into separate PR(s), but I did them inline with this work, and if they seem acceptable, I think they are improvements we should merge.

pd/src/consensus/worker.rs
```diff
@@ -96,11 +99,8 @@ impl Worker {
 
         // Now start building the genesis block:
         self.note_commitment_tree = NoteCommitmentTree::new(0);
-        let mut genesis_block = PendingBlock::new(
-            self.note_commitment_tree.clone(),
-            app_state.chain_params.epoch_duration,
-        );
-        genesis_block.set_height(0);
+        let mut genesis_block = PendingBlock::new(self.note_commitment_tree.clone());
```
plaidfinch, line 102: We don't need the epoch duration to construct a `PendingBlock` anymore, because it doesn't directly contain this information anymore. This was duplicated from the chain parameters, so I removed the duplication.

pd/src/consensus/worker.rs
```diff
@@ -247,164 +241,235 @@ impl Worker {
             .height
             .try_into()
             .expect("height should be nonnegative");
-        let epoch = pending_block.set_height(height);
+        let epoch = pending_block.set_height(
```
plaidfinch, line 244: However, removing the `epoch_duration` from the inside of `PendingBlock` means we have to supply it here.

pd/src/consensus/worker.rs
```diff
@@ -247,164 +241,235 @@ impl Worker {
...
+                .chain_params_rx()
+                .borrow()
+                .epoch_duration,
+        );
 
         tracing::debug!(?height, ?epoch, end_height = ?epoch.end_height());
 
+        // Find out which validators were slashed in this block
+        let slashed_validators = pending_block
+            .validator_state_changes
```
plaidfinch, line 257: @zbuc is this where this information will be found in the refactored version you're working on? Can you take a look here to see what needs to be adjusted?

pd/src/consensus/worker.rs
```diff
@@ -247,164 +241,235 @@ impl Worker {
...
-
-            tracing::info!(
-                ?height,
-                ?prev_epoch,
-                ?current_epoch,
-                ?next_epoch,
-                "crossed epoch boundary, processing rate updates"
-            );
-            metrics::increment_counter!("epoch");
+            self.end_epoch().await?;
```
plaidfinch, line 279: All the end-of-epoch code is factored out into this one function now :)

pd/src/consensus/worker.rs
```diff
@@ -247,164 +241,235 @@ impl Worker {
...
 
-            /// FIXME: set this less arbitrarily, and allow this to be set per-epoch
-            /// 3bps -> 11% return over 365 epochs, why not
-            const BASE_REWARD_RATE: u64 = 3_0000;
+        // Find all the validators which were *not* slashed in this block
+        let well_behaved_validators = pending_block
+            .validator_state_changes
+            .iter()
+            // THIS IS A LOAD-BEARING NEGATION: we want all validators which are *NOT* slashed
+            .filter(|p| !matches!(p.1, ValidatorState::Slashed))
```
plaidfinch, line 328: We filter out validators who have been slashed in this block, so that we don't simultaneously unbond and slash any notes/nullifiers. While the database ordering currently gives precedence to slashing, it would be bad if a refactor changed this and silently introduced a bug where undelegations from a slashed validator unbond without being slashed if the validator is slashed in the same block that the unbonding should occur. This is prevented doubly!

pd/src/consensus/worker.rs
```diff
@@ -247,164 +241,235 @@ impl Worker {
...
+            .map(|p| p.0.clone()) // get the validator ID
+            .collect::<Vec<_>>();
+
+        // Process unbonding notes and nullifiers for this epoch
```
plaidfinch, line 332: This is where unbonding actually happens in the `PendingBlock`.

Notice that we select only from the well behaved validators, and we select all notes/nullifiers whose unbonding height is equal to or less than the height of this epoch-ending block. This naturally means that if the epoch duration or the unbonding period are changed, two things are true:

1. unbonding still happens always on epoch boundaries, but
2. a change to either of these parameters will delay the unbonding of notes delegated before the change is applied by at most the new epoch duration.

pd/src/consensus/worker.rs
```diff
@@ -247,164 +241,235 @@ impl Worker {
...
+                .epoch_duration,
+        );
 
         tracing::debug!(?height, ?epoch, end_height = ?epoch.end_height());
 
+        // Find out which validators were slashed in this block
+        let slashed_validators = pending_block
+            .validator_state_changes
+            .iter()
+            .filter(|p| matches!(p.1, ValidatorState::Slashed))
```
plaidfinch, line 259: We filter out validators who were slashed in *this block* here: this is correct because we don't need to worry about previous blocks' slashings, since we should already have removed their quarantined items from the database during the commit of that block.

pd/src/consensus/worker.rs
```diff
@@ -247,164 +241,235 @@ impl Worker {
...
+            .map(|p| p.0) // get the validator ID
+            .collect::<Vec<_>>();
+
+        // Immediately revert notes and nullifiers immediately from slashed validators in this block
```
plaidfinch, line 263: We immediately revert quarantined notes and nullifiers undelegated from a slashed validator; we do not wait for the next epoch boundary.

pd/src/consensus/worker.rs
```diff
@@ -247,164 +241,235 @@ impl Worker {
...
+        while let Some(result) = unbonding_notes.next().await {
+            let (_, commitment, data) = result?;
+            pending_block.add_note(commitment, data);
+        }
+        while let Some(result) = unbonding_nullifiers.next().await {
+            pending_block.unbonding_nullifiers.insert(result?.1); // insert the nullifier
+        }
+        drop(unbonding_notes);
+        drop(unbonding_nullifiers);
+
```
plaidfinch, line 346: The rest of this file's changes are not things I've touched; this is re-arrangement noise due to the rebase and the creation of the separated `end_epoch` function.

pd/src/genesis.rs
```diff
@@ -116,7 +116,7 @@ impl TryFrom<pb::genesis_app_state::ValidatorPower> for ValidatorPower {
 impl Protobuf<pb::genesis_app_state::ValidatorPower> for ValidatorPower {}
 
 /// The application state at genesis.
-#[derive(Deserialize, Serialize, Debug, Clone)]
+#[derive(Deserialize, Serialize, Debug, Clone, Default)]
```
plaidfinch, line 119: The genesis app state can derive `Default` rather than implementing it directly.

pd/src/main.rs
```diff
@@ -61,6 +61,9 @@ enum Command {
         /// Number of blocks per epoch.
         #[structopt(short, long, default_value = "60")]
         epoch_duration: u64,
+        /// Number of epochs before unbonding stake is released.
+        #[structopt(short, long, default_value = "60")]
+        unbonding_epochs: u64,
```
plaidfinch, line 66: When generating a testnet, we now need to specify how many epochs unbonding takes, or the default will be used. @zbuc is this a good default?

pd/src/pending_block.rs
```diff
@@ -112,6 +134,13 @@ impl PendingBlock {
             transaction_id: [0; 32],
         };
 
+        self.add_note(commitment, note_data);
+
+        self.reward_counter += 1;
+    }
+
+    /// Adds a new note to this pending block.
+    pub fn add_note(&mut self, commitment: note::Commitment, data: NoteData) {
```
plaidfinch, line 143: The same code was being used to add a note in `add_transaction` and `add_validator_reward_note`; we also need to call it directly to add the unbonding notes to the `PendingBlock` when they unbond. I've factored it out into a function here.

pd/src/state.rs
```diff
@@ -14,6 +14,7 @@ pub use writer::Writer;
 pub async fn new(uri: &str) -> Result<(Reader, Writer)> {
     // Maintain two connection pools, so that reader contention cannot starve
     // the writer.
+    #[allow(clippy::zero_prefixed_literal)]
```
plaidfinch, line 17: Hush, clippy, hush :paperclip:

pd/src/state/reader.rs
```diff
@@ -476,6 +477,108 @@ impl Reader {
...
+    /// equal to that height will be returned.
+    ///
+    /// If `validators` is `Some`, only notes which were associated with an undelegation from some
+    /// validator in that set will be returned. (This is more efficient than filtering after
+    /// receiving he stream, because the database is performing the filtration.)
+    pub fn quarantined_notes<'a>(
+        &self,
+        maximum_unbonding_height: Option<u64>,
+        validators: Option<impl IntoIterator<Item = impl Borrow<&'a IdentityKey>>>,
+    ) -> impl Stream<Item = Result<(IdentityKey, note::Commitment, NoteData)>> + Send + Unpin + '_
```
plaidfinch, line 493: This function is called during slashing and unbonding of notes; it's generalized so that it is applicable in both situations.

pd/src/state/reader.rs
```diff
@@ -476,6 +477,108 @@ impl Reader {
...
+    /// equal to that height will be returned.
+    ///
+    /// If `validators` is `Some`, only nullifiers which were associated with an undelegation from some
+    /// validator in that set will be returned. (This is more efficient than filtering after
+    /// receiving he stream, because the database is performing the filtration.)
+    pub fn quarantined_nullifiers<'a>(
+        &self,
+        maximum_unbonding_height: Option<u64>,
+        validators: Option<impl IntoIterator<Item = impl Borrow<&'a IdentityKey>>>,
+    ) -> impl Stream<Item = Result<(IdentityKey, Nullifier)>> + Send + Unpin + '_ {
```
plaidfinch, line 547: This function is called during slashing and unbonding of notes; it's generalized so that it is applicable in both situations.

pd/src/state/writer.rs
```diff
@@ -238,6 +238,36 @@ impl Writer {
         .execute(&mut dbtx)
         .await?;
 
+        // Drop quarantined notes associated with a validator slashed in this block
```
plaidfinch, line 241: Here is where we apply the changes resulting from slashing: by dropping the notes associated with a validator which are in quarantine, and removing the nullifiers _both_ from quarantine and from the **main nullifier set** which makes them spendable again.

pd/src/state/writer.rs
```diff
@@ -238,6 +238,36 @@ impl Writer {
...
+            .execute(&mut dbtx)
+            .await?;
+        }
+
+        // Drop quarantined nullifiers from the main nullifier set if they were associated with a
+        // validator slashed in this block (thus reverting their spend)
+        for nullifier in block.reverting_nullifiers {
+            // Forget about this nullifier, making the associated note spendable again
+            query!(
+                "DELETE FROM nullifiers WHERE nullifier = $1",
```
plaidfinch, line 256: This is very dangerous; a bug that introduced a nullifier into `reverting_nullifiers` would enable a double-spend. Please check my work!

pd/src/state/writer.rs
```diff
@@ -259,6 +289,75 @@ impl Writer {
...
+        // Calculate the height at which notes quarantined in this block should unbond. If the
+        // unbonding period or the epoch duration change, notes will unbond at the nearest epoch
+        // boundary following this height.
+        let unbonding_epochs = self
+            .private_reader()
+            .chain_params_rx()
+            .borrow()
+            .unbonding_epochs;
+        let epoch_duration = block.epoch.as_ref().unwrap().duration;
+        let unbonding_height = height + (epoch_duration * unbonding_epochs);
```
plaidfinch, line 311: We calculate _at the time of undelegation submission_ the height at which the notes should become available (modulo slashing) and store _that_ height in the database. This naturally means that if the chain's epoch duration or unbonding epochs are altered, undelegations already in flight are only affected by at most the distance to the following newly-defined epoch, rather than being multiplied out into the future or suddenly made available all at once in the present. This is the least-surprising behavior in case we decide to change these parameters (however we probably won't!).

pd/src/state/writer.rs
```diff
@@ -259,6 +289,75 @@ impl Writer {
...
+                )
+                .execute(&mut dbtx)
+                .await?;
+            }
+
+            // Quarantine all nullifiers associated with this quarantine group
+            for &nullifier in nullifiers.iter() {
+                let nullifier_bytes = &<[u8; 32]>::from(nullifier)[..];
+
+                // Keep track of the nullifier associated with the block height
```
plaidfinch, line 349: You might think that we need to also insert the nullifier into the main nullifier set here, but that's taken care of already by this code in `PendingBlock::add_transaction`: https://github.com/penumbra-zone/penumbra/blob/8fc17d4606ad03d5cabb6824606267b9c91966ef/pd/src/pending_block.rs#L177-L181

pd/src/verify.rs
```diff
@@ -39,8 +39,8 @@ pub struct PendingTransaction {
     pub spent_nullifiers: BTreeSet<Nullifier>,
     /// Delegations performed in this transaction.
     pub delegations: Vec<Delegate>,
-    /// Undelegations performed in this transaction.
-    pub undelegations: Vec<Undelegate>,
+    /// Undelegation, if any, performed in this transaction (there must be no more than one).
+    pub undelegation: Option<Undelegate>,
```
plaidfinch, line 43: We now have a consensus rule saying there's no more than one undelegation per transaction, to simplify user comprehension and our own implementation.

pd/src/verify/stateful.rs
```diff
@@ -171,5 +172,6 @@ pub fn mark_genesis_as_verified(transaction: Transaction) -> VerifiedTransaction
         new_notes,
         spent_nullifiers: BTreeSet::<Nullifier>::new(),
         delegation_changes: BTreeMap::new(),
+        undelegation_validator: None,
```
plaidfinch, line 175: Genesis contains no undelegations.

pd/src/verify/stateless.rs
```diff
@@ -99,23 +99,36 @@ impl StatelessTransactionExt for Transaction {
...
                 }
                 Action::Undelegate(undelegate) => {
-                    // There are currently no stateless verification checks than the ones implied by
-                    // the binding signature.
-                    undelegations.push(undelegate);
+                    if undelegation.is_none() {
+                        undelegation = Some(undelegate);
+                    } else {
+                        return Err(anyhow::anyhow!("Multiple undelegations in one transaction"));
+                    }
```
plaidfinch, line 106: The consensus rule of one undelegation per transaction is added here.

pd/src/verify/stateless.rs
```diff
@@ -99,23 +99,36 @@ impl StatelessTransactionExt for Transaction {
...
+        // We prohibit actions other than `Spend`, `Delegate`, `Output` and `Undelegate` in
+        // transactions that contain `Undelegate`, to avoid having to quarantine them.
+        if undelegation.is_some() {
+            use Action::*;
+            for action in self.transaction_body().actions {
+                if !matches!(action, Undelegate(_) | Delegate(_) | Spend(_) | Output(_)) {
+                    return Err(anyhow::anyhow!("transaction contains an undelegation, but also contains an action other than Spend, Delegate, Output or Undelegate"));
+                }
+            }
+        }
```
plaidfinch, line 123: Additionally, we prohibit as a consensus rule any actions from sharing a transaction that contains an `Undelegate` other than `Delegate` (to allow redelegation), `Spend`, and `Output`. This means we don't need to handle the logic for things like "what if someone's swap gets quarantined" or "do quarantined votes count, and if so, when?"

pd/src/wallet.rs
```diff
@@ -60,10 +60,9 @@ impl LightWallet for state::Reader {
             .await
             .map_err(|_| tonic::Status::unavailable("error retrieving genesis configuration"))?;
 
-        Ok(tonic::Response::new(ChainParams {
-            chain_id: genesis_configuration.chain_params.chain_id,
-            epoch_duration: genesis_configuration.chain_params.epoch_duration,
-        }))
+        Ok(tonic::Response::new(
+            genesis_configuration.chain_params.into(),
```
plaidfinch, line 64: We can just copy the chain parameters wholesale here.

proto/proto/chain.proto
```diff
@@ -7,8 +7,10 @@ import "crypto.proto";
 message ChainParams {
   // The identifier of the chain.
   string chain_id = 1;
-  // The transaction fee.
+  // The duration of each epoch, in number of blocks.
```
plaidfinch, line 10: Fix outdated comment.

proto/proto/chain.proto
```diff
@@ -7,8 +7,10 @@ import "crypto.proto";
...
   uint64 epoch_duration = 2;
+  // The number of epochs an unbonding note for before being released.
+  uint64 unbonding_epochs = 3;
```
plaidfinch, line 13: Add this to the `ChainParams` proto, so it matches the domain type.

plaidfinch (commented):
I've done a self-review to call out what's going on in these changes, especially noting correctness-critical logic and changes to consensus rules.

pd/src/main.rs
zbuc: 60 will coincide with the epoch duration. I think that makes sense as a default for now, but for mainnet we'll likely want different values for both.

pd/src/pending_block.rs
```diff
@@ -43,31 +41,55 @@ pub struct PendingBlock {
     reward_counter: u64,
     /// Records pending state changes to validators.
     pub validator_state_changes: BTreeMap<IdentityKey, ValidatorState>,
+    /// Records all the quarantined inputs/outputs from this block.
```
zbuc, line 44: The pending block is getting pretty heavy, we probably want to encapsulate the unbonding logic and data

pd/src/state/writer.rs
zbuc: I'm not seeing where `reverting_nullifiers` has any nullifiers added to it

pd/src/consensus/worker.rs
hdevalence: I don't follow -- isn't this filtering *in* validators who were slashed in this block, not filtering them out?

hdevalence: (The code looks correct, just confused by your review comment)

pd/src/consensus/worker.rs
hdevalence: Going forward, we'll need to change the signature of `end_epoch()` at some point so that we can propagate validator info into the `EndBlock` response.

pd/src/consensus/worker.rs
hdevalence: nit: I think this could be
```
.filter(|p| {
    // THIS IS A LOAD-BEARING NEGATION: we want all validators which are *NOT* slashed
    !matches!(...)
})
```
so that the comment can't detach from the closure in refactoring.

pd/src/main.rs
hdevalence: coinciding isn't a problem, right? like with these defaults, the unbonding period is 3600 blocks, correct?

pd/src/state/writer.rs
hdevalence: LGTM

pd/src/verify.rs
hdevalence: Question: is the consensus rule that there is *no more than one undelegation per transaction*, or that there is *no more than one validator undelegated from*? i.e., is it allowed to have multiple `Undelegate` descriptions for the same validator?

(I don't think the choice matters, since there's no reason to have multiple, but just want to be precise about it).

hdevalence: NVM, resolved below.

hdevalence (approved):
LGTM!

PR #430, opened 2022-02-09
feat: jumbled address encoding

redshiftzero:
This is a breaking change to the address format.

The address encodings are now:
1. The 75 bytes we had before: `d || pk_d || ck_d`
2. Padded to 80 bytes: `d || pk_d || ck_d || pen00`
3. Permuted using F4jumble (using the upstream zcash crate)
4. Bech32m encoded

If this looks reasonable, I could write code to Bech32m decode the discord addresses and re-encode in the new jumbled format, i.e. to update: https://github.com/penumbra-zone/penumbra/blob/main/testnets/discord-addresses


hdevalence:
This should probably update the prefix, right?

plaidfinch:
Could we have the address parser parse old-format addresses and pretty-print only as new-format? This would mean that people who wrote down an old-format address during a prior testnet would not run into a hard error when using it.

crypto/src/address.rs
```diff
@@ -128,8 +144,9 @@ impl std::str::FromStr for Address {
     type Err = anyhow::Error;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let inner = bech32str::decode(s, bech32str::address::BECH32_PREFIX, bech32str::Bech32m)?;
         pb::Address {
-            inner: bech32str::decode(s, bech32str::address::BECH32_PREFIX, bech32str::Bech32m)?,
+            inner: f4jumble_inv(&inner).ok_or(std::fmt::Error)?,
```
hdevalence, line 149: I'm not sure this is the right place for the jumbling, because it's only being applied to the displayed bytes, not to the proto-encoded data.

Instead, the jumbling should happen at the boundary between the domain type and the proto type, so that the proto type always records jumbled bytes, and we don't need any special-cased paths.

crypto/src/address.rs
```diff
@@ -81,6 +86,9 @@ impl From<Address> for pb::Address {
         bytes
             .write_all(&a.clue_key().0)
             .expect("can write clue key into vec");
+        bytes
+            .write_all(ADDR_PADDING)
+            .expect("can write padding into vec");
```
hdevalence, line 92: i.e., jumble here before conversion to proto

hdevalence:
> Could we have the address parser parse old-format addresses and pretty-print only as new-format? This would mean that people who wrote down an old-format address during a prior testnet would not run into a hard error when using it.

Doing this means we have to maintain multiple sets of parsing code, which seems preferable to avoid. Since we retain the seed on `wallet reset`, we can just regenerate the addresses as needed.

hdevalence:
> If this looks reasonable, I could write code to Bech32m decode the discord addresses and re-encode in the new jumbled format, i.e. to update: https://github.com/penumbra-zone/penumbra/blob/main/testnets/discord-addresses

I think this could be put in a `pcli tmp` subcommand, like `pcli tmp jumble`; then we can delete it in a few weeks.

PR #434, opened 2022-02-15
wip: threshold spec updates

avahowell:
this PR adds to the spec a concrete instantiation of our homomorphic threshold encryption system


hdevalence:
@avahowell happy to do an optimistic rollup so we get rendered docs

PR #436, opened 2022-02-15
pd: use `tokio_util::sync::PollSender` in consensus svc

hawkw:
In v0.7 of `tokio-util`, the `PollSender` API was changed to have the
semantics that most users expect (see tokio-rs/tokio#4214). The
`poll_send_done` method was replaced with a `poll_reserve` method, and
the implementation rewritten to drive a `Sender::reserve_owned` future
that's consumed in `send_item`. Now, `PollSender` essentially implements
exactly the same code that was written by hand in the consensus service.
We can simplify the consensus service significantly by upgrading the
`tokio-util` dependency to 0.7 and replacing the hand-written version
with `PollSender`.

There should be no functional change as a result of this refactor.

### DIFF
--- a/pd/src/consensus/service.rs
+++ b/pd/src/consensus/service.rs
@@ -4,28 +4,18 @@ use std::{
     task::{Context, Poll},
 };
 
-use futures::{ready, FutureExt};
+use futures::FutureExt;
 use tendermint::abci::{ConsensusRequest, ConsensusResponse};
-use tokio::sync::{
-    mpsc::{self, error::SendError, OwnedPermit},
-    oneshot,
-};
-use tokio_util::sync::ReusableBoxFuture;
+use tokio::sync::{mpsc, oneshot};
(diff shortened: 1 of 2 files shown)


pd/src/consensus/service.rs
```diff
@@ -35,60 +25,32 @@ impl Consensus {
...
-
         let span = req.create_span();
         let (tx, rx) = oneshot::channel();
 
-        permit.send(Message {
-            req,
-            rsp_sender: tx,
-            span,
-        });
+        self.queue
```
hawkw, line 47: we may want to check if the channel is closed here and bail early (in case it closed between a successful `poll_ready` call and the call to `call`). i don't know how that matters in `pd`, specifically, though, since there's only ever going to be a single consensus background worker and it's expected to live for the entire lifetime of the node...if it's gone, panicking in the `send_item` call is probably also fine.

I didn't change this because the previous code didn't have such a check.

PR #441, opened 2022-02-17
spec: threshold: fix some rendering bugs and make notation more consistent

avahowell:
this PR fixes some rendering bugs in the rendered LaTex, changes the notation a bit to be more consistent, and adds a note on the encryption SNARK note to the threshold spec


aubrika:
I'm getting a `ParseError: KaTeX parse error: Can't use function '$' in math mode at position 49: …2} + v_3 2^{48}$̲$ with $v_i \in…` on trying to preview the markdown - looks like an error on line ~63?

avahowell:
huh, is that when running `mdbook serve`?

aubrika:
> huh, is that when running `mdbook serve`?

`mdbook serve` doesn't seem to(?) throw an error there, I saw the above when viewing as preview in VSCode using the  markdownlint extension

avahowell:
optimistically merging to fix the rendering issues on protocol.penumbra.zone

PR #442, opened 2022-02-17
Add link to Barry Whitehat's post on why a private Uniswap with ZKPs doesn't work

bertmiller:
I think this is the intended link


avahowell:
Thanks!

PR #443, opened 2022-02-18
feat: derive `SpendSeed` from seed phrase

### LINKED ISSUE #419: Derive `SpendSeed` from seed phrase
Our [`SpendSeed`](https://github.com/penumbra-zone/penumbra/blob/main/crypto/src/keys/spend.rs#L16) is currently generated randomly. We should instead derive the `SpendSeed` from a series of randomly generated words as described in [BIP39](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki) since it's easier to remember English words. Users then just need to remember the seed phrase. This would require changes to `SpendKey::generate` in the crypto crate as well as changes to the `pcli wallet` commands (e.g. to `import` from a seed phrase instead of the spend seed). 

A recommended wordlist: https://github.com/bitcoin/bips/blob/master/bip-0039/english.txt

redshiftzero:
Closes #419 

This follows BIP39 which both Ledger and Trezor are using for seed phrase generation. Right now this also allows for legacy spend seeds (i.e. those generated randomly and not from a seed phrase) so current users don't need to regenerate their wallets. There's a `pcli wallet` command to import from spend seed (unchanged from before) as well as a new import command to import from seed phrase.  

Two questions/notes:
* I made a decision here to always use 256 bit entropy (24 word) phrases. 
* ~Do we want to persist the seed phrase to the wallet file? Right now in `pcli` it's just displaying the seed phrase to the user, then persisting the `SpendSeed` derived from the seed phrase.~ discussed with Henry, seems like this is good for now


Bleujj1:
Have you gotten help with your complaints?

Issue #82, opened 2021-10-16
Avoid full expansion of clue keys in `decaf377_fmd`


hdevalence:
The FMD variant we use for Penumbra derives the child keypairs used for each bit of detection precision from a single root keypair (the compact clue key / compact detection key). In the current implementation of `decaf377_fmd`, all possible child keypairs are derived when a clue key is expanded, regardless of whether or not they will actually be used.

Two alternatives:

- the child points could be derived each time the address is used, slowing down the case where the same address is used multiple times, but aiding simplicity;
- the child points could be stored in a `RefCell` allowing interior mutability, and transparently cached inside the `ExpandedClueKey` as needed, at the cost of removing the `Sync` bound from `ExpandedClueKey`. 

This is an optimization, so it's not important to do it now.

hdevalence:
I think the best way to do this would be #2:

- change the storage to a `RefCell` holding a `Vec` instead of an array;
- add a private `fn ensure_at_least(&self, n: usize)` to `ExpandedAddress` that ensures the expanded array has at least `n` subkeys;
- call `ensure_at_least` before creating a clue.

hdevalence:
Closed in #444

PR #444, opened 2022-02-20
feat(decaf377_fmd): improve subkey generation logic

erwanor:
This PR implements #82, it's still a draft but the main parts are here. Beside evaluating whether adding nice-to-haves (e.g. `Zeroize` on `ClueKey`), fmt and tests, I need to improve error propagation if `ensure_at_least` fails.

### DIFF
--- a/decaf377-fmd/src/clue_key.rs
+++ b/decaf377-fmd/src/clue_key.rs
@@ -1,4 +1,4 @@
-use std::{convert::TryInto, cell::RefCell};
+use std::cell::RefCell;
 
 use ark_ff::{Field, UniformRand};
 use bitvec::{array::BitArray, order};
@@ -34,7 +34,7 @@ impl ClueKey {
     #[allow(non_snake_case)]
     pub fn expand(&self, precision: usize) -> Result<ExpandedClueKey, Error> {
         if precision > MAX_PRECISION {
-            return Err(Error::PrecisionTooLarge(precision))
+            return Err(Error::PrecisionTooLarge(precision));
         }
 
         let root_pub_enc = decaf377::Encoding(self.0);
@@ -49,7 +49,9 @@ impl ClueKey {
             .map(|i| hkd::derive_public(&root_pub, &root_pub_enc, i as u8))
             .collect::<Vec<_>>();
 
-        Ok(ExpandedClueKey { subkeys: RefCell::new(Xs) })
+        Ok(ExpandedClueKey {
+            subkeys: RefCell::new(Xs),
+        })
     }
 }
 
@@ -75,7 +77,7 @@ impl ExpandedClueKey {
         }
 
         if !self.ensure_at_least(precision_bits) {
-            return Err(Error::PrecisionTooLarge(precision_bits))
+            return Err(Error::PrecisionTooLarge(precision_bits));
         }
 
         let r = Fr::rand(&mut rng);
@@ -111,7 +113,6 @@ impl ExpandedClueKey {
     /// Checks that the expanded clue key has at least `precision` subkeys
     fn ensure_at_least(&self, precision: usize) -> bool {
         // TODO: handle try_borrow failure via Result
-        return self.subkeys.try_borrow().unwrap().len() >= precision
+        return self.subkeys.try_borrow().unwrap().len() >= precision;
     }
-
 }
--- a/decaf377-fmd/src/clue_key.rs
+++ b/decaf377-fmd/src/clue_key.rs
@@ -1,4 +1,4 @@
-use std::convert::TryInto;
+use std::{convert::TryInto, cell::RefCell};
 
 use ark_ff::{Field, UniformRand};
 use bitvec::{array::BitArray, order};
@@ -14,8 +14,12 @@ use crate::{hash, hkd, Clue, Error, MAX_PRECISION};
 /// situations where clue key might or might not actually be used.  This saves
 /// computation; at the point that a clue key will be used to create a [`Clue`],
 /// it can be expanded to an [`ExpandedClueKey`].
-#[derive(Copy, Clone, PartialEq, Eq, Debug)]
-pub struct ClueKey(pub [u8; 32]);
+/// Implement copy
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub struct ClueKey {
+    pub root: [u8; 32],
+    subkeys: RefCell<Vec<decaf377::Element>>,
+}
 
 /// An expanded and validated clue key that can be used to create [`Clue`]s
 /// intended for the corresponding [`DetectionKey`](crate::DetectionKey).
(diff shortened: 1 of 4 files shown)


decaf377-fmd/src/clue_key.rs
```diff
@@ -14,15 +14,15 @@ use crate::{hash, hkd, Clue, Error, MAX_PRECISION};
 /// situations where clue key might or might not actually be used.  This saves
 /// computation; at the point that a clue key will be used to create a [`Clue`],
 /// it can be expanded to an [`ExpandedClueKey`].
+/// TODO: use Zeroize?
```
hdevalence, line 17: I don't think it's important to use `Zeroize` for `ClueKey`, because the clue key is public, so there aren't secrets to leak.

To be more precise, the clue key might not be completely public, in the sense that someone might not want to give out their address, and in the sense that it's not included in any transactions, but it doesn't have cryptographic secrets, and this isn't any different level of protection than any other data involved in creating a transaction.

decaf377-fmd/src/clue_key.rs
```diff
@@ -99,4 +109,10 @@ impl ExpandedClueKey {
 
         Ok(Clue(buf))
     }
+
+    /// Checks that the expanded clue key has at least `precision` subkeys
+    fn ensure_at_least(&self, precision: usize) -> bool {
+        // TODO: handle try_borrow failure via Result
+        return self.subkeys.try_borrow().unwrap().len() >= precision;
+    }
```
hdevalence, line 117: This `ensure_at_least` just checks whether the existing expansion was large enough to handle the specified precision, but it would be more convenient as an API user to not have to think about the expansion at all, and have it happen automatically.

To do that, I think the expansion logic from the current `expand` function should move into `ensure_at_least`, so that `ensure_at_least` will expand the key up to the specified precision.  This means that the `ExpandedClueKey` would need to store the `root_pub: decaf377::Element`, so that it's accessible to `ensure_at_least`.

The existing `expand` function does three things:

1. It converts a `ClueKey` to an `ExpandedClueKey`;
2. It validates that the bytes of the `ClueKey` are a valid decaf377 encoding, and decodes them to a group element;
3. It does the expansion.

With the new `ensure_at_least` behavior, the user wouldn't have to think about expansion at all, so the only remaining behaviors are (1) and (2).  In that case, I think it would make sense to rename it to `ExpandedClueKey::new(clue_key: &ClueKey) -> Result<Self, ...>`, and to provide a convenience wrapper `ClueKey::expand(&self) -> Result<ExpandedClueKey, ...>` that calls `ExpandedClueKey::new(&self)`.

erwanor: Got it, thanks for your feedback; this is helpful and an improvement for sure!

decaf377-fmd/src/clue_key.rs
erwanor: Ok, I wasn't sure how toxic the `ClueKey` was. I know how it's used in the spec, and I know we feed it to an HKDF so I thought I'd check. Thanks that makes sense!

erwanor:
Notable changes compared to the first attempt:

1. Key expansion happens in [`ensure_at_least`](https://github.com/penumbra-zone/penumbra/pull/444/commits/ec6aa54faa15fc9023cee54baa5c450ef3f29f6b#diff-3a2138827059c0bf8c39590cc4ac23bb3c151a614030b0284c91b6e1ebde9cd1R56)
2. `ExpandedClueKey` stores both the root key group element and its encoding
3. We still use a `RefCell` to store the subkeys, but runtime borrow errors are propagated downstream. I added an `InternalError` variant and implement `From<BorrowError>` [here](https://github.com/penumbra-zone/penumbra/pull/444/commits/ddfe0c1796a196096ca5407321a754500265e013#diff-4e19fe021395cf4fdff096b0d5be72e456c958f561e41dd884c8d7317c6d7fd7R22).

I will add documentation and perhaps a couple more tests after a second round of feedback.

decaf377-fmd/src/clue_key.rs
```diff
@@ -33,25 +34,47 @@ impl ClueKey {
...
-            .try_into()
-            .expect("iterator of length `MAX_PRECISION`");
-
-        Ok(ExpandedClueKey { Xs })
+            .map(|i| hkd::derive_public(&self.root_pub, &self.root_pub_enc, i as u8))
+            .collect::<Vec<_>>();
+
+        self.subkeys.try_borrow_mut()?.append(&mut expanded_keys);
+
+        return Ok(());
```
hdevalence, line 75: This function only returns an `Err` if there's a problem borrowing `self.subkeys`.  That can only happen if there's a conflicting borrow of `subkeys`, but we know that can't happen, because `subkeys` is only used internally to the `ExpandedClueKey`, and none of the borrows live longer than a single access.

Instead, I think it'd be cleaner to use the panicing `borrow()`/`borrow_mut()` functions (which we know won't panic), so that this function can be infallible.  Making this function infallible simplifies the downstream API:

* the new error conversions from the `RefCell` errors can be removed
* the `PrecisionTooLarge` error can stay as it is, only erroring if it's larger than `MAX_PRECISION`

decaf377-fmd/src/clue_key.rs
```diff
@@ -61,6 +84,7 @@ impl ExpandedClueKey {
     /// # Errors
     ///
     /// `precision_bits` must be smaller than [`MAX_PRECISION`].
+    /// `precision_bits` must be smaller or equal to the `ExpandedClueKey` precision.
```
hdevalence, line 87: I don't think this condition should cause an error -- we should expand the clue key as needed, as long as it's less than `MAX_PRECISION`, and in fact the code works like that, so we don't need any extra conditions.

decaf377-fmd/src/clue_key.rs
```diff
@@ -21,8 +21,9 @@ pub struct ClueKey(pub [u8; 32]);
 /// intended for the corresponding [`DetectionKey`](crate::DetectionKey).
 #[allow(non_snake_case)]
```
hdevalence, line 22: nit: now that the `Xs` field is gone, the `#[allow(non_snake_case)]` could be removed too

hdevalence:
> Notable changes compared to the first attempt:
> 
>     1. Key expansion happens in [`ensure_at_least`](https://github.com/penumbra-zone/penumbra/pull/444/commits/ec6aa54faa15fc9023cee54baa5c450ef3f29f6b#diff-3a2138827059c0bf8c39590cc4ac23bb3c151a614030b0284c91b6e1ebde9cd1R56)
> 
>     2. `ExpandedClueKey` stores both the root key group element and its encoding

Awesome!


>     3. We still use a `RefCell` to store the subkeys, but runtime borrow errors are propagated downstream. I added an `InternalError` variant and implement `From<BorrowError>` [here](https://github.com/penumbra-zone/penumbra/pull/444/commits/ddfe0c1796a196096ca5407321a754500265e013#diff-4e19fe021395cf4fdff096b0d5be72e456c958f561e41dd884c8d7317c6d7fd7R22).

I think it'd be cleaner and easier to use the panicking methods, since we aren't doing anything complicated with the `RefCell` and we know that those errors won't happen (because we don't use the `RefCell` in a way that could cause them).

hdevalence:
This looks great! Thank you for your help with this, I merged it in.

PR #453, opened 2022-03-01
crypto: add `threshold` module

avahowell:
this PR adds a new module `threshold` to the existing crypto crate, which implements the encryption and threshold decryption described in the [flow encryption](https://protocol.penumbra.zone/main/crypto/flow-encryption/threshold-encryption.html) spec. 

a few properties are tested here: basic encryption/decryption, the additive homomorphism, as well as full threshold decryption (using FROST's DKG) using a randomly selected (t/n) decryption shares.


plaidfinch:
For the public API (and probably the private API too), I think it makes sense to use triple-slash `///` doc-comments so that `rustdoc` will document those functions, and to use the Rust documentation standard of capitalizing these doc comments and terminating them with a period. Also note that you can do intra-doc links with markdown links, like:

```rust
/// Verify the `EncryptionProof` for this EncryptedValue.
///
/// See the [spec](https://...) for more details.
```

crypto/src/threshold/threshold.rs
```diff
@@ -0,0 +1,387 @@
...
+        share.verify(encrypted_value.c1, participant_commitments[i])?;
+    }
+
+    let indicies = decryption_shares
+        .iter()
+        .map(|share| share.participant_index)
+        .collect::<Vec<_>>();
+
+    let mut d = decaf377::Element::default();
+    for share in decryption_shares {
```
hawkw, line 110: note that the allocation of an intermediate `Vec` can be avoided here --- it isn't necessary to store the vector of decryption shares on the heap, since they're only iterated over once:
```suggestion
        .map(|share| share.participant_index);

    let mut d = decaf377::Element::default();
    for share in decryption_shares {
```

crypto/src/threshold/threshold.rs
```diff
@@ -0,0 +1,387 @@
...
+/// Sum a set of encrypted values together using the additive homomorphism.
+pub fn aggregate_values(
+    values: &[EncryptedValue],
+    dkg_pubkey: decaf377::Element,
+) -> Result<EncryptedValue, anyhow::Error> {
+    let mut res = values.first().unwrap().clone();
+    res.verify(dkg_pubkey)?;
+    for i in 1..values.len() {
+        values[i].verify(dkg_pubkey)?;
+        res = res.add(&values[i]);
```
hawkw, line 125: consider:
```suggestion
    for value in values.iter().skip(1) {
        value.verify(dkg_pubkey)?;
        res = res.add(&value);
```
(this should elide unnecessary bounds checks)

crypto/src/threshold/threshold.rs
```diff
@@ -0,0 +1,387 @@
...
+}
+
+/// Decrypt a given [`EncryptedValue`] in the threshold setting using a set of decryption shares.
+pub fn decrypt_value(
+    encrypted_value: &EncryptedValue,
+    decryption_shares: &[DecryptionShare],
+    participant_commitments: &[decaf377::Element],
+) -> Result<decaf377::Element, anyhow::Error> {
+    for (i, share) in decryption_shares.iter().enumerate() {
+        share.verify(encrypted_value.c1, participant_commitments[i])?;
```
hawkw, line 101: array indexing could be replaced with a second iterator here, which may elide bounds checks:
```suggestion
    for (participant_commitment, share) in decryption_shares.iter().zip(participant_commitments) {
        share.verify(encrypted_value.c1, participant_commitment)?;
```

hawkw (commented):
some very minor drive-by suggestions, hope they're useful! :)

crypto/src/threshold/threshold.rs
avahowell: the set of all the indices is needed in each iteration through decryption_shares (not just the index for that particular share) , so I'm not sure how to avoid collecting all the indices into a separate vec here

avahowell:
thanks!

hdevalence (approved):
LGTM as a first stub -- let's make a tracking issue!

PR #455, opened 2022-03-01
Refactor ValidatorSet

zbuc:
I haven't had a chance to fully test this yet.


pd/src/components/validator_set.rs
```diff
@@ -83,61 +106,256 @@ impl ValidatorSet {
...
+                    name,
+                    website,
+                    description,
+                    voting_power,
+                    validator_state,
+                    unbonding_epoch
+                ) VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9)",
+                v.validator.identity_key.encode_to_vec(),
+                v.validator.consensus_key.to_bytes(),
+                v.validator.sequence_number as i64,
```
avahowell, line 189: could there be an overflow issue here, where if the validator sets the sequence number to some value greater than max_i64, this will overflow to being negative, and cause replay to be possible for validator definitions?

pd/src/components/validator_set.rs
```diff
@@ -83,61 +106,256 @@ impl ValidatorSet {
...
+                v.validator.name,
+                v.validator.website,
+                v.validator.description,
+                v.status.voting_power as i64,
```
avahowell, line 193: ditto here, i think this should probably be a checked cast (though it would be really nice if our DB could support u64s :p)

pd/src/components/validator_set.rs
```diff
@@ -373,24 +598,28 @@ impl ValidatorSet {
...
+        new_epoch: Epoch,
+    ) -> impl Future<Output = Result<()>> + Send + Unpin + '_ {
         let chain_params = self.reader.chain_params_rx().borrow();
         let unbonding_epochs: u64 = chain_params.unbonding_epochs;
         let active_validator_limit: u64 = chain_params.active_validator_limit;
         drop(chain_params);
 
+        let prev_epoch = self.epoch().clone();
+        tracing::debug!(?new_epoch, "Advancing epoch");
+        self.set_epoch(new_epoch);
```
avahowell, line 612: should we be validating monotonicity of the new epoch here?

pd/src/state/writer.rs
```diff
@@ -161,30 +179,181 @@ impl Writer {
...
+
+        // write the token supplies to the database
+        for (id, asset) in &supply_updates {
+            query!(
+                "INSERT INTO assets (asset_id, denom, total_supply)
+            VALUES ($1, $2, $3)
+            ON CONFLICT (asset_id) DO UPDATE SET denom=$2, total_supply=$3",
+                &id.to_bytes()[..],
+                asset.0.to_string(),
+                asset.1 as i64
```
avahowell, line 322: ditto on unchecked cast here: if we have a total supply close to max u64 this could overflow

avahowell (commented):
this looks great!

pd/src/components/validator_set.rs
zbuc: Good catch!

hdevalence (approved):
LGTM!

PR #457, opened 2022-03-01
fix: resolve error when base rate is updated at epoch boundary

### LINKED ISSUE #456: pd: error updating base rates at epoch boundary
Flagged by @zbuc

## Steps to reproduce

1. Start `pd`
2. Run node until the epoch boundary is reached (recommend a short epoch duration for testing purposes, not required for reproduction)

## Expected behavior

Next base rate is saved in the `base_rates` table

## Actual behavior

```
Mar 01 13:15:06.761 DEBUG abci{addr=127.0.0.1:53888}:abci:Commit: pd::state::writer: Saving next base rate to the database base_rate_data=BaseRateData { epoch_index: 2, base_reward_rate: 30000, base_exchange_rate: 100030000 }
Mar 01 13:15:06.765  INFO abci{addr=127.0.0.1:53888}:abci:Commit: sqlx::query: INSERT INTO base_rates VALUES …; rows: 0, elapsed: 3.755ms

INSERT INTO
  base_rates
VALUES
  ($1, $2, $3)

thread 'tokio-runtime-worker' panicked at 'commit must succeed: error returned from database: duplicate key value violates unique constraint "base_rates_pkey"

Caused by:
    duplicate key value violates unique constraint "base_rates_pkey"', /Users/redshiftzero/Documents/penumbra/pd/src/consensus/worker.rs:89:26
thread 'tokio-runtime-worker' panicked at 'worker error??: RecvError(())', pd/src/consensus/service.rs:93:34
Mar 01 13:15:06.768  INFO sqlx::query: /* SQLx ping */; rows: 0, elapsed: 427.458µs
```

redshiftzero:
Resolves #456

### DIFF
--- a/pd/sqlx-data.json
+++ b/pd/sqlx-data.json
@@ -69,6 +69,20 @@
       ]
     }
   },
+  "155f64fb6352d1170eb4df09f7bf64c98edbfbf88c03168d6b80d84905eb0517": {
+    "query": "INSERT INTO base_rates VALUES ($1, $2, $3) ON CONFLICT ON CONSTRAINT base_rates_pkey DO UPDATE SET base_reward_rate=$2, base_exchange_rate=$3",
+    "describe": {
+      "columns": [],
+      "parameters": {
+        "Left": [
+          "Int8",
+          "Int8",
+          "Int8"
+        ]
+      },
+      "nullable": []
+    }
+  },
   "2b00fd7700707a635a3d5827f69f2a16a45737fe24797d9aae3874c95d640524": {
     "query": "DELETE FROM nullifiers WHERE nullifier = $1",
     "describe": {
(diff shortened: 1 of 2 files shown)


zbuc:
Thanks for picking that up. 

https://github.com/penumbra-zone/penumbra/pull/455 also fixes this issue, sorry for not bringing it to your attention.

I resolved the issue for the base rates table, but a similar issue existed for the validator rates table for complicated reasons. Basically delegations require the next epoch's rates to be available, however they can also change during the epoch (I.e. if the validator is slashed).

So, I went with `ON CONFLICT UPDATE` instead of `ON CONFLICT DO NOTHING` because the rate needs to be overwritten if slashed. Let me know if you think that makes sense.

zbuc:
Agreed!

PR #463, opened 2022-03-02
Track unbonding stake in `pcli`

plaidfinch:
This will be a breaking change for `pcli`/`pd` communication, because although the protobufs are shape-compatible, it changes the circumstances in which a nullifier appears in the `CompactBlock`: namely, it is only to appear in the `CompactBlock`'s `nullifiers` field after it has been fully unbonded; before then, it appears only in the `quarantined_nullfiers` field.

I have not yet tested any of this; that's for tomorrow.

One thing not accomplished by this so far is to display which validator the stake is being unbonded from. At present, this information is not exposed by the `CompactBlock`, but it would be easy to do so, and then to thread through tracking it into the wallet, and thereby into `pcli`. Is this worth doing now?

### DIFF
--- a/pcli/src/command/stake.rs
+++ b/pcli/src/command/stake.rs
@@ -1,8 +1,9 @@
-use std::collections::{BTreeMap, HashMap};
+use std::{collections::BTreeMap, time::Duration};
 
 use anyhow::{anyhow, Context, Result};
 use comfy_table::{presets, Table};
 use futures::stream::TryStreamExt;
+use itertools::Itertools;
 use penumbra_crypto::Value;
 use penumbra_proto::{light_wallet::ValidatorInfoRequest, thin_wallet::ValidatorRateRequest};
 use penumbra_stake::{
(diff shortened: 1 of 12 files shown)


pcli/src/command/stake.rs
```diff
@@ -242,17 +244,65 @@ impl StakeCmd {
...
+                    );
+                    let approx_eta = eta.to_string().split(' ').take(2).join(" ");
+
+                    table.add_row(vec![
+                        format!("Unbonding (ETA: {approx_eta})"),
+                        value.try_format(state.asset_cache()).unwrap(),
+                        format!("{:.4}", 1.0),
+                        value.try_format(state.asset_cache()).unwrap(),
+                    ]);
+                }
```
hdevalence, line 305: Am I reading correctly that this code is only showing the estimated ETA (in units of time) and not the precise ETA (in blocks)?

I think it would be useful to display the exact unbonding time (in blocks), and then have the ETA in parentheses as an estimate afterwards.

That way, if the estimate bounces around, it's more clear that what's happening is that the *estimate* is moving, rather than the actual unbonding height.

pd/migrations/20211216025603_init.sql
```diff
@@ -113,30 +106,59 @@ CREATE TABLE IF NOT EXISTS delegation_changes (
...
+    ),
+    -- unbonding_height can't be negative
+    CONSTRAINT positive_unbonding_height
+        CHECK (unbonding_height IS NULL OR unbonding_height >= 0),
+    -- if unbonding_height is not null, it is greater than or equal to height
+    CONSTRAINT unbonding_height_greater_than_height
+        CHECK (unbonding_height IS NULL OR height <= unbonding_height),
+    -- if reverted_height is not null, it is less than or equal to unbonding_height
+    CONSTRAINT reverted_height_less_than_or_equal_unbonding_height
+        CHECK (reverted_height IS NULL OR reverted_height <= unbonding_height)
```
hdevalence, line 138: These constraints are fine as a backup, but we should not be pushing our application logic into the database, and I worry that adding them now commits to maintaining them in the future.

pd/src/lib.rs
```diff
@@ -1,7 +1,7 @@
 //! Source code for the Penumbra node software.
 
 // This is for the async_stream macros
-#![recursion_limit = "512"]
+#![recursion_limit = "1024"]
 #![allow(clippy::clone_on_copy)]
```
hdevalence, line 5: Why did this get bigger?

pd/migrations/20211216025603_init.sql
```diff
@@ -113,30 +106,59 @@ CREATE TABLE IF NOT EXISTS delegation_changes (
...
--- be inserted into the NCT and the notes table (and removed from this table)
+-- Nullifiers, indexed by height
+CREATE TABLE IF NOT EXISTS nullifiers (
+    nullifier bytea PRIMARY KEY,
+    height bigint NOT NULL REFERENCES blocks (height),
+
+    -- if this nullifier was quarantined at any point in the past:
+    unbonding_height bigint, -- height at which to make the spend permanent
+    reverted_height bigint REFERENCES blocks (height), -- height at which the revert was made, if it was reverted
+    validator_identity_key bytea REFERENCES validators (identity_key),
```
hdevalence, line 117: I don't think it's a good idea to mix this data in with the nullifier set.  Checking whether something is in the nullifier set should be a simple set membership query, we should not have to handle other data, or deal with whether or not these fields are present or absent.

hdevalence (changes requested):
I don't think this is a good approach. It pushes a lot of application logic into the database, which is the opposite direction we want to be moving in, and it mixes two different roles into the same database table, distinguished by nullable fields.

hdevalence:
> One thing not accomplished by this so far is to display which validator the stake is being unbonded from. At present, this information is not exposed by the `CompactBlock`, but it would be easy to do so, and then to thread through tracking it into the wallet, and thereby into `pcli`. Is this worth doing now?

I think this could be worth doing at some point, but probably isn't worth doing right now (i.e., before trying to get the testnet out).

pcli/src/command/stake.rs
```diff
@@ -242,17 +244,65 @@ impl StakeCmd {
...
+
+                // Lifted out because borrow checker
+                let empty_notes = BTreeMap::default();
+
+                for note in notes
+                    .get(&*STAKING_TOKEN_DENOM)
+                    .unwrap_or(&empty_notes)
+                    .values()
+                    .flatten()
+                {
```
hdevalence, line 266: I don't think this logic is correct, because it assumes that every quarantined note can only hold the staking token, but this is not true (it could also be a different delegation token).

hdevalence:
Closing in favor of #588

PR #468, opened 2022-03-02
Automatically detect latest testnet when generating testnet

plaidfinch:
This uses the build script to compute the chronologically most recent testnet and embed the allocations and validators of it in the binary, which means several things:

1. Running `pd generate-testnet` is no longer dependent on being run from the root of the `penumbra` repository, as it previously implicitly was.
2. Running `pd generate-testnet` now appropriately uses the latest available testnet as a default for the chain id, allocations, and validator set, although each can be individually overridden by a command line flag as before.
3. We do not need to maintain these defaults every time we do a new testnet (and indeed, we had forgotten to maintain them before; I found this issue because they were still pointed at Thelxinoe).

This also incorporates a smattering of improvements to error reporting during testnet generation, most notably better error contexts for file reading and parsing, which should lead to easier diagnostics when deploying testnets in the future.

### DIFF
--- a/pd/Cargo.toml
+++ b/pd/Cargo.toml
@@ -68,3 +68,4 @@ once_cell = "1.7.2"
 
 [build-dependencies]
 vergen = "5"
+anyhow = "1"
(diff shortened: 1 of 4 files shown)


zbuc:
LGTM!

PR #476, opened 2022-03-10
ci: add gh action with pd smoke test

### LINKED ISSUE #475: CI: run testnet for X blocks
We've had a few `pd`-crashing bugs in `main` at:
1. `InitChain`
2. the `Commit` phase at the end of a block
3. During the transition between epochs

Right now we don't catch these in CI or test (related: #230).

We could have a CI job that spins up the testnet, runs for X blocks (or X * (10 seconds) time) ideally up through the first epoch boundary. If `pd` crashes, the job would fail. This would also provide a clear indication to developers as to the state of the `main` branch.

redshiftzero:
Closes #475 (https://github.com/penumbra-zone/penumbra/issues/475#issuecomment-1063425837 I'll put in a new ticket since that's a good next step)

This adds a gh action that builds and runs a testnet for a configurable amount of time. It will fail the job (non-zero exit code from pd) if a crash is encountered in that time. See [here](https://github.com/redshiftzero/penumbra/runs/5487945502?check_suite_focus=true) for a successful build and [here](https://github.com/redshiftzero/penumbra/runs/5487953621?check_suite_focus=true) for a failed build where a panic occurred during `InitChain`. 

Currently this is configured to run only on the `main` branch since the job is slow. I think it's better to put slow jobs on `main` and act on build failures, but if we are finding that we're having a lot of crashers hit `main` then we could reconsider and add to PRs (and maybe spend some cycles speeding up this job).

### DIFF
--- a/.github/workflows/smoke.yml
+++ b/.github/workflows/smoke.yml
@@ -1,0 +1,26 @@
+on: [push, pull_request]
+
+jobs:
+  smoke_test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+
+      - name: Install sqlx.
+        run: cargo install sqlx-cli
+
+      - name: Build the testnet.
+        run: |
+          mkdir ~/scratch
+          ./scripts/docker_compose_freshstart.sh ~/scratch/testnet_build
+          chmod -R 777 ~/scratch/testnet_build
+          docker-compose build
+
+      - name: Run testnet for smoke test duration.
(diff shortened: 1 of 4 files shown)


avahowell:
LGTM

hdevalence:
> Currently this is configured to run only on the `main` branch since the job is slow. I think it's better to put slow jobs on `main` and act on build failures, but if we are finding that we're having a lot of crashers hit `main` then we could reconsider and add to PRs (and maybe spend some cycles speeding up this job).

What if we make the job run by default, but mark it as non-required for merging? Then for changes we're pretty sure won't have an impact, or that we're okay with catching later, we can merge optimistically, but for larger changes, we don't have to wait for a merge to main to find out about breakage.

Issue #464, opened 2022-03-02
Add IBC Packets to Penumbra transactions


hdevalence:
We need a way for relayers to add IBC messages to the Penumbra chain state, so they can be processed by Penumbra's IBC component. This mechanism should allow essentially arbitrary data packets, to be processed by the IBC component.

Our transactions are proto-based, and IBC is also proto-oriented, so this probably looks like having an additional transaction action, say `IBCPacket`, that contains a single IBC message; relayers can create transactions with arbitrarily many `IBCPackets` (subject to usual concerns about gas). 

Questions:

- [x] How do we represent the contents of an `IBCPacket`? Do we have plain bytes, or do we import the IBC protos and refer to those types in the Penumbra protos, or do we use some kind of proto "any type" / dynamic downcasting / etc?
- [x] How do IBC messages work with `CheckTx` vs `DeliverTx`? Presumably there's validity checking before block inclusion, but how do we actually implement this?  Do we need the mempool to have a copy of the IBC state machine to do "trial execution" of IBC messages?

avahowell:
this is completed with #477: we now have the IBCAction type that encapsulates IBC packets

PR #477, opened 2022-03-10
add IBCAction protobuf message, specify handling of IBC messages

avahowell:
This PR adds a new `IBCAction` proto, which includes all of the IBC messages required for basic IBC support (connections, channels, and packets) for penumbra. It also expands the spec to specify how we do basic IBC message encapsulation and vendors `ibc-go`'s protobufs into the `penumbra/proto` directory.

This should address the two main questions in #464 as follows:

* IBC packets are encapsulated by an `IBCAction` protobuf, which uses the Proto3 `oneof` directive to match it to one of the valid IBC message protos.


avahowell:
note that this does *NOT* add the IBCAction penumbra domain type or its handling in action.rs, just the IBC protos and the IBCAction proto.

hdevalence:
This looks good to me!

One thing I'm not sure about (non-blocking) is whether we want to be pulling all of the protos ourselves, or using some shared rust crates for cosmos-sdk proto types as a crate dependency.  If we did that, we might be able to save maintenance (?) but I'm not sure how we'd be able to compose protos together.

PR #480, opened 2022-03-10
Add support for uploading validators to pcli [WIP]

zbuc:


proto/src/lib.rs
```diff
@@ -77,7 +82,16 @@ pub mod sighash {
                 Some(TxAction::Output(o)) => Some(SHAction::Output(o)),
                 Some(TxAction::Delegate(d)) => Some(SHAction::Delegate(d)),
                 Some(TxAction::Undelegate(d)) => Some(SHAction::Undelegate(d)),
-                Some(TxAction::ValidatorDefinition(d)) => Some(SHAction::ValidatorDefinition(d)),
+                // The `ValidatorDefinition` contains sig bytes, so we only want to hash the `Validator` itself.
+                // Some(TxAction::ValidatorDefinition(v)) => Some(SHAction::Validator(v)),
```
hdevalence, line 86: Either way would be fine here, but also, we should do a proper sighash definition later, so that we don't need to do ad-hoc proto stuff for signing.

proto/src/lib.rs
```diff
@@ -23,6 +23,9 @@ pub mod serializers;
 mod protobuf;
 pub use protobuf::Protobuf;
 
+mod error;
+pub use error::ProtoError;
```
hdevalence, line 27: What's the motivation for pulling `ProtoError` into `penumbra_proto` from the `transaction` crate?  I'm not 100% sure, but my intuition is that we probably don't want it here (if we want it at all -- it might be better to remove it entirely), since the `ProtoError` has errors relating to the conversion between a proto type and a domain type, and the domain types don't live in the `proto` crate.

transaction/src/transaction/builder.rs
```diff
@@ -271,6 +284,9 @@ impl Builder {
         self.outputs.shuffle(rng);
         self.delegations.shuffle(rng);
         self.undelegations.shuffle(rng);
+        self.validator_definitions.shuffle(rng);
+
+        println!("Shuffled... {:#?}", self.validator_definitions);
```
hdevalence, line 289: nit: this should probably be a `tracing` invocation? (no worries if it's just for debugging in a WIP state, though)

transaction/src/transaction/builder.rs
```diff
@@ -193,6 +197,15 @@ impl Builder {
...
 
+    pub fn add_validator_definition(
+        &mut self,
+        validator: Validator,
+        signing_key: SigningKey<SpendAuth>,
+    ) -> &mut Self {
+        self.validator_definitions.push((validator, signing_key));
+        self
+    }
+
```
hdevalence, line 208: It looks like this is doing deferred signing, but I think that it might be easier and cleaner to do immediate signing, or even to just take a `ValidatorDefinition` and have the caller do the signing.

Unlike the case of a spend, where it's super important that the spend happen in the context of *a particular transaction* (i.e., the spend authorization should be bound also to the outputs, so you can't take my spends and put them in your transaction), the validator definitions are essentially context-free: they're signed by the validator's identity key, but that's it, and the purpose of the transaction is just to get the self-authenticating `ValidatorDefinition` to the chain.

If we had the code take a signed `ValidatorDefinition`, it'd also be possible to support offline setups mostly-free: the offline machine with the identity key signs the validator definition, it gets copied over to the machine with the hot wallet, then uploaded with no further intervention, etc.

wallet/src/state.rs
```diff
@@ -414,6 +416,86 @@ impl ClientState {
...
+                        amount: change,
+                        asset_id: denom.id(),
+                    },
+                    memo,
+                    self.wallet.outgoing_viewing_key(),
+                );
+
+                self.register_change(note);
+            }
+        }
```
hdevalence, line 488: I don't think we should try to change this in this PR (duplicating the code lets us get it working, and highlights opportunities for commonality), but in later work, we should probably figure out a way to get this kind of common logic to actually be common.

proto/src/lib.rs
zbuc: How do you mean? Is there an example in the codebase?

proto/src/lib.rs
zbuc: There was a circular dependency I was dealing with, but it's no longer necessary. Agreed about removing it -- there was a comment about just replacing it with `anyhow`. Is the idea that we don't want to use error enums at all in the codebase?

transaction/src/transaction/builder.rs
zbuc: Yeah the `println` statements are just for debugging, will remove in a polish up commit

wallet/src/state.rs
zbuc: Agreed -- it's also subtly different from the spend code, in that the spend code needs to take account of the spend action's output as well, and for validator definitions we only care about the fee. However it still should be possible to be genericized somehow.

proto/src/lib.rs
zbuc: > In this specific case, it's fine to have the full validator definition be part of the sighash data, because the validator definition signature is part of the transaction data to be signed, rather than itself being a spend authorization signature (the validator definition signature binds only to the definition, not to the whole transaction it's part of).

This is not actually true, right now the entire transaction is signed (and the verification checks that the validator's identity key made the signature), not just the validator definition itself

transaction/src/transaction/builder.rs
hdevalence: Yeah, the signature in the `ValidatorDefinition` has to only be a signature over the `Validator` itself, because the point is that the `ValidatorDefinition` is supposed to be a self-authenticating definition of some validator configuration, and we should be able to pass it around the program without needing to hold on to other context.  So we don't want to have the signature sign over other data (e.g., the rest of the transaction that broadcast it), because then we wouldn't be able to verify it as a standalone item.

But in that case, there's no need for deferred signing, since the `ValidatorDefinition` isn't bound to any of the other transaction data.

proto/src/lib.rs
hdevalence: Hmm, I think the signature in the `ValidatorDefinition` should only be signing the `Validator`, since otherwise we have no way to verify the signatures later: https://github.com/penumbra-zone/penumbra/pull/480/files#r827261810

hdevalence:
Included in #495

PR #481, opened 2022-03-10
docker: add log rotation for pd container in prod configuration

### LINKED ISSUE #437: Add log rotation / pruning to docker configuration
Currently, we don't do any automatic log rotation or pruning, so we generate potentially large amounts of log data.  This took down the Mneme node when it filled the disk with logs. Instead we should probably configure some kind of log data; the limit could be pretty big (e.g., 50GB), just some limit so we don't fill the disk.

redshiftzero:
Closes #437

### DIFF
--- a/docker-compose.prod.yml
+++ b/docker-compose.prod.yml
@@ -1,10 +1,5 @@
 version: "3.7"
-# docker-compose oerrides for local development.
-#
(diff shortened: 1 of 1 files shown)


hdevalence (approved):
LGTM

PR #483, opened 2022-03-11
expose IBC module in `proto`

avahowell:
This is a small PR that exposes the `ibc` proto types as part of the `proto` crate.

### DIFF
--- a/proto/Cargo.toml
+++ b/proto/Cargo.toml
@@ -14,7 +14,9 @@ hex = "0.4"
 anyhow = "1.0"
 subtle-encoding = "0.5"
 bech32 = "0.8"
+ibc-proto = { git = "https://github.com/penumbra-zone/ibc-rs" }
 
 [build-dependencies]
 prost-build = "0.9"
+ibc-proto = { git = "https://github.com/penumbra-zone/ibc-rs" }
 tonic-build = "0.6"
--- a/proto/Cargo.toml
+++ b/proto/Cargo.toml
@@ -14,9 +14,9 @@ hex = "0.4"
 anyhow = "1.0"
 subtle-encoding = "0.5"
 bech32 = "0.8"
-ibc-proto = { git = "https://github.com/penumbra-zone/ibc-rs" }
+ibc-proto = "0.12.0"
 
 [build-dependencies]
 prost-build = "0.9"
-ibc-proto = { git = "https://github.com/penumbra-zone/ibc-rs" }
+ibc-proto = "0.12.0"
 tonic-build = "0.6"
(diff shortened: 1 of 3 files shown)


proto/Cargo.toml
```diff
@@ -14,7 +14,9 @@ hex = "0.4"
 anyhow = "1.0"
 subtle-encoding = "0.5"
 bech32 = "0.8"
+ibc-proto = { git = "https://github.com/penumbra-zone/ibc-rs" }
```
hdevalence, line 17: What's the motivation for having our own fork here?

proto/build.rs
```diff
@@ -39,6 +39,7 @@ fn main() -> Result<()> {
     config.compile_protos(&["proto/stake.proto"], &["proto/"])?;
     config.compile_protos(&["proto/chain.proto"], &["proto/"])?;
     config.compile_protos(&["proto/genesis.proto"], &["proto/"])?;
+    config.extern_path(".ibc", "::ibc-proto::ibc");
```
hdevalence, line 42: I think you mentioned on Discord that this is related to some problem with import paths for the `ibc-rs` types, would it be possible to have a comment explaining what it's doing?

Side Q: wait, is `::ibc-proto::ibc` a Rust module path? are dashes allowed there?

hdevalence:
Looks like tihs is failing the build because the `ibc-rs` MSRV is 1.58, but we're currently using 1.56 ? Odd that the CI check passes but the smoke test doesn't, is the smoke test GH actions using a different rust version (cc @redshiftzero )?

proto/Cargo.toml
avahowell: Originally it's so that we could derive the `prost::Message` trait, which is required for the `Oneof` in the IBCAction proto, but in the latest ibc-rs version that appears to already exist. We can just use that, I don't see a reason to use our own fork at this point

proto/build.rs
avahowell: The codegen apparently converts the dash to an underscore

redshiftzero:
I see that this was resolved with @avahowell's latest push, but the smoke test GH action and the rust CI GH actions are both using `actions-rs/toolchain@v1` with the stable [toolchain](https://github.com/actions-rs/toolchain#inputs). However, it looks like the `rustc` version issue reported in the [failed smoke test build](https://github.com/penumbra-zone/penumbra/runs/5505524664?check_suite_focus=true) was due to the version of rustc specified (1.56.x) in the [base image](https://github.com/penumbra-zone/penumbra/blob/main/Dockerfile#L1) used for `pd`, so if we see a similar error we could bump up the base image version.

proto/build.rs
hdevalence: wild!

PR #490, opened 2022-03-15
Implement needed changes following refactor and asyncification of the JMT crate

### LINKED ISSUE #486: Update `jmt` crate to current `main` branch
The current `pd` code uses a proof-of-concept version of the `jmt` crate, on the `async-poc` branch.  The `main` branch has now been updated to support ICS23 proofs, and has a reworked API.

As a first step, we should swap out the `async-poc` branch with the new `main` branch, leaving the rest of the code unchanged.  This should be a relatively minor change, and the upside to doing (& merging) it first is that we'll start rendering the correct `jmt` version on our workspace Rustdoc output (so we can see the API).

aubrika:
Retarget `jmt main` from `penumbra` following asyncification. Closes #486.

### DIFF
--- a/pd/src/state/jellyfish.rs
+++ b/pd/src/state/jellyfish.rs
@@ -1,14 +1,11 @@
 use anyhow::Result;
 use futures::future::BoxFuture;
 use jmt::{
-    define_hasher,
-    hash::{CryptoHasher, DefaultHasher, HashValue},
-    node_type::{LeafNode, Node, NodeKey},
-    NodeBatch, TreeReaderAsync, TreeWriterAsync, Value,
+    storage::{LeafNode, Node, NodeBatch, NodeKey, TreeReader, TreeWriter},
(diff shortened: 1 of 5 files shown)


pd/src/state/jellyfish.rs
```diff
@@ -17,39 +14,26 @@ pub enum Key {
...
-                let mut state = NoteCommitmentAnchorHasher::default();
-                state.update(b"");
-                state.finish()
+                let mut state = sha2::Sha256::new();
+                state.update(b"nct");
+                KeyHash(*state.finalize().as_ref())
             }
         }
     }
 }
```
hdevalence, line 26: I think we can get rid of the `Key` type entirely here and just write the NCT data in under the key `b"nct"`?

hdevalence:
Merged now so that we get a rendering of the new API.

Issue #488, opened 2022-03-15
Refactor application logic into components.


hdevalence:
As described in #487, we'd ideally like to have the apphash bind to all of the application's persistent data.  After getting a RocksDB backing store working in #487, we should migrate the application state over to use a `jmt::WriteOverlay` on top of that store as the application's primary state.  This should dramatically simplify the logic in, e.g., the `ValidatorSet`, which can just read from and write to a `WriteOverlay` owned by the consensus worker, rather than having to manually save up pending writes in objects and implement custom read-through behavior.

Our initial development focused on building an MVP of the basic, multi-asset shielded pool functionality, which we then extended to support the mechanism design for shielded staking.  However, implementing the full staking state machine, dynamic validator set and all, showed that we'd reached the limits of this design.  At the same time, we're also reaching the limits of our (also MVP) Postgres-backed state system, which doesn't allow us to provide authentication proofs about any of the chain state.

To fix this, #500 introduces a new design based around the [`Component` trait](https://rustdoc.penumbra.zone/main/pd/trait.Component.html), which works roughly as follows.  The `Application` is a copy-on-write model of the ledger state and consensus rules, with some finalized portion (in the `Storage`) and some uncommitted portion (in a `jmt::WriteOverlay`).  The `Application` is structured a bundle of `Component`s, with a `jmt::WriteOverlay` shared among all of that `Application` instance's components: 
```ascii
┌────────────┐          ┌───────────┐                           
│WriteOverlay│          │ Component │                           
│  ::new()   │═════════▶│  ::new()  │      ═══▶ Execution Flow  
└────────────┘          └───────────┘           (Approximate)   
       │                      ║            ───▶ Data Flow       
       ▼               ╔══════╩══════╗                          
┌────────────┐         ▼             ║                          
│            │   ┌───────────┐       ║         ┌───────────────┐
│            │◀──│init_chain │◀──────╬─────────│ Genesis State │
│            │   └───────────┘       ║         └───────────────┘
│            │         ║             ▼                          
│            │         ║       ┌───────────┐   ┌───────────────┐
│            │◀────────╬──────▶│begin_block│◀──│ABCI BeginBlock│
│            │         ║       └───────────┘   └───────────────┘
│            │         ║             ║                          
│            │         ║    ╔═══════▶║                          
│            │         ║    ║        ▼                          
│            │         ║    ║  ┌───────────┐   ┌───────────────┐
[...]

PR #495, opened 2022-03-17
Properly calculate and store voting power

zbuc:


pcli/src/command/validator.rs
```diff
@@ -31,6 +52,43 @@ impl ValidatorCmd {
...
+            ValidatorCmd::UploadDefinition { file, fee, source } => {
+                // The definitions are stored in a JSON document,
+                // however for ease of use it's best for us to generate
+                // the signature here based on the configured wallet.
+                //
+                // TODO: eventually we'll probably want to support defining the
+                // identity key in the JSON file.
+                //
+                // We could also support defining multiple validators in a single
+                // file.
```
hdevalence, line 64: maybe we should have a command with some command-line options that would generate the definition?

pcli/src/command/validator.rs
```diff
@@ -1,24 +1,45 @@
...
+use anyhow::Context;
 use anyhow::Result;
-use penumbra_stake::IdentityKey;
-use penumbra_wallet::ClientState;
+use penumbra_stake::Validator;
+use penumbra_stake::ValidatorDefinition;
+use rand_core::OsRng;
 use structopt::StructOpt;
 
+use penumbra_proto::{stake::Validator as ProtoValidator, Message};
```
hdevalence, line 10: I think a slightly cleaner pattern here would be:
```
use penumbra_proto::{stake as pb, Message};
```
so that the protobuf type can be referred to as `pb::Validator`.

hdevalence: (this is non-blocking, though!)

pcli/src/command/validator.rs
```diff
@@ -31,6 +52,43 @@ impl ValidatorCmd {
...
+                let v_bytes = protobuf_serialized.encode_to_vec();
+                let signing_key = state.wallet().spend_key().spend_auth_key().clone();
+                let auth_sig = signing_key.sign(&mut OsRng, &v_bytes);
+                let vd = ValidatorDefinition {
+                    validator: new_validator,
+                    auth_sig,
+                };
+                // Construct a new transaction and include the validator definition.
+                // TODO: is it possible to get rid of this clone? It's only used because we can't
+                // borrow state mutably & immutably.
```
hdevalence, line 81:
```suggestion
```
Looks like the clone was already eliminated?

pd/src/components/validator_set.rs
```diff
@@ -246,21 +243,6 @@ impl BlockChanges {
...
-            VALUES ($1, $2, $3)
-            ON CONFLICT (asset_id) DO UPDATE SET denom=$2, total_supply=$3",
-                &id.to_bytes()[..],
-                asset.0.to_string(),
-                asset.1 as i64
-            )
-            .execute(&mut *dbtx)
-            .await?;
-        }
-
```
hdevalence, line 263: I didn't see where this logic moved to, so maybe we're not computing the asset balances correctly any more (?), but probably it's easier to fix that as part of the state refactoring, so we should probably merge this in as-is?

hdevalence (approved):
This LGTM, the comments I have are all non-blocking. I think we should merge these changes in and aim to do some cleanup when we move to a JMT-only design (#488).

PR #523, opened 2022-03-30
Staking component scaffolding

zbuc:
Optimistically merging prior to writing the rest of the implementation

### DIFF
--- a/pd/src/components/staking.rs
+++ b/pd/src/components/staking.rs
@@ -61,18 +61,18 @@ impl Component for Staking {
     }
 
     fn check_tx_stateless(_tx: &Transaction) -> Result<()> {
-        todo!()
+        Ok(())
     }
 
     async fn check_tx_stateful(&self, _tx: &Transaction) -> Result<()> {
-        todo!()
+        Ok(())
     }
 
     async fn execute_tx(&mut self, _tx: &Transaction) -> Result<()> {
-        todo!()
+        Ok(())
     }
 
     async fn end_block(&mut self, _end_block: &abci::request::EndBlock) -> Result<()> {
-        todo!()
+        Ok(())
     }
 }
--- a/pd/src/components/staking.rs
+++ b/pd/src/components/staking.rs
@@ -1,28 +1,62 @@
 use anyhow::Result;
 use async_trait::async_trait;
+use penumbra_chain::params::ChainParams;
+use penumbra_stake::Epoch;
 use penumbra_transaction::Transaction;
 use tendermint::abci;
 
 use super::{Component, Overlay};
-use crate::genesis;
+use crate::{components::validator_set::BlockChanges, genesis};
 
 // Stub component
 pub struct Staking {
     overlay: Overlay,
+
+    // TODO: this shouldn't be here -- we should grab this from a common location
+    // in the JMT, i.e. genesis/AppState but that's not available until #511 lands.
+    // since consensus can't yet change chain params on the fly, it's not an issue,
+    // but it definitely will be in the future!
+    chain_params: Option<ChainParams>,
 }
 
 #[async_trait]
 impl Component for Staking {
     async fn new(overlay: Overlay) -> Result<Self> {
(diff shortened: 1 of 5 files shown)


pd/src/components/staking.rs
```diff
@@ -0,0 +1,78 @@
...
+
+// Stub component
+pub struct Staking {
+    overlay: Overlay,
+
+    // TODO: this shouldn't be here -- we should grab this from a common location
+    // in the JMT, i.e. genesis/AppState but that's not available until #511 lands.
+    // since consensus can't yet change chain params on the fly, it's not an issue,
+    // but it definitely will be in the future!
+    chain_params: Option<ChainParams>,
```
hdevalence, line 19: Non-blocking:

Does this need to be here (i.e., in-memory, as a field on `Staking`) at all?  It looks like the only place it's read from is when getting the epoch duration.

But alternatively, we could have the epoch duration written into the JMT somewhere and have an `epoch_duration` method on `Staking` that would pull it out of the `Overlay` whenever it was needed.

There's a question of how we might want to organize code for reading / writing data that would be used by multiple components (e.g., the epoch duration).  We should break out discussion on this, but I think making another extension trait could be a good option.

pd/src/components/staking.rs
```diff
@@ -0,0 +1,78 @@
+use anyhow::Result;
+use async_trait::async_trait;
+use penumbra_chain::params::ChainParams;
+use penumbra_stake::Epoch;
+use penumbra_transaction::Transaction;
+use tendermint::abci;
+
+use super::{Component, Overlay};
+use crate::{components::validator_set::BlockChanges, genesis, WriteOverlayExt};
```
hdevalence, line 9: I think it'd be better to firewall off this code from the previous `validator_set` code, and just copy things over as needed, so that when we're done, we can cleanly delete the old code.  In this case, for instance, I think the design of the `WriteOverlay` will probably mean we don't need the `BlockChanges` at all.

pd/src/components/staking.rs
zbuc: Yeah, I'm removing this as part of the implementation and only fetching it from the JMT (and for now it'll be set as part of the `App` as discussed)

pd/src/components/staking.rs
```diff
@@ -0,0 +1,78 @@
...
+        todo!()
+    }
+
+    async fn execute_tx(&mut self, _tx: &Transaction) -> Result<()> {
+        todo!()
+    }
+
+    async fn end_block(&mut self, _end_block: &abci::request::EndBlock) -> Result<()> {
+        todo!()
+    }
```
hdevalence, line 77: These `todo!()`s will cause a panic in the whole app when we run it as a sidecar (as discovered in #518), so it'd be better to replace them with an `Ok(())` so that we pass through the incomplete implementation until we're done writing it.

stake/src/epoch.rs
```diff
@@ -1,8 +1,9 @@
+use serde::{Deserialize, Serialize};
 use tendermint::block;
 
 /// Epoch represents a given epoch for Penumbra and is used
 /// for calculation of staking exchange rates.
-#[derive(Debug, Eq, PartialEq, Copy, Clone)]
+#[derive(Debug, Eq, PartialEq, Copy, Clone, Serialize, Deserialize)]
```
hdevalence, line 6: We should be careful to use the `try_from` / `into` Serde annotations to make sure that we always serialize *through* the protobuf types, and don't end up accidentally defining a serialization format from our internal implementation details.

pd/src/components/validator_set.rs
```diff
@@ -251,7 +253,7 @@ impl BlockChanges {
     }
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Serialize, Deserialize)]
```
hdevalence, line 256: same comment as https://github.com/penumbra-zone/penumbra/pull/523/files#r838918448

PR #533, opened 2022-03-31
store the current block height and block timestamp in app state

avahowell:
This PR adds some methods to `PenumbraStore` and to App::begin_block that stores the block height and block timestamp in the application state, so that components (such as the IBC component) can access them

### DIFF
--- a/pd/src/storage/penumbra_store.rs
+++ b/pd/src/storage/penumbra_store.rs
@@ -1,6 +1,9 @@
+use std::str::FromStr;
+
 use anyhow::{anyhow, Result};
 use async_trait::async_trait;
 use penumbra_chain::params::ChainParams;
+use tendermint::Time;
 
 use crate::WriteOverlayExt;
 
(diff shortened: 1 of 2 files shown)


pd/src/storage/penumbra_store.rs
```diff
@@ -32,6 +35,37 @@ pub trait PenumbraStore: WriteOverlayExt {
...
+            .ok_or_else(|| anyhow!("Missing block_timestamp"))?;
+
+        Ok(Time::from_str(&timestamp_string).unwrap())
+    }
+
+    /// Writes the block timestamp to the JMT
+    async fn put_block_timestamp(&self, timestamp: Time) {
+        self.put_proto(b"block_timestamp".into(), timestamp.to_rfc3339())
+            .await
+    }
```
hdevalence, line 68: Do we want only the most recent block timestamp to be recorded, or every block's timestamp?

pd/src/storage/penumbra_store.rs
```diff
@@ -32,6 +35,37 @@ pub trait PenumbraStore: WriteOverlayExt {
...
+            .get_proto(b"block_height".into())
+            .await?
+            .ok_or_else(|| anyhow!("Missing block_height"))?;
+
+        Ok(height_bytes)
+    }
+
+    /// Writes the block height to the JMT
+    async fn put_block_height(&self, height: u64) {
+        self.put_proto(b"block_height".into(), height).await
```
hdevalence, line 51: Wait, why does it work to use a raw u64 as a proto type? What's happening here?

PR #534, opened 2022-03-31
Commit App state following init_chain

hdevalence:
The initial sidecar use of the `App` code didn't call `commit` following `init_chain`, so the genesis state was never actually committed.

However, trying to commit that state causes new problems, which are probably related to the JMT's handling of the empty tree:
```
Mar 31 14:55:03.809 DEBUG abci:InitChain{chain_id="penumbra-orthosie-23a3e8b8"}:commit: pd::storage: node_key=NodeKey { version: 0, nibble_path:  } value=None
thread 'tokio-runtime-worker' panicked at 'called `Option::unwrap()` on a `None` value', /Users/hdevalence/.cargo/git/checkouts/jellyfish-merkle-6b18a07b8fd5fec6/445c51a/src/tree_cache.rs:188:58
```

### DIFF
--- a/pd/src/consensus/worker.rs
+++ b/pd/src/consensus/worker.rs
@@ -168,6 +168,8 @@ impl Worker {
 
         // Begin new sidecar code
         self.app.init_chain(&app_state).await?;
+        // Note: App::commit resets internal components, so we don't need to do that ourselves.
+        self.app.commit(self.storage.clone()).await?;
         // End new sidecar code
 
         // Initialize the database with the app state.
--- a/pd/src/consensus/worker.rs
+++ b/pd/src/consensus/worker.rs
@@ -167,6 +167,17 @@ impl Worker {
             .expect("can parse app_state in genesis file");
 
         // Begin new sidecar code
+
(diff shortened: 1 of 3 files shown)


pd/src/consensus/worker.rs
```diff
@@ -167,7 +167,20 @@ impl Worker {
...
 
         // Begin new sidecar code
+
+        // We want to write the genesis state as version 0 of the tree, so we
+        // need to initialize the write overlay with the PRE_GENESIS_VERSION.
+        // To do this, it's easiest to just overwrite the app here.
+        self.app = App::new(Arc::new(Mutex::new(WriteOverlay::new(
+            self.storage.clone(),
+            // Some kind of visibility issue on associated consts?
+            //WriteOverlay::PRE_GENESIS_VERSION,
```
hdevalence, line 177: I'm not sure why, but using `WriteOverlay::PRE_GENESIS_VERSION` here doesn't work, because rustc thinks that the associated constant isn't there.  For now, this just puts in the value directly.

PR #535, opened 2022-03-31
ibc client: implement handling of CreateClient IBC message

avahowell:
this PR adds support to the IBC component for handling incoming CreateClient IBC messages.

### DIFF
--- a/pd/src/components/ibc.rs
+++ b/pd/src/components/ibc.rs
@@ -1,5 +1,6 @@
 use anyhow::Result;
 use async_trait::async_trait;
+use penumbra_ibc::IBCAction;
 use penumbra_transaction::{Action, Transaction};
 use serde::{Deserialize, Serialize};
 use std::convert::TryFrom;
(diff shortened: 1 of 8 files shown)


pd/src/components/ibc.rs
```diff
@@ -32,11 +50,98 @@ impl Component for IBCComponent {
...
+    async fn execute_tx(&mut self, tx: &Transaction) -> Result<()> {
+        // handle client transactions
+        for action in tx.transaction_body.actions.iter() {
+            let ibc_action = match action {
+                Action::IBCAction(ibc_action) => Some(ibc_action),
+                _ => None,
+            };
+            if ibc_action.is_none() {
+                continue;
+            }
```
hdevalence, line 62: The control flow here could be cleaned up with `filter_map`, by doing something like
```rust
        for action in tx.transaction_body.actions.iter()
            .filter_map(|action| match action {
                Action::IBCAction(ibc_action) => Some(ibc_action),
                _ => None,
            })
        {
            // ...
        }
```
so that there's no need for further filtering or unwrapping.

pd/src/components/ibc.rs
```diff
@@ -32,11 +50,98 @@ impl Component for IBCComponent {
...
+
+            match &ibc_action.unwrap().action {
```
hdevalence, line 64: Pre-filtering avoids the need for this `unwrap`, since the `action` would already be an IBC action

pd/src/components/ibc.rs
```diff
@@ -32,11 +50,98 @@ impl Component for IBCComponent {
...
+
+                    // get the current client counter
+                    let id_counter = self.client_counter().await?;
+                    let client_id =
+                        ClientId::new(msg_create_client.client_state.client_type(), id_counter)?;
+
+                    tracing::info!("creating client {:?}", client_id);
+
+                    self.store_new_client(client_id, msg_create_client).await?;
+                }
```
hdevalence, line 90: What's the best way to factor this code out, so that we don't end up with all of the IBC state machine directly in the body of `execute_tx`?

pd/src/components/ibc.rs
```diff
@@ -32,11 +50,98 @@ impl Component for IBCComponent {
...
+
+        todo!()
     }
 
     async fn end_block(&mut self, _end_block: &abci::request::EndBlock) -> Result<()> {
         Ok(())
     }
 }
+
+#[derive(Serialize)]
```
hdevalence, line 103: We shouldn't use `#[derive(Serialize)]` on Rust types that we're putting in our data structures, since it means that we can't change the Rust types without breaking the data format.  Instead, we should either (a) define our own proto + domain type or (b) use an existing proto + domain type (not sure if there is one for `ClientData`), so that we can use the `get_domain`/`put_domain` methods.

hdevalence (changes requested):

avahowell:
updated this by refactoring the handling of ibc actions into its own fn, implemented the new View trait, and added domain types for ClientData and ClientCounter.

hdevalence (approved):

PR #538, opened 2022-04-01
First pass at documentation for running fullnodes

zbuc:
Some things aren't ready yet, specifically:

1. Specifying the persistent peers requires the node ID, which is determined by the public key of the node. This will need to be grabbed and filled in when we deploy the next testnet (and we should probably use the same key between testnet deployments to avoid needing to update the value in the documentation)
2. The `genesis.json` file is not available for the newest testnet deployment. We'll want to have these available at a constant well-known URL.

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,6 +159,100 @@ the `#testnet-feedback` channel on our [Discord]. We would love to know about bu
 confusing error messages, or any of the many other things that inevitably won't quite work yet. Have
 fun! :)
 
+## Connecting a node to a testnet
+
+A fullnode can be joined to our testnet deployments and optionally behave as a validator.
+
+The node binary (`pd`) software is part of the same repository as `pcli`, so ensure that you've [cloned the repository](https://github.com/penumbra-zone/penumbra#cloning-the-repository).
+
+### Building the `pd` node binary
+
+Then, build the `pd` binary using `cargo`:
+
+```bash
+cargo build --release --bin pd
+```
+
+Because you are building a work-in-progress version of the node, you may see compilation warnings,
+which you can safely ignore.
+
+### Installing tendermint
+
+You'll need to have [tendermint installed](https://docs.tendermint.com/master/introduction/install.html) on your system to join your node to the testnet.
+
+### Setting up postgres
+
+Next, you'll need to set up a Postgres instance. Here is one way:
+```console
+$ docker volume create tmp_postgres_data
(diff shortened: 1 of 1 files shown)


hdevalence:
On (2), maybe we should just tell people to grab the latest `genesis.json` file from the `testnets` directory in the repo?

README.md
```diff
@@ -159,6 +159,100 @@ the `#testnet-feedback` channel on our [Discord]. We would love to know about bu
...
+A fullnode can be joined to our testnet deployments and optionally behave as a validator.
+
+The node binary (`pd`) software is part of the same repository as `pcli`, so ensure that you've [cloned the repository](https://github.com/penumbra-zone/penumbra#cloning-the-repository).
+
+### Building the `pd` node binary
+
+Then, build the `pd` binary using `cargo`:
+
+```bash
+cargo build --release --bin pd
```
hdevalence, line 173: we could add `--quiet` to suppress warnings here

README.md
```diff
@@ -159,6 +159,100 @@ the `#testnet-feedback` channel on our [Discord]. We would love to know about bu
...
+```
+
+Because you are building a work-in-progress version of the node, you may see compilation warnings,
+which you can safely ignore.
```
hdevalence, line 177: 👍

README.md
```diff
@@ -159,6 +159,100 @@ the `#testnet-feedback` channel on our [Discord]. We would love to know about bu
...
+config	data
+
+$ ls $HOME/.tendermint/config/
+config.toml	genesis.json	node_key.json
+```
+
+Next you'll need to modify the persistent peers list to specify our primary testnet node as a permanent peer for your node. Open `$HOME/.tendermint/config/config.toml` and find the `persistent-peers` line and make it look like so:
+
+```toml
+persistent-peers = "TODO_IDENTIFIER_GOES_HERE@192.167.10.21:26656"
```
hdevalence, line 227: is there a way to skip the identifier here, and just do TOFU for whatever that IP gives out?

hdevalence (approved):
LGTM, I think we can fill in the remaining gaps once we stand up the network and check the instructions against the actually running version.

README.md
zbuc: I'll play around with it but my recollection from the local multinode testnet is that it didn't work without explicitly defining the identifier here :(

PR #544, opened 2022-04-03
pd: change RPC to only provide the next epoch's RateData

hdevalence:
To do staking transactions, we only need the *next* epoch's rate data, not any
historical rate data.  Restricting the RPC to only what we actually need will
make porting the staking component much easier.


hdevalence:
cc @zbuc this means that when we use the JMT as the backing state store in the `Component` rewrite, we only need to record `current_rate` and `next_rate` in the tree to be able to implement our RPCs

Issue #507, opened 2022-03-23
Implement `Staking` logic as a `Component`


hdevalence:
As part of #488, we want to implement a second version of the staking logic (currently in `components/validator_set.rs`, though it is not a component) as a `Component`.

PR #550, opened 2022-04-04
pd: add ShieldedPoolStore extension trait

hdevalence:
This allows other components to read relevant data stored by the ShieldedPool component, initially just the token supply.

This is needed for #507, which needs to be able to read the delegation token supplies.

### DIFF
--- a/pd/src/components/shielded_pool.rs
+++ b/pd/src/components/shielded_pool.rs
@@ -23,9 +23,6 @@ use crate::{genesis, WriteOverlayExt};
 pub struct ShieldedPool {
     overlay: Overlay,
     note_commitment_tree: NoteCommitmentTree,
-    // TODO: change the on-chain registry to just store asset id -> amount
-    // and asset id -> denom separately
-    new_denoms: BTreeMap<asset::Id, Denom>,
     /// The in-progress CompactBlock representation of the ShieldedPool changes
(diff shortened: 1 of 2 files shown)


pd/src/components/shielded_pool.rs
```diff
@@ -321,3 +321,54 @@ impl ShieldedPool {
...
+    async fn update_token_supply(&self, asset_id: &asset::Id, change: i64) -> Result<()> {
+        let key = format!("shielded_pool/assets/{}/token_supply", asset_id).into();
+        let current_supply = match self.get_proto(key).await {
+            Ok(Some(value)) => value,
+            Ok(None) => 0u64,
+            // We want to handle the MissingRootError specially here, so that we can
+            // use 0 as a default value if the tree is empty, without ignoring all errors.
+            Err(e) if e.downcast_ref::<jmt::MissingRootError>().is_some() => 0u64,
+            Err(e) => return Err(e),
+        };
```
hdevalence, line 345: cc @aubrika ; IIRC we had a discussion about using downcasting inside of the `WriteOverlay` in the JMT crate.

 I think it'd be much cleaner if the `WriteOverlay` did the downcasting itself, and always replaced `Err(MissingRootError)` with `Ok(None)`, so we could get this behavior everywhere.

PR #560, opened 2022-04-05
Implement slashing in staking component

zbuc:


pd/src/components/staking.rs
```diff
@@ -130,6 +131,9 @@ impl Component for Staking {
             .set_base_rates(cur_base_rate, next_base_rate)
             .await;
 
+        // All genesis validators start in the "Active" state:
+        let state = ValidatorState::Active;
```
hdevalence, line 135: nit: I think it'd be clearer if, rather than having a variable here, we just passed `ValidatorSet::Active` into `save_validator` at 165 with this comment, so that when reading the `save_validator` invocation, there's no need to refer back here

pd/src/components/staking.rs
```diff
@@ -280,7 +285,7 @@ impl Component for Staking {
 ///
 /// TODO: should this be split into Read and Write traits?
 #[async_trait]
-pub trait View: WriteOverlayExt + Send + Sync {
+pub trait View: WriteOverlayExt + Send + Sync + Sized {
```
hdevalence, line 288: Hmm, maybe we should move the `Send + Sync + Sized` bounds to the `WriteOverlayExt` trait, so that they're implied for every `View`?

zbuc: That sounds like a good plan

pd/src/components/staking.rs
```diff
@@ -354,16 +369,59 @@ pub trait View: WriteOverlayExt + Send + Sync {
...
+            })?;
+
+        cur_rate = cur_rate.slash(slashing_penalty);
+
+        // TODO: would it be better to call `current_base_rate.next`? the same logic exists
+        // within there, but it requires passing in the current base rates & funding streams,
+        // which aren't actually used because the rate is held constant. So, doing it this way
+        // avoids a couple unnecessary JMT reads that the `current_base_rate.next` API would require.
+        //
+        // At any rate, the next rate is held constant for slashed validators.
```
hdevalence, line 419: I think it's better not to call `next` here, since I think that will evolve the rates through time, but we know that the next rate should be constant.

zbuc: The implementation of `next` is state-aware and will hold a constant rate for a slashed validator, since `next` is what's called for all validators during `end_epoch`.

stake/src/status.rs
```diff
@@ -27,34 +27,48 @@ impl From<ValidatorStatus> for pb::ValidatorStatus {
...
-                ValidatorState::Active => pb::validator_status::ValidatorState::Active,
-                ValidatorState::Unbonding { .. } => pb::validator_status::ValidatorState::Unbonding,
-                ValidatorState::Slashed => pb::validator_status::ValidatorState::Slashed,
-            } as i32,
-            unbonding_epoch: match v.state {
-                ValidatorState::Unbonding { unbonding_epoch } => Some(unbonding_epoch),
-                _ => None,
-            },
+            // TODO: Why is this becoming an Option type? the protobuf type
+            // doesn't define it as optional, and it should always be present.
```
hdevalence, line 31: This is becoming a Rust option because all proto fields are optional; primitive types have default values if missing, but proto messages are either there or not, so the Rust types get filled up with `Option`s.  I think the `optional` field in proto3 just records extra info that lets the application determine whether a primitive type was really missing or whether it was just a default value.

stake/src/validator_state.rs
```diff
@@ -16,8 +21,51 @@ pub enum ValidatorState {
...
+                pb::validator_state::ValidatorStateEnum::Slashed => ValidatorState::Slashed,
+            },
+        )
+    }
+}
+
 /// The name of a validator state, as a "C-style enum" without the extra information such as the
 /// `unbonding_epoch`.
+/// TODO: is this necessary now? Is `ValidatorStatus` necessary or are we decomposing into different
+/// paths in the JMT?
```
hdevalence, line 68: idk? it seems possibly useful to bundle all the state data into one path, so some client can authenticate a proof of inclusion for whatever the state is

hdevalence (approved):
looks great, all my comments are non-blocking!

pd/src/components/staking.rs
hdevalence: Ah, great, never mind then!  I think just bumping up the epoch here is simpler (from the reader's point of view) than having to fetch a bunch of extra data.

stake/src/validator_state.rs
zbuc: The `ValidatorStatus` struct also includes the voting power (and another copy of the identity key). It's not clear to me yet how we'll handle voting power in the staking component yet. It might make more sense to serialize the entire `ValidatorStatus` to a single path rather than splitting power and state, idk, I'll figure it out when I get there

PR #561, opened 2022-04-06
Update next rates on epoch transition in Staking component

zbuc:
It was unclear how to handle reward notes right now so those aren't working. The implementation also had to change a fair bit from the original, so a close review would be appreciated.

### DIFF
--- a/pd/src/components/staking.rs
+++ b/pd/src/components/staking.rs
@@ -239,7 +239,7 @@ impl Staking {
             identity_key: IdentityKey,
             power: u64,
             state: ValidatorState,
-        };
+        }
 
         let mut validator_power_list = Vec::new();
         for v in self.overlay.validator_list().await?.iter() {
@@ -329,7 +329,7 @@ impl Staking {
                 .ok_or_else(|| anyhow::anyhow!("validator missing power"))?;
             let validator = self
                 .overlay
-                .validator(&v)
+                .validator(v)
                 .await?
                 .ok_or_else(|| anyhow::anyhow!("validator missing"))?;
             updates.push(ValidatorUpdate {
@@ -406,7 +406,7 @@ impl Component for Staking {
                     // All genesis validators start in the "Active" state:
                     ValidatorState::Active,
                 )
-                .await;
+                .await?;
         }
 
         // Finally, record that there were no delegations in this block, so the data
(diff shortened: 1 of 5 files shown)


pd/src/components/staking.rs
```diff
@@ -29,6 +30,14 @@ pub struct Staking {
...
     /// epoch.
     delegation_changes: DelegationChanges,
+    /// New validator definitions accumulated over the course of this block, to be
+    /// persisted at the end of the block for processing at the end of the next
+    /// epoch.
+    new_validator_definitions: ValidatorDefinitions,
+    /// Updates to existing validator definitions accumulated over the course of this block, to be
+    /// persisted at the end of the block for processing at the end of the next
+    /// epoch.
+    updated_validator_definitions: ValidatorDefinitions,
```
hdevalence, line 40: Do we need to accumulate these over the course of the block, and only add them in at the end?

If we could apply them to the tree state immediately, I think a bunch of our logic would get simpler, because we wouldn't need to, e.g., resolve conflicting validator definitions (we'd just apply the first one, and reject the second one), etc.

There's a comment further down at line 684:
```
        // The new validator definitions can be applied now. We didn't apply them earlier because
        // we don't want rate calculations to apply to newly defined validators.
```
but I'm not sure why we don't want this?

pd/src/components/staking.rs
```diff
@@ -49,16 +58,31 @@ impl Staking {
...
+            .into_iter()
+            .map(|(k, v)| (k, v.collect()))
+            .collect();
+        let undelegations_by_validator: BTreeMap<IdentityKey, Vec<Undelegate>> = total_changes
+            .undelegations
+            .into_iter()
+            .group_by(|u| u.validator_identity.clone())
+            .into_iter()
+            .map(|(k, v)| (k, v.collect()))
+            .collect();
```
hdevalence, line 76: I'm not sure this code is correct; looking at the [`group_by`](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.group_by) docs, it groups *consecutive* runs of the same key, but we need to collect *all* the (un)delegations with that key, and that's not going to be possible without allocating (which the `group_by`) doesn't do.  I think this is going to drop all of the (un)delegations except the last consecutive runs by identity key.

Instead I think it'd be better to iterate over the (un)delegations directly, and then manually push into the right place with the Entry API, like (sketch):
```rust
let mut delegations_by_validator = BTreeMap::<IdentityKey, Vec<Delegate>::new();
for d in total_changes.delegations {
    delegations_by_validator.entry(d.validator_identity.clone())
        .or_default()
        .push(d);
}
```

pd/src/components/staking.rs
```diff
@@ -69,33 +93,252 @@ impl Staking {
...
+                anyhow::anyhow!("validator had ID in validator_list but not found in JMT")
+            })?;
+            // The old epoch's "current rate" is going to become the "previous rate".
+            let prev_rate = self
+                .overlay
+                .current_validator_rate(v)
+                .await?
+                .ok_or_else(|| {
+                    anyhow::anyhow!("validator had ID in validator_list but rate not found in JMT")
+                })?;
```
hdevalence, line 113: It looks like we don't use this rate at all, except for an assert later.  Could we skip loading it entirely?

pd/src/components/staking.rs
hdevalence: Per discussion in discord, let's move this to a future PR.

PR #569, opened 2022-04-07
Handle validator definition transactions in real time rather than batching

### LINKED ISSUE #566: Simplify `Staking` component logic by applying validator definitions immediately
> Do we need to accumulate these over the course of the block, and only add them in at the end?
>
> If we could apply them to the tree state immediately, I think a bunch of our logic would get simpler, because we wouldn't need to, e.g., resolve conflicting validator definitions (we'd just apply the first one, and reject the second one), etc.
>
> There's a comment further down at line 684:
> ```
>         // The new validator definitions can be applied now. We didn't apply them earlier because
>         // we don't want rate calculations to apply to newly defined validators.
> ```
> but I'm not sure why we don't want this?

_Originally posted by @hdevalence in https://github.com/penumbra-zone/penumbra/pull/561#discussion_r844312899_

zbuc:
Fixes #566


hdevalence:
@zbuc looks like this deleted the orthosie allocations by accident (?)

pd/src/components/staking.rs
```diff
@@ -607,12 +589,10 @@ impl Component for Staking {
...
             } else {
                 // This is a new validator definition.
-                validator_definitions.push(v.clone());
                 continue;
             }
 
-            // the validator definition has now passed all verification checks, so add it to the list
-            validator_definitions.push(v.clone());
+            // the validator definition has now passed all verification checks
         }
```
hdevalence, line 596: Hmm, what's going on here? From the warning it looks like we're iterating over an empty list?

pd/src/components/staking.rs
```diff
@@ -634,11 +614,51 @@ impl Component for Staking {
...
-            .definitions
-            .append(&mut definitions);
+        // The validator definitions have been completely verified, so we can add them to the JMT
+        let definitions = tx.validator_definitions().map(|v| v.to_owned());
+        let cur_epoch = self.overlay.get_current_epoch().await?;
+
+        for v in definitions {
+            let existing_v = self.overlay.validator(&v.validator.identity_key).await?;
+
+            if let Some(_existing_v) = existing_v {
```
hdevalence, line 624: tioli: I think this could be combined into
```rust
if self.overlay.validator(&v.validator.identity_key).await?.is_some() {
    // This is an existing validator....
} else {
    // This is a new validator ...
}
```

pd/src/components/staking.rs
```diff
@@ -669,40 +683,6 @@ impl Component for Staking {
         // The epoch has now transitioned, if necessary
         let cur_epoch = self.overlay.get_current_epoch().await?;
```
hdevalence, line 684: Looks like this could be pruned as well?

zbuc:
> @zbuc looks like this deleted the orthosie allocations by accident (?)

Oops, I had removed a bunch of them to make the testing feedback loop tighter, will fix

pd/src/components/staking.rs
zbuc: Which warning? I don't see one locally

hdevalence: never mind, it was a stale comment

hdevalence (approved):

PR #570, opened 2022-04-07
Remove ValidatorPower struct, rework genesis

### LINKED ISSUE #564: Remove `genesis::ValidatorPower`

Currently, we use a `ValidatorPower` structure in the genesis state, which specifies a validator power. But this gets overridden at the first epoch by the delegation tokens, and we already have to create delegation token allocations to avoid a chain halt there, so the `ValidatorPower` serves no purpose.

Remove it and compute powers from the delegation token supplies.

zbuc:
Fixes #564


pd/src/components/staking.rs
```diff
@@ -361,14 +361,26 @@ impl Component for Staking {
...
-            .set_base_rates(cur_base_rate, next_base_rate)
+            .set_base_rates(cur_base_rate.clone(), next_base_rate)
             .await;
 
+        let mut allocations_by_validator = HashMap::new();
+        for allocation in &app_state.allocations {
+            if allocation.amount == 0 {
+                continue;
+            }
+
```
hdevalence, line 372: non-blocking: this could be cut out, since we'll just be adding 0 to the total, so it's a no-op

hdevalence (approved):
LGTM!

PR #571, opened 2022-04-07
Changes to RPC to allow backing by the JMT

### LINKED ISSUE #554: JMT-backed RPC implementation
As part of #488, we need to have a new implementation of our RPC interfaces backed by the JMT.

This is also a good opportunity to clean up some of the RPC methods (e.g., the "light wallet"/"thin wallet" naming, which is horrible).

aubrika:
Closes #554


aubrika:
Questions:

1. What kind of logging do we want around these? Should I try to emulate the existing logging using what's available in the request?
2. How do we want to map the errors from the tonic results? I'm not confident the mappings I migrated over are still ideal.

pd/src/info/light_wallet.rs
```diff
@@ -0,0 +1,60 @@
...
+
+use crate::WriteOverlayExt;
+
+struct WalletOverlay<T: WriteOverlayExt>(T);
+
+#[tonic::async_trait]
+impl<T> LightWallet for WalletOverlay<T>
+where
+    T: 'static + WriteOverlayExt + Send + Sync,
+{
```
hdevalence, line 21: I think we should be able to avoid having to have the wrapper struct by doing something like
```
impl<T: WriteOverlayExt> LightWallet for T
```
The `Send + Sync` bounds are already implied by `WriteOverlayExt`, and if we need it to be `'static` here, we could add a `'static` bound to `WriteOverlayExt` so that it's available for every user

aubrika: I tried this originally & it was in violation of the orphan rules 🤔

hdevalence:
>     1. What kind of logging do we want around these? Should I try to emulate the existing logging using what's available in the request?

I think this isn't essential to get a final answer on now, we can iterate on the tracing output as we like once we have the functionality replicated.

>     2. How do we want to map the errors from the tonic results? I'm not confident the mappings I migrated over are still ideal.

Hmm, do you mean mapping our internal errors into tonic (gRPC) errors or mapping gRPC errors from tonic into our errors (which I don't *think* we need to do?) ?  In the first case, I think this also isn't super essential to get perfectly, as long as we have a reasonably sane mapping.

pd/src/info/light_wallet.rs
hdevalence: ah boo

hdevalence: i'm surprised because `WriteOverlayExt` is defined in-crate. but in that case, maybe it would be better to have one wrapper struct instead of two, and lift it into the parent `info` module as an `RpcWrapper`, that we can then add whatever extra extension traits to

hdevalence:
Picked this up since @aubrika is out today.  There's two parts to the changes: changes to the RPC itself, to allow the new implementation, and changes to add a new, JMT-backed implementation. 

I tested locally that the changes to the RPC preserve basic functionality (scanning, tx submission, delegation, undelegation, ...), but this branch doesn't (yet) switch over to the new RPC.

pd/src/wallet/server/thin.rs
```diff
@@ -1,91 +1,23 @@
...
-                    tx.send(Ok(asset.clone())).await.unwrap();
-                }
-            }
-            .instrument(Span::current()),
-        );
-
-        Ok(tonic::Response::new(Self::AssetListStream::new(rx)))
+        _request: tonic::Request<NoteCommitment>,
+    ) -> Result<tonic::Response<NoteSource>, Status> {
+        unimplemented!()
```
hdevalence, line 20: I left this unimplemented since we don't use it and we'll be deleting this code.

chain/src/known_assets.rs
```diff
@@ -0,0 +1,34 @@
+use penumbra_crypto::asset::Asset;
+use penumbra_proto::{chain as pb, Protobuf};
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Debug, Default, Serialize, Deserialize)]
+#[serde(try_from = "pb::KnownAssets", into = "pb::KnownAssets")]
+pub struct KnownAssets(pub Vec<Asset>);
```
hdevalence, line 7: We need a newtype so that we can protoize it in the JMT

chain/src/note_source.rs
```diff
@@ -50,3 +53,22 @@ impl TryFrom<[u8; 32]> for NoteSource {
         }
     }
 }
+
+impl Protobuf<pb::NoteSource> for NoteSource {}
```
hdevalence, line 57: Protoizing this saves some ad-hoc byte manipulation.

crypto/src/asset.rs
```diff
@@ -10,6 +13,40 @@ pub use denom::{Denom, Unit};
 pub use id::Id;
 pub use registry::{Registry, REGISTRY};
 
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(try_from = "pb::Asset", into = "pb::Asset")]
+pub struct Asset {
+    pub id: Id,
+    pub denom: Denom,
+}
```
hdevalence, line 21: this also exists so we can protoize it into the JMT

pd/src/components/shielded_pool.rs
```diff
@@ -408,48 +407,57 @@ pub trait View: WriteOverlayExt {
...
+            // ... and we want to record it in the list of known asset IDs
+            // (this requires reading the whole list, which is sad, but hopefully
+            // we don't do this often).
+            let mut known_assets = self.known_assets().await?;
+            known_assets.0.push(Asset {
+                id,
+                denom: denom.clone(),
+            });
+            self.put_domain("shielded_pool/known_assets".into(), known_assets)
+                .await;
```
hdevalence, line 445: without writing a list of all of the known asset IDs, we can't iterate over the denoms, and if we're writing a list anyways, we might as well bundle in the denoms.

pd/src/info/light_wallet.rs
```diff
@@ -0,0 +1,160 @@
...
+            .map_err(|_| tonic::Status::unavailable("database error"))?;
+
+        let _show_inactive = request.get_ref().show_inactive;
+        let store = self.0.clone();
+        let s = try_stream! {
+            for validator in validators {
+                let info = store.validator_info(&validator)
+                    .await?
+                    .expect("known validator must be present");
+                // TODO: filter by show_inactive
```
hdevalence, line 88: Currently, the implementation doesn't respect the `show_inactive` filter, but this doesn't affect the client except to return more data than was desired.

pd/src/info/light_wallet.rs
```diff
@@ -0,0 +1,160 @@
...
+};
+use tonic::Status;
+use tracing::instrument;
+
+// TODO(hdevalence): this still doesn't work, giving up for now
+// We need to use the tracing-futures version of Instrument,
+// because we want to instrument a Stream, and the Stream trait
+// isn't in std, and the tracing::Instrument trait only works with
+// (stable) std types.
+// use tracing_futures::Instrument;
```
hdevalence, line 22: Supposedly there's a way to use the `Instrument` trait to attach a tracing span to a `Stream`, but I couldn't get it to work, and didn't think it was worth spending more time on right now.

proto/proto/light_wallet.proto
```diff
@@ -11,16 +12,23 @@ service LightWallet {
...
+message AssetListRequest {
+  // The expected chain id (empty string if no expectation).
+  string chain_id = 1;
 }
 
 // Requests a range of compact block data.
 message CompactBlockRangeRequest {
   // The expected chain id (empty string if no expectation).
-  string chain_id = 3;
+  string chain_id = 1;
```
hdevalence, line 27: Since we're breaking the RPC anyways, I took the opportunity to reorder the fields so that `chain_id` came first.

proto/proto/thin_wallet.proto
```diff
@@ -10,48 +10,13 @@ import "stake.proto";
 // Unlike the "light wallet" service, this protocol does not attempt to be
 // trust-minimized, either in terms of integrity or privacy.
 service ThinWallet {
-  rpc TransactionByNote(TransactionByNoteRequest) returns (TransactionDetail);
-  rpc AssetLookup(AssetLookupRequest) returns (chain.AssetInfo);
-  rpc AssetList(AssetListRequest) returns (stream Asset);
-  // TODO: return ValidatorStatus?
+  rpc TransactionByNote(crypto.NoteCommitment) returns (chain.NoteSource);
```
hdevalence, line 13: We never used `AssetLookup`, so I deleted it.  The `AssetList` doesn't request any specific user data, so I moved it to the "light" protocol.

redshiftzero (approved):
LGTM!

PR #572, opened 2022-04-07
Mint reward notes within the shielded pool component based on staking rewards

### LINKED ISSUE #549: Add validator reward notes to the `ShieldedPool`

Currently, the new `ShieldedPool` component's note commitment tree diverges from the old implementation. #548 unified note minting so that both implementations would agree on the genesis data, but it still falls out of agreement after the end of the first epoch, when the legacy implementation adds the staking reward notes.

We need to use the `ShieldedPool::mint_note` method to add reward notes to the shielded pool.  There are two options on how we could structure this:

1. Make adding the notes the `ShieldedPool`'s responsibility.  This means that the `ShieldedPool` needs to be able to read staking-relevant data in order to compute what the reward notes should be.
2. Give the `Staking` component access to the `ShieldedPool` somehow.  This would mean (at least) changing the `Component` trait to remove the `new` method, and then figuring out a way to share access to components.

For now, let's aim at (1) and then switch gears if we experience difficulty.

zbuc:
Fixes #549


pd/src/components/shielded_pool.rs
```diff
@@ -277,6 +279,33 @@ impl ShieldedPool {
...
+            .await?
+            .unwrap_or_default();
+
+        // TODO: should we calculate this here or include it directly within the PendingRewardNote
+        // to prevent a potential mismatch between Staking and ShieldedPool?
+        let epoch_index = Epoch::from_height(
+            self.compact_block.height,
+            self.overlay.get_epoch_duration().await?,
+        )
+        .index;
```
hdevalence, line 295: I think this is fine, it's not a super complicated calculation to duplicate.

pd/src/components/staking.rs
```diff
@@ -211,6 +215,18 @@ impl Staking {
...
+        // so they can be processed by the ShieldedPool.
+        let height = self.overlay.get_block_height().await?;
+        self.overlay
+            .put_domain(
+                format!("reward_notes/{}", height).into(),
+                RewardNotes {
+                    notes: reward_notes,
+                },
+            )
+            .await;
```
hdevalence, line 228: This should be wrapped up in methods on `staking::View`, like
```rust
async fn set_reward_notes(&self, height: block::Height, notes: RewardNotes) { /* ... */ }
async fn reward_notes(&self, height: block::Height) -> Result<Option<RewardNotes>> { /* ... */ }
```
so that we can separate the tree accesses (working with the paths, etc) from the app logic that uses the data, and so that we only ever define the paths in one place (the `View` trait for the component).

Also, I think the key should probably be `staking/reward_notes/{}` so that it has the same prefix as the other keys.

pd/src/components/shielded_pool.rs
```diff
@@ -277,6 +279,33 @@ impl ShieldedPool {
 
     #[instrument(skip(self))]
     async fn write_block(&mut self) -> Result<()> {
+        // Handle any pending reward notes from the Staking component
+        let notes: RewardNotes = self
+            .overlay
+            .get_domain(format!("reward_notes/{}", self.compact_block.height).into())
+            .await?
+            .unwrap_or_default();
```
hdevalence, line 287: Instead of constructing the path manually here (meaning we have to check that it's constructed consistently), I think we should `use super::staking::View as _;` so that the paths can be in a single place.

pd/src/components/staking.rs
zbuc: Yeah I wasn't sure what prefix to use since it's shared between two components. Is the pattern going to be "prefix it with the setter of the field"?

hdevalence (changes requested):
This looks good to me except for the path construction -- we should have a design guideline that explicitly constructed paths into JMT data are only ever used in the `View` traits, so that we don't end up needing to coordinate a bunch of different places in the code all having different copies of hardcoded paths.

pd/src/components/staking.rs
hdevalence: I think if we have a rule that we only ever define paths in a `View` trait, I think we get the answer "for free", in that we can just use the prefix of whatever component is defining that view

hdevalence (approved):

PR #578, opened 2022-04-09
add basic IBC updateclient handling

### LINKED ISSUE #634: Implement IBC UpdateClient
Tracking issue for implementation of IBC UpdateClient messages, which implement light client updates for IBC clients.

avahowell:
implements handling `MsgUpdateClient` IBC messages, which perform the core of IBC light client verification. Closes #634


hdevalence:
Looking through the code, one thing I'm noticing is that the implementation of `check_tx_stateful` is empty, and the validation logic is done in the `execute_tx` path.  This is a problem, because the check/execute split is how we ensure that we're actually executing things transactionally, and it seems like another example of how not having #632 is causing problems for us.

Making the change in #632 will require changing the implementation of the `Component`s, so it interacts with this code.  I think the easiest thing to do would be to merge a version of this code, and then refactor it as part of implementing #632.

avahowell:
Refactored this to use check_tx_stateless, check_tx_stateful correctly

hdevalence (approved):

PR #586, opened 2022-04-11
Remove all references to postgres from README/code/docker configs

### LINKED ISSUE #581: Remove database connection parameter from `pd`

As of #579, we don't load a Postgres connection any more, and the `database_uri` parameter is only used in a log message.  However, our documentation still tells people to set up Postgres and connect to it.

We should:

1. Remove the `database_uri` parameter from the `Start` command;
2. Update the docker-compose files to not pass that parameter, and to strip out the Postgres-related parts;
3. Simplify the instructions on how to build and run `pd` by removing all of the Postgres-related parts.

At the moment, we *shouldn't* need a Postgres instance to build `pd`, because we have the `sqlx-data.json` checked in to the source tree.  We won't be making any more changes to the existing query code, so this shouldn't need to be regenerated until we go through and delete all of the code.

However, updating the deployment scripts and docs first means that we don't need to block launching+deploying Herse on going through and deleting all of our legacy application code #582 (which we should do, just, probably not tomorrow morning).

zbuc:
Closes #581


hdevalence:
This looks good, except it looks like the Orthosie allocations got chomped?

Issue #587, opened 2022-04-11
Restructure client code into a bundle of microservices


hdevalence:
## Problem

We're now (or about to be) facing a similar situation on the client side (`pcli`) as we were in `pd` prior to #488: we started with code designed to be an MVP, it's been extended beyond that scope, and we're now at the point where further extension risks blowing up the complexity of the code.

At the same time, we're getting a better sense of our needs for the client code, and there are more use-cases to support:

- `pcli` itself: a standalone command-line utility for interacting with the chain;
- `galileo`, the discord faucet bot (as a test-case for "an online service that transacts on-chain");
- IBC relayers, which need to be able to at least create transactions, and possibly manage shielded funds to be able to pay fees;
- active LPs, who would want to be able to manage liquidity (this is a speculative use-case)

The current client code (`pcli` + `penumbra-wallet`) was designed with a much more vague version of the above use cases in mind: `penumbra-wallet` was supposed to have reusable code that would handle Penumbra-specific logic (like how to scan a block, and what state to serialize as part of the wallet), which `pcli` would call to do whatever operations it needed.  Other applications could use the `penumbra-wallet` code directly.  However, then they have to use the Rust crate (trivial for Rust applications, potentially nontrivial for other languages), and implement the logic required to drive the wallet code, and this is a bunch of effort.

## Solution

Instead, it would be useful to be able to completely offload responsibility for synchronizing the chain state and performing transactions to a standalone **view service**.  This service would:

- maintain user data in a database (SQLite seems like a [good choice](https://www.sqlite.org/appfileformat.html));
- synchronize with the chain state on launch;
- stay in sync with the chain state as long as it's running;
- implement a gRPC service that:
  - buffers all requests until the service is ready (synchronized with the chain state);
  - provides queries for private chain state (UTXOs, balances, transactions, ...);
  
This service should be written as a library, so that it can be used in multiple ways:

- by a standalone `pviewd` server that exposes the gRPC service (access control for the gRPC service is left as a deployment concern);
- by a Rust application like `pcli`, which would create an instance of the view service internally, but call its (gRPC-derived) methods directly, then shut down once it was finished.

To clarify terminology:

- The **client protocol** is spoken between a full node (the server) and a client wishing to learn about the public state (including encrypted data).  This is split into two sub-protocols:
  - The **oblivious client protocol** has requests that are intended to not reveal specific user information (e.g., downloading a range of `CompactBlocks`, or downloading information on every validator);
[...]

PR #600, opened 2022-04-12
Remove old postgres code, rename light wallet/thin wallet -> light client protocol/thin client protocol

zbuc:
Part of #587


hdevalence:
I think it might be cleaner to do the removal and the scaffolding as two separate PRs?

zbuc:
@hdevalence I split out the scaffolding into #600 though I'm leaving that one in draft mode because it builds on the changes from this PR so I'll need to rebase on main after this is merged

pd/src/mempool.rs
```diff
@@ -32,29 +32,22 @@ use tokio::sync::{watch, Mutex as AsyncMutex};
...
 
-use crate::{state, verify::StatelessTransactionExt, RequestExt};
+use crate::RequestExt;
 
 #[derive(Clone)]
 pub struct OldMempool {
     // sidecar
     new_mempool: Mempool,
     // old code
     nullifiers: Arc<AsyncMutex<BTreeSet<Nullifier>>>,
```
hdevalence, line 42: The whole `OldMempool` struct can be deleted, since we don't use it at all any more.

pd/src/state.rs
```diff
@@ -11,66 +10,7 @@ pub use reader::Reader;
 pub use writer::Writer;
 
 #[instrument]
+// TODO: killme
 pub async fn new(uri: &str) -> Result<(Reader, Writer)> {
```
hdevalence, line 14: The whole `state` module subtree can also be cleanly deleted, since it only did Postgres stuff, and the new `storage` module is completely separate from it.

pd/src/wallet/server.rs
```diff
@@ -1,22 +1,21 @@
...
+//             Ok(())
+//         } else {
+//             Err(tonic::Status::failed_precondition(format!(
+//                 "provided chain_id {} does not match chain_id {}",
+//                 provided,
+//                 self.chain_params_rx().borrow().chain_id
+//             )))
+//         }
+//     }
+// }
```
hdevalence, line 21: This is already implemented, I think as part of the `app::View` trait, so it can be deleted here as well. 

In fact the whole `wallet` module subtree can be deleted, since the new RPC implementation lives under `info` (in hope of some future unification with the ABCI query interface).

pd/src/wallet/server/light.rs
```diff
@@ -13,113 +13,112 @@ use penumbra_proto::{
...
-            .compact_blocks(
-                start_height.try_into().unwrap(),
-                end_height.try_into().unwrap(),
-            )
-            .map_err(|e| tonic::Status::internal(e.to_string()));
-
-        Ok(tonic::Response::new(stream.boxed()))
-    }
-}
+// TODO: implement for WriteOverlay
```
hdevalence, line 16: This is implemented already, so it can be deleted with the rest of the `wallet` code.

pd/src/wallet/server/thin.rs
```diff
@@ -8,48 +8,47 @@ use penumbra_proto::{
...
-
-        let rate = rates
-            .get(&identity_key)
-            .ok_or_else(|| tonic::Status::not_found("validator not found"))?
-            .clone();
-
-        Ok(tonic::Response::new(rate.into()))
-    }
-}
+// TODO: implement for WriteOverlay
```
hdevalence, line 11: Ditto above comment on deletion

hdevalence:
This looks good, I think there's even more code that could be deleted and left some comments on where.  Following rounds of deletions, it'd be useful to run `cargo fix` and then `cargo fmt` (`cargo fix` only applies on a clean checkout, but does not reformat the source code) to clean up newly unused imports.

PR #606, opened 2022-04-14
Add wallet.proto

aubrika:


proto/proto/wallet.proto
```diff
@@ -0,0 +1,62 @@
...
+}
+
+// Requests all unspent transactions associated with the wallet.
+message UtxoRequest {
+    uint64 id = 1;
+}
+
+// Contains an unspent transaction.
+message UtxoResult {
+}
```
zbuc, line 51: It wasn't clear what this would actually look like or what the analog in the current pcli implementation is

zbuc (approved):

proto/proto/wallet.proto
hdevalence: The analog would be unspent notes, which are currently visible in `pcli` only through `balance --by-note` but otherwise aren't exposed.

proto/proto/wallet.proto
```diff
@@ -0,0 +1,62 @@
...
+// Returns whether the creation of the transaction was successful or not.
+message CreateTransactionResult {
+    bool success = 1;
+}
+
+// Returns the status of the wallet server and whether it is synchronized with the chain state.
+message StatusResult {
+  /// Whether the wallet service is synchronized with the chain state.
+  bool synchronized = 1;
+}
```
hdevalence, line 37: It might be handy to return heights, like
```
// The height reported by the full node
uint64 chain_height = 2;
// The height the wallet has synchronized to so far
uint64 sync_height = 3;
```
so that the client would be able to get progress info.

proto/proto/wallet.proto
```diff
@@ -0,0 +1,62 @@
...
+  /// List unspent transaction outputs
+  rpc Utxo(UtxoRequest) returns (stream UtxoResult);
+  /// Display current wallet balance
+  rpc Balance(BalanceRequest) returns (stream BalanceResult);
+}
+
+// Requests the wallet server to sign and submit a transaction.
+message CreateTransactionRequest {
+    transaction.TransactionBody body = 1;
+}
```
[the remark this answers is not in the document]
hdevalence: Update: after thinking about it some more, I think this comment points in the wrong direction, and will update #614 accordingly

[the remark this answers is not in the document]
hdevalence: Updated: https://github.com/penumbra-zone/penumbra/issues/614#issue-1205443086

hdevalence:
I updated #614 with some reasoning about where we might want to go with this design, but I think we should merge this now to be able to continue iterating on it.

PR #617, opened 2022-04-15
TCT integration

plaidfinch:
This PR comprises a variety of changes along the way to integrating the TCT into the rest of the Penumbra codebase. They can be seen as falling into several distinct categories:

- internal improvements to the TCT discovered during the course of integration,
- a full executable specification and property test suite for the TCT developed to gain confidence before integration,
- interface improvements to the TCT discovered during the course of integration, and
- changes to various parts of the Penumbra codebase outside of the TCT to integrate it and remove the `incrementalmerkletree` crate, notably:
  - breaking changes to the proto formats for transparent proofs,
  - drop-in integration into the `crypto` and `transaction` crates to replace equivalent parts of the previous NCT,
  - changes to the `wallet` crate to support scanning compact blocks into the new TCT design: this is currently _not_ optimized using the TCT's unique summarizing features, but it is written to automatically roll into new epochs based on the block height,
  - fixes to `pcli` to update code written while this PR was pending to use the new crate and its API when necessary in transaction construction, and
  - changes and fixes to `pd` to integrate the TCT into the new component architecture for the shielded pool.

I should note that I have not yet tested the combination of these changes to `pd` and `pcli` to ensure that they produce a coherent result. That should happen before this merges.


Cargo.toml
```diff
@@ -15,9 +15,14 @@ members = [
   "pcli",
   "pcli-next",
   "tct",
+  "tct-property-test",
   "ibc",
 ]
 
+[profile.test]
+# Maximum optimization level for tests, so that they are fast (especially the proptest ones)
+opt-level = 3
```
hdevalence, line 24: This is changing the entire workspace configuration to do more expensive builds, regardless of whether or not the expensive tests are actually going to be run.

Could we just do `cargo test --release` instead when running the slow, long-running property tests?

crypto/src/note.rs
```diff
@@ -333,6 +333,24 @@ pub struct Commitment(pub Fq);
...
+    fn from(commitment: Commitment) -> penumbra_tct::Commitment {
+        commitment.0.into()
+    }
+}
+
+impl From<penumbra_tct::Commitment> for Commitment {
+    fn from(commitment: penumbra_tct::Commitment) -> Commitment {
+        Commitment(commitment.into())
+    }
+}
```
hdevalence, line 352: Having two copies of the (logically) same type in two different crates is pretty awkward, although I guess it's not the worst thing for now.

crypto/src/proofs/transparent.rs
```diff
@@ -652,20 +606,17 @@ mod tests {
         let rsk = sk_sender.spend_auth_key().randomize(&spend_auth_randomizer);
         let nk = *sk_sender.nullifier_key();
         let ak = sk_sender.spend_auth_key().into();
-        let mut nct = merkle::BridgeTree::<note::Commitment, 32>::new(5);
-        nct.append(&note_commitment);
-        let anchor = nct.root2();
-        nct.witness();
-        let merkle_path = nct.authentication_path(&note_commitment).unwrap();
+        let mut nct = penumbra_tct::Eternity::new();
```
hdevalence, line 609: This name is kind of confusing out of context. I think it would be better if the structures from the `tct` crate had `Tree` in the name, like `EternityTree`, `EpochTree`, `BlockTree`, etc.

crypto/src/proofs/transparent.rs
```diff
@@ -652,20 +606,17 @@ mod tests {
...
+        nct.insert(penumbra_tct::Keep, note_commitment).unwrap();
```
hdevalence, line 610: Rather than re-exporting the inner `Witness::Keep` and `Witness::Forget` variants from the crate root, then calling them `tct::Keep`, it would be clearer if the calling code just imported `Witness` and wrote
```
nct.insert(Witness::Keep, note_commitment);
```

pd/src/components/shielded_pool.rs
```diff
@@ -271,27 +269,48 @@ impl ShieldedPool {
...
         )
-        .await;
+        .await?;
 
         Ok(())
     }
 
     #[instrument(skip(self, source, output_body))]
-    async fn add_note(&mut self, output_body: output::Body, source: NoteSource) {
+    async fn add_note(&mut self, output_body: output::Body, source: NoteSource) -> Result<()> {
```
hdevalence, line 278: This function shouldn't return a `Result`, because we cannot fail to execute a transaction once we've accepted it with `check_tx_stateful`.

Instead, we need to check that the `BlockTree` is not full in the `check_tx_stateful` method, and then `.expect` (with a reference to that check) that the `BlockTree` is not full in `execute_tx`.

pd/src/components/shielded_pool.rs
```diff
@@ -271,27 +269,48 @@ impl ShieldedPool {
...
+        let height = self.compact_block.height;
+        if height != 0 {
+            if height % epoch_duration == 0 {
+                self.note_commitment_tree
+                    .insert_epoch(penumbra_tct::Epoch::new())?;
+            } else {
+                self.note_commitment_tree
+                    .insert_block(penumbra_tct::Block::new())?;
+            }
+        }
```
hdevalence, line 306: This code special-cases `0`, and skips inserting a new block after processing the genesis state.  This means that the genesis allocations will end up in the same tree as any notes created in the first block, which seems wrong.

I think the root cause of this problem is that this function was originally just supposed to write the block, but ended up containing the `end_block` implementation.

pd/src/components/shielded_pool.rs
```diff
@@ -271,27 +269,48 @@ impl ShieldedPool {
...
+
+        // Check that the position we're starting at matches the height of the block
+        assert_eq!(
+            u64::from(self.note_commitment_tree.position()) >> 16,
+            height
+        );
```
hdevalence, line 312: This assertion is wrong, because it does not compute the height correctly (it will fail as soon as the first epoch ends, halting the chain).

pd/src/components/shielded_pool.rs
```diff
@@ -271,27 +269,48 @@ impl ShieldedPool {
...
+        // if so; otherwise, just create a new block
+        let epoch_duration = self.overlay.get_epoch_duration().await?;
+        let height = self.compact_block.height;
+        if height != 0 {
+            if height % epoch_duration == 0 {
+                self.note_commitment_tree
+                    .insert_epoch(penumbra_tct::Epoch::new())?;
+            } else {
+                self.note_commitment_tree
+                    .insert_block(penumbra_tct::Block::new())?;
```
hdevalence, line 304: It's a bit unintuitive that the way to finish a block or epoch is to insert an additional, empty structure.  I think it would be better to have `finish_block` and `finish_epoch` methods (even if they have this implementation internally).

tct/Cargo.toml
```diff
@@ -6,18 +6,35 @@ edition = "2021"
...
 hash_hasher = "2"
 thiserror = "1"
 serde = { version = "1.0", features = ["derive"] }
+parking_lot = "0.12"
 ark-ff = { git = "https://github.com/penumbra-zone/algebra", branch = "ours" }
 ark-serialize = { git = "https://github.com/penumbra-zone/algebra", branch = "ours" }
 poseidon377 = { git = "https://github.com/penumbra-zone/poseidon377" }
 decaf377 = { git = "https://github.com/penumbra-zone/decaf377" }
+# this allows us to implement Decode for various types without orphan rule issues
+sqlx = { version = "0.5.9", features = ["postgres"], optional = true }
```
hdevalence, line 24: We should not include `sqlx` here, since we've just finished standardizing on using our canonical, proto-based serialization for persistent storage.

tct/src/epoch.rs
```diff
@@ -94,35 +165,56 @@ impl Epoch {
...
         let index = *self.index.get(&commitment)?;
 
-        let (auth_path, leaf) = self.inner.witness(index)?;
+        let (auth_path, leaf) = match self.inner.witness(index) {
+            Some(witness) => witness,
+            None => panic!(
+                "commitment `{:?}` indexed with position `{:?}` must be witnessed",
+                commitment, index
+            ),
+        };
```
hdevalence, line 213: Why are we panicking here instead of just returning `None` as before?

tct/src/epoch.rs
```diff
@@ -404,3 +521,33 @@ impl EpochMut<'_> {
...
+            self.0.encode_by_ref(buf)
+        }
+    }
+
+    impl Type<Postgres> for Root {
+        fn type_info() -> <Postgres as Database>::TypeInfo {
+            <[u8] as Type<Postgres>>::type_info()
+        }
+    }
+}
```
hdevalence, line 553: We don't need this code, because we don't use Postgres any more.

tct/src/internal/hash.rs
```diff
@@ -103,14 +115,98 @@ impl Hash {
...
+            <&[u8] as sqlx::Encode<Postgres>>::encode(&bytes[..], buf)
+        }
+    }
+
+    impl Type<Postgres> for Hash {
+        fn type_info() -> <Postgres as Database>::TypeInfo {
+            <[u8] as Type<Postgres>>::type_info()
+        }
+    }
+}
```
hdevalence, line 171: This code should also not be added, since we don't use Postgres any more.

tct/src/internal/hash/cache.rs
```diff
@@ -0,0 +1,104 @@
...
+    once: Once,
+}
+
+// This is safe because the restricted API of `CachedHash` only sets the internal hash *at most
+// once*, from only one thread, using a `Once` to ensure multiple threads do not race on setting the
+// hash. The only way to do interior mutability is via `set_if_empty`, which mutates the inner
+// `Cell` exactly once. Additionally, the value of the inner `Cell` cannot be observed until after
+// it has been fully set, which means it is impossible to observe partial writes, regardless of
+// whether this would be possible in practice on any supported architecture.
+unsafe impl Sync for CachedHash {}
```
hdevalence, line 29: We should absolutely not be doing an `unsafe impl Sync` in our Merkle tree implementation.  It is critical that the Merkle tree work correctly, and having to reason about synchronization behaviour on top of basic functional correctness adds a huge amount of risk for very little gain.

tct/src/internal/hash/cache.rs
```diff
@@ -0,0 +1,104 @@
...
+
+        CachedHash { once, cell }
+    }
+}
+
+// A derived `Debug` implementation won't do, because the generated code would access the `Cell`
+// without going through the `Once` synchronization barrier. Our manual implementation, in addition
+// to ensuring that torn reads cannot be observed (by using the safe method `get`), gives a succinct
+// representation of the cached hash, using `_` to denote a hash that hasn't yet been populated.
+impl Debug for CachedHash {
```
hdevalence, line 61: The fact that accidentally leaving a `#[derive(Debug)]` on the struct (or adding one in the future, or ...) could break the soundness of the code we use to verify note inclusion seems like a very clear sign that using `unsafe` here is a bad way to allocate implementation risk.

tct/src/internal/hash/fast_hash.rs
```diff
@@ -0,0 +1,25 @@
+//! This module, when enabled with the `fast_hash` feature flag, replaces the Poseidon hash
+//! implementation with `blake2b_simd`, which is much faster. **This is useless for production use
+//! of this crate; it is only useful to accelerate testing, when we don't care about producing
+//! zero-knowledge proofs.**
```
hdevalence, line 4: We shouldn't make this change in this PR, because it is already far too large, but wouldn't this be a non-problem if the tree were generic over the leaf type, like the `incrementalmerkletree` API?

wallet/src/state.rs
```diff
@@ -794,9 +787,48 @@ impl ClientState {
...
+            self.note_commitment_tree
+                .insert_epoch(penumbra_tct::Epoch::new())
+                .map_err(|e| anyhow!(e.to_string()))?;
+        }
+
+        // Insert the block into the full commitment tree
+        self.note_commitment_tree
+            .insert_block(block)
+            // Errors returned by `insert_block` are not `Sync` because they contain a `Block`,
+            // which is not `Sync`, so we can't return them directly
```
hdevalence, line 816: The errors should be `Sync` to avoid this type of problem.

hdevalence (changes requested):
We should not use unsafe `Sync` impls in our Merkle tree implementation.  There are some other minor API design points, but this is the major blocker.  It would be better to just use a `parking_lot` mutex.

We should also run the code to see whether it works correctly before merging -- currently, it panics on the first epoch transition, because of https://github.com/penumbra-zone/penumbra/pull/617#discussion_r851674520 , but even after that, the client can't submit any transactions, because it doesn't compute the same NCT root.

pd/src/components/shielded_pool.rs
hdevalence: Fixed in https://github.com/penumbra-zone/penumbra/pull/617/commits/cda7931722303f4894e333fad8e1d29b65a65305

pd/src/components/shielded_pool.rs
hdevalence: Fixed in https://github.com/penumbra-zone/penumbra/pull/617/commits/cda7931722303f4894e333fad8e1d29b65a65305

hdevalence:
Fixed the chain halt and block 0 / block 1 merge in cda7931, but transactions still fail because the submitted anchor is not valid.  I noticed that 
```
Apr 17 00:58:37.965 DEBUG abci:EndBlock{height=344}:end_block:shielded_pool:write_block: pd::components::shielded_pool: writing anchor height=0 anchor=Root(d132c20a)
```
The `height` field is obviously wrong there, but there's also something wrong with the `Debug` implementation for `Root` that's causing it to print a truncated hash.  Looking into the code, the `Root`'s `Debug` implementation is derived from that of `Hash`, which is doing something mysterious:
```rust
impl Debug for Hash {
    fn fmt(&self, f: &mut std::fmt::Formatter) -> Result<(), std::fmt::Error> {
        let mut bytes = Vec::with_capacity(4 * 8);
        self.0.write(&mut bytes).unwrap();
        write!(f, "{}", hex::encode(&bytes[3 * 8 + 4..]))
    }
}
```

hdevalence:
Why does the NCT root change every fourth block, even when there are no new notes?
```
Apr 17 02:05:22.896 DEBUG scan_block{height=16}: penumbra_wallet::state: finished scanning block self.last_block_height=16 root=Root(43570e07)
Apr 17 02:05:22.899 DEBUG scan_block{height=17}: penumbra_wallet::state: starting block scan outputs_len=0
Apr 17 02:05:22.899 DEBUG scan_block{height=17}: penumbra_wallet::state: inserting block tree
Apr 17 02:05:22.899 DEBUG scan_block{height=17}: penumbra_wallet::state: finished scanning block self.last_block_height=17 root=Root(43570e07)
Apr 17 02:05:22.899 DEBUG scan_block{height=18}: penumbra_wallet::state: starting block scan outputs_len=0
Apr 17 02:05:22.899 DEBUG scan_block{height=18}: penumbra_wallet::state: inserting block tree
Apr 17 02:05:22.899 DEBUG scan_block{height=18}: penumbra_wallet::state: finished scanning block self.last_block_height=18 root=Root(43570e07)
Apr 17 02:05:22.899 DEBUG scan_block{height=19}: penumbra_wallet::state: starting block scan outputs_len=0
Apr 17 02:05:22.899 DEBUG scan_block{height=19}: penumbra_wallet::state: inserting block tree
Apr 17 02:05:22.899 DEBUG scan_block{height=19}: penumbra_wallet::state: finished scanning block self.last_block_height=19 root=Root(43570e07)
Apr 17 02:05:22.899 DEBUG scan_block{height=20}: penumbra_wallet::state: starting block scan outputs_len=0
Apr 17 02:05:22.899 DEBUG scan_block{height=20}: penumbra_wallet::state: inserting block tree
Apr 17 02:05:22.899 DEBUG scan_block{height=20}: penumbra_wallet::state: finished scanning block self.last_block_height=20 root=Root(bd866301)
Apr 17 02:05:22.899 DEBUG scan_block{height=21}: penumbra_wallet::state: starting block scan outputs_len=0
Apr 17 02:05:22.899 DEBUG scan_block{height=21}: penumbra_wallet::state: inserting block tree
Apr 17 02:05:22.899 DEBUG scan_block{height=21}: penumbra_wallet::state: finished scanning block self.last_block_height=21 root=Root(bd866301)
Apr 17 02:05:22.899 DEBUG scan_block{height=22}: penumbra_wallet::state: starting block scan outputs_len=0
Apr 17 02:05:22.899 DEBUG scan_block{height=22}: penumbra_wallet::state: inserting block tree
Apr 17 02:05:22.899 DEBUG scan_block{height=22}: penumbra_wallet::state: finished scanning block self.last_block_height=22 root=Root(bd866301)
Apr 17 02:05:22.899 DEBUG scan_block{height=23}: penumbra_wallet::state: starting block scan outputs_len=0
Apr 17 02:05:22.899 DEBUG scan_block{height=23}: penumbra_wallet::state: inserting block tree
Apr 17 02:05:22.899 DEBUG scan_block{height=23}: penumbra_wallet::state: finished scanning block self.last_block_height=23 root=Root(bd866301)
Apr 17 02:05:22.899 DEBUG scan_block{height=24}: penumbra_wallet::state: starting block scan outputs_len=0
Apr 17 02:05:22.899 DEBUG scan_block{height=24}: penumbra_wallet::state: inserting block tree
Apr 17 02:05:22.899 DEBUG scan_block{height=24}: penumbra_wallet::state: finished scanning block self.last_block_height=24 root=Root(a1e1ac04)
```

hdevalence:
After some further rearrangement to fix NCT root desyncing between the client and the server, it no longer falls out of sync immediately.  However, it looks like there's something wrong with the way that adding a complete block tree works, because it ends up different from adding an empty one and then appending to it.

On the `pd` side, we have
```
Apr 17 02:47:23.263 DEBUG abci:DeliverTx{txid="53acd053b9a070de23be87a4243b2da165c72afda856a560cb954e9441167f1a"}:execute_tx:shielded_pool:add_note: pd::components::shielded_pool: appending to NCT in component commitment=note::Commitment(e4c09957dc8b720e909b47a84d6737bdc464b9f5053f6331b291aaa21e7cb907) position=Position(Eternity { epoch: 1, block: 19, commitment: 0 })
```
so the note commitment is inserted at position `1/19/0`, and then in `end_block`, we have
```
Apr 17 02:47:23.264 DEBUG abci:EndBlock{height=60}:end_block:shielded_pool: pd::components::shielded_pool: inserting block tree for next block
```
and then
```
Apr 17 02:47:23.265 DEBUG abci:EndBlock{height=60}:end_block:shielded_pool:write_block: pd::components::shielded_pool: position=Position(Eternity { epoch: 1, block: 20, commitment: 0 })
Apr 17 02:47:23.265 DEBUG abci:EndBlock{height=60}:end_block:shielded_pool:write_block: pd::components::shielded_pool: writing anchor height=60 anchor=Root(5362b403)
```

On the `pcli` side, we have
```
Apr 17 02:47:29.305 DEBUG scan_block{height=60}: penumbra_wallet::state: position=Position(Eternity { epoch: 1, block: 19, commitment: 0 })
Apr 17 02:47:29.305 DEBUG scan_block{height=60}: penumbra_wallet::state: starting block scan outputs_len=1
Apr 17 02:47:29.305 DEBUG scan_block{height=60}: penumbra_wallet::state: found note while scanning note_commitment=note::Commitment(e4c09957dc8b720e909b47a84d6737bdc464b9f5053f6331b291aaa21e7cb907) position=Position(Eternity { epoch: 1, block: 19, commitment: 0 }) note=Note { value: Value { amount: 100000000000, asset_id: passet1984fctenw8m2fpl8a9wzguzp7j34d7vravryuhft808nyt9fdggqxmanqm }, note_blinding: Fp256(BigInteger256([1925076761441888379, 10327062928397409300, 16562428622360418953, 14453853609863477])), diversifier: Diversifier(e21c212d0d11002a7f5ab5), transmission_key: decaf377_ka::Public(025df1c0885c83a1f9f4370ec0e3edec336ffc513bd80611c15b0c9df6d0c901), transmission_key_s: Fp256(BigInteger256([15986969656633463360, 6500317334198110268, 4067783070624381186, 1271587261249631793])) }
Apr 17 02:47:29.306 DEBUG scan_block{height=60}: penumbra_wallet::state: inserting block tree
Apr 17 02:47:29.306 DEBUG scan_block{height=60}: penumbra_wallet::state: found nullifier for submitted spend note, marking it as spent value=Value { amount: 100000000000, asset_id: passet1984fctenw8m2fpl8a9wzguzp7j34d7vravryuhft808nyt9fdggqxmanqm } nullifier=Nullifier(e292ad67f1a986df243a3d5ce7ec385c4fadfc6ce5ff333f221efc760f8f0604)
Apr 17 02:47:29.306 DEBUG scan_block{height=60}: penumbra_wallet::state: finished scanning block self.last_block_height=60 root=Root(77f69003) position=Position(Eternity { epoch: 1, block: 20, commitment: 1 })
```
so although the note is inserted in the same `1/19/0` position, we end up with a different position and a different root hash.

hdevalence:
Adding more tracing suggests there's a bug in `insert_block`, since inserting a block tree to advance the position doesn't do so:
```
Apr 17 03:29:07.985 DEBUG abci:EndBlock{height=40}:end_block:shielded_pool: pd::components::shielded_pool: position=Position(Eternity { epoch: 1, block: 0, commitment: 0 })
Apr 17 03:29:07.985 DEBUG abci:EndBlock{height=40}:end_block:shielded_pool: pd::components::shielded_pool: inserting block tree for next block
Apr 17 03:29:07.985 DEBUG abci:EndBlock{height=40}:end_block:shielded_pool: pd::components::shielded_pool: position=Position(Eternity { epoch: 1, block: 0, commitment: 0 })
```

pd/src/components/shielded_pool.rs
hdevalence: The `Component` API doesn't make it clear that this is a problem, because there's no distinction in what *kind* of `Result`s are returned.  Discussion of a potential fix: https://github.com/penumbra-zone/penumbra/issues/632

hdevalence:
From discussion with @plaidfinch, since this PR is very large but not immediately mergeable, the current plan is to try to separate this branch into two parts, one that only touches the `tct` code, and another that touches the rest of the workspace.

tct/src/internal/hash/cache.rs
plaidfinch: This is now resolved in the tct-integration-tct-only branch.

tct/src/internal/hash/cache.rs
plaidfinch: This is now resolved in the tct-integration-tct-only branch.

tct/Cargo.toml
plaidfinch: Sqlx has been removed in the tct-integration-tct-only branch.

tct/src/epoch.rs
plaidfinch: Removed in tct-integration-tct-only.

tct/src/internal/hash.rs
plaidfinch: Removed in tct-integration-tct-only.

plaidfinch:
> The `height` field is obviously wrong there, but there's also something wrong with the `Debug` implementation for `Root` that's causing it to print a truncated hash. Looking into the code, the `Root`'s `Debug` implementation is derived from that of `Hash`, which is doing something mysterious:
> 
> ```rust
> impl Debug for Hash {
>     fn fmt(&self, f: &mut std::fmt::Formatter) -> Result<(), std::fmt::Error> {
>         let mut bytes = Vec::with_capacity(4 * 8);
>         self.0.write(&mut bytes).unwrap();
>         write!(f, "{}", hex::encode(&bytes[3 * 8 + 4..]))
>     }
> }
> ```

This was useful for debugging purposes when I was developing the tree, but it's now been fixed to print the whole hash in the tct-integration-tct-only branch.

PR #628, opened 2022-04-18
Corrected Genesis

dvjromashkin:
There was a json RPC element in the Genesis file, because of which Tendermint would not start with it, and gave an error. I've installed the corrected Genesis that works in Tendermint.


hdevalence:
@dvjromashkin thanks for the PR! I edited the commit to replace the existing genesis file completely.

PR #629, opened 2022-04-18
pd: refactor `ShieldedPool::write_block`

hdevalence:
The existing implementation had application logic related to ending the block
(e.g., minting reward notes) in `ShieldedPool::write_block`, which is used by
both `end_block` and `init_chain`.  This worked because the reward notes would
be empty during `init_chain`, but in general isn't great to mix application
logic between the two cases.

Instead, this commit pulls out that logic into `end_block`, and renames
`write_block` to `write_compactblock_and_nct`, so that it's clearer from the
name of the method that it's just about writing specific data, not doing
general processing.

### DIFF
--- a/pd/src/components/shielded_pool.rs
+++ b/pd/src/components/shielded_pool.rs
@@ -74,7 +74,7 @@ impl Component for ShieldedPool {
         }
 
         self.compact_block.height = 0;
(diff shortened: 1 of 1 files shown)


zbuc (approved):
LGTM

Issue #619, opened 2022-04-15
Delegating to non-genesis validator causes `pcli stake show` to crash


zbuc:
**Describe the bug**
If a validator is created post-genesis and you delegate to it, `pcli stake show` will begin crashing because it does not know about the delegation token type.

**To Reproduce**
1. Start a local testnet
2. Upload a new validator definition
3. Stake to the new validator
4. `pcli stake show`

```console
thread 'main' panicked at 'all asset IDs should have denominations stored locally', wallet/src/state.rs:618:22
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I think the best way to handle this is during the chain sync.

hdevalence:
Along the way to testing this, I noticed some awkwardness with the process of defining and uploading validators: #648

PR #649, opened 2022-04-19
pcli: add validator template-definition and fetch-definition commands

### LINKED ISSUE #648: Improve ergonomics for validator definitions

We have a command for uploading validator definitions, but the process of creating the definitions could be more ergonomic.

Add commands to `pcli validator`:

- [ ] `template-definition`: generates a template JSON file for editing
- [ ] `fetch-definition IDENTITY`: fetches an existing validator definition by identity key

hdevalence:
Closes #648

Along the way to this, I noticed that we have a lot of overlapping "validator"
structures, and maybe we should try to streamline those later?


hdevalence:
I tested locally that this fixes #619.

PR #654, opened 2022-04-19
Add configurations for Cyllene testnet

zbuc:
This should be left un-merged until we are ready to deploy Cyllene, but I wanted to make sure we didn't forget to add additional allocations to Galileo so things go more smoothly.


docs/guide/src/pd/join-testnet.md
```diff
@@ -1,67 +1,12 @@
...
 
-Currently we only provide instructions for running a fullnode deployment. A
+We provide instructions for running both fullnode deployments and validator deployments. A
 fullnode will sync with the network but will not have any voting power, and will
 not be eligible for staking or funding stream rewards. For more information on
 what a fullnode is, see the [Tendermint
 documentation](https://docs.tendermint.com/v0.35/nodes/#full-node).
 
-## Configuring your node
+A regular validator will participate in voting and rewards.
```
hdevalence, line 9:
```suggestion
A regular validator will participate in voting and rewards. (Of course, these rewards, like all of the other testnet tokens, have no value). 
```

docs/guide/src/pd/join-testnet/validator.md
```diff
@@ -0,0 +1,170 @@
...
+
+```console
+\$ grep -A3 pub_key ~/.tendermint/config/priv_validator_key.json
+  "pub_key": {
+    "type": "tendermint/PubKeyEd25519",
+    "value": "Fodjg0m1kF/6uzcAZpRcLJswGf3EeNShLP2A+UCz8lw="
+  },
+```
+
+The `identity_key` value placed by the `template-definition` command in your `validator.json` will correspond to the key used by your wallet configs for `pcli`, i.e. `pcli` will have access to any balance delegated to the validator.
```
hdevalence, line 114: Flagging this because it's not actually true -- the identity key for a validator doesn't get to control delegations, since delegations are totally non-custodial.

zbuc: Good catch, what would be more accurate to say? "The validator's address will be determined by the keys configured for the pcli wallet"?

zbuc:
@hdevalence we should probably link back to the validators page in the spec doc or put a summary version here so the mechanics are clear

docs/guide/src/pd/join-testnet/validator.md
hdevalence: I put this:

>  The root of a validator's identity is their *identity key*.  Currently, `pcli`
reuses the spend authorization key in whatever wallet is active as the
validator's identity key.  This key is used to sign validator definitions that
update the configuration for a validator.

Issue #491, opened 2022-02-04
Tiered Note Commitment Tree (re)Design


hdevalence:
## Context

Penumbra uses two Merkle trees: 

* one for the public chain state, storing key/value pairs of public data (like validator info, price info, IBC messages, etc.), and implemented using a Jellyfish Merkle tree;
* one for the private chain state, storing note commitments, and implemented using a SNARK-friendly incremental Merkle tree.

These trees are used differently: key/value pairs in the public chain state can be modified, and inclusion proofs work by providing the authentication path from the leaf node up to the root, but the private chain state is append-only, and inclusion proofs work by proving knowledge of an authentication path from the leaf node up to the root inside of a zk-SNARK, so that transactions can prove that some data *was included* without revealing *which data* was included.

Both are authenticated data structures, in the sense of this description from the [Jellyfish tree paper](https://diem-developers-components.netlify.app/papers/jellyfish-merkle-tree/2021-01-14.pdf):

<img width="769" alt="Screen Shot 2022-02-04 at 11 15 45 AM" src="https://user-images.githubusercontent.com/44879/152589452-41be2be7-255a-40ff-bd63-bddc495e3022.png">

However, there's somewhat of a role reversal between the (more conventional) Merkle tree of public state, and the (less conventional) Merkle tree of private state.  For the public state, the data structure is held by full nodes, who construct proofs of correctness when responding to queries made by clients; the clients hold an authenticator (in the form of the app hash, the root of the tree), and can verify those proofs.  But for the private state, clients seek to prove to the chain that their state updates are honestly constructed (in zero knowledge), so the *client* holds the data structure, and the *chain* (i.e., the full nodes) use a short authenticator to verify the client's proof-of-inclusion.

This means that every client must maintain a copy of the note commitment tree, so that they can construct the authentication proofs required to build their transactions, and this poses a practical difficulty, in that maintaining a full copy of the entire tree is resource-intensive and inefficient.

To address this, the current note commitment tree design and implementation (adopted from Zcash) allows the user to mark leaf nodes as "witnessed", and, while appending new leaves, forgets any leaf nodes that have not been explicitly witnessed, instead retaining only a minimal frontier that includes every witnessed leaf node.  This structure is referred to as a "bridge tree", since it stores the frontier as a sequence of arched "bridges" between adjacent leaf nodes.

This design is a huge space saving over a dense tree, but still has two issues:

[...]

PR #663, opened 2022-04-21
Written spec for TCT

### LINKED ISSUE #658: TCT Hashing Written Specification
Before proceeding with changes outlined in #657, there should be a written specification of the TCT's hashing behavior.

In particular, this specification does not need to involve itself with any of three important properties of the real implementation: its _sparsity_ (it stores only the minimum structure needed to produce authentication paths to commitments marked `Witness::Keep`), its _incrementalism_ (it takes amortized constant time to insert a new commitment into an existing tree), and its _laziness_ (it only computes hashes on demand).

Instead, the specification should describe two functions:

1. a mapping from a sequence (an eternity) of sequences (epochs) of sequences (blocks) of commitments (`Fq` elements), to a dense three-tiered quad-tree structure, without mention of how to compute internal hashes for the tree's nodes, and
2. a mapping from any node within the tree (including leaf nodes) to its hash (an element of `Fq` also), defined inductively on the structure of the tree.

This specification should be added as a section to the mdbook for the penumbra spec.

plaidfinch:
Resolves #658.


docs/protocol/src/crypto.md
```diff
@@ -4,16 +4,21 @@ Penumbra uses the following cryptographic primitives, described in the following
...
 
 - The [Poseidon for BLS12-377](./crypto/poseidon.md) section describes parameter
 selection for an instantiation of [Poseidon], a SNARK-friendly sponge
-construction, over the BLS12-377 scalar field;
+construction, over the BLS12-377 scalar field.
+
+- The [Tiered Commitment Tree](./crypto/tct.md) section describes a sparse
+incremental three-tiered Merkle quad-tree which uses the above
+[Poseidon for BLS12-377](./crypto/poseidon.md) construction as its hash function,
+allowing zero knowledge proofs of inclusion in the tree.
```
hdevalence, line 21: The description of an "incremental three-tiered Merkle quad-tree" drops a lot of words that don't make much sense without already having read the section.  Contrast with the FMD description below, where the details of the construction are mostly skipped in favor of motivation. It would be better for the TOC text to read:

```suggestion
- The [Tiered Commitment Tree](./crypto/tct.md) section describes the Merkle tree Penumbra uses to record private state, designed to allow efficient zero-knowledge proofs of inclusion as well as efficient state synchronization for the clients who make proofs.
```

docs/protocol/src/crypto/tct.md
```diff
@@ -0,0 +1,383 @@
+# Tiered Note Commitment Tree
+
+## Motivation and Relevant Context
```
hdevalence, line 3: Hmm, the point of doing the design writeup in #491 was to already have a writeup of the problem context and motivation -- was there a reason not to include it here?

docs/protocol/src/crypto/tct.md
```diff
@@ -0,0 +1,383 @@
...
+
+Each commitment is represented by a single element from the $\mathbb{F}_q$ field of
+[BLS12-377](https://docs.rs/ark-ed-on-bls12-377/0.3.0/ark_ed_on_bls12_377/). The [Poseidon for
+BLS12-377](./poseidon.md) hash function is defined over that same field.
+
+### Elementary Hashes and Domain Separators
+
+We define the hash of each node in the tree in terms of two arities of the Poseidon hash function:
+
+- $\mathtt{hash_1} : \mathbb{F}_q \times \mathbb{F}_q \to \mathbb{F}_q$
```
hdevalence, line 194: Do we need to prehash the leaves?

docs/protocol/src/crypto/tct.md
```diff
@@ -0,0 +1,383 @@
...
+
+If a node has an $n$th child (for $n > 0$), it must have an $(n - 1)$th child (there cannot be gaps
+in the list of children).
+
+Each node must be _left-to-right filled_:
+
+- if its children are nodes, all its children but its rightmost child are _perfect_, and its
+  rightmost child is left-to-right filled (where a _perfect_ node has exactly four children and if its
+  children are nodes, they are also perfect) or
+- if its children are leaves or tiers, the node is left-to-right filled.
```
hdevalence, line 122: It would be clearer if this defined "leaf nodes" and "internal nodes", since conventionally, a node is either a leaf or an internal node, rather than being defined to exclude leaves.

docs/protocol/src/crypto/tct.md
```diff
@@ -0,0 +1,383 @@
...
+
+- if the sequence is empty, return the empty tier;
+- otherwise, do the following exactly 8 times:
+  - in left to right order, greedily partition the sequence into chunks of at most 4 items, so
+    that all but the last chunk has exactly 4 items, and the last chunk has at most 4 items;
+  - for each chunk, construct a single node from it; and
+  - unless done with the 8 iterations, repeat again, starting with this sequence of nodes (which is
+    now shorter by about a factor of 4)
+- if the input sequence was less than $4^8$ items long, the resultant sequence will now contain
+  exactly one node; return it as the top-most node contained in a tier.
```
hdevalence, line 169: I'm not sure this adds much over the formulas; I think we just need one definition of the tree structure.

docs/protocol/src/crypto/tct.md
```diff
@@ -0,0 +1,383 @@
...
+\mathtt{blake2b}(\mathtt{"penumbra.tct"})$ interpreted in little-endian order as an element of
+$\mathbb{F}_q$, i.e. the following hash output from
+[BLAKE2](https://datatracker.ietf.org/doc/html/rfc7693):
+
+```plain_text
+0x197e501077d94b004b29e91196b9da397b8acbc98ac2d13d9099ad99f5215e490f171a0406f223ed25bafe4bc0451b649ba4b3fe0374d4a0b1ffd2ead37f531e
+```
+
+...interpreted as a (large) integer in little-endian order, and reduced by the modulus of
+$\mathbb{F}_q$.
```
hdevalence, line 209: We can skip the test vectors here, since they're not enough to determine the implementation behavior anyways, and so it mostly serves to clutter the description.

Issue #613, opened 2022-04-15
Scaffold the wallet server state machine;


hdevalence:
The wallet service is responsible for using a full viewing key (keys?) to scan the chain state and build a local copy of the private part of the chain state that the FVK has access to.

Initially, we can design the implementation around the assumption that there is only a single FVK, set during initialization, to simplify the initial code (since the current client only ever works with one viewing key at a time, this is not a big deal).  However, when designing the gRPC interface (#614), we should parametrize all of the requests by a hash of the FVK, so that the protocol is compatible with a future where the wallet service has viewing keys for multiple accounts.

The wallet service will need to have a task that performs scanning and writes the results to SQLite.

The most basic level of scanning is the kind required for spending, which involves learning about new spendable notes and tracking which notes have been spent:

- Iterate over all of the `output::Body`ies in the block, and for each one:
  - append the note commitment to the note commitment tree;
  - trial decrypt the note ciphertext with our incoming viewing key (this can be derived from the FVK on demand). If successful, we should:
    - witness the (just inserted) note commitment in the NCT;
    - derive the nullifier for the note, using the FVK;
    - decrypt the diversifier to the diversifier index;
    - record the following data: the note, the nullifier (and its association with the note), the diversifier index (and its association with the note).
- Iterate over the nullifiers in the block, and for each one:
  - check if the nullifier is one we know;
  - if so, mark the associated note as spent, and also forget the entry in the NCT.

The data we need to store at this stage is:

- The note plaintexts, which are tuples of (diversifier, amount, asset id, transmission key, blinding factor).  It is probably also useful to store the (diversifier index, note commitment), even though those could in principle be computed from the other note data, so that the note plaintexts can be queried by note commitment or by diversifier index, and also probably useful to store the height at which the note was created (since this isn't possible to recover later without querying a remote server for specific user data).
- The relation between notes (identified by note commitment) and nullifiers.
- The state of each note, whether it is spent or unspent.

We also need to store some application state:

- the chain params (this can be a blob);
- the full viewing key (this can be a blob);
- the height we scanned up to.

PR #667, opened 2022-04-21
Scaffold wallet server data model and sync method

aubrika:
Breaking out sync functionality into 'pwalletd' from existing wallet/pcli code

Targets #613


wallet-next/src/wallet.rs
```diff
@@ -0,0 +1,144 @@
...
+/// The contents of the wallet file that share a spend authority.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(from = "serde_helpers::WalletHelper")]
+#[serde(into = "serde_helpers::WalletHelper")]
+pub struct Wallet {
+    /// A list of human-readable labels for addresses.
+    ///
+    /// The label at index `i` is used for the address with `DiversifierIndex(i)`.
+    address_labels: Vec<String>,
+    spend_key: SpendKey,
```
hdevalence, line 20: As a heads' up, we shouldn't have the wallet service hold onto a `SpendKey`, since it shouldn't have spending authority.  Instead, it should just hold a [`FullViewingKey`](https://rustdoc.penumbra.zone/main/penumbra_crypto/keys/struct.FullViewingKey.html), so that the wallet service isn't capable of spending funds.

wallet-next/src/wallet.rs
```diff
@@ -0,0 +1,144 @@
...
+            .payment_address(next_index.into());
+        (next_index, address, dtk)
+    }
+
+    /// Get address by index.
+    pub fn address_by_index(&self, index: usize) -> Result<(String, Address), anyhow::Error> {
+        let label = self
+            .address_labels
+            .get(index)
+            .ok_or_else(|| anyhow::anyhow!("no address with index {}", index))?;
```
hdevalence, line 81: The address labels are just client-side information, so if we kept them in the wallet service, we'd need a way to sync them back and forth between the wallet client and the wallet service.

But there's no need for the wallet service to track them at all.  Instead of working with the labels, or even with the indexes, the wallet service should just work with 11-byte [`DiversifierIndex`es](https://rustdoc.penumbra.zone/main/penumbra_crypto/keys/struct.DiversifierIndex.html), and let the wallet client figure out how to interpret them.

wallet-next/src/wallet.rs
hdevalence: #671 will be helpful for this.

hdevalence:
This PR has been open for long enough to start generating conflicts; it'd be good to get some version of this code merged, so we can keep iterating on it in-tree.

wallet-next/src/bin/pwalletd.rs
```diff
@@ -1,22 +1,210 @@
...
+        /// The full viewing key
+        #[structopt(short, long)]
+        fvk: String,
+    },
+    Start {
+        /// Bind the services to this host.
+        #[structopt(short, long, default_value = "127.0.0.1")]
+        host: String,
+        /// Bind the wallet gRPC server to this port.
+        #[structopt(short, long, default_value = "26668")]
```
hdevalence, line 49: Let's set this to `8081` (to be different but similar to the port `8080` used by `pd`'s gRPC server).

wallet-next/src/bin/pwalletd.rs
```diff
@@ -1,22 +1,210 @@
...
+        grpc_port: u16,
```
hdevalence, line 50: Let's rename this `wallet_port`, since there are multiple gRPC servers in play.

wallet-next/src/bin/pwalletd.rs
```diff
@@ -1,22 +1,210 @@
...
+        /// The address of the pd+tendermint node.
+        #[structopt(short, long, default_value = "testnet.penumbra.zone")]
+        node: String,
+        /// The port to use to speak to pd's light wallet server.
+        #[structopt(short, long, default_value = "26666")]
+        oblivious_query_port: u16,
+        /// The port to use to speak to pd's thin wallet server.
+        #[structopt(short, long, default_value = "26667")]
+        specific_query_port: u16,
```
hdevalence, line 59: These options should be folded together into a single `pd_port` with default value `8080`, now that #730 unified them into a single endpoint.

wallet-next/src/bin/pwalletd.rs
```diff
@@ -1,22 +1,210 @@
...
+            //             Some(remote_addr) => tracing::error_span!("wallet_query", ?remote_addr),
+            //             None => tracing::error_span!("wallet_query"),
+            //         })
+            //         .add_service(WalletServer::new(storage.clone()))
+            //         .serve(
+            //             format!("{}:{}", host, wallet_query_port)
+            //                 .parse()
+            //                 .expect("this is a valid address"),
+            //         ),
+            // );
```
hdevalence, line 130: This logic isn't going to be what we really want (we'll need to start and manage tasks to run the wallet service), so let's replace the whole block with an `unimplemented!()` and fill it in later -- there shouldn't be much code in the `pwalletd` binary itself anyways, since that's supposed to just be another way to run the common wallet service code that could also be in-process for `pcli`.

wallet-next/src/bin/pwalletd.rs
```diff
@@ -1,22 +1,210 @@
...
+            storage.commit()?;
+            tracing::info!(height = ?storage.last_block_height().await.unwrap(), "syncing...");
+        }
+    }
+
+    storage.prune_timeouts();
+    storage.commit()?;
+    tracing::info!(end_height = ?storage.last_block_height().await.unwrap(), "finished sync");
+    Ok(())
+}
```
hdevalence, line 172: This method should be deleted, since it's copied from `pcli`, but its state model here is different, and this code doesn't fit (we don't want a bare method on a `Storage` object, we need a way to instantiate a service that does the syncing).

wallet-next/src/bin/pwalletd.rs
```diff
@@ -1,22 +1,210 @@
...
+    for asset in assets.0 {
+        storage
+            .asset_cache_mut()
+            .extend(std::iter::once(asset.denom));
+    }
+
+    storage.commit()?;
+    tracing::info!("updated asset registry");
+    Ok(())
+}
```
hdevalence, line 189: This method can be deleted, since we don't currently record any information about asset IDs and denominations in the state we have (it's all public state, anyways).

wallet-next/src/bin/pwalletd.rs
```diff
@@ -1,22 +1,210 @@
...
+    tracing::info!(?params, "saving chain params");
 
-    println!(
-        "Hello, pwalletd! I got stuff from sqlite: row {} value {}",
-        row, x
-    );
+    *storage.chain_params_mut() = Some(params);
+    storage.commit()?;
     Ok(())
 }
```
hdevalence, line 210: This method is also probably deletable; we'll need to get the chain params, but we won't want to do so here, and the API this implementation uses (using a `_mut()` method to get a mutable reference, setting a value there, and then calling a commit function) isn't compatible with the state model we're using (SQLite queries).  So there's not much code to save for reuse later other than the `client.chain_params` invocation, which we can redo whenever.

wallet-next/src/storage.rs
```diff
@@ -19,9 +23,9 @@ impl Storage {
         // Insert the task, then obtain the ID of this row
         let id = sqlx::query!(
             r#"
-INSERT INTO penumbra ( value )
-VALUES ( ?1 )
-        "#,
+            INSERT INTO penumbra ( value )
+            VALUES ( ?1 )
+            "#,
```
hdevalence, line 28: This whole method can be dropped, since we don't need it for testing the sqlx queries any more.

wallet-next/src/storage.rs
```diff
@@ -34,15 +38,69 @@ VALUES ( ?1 )
...
         let recs = sqlx::query!(
             r#"
-SELECT id, value
-FROM penumbra
-ORDER BY id
-LIMIT 1
+            SELECT id, value
+            FROM penumbra
+            ORDER BY id
+            LIMIT 1
```
hdevalence, line 44: We can also drop this whole method, for the same reason as above.

wallet-next/src/storage.rs
```diff
@@ -34,15 +38,69 @@ VALUES ( ?1 )
...
+            r#"
+            SELECT bytes
+            FROM chain_params
+            LIMIT 1
+        "#
+        )
+        .fetch_all(&self.pool)
+        .await?;
+
+        Ok(result[0].value.clone())
```
hdevalence, line 78: This method should return a `Result<ChainParams>` rather than a `Result<Vec<u8>>`; I think that all that's needed to do that is to replace this line with
```
Ok(ChainParams::decode(&result[0].value)?)
```
using the [`decode` method from the `Protobuf` marker trait](https://rustdoc.penumbra.zone/main/penumbra_proto/trait.Protobuf.html#method.decode)

wallet-next/src/storage.rs
```diff
@@ -34,15 +38,69 @@ VALUES ( ?1 )
...
+            r#"
+            SELECT bytes
+            FROM full_viewing_key
+            LIMIT 1
+            "#
+        )
+        .fetch_all(&self.pool)
+        .await?;
+
+        Ok(result)
```
hdevalence, line 91: Similarly, this should return a `FullViewingKey`, which should be deserializable from proto-encoded bytes in the same way.

wallet-next/src/storage.rs
```diff
@@ -34,15 +38,69 @@ VALUES ( ?1 )
...
+            r#"
+            SELECT bytes
+            FROM note_commitment_tree
+            LIMIT 1
+            "#
+        )
+        .fetch_all(&self.pool)
+        .await?;
+
+        Ok(result)
```
hdevalence, line 104: This should return a `NoteCommitmentTree`; unlike the other methods, the `incrementalmerkletree`-provided `NoteCommitmentTree` isn't proto-serialized, but it can be serialized with `bincode` similarly to the existing `get_nct` and `put_nct` methods on the shielded pool component by doing
```
let nct_data = bincode::serialize(&self.note_commitment_tree)?;
```
on the way in and
```
bincode::deserialize(&bytes).map_err(Into::into)
```
on the way out.

This is awkward, but any serialization code for the current NCT will be rewritten soon when we implement the NCT using the TCT library, so that serialization code has a lifespan of at most several weeks.

wallet-next/src/wallet.rs
```diff
@@ -0,0 +1,73 @@
...
+    /// Computes the change address for the given note.
+    pub fn change_address(&self, note: &Note) -> Result<Address, anyhow::Error> {
+        let index: u64 = self
+            .incoming_viewing_key()
+            .index_for_diversifier(&note.diversifier())
+            .try_into()
+            .context("cannot convert DiversifierIndex to u64")?;
+        self.address_by_index(index as usize)
+    }
+}
```
hdevalence, line 73: This whole file should be deleted, since we're going to have the wallet server only store a single FVK for now, so we don't need it to manage all of this information.

wallet-next/src/sync.rs
```diff
@@ -0,0 +1,93 @@
...
+    pub diversifier: Vec<u8>,
+    pub amount: u64,
+    pub asset_id: Vec<u8>,
+    pub transmission_key: Vec<u8>,
+    pub blinding_factor: Vec<u8>,
+    // precomputed decryption of the diversifier
+    pub diversifier_index: Vec<u8>,
+    // the nullifier for this note, used to detect when it is spent
+    pub nullifier: Vec<u8>,
+}
```
hdevalence, line 24: Let's move this out to `src/note_record.rs` and make the `NoteRecord` be a domain type for the [`NoteRecord` proto type](https://rustdoc.penumbra.zone/main/penumbra_proto/wallet/struct.NoteRecord.html) introduced in #732:
```rust
pub struct NoteRecord {
    pub note_commitment: note::Commitment,
    pub note: Note,
    pub diversifier_index: DiversifierIndex,
    pub nullifier: Nullifier,
    pub height_created: u64,
    pub height_spent: Option<u64>,
}

impl Protobuf<pb::NoteRecord> for NoteRecord {}
```

Folding the note contents up into the existing `Note` domain type adds a tiny amount of code when moving in/out of the database, since the `Note` has to [be built `from_parts`](https://rustdoc.penumbra.zone/main/penumbra_crypto/note/struct.Note.html#method.from_parts), but the upshot is saving a lot of code for protoizing the data for downstream consumers.

wallet-next/src/sync.rs
```diff
@@ -0,0 +1,93 @@
...
+            let record = NoteRecord {
+                note_commitment: note_commitment.encode_to_vec(),
+                height_spent: None,
+                height_created: height,
+                diversifier: note.diversifier().0.to_vec(),
+                amount: note.amount(),
+                asset_id: note.asset_id().encode_to_vec(),
+                transmission_key: note.transmission_key().0.to_vec(),
+                blinding_factor: note.note_blinding().0.to_vec(),
+                diversifier_index: note.diversifier().0.to_vec(),
```
hdevalence, line 81: The `DiversifierIndex` is the decryption of the `Diversifier`, so it has to [be computed](https://rustdoc.penumbra.zone/main/penumbra_crypto/keys/struct.IncomingViewingKey.html#method.index_for_diversifier); probably the easiest way is something like
```
fvk.incoming().index_for_diversifier(note.diversifier())
```

wallet-next/src/sync.rs
```diff
@@ -0,0 +1,93 @@
...
+
+            let record = NoteRecord {
+                note_commitment: note_commitment.encode_to_vec(),
+                height_spent: None,
+                height_created: height,
+                diversifier: note.diversifier().0.to_vec(),
+                amount: note.amount(),
+                asset_id: note.asset_id().encode_to_vec(),
+                transmission_key: note.transmission_key().0.to_vec(),
+                blinding_factor: note.note_blinding().0.to_vec(),
```
hdevalence, line 80: This code will get cleaner once the `NoteRecord` is a domain type that contains a `Note`, since the note can just be used directly.

xref https://github.com/penumbra-zone/penumbra/pull/667/files#r862422793

hdevalence:
Minimal path to mergeability:

- [ ] delete the code copied over from `pcli` that doesn't really fit (marked in review comments)
- [ ] move the `NoteRecord` struct out to its own file, and make it a domain type for the proto type introduced in #732 (as marked in comments, also needs `From`/`TryFrom` impls)
- [ ] compute the `DiversifierIndex` using the incoming viewing key, rather than copying the diversifier bytes
- [ ] update ports and command-line options

It's not essential to get the rest of the sync process working, or the interaction with the sqlite state working, since we can do those in follow-on PRs.

PR #671, opened 2022-04-22
crypto: make FullViewingKey serializable

hdevalence:
This will be useful for things like `pwalletd` where we want to give only viewing capability.


aubrika (approved):
LGTM!

PR #674, opened 2022-04-22
Add validator spec doc based on original ticket

zbuc:
There was nothing in the spec about how validators work and the original ticket seemed like a good starting place: https://github.com/penumbra-zone/penumbra/issues/375


hdevalence (approved):

hdevalence:
This looks good. Eventually, we'll probably also want to have another section that describes exactly how all of the actions work and what their consensus rules are.

PR #678, opened 2022-04-23
Implement downtime-based slashing

### LINKED ISSUE #664: Implement uptime-based slashing for validators

Currently, we don't track whether validators participate in block signing, so we can't slash them when they go offline.  This means nothing prevents someone from "running" a validator but not actually participating in the network.

- [ ] Learn what criteria the Cosmos SDK uses to slash validators for downtime
- [ ] Copy exactly those criteria into the staking component.

This isn't absolutely a blocker for multi-validator testnets, but we should do it soonish.

hdevalence:
Closes #664


hdevalence:
I tested this locally on a single validator testnet, but haven't tested a multi-validator setup to see that one of the nodes gets slashed if it goes offline.

hdevalence:
Going to merge this optimistically so it's easy to test on `testnet-next.penumbra.zone`, but review would still be good

PR #682, opened 2022-04-24
guide: explain how to get the latest node id when joining testnet

hdevalence:
This way we don't have to constantly update the guide text.

### DIFF
--- a/docs/guide/src/pd/join-testnet.md
+++ b/docs/guide/src/pd/join-testnet.md
@@ -24,28 +24,31 @@ config	data
 config.toml	genesis.json	node_key.json
 ```
 
+The `genesis.json` file Tendermint generates needs to be replaced with the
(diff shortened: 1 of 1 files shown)


zbuc:
I have some changes to this documentation in https://github.com/penumbra-zone/penumbra/pull/654 as well, I was planning to merge it alongside adding the testnet allocations

hdevalence:
@zbuc ah okay, those docs changes look a lot better, any reason not to merge that PR now (and update the allocations or make whatever other changes needed later) rather than leaving it as a draft?

zbuc:
Not really, no, it will just make the `main` branch default to the new
testnet which isn't a big deal since we have the tag as well.

hdevalence:
Cool, I'll close this PR then and port the changes over to that one.

PR #683, opened 2022-04-24
pd: initial tokio-console integration

### LINKED ISSUE #13: Basic tokio-console support.
Add support for `tokio-console`, allowing inspection of running tasks.

hawkw:
This commit adds initial support for the [Tokio Console], using the
basic [`console_subscriber::init`] API.  This should not interfere with
the existing logging configuration using `tracing`, as
`console_subscriber::init` is essentially a superset of
`tracing_subscriber::fmt::init`. Note that the console filter
configuration does not interfere with the filter configuration for
logging.

The console endpoint is served on the default address,
http://127.0.0.1:6669/. 

## Future Work

Future commits will improve the console instrumentation by configuring
the console endpoint using `pd`'s command line arguments, adding the
filter reloading described in
https://github.com/penumbra-zone/penumbra/issues/13#issuecomment-1100969950,
and integrate Tokio Console with the existing Grafana setup, but this is
a working starting point.

## Examples

image

Note that:
- Only the `pd=trace` filter setting is logged (tokio events are not
  logged)
- Named tasks show up in the console
- Blocking tasks (from rocksdb) are also displayed in the console

Closes #13

[Tokio Console]: https://docs.rs/tokio-console/latest/tokio_console/
[`console_subscriber::init`]: https://docs.rs/console-subscriber/latest/console_subscriber/fn.init.html

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,0 +1,3 @@
+[build]
+# Enable Tokio's `tracing` support for `tokio-console`
+rustflags = ["--cfg", "tokio_unstable"]
--- a/pcli/Cargo.toml
+++ b/pcli/Cargo.toml
@@ -41,7 +41,7 @@ tower = { version = "0.4", features = ["full"]}
 tracing = "0.1"
 structopt = "0.3"
 tonic = "0.6.1"
-tracing-subscriber = "0.2"
+tracing-subscriber = "0.3"
 pin-project = "1"
 serde_json = "1"
 serde = { version = "1", features = ["derive"] }
--- a/pd/Cargo.toml
+++ b/pd/Cargo.toml
@@ -48,7 +48,7 @@ regex = "1.5"
 prost-types = "0.9"
 structopt = "0.3"
 tonic = "0.6.1"
-tracing-subscriber = "0.2"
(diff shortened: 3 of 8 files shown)


hdevalence:
Question about the `console_subscriber::init` and the endpoint -- I think it'd be good to have a command-line option for the console endpoint and use its presence/absence to control whether the console is enabled.  Presumably we could do that by only doing `console_subscriber::init` if the option is present?

hawkw:
> Question about the `console_subscriber::init` and the endpoint -- I think it'd be good to have a command-line option for the console endpoint and use its presence/absence to control whether the console is enabled. Presumably we could do that by only doing `console_subscriber::init` if the option is present?

Yes, I think that would be ideal; I didn't do it in the initial PR as it would involve using the `console-subscriber` builder API, requiring the tracing setup code to be rewritten. I thought I'd do this in a follow-up branch along with the filter reloading stuff you mentioned in issue #13. I was going to do this in a follow up because I thought it would be nice to ensure that the initial instrumentation gets into the next testnet for sure.

We should maybe open issues for additional follow-up work like that?

hdevalence (approved):
Thanks!

PR #684, opened 2022-04-24
pd: fail service if consensus worker task panics

hawkw:
Currently, if the consensus worker task panics (or terminates
unexpectedly), the rest of the `pd` process will simply continue
running. This is because the consensus `Service` calls `.expect(...)` on
the oneshot reciever from the worker task, and the `tower_acbi` server
may spawn the response future in a background task whose panicking may
not be propagated. Additionally, it is possible that the `Service` may
be driven to readiness successfully before the worker panicked or
terminated. If this occurs, the message will be sent successfully, but
the worker task may never respond.

This branch should fix these issues by changing the consensus service
to:
1. proactively check if the worker is still there in `call`, and return
   a failed response future if the worker has died;
2. return a response future that completes with an error if the oneshot
   receiver is dropped, rather than panicking, which should ensure the
   request fails.

Hopefully, this means that `tower-acbi` will propagate the failure up to
the main function and terminate the application gracefully.

### DIFF
--- a/pd/src/consensus/service.rs
+++ b/pd/src/consensus/service.rs
@@ -52,6 +52,16 @@ impl tower::Service<ConsensusRequest> for Consensus {
     }
 
     fn call(&mut self, req: ConsensusRequest) -> Self::Future {
+        // Check if the worker has terminated. We do this again in `call`
+        // because the worker may have terminated *after* `poll_ready` reserved
+        // a send permit.
+        if self.queue.is_closed() {
+            return async move {
+                Err(anyhow::anyhow!("consensus worker terminated or panicked").into())
+            }
+            .boxed();
+        }
+
         let span = req.create_span();
         let (tx, rx) = oneshot::channel();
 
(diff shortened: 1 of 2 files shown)


hdevalence:
I checked what goes on `tower-abci`, and it looks like there's also something to patch up there to fully propagate issues.

The `pd` code spawns a `Server::listen` task, which is internally implemented as:
```rust
    pub async fn listen<A: ToSocketAddrs + std::fmt::Debug>(self, addr: A) -> Result<(), BoxError> {
        tracing::info!(?addr, "starting ABCI server");
        let listener = TcpListener::bind(addr).await?;
        let local_addr = listener.local_addr()?;
        tracing::info!(?local_addr, "bound tcp listener");

        loop {
            match listener.accept().await {
                Ok((socket, _addr)) => {
                    let conn = Connection {
                        consensus: self.consensus.clone(),
                        mempool: self.mempool.clone(),
                        info: self.info.clone(),
                        snapshot: self.snapshot.clone(),
                    };
                    tokio::spawn(async move { conn.run(socket).await.unwrap() });
                }
                Err(e) => {
                    tracing::warn!({ %e }, "error accepting new tcp connection");
                }
            }
        }
    }
```
so when one of the connections fails, nothing happens, because the per-connection task panic isn't propagated anywhere.  To fix that, I guess we'd need to have the `listen` method keep track of the join handles for the tasks it spawns, and then wait for either a new connection, or a result from one of the tasks, and propagate errors that way.

PR #695, opened 2022-04-24
ci: update deploy workflow

hdevalence:
The previous workflow broke with #691; I used `rg testnet_build` to try to find
all references to the old paths, but the deployment workflow is in the
`.github/` directory, which ripgrep ignores by default.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -37,11 +37,7 @@ jobs:
         script: |
           cd ~/penumbra
           docker-compose stop
-          rm -rf ~/scratch/testnet_build/node0/tendermint/data/
-          rm -rf ~/scratch/testnet_build/node0/pd/rocksdb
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -37,11 +37,7 @@ jobs:
         script: |
           cd ~/penumbra
           docker-compose stop
-          rm -rf ~/scratch/testnet_build/node0/tendermint/data/
-          rm -rf ~/scratch/testnet_build/node0/pd/rocksdb
-          mkdir ~/scratch/testnet_build/node0/tendermint/data/
-          echo -n '{}' > ~/scratch/testnet_build/node0/tendermint/data/priv_validator_state.json
-          chmod -R 777 ~/scratch/testnet_build
```
hdevalence, line 44: @zbuc I wasn't sure why this needed to manually construct this data, rather than relying on `generate-testnet`, but it seems to work just to just blow away the whole testnet directory.

PR #719, opened 2022-04-27
add IBC ConnectionOpenTry message handling

avahowell:


ibc/src/client.rs
```diff
@@ -246,3 +256,90 @@ impl From<ClientConnections> for pb::ClientConnections {
...
+        ));
+    }
+
+    // check that the trust level is correct
+    validate_trust_threshold(tm_client_state.trust_level)?;
+
+    // TODO: check that the unbonding period is correct
+    //
+    // - check unbonding period is greater than trusting period
+    if tm_client_state.unbonding_period < tm_client_state.trusting_period {
```
hdevalence, line 336: What's the relation between `tm_client_state.unbonding_period` and the `unbonding_epochs` in our chain parameters? We need these to be aligned, right?

hdevalence (approved):
This seems reasonable but it's a little hard to follow the logic from the diff. I'm wondering if it would make sense to merge this optimistically now, then next week, move the IBC component code into the `ibc` crate, and pair on doing a global read-through / docs / (re)structuring pass?

Issue #709, opened 2022-04-26
Restructure `pd` to eliminate the split between "X crates" and "X components"


hdevalence:
Following the big refactoring in #488, we've ended up with a much better architecture for the Penumbra application logic, with an `App` that holds a bunch of `Component`s; the `Components` access a common `Overlay` over the backing `Storage`, and expose `View`s that read and write data.

However, this means that we end up having a bit of a weird split between our crates that handle some area of the application logic, and the components that actually implement that logic in `pd`.  For instance, we have a `stake` crate that has staking-related structures, but the logic about validator state transitions is handled by the staking component in the `pd` crate.

It would be cleaner if we could have each component in a different crate, because:
- we could avoid a split between the structures for that part of the logic, and the logic itself;
- we could enforce privacy boundaries between components, by choosing whether or not to export a `View` trait from the component's crate;
- we could enforce (non-cycles) of dependencies between components; if one component needs to use another component's data, it has to depend on that component's crate, and that can't be cyclic.

To do this, we'd pull out the current contents of the `pd` crate as follows:

- Create a `penumbra-storage` crate that would have the `Storage` wrapper around RocksDB, as well as the `Overlay` type definition, and the `OverlayExt` trait that handles reading/writing domain and proto types.  The `penumbra-storage` crate would depend just on `penumbra-proto` to get the proto trait definitions.
- Create a `penumbra-component` crate that would have the definition of the `Component` trait.  This crate would depend on `penumbra-storage`.
- Move the staking component code into the `penumbra-stake` crate, and add a dependency on `penumbra-component` to `penumbra-stake`.
- Move the ibc component code into the `penumbra-ibc` crate, and add a dependency on `penumbra-component` to `penumbra-ibc`.
- Create a new `penumbra-shielded-pool` crate, dependent on `penumbra-component` and move the shielded pool component into it.
- Change the `pd` code so that only the `App` is present, and built out of components from other crates.

This refactoring would be fairly mechanical, but obviously clobbers any work that touches any of the components or any of the other crates, so if we decide to do it, we should plan when to slip it in.

Update: The groundwork for this is done now; what's left is moving the IBC component and the staking components out into their crates:

- [x] #740
- [x] #741

hdevalence:
As written above, this doesn't work, because the `Component` trait needs the `Transaction` type, and the `Transaction` type uses actions from different crates (e.g. `Delegation` is from the stake crate, etc). 

As a start, though, we can pull out the storage and the component traits.

[...]

PR #722, opened 2022-04-27
Break dependencies to let components live in their crates

hdevalence:
A blocker for #709 is that the transaction crate can't depend on any of the component crates.  This PR breaks that dependency so that we can move components into the topic crates.

To do this, the following changes were required:

- The `Delegate` and `Undelegate` actions move to `transaction`;
- The `DelegationToken` and the `IdentityKey` moves to `crypto`;
- The `ValidatorDefinition` action now only holds a proto, not a domain type;
- The `IBCAction` now only holds a proto, instead of a wrapper around a proto.


zbuc (approved):

avahowell:
LGTM

PR #731, opened 2022-04-30
update to upstream tendermint 0.24.0-pre.1

avahowell:
this PR unifies the version of tendermint-rs that we are using to `0.24.0-pre.1`, across ibc-rs and the main penumbra crates.


hdevalence (approved):

hdevalence:
@avahowell This looks good, I pushed up an extra commit that removes the hardcoded version from the notes deploy workflow, now that there's just one version.

PR #732, opened 2022-05-01
Rework wallet protocol

### LINKED ISSUE #614: Define a new gRPC interface for the wallet server in the `proto` crate
From #587:

> - implement a gRPC service that:
>  - buffers all requests until the service is ready (synchronized with the chain state);
>  - provides queries for private chain state (UTXOs, balances, transactions, ...);
>  - ~~provides higher-level transaction creation methods for common use cases (sending funds, delegating, undelegating, IBC relaying ...) that build, sign, and submit the transaction;~~

A first pass of this is in #606.

@hdevalence:

> However, after thinking about it a bit more, I think my original [suggestion](https://github.com/penumbra-zone/penumbra/pull/606#discussion_r850881851) wasn't a good one, and we should keep refining the design before we move to implementing it. Specifically, I think that the last bullet point on the original list will mix responsibilities, and block us from having a simple, clean data model.

# Design Questions

- Specific/Oblivious: how is this enforced in code? What does it mean for developers? Is specific e2e encrypted?
- pcli breakdown: should we be designing the wallet service/wallet client/custody service to be composable and utilizable by any other component? i.e. the pcli boxes variously encapsulate these components -- are they standalone binaries communicated w/ via gRPC, or crates compiled into `pcli`?
- Development priority: what is most useful to build first? Should we leave the current "custody" logic within `pcli` for now and focus on the wallet server, or build a new custody service in parallel?
- Any organizational opinions on where this logic lives? New crates? `pcli` crate?

# Todo

- [ ] Migrate sync logic and fullnode interface (oblivious/specific) into pwalletd
- [ ] Extract client functionality into wallet client

hdevalence:
This brings the wallet protocol in line with the design in #587, where we limit
the responsibilities of the wallet service to just sync and scan the public
state, leaving all of the logic of transaction creation to the client and
transaction signing to the (future) custody service.

In the future, this RPC could be extended with other parts of private chain
state (e.g., memo contents) that we don't currently download.

Closes #614


crypto/src/keys/fvk.rs
```diff
@@ -81,6 +86,15 @@ impl FullViewingKey {
     pub fn spend_verification_key(&self) -> &VerificationKey<SpendAuth> {
         &self.ak
     }
+
+    /// Hashes the full viewing key into a [`FullViewingKeyHash`].
+    pub fn hash(&self) -> FullViewingKeyHash {
+        let hash_result = prf::expand(b"Penumbra_HashFVK", &self.nk.0.to_bytes(), self.ak.as_ref());
```
hdevalence, line 92: This is the same way that the outgoing viewing key and diversifier key are derived from the FVK, but with a different label to ensure different outputs.

proto/proto/wallet.proto
```diff
@@ -4,58 +4,29 @@ package penumbra.wallet;
...
 
+// The wallet protocol is used by a wallet client, who wants to do some
+// transaction-related actions, to request data from a wallet service, which is
+// responsible for synchronizing and scanning the public chain state with one or
+// more full viewing keys.
+//
+// Wallet protocol requests include a hash of the full viewing key, used to
+// identify which set of data to query.  This also works as a pseudo-auth system
+// (assuming transport security, the client has to know the FVK to request its
+// data).  (TODO: refine this)
```
hdevalence, line 15: The idea behind including a hash of the FVK is:

(1) it means that the protocol doesn't bake in the assumption that there's only one FVK per wallet service, even if we make that assumption in the current implementation;

(2) it's marginally better than sending the FVK as a way to identify the account data, since it doesn't reveal the full viewing key itself.  However, it's not *that* much better, since someone who learns the FVK hash (e.g., by being on the other end of a misconfigured server, by snooping on unencrypted network traffic, etc) can use the FVK hash as a bearer token to make requests to the wallet service (if they know which service to access).

proto/proto/wallet.proto
```diff
@@ -4,58 +4,29 @@ package penumbra.wallet;
...
 service WalletProtocol {
-    // TODO: do we want to include auth/signatures in all requests here?
-    /// Get current status of chain sync
-    rpc Status() returns (StatusResult);
+    // Get current status of chain sync
+    rpc Status(crypto.FullViewingKeyHash) returns (StatusResult);
```
hdevalence, line 18: Parameterizing this by the FVK hash means the service can report on sync progress of different accounts, if we decide to implement that.

proto/proto/wallet.proto
```diff
@@ -68,90 +39,56 @@ message StatusResult {
...
+    // The note plaintext itself.
+    crypto.Note note = 2;
+    // A precomputed decryption of the note's diversifier index.
+    crypto.DiversifierIndex diversifier_index = 3;
+    // The note's nullifier.
+    crypto.Nullifier nullifier = 4;
+    // The height at which the note was created.
+    uint64 height_created = 5;
+    // Records whether the note was spent (and if so, at what height).
+    optional uint64 height_spent = 6;
```
hdevalence, line 55: The `NoteRecord` type in #667 should turn into a domain type for this proto type.

proto/proto/wallet.proto
hdevalence: (To be clear, we should think of this as a first-pass mechanism, and figure out what a reasonable auth story looks like later. [Some ideas are here](https://twitter.com/FiloSottile/status/1520532532139339776).)

PR #736, opened 2022-05-02
use ibc enabled chain param to gate IBC txs

avahowell:
this PR updates the root IBC component to use the ibc_enabled chain param to enable or disable IBC txs


hdevalence:
Included in #742

PR #744, opened 2022-05-03
[WIP] move staking component to staking crate

zbuc:


shielded-pool/src/component.rs
```diff
@@ -565,6 +563,22 @@ pub trait View: StateExt {
...
+            .await
+    }
+
+    async fn set_commission_amounts(&self, height: u64, notes: CommissionAmounts) {
+        self.put_domain(
+            format!("staking/commission_amounts/{}", height).into(),
+            notes,
+        )
+        .await
+    }
```
hdevalence, line 581: @zbuc @avahowell I had to move these into the `shielded-pool` crate to avoid a circular dependency, but I'm wondering if it actually makes sense for it to be in the `shielded-pool` crate (with a different, more general name) so that we could use the same mechanism for IBC transfers as well as minting commission notes.

PR #747, opened 2022-05-03
Fix tendermint validator power handling

zbuc:
This refactors the tendermint validator power handling to be managed in-memory during the course of the block, making it easier to report only updates to tendermint, as well as not report 0 power for new, inactive validators.

### DIFF
--- a/stake/src/component.rs
+++ b/stake/src/component.rs
@@ -40,9 +40,40 @@ pub struct Staking {
     /// persisted at the end of the block for processing at the end of the next
     /// epoch.
     delegation_changes: DelegationChanges,
+    /// List of changes to the tendermint validator set accumulated throughout
+    /// this block, to be returned during `EndBlock`.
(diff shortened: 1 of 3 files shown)


stake/src/component.rs
```diff
@@ -40,9 +40,40 @@ pub struct Staking {
...
+    /// List of changes to the tendermint validator set accumulated throughout
+    /// this block, to be returned during `EndBlock`.
+    tm_validator_updates: Vec<ValidatorUpdate>,
 }
 
 impl Staking {
+    #[instrument(skip(self))]
+    fn update_tm_validator_power(&mut self, ck: &PublicKey, power: u64) -> Result<()> {
+        // TODO: would it be more sensible to have `tm_validator_updates` be a
+        // BTreeMap instead, and construct the `ValidatorUpdate` vec on-demand?
```
hdevalence, line 52: Yeah, this sounds reasonable.

PR #749, opened 2022-05-03
wallet-next: scaffold the wallet server state machine

### LINKED ISSUE #613: Scaffold the wallet server state machine;

hdevalence:
This commit scaffolds the basic wallet server state machine, which uses a
pub-sub architecture mediated by SQLite:

- a single `Worker` task runs indefinitely, scanning the chain, and writing the
  results to SQLite;
- any number of `WalletService` readers can serve gRPC requests.

Along the way, tidy the proto files slightly for consistency.

Closes #613

### DIFF
--- a/proto/proto/wallet.proto
+++ b/proto/proto/wallet.proto
@@ -15,10 +15,10 @@ import "crypto.proto";
 // data).  (TODO: refine this)
 service WalletProtocol {
     // Get current status of chain sync
-    rpc Status(crypto.FullViewingKeyHash) returns (StatusResult);
+    rpc Status(StatusRequest) returns (StatusResponse);
 
     // Queries for notes.
-    rpc Notes(NoteRequest) returns (stream NoteRecord);
+    rpc Notes(NotesRequest) returns (stream NoteRecord);
 
     // Returns authentication paths for the given note commitments.
     //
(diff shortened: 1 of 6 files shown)


aubrika (approved):
LGTM!

PR #772, opened 2022-05-04
add go protos, build script for go protos

avahowell:
this PR adds go codegen for the penumbra proto types to the repo, and a script for generating that Go code


avahowell:
the immediate use for this is so we can integrate with https://github.com/penumbra-zone/relayer/pull/1, but it may also be useful in the future if any other Go apps want to do stuff with penumbra domain types

hdevalence (approved):
Cool -- I think we might run into issues ensuring that the generated Go code is up to date (since our Rust code is generated from protos automatically), but we can revisit this setup if/when we do .

PR #776, opened 2022-05-05
Add commit logic to wallet-next storage

### LINKED ISSUE #751: Add method to `wallet_next::Storage` to ingest a `ScanResult` and `NoteCommitmentTree`
The `Storage` interface added in #613 only has individual read methods, not write methods. We don't want to have individual write methods, since we should only update the database state in two cases: 

1. when we initialize it with a new FVK;
2. when we scan a new block.

To do (2), in this issue, we should have a `record_block` method on `Storage`:
```
pub async fn record_block(&self, scan_result: ScanResult, nct: &mut NoteCommitmentTree) -> anyhow::Result<Self>
```
This would:

- [ ] Insert all of the new `NoteRecord`s into the `notes` table, destructuring the inner `Note` and serializing the fields as follows:
  - [ ] the `height_created` as the current `height`;
  - [ ] the `height_spent` as `None`;
  - [ ] the `note_commitment` as a `[u8; 32]` with `to_bytes`;
  - [ ] the `diversifier` as its inner `[u8; 11]`;
  - [ ] the `amount` as an integer;
  - [ ] the `asset_id` as a `[u8; 32]` with `to_bytes`;
  - [ ] the `transmission_key` as its inner `[u8; 32]`;
  - [ ] the `blinding_factor` as a `[u8; 32]` with `to_bytes`;
  - [ ] the `diversifier_index` as its inner `[u8; 11]`;
  - [ ] the `nullifier` as a `[u8; 32]` with `to_bytes`;
- [ ] **Detect spends**, by using the set of nullifiers in the `ScanResult` to update any rows of the table with matching nullifiers to have `height_spent` set to `Some(height)`.
- [ ] Construct the list of note commitments that were marked as spent in the previous step, and forget them from the NCT (this is why it's `&mut`); 
- [ ] Serialize the current NCT state to bytes with `bincode::serialize(nct)`, and update the (single) row of the `note_commitment_tree` table with the encoding.

This should all be done transactionally, since we're using SQLite as the synchronization mechanism between the worker task and the wallet service.

aubrika:
Closes #751

### DIFF
--- a/wallet-next/src/storage.rs
+++ b/wallet-next/src/storage.rs
@@ -1,7 +1,13 @@
 use penumbra_chain::params::ChainParams;
-use penumbra_crypto::merkle::NoteCommitmentTree;
+use penumbra_crypto::{
+    merkle::{NoteCommitmentTree, Tree},
+    note::Commitment,
+    FieldExt,
+};
 use penumbra_proto::{crypto::FullViewingKey, Message, Protobuf};
 use sqlx::{query, Pool, Sqlite};
+
+use crate::sync::ScanResult;
 
 #[derive(Clone)]
 pub struct Storage {
@@ -27,10 +33,10 @@ impl Storage {
             LIMIT 1
         "#
         )
-        .fetch_all(&self.pool)
+        .fetch_optional(&self.pool)
         .await?;
 
-        Ok(result[0].height.map(|h| h as u64))
+        Ok(result.map(|row| row.height as u64))
     }
 
     pub async fn chain_params(&self) -> anyhow::Result<ChainParams> {
@@ -41,10 +47,10 @@ impl Storage {
             LIMIT 1
         "#
         )
-        .fetch_all(&self.pool)
+        .fetch_one(&self.pool)
         .await?;
 
-        ChainParams::decode(result[0].bytes.as_ref().unwrap().as_slice())
+        ChainParams::decode(result.bytes.as_slice())
     }
 
     pub async fn full_viewing_key(&self) -> anyhow::Result<FullViewingKey> {
@@ -55,12 +61,10 @@ impl Storage {
             LIMIT 1
             "#
         )
-        .fetch_all(&self.pool)
+        .fetch_one(&self.pool)
         .await?;
 
-        Ok(FullViewingKey::decode(
-            result[0].bytes.as_ref().unwrap().as_slice(),
-        )?)
+        Ok(FullViewingKey::decode(result.bytes.as_slice())?)
     }
 
     pub async fn note_commitment_tree(&self) -> anyhow::Result<NoteCommitmentTree> {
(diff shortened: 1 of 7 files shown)


wallet-next/src/storage.rs
```diff
@@ -80,4 +82,75 @@ impl Storage {
...
+                nullifier
+            )
+            .fetch_all(&self.pool)
+            .await?;
+
+            //Forget spent note commitments from the NCT
+
+            let spent_commitment = Commitment::try_from(result[0].note_commitment.as_slice())?;
+
+            todo!("forget spent commitments appropriately - unclear how to do this, simply remove?")
```
hdevalence, line 141: This is done with
```
note_commitment_tree.remove_witness(&spent_commitment);
```

wallet-next/src/storage.rs
```diff
@@ -80,4 +86,73 @@ impl Storage {
...
+                note_record.note.diversifier().0,
+                note_record.note.amount(),
+                note_record.note.asset_id().to_bytes(),
+                note_record.note.transmission_key().0,
+                note_record.note.note_blinding().to_bytes(),
+                note_record.diversifier_index.0,
+                note_record.nullifier.to_bytes()
+            );
+
+            let _result = &self.pool.execute(&*query).await?;
```
hdevalence, line 122: This is going to do each update on its own, rather than having them be done transactionally, right? Can we use a transaction instead so that other readers of the state see it ratchet forward an entire block at a time?

wallet-next/src/storage.rs
```diff
@@ -80,4 +86,73 @@ impl Storage {
...
+            nct.remove_witness(&Commitment::try_from(result[0].note_commitment.as_slice())?);
+        }
+
+        // Update NCT table with current NCT state
+
+        let query = format!(
+            "UPDATE note_commitment_tree
+            SET bytes = {:?}",
+            bincode::serialize(nct)
+        );
```
hdevalence, line 152: Any reason not to use `query!` here rather than building a string?

aubrika: Either can work, but the string interpolation for `query!` doesn't support inline functions, so, personal preference vs. serializing the `nct` into a variable beforehand, I suppose. `query!` also doesn't infer string formatting for every type, so that may also need to be done outside the macro. No strong opinions either way, I still haven't found a way to use `sqlx` that feels succinct & maximally readable to me.

wallet-next/src/storage.rs
aubrika: I didn't see an exactly straightforward way to parameterize a set of sets of distinct values using `sqlx` here - we have to iterate over all the `NoteRecords` and map each the `NoteRecord` to the `notes` table rows, but I think we could write some dynamic SQL generation logic to build a valid query string of arbitrary length. One way we might be able to do this is to concatenate the `query` string or sub-string built for each `NoteRecord` into one long query and execute that as a single transaction.

wallet-next/src/storage.rs
```diff
@@ -80,4 +86,76 @@ impl Storage {
...
+
+            let result = query!(
+                r#"UPDATE notes 
+                SET height_spent = ?
+                WHERE nullifier = ?
+                RETURNING note_commitment"#,
+                height_spent,
+                nullifier
+            )
+            .fetch_all(&self.pool)
```
aubrika, line 141: Unsure how to include this in the `note_record` insert transaction - because this query is returning the `note_commitment` for each spent nullifier that we need to forget the corresponding NCT entry, if we execute these in a batch with the note inserts, I think we'd need to refactor to return a set of the needed `note_commitment`s on transaction commit, then update the NCT table in a separate transaction - seems doable, but is it preferable to doing this part sequentially?

PR #777, opened 2022-05-06
ibc: refactor connection handshake validation

avahowell:
Similarly to #774 , this PR refactors IBC connection handshake validation into new `stateful` and `stateless` modules, and adds documentation for the IBC handshake.

### DIFF
--- a/ibc/src/component/connection.rs
+++ b/ibc/src/component/connection.rs
@@ -3,7 +3,6 @@ use async_trait::async_trait;
 use ibc::core::ics02_client::client_consensus::ConsensusState;
 use ibc::core::ics02_client::client_def::AnyClient;
 use ibc::core::ics02_client::client_def::ClientDef;
-use ibc::core::ics02_client::client_state::AnyClientState;
 use ibc::core::ics02_client::client_state::ClientState;
 use ibc::core::ics03_connection::connection::Counterparty;
 use ibc::core::ics03_connection::connection::{ConnectionEnd, State as ConnectionState};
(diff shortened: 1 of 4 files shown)


ibc/src/component/connection.rs
```diff
@@ -55,7 +55,26 @@ impl Component for ConnectionComponent {
...
+                Some(ConnectionOpenTry(msg)) => {
+                    let _ = MsgConnectionOpenTry::try_from(msg.clone())?;
+                }
+                Some(ConnectionOpenAck(msg)) => {
+                    let _ = MsgConnectionOpenAck::try_from(msg.clone())?;
+                }
+
+                Some(ConnectionOpenConfirm(msg)) => {
+                    let _ = MsgConnectionOpenConfirm::try_from(msg.clone())?;
+                }
```
hdevalence, line 74: Are these supposed to have more checks added later? If so, could we have a TODO comment, so that when reading the code we can assume that either the checks listed there are all of the checks, or else there's an explicit marker that some are left to be implemented?

avahowell: I wasn't sure if there was any stateless validation that could be pulled out of the stateful validation, but there is for OpenTry and OpenAck (that client state, client proof, and consensus proof are not None), so I pushed a commit to pull those validations into the stateless mod and document that OpenConfirm has no additional stateless validation

ibc/src/component/connection/stateful.rs
```diff
@@ -0,0 +1,496 @@
...
+                .await?
+                .0;
+
+            let client_def = AnyClient::from_client_type(trusted_client_state.client_type());
+
+            // PROOF VERIFICATION
+            // note: here we only verify connection state inclusion, not client state or consensus
+            // state.
+            //
+            // 1. verify that the counterparty chain committed the expected_conn to its state
```
hdevalence, line 74: Are those verified somewhere else? Could the comment point to where, if so?

If there's only one step, is it better not to number it, so as not to give the impression there are also steps 2,3, .... ?

avahowell: yep, there's only one step here. i'll write this comment so it's a bit clearer

PR #787, opened 2022-05-09
[WIP] refactor staking component

hdevalence:
So far the main work is (1) expanding the set of validator states and (2) pulling the state transitions into a single method on the staking component that can exhaustively match them.

### DIFF
--- a/stake/src/component.rs
+++ b/stake/src/component.rs
@@ -42,36 +42,92 @@ pub struct Staking {
     delegation_changes: DelegationChanges,
     /// List of changes to the tendermint validator set accumulated throughout
     /// this block, to be returned during `EndBlock`.
-    tm_validator_updates: Vec<ValidatorUpdate>,
+    tm_validator_updates: BTreeMap<tendermint::PublicKey, u64>,
 }
 
 impl Staking {
     #[instrument(skip(self))]
-    fn update_tm_validator_power(&mut self, ck: &PublicKey, power: u64) -> Result<()> {
-        // TODO: would it be more sensible to have `tm_validator_updates` be a
-        // BTreeMap instead, and construct the `ValidatorUpdate` vec on-demand?
-        let existing = self
-            .tm_validator_updates
-            .iter()
-            .enumerate()
-            .find(|(_i, v)| v.pub_key == *ck);
+    async fn set_validator_state(
+        &self,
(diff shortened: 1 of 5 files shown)


stake/src/component.rs
```diff
@@ -42,36 +42,208 @@ pub struct Staking {
...
+            }
+            (Active, Disabled) => {
+                todo!("happens when a validator operator disables their validator");
+            }
+            (Inactive, Disabled) => {
+                todo!("happens when a validator operator disables their validator");
+            }
+            (Jailed, Disabled) => {
+                todo!("happens when a validator operator disables their validator");
+            }
```
hdevalence, line 178: Leaving these as `todo!`s for now shouldn't allow a chain halt, since we don't handle these state transitions, so the function won't be called with these parameters.

hdevalence:
I think this is good to merge, although we should test it on `testnet-preview` before deploying it.

stake/src/component.rs
```diff
@@ -42,36 +42,208 @@ pub struct Staking {
...
+
+                // Finally, set the validator to be inactive.
+                self.state.put_domain(state_key, Inactive).await;
+
+                Ok(())
             }
-        };
+            (Jailed, Inactive) => {
+                todo!("happens when a validator operator unjails themself");
+            }
```
hdevalence, line 166: I'm not sure how we want to handle unjailing.  One option would be to add an `Unjail` action to the transaction list. Another would be to define an updated validator definition (incremented sequence number) as an unjailing.

stake/src/component.rs
```diff
@@ -42,36 +42,208 @@ pub struct Staking {
...
+            (Disabled, Inactive) => {
+                todo!("happens when a validator operator enables their validator");
+            }
+            (Active, Disabled) => {
+                todo!("happens when a validator operator disables their validator");
+            }
+            (Inactive, Disabled) => {
+                todo!("happens when a validator operator disables their validator");
+            }
+            (Jailed, Disabled) => {
```
hdevalence, line 176: To handle these we'll probably want to add an `enabled` field to the validator definitions.

stake/src/component.rs
zbuc: The former might be more intuitive. I don't have a strong preference either.

PR #788, opened 2022-05-10
Add intialize/load methods to wallet-next storage.rs

### LINKED ISSUE #750: Add `wallet_next::Storage::initialize`
The `Storage` interface added in #613 only has individual read methods, not write methods. We don't want to have individual write methods, since we should only update the database state in two cases: 

1. when we initialize it with a new FVK;
2. when we scan a new block.

To do (1), in this issue, we should have an initialize method on `Storage`:
```
pub async fn initialize(storage_path: PathBuf, fvk: FullViewingKey, params: ChainParams) -> anyhow::Result<Self>
```
This would:

- [x] Check that the file at the given path does not exist;
- [x] Create a SQLite database there and run the migrations;
- [x] Initialize the database state with:
  - the FVK
  - the chain parameters;
  - an empty note commitment tree;
- [x] Remove the `Storage::migrate` (it's folded into `initialize`) method;
- [x] Change `Storage::new` to `Storage::load(storage_path: PathBuf)`, so that it's clear that it's loading an *existing* database state, and so that the caller doesn't have to manage opening the SQLite database / constructing a `Pool`.
  
(Not writing the sync height means it will continue to be reported as `None`, which is what we want, since we haven't synced the genesis state).

aubrika:
@zbuc - looks like we had a naming collision on branches for the same issue, which I realized only after I'd pushed my changes to remote, whoops. Doesn't look like there are any merge conflicts as it stands, but this should close #750. Let me know if you're still working on this branch, we can separate our changes or not as preferred.

### DIFF
--- a/wallet-next/src/storage.rs
+++ b/wallet-next/src/storage.rs
@@ -1,3 +1,6 @@
+use std::path::PathBuf;
+
+use anyhow::anyhow;
 use penumbra_chain::params::ChainParams;
 use penumbra_crypto::{
     merkle::{NoteCommitmentTree, Tree},
@@ -5,9 +8,11 @@ use penumbra_crypto::{
     FieldExt,
 };
 use penumbra_proto::{crypto::FullViewingKey, Message, Protobuf};
-use sqlx::{query, Pool, Sqlite};
+use sqlx::{migrate::MigrateDatabase, query, Pool, Sqlite};
 
 use crate::sync::ScanResult;
+
+const MAX_MERKLE_CHECKPOINTS_CLIENT: usize = 10;
 
 #[derive(Clone)]
 pub struct Storage {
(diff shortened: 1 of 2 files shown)


wallet-next/src/storage.rs
```diff
@@ -1,30 +1,84 @@
...
+        self: &Storage,
+        storage_path: PathBuf,
+        fvk: FullViewingKey,
+        params: ChainParams,
+    ) -> anyhow::Result<()> {
+        //   Check that the file at the given path does not exist;
+        if !storage_path.exists() {
+            // Create the SQLite database
+
+            let path_str = storage_path.as_os_str().to_str().unwrap();
```
aubrika, line 36: Is there a more succinct/less unwieldy method of converting `PathBuf` to `&str`?

hdevalence: No, the `PathBuf` type is unwieldy because it isn't required to be UTF-8 (because system file paths aren't). But if SQLite wants a `&str`, then we could avoid the whole thing by replacing `PathBuf` with `String`, and then `&storage_path` could just be passed to `create_database` directly.

wallet-next/src/storage.rs
```diff
@@ -1,30 +1,84 @@
...
+use sqlx::{migrate::MigrateDatabase, query, Pool, Sqlite};
 
 use crate::sync::ScanResult;
 
+const MAX_MERKLE_CHECKPOINTS_CLIENT: usize = 10;
+
 #[derive(Clone)]
 pub struct Storage {
-    pool: Pool<Sqlite>,
+    storage_path: PathBuf,
```
hdevalence, line 18: I think we probably want to keep the `Pool` on `Storage`, so we don't have to be constantly connecting to it.

wallet-next/src/storage.rs
```diff
@@ -1,30 +1,84 @@
...
     }
 
-    pub async fn migrate(self: &Storage) -> anyhow::Result<()> {
-        sqlx::migrate!().run(&self.pool).await.map_err(Into::into)
+    pub async fn initialize(
+        self: &Storage,
+        storage_path: PathBuf,
+        fvk: FullViewingKey,
+        params: ChainParams,
+    ) -> anyhow::Result<()> {
```
hdevalence, line 31: I think this method should be 
```suggestion
    pub async fn initialize(
        storage_path: PathBuf,
        fvk: FullViewingKey,
        params: ChainParams,
    ) -> anyhow::Result<Self> {
```
, so that the `initialize` method is also a constructor, and there are just two different constructors: one for already-existing data (`load`) and one that initializes (`initialize`).  Then there's no reason to move the `Pool` out of the `Storage` struct.

wallet-next/src/storage.rs
```diff
@@ -1,26 +1,78 @@
...
 
-    pub async fn migrate(self: &Storage) -> anyhow::Result<()> {
-        sqlx::migrate!().run(&self.pool).await.map_err(Into::into)
+    pub async fn initialize(
+        storage_path: String,
+        fvk: FullViewingKey,
+        params: ChainParams,
+    ) -> anyhow::Result<Self> {
+        //   Check that the file at the given path does not exist;
+        if !PathBuf::from(&storage_path).exists() {
```
hdevalence, line 35: The eventual error ends up fairly distant at the bottom of the function; I think it'd be cleaner and avoid rightward drift if the conditional were reversed, to do if (exists) { return Err; }, so the error can be local to the conditional and the rest of the happy path can be unindented

hdevalence (approved):

hdevalence:
Looks like the CI failure is #797

PR #791, opened 2022-05-10
stake: correct handling of validator updates

hdevalence:
Prior to this commit, setting validators with enabled: false did not trigger state updates.


hdevalence:
Rebased without the test run since the CI hung for some reason and won't be cancelled (??)

PR #793, opened 2022-05-10
Add sync_to_latest worker to wallet-next

### LINKED ISSUE #752: Implement `wallet_next::Worker::sync_to_latest`
The logic should be similar to the existing logic in `pcli`: construct a `CompactBlockRangeRequest`, then iterate over the returned blocks.  For each one, produce a `ScanResult`, then feed it to `record_block` from #751.

To do this, the worker task will need to own an in-memory NCT and an in-memory FVK, which it should load from the database on construction and save as struct members.

aubrika:
Closes #752

### DIFF
--- a/wallet-next/src/storage.rs
+++ b/wallet-next/src/storage.rs
@@ -5,10 +5,10 @@ use penumbra_chain::params::ChainParams;
 use penumbra_crypto::{
     merkle::{NoteCommitmentTree, Tree},
     note::Commitment,
-    FieldExt,
+    FieldExt, FullViewingKey,
 };
-use penumbra_proto::{crypto::FullViewingKey, Message, Protobuf};
-use sqlx::{migrate::MigrateDatabase, query, Pool, Sqlite};
+use penumbra_proto::Protobuf;
+use sqlx::{query, Pool, Sqlite};
 
 use crate::sync::ScanResult;
 
@@ -225,6 +225,13 @@ impl Storage {
 
         let nct_bytes = bincode::serialize(nct)?;
         sqlx::query!("UPDATE note_commitment_tree SET bytes = ?", nct_bytes)
+            .execute(&mut tx)
+            .await?;
+
+        // Record block height as latest synced height
+
+        let latest_sync_height = scan_result.height as i64;
+        sqlx::query!("UPDATE sync_height SET height = ?", latest_sync_height)
             .execute(&mut tx)
             .await?;
 
--- a/wallet-next/src/storage.rs
+++ b/wallet-next/src/storage.rs
@@ -137,6 +137,23 @@ impl Storage {
         scan_result: ScanResult,
         nct: &mut NoteCommitmentTree,
     ) -> anyhow::Result<()> {
+        //Check that the incoming block height follows the latest recorded height
+        let last_sync_height = self.last_sync_height().await?;
+
+        let correct_height = match last_sync_height {
+            // Require that the new block follows the last one we scanned.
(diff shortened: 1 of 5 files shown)


wallet-next/src/storage.rs
```diff
@@ -39,6 +39,14 @@ impl Storage {
         Ok(result.map(|row| row.height as u64))
     }
 
+    pub async fn set_last_sync_height(&self, height: i64) -> anyhow::Result<()> {
```
aubrika, line 42: Should this be incorporated into `record_block`, so any block recorded with height > latest will update the `sync_height' table?

hdevalence: Yeah, I think so, my instinct is that we probably don't want to have any methods that set individual parts of the storage, because we know that at the moment, we only need to (1) initialize it or (2) ratchet forward a block at a time

wallet-next/src/worker.rs
```diff
@@ -1,24 +1,71 @@
...
+
+        while let Some(block) = stream.message().await? {
+            let scan_result = scan_block(&self.fvk, &mut self.nct, block.try_into()?);
+
+            end_height = std::cmp::max(end_height, scan_result.height);
+
+            self.storage.record_block(scan_result, &mut self.nct);
+        }
+
+        self.storage.set_last_sync_height(end_height as i64);
```
hdevalence, line 61:
```suggestion
            self.storage.record_block(scan_result, &mut self.nct);
        }
```

If the `record_block` impl sets the height internally, these lines and the entire setter function could disappear.

wallet-next/src/worker.rs
```diff
@@ -1,24 +1,71 @@
...
+                end_height: 0,
+                chain_id: self.storage.chain_params().await?.chain_id,
+            }))
+            .await?
+            .into_inner();
+
+        while let Some(block) = stream.message().await? {
+            let scan_result = scan_block(&self.fvk, &mut self.nct, block.try_into()?);
+
+            end_height = std::cmp::max(end_height, scan_result.height);
```
aubrika, line 56: The intent here is to ratchet the local `end_height` up to the highest witnessed block height, but as a sanity check, is there any possibility that this could result in missed blocks of a lower height which would only come in later?

hdevalence: Yeah, but this problem is much worse -- if we miss blocks, we'll get a corrupted state, because we could miss notes that we should have seen, or else miss spends (e.g., if we would have got a note in block H, and we miss it, and it's spent in block H+K, then when we scan block H+K we won't see that it's something whose spend we should record, even if we scan block H after the fact).

There's no way around the fact that we need to apply each delta update (block) in sequence.

As a sanity check, I think it would make sense to have `record_block` check against `sync_height: Option<u64>` like so:
```rust
let correct_height = match sync_height {
    // Require that the new block follows the last one we scanned.
    Some(cur_height) => scan_result.height == cur_height + 1,
    // Require that the new block represents the initial chain state.
    None => scan_result.height == 0,
};

if !correct_height {
    return Err(anyhow::anyhow!("Wrong block height {:?} for sync height {:?}", scan_result.height, sync_height));
}
```

aubrika: Cool, this makes sense - then presumably just increment the latest height stored in sqlite by 1?

I'm mostly concerned about the role the sqlite `sync_height` table is playing here and want to take extra care to ensure it isn't incremented in such a way that we could end up with corrupted state.

Are there additional checks or mitigations we could be doing, or additional means of verifying we are storing the correct latest height in sqlite?

PR #795, opened 2022-05-11
remove wrapper types for IBC domain types

avahowell:
This PR removes our wrapper types for the IBC domain types, and instead does `impl Protobuf` directly in the proto crate on the IBC types. Also, the state paths for client related data have been brought into spec with ICS2.

### DIFF
--- a/ibc/src/component/client.rs
+++ b/ibc/src/component/client.rs
@@ -39,7 +39,7 @@ use tendermint_light_client_verifier::{
 };
 use tracing::instrument;
 
-use crate::{ClientConnections, ClientCounter, ConsensusState, VerifiedHeights, COMMITMENT_PREFIX};
+use crate::{ClientConnections, ClientCounter, VerifiedHeights, COMMITMENT_PREFIX};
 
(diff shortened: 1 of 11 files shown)


ibc/src/component/client.rs
```diff
@@ -352,44 +354,52 @@ impl Ics2Client {
...
     }
 
     async fn get_verified_heights(&self, client_id: &ClientId) -> Result<Option<VerifiedHeights>> {
         self.get_domain(
             format!(
-                "ibc/ics02-client/clients/{}/verified_heights",
-                hex::encode(client_id.as_bytes())
+                // NOTE: this is an implementation detail of the Penumbra ICS2 implementation, so
+                // it's not in the same path namespace.
+                "penumbra_verified_heights/{}/verified_heights",
```
hdevalence, line 401: non-blocking nit: maybe these should be `penumbra_ibc/verified_heights/`, so that they're all under the same path namespace, even though it's distinct from the `ibc` namespace?

hdevalence (approved):

PR #801, opened 2022-05-12
wip: ibc: add ICS4 channels support

avahowell:
this PR adds ICS4 channels support.


hdevalence (approved):
This looks reasonable to me, although I didn't cross-check against the ICS4 spec.

Could we split out the TODO items into issues, so we don't lose track of them?

ibc/src/component/channel.rs
```diff
@@ -0,0 +1,269 @@
...
+        self.put_channel_counter(ctr + 1).await;
+
+        Ok(ChannelId::new(ctr))
+    }
+    async fn get_channel(
+        &self,
+        channel_id: &ChannelId,
+        port_id: &PortId,
+    ) -> Result<Option<ChannelEnd>> {
+        self.get_domain(format!("channelEnds/ports/{}/channels/{}", port_id, channel_id).into())
```
hdevalence, line 224: Kind of surprised that these paths aren't prefixed with `ibc/` or something.

ibc/src/component/channel.rs
```diff
@@ -0,0 +1,269 @@
...
+use penumbra_transaction::Transaction;
+use tendermint::abci;
+use tracing::instrument;
+
+mod stateful;
+mod stateless;
+
+mod execution;
+
+pub struct ICS4Channel {
```
hdevalence, line 36: nit: could we rename this to `Ics4Channel` for consistency with the other IBC components?

ibc/src/component/channel.rs
```diff
@@ -0,0 +1,269 @@
...
+                _ => {}
+            }
+        }
+    }
+
+    #[instrument(name = "ics4_channel", skip(self, _end_block))]
+    async fn end_block(&mut self, _end_block: &abci::request::EndBlock) {}
+}
+
+impl ICS4Channel {}
```
hdevalence, line 200: nit: this empty block could be removed ?

hdevalence:
Going to squash+merge this, so that we don't risk generating conflicts, but it would be good to split out the TODOs into issues so we don't lose track.

PR #807, opened 2022-05-12
Track the TCT alongside the NCT

plaidfinch:
This commits the TCT root alongside the NCT root to the JMT state, and stores the TCT in the sidecar non-consensus storage in RocksDB. It prints the TCT root every block to `debug!` level tracing.


shielded-pool/src/component.rs
```diff
@@ -246,6 +248,29 @@ impl Component for ShieldedPool {
             .unwrap();
         }
 
+        // Close the block in the TCT
+        self.tiered_commitment_tree
+            .end_block()
+            .expect("ending TCT block should never fail");
```
hdevalence, line 254: If we `expect` here, we'll halt `pd` in case of an invariant violation. We'll want that eventually, but initially, we want the TCT to be a sidecar that gets exercised but not fed into consensus.

Instead, I we should match on the result and do a `tracing::error!` on failure, but otherwise continue executing.

shielded-pool/src/component.rs
```diff
@@ -246,6 +248,29 @@ impl Component for ShieldedPool {
...
+                .get_chain_params()
+                .await
+                .expect("chain params request must succeed")
+                .epoch_duration,
+        )
+        .is_epoch_end(self.compact_block.height)
+        {
+            self.tiered_commitment_tree
+                .end_epoch()
+                .expect("ending TCT block should never fail");
```
hdevalence, line 269: ditto expect above

shielded-pool/src/component.rs
```diff
@@ -344,6 +369,9 @@ impl ShieldedPool {
         // 1. Insert it into the NCT
         self.note_commitment_tree
             .append(&output_body.note_commitment);
+        self.tiered_commitment_tree
+            .insert(penumbra_tct::Witness::Forget, output_body.note_commitment)
+            .expect("inserting a commitment into the TCT should never fail");
```
hdevalence, line 374: ditto expect above

shielded-pool/src/component.rs
```diff
@@ -360,7 +388,11 @@ impl ShieldedPool {
             .await;
         // and the note commitment tree data and anchor:
         self.state
-            .set_nct_anchor(self.compact_block.height, self.note_commitment_tree.root2())
+            .set_nct_anchor(
+                self.compact_block.height,
+                self.note_commitment_tree.root2(),
+                self.tiered_commitment_tree.root(),
```
hdevalence, line 394: This is writing the NCT anchor from the TCT sidecar into the consensus-critical state, which is what we don't want to do yet, right?

shielded-pool/src/component.rs
```diff
@@ -481,18 +513,38 @@ pub trait View: StateExt {
...
+        // Write the TCT anchor both as a value, so we can look it up,
+        self.put_domain(
+            format!("shielded_pool/tct_anchor/{}", height).into(),
+            tct_anchor,
         )
         .await;
         // and as a key, so we can query for it.
         self.put_proto(
-            format!("shielded_pool/valid_anchors/{}", anchor).into(),
+            format!("shielded_pool/valid_tct_anchors/{}", tct_anchor).into(),
```
hdevalence, line 547: I think we want to roll back all the changes to this method, and leave it as-is, then, when we execute the switch-over, we'll swap
```
nct_anchor: merkle::Root
```
for
```
nct_anchor: penumbra_tct::Root
```
along with the other changes we'll need to make to the client protocol.

hdevalence (changes requested):
This looks good, but goes slightly too far: we should roll back the changes that add the TCT into the consensus-critical state, and only keep the storage sidecar.

shielded-pool/src/component.rs
plaidfinch: Ah, yeah, I figured it was fine to prepone doing that, but taking it out for now.

shielded-pool/src/component.rs
plaidfinch: Makes sense!

plaidfinch:
This is now rebased relative to the latest main branch and addresses the changes above. Look good to you @hdevalence?

hdevalence (approved):
LGTM!

PR #808, opened 2022-05-12
storage: create database and column families if they are missing

avahowell:
when we changed how we use the RocksDB API, from `open_default` to `open_cf`, we removed the create_if_missing and create_missing_column_families flag, which exist by default in `open_default`. this was causing our smoke test and the IBC test to fail.


hdevalence (approved):
LGTM

PR #809, opened 2022-05-13
storage bugfix: use `jmt` column family in rocksdb

avahowell:
When we updated storage to use rocks column families, we didn't update the JMT reader and writers to use the JMT column family. This changes the JMT column family to 'jmt' and updates the readers/writers to use it


hdevalence (approved):
LGTM, let's check the status endpoint on `testnet-preview` before retagging

PR #814, opened 2022-05-13
update guide to use Penumbra tendermint config instead of default

avahowell:


docs/guide/src/pd/join-testnet/fullnode.md
```diff
@@ -28,6 +28,12 @@ curl -s http://testnet.penumbra.zone:26657/genesis | jq ".result.genesis" > \$HO
 ```
 The response data is nested, so we use `jq` to grab a specific key.
 
+Next, grab the correct tendermint configuration from our repo. This is important to make sure your node can stay in sync with the network:
```
hdevalence, line 31: This is a good idea but the config template has other slots that need to be filled in, so it's not usable off the shelf.

Instead of trying to enumerate them, maybe we should tell people to run

```
grep "{}" \$HOME/.tendermint/config/config.toml
```

to find the fields they need to fill in?

avahowell: shouldn't all of the changes that need be made be in this doc already? just setting the bootstrap peers AFAIK

hdevalence: Nope, there's also the `moniker` field, but the point of giving the command is so that, if we change the config template later (say we need to make changes for generate-testnet), we don't leave the instructions out of sync

hdevalence (approved):

docs/guide/src/pd/join-testnet/fullnode.md
avahowell: gotcha, added the grep. maybe in the future we can just have people use docker deployments that automatically pull in our tendermint config

hdevalence: Yeah, I think that'd be the best thing

PR #816, opened 2022-05-14
Add storage::notes function

### LINKED ISSUE #754: Implement `WalletService::notes`
This request will generate one of several queries, based on which fields are present in the request.  To start, we can buffer the entire query results in-memory, rather than trying to stream them to the client.

To reconstruct the `NoteRecord`s, we'll need to do the opposite steps as were used to save them, in #751.

aubrika:
Closes #754

### DIFF
--- a/wallet-next/src/service.rs
+++ b/wallet-next/src/service.rs
@@ -3,7 +3,12 @@ use std::{
     sync::{Arc, Mutex},
 };
 
-use penumbra_crypto::keys::FullViewingKeyHash;
+use async_stream::try_stream;
+use futures::stream::{StreamExt, TryStreamExt};
+use penumbra_crypto::{
+    asset,
+    keys::{DiversifierIndex, FullViewingKeyHash},
+};
 use penumbra_proto::{
     client::oblivious::oblivious_query_client::ObliviousQueryClient,
     crypto as pbc,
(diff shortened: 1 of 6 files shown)


wallet-next/src/service.rs
```diff
@@ -114,7 +119,34 @@ impl WalletProtocol for WalletService {
...
+        });
+        let amount_to_spend = request.get_ref().amount_to_spend;
+
+        let notes = self
+            .storage
+            .notes(include_spent, asset_id, diversifier_index, amount_to_spend)
+            .await
+            .map_err(|_| tonic::Status::unavailable("database error"))?;
+
+        let stream = try_stream! {
```
aubrika, line 139: Should the `storage` method be refactored to allow iteration over a range of heights, which would make returning a stream here make more sense? Or should we change the method signature in the service to return a `Vec<NoteRecord>`? Leaving as-is for now, but seems like one side or the other should change.

wallet-next/src/storage.rs
```diff
@@ -132,6 +134,86 @@ impl Storage {
...
+            None => "diversifier_index".to_string(),
+        };
+
+        // If set, stop returning notes once the total exceeds this amount.
+        //
+        // Ignored if `asset_id` is unset or if `include_spent` is set.
+        // uint64 amount_to_spend = 5;
+
+        if asset_id.is_none() || include_spent {
+            //TODO: figure out a clever way to only return notes up to the sum using SQL
```
aubrika, line 173: I don't know a succinct/straightforward way in SQL to truncate the size of the returned set by the sum of a column value, but if we refactored to stream smaller sets of `NoteRecord`s we could just sum those smaller sets and check against the `amount_to_spend` - as it stands, it's just returning the whole set of `NoteRecord`s at once, so to implement this cutoff, we'd need that to change.

hdevalence: For now I think it's fine to load everything in memory and then throw it away. We can always do the better DB query if/when we notice a perf problem.

wallet-next/src/service.rs
hdevalence: Seems like if we leave it as is, we have the option of changing just the wallet impl to steam data at a later point without having to change anything on the client side

wallet-next/src/storage.rs
```diff
@@ -132,6 +134,86 @@ impl Storage {
...
+            };
+            let note_blinding = Fq::from_bytes(record.blinding_factor[..].try_into()?)?;
+
+            output.push(NoteRecord {
+                note_commitment: Commitment::try_from(&record.note_commitment[..])?,
+                note: Note::from_parts(diversifier, transmission_key, value, note_blinding)?,
+                diversifier_index: DiversifierIndex(record.diversifier_index[..].try_into()?),
+                nullifier: Nullifier::try_from(record.nullifier)?,
+                height_created: record.height_created as u64,
+                height_spent: if record.height_spent == None {
```
hdevalence, line 206: maybe this could use Option::map to avoid the if?

hdevalence (approved):
Happy to merge on green CI

PR #817, opened 2022-05-16
Improve transaction modeling and authorization hashing.

hdevalence:
This is a prelude to being able to do a useful custody / signing protocol.

As mentioned in #810:

> In order to have a useful signing/custody protocol, we need to classify transaction data
into authorizing data (which describes the actions that signatures authorize)
and effecting data (which effects those actions, but is not necessary to
authorize, e.g., proof data). Then we'll want to have a protocol that allows
us to bundle up only the authorizing data and send it to a signer, who signs an
authorization hash (a renamed sighash) of the authorization data. However,
this means that we want to pull apart the proto structures into chunks, where
each chunk has only authorizing /or/ effecting data. To do /that/, it'd be
useful to be able to describe an OutputBody, which the NotePayload isn't
really.

This PR makes two substantive changes:

First, it changes the definition of the transaction structures and protos to
split out "authorizing data", data which describes what actions the transaction
does, which should be fed into an authorization hash for signing, and
"effecting data", which effects the actions described by the authorizing data,
but does not itself describe them, and which the authorization does not need to
be bound to.

This distinction is now consistently mapped to the `Foo`/`FooBody` distinction:
`FooBody` contains the authorizing data for `Foo`, and `Foo` contains both a
`FooBody` as well as any effecting data.

In particular, we move the anchor and the ZK proofs out of the `Body`
structures and consider the effecting data.  This allows a custodian to
authorize a transaction without restricting when the prover can create the
transaction proof.

Otherwise, if the anchor were included in the authorization hash, the prover
would be forced to use whatever anchor was available at the point that they
requested signing, and this creates a side channel that allows a third party to
learn information about what kind of custody was used for transaction
authorization.  For instance, threshold signing with offline keys necessarily
has high latency, which would force a prover to use an anchor that would be
hours or days out of date, distinguishing it from other transactions that used
recent anchors.

Second, it removes the old sighash code, and replaces with a new `AuthHash`
implementation. This serves the same purpose, but:

* has a better name: it's the hash of the authorizing data (as opposed to
  effecting data);

* is specified explicitly rather than implicitly, so that we don't require a
  signer to have the exact same protobuf implementation we use.
  
These changes lay the groundwork for a custody protocol, as mentioned in #587.


hdevalence:
I think this design is basically good, and requested review from everyone primarily to flag the design changes mentioned in the PR description rather than to request careful, line-by-line auditing, but if I missed something feel free to flag it.

redshiftzero (approved):

PR #821, opened 2022-05-17
Bubble up errors in wallet service/worker, shut down worker when service shuts down

### LINKED ISSUE #758: Implement shutdowns and error handling for `WalletService`
The scaffolding in #749 has some comments pointing to error handling, but doesn't actually do it.

There are two basic cases to cover:

- [x] If the worker errors, we should report errors through the `WalletService` (we can try to recover from them later).   To do this, we can have an `Arc<Mutex<Option<anyhow::Error>>>` shared between the `Worker` and all of the `WalletService`s.  This can be a `std::sync::Mutex` rather than a Tokio mutex, since it should be uncontended.  On the worker side, the `run` method should be wrapped in another method that checks for an error returned by the worker task, and sets it into the error slot.  On the wallet service side, the `check_worker` method should do `if self.error_slot.lock().is_some() { ... }` (which should be fast), and only if it is set, pull the error contents out as a string and put them into a `tonic::Status`.
- [x] If all of the `WalletService`s are dropped, the worker should exit, rather than continuing to run.  One way to do this is to use a [`mpsc` channel](https://docs.rs/tokio/latest/tokio/sync/mpsc/index.html); the worker has a `mpsc::Receiver<()>` and the services each have a `mpsc::Sender<()>`.  When all `Sender`s are dropped, the channel closes, and the worker can detect this and exit.

This isn't absolutely essential for a `pwalletd` PoC but is important when integrating a `WalletService` into other applications.

zbuc:
Closes #758, this will cause any errors occurring in the wallet workers to be bubbled up to the wallet service, as well as adds an `mpsc` channel between the worker and the wallet services, so that the worker can shut down when the services halt.

### DIFF
--- a/wallet-next/src/service.rs
+++ b/wallet-next/src/service.rs
@@ -1,4 +1,7 @@
-use std::pin::Pin;
+use std::{
+    pin::Pin,
+    sync::{Arc, Mutex},
+};
 
 use penumbra_crypto::keys::FullViewingKeyHash;
 use penumbra_proto::{
@@ -23,7 +26,7 @@ pub struct WalletService {
     storage: Storage,
     // TODO: add a way for the WalletService to poll the state of its worker task, to determine if it failed
     // this probably looks like a handle for the task, wrapped in an arc'd mutex or something
-    // error_slot: Arc<Mutex<Option<anyhow::Error>>>,
+    error_slot: Arc<Mutex<Option<anyhow::Error>>>,
     // TODO: add a way for the WalletService to signal the worker that it should shut down
     // this probably looks like an Arc<oneshot::Sender<()>> or something,
     // where the receiver is held by the worker and the worker checks if it's closed (=> all sender handles dropped)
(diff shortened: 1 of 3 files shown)


wallet-next/src/service.rs
```diff
@@ -74,7 +88,20 @@ impl WalletService {
...
+            return Err(tonic::Status::new(
+                tonic::Code::Internal,
+                format!(
+                    "Worker failed: {}",
+                    self.error_slot.lock().unwrap().as_ref().unwrap()
+                ),
+            ));
+        }
+
         // TODO: check whether the worker is still alive, else fail, when we have a way to do that
```
hdevalence, line 103: I think this TODO is closed out by these changes(?)

wallet-next/src/worker.rs
```diff
@@ -1,21 +1,29 @@
...
+use tokio::sync::mpsc::{error::TryRecvError, Receiver};
 use tonic::transport::Channel;
 #[derive(Clone)]
 pub struct Worker {
     storage: Storage,
     client: ObliviousQueryClient<Channel>,
     nct: NoteCommitmentTree,
     fvk: FullViewingKey, // TODO: notifications (see TODOs on WalletService)
+    error_slot: Arc<Mutex<Option<anyhow::Error>>>,
+    // This is in an `Arc` because `Receiver` is not `Clone`.
```
hdevalence, line 17: Why does Receiver need to be Clone? We only have one worker, right?

wallet-next/src/worker.rs
```diff
@@ -64,12 +74,39 @@ impl Worker {
...
+                    tracing::info!(?e, "wallet worker error");
+                    self.error_slot.lock().unwrap().replace(e);
+                }
+            };
+        }
+
+        Ok(())
+    }
+
+    async fn _run(&mut self) -> Result<(), anyhow::Error> {
```
hdevalence, line 94: nit: I think it would be more consistent with other Tokio code to call this run_inner

wallet-next/src/worker.rs
zbuc: The Worker type had an existing Clone annotation.

hdevalence: Ah, cool. I don't think we need it -- we consume the worker to run it, so it'd be better to drop the Clone bound and hold the receiver directly.

wallet-next/src/service.rs
zbuc: I think there's one case that's not covered: if the worker is to crash without setting the `error_slot`, I don't think it would cause the service to die.

PR #824, opened 2022-05-18
implement IBC packets, acknowledgements, timeouts

avahowell:


ibc/src/component/channel/stateful/proof_verification.rs
```diff
@@ -0,0 +1,306 @@
...
+use ibc::core::ics24_host::Path;
+use ibc::downcast;
+use ibc_proto::ibc::core::commitment::v1::MerkleProof as RawMerkleProof;
+use prost::Message;
+
+use ibc::core::ics02_client::client_state::AnyClientState;
+use penumbra_chain::View as _;
+use sha2::{Digest, Sha256};
+
+// NOTE: this is underspecified.
```
hdevalence, line 24: Should we keep a list of things that are underspecified, so we can contribute them to spec improvements?

PR #827, opened 2022-05-18
Fix bug where inserting a into finalized blocks/epochs failed incorrectly

plaidfinch:
When an epoch was finalized, subsequent insertions into the tree would fail, because they incorrectly attempted to insert into the finalized epoch, rather than creating a new, unfinalized epoch into which to insert. This fixes that bug and shows no errors when testing with `pd` across several epochs from genesis.


hdevalence (approved):
This looks reasonable and definitely makes the code clearer!

PR #831, opened 2022-05-18
Use `get_block_height` as canonical source of block height

plaidfinch:
Previously there were a couple subtle order-of-execution bugs in the shielded pool because we sourced height from the under-construction `CompactBlock`. This PR uniformly uses `get_block_height` throughout the shielded pool to get the height, and it sets it in the compact block immediately upon begin-block. Log messages indicate that the height now correctly reflects the increasing sequence of blocks.

### DIFF
--- a/shielded-pool/src/component.rs
+++ b/shielded-pool/src/component.rs
@@ -222,8 +222,8 @@ impl Component for ShieldedPool {
         //}
     }
 
-    #[instrument(name = "shielded_pool", skip(self, end_block))]
-    async fn end_block(&mut self, end_block: &abci::request::EndBlock) {
+    #[instrument(name = "shielded_pool", skip(self, _end_block))]
+    async fn end_block(&mut self, _end_block: &abci::request::EndBlock) {
         // Get the current block height
         let height = self
             .state
--- a/shielded-pool/src/component.rs
+++ b/shielded-pool/src/component.rs
@@ -88,12 +88,23 @@ impl Component for ShieldedPool {
             .unwrap();
         }
 
-        self.compact_block.height = 0;
+        self.compact_block.height = self
+            .state
+            .get_block_height()
+            .await
+            .expect("block height must be set");
+
         self.write_compactblock_and_nct().await.unwrap();
     }
 
     #[instrument(name = "shielded_pool", skip(self, _begin_block))]
-    async fn begin_block(&mut self, _begin_block: &abci::request::BeginBlock) {}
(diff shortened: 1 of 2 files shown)


shielded-pool/src/component.rs
```diff
@@ -88,12 +88,23 @@ impl Component for ShieldedPool {
             .unwrap();
         }
 
-        self.compact_block.height = 0;
+        self.compact_block.height = self
+            .state
+            .get_block_height()
+            .await
+            .expect("block height must be set");
+
```
hdevalence, line 96: In `init` we should be hardcoding the block height as 0; I'm not sure that the block height will be set at this point, because there hasn't been a `BeginBlock`.  The 0th CompactBlock is special-cased out of the genesis app state in `InitChain`, so it doesn't actually correspond to a real Tendermint block, it's just a way to hook the genesis state up with the client sync code.

shielded-pool/src/component.rs
```diff
@@ -216,25 +227,34 @@ impl Component for ShieldedPool {
...
+        if height != (end_block.height as u64) {
+            tracing::error!(
+                "end block height {} does not match current height {}",
+                end_block.height,
+                height
+            );
+        }
```
hdevalence, line 243: I'm not sure there's much point of doing this -- at this point we're going to be badly corrupted, so adding an extra tracing message doesn't help much. Better to leave this kind of invariant checking out of the individual components IMO.

shielded-pool/src/component.rs
```diff
@@ -391,13 +411,15 @@ impl ShieldedPool {
 
     #[instrument(skip(self))]
     async fn write_compactblock_and_nct(&mut self) -> Result<()> {
+        // Extract the compact block, resetting it
+        let compact_block = std::mem::take(&mut self.compact_block);
+        let height = compact_block.height;
```
hdevalence, line 416: We should use the `get_block_height` method here, since that's where the actual bug was -- this is still dependent on someone not refactoring this to reorder changes to the compact block with grabbing the height, or grabbing the height off of `self.compact_block` instead.

plaidfinch: Right! Done.

shielded-pool/src/component.rs
plaidfinch: Makes sense to me.

shielded-pool/src/component.rs
plaidfinch: Ah, I see, that makes sense.

Issue #806, opened 2022-05-12
Add typed state query methods to `pcli q`


hdevalence:
**Is your feature request related to a problem? Please describe.**

Following #805 and #802, we have a minimum viable state query interface.  It would also be useful to be able to make typed queries to the state

**Describe the solution you'd like**

Add query commands for each component:

- [ ] `pcli q shielded-pool`
- [ ] `pcli q stake`
- [ ] `pcli q ibc`

Then, for each component's query command, implement one subcommand that looks up one of the items in `state_keys`, e.g.,
```
pcli q shielded-pool token-supply ASSET_ID
```
which parses an input `ASSET_ID`, uses `penumbra_shielded_pool::state_keys::token_supply(asset_id)` to compute the key hash, does the RPC, and then does proto decoding from the response bytes into the appropriate domain type (in this case `u64`).

It's only necessary to implement one typed query for each component, in order to define a structure that we can add additional query commands to later. It's desirable to only do one to start, because while it's useful to be able to make typed queries, each typed query command is more code to maintain, and we're still iterating on the design of the state. So filling in additional queries as needed seems like a better strategy.

PR #835, opened 2022-05-19
Add `chain params` and `chain_info` `pcli` commands

### LINKED ISSUE #759: Add `chain` subcommand to `pcli` to view chain parameters and stats
There's no easy way to view the chain parameters or info (e.g. chain id/height/epoch) on-demand.

A `pcli chain` subcommand could be added.

Subcommands:

* [ ] `pcli chain params`
* [ ] `pcli chain info` (with `--verbose` and it would print params as well?)

zbuc:
This implements new `pcli chain params` and `pcli chain info` commands.

`pcli chain params` prints the current chain parameters.

`pcli chain info` prints some summary information about the chain. Maybe this would be better off implemented in the general state query interface later on.

Closes #759

### DIFF
--- a/pcli/src/command.rs
+++ b/pcli/src/command.rs
@@ -2,6 +2,7 @@ use structopt::StructOpt;
 
 mod addr;
 mod balance;
+mod chain;
 mod stake;
 mod temp;
 mod tx;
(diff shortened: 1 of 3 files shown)


hdevalence:
Wondering if there's a way we could align this with a general purpose query interface?

zbuc:
> Wondering if there's a way we could align this with a general purpose query interface like in #806?

Yeah, I was thinking a lot of this stuff would be nice to have behind some kind of RPC backed by the same metrics we send to Prometheus.

PR #840, opened 2022-05-19
Make TCT the only note commitment tree across the whole workspace

plaidfinch:


crypto/src/note.rs
```diff
@@ -236,6 +236,27 @@ impl Note {
...
+            value.amount.into(),
+            value.asset_id.0,
+            diversified_generator.compress_to_field(),
+            transmission_key_s,
+        ),
+    );
+
+    Commitment(commit)
+}
+
```
hdevalence, line 259: Should this be a method on `Note`?

hdevalence: i.e., do we have a reason to construct a note commitment from raw parts without constructing a note?

hdevalence:
This looks basically good to me! I had one (non-blocking) question about the `commitment` function, and it'd be good to do a pass to remove the warnings, but otherwise this looks great, awesome work!

crypto/src/note.rs
plaidfinch: We do construct a commitment this way without starting from a note, in two places, where we verify spend and output proofs, respectively:

https://github.com/penumbra-zone/penumbra/blob/e45943eb45ba4f22575b18a3b0d6726a276d3723/crypto/src/proofs/transparent.rs#L82-L87

https://github.com/penumbra-zone/penumbra/blob/e45943eb45ba4f22575b18a3b0d6726a276d3723/crypto/src/proofs/transparent.rs#L172-L177

Is there a better way to do this? I also agree it's a little awkward, but I didn't feel like a static method on `Note` would make more sense than this.

avahowell:
This LGTM as well. The only relevant changes to the ibc piece was swapping out the incrementalmerkletree for the TCT in the client update integration test

plaidfinch:
The only iffy thing to me right now is the way the `wallet-next` integration is: @aubrika can you let me know what you think? There is a `todo!` I've placed for where we'll need to fill in the epoch duration for the chain; I'm not sure (because I don't know how the wallet daemon is designed) how best to source this:

https://github.com/penumbra-zone/penumbra/blob/8a884db3b7065936e6078cb214e2902a5b7cc377/wallet-next/src/sync.rs#L76-L82

Would it be acceptable to merge this as-is and fix up that part of the wallet in a separate PR, given that it's not yet primary?

aubrika:
> The only iffy thing to me right now is the way the `wallet-next` integration is: @aubrika can you let me know what you think? There is a `todo!` I've placed for where we'll need to fill in the epoch duration for the chain; I'm not sure (because I don't know how the wallet daemon is designed) how best to source this:
> 
> https://github.com/penumbra-zone/penumbra/blob/8a884db3b7065936e6078cb214e2902a5b7cc377/wallet-next/src/sync.rs#L76-L82
> 
> Would it be acceptable to merge this as-is and fix up that part of the wallet in a separate PR, given that it's not yet primary?

Currently `wallet-next` is getting `ChainParams` through an ObliviousClient instance, so that is its source for the `epoch_duration` along with all other chain params. It is currently also passing no chain_id in particular to this request, so I think it will always be getting the default ChainParams, for now.

crypto/src/note.rs
hdevalence: The alternative would be to construct a `Note` out of the parts, and then call `note.commit` on the resulting instance.

PR #850, opened 2022-05-20
Conditionally enable console-subscriber with a console-subscriber feature

zbuc:
`console-subscriber` seems to cause OOM errors when `pd` runs long enough, and at any rate, we don't want it enabled on our production deployments.

This PR adds a default feature to enable `console-subscriber` so developers building locally will keep behaving as expected. The prod Dockerfiles have been updated to disable default features on the `pd` crate, and will use the regular tracing subscriber instead.

### DIFF
--- a/docker-compose.prod.yml
+++ b/docker-compose.prod.yml
@@ -8,7 +8,7 @@ services:
       [
         "bash",
         "-c",
-        "pd start --host 0.0.0.0 -r /pd/rocksdb"
+        "pd start --host 0.0.0.0 -r /pd/rocksdb --no-console"
       ]
     logging:
       driver: "json-file"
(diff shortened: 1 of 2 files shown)


hdevalence:
I'd rather not have an extra conditional compilation here. Can we just make a --enable-console flag instead? I think it's fine if it's off by default. But that way we don't have to worry about multiple build configurations.

hawkw:
Memory leak should be fixed in the next `tokio-console` release; I'm hoping to get one out today.

hawkw:
Should be fixed in https://github.com/tokio-rs/console/releases/tag/console-subscriber-v0.1.6

zbuc:
Closing and using latest `console-subscriber` for now. @hdevalence mentioned future improvements like adding "layering" of subscribers for pd rather than conditional enabling.

PR #855, opened 2022-05-23
Add additional PD metrics for staking and TCT

zbuc:
This introduces some simple metrics on validator state and TCT size.

One thing that wasn't clear was how we should handle the missed blocks metric. Missed blocks are only relevant for Active validators, so the metric is only updated for Active validators -- but if a validator transitions out of Active state, the metric won't be updated. Maybe it should be zeroed out if the validator transitions out of Active?

However that would mismatch the state machine code, which never resets the missed blocks counter, even if the validator goes disabled and then active again. I'm thinking we can try the metric as implemented and tweak if needed.

### DIFF
--- a/stake/src/component.rs
+++ b/stake/src/component.rs
@@ -93,6 +93,22 @@ impl Staking {
     ) -> Result<()> {
         let state_key = format!("staking/validators/{}/state", identity_key).into();
 
+        // Update metrics
+        match cur_state {
+            Inactive => decrement_gauge!(metrics::INACTIVE_VALIDATORS, 1.0),
+            Active => decrement_gauge!(metrics::ACTIVE_VALIDATORS, 1.0),
+            Disabled => decrement_gauge!(metrics::DISABLED_VALIDATORS, 1.0),
+            Jailed => decrement_gauge!(metrics::JAILED_VALIDATORS, 1.0),
+            Tombstoned => decrement_gauge!(metrics::TOMBSTONED_VALIDATORS, 1.0),
+        };
+        match new_state {
+            Inactive => increment_gauge!(metrics::INACTIVE_VALIDATORS, 1.0),
+            Active => increment_gauge!(metrics::ACTIVE_VALIDATORS, 1.0),
+            Disabled => increment_gauge!(metrics::DISABLED_VALIDATORS, 1.0),
+            Jailed => increment_gauge!(metrics::JAILED_VALIDATORS, 1.0),
+            Tombstoned => increment_gauge!(metrics::TOMBSTONED_VALIDATORS, 1.0),
+        };
+
         // Doing a single tuple match, rather than matching on substates,
         // ensures we exhaustively cover all possible state transitions.
         use validator::BondingState::*;
@@ -149,8 +165,6 @@ impl Staking {
                 self.state.put_domain(state_key, Active).await;
 
                 // Update metrics
-                decrement_gauge!(metrics::INACTIVE_VALIDATORS, 1.0);
-                increment_gauge!(metrics::ACTIVE_VALIDATORS, 1.0);
                 gauge!(metrics::MISSED_BLOCKS, 0.0, "identity_key" => identity_key.to_string());
 
                 tracing::debug!(?power, "validator became active");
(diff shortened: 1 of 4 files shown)


elsehow:
I think this is a reasonable thing to try for now. Intuitively, most people would probably expect that an inactive validator is "missing" all the blocks that happen while it is inactive. But, that *also* causes a mismatch between the metric we record on and the state machine's state, which is not ideal.

zbuc:
Just realized this is missing incrementing the counters for genesis validators and newly added validators, will fix.

hdevalence:
> However that would mismatch the state machine code, which never resets the missed blocks counter, even if the validator goes disabled and then active again. I'm thinking we can try the metric as implemented and tweak if needed.

Hmm, are you sure? I think the counter is reset whenever the validator becomes active:
```rust
(Inactive, Active) => {
    // The validator's delegation pool becomes bonded.
    self.state
        .set_validator_bonding_state(identity_key, Bonded)
        .await;

    // Start tracking the validator's uptime with a new uptime tracker.
    // This overwrites any existing uptime tracking, regardless of whether
    // the validator was recently in the active set.
    self.state
        .set_validator_uptime(
            identity_key,
            Uptime::new(
                self.state.get_block_height().await?,
                self.state.signed_blocks_window_len().await? as usize,
            ),
        )
        .await;
```
The call to `Uptime::new()` initializes a blank slate bitflag.

We could also wipe the bitflag when the validator becomes *inactive*, in order to get this behavior?

stake/src/metrics.rs
```diff
@@ -15,15 +15,23 @@ pub use metrics::*;
...
 }
 
-// Sample code for reference -- delete when adding the first metric
-// pub const MEMPOOL_CHECKTX_TOTAL: &str = "penumbra_pd_mempool_checktx_total";
+pub const MISSED_BLOCKS: &str = "penumbra_pd_missed_blocks";
+pub const ACTIVE_VALIDATORS: &str = "penumbra_pd_active_validators";
+pub const INACTIVE_VALIDATORS: &str = "penumbra_pd_inactive_validators";
+pub const JAILED_VALIDATORS: &str = "penumbra_pd_jailed_validators";
+pub const DISABLED_VALIDATORS: &str = "penumbra_pd_disabled_validators";
+pub const TOMBSTONED_VALIDATORS: &str = "penumbra_pd_tombstoned_validators";
```
hdevalence, line 37: Some nitpicking on the metrics keys:

* We're in the `stake` crate rather than the `pd` crate, so I think we should use `penumbra_stake` rather than `penumbra_pd`.
* Since metrics are sorted alphabetically, I think it'd be better to have `validators` come first, to group the metrics.

This way we'd end up with
```
penumbra_stake_validators_active
penumbra_stake_validators_inactive
penumbra_stake_validators_jailed
penumbra_stake_validators_disabled
penumbra_stake_validators_tombstoned
```

Also, are we supposed to have a unit suffix, per the [naming guidelines](https://prometheus.io/docs/practices/naming/)? Or maybe not, since it's a unitless *gauge*, rather than a unitless *count*?

storage/src/metrics.rs
```diff
@@ -15,15 +15,8 @@ pub use metrics::*;
...
-        "The total number of checktx requests made to the mempool"
-    );
-     */
+    register_gauge!(TCT_SIZE_BYTES);
+    describe_gauge!(TCT_SIZE_BYTES, "The size of the serialized TCT in bytes");
 }
 
-// Sample code for reference -- delete when adding the first metric
-// pub const MEMPOOL_CHECKTX_TOTAL: &str = "penumbra_pd_mempool_checktx_total";
+pub const TCT_SIZE_BYTES: &str = "penumbra_pd_tct_size_bytes";
```
hdevalence, line 22: ditto above, how about
```
penumbra_storage_tct_size_bytes
```

storage/src/storage.rs
```diff
@@ -6,13 +6,14 @@ use jmt::{
...
 };
+use metrics::gauge;
 use rocksdb::{Options, DB};
 use tokio::sync::RwLock;
 use tracing::{instrument, Span};
 
 use penumbra_tct as tct;
 
-use crate::State;
+use crate::{metrics::TCT_SIZE_BYTES, State};
```
hdevalence, line 16: Rather than importing everything from two places, if we just do `use crate::metrics;`, we can write code like
```rust
metrics::gauge!(metrics::TCT_SIZE_BYTES, /* ... */);
```
and then it's clear that all of the constants are metrics-related, rather than being part of some other code.

stake/src/component.rs
```diff
@@ -580,6 +635,8 @@ impl Staking {
                     ?params.missed_blocks_maximum,
                     "recorded vote info"
                 );
+                // TODO: We may want to update or clear this metric for non-active validators.
```
hdevalence, line 638: I think setting this to `0` as the validator moves out of `Active` state is a good idea.

stake/src/metrics.rs
zbuc: Wouldn't `validators` be the correct plural unit suffix?

hdevalence: I guess, but I think it's probably better to put validators first, so the labels are grouped alphabetically, and not have a unit

stake/src/component.rs
```diff
@@ -172,6 +172,21 @@ impl Staking {
...
+                        .set_validator_uptime(
+                            identity_key,
+                            Uptime::new(
+                                self.state.get_block_height().await?,
+                                self.state.signed_blocks_window_len().await? as usize,
+                            ),
+                        )
+                        .await;
+                }
+
```
hdevalence, line 189: Do we need to write this in, specially, only when the validator becomes `Inactive`, rather than just updating the gauge? We'll clear the uptime tracker if it ever becomes active, and we'll never read it while it's non-Active.

zbuc: It doesn't seem strictly necessary but I thought it was what you were suggesting in your comment:

> We could also wipe the bitflag when the validator becomes inactive, in order to get this behavior?

hdevalence (approved):

PR #856, opened 2022-05-23
Correct flag in testnet instructions

elsehow:
This PR fixes an incorrect flag for validator nodes listed in the docs.


zbuc:
Thanks!

PR #857, opened 2022-05-23
Suggest removing RocksDB on Tendermint error

elsehow:
People may need to reset RocksDB when upgrading between testnets. This PR makes a suggestion to that effect.


avahowell:
hmm should we add this as a primary thing to do in the steps, instead of a common error?

PR #862, opened 2022-05-24
doc: testnet RM checklist

redshiftzero:


.github/ISSUE_TEMPLATE/release_testnet.md
```diff
@@ -0,0 +1,36 @@
...
+
+Monday (week before release):
+- [ ] Create GitHub milestone, work with team to populate the milestone with tickets targeted for the release.
+
+Thursday:
+- [ ] Check in with team again in a release meeting and update the GitHub milestone to ensure it represents what will make it into the testnet.
+
+Friday: Prepare a PR (for merge on Monday) with the following:
+- [ ] Update `discord-addresses` file using the Galileo bot (`cargo run --release history --channel <channel>`)
+- [ ] Create a new `testnet/<moon>/allocations.csv` file describing the initial genesis allocations for this testnet (make sure all current team members and the Galileo bot have some assets!)
```
hdevalence, line 27: This should be done on Monday, immediately following the release, so that the `testnet-preview` instances have a randomized version of the *next* chain id, rather than of the current one.

hdevalence: Hmm, looks like this is still in the Friday (pre-release) section rather than the Monday (post-release) section ?

redshiftzero: so I moved the creation of the testnet directory and initial team allocations to Monday, but I left the step where we update the list of discord bot addresses (and adding allocations for them) on Friday (such that the faucet addresses posted during the week are included) - do we want to do them both on Monday?

.github/ISSUE_TEMPLATE/release_testnet.md
```diff
@@ -0,0 +1,37 @@
...
+- [ ] Check in with team again in a release meeting and update the GitHub milestone to ensure it represents what will make it into the testnet.
+
+Friday: Prepare a PR (for merge on Monday) with the following:
+- [ ] Update `discord-addresses` file using the Galileo bot (`cargo run --release history --channel <channel>`)
+- [ ] Update `testnet/<moon>/allocations.csv` file with the above Discord faucet addresses
+- [ ] Update the User Guide to mention the git tag
+
+Monday (release day):
+- [ ] Create new git tag e.g. `006-orthosie`, push to shared remote: `git tag -a <tag_name>` - must be annotated tag for Vergen build. This will begin the release process. Monitor the GitHub action to ensure it completes.
+- [ ] Update peer configuration on `testnet.penumbra.zone`: in Tendermint's `config.toml`, update the `bootstrap-peers` and `persistent-peers` fields to contain the IPs of the Penumbra-operated full nodes.
```
zbuc, line 33: The node ID changes during redeployment unless the consensus key is persisted between testnets, which the automation currently doesn't support.

zbuc: i.e. you need to bring up the new testnet, check the node id(`curl -s http://testnet.penumbra.zone:26657/status | jq ".result.node_info.id"` or look at `priv_key.json` in the tendermint data directory), then update the `persistent-peers` to contain the main Penumbra validator's node ID + IP

@avahowell should we be updating `bootstrap-peers` anywhere during a deployment?

.github/ISSUE_TEMPLATE/release_testnet.md
hdevalence: I think it's simpler to do them both on Monday -- it means there's a week-long lead time for faucet-to-genesis inclusion, but it's a simpler workflow, which seems like a win.

.github/ISSUE_TEMPLATE/release_testnet.md
redshiftzero: ah so I linked out to https://guide.penumbra.zone/main/pd/join-testnet/fullnode.html (in the below point) for how to fetch the updated node ID and update the peers since that's described there

PR #866, opened 2022-05-24
thread merkle proofs into general-purpose state query api

### LINKED ISSUE #804: Thread Merkle proofs through general-purpose state query API

Following #803, which should be initially implemeted without proofs, we should extend the RPC messages to add Merkle proofs. Adding this to the proto is relatively straightforward.  The proofs also exist in the underlying JMT store.  However, all of the other levels of abstraction discard proof data, so we'll need to create query methods that thread the inclusion proofs up through the `WriteOverlay` and other layers.

- [ ] Add ICS23 proofs to the proto messages
- [ ] Plan API extensions that thread Merkle proofs from the JMT up to the `Info` service that handles RPC:
  - (fill in list before starting)

avahowell:
closes #804. requires https://github.com/penumbra-zone/jmt/pull/43

### DIFF
--- a/pd/src/info/specific.rs
+++ b/pd/src/info/specific.rs
@@ -97,16 +97,33 @@ impl SpecificQuery for Info {
     ) -> Result<tonic::Response<KeyValueResponse>, Status> {
         let state = self.state_tonic().await?;
         state.check_chain_id(&request.get_ref().chain_id).await?;
-
         let key = request.into_inner().key_hash;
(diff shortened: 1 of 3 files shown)


hdevalence:
@avahowell I think this might need a `cargo update` now that we commit a lockfile (the JMT pr is merged)

avahowell:
updated, seems to be working

Issue #755, opened 2022-05-04
Implement `ViewService::witness`


hdevalence:
To implement this query, the `ViewService` needs to have a copy of the same `NoteCommitmentTree` data as used by the worker task.  One way to do this would be to have the `ViewService` load it from the database.  A better way would be to share the worker's `note_commitment_tree` added in #752 with the view service, by wrapping it in a [Tokio `RwLock`](https://docs.rs/tokio/latest/tokio/sync/struct.RwLock.html) (to share access) and an `Arc` (to share ownership), adding a
```
note_commitment_tree: Arc<RwLock<NoteCommitmentTree>>,
```
to the `ViewService`.  The `note_commitment_tree` should be returned from `Worker::new`, which will need to become an async function.

Then, the view service can answer the `AuthPathsRequest` by:

1. Acquiring a read lock on the NCT;
2. Reading the NCT root;
3. Obtaining an auth path for each requested note commitment;
4. Releasing the read lock on the NCT;

It's important to lock the NCT once, rather than for each operation, so we can provide the semantics of the request, which is that all of the auth paths should be made relative to the *same* root (since our transactions use a common merkle root for all proofs).

PR #877, opened 2022-05-26
Use shared NCT between ViewService and Worker

zbuc:
First part of #755. This wraps the NCT with `Arc<RwLock<>>` so it's shared between the `ViewService` and view `Worker`.


hdevalence (approved):
LGTM

PR #889, opened 2022-05-26
Implement witness call in ViewService

### LINKED ISSUE #755: Implement `ViewService::witness`

zbuc:
Closes #755

### DIFF
--- a/view/src/service.rs
+++ b/view/src/service.rs
@@ -14,6 +14,8 @@ use penumbra_proto::{
     crypto as pbc, transaction as pbt,
     view::{self as pb, view_protocol_server::ViewProtocol, StatusResponse},
 };
+use penumbra_tct::{Commitment, Proof};
+use penumbra_transaction::WitnessData;
 use tokio::sync::{mpsc, RwLock};
 use tonic::{async_trait, transport::Channel};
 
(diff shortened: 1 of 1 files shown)


view/src/service.rs
```diff
@@ -187,6 +189,38 @@ impl ViewProtocol for ViewService {
...
+
+        // Read the NCT root
+        let anchor = nct.root();
+
+        // Obtain an auth path for each requested note commitment
+        let requested_note_commitments = request
+            .get_ref()
+            .note_commitments
+            .iter()
+            .map(|nc| {
```
hdevalence, line 204: We shouldn't expect here, since that will bring down the whole server on a malformed request.

Instead, if you do .map(Commitment::try_from), you'll get an iterator of Result s, and if you do .collect::<Result<Vec<_>, _>>()? you can collect the iterator of Results into a Result of a Vec, then bubble up any conversion errors.

zbuc: I was wondering about that. FYI we `expect` on protobuf conversions from request values elsewhere in the service (e.g. https://github.com/penumbra-zone/penumbra/blob/main/view/src/service.rs#L160), I can fix those up as well?

hdevalence: Yeah, that sounds good!

hdevalence (approved):

Issue #881, opened 2022-05-26
Change `NoteRecord` to record note positions;


hdevalence:
To build transaction plans, the wallet logic needs to know the position of each note. This information is known by the view service but not currently recorded in the `NoteRecord`:

- [ ] Add a `tct::Position` to the `NoteRecord` domain type, and a `uint64` position to the proto type;
- [ ] Record the position in the `notes` table;

PR #890, opened 2022-05-26
Implement `plan::validator_definition` in `wallet-next`

### LINKED ISSUE #884: Add `plan::validator_definition` to `wallet-next`
Following the `plan::delegate` example, port the `wallet::ClientState::validator_definition` method to `wallet_next::plan::validator_definition`.

zbuc:
I noticed that this module isn't being imported anywhere so we've probably got some compile-time errors in it still, but we can fix those later.

Closes #884


zbuc:
This has a dependency on #881 and will need to be changed to use the note positions properly after #881 is merged

PR #893, opened 2022-05-26
add chain params to view protocol

### LINKED ISSUE #879: Extend the `ViewProtocol` to allow fetching the `ChainParams`
Wallet logic needs to know the chain parameters to be able to build transactions, and it's better if it just requests it from the view service (inside the privacy boundary) than a full node.

avahowell:
closes #879

### DIFF
--- a/view/src/service.rs
+++ b/view/src/service.rs
@@ -10,6 +10,7 @@ use penumbra_crypto::{
     keys::{DiversifierIndex, FullViewingKeyHash},
 };
 use penumbra_proto::{
+    chain as pbp,
     client::oblivious::oblivious_query_client::ObliviousQueryClient,
     crypto as pbc, transaction as pbt,
     view::{self as pb, view_protocol_server::ViewProtocol, StatusResponse},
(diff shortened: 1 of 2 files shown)


view/src/service.rs
```diff
@@ -233,4 +234,21 @@ impl ViewProtocol for ViewService {
...
         Ok(tonic::Response::new(witness_data.into()))
     }
+
+    async fn chain_params(
+        &self,
+        _request: tonic::Request<pb::ChainParamsRequest>,
+    ) -> Result<tonic::Response<pbp::ChainParams>, tonic::Status> {
+        self.check_worker().await?;
+
+        // NOTE: should we check the chain ID here?
```
hdevalence, line 244: Yeah, probably. There's a merge conflict now, so I'll rebase the PR and add the check while I do.

hdevalence: Hmm, wait. How does the wallet client know the chain ID? It's supposed to be stateless.

hdevalence: Removed the parameter in https://github.com/penumbra-zone/penumbra/pull/893/commits/be562908215a29135414917b19b51831a446b390

PR #898, opened 2022-05-27
Add status() to pviewd.rs

### LINKED ISSUE #839: Correctly report sync status in wallet service
Currently the wallet service doesn't correctly report its synced status, because it didn't have knowledge of the Tendermint RPC endpoint.

Now that we set that in the command line options, we can correctly report sync status by: (1) fetching the JSON from the /status endpoint (2) extracting the block height from the JSON (3) comparing the remote node's block height with the local sync height. 

I think it would make sense to report that the wallet service is synced, as long as it's within 1 block of the tip. That way, we won't report that we're still syncing if a new block was produced right before we report our status.

aubrika:
Closes #839

### DIFF
--- a/view/src/bin/pviewd.rs
+++ b/view/src/bin/pviewd.rs
@@ -5,11 +5,13 @@ use penumbra_crypto::FullViewingKey;
 use penumbra_proto::client::oblivious::oblivious_query_client::ObliviousQueryClient;
 use penumbra_proto::client::oblivious::ChainParamsRequest;
 use penumbra_proto::view::view_protocol_server::ViewProtocolServer;
+use penumbra_proto::view::StatusResponse;
 use penumbra_view::ViewService;
 use std::env;
 use std::str::FromStr;
 use structopt::StructOpt;
 use tonic::transport::Server;
+use tracing::instrument;
 #[derive(Debug, StructOpt)]
 #[structopt(
     name = "pviewd",
@@ -81,7 +83,7 @@ async fn main() -> Result<()> {
 
             let storage = penumbra_view::Storage::load(opt.sqlite_path).await?;
 
-            let service = ViewService::new(storage, client).await?;
+            let service = ViewService::new(storage, client, opt.node, opt.tendermint_port).await?;
 
             tokio::spawn(
                 Server::builder()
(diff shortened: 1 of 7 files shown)


view/src/bin/pviewd.rs
```diff
@@ -34,6 +36,49 @@ struct Opt {
     pd_port: u16,
 }
 
+impl Opt {
+    #[instrument(skip(self))]
+    pub async fn status(&self) -> Result<StatusResponse, anyhow::Error> {
```
hdevalence, line 41: We'll need to get the status from inside of the `ViewService` while processing the status request, so we need the code that fetches the Tendermint status to be in the `ViewService`.

We should probably change `ViewService::new` to take `node` and `tendermint_port` and store them internally, so we can put this code in a
```rust
impl ViewService {
    /// Returns the `latest_block_height` and `catching_up` fields from the full node we're syncing from.
    async fn node_block_height(&self) -> Result<(u64, bool)> { ... }
}
```
that we can call from inside the `ViewService`.

aubrika: Cool, will fix!

view/src/bin/pviewd.rs
```diff
@@ -34,6 +36,49 @@ struct Opt {
...
+        let chain_height = sync_info
+            .get("max_peer_block_height")
+            .and_then(|c| c.as_u64())
+            .ok_or_else(|| anyhow::anyhow!("could not parse JSON response"))?;
+
+        Ok(StatusResponse {
+            sync_height,
+            synchronized,
+            chain_height,
+        })
```
hdevalence, line 78: We should change the `StatusResponse` to:
```proto
message StatusResponse {
    // The height the view service has synchronized to so far
    uint64 sync_height = 1;
    // Whether the view service is catching up with the chain state
    bool catching_up = 2;
}
```

Then `sync_height` is set to the sync height of the view service, and `catching_up` could be determined as follows:
```
let height_diff = latest_block_height.checked_sub(sync_height).ok_or_else(|| anyhow!("sync height ahead of node height"))?;
let catching_up = match (node_catching_up, height_diff) {
    // We're synced to the same height as the node
    (false, 0) => false,
    // We're one block behind, and will learn about it soon, close enough
    (false, 1) => false,
    // We're behind the node
    (false, _) => true,
    // The node is behind the network
    (true, _) => true,
};
```

PR #899, opened 2022-05-27
Implement assets in pviewd/pcli-next

### LINKED ISSUE #880: Extend the `ViewProtocol` to allow fetching an `AssetCache`
Like #879, it's better to fetch this from inside the privacy boundary, now that we don't have any state in the wallet itself.

Unlike #879, we don't already store this state in the view service, so as part of this work, we also have to have the view service fetch assets and store them in the SQLite database.  To do this, we should:

- [ ] Add a new table, `assets`, with an asset id (blob) column and a denom (string) column;
- [ ] Port the asset fetching code from current `pcli` to the view crate.  Because we don't have any new assets, we can just fetch them at genesis for now.
- [ ] Add an RPC method to the view protocol that pulls the entire asset cache. (We could do specific lookups later).

zbuc:
Closes #880

### DIFF
--- a/crypto/src/asset/id.rs
+++ b/crypto/src/asset/id.rs
@@ -1,4 +1,5 @@
 use ark_ff::fields::PrimeField;
+use ark_serialize::CanonicalDeserialize;
 use decaf377::FieldExt;
 use once_cell::sync::Lazy;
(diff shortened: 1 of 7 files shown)


hdevalence (approved):
LGTM!

Issue #802, opened 2022-05-12
Extract state keys from component views


hdevalence:
**Is your feature request related to a problem? Please describe.**

Currently, we isolate all of the keys used to query the state into the implementations of the `View` traits.  This is good because it isolates all of the keys into one place (the read and write methods of the view traits), but it's still difficult to see a comprehensive list of all of the keys we use to store data in the state.

**Describe the solution you'd like**

Each component crate should have a (public) `crate::state_key` module.  This module should contain a bunch of bare functions, all of which return a `KeyHash`, and which take whatever arguments they need to produce the key, for instance:

```rust
fn token_supply(asset_id: &asset::Id) -> KeyHash {
    format!("shielded_pool/assets/{}/token_supply", asset_id).into()
}

fn known_assets() -> KeyHash { 
    "shielded_pool/known_assets".into()
}
```

Then the rest of the component crate can access these keys as `state_key::token_supply(asset_id)`, `state_key::known_assets()`, etc:
```rust
async fn token_supply(&self, asset_id: &asset::Id) -> Result<Option<u64>> {
    self.get_proto(state_key::token_supply(asset_id)).await
}
```

This is a totally mechanical refactoring, so it should be easy, if a bit tedious (`rg format!` would be helpful to catch any stray keys), but the upshot is that at the end, we'll have one file in each component that has a comprehensive list of keys we use to query the state.

borngraced:
Interested...can assign please

hdevalence:
> Interested...can assign please

@borngraced sure! I assigned you just now. One thing though is that this might end up on our critical path, so it'd be preferable if you could make a draft PR as soon as you start working on it, rather than when you're finished.  That way, if we discover that some later tasks need this in a time-critical way, and that we need to pick it up and carry it over the finish line to avoid blocking work, we'll be able to use the work you did.

borngraced:
Ok sure will do that..👍

PR #901, opened 2022-05-27
state keys from component views refactoring in #802

borngraced:
Refactored state keys by extracting keys from component views to a pub(crate) module `state_keys` within component crates


chain/src/view.rs
```diff
@@ -71,7 +71,7 @@ pub trait View: StateExt {
     /// Gets the current block timestamp from the JMT
     async fn get_block_timestamp(&self) -> Result<Time> {
         let timestamp_string: String = self
-            .get_proto(b"block_timestamp".into())
+            .get_proto(state_key::block_height())
```
avahowell, line 74: this should be block_timestamp instead of block_height

chain/src/view.rs
```diff
@@ -80,7 +80,7 @@ pub trait View: StateExt {
 
     /// Writes the block timestamp to the JMT
     async fn put_block_timestamp(&self, timestamp: Time) {
-        self.put_proto(b"block_timestamp".into(), timestamp.to_rfc3339())
+        self.put_proto(state_key::block_height(), timestamp.to_rfc3339())
```
avahowell, line 83: ditto

chain/src/view.rs
borngraced: oh blocker noticed now...sorry!

ibc/src/component/connection.rs
```diff
@@ -279,12 +281,7 @@ pub trait View: StateExt + Send + Sync {
         connection: ConnectionEnd,
     ) -> Result<()> {
         self.put_domain(
-            format!(
-                "{}/connections/{}",
-                COMMITMENT_PREFIX,
-                connection_id.as_str()
-            )
-            .into(),
+            state_key::connection(COMMITMENT_PREFIX, connection_id),
```
hdevalence, line 284: Since `COMMITMENT_PREFIX` is a crate-level constant, it would be cleaner if it were folded into the definition of `state_key::connection`, so that this call would be
```
state_key::connection(connection_id)
```
with the `COMMITMENT_PREFIX` inserted internally.

shielded-pool/src/state_key.rs
```diff
@@ -0,0 +1,55 @@
...
+    format!("shielded_pool/known_assets").into()
+}
+
+pub fn denom_by_asset(asset_id: &Asset_Id) -> KeyHash {
+    format!("shielded_pool/assets/{}/denom", asset_id).into()
+}
+
+pub fn register_denom(denom: &Denom) -> KeyHash {
+    format!("shielded_pool/assets/{}/denom", denom).into()
+}
```
hdevalence, line 23: These are the same function; the `register_denom` function should be removed, and replaced by calls to `denom_by_asset`.

That way, we just have one function _per key_, regardless of the context in which that key is used.

shielded-pool/src/state_key.rs
```diff
@@ -0,0 +1,55 @@
...
+    format!("shielded_pool/note_source/{}", note_commitment).into()
+}
+
+pub fn set_compact_block(compact_block: &CompactBlock) -> KeyHash {
+    format!("shielded_pool/compact_block/{}", compact_block.height).into()
+}
+
+pub fn compact_block(height: u64) -> KeyHash {
+    format!("shielded_pool/compact_block/{}", height).into()
+}
```
hdevalence, line 35: Ditto above, `set_compact_block` should be removed and replaced by calls to `compact_block`.

shielded-pool/src/state_key.rs
```diff
@@ -0,0 +1,55 @@
...
+
+pub fn nct_anchor_valid(nct_anchor: &Root) -> KeyHash {
+    format!("shielded_pool/valid_anchors/{}", nct_anchor).into()
+}
+
+pub fn claimed_anchor(anchor: &Root) -> KeyHash {
+    format!("shielded_pool/valid_anchors/{}", anchor).into()
+}
+
+pub fn spend_nullifier(nullifier: &Nullifier) -> KeyHash {
```
hdevalence, line 49: This should be `spent_nullifier` to match the key, so that it's reflecting the state ("spent") that's recorded in that part of the tree, rather than the action of writing the nullifier there (spending it).

shielded-pool/src/state_key.rs
```diff
@@ -0,0 +1,55 @@
...
+    format!("shielded_pool/tct_anchor/{}", nct_anchor).into()
+}
+
+pub fn nct_anchor_valid(nct_anchor: &Root) -> KeyHash {
+    format!("shielded_pool/valid_anchors/{}", nct_anchor).into()
+}
+
+pub fn claimed_anchor(anchor: &Root) -> KeyHash {
+    format!("shielded_pool/valid_anchors/{}", anchor).into()
+}
```
hdevalence, line 47: Ditto above, these should be merged into one function.  It's used in different places, and neither name captures both, so I'd suggest calling it `anchor_lookup`.

shielded-pool/src/state_key.rs
hdevalence: Edit: for the same reason as the comment above, I think it'd be better to call this `spent_nullifier_lookup`.

shielded-pool/src/component.rs
```diff
@@ -491,50 +486,40 @@ pub trait View: StateExt {
...
         tracing::debug!(?height, ?nct_anchor, "writing anchor");
 
         // Write the NCT anchor both as a value, so we can look it up,
-        self.put_domain(
-            format!("shielded_pool/tct_anchor/{}", height).into(),
-            nct_anchor,
-        )
-        .await;
+        self.put_domain(state_key::nct_anchor_tct(&nct_anchor), nct_anchor)
+            .await;
```
hdevalence, line 519: This isn't quite the right translation: the key should be derived from the _height_, not the anchor, so that we can query for what the anchor at a given height is.

To fix this, let's replace `nct_anchor_tct(nct_anchor: &Root)` by `anchor_by_height(height: u64)`, to match the original code.

shielded-pool/src/state_key.rs
borngraced: can just use generic right? Where T can be denom asset_id or denom??

hdevalence (approved):
LGTM! I patched up two small changes.

hdevalence:
Merged, thanks so much for your help with this!

PR #902, opened 2022-05-27
implement ABCI query interface for JMT

avahowell:
this adds ABCI query handling to `pd`, currently only handling one path: `/jmt/key`, which queries the JMT state for an arbitrary key. The query response is a proto-encoded `KeyValueResponse` that contains the value as well as an ICS23 CommitmentProof of that value's inclusion.

### DIFF
--- a/pd/src/info.rs
+++ b/pd/src/info.rs
@@ -67,13 +67,13 @@ impl Info {
 
         match query.path.as_str() {
             "/jmt/key" => {
-                let state = self.state_tonic().await?;
-                let (value, proof) = state
-                    .read()
-                    .await
-                    .get_with_proof(query.data.to_vec())
-                    .await
-                    .map_err(|e| tonic::Status::internal(e.to_string()))?;
+                let height: u64 = query.height.into();
+                let key = query.data.to_vec();
+
+                let proof = jmt::JellyfishMerkleTree::new(&self.storage)
+                    .get_with_ics23_proof(key, height)
+                    .await?;
+                let value = proof.value.clone();
 
                 let commitment_proof = ics23::CommitmentProof {
                     proof: Some(ics23::commitment_proof::Proof::Exist(proof)),
@@ -84,8 +84,6 @@ impl Info {
                     proof: Some(commitment_proof),
                 };
 
-                let height = state.get_block_height().await?;
-
                 Ok(abci::response::Query {
                     code: 0,
                     key: query.data,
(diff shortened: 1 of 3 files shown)


pd/src/info.rs
```diff
@@ -56,9 +58,48 @@ impl Info {
         &self,
         query: abci::request::Query,
     ) -> Result<abci::response::Query, anyhow::Error> {
-        tracing::warn!(?query, "unhandled query");
+        tracing::info!(?query);
+
+        match query.path.as_str() {
+            "/jmt/key" => {
```
avahowell, line 64: note: should we call this something else? for reference, the cosmos SDK implements ABCI query interfaces separately for each module: the kv query for ibc is `store/ibc/key` for instance. Since we don't have modules, I think just having a single kv query interface is probably good, but maybe we don't want to call it `jmt` as to avoid binding the name to an implementation detail

pd/src/info.rs
```diff
@@ -56,9 +58,48 @@ impl Info {
...
+
+                let height = state.get_block_height().await?;
+
+                Ok(abci::response::Query {
+                    code: 0,
+                    key: query.data,
+                    log: "".to_string(),
+                    value: kvr.encode_to_vec().into(),
+                    // NOTE: the ABCI query proof is not the same as ICS-23 proofs.
+                    proof: None,
```
avahowell, line 90: this is a bit ugly. the ABCI query interface includes a `proof` field, but it's a tendermint::merkle:Proof, which is *not* the same as an ics23 proof (and not as general as an ics23 proof). SDK chains actually have to convert this to an ics23 proof to work. Instead, we don't use this field, and instead directly include the ics23 proof in the `value` field of the query (encoded into the KeyValueResponse).

avahowell:
This should be the MVP functionality for our relayers to function

pd/src/info.rs
hdevalence: How about `/state/key/`?

pd/src/info.rs
hdevalence: Hmm, that is ugly. Maybe we can get them to fix this upstream? It'd be a breaking change, but getting it into the ecosystem at some poitn seem better...

pd/src/info.rs
```diff
@@ -66,31 +65,32 @@ impl Info {
...
+                    proof: Some(ics23::commitment_proof::Proof::Exist(jmt_proof)),
                 };
 
-                let kvr = KeyValueResponse {
-                    value,
-                    proof: Some(commitment_proof),
+                let op = tendermint::merkle::proof::ProofOp {
+                    field_type: "jmt:v".to_string(),
+                    key,
+                    data: commitment_proof.encode_to_vec(),
```
hdevalence, line 84: I'm surprised this is all we have to do -- where does `jmt:v` come from?

avahowell: https://github.com/cosmos/ibc-go/blob/main/modules/core/23-commitment/types/utils.go#L9 this is how relayers convert ABCI Proof to ICS23 proof

PR #908, opened 2022-05-29
Pull together the pieces into `pcli-next`

### LINKED ISSUE #874: Reassemble client microservices into `pcli-next`, and then use it to replace `pcli`

Following #587, which splits up the responsibility for the existing client code into microservices, we should reassemble that code into a new version of `pcli` that uses them. We've been doing this in duplicate `-next` crates, so that we can do the rewrite without blocking other work.  This issue tracks the remaining work to do before we can replace `pcli` with `pcli-next`.

- [x] Port over functionality from the current `wallet` crate:
  - [x] #755
  - [x] #839
  - [x] #879
  - [x] #880
  - [x] #881
  - [x] #882
  - [x] #883
  - [x] #884
- [ ] Copy the `pcli` source over to `pcli-next`, and update it to use the new API (sketch):
  - [x] Add dependencies on `wallet-next`, `view`, `custody`, and remove the dependency on `wallet`;
  - [x] Remove `state.rs` and `ClientStateFile`, since all the state will be in the view service;
  - [x] Copy the `Wallet` structure from the `wallet` crate, and remove the address labels, leaving only the key material (or maybe we want `LegacyWallet` that's shaped to be able to pull out data from the existing json?);
  - [x] Change the storage from a single JSON file to a directory with two files, `view.sqlite` and `wallet.json`;
  - [x] Change `main.rs` to instantiate a `ViewService` and a `SoftHSM`, with the `ViewService` backed by `view.sqlite` and the `SoftHSM` instantiated with the key stored in the `Wallet` in `wallet.json`;
  - [ ] Patch up the rest of the code until it works again.

hdevalence:
Closes #874.

Current status:

- [x] compiles
- [x] per tracing output, syncs the chain
- [x] returns notes

Functionality to test:
- [x] `balance`
- [x] `tx send`
- [x] `stake delegate`
- [x] `stake show`
- [x] `stake undelegate`
- [ ] `validator upload-definition`


pcli-next/src/command/balance.rs
```diff
@@ -0,0 +1,109 @@
+use anyhow::Result;
+use comfy_table::{presets, Table};
+use penumbra_crypto::{keys::DiversifierIndex, FullViewingKey, Value};
+use penumbra_view::ViewClient;
+use structopt::StructOpt;
+
+// TODO
```
avahowell, line 7: stray TODO?

avahowell:
lgtm

pcli-next/src/command/balance.rs
hdevalence: Yeah, looks like it. I think I might have been about to suggest that we move `balance` into a subcommand of a new `view` command.

PR #924, opened 2022-05-31
Update pd testnet guide

elsehow:
- Recommends everyone wipe RocksDB.
- Fixes Latex rendering issue with `$HOME`.


hdevalence (approved):
LGTM.  We should upstream a fix for mdbook-katex to not mess with code blocks.

PR #933, opened 2022-06-01
don't error on IBC action in shielded pool

avahowell:
This was preventing the IBC component from receiving IBC actions


hdevalence:
I think a better fix would be to ignore unknown actions? Each component should ideally not have to be aware of every possible message. We ended up with this problem because there was an unknown action, but if we just add the action to the whitelist, the next time we add a new action we'll be back where we started.

PR #934, opened 2022-06-01
fix ibc enabled: use chain params from state

avahowell:
prior to this change, we only set an `enabled` boolean on init-chain, meaning that whether IBC was enabled would depend on if the node had restarted. this PR instead directly pulls the chain parameters when validating IBC transactions.


hdevalence (approved):
LGTM.  We should avoid having components keep in-memory state unless it's really necessary.

PR #944, opened 2022-06-01
Test tct-property-test using release build

plaidfinch:
The property tests crate for the TCT takes some time to run, because it does a lot of hashing; this PR speeds it up by running it in release mode. This decreases the total CI time spent in the test suite from 15 minutes to 9 minutes.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -32,6 +32,11 @@ jobs:
       - uses: actions-rs/cargo@v1
         with:
           command: test
+          args: --exclude penumbra-tct-property-test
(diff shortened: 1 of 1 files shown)


plaidfinch:
Wondering whether we should also run the tests in the penumbra-tct crate itself in release mode in CI. The only downside I see is that we'd miss `debug_assert!`s, but maybe those should be `assert!`s instead. Thoughts @hdevalence?

hdevalence:
Rather than changing the internals of the code, I think we might want to move towards a split model where we have some tests that run pre-merge, and some longer tests that run post-merge and alert on failure.

Issue #936, opened 2022-06-01
Initial event recording system


hdevalence:
**Is your feature request related to a problem? Please describe.**

We want to be able to support [Tendermint events](https://rustdoc.penumbra.zone/main/tendermint/abci/struct.Event.html), so that Penumbra data is indexable.

**Describe the solution you'd like**

To do this, we need (1) to have a way to record events (2) to have a systematic way to produce events.

To do (1), we should add a `Context` type to `penumbra_component`, and thread it through the methods in the `Component` trait.  The `Context` needs basically three methods: initialization, event recording, and extracting a list of events.

In the future, we want to be able to support spawning tasks to perform concurrent work.  Because spawned futures need to be `'static`, this means that it'd be convenient if the method signatures on the `Component` trait didn't impose lifetime bounds.  For this reason, it seems preferable to make the `Context` an owned handle to some shared data:

```rust
#[derive(Clone)]
pub struct Context {
    inner: Arc<Mutex<Option<Vec<Event>>>>,
}

impl Context {
    pub fn new() -> Self {
        Self { inner: Arc::new(Mutex::new(Some(Vec::new()))) }
    }
    pub fn record(&self, e: Event) {
        self.inner.lock().as_mut().expect("record called after into_events").push(e);
    }
    pub fn into_events(self) -> Vec<Event> {
        self.inner.lock().take()
    }
}
```

Then, to thread this through the API, we can add a `ctx: Context` parameter to each `Component` method.  Because we take an owned context, rather than a borrowed one, a component may need to clone the context it passes to its children, but this is cheap, because the inner data is `Arc`d, so this is just incrementing a refcount.

To do (2), we want to have a way to produce events.  The `Event` constructor is relatively ergonomic:
```rust
use tendermint::abci::{Event, EventAttributeIndexExt};

let event = Event::new(
    "app",
    vec![
        ("key1", "value1").index(),
        ("key2", "value2").index(),
        ("key3", "value3").no_index(), // will not be indexed
    ],
);
```
However, this API is operating at the level of string keys and values.  We want to be able to see all of the events we define at a glance.  To do this, I think we should use a similar approach as in #802.  Each component crate should define a top-level `event` module.  That module should have functions like
```rust
fn readable_name(data: Whatever) -> Event {
    Event::new(
        "app", // TODO: check what should go here
        vec![
            ("my_event_key", data.to_string()).index(),
        ],
    )
}
```
[...]

PR #948, opened 2022-06-02
Add ABCI event recording capability to penumbra Components

avahowell:
this PR implements part of #936 , adding a `Context`, threading it through the component api, and reporting events back to ABCI responses. Currently, no events are created, just the context functionality itself.

### DIFF
--- a/ibc/src/component/client.rs
+++ b/ibc/src/component/client.rs
@@ -25,7 +25,7 @@ use ibc::{
     },
 };
 use penumbra_chain::{genesis, View as _};
-use penumbra_component::Component;
+use penumbra_component::{Component, Context};
 use penumbra_proto::ibc::{
     ibc_action::Action::{CreateClient, UpdateClient},
     IbcAction,
@@ -67,8 +67,8 @@ impl Component for Ics2Client {
         self.state.put_client_counter(ClientCounter(0)).await;
     }
 
-    #[instrument(name = "ics2_client", skip(self, begin_block))]
-    async fn begin_block(&mut self, begin_block: &abci::request::BeginBlock) {
+    #[instrument(name = "ics2_client", skip(self, _ctx, begin_block))]
+    async fn begin_block(&mut self, _ctx: Context, begin_block: &abci::request::BeginBlock) {
         // In BeginBlock, we want to save a copy of our consensus state to our
         // own state tree, so that when we get a message from our
         // counterparties, we can verify that they are committing the correct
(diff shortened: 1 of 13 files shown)


ibc/src/component/client.rs
```diff
@@ -748,26 +748,31 @@ mod tests {
...
         assert_eq!(client_component.state.client_counter().await.unwrap().0, 1);
 
         // now try update client
 
-        Ics2Client::check_tx_stateless(&update_client_tx).unwrap();
+        Ics2Client::check_tx_stateless(ctx.clone(), &update_client_tx).unwrap();
         // verify the ClientUpdate proof
         client_component
-            .check_tx_stateful(&update_client_tx)
+            .check_tx_stateful(ctx.clone(), &update_client_tx)
```
avahowell, line 769: it occurred to me that the new Context will be useful for testing components (can assert against events etc)

hdevalence:
It'd be useful to create at least one event, just so we can check that there's not something subtle in the API that blocks it from actually being useful.  Maybe an easy one to start would be adding note commitment / nullifier events in the shielded pool? That would also be useful for the view service, since we'd be able to query for the full transactions that produced the notes we detected (though we'd probably want to make some chaff requests too...)

shielded-pool/src/event.rs
```diff
@@ -0,0 +1,13 @@
+use penumbra_chain::NoteSource;
+use penumbra_crypto::Nullifier;
+use tendermint::abci::{Event, EventAttributeIndexExt};
+
+pub fn spend(nullifier: Nullifier, source: NoteSource) -> Event {
+    Event::new(
+        "spend",
+        vec![
+            ("nullifier", nullifier.to_string()).index(),
+            ("source", source.to_string()).index(),
```
hdevalence, line 10: I don't think we need to index the `source` here, since the event is already happening in the context of the transaction that `source` is pointing to. So we can just record the nullifier and omit the source.

chain/src/note_source.rs
```diff
@@ -87,3 +87,18 @@ impl std::fmt::Debug for NoteSource {
...
+                f.write_fmt(format_args!("NoteSource::Transaction({})", hex::encode(id)))
+            }
+            NoteSource::Genesis => f.write_fmt(format_args!("NoteSource::Genesis")),
+            NoteSource::FundingStreamReward { epoch_index } => f.write_fmt(format_args!(
+                "NoteSource::FundingStreamReward({})",
+                epoch_index
+            )),
+        }
+    }
+}
```
hdevalence, line 104: I think we should probably drop this impl (for the reason below), but if we did keep it, rather than printing out a Rust-style enum, we should just print the hex-formatted bytes.  The point of the `NoteSource` is to be able to stuff some extra data into something that looks and feels like a transaction hash.

PR #955, opened 2022-06-02
935 await change rpc

### LINKED ISSUE #935: Add an RPC to the view service that allows awaiting change outputs

While doing note management, it's useful to learn when a change note is available for spending, or to be able to wait for a change note to become available to spend.

In this case, we already know the note and its note commitment, but we want to be notified when it's detected through scanning.

We can support this use case by adding a new RPC to the view protocol:
```proto
rpc AwaitChange(AwaitChangeRequest) returns (NoteRecord);

message AwaitChangeRequest {
  crypto.NoteCommitment note_commitment = 2;
}
```
This RPC should have the following behavior:

- if the provided `note_commitment` is known to the view service, it returns the corresponding `NoteRecord` immediately.
- if the provided `note_commitment` is not known to the view service, it waits for it to become known (however long that takes), and then returns the corresponding `NoteRecord`.

Implementing the first part is easy with the current view service code: just query the storage. Implementing the second part is more work, because we need a way to notify tasks when notes are detected by the worker scanning new blocks and committed to storage.

To do this, we can use a [`broadcast` channel](https://docs.rs/tokio/latest/tokio/sync/broadcast/index.html) to the `Storage`.  The sender side of the channel is held by the `Storage`, and when data is committed to the database, the `Storage` should also feed every `NoteRecord` into the sender side of the channel.

The logic to process the request would then look like:

- Create a new `Receiver` by [calling `subscribe`](https://docs.rs/tokio/latest/tokio/sync/broadcast/struct.Sender.html#method.subscribe) on the sender side of the channel
- Check if the provided `note_commitment` is in the database, and return it if so;
- Otherwise, call `recv().await` in an infinite `loop`, and each time a new NoteRecord is received, check if it matches the requested `note_commitment` and return it if so.

The ordering of these steps is important for the following reason: `subscribe` will return a receiver that sees all messages *after* the call to subscribe, so by creating the receiver before querying the database, we ensure that, even if our database read raced a write and missed a record, we'd still be notified of the record through the channel.

This could be bundled up into an `await_change` method on the `Storage`, which has access to the sender to call `subscribe` on.

Handling timeouts on the server side doesn't seem important for now, since the future will be dropped if the client stops holding the connection open.  The caller on the client side can choose to apply a timeout if they want.

aubrika:
Closes #935


view/src/service.rs
```diff
@@ -211,6 +212,16 @@ impl ViewService {
             catching_up,
         })
     }
+
+    #[instrument(skip(self))]
+    pub async fn await_change(
```
hdevalence, line 217: I think we could probably elide this 
method and pass directly from the proto generated trait to the method on the underlying storage

view/src/service.rs
```diff
@@ -223,6 +234,31 @@ impl ViewProtocol for ViewService {
...
     >;
 
+    async fn await_change(
+        &self,
+        request: tonic::Request<pb::AwaitChangeRequest>,
+    ) -> Result<tonic::Response<pb::NoteRecord>, tonic::Status> {
+        self.check_worker().await?;
+        self.check_fvk(request.get_ref().fvk_hash.as_ref()).await?;
+
+        let note_commitment = request.get_ref().note_commitment.as_ref();
```
hdevalence, line 244: This might be cleaner with something like into_inner(), then ok_or_else(|| tonic::Status ... )? to get rid of the option, then try_into() to convert to the domain type. Then you just have the note commitment domain type and can pass it directly to the storage method

view/src/storage.rs
```diff
@@ -117,9 +122,49 @@ impl Storage {
...
             uncommitted_height: Arc::new(Mutex::new(None)),
+            scanned_notes_tx: broadcast::channel(10).0,
         })
     }
 
+    pub fn await_change(
+        &self,
+        note_commitment: tct::Commitment,
+    ) -> impl Future<Output = anyhow::Result<NoteRecord>> + '_ {
+        let mut rx = self.scanned_notes_tx.clone().subscribe();
```
hdevalence, line 133: I think the clone here isn't needed, since subscribe() will create an owned receiver, so it's already unlinked from the lifetime of the sender.

view/src/storage.rs
```diff
@@ -117,9 +122,49 @@ impl Storage {
...
+
+        async move {
+            match self.note_record_by_commitment(note_commitment).await? {
+                Some(record) => Ok(record),
+
+                None => loop {
+                    let record = rx.recv().await.unwrap();
```
hdevalence, line 140: Should we bubble up the error here?

view/src/storage.rs
```diff
@@ -412,6 +457,10 @@ impl Storage {
             )
             .execute(&mut tx)
             .await?;
+
+            //Broadcast note record to channel
+
+            self.scanned_notes_tx.send(note_record).unwrap();
```
hdevalence, line 463: This is slightly nitpicky, but maybe we want to skip sending the notes into the channel here, wait until /after/ calling commit, and then loop over all the records we just inserted and push them into the channel.

That way, we avoid the possibility of a data race, where someone could be notified with a NoteRecord before it's been committed to the database. 

I think this is unlikely to be a problem in practice, but ...

aubrika: > unlikely to be a problem in practice

last words famous enough to go ahead & fix it now

hdevalence (approved):
LGTM, especially with the streamlining from pairing.

PR #959, opened 2022-06-03
Specify units in metrics

zbuc:
The metrics library has the ability to specify units for metrics that Prometheus can use. This PR adds appropriate units to all existing metrics calls.


hdevalence:
nice find

PR #960, opened 2022-06-03
Quarantine

plaidfinch:


component/src/shielded_pool/component.rs
```diff
@@ -1,38 +1,51 @@
...
-            state,
             note_commitment_tree,
-            compact_block: Default::default(),
+            // If a previous component started writing into the compact block (i.e. staking
+            // recording rolled back nullifiers and notes), retrieve it
+            compact_block: state
+                .compact_block(state.get_block_height().await.expect("can get height"))
+                .await
+                .expect("can get compact block")
+                .unwrap_or_default(),
```
hdevalence, line 46: I don't understand how this can work -- if other components can write into the compact block, we can't keep it in memory.  This won't pick up any writes made by other components, because it's called when the components are being re-initialized, not after their writes.

Data sharing between components is *only* done through the `State`, so if we need other components to write into the `CompactBlock` (why?), we can't maintain an unsynchronized copy in memory.

component/src/shielded_pool/component.rs
```diff
@@ -168,31 +181,71 @@ impl Component for ShieldedPool {
...
-            .transaction_body
-            .actions
-            .iter()
-            .any(|action| matches!(action, Action::Undelegate { .. }));
+        let mut should_quarantine = tx.transaction_body.actions.iter().find_map(|action| {
+            if let Action::Undelegate(Undelegate {
+                validator_identity, ..
+            }) = action
+            {
+                Some(validator_identity)
```
hdevalence, line 189: How do we know this `validator_identity` is unique?

component/src/shielded_pool/component.rs
```diff
@@ -168,31 +181,71 @@ impl Component for ShieldedPool {
...
+                self.state
+                    .validator_bonding_state(identity_key)
+                    .await
+                    .expect("validator lookup in state succeeds")
+                    .expect("validator is present in state"),
+                validator::BondingState::Unbonded
+            ) {
+                should_quarantine = None;
+            }
+        }
```
hdevalence, line 207: Let's extract this logic into a `should_quarantine` method.

component/src/shielded_pool/state_key.rs
```diff
@@ -37,3 +37,15 @@ pub fn spent_nullifier_lookup(nullifier: &Nullifier) -> KeyHash {
 pub fn commission_amounts(height: u64) -> KeyHash {
     format!("staking/commission_amounts/{}", height).into()
 }
+
+pub fn quarantined_note_source(note_commitment: &note::Commitment) -> KeyHash {
+    format!("shielded_pool/quarantined_note_source/{}", note_commitment).into()
+}
```
hdevalence, line 43: This should just be the same as `note_source`; otherwise, how can a client query the source of a note, without knowing whether or not it was already quarantined?

component/src/shielded_pool/component.rs
```diff
@@ -541,33 +764,178 @@ pub trait View: StateExt {
...
+    // Unschedule the unquarantining of all notes and nullifiers for the given validator, in any
+    // epoch which could possibly still be unbonding
+    async fn unschedule_unquarantine_all(&self, identity_key: IdentityKey) -> Result<()> {
+        let height = self.get_block_height().await?;
+        let epoch_duration = self.get_epoch_duration().await?;
+        let this_epoch = Epoch::from_height(height, epoch_duration);
+        let unbonding_epochs = self.get_chain_params().await?.unbonding_epochs;
+
+        // We will edit the compact block to note the rolled back nullifiers and commitments
+        let mut compact_block = self.compact_block(height).await?.unwrap_or_default();
```
hdevalence, line 913: We cannot edit the compact block like this, because it's stored in memory in the shielded pool component, not in the state.  This method needs to be a method on the shielded pool component, because that's the only thing that has access to the compact block.

hdevalence: If we want the staking component to trigger changes to the compact block, we need to use the state to do message-passing to the shielded component, which owns it.

component/src/shielded_pool/component.rs
```diff
@@ -401,6 +474,115 @@ impl ShieldedPool {
...
+        Epoch::from_height(height, epoch_duration)
+    }
+
+    /// Get the epoch duration.
+    async fn epoch_duration(&self) -> u64 {
+        self.state
+            .get_epoch_duration()
+            .await
+            .expect("can get epoch duration")
+    }
```
hdevalence, line 496: The epoch isn't a property of the shielded pool, so it shouldn't be in the shielded pool's `View` trait. A better place would be the common `penumbra_chain::View` trait.

component/src/shielded_pool/component.rs
hdevalence: Cool, I remembered we'd talked about that, but I wasn't sure where the check is.

Not blocking on this PR, but I'm a little worried about how to manage the complexity of having different checks in different components, since it's hard to refer to "lines 846 to 848 in blahblah.rs".

In the IBC component, we did a refactor pass where we put each logical check into a separate function, so that the verification function reads as a list of explicit, named checks.

We don't do this in the SP or staknig component, for historical reasons. But if we refactored the code that way, we'd be able to put a comment in referring to a specific check by name, like
```rust
// We know there can only be one validator identity, because of
// stake::checks::undelegation_only_from_one_validator()
```

However, that's a bunch of refactoring (albeit mechanical and easy), so we should do it later.

component/src/shielded_pool/state_key.rs
plaidfinch: Resolved: now there is just one note source, from which we can delete things.

chain/src/quarantined.rs
```diff
@@ -0,0 +1,371 @@
...
+                        identity_key: Some(identity_key.into()),
+                        unbonding: if unbonding.is_empty() {
+                            None
+                        } else {
+                            Some(unbonding.into())
+                        },
+                    },
+                )
+                .collect(),
+        }
```
hdevalence, line 257: It'd be nice to have a comment here that highlights the canonical serialization issue and how the code addresses it: we know that the items will always be written in order, because they're stored in a BTreeMap, which iterates in order.

However, this means that the `Ord` impl on the relevant types is consensus-critical, and that could be good to note..

redshiftzero (approved):
I realize full testing here can't be done until pcli changes are complete, but the changes here LGTM

PR #969, opened 2022-06-07
pcli: add query support

### LINKED ISSUE #805: Add basic `pcli q` interface for state queries

Following #803, we should add a basic command-line query interface.

- [ ] Add a `pcli q` subcommand
- [ ] Add a `pcli q key KEY` subsubcommand, which takes a string `KEY` as input, hashes `KEY` to a `KeyHash`, queries the state for that key, and prints the returned bytes in hex format.

This is the minimum viable state query interface.

### LINKED ISSUE #806: Add typed state query methods to `pcli q`

hdevalence:
Closes #805
Closes #806

One difficulty in using the existing state query API is that the `state_keys`
framework we have generates `KeyHash`es, but the RPC expects the full keys.  We
can't change the RPC to only take key hashes, because ICS23 proofs require that
we have the full key to put in the proof.

There are probably two ways out:

1. Change the RPC to accept either a key hash or a key (what this commit does);
2. Change the `state_key` methods to return `String`s.

The downside of (2) is that it means that every use of a `state_key::foo()`
method will need to be replaced by `state_key::foo().into()`. But maybe this is
more coherent overall.


pd/src/info/specific.rs
```diff
@@ -98,15 +98,23 @@ impl SpecificQuery for Info {
...
+        let request = request.into_inner();
+        tracing::debug!(?request);
+
+        if request.proof {
+            if request.key.is_empty() {
+                return Err(Status::invalid_argument("key is empty"));
+            }
+            if !request.key_hash.is_empty() {
+                return Err(Status::invalid_argument(
+                    "key_hash is nonempty but proof was requested",
```
avahowell, line 110: why do we require key_hash when a proof is requested? couldn't we just derive it from key?

hdevalence: we require that `key_hash` is unset, because we can only generate proofs with a `key`, and we don't want to accept both a key and a hash. definitely ugly

hdevalence:
I think that probably it'd be cleaner to only use keys, not key hashes, but let's merge this now and keep iterating.

PR #987, opened 2022-06-08
pcli: Add await change functionality to transaction submission

### LINKED ISSUE #938: Use `AwaitChange` to wait for transaction confirmation in `pcli`

Currently, `pcli` doesn't wait for transactions to be confirmed after sending them. This can cause problems when `pcli` is used twice in quick succession, because it might accidentally try to spend the same notes twice.

We can use the change notification system in #935 to avoid this problem.  If we know we should detect an output of a transaction when it is committed (because we sent it to ourselves), we can use it to wait for transaction confirmation.

Transaction submission happens via
```rust
pub async fn submit_transaction(&self, transaction: &Transaction) -> Result<(), anyhow::Error>
```
Changing this to
```rust
pub async fn submit_transaction(&self, transaction: &Transaction, await_change: Option<note::Commitment>) -> Result<(), anyhow::Error>
```
would allow us to wait for confirmation, by making a view service `AwaitChange` RPC call after we submit the transaction and waiting for it to complete (or time out and error, say 30 seconds later).

We have to pass the `note::Commitment` separately from the `Transaction`, because at this point, the transaction has been created, and it doesn't reveal anything about which outputs are ours.

We only need to specify one change note per submission, even if there are multiple self-addressed outputs, since as soon as we detect one, we will have detected them all.

Tendermint provides an RPC that waits for confirmation.  However, it also says not to use it in production, and to use the websocket event notification system instead.  That system is now being deprecated in favor of a polling-based API, which isn't what we want here -- we want to be notified as soon as the transaction is seen.

Waiting on something from the view service is better than waiting on something from Tendermint anyways, since that way, we know that once we're notified, the data will be in our view, which we don't get from Tendermint -- we might be notified before the view service has a chance to scan, and then we'd get confused about why it's not there.

aubrika:
Closes #938

### DIFF
--- a/pcli/src/network.rs
+++ b/pcli/src/network.rs
@@ -9,7 +9,7 @@ use penumbra_proto::{
     Protobuf,
 };
 use penumbra_transaction::{plan::TransactionPlan, Transaction};
-use penumbra_view::{NoteRecord, ViewClient};
+use penumbra_view::ViewClient;
 use rand::Rng;
 use rand_core::OsRng;
 use std::future::Future;
--- a/pcli/src/network.rs
+++ b/pcli/src/network.rs
@@ -1,5 +1,6 @@
 use anyhow::{Context as _, Result};
 use penumbra_component::Context;
+use penumbra_crypto::note;
 use penumbra_proto::{
     client::{
         oblivious::oblivious_query_client::ObliviousQueryClient,
@@ -8,6 +9,7 @@ use penumbra_proto::{
     Protobuf,
 };
 use penumbra_transaction::{plan::TransactionPlan, Transaction};
+use penumbra_view::{NoteRecord, ViewClient};
 use rand::Rng;
 use rand_core::OsRng;
 use std::future::Future;
@@ -21,8 +23,14 @@ impl App {
         &mut self,
         plan: TransactionPlan,
     ) -> anyhow::Result<()> {
+        let self_addressed_output = plan
+            .output_plans()
+            .find(|output| output.is_viewed_by(self.fvk.incoming()))
+            .map(|output| output.output_note().commit());
+
         let tx = self.build_transaction(plan).await?;
-        self.submit_transaction(&tx).await
+
+        self.submit_transaction(&tx, self_addressed_output).await
     }
 
     fn build_transaction<'a>(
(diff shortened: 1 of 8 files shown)


pcli/src/network.rs
```diff
@@ -55,6 +63,20 @@ impl Opt {
 
         if code == 0 {
             println!("transaction submitted successfully");
+
+            if let Some(note_commitment) = await_change {
+                let mut view_client = self.view_client(fvk, self.data_path.as_ref()).await?;
```
hdevalence, line 68: We don't want to do this, because with the current `view_client` implementation, this will potentially spin up a new view service for every submitted transaction. (If `pcli` is using an external `pviewd`, it will just create a new connection, but even that is suboptimal relative to reusing it).

pcli/src/network.rs
```diff
@@ -55,6 +63,20 @@ impl Opt {
...
+                tracing::debug!("awaiting change note");
+                view_client
+                    .note_by_commitment(NoteByCommitmentRequest {
+                        fvk_hash: Some(fvk.hash().into()),
+                        note_commitment: Some(note_commitment.into()),
+                        await_detection: true,
+                    })
```
hdevalence, line 75: This forces the client to manually construct the proto type; the `ViewClient` should handle conversions from domain types.

hdevalence:
Rebased these changes on top of the `pcli` cleanup in #1003

hdevalence:
This doesn't seem to work, as yet:
```
Scanning blocks from last sync height 923 to latest height 936
[0s] ██████████████████████████████████████████████████      13/13      769/s ETA: 0s
pre-checking transaction...
broadcasting transaction...
transaction submitted successfully, waiting for finality...
Error: error while waiting for detection of submitted transaction

Caused by:
    status: Unknown, message: "unknown error getting requested note record", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc"} }
```

There should at least be a better error message...

hdevalence:
```
pre-checking transaction...
broadcasting transaction...
transaction submitted successfully, waiting for finality...
Error: error while waiting for detection of submitted transaction

Caused by:
    status: Internal, message: "error: error returned from database: unrecognized token: \":\"", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc"} }
```

hdevalence:
Looks like that failure was caused by writing the SQLite query with a Rust value, rather than correctly hex-encoding it: https://github.com/penumbra-zone/penumbra/pull/987/commits/fdbf561777307d6673a8b2c2ac315f8ed47dcf50

hdevalence:
Looks like this works now:
```
Scanning blocks from last sync height 1076 to latest height 1106
[0s] ██████████████████████████████████████████████████      30/30      394/s ETA: 0s
pre-checking transaction...
broadcasting transaction...
transaction submitted successfully, waiting for finality...
successfully detected transaction outputs on-chain
```

PR #988, opened 2022-06-09
pcli: Erase gRPC service types

### LINKED ISSUE #986: Erase types in internal `pcli` services

`pcli` now uses a "microservices in one process" architecture: internally, the code uses a view service and a custody service, and these services could be either in process or out of process.  (Currently, this only works with an external view service, but only because we don't have a standalone custody daemon).

The way this works is as follows:

- The services are defined in `proto` files, and compiled with Tonic, which generates for each `FooProtocol`:
  - a `FooProtocol` trait that can be used to implement the server;
  - a `FooProtocolServer` that wraps a type implementing `FooProtocol`;
  - a `FooProtocolClient` that wraps a `tower::Service` with various bounds implying that it's a gRPC service.
- To use a `FooProtocolClient`, one can either:
  1. Use `connect`, which creates a `FooProtocolClient<Channel>`, treating the transport channel to the remote service as the `Service` impl;
  2. Construct the client from a `Service` impl directly.

To use an in-process service, we construct the `FooProtocolServer`, and stick it inside the `FooProtocolClient`.  To use an out-of-process service, we use `connect`.

The problem, though, is that we then end up with client instances with different types.  Currently we work around this in an awful way, by passing to a `main_inner` that has some other type bounds. But this is only manageable because we only have one branch determining the type (remote or local view service); if we were to have more options, we'd need to have combinatorially many calls to `main_inner`, which is unworkable.  Also, this entire code structure is a mess, and we need to clean it up for other reasons.

Ideally, we'd like to be able to store the view and custody clients in a struct, with erased types.  We'd like to be able to access the `FooProtocolClient` methods, which means we want to store a `FooProtocolClient<Something>`.  That `Something` is likely to be a `tower` `BoxService`.

However, my original attempts to do this didn't work, because I had a lot of trouble getting the service request and response types to line up.  There's probably some amount of service combinators that would have to be applied to unify the request and response types.

In `pcli`'s `main.rs`, we should be able to write something like
```rust
let view = if let Some(view_address) = opt.view_address {
    /* something involving .connect() */
} else {
    /* local instance */
};
```
showing that the two clients can be unified into the same type.  This unlocks the ability to refactor the `pcli` code (later) into something where all the data (custody and view clients, fvk, etc) are fields of a struct, rather than just passed around as in the current mess.

hawkw:
`pcli` may use either an in-process view/custody service, or be
configured to talk to an out-of-process view or custody daemon. This
means that the types of the `tower::Service` given to the `tonic` client
for these services may differ depending on how `pcli` is configured.
Currently, this is handled by splitting out half of the `main` function
into a separate function and making it generic, which is kind of
unfortunate.

This branch adds a `box_grpc_svc` module in `pcli`, with utilities for
type-erasing a local or remote gRPC service. It changes the view service
constructed in `main` to use these utilities and removes the (now
unnecessary) `main_inner`. The `box_grpc_service` module could also be
used to do the same for a custody service, once a standalone custody
daemon is implemented.

I also went ahead and factored out the construction of the view service
into a method on `Opts`, which I thought seemed a bit nicer.

Closes #986

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2604,6 +2604,7 @@ dependencies = [
  "fslock",
  "futures",
  "hex",
+ "http-body",
  "indicatif",
  "jmt",
  "pd",
--- a/pcli/Cargo.toml
+++ b/pcli/Cargo.toml
@@ -64,6 +64,7 @@ rand = "0.8"
 rand_chacha = "0.3.1"
 rand_core = { version = "0.6.3", features = ["getrandom"] }
 indicatif = "0.16"
+http-body = "0.4.5"
 
 [build-dependencies]
 vergen = "5"
--- a/pcli/src/box_grpc_svc.rs
+++ b/pcli/src/box_grpc_svc.rs
@@ -1,0 +1,53 @@
+use bytes::Bytes;
+use http_body::{combinators::UnsyncBoxBody, Body};
+use tonic::{
+    body::BoxBody as ReqBody,
+    codegen::http as grpc,
+    transport::{self, Endpoint},
+};
+use tower::{util::UnsyncBoxService, Service, ServiceBuilder};
+
+/// A type-erased gRPC service.
+pub(crate) type BoxGrpcService =
+    UnsyncBoxService<grpc::Request<ReqBody>, grpc::Response<RspBody>, BoxError>;
+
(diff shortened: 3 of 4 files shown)


hdevalence:
Awesome, thanks!

Issue #688, opened 2022-04-24
Unify smoke test and deploy workflows


hdevalence:
**Is your feature request related to a problem? Please describe.**

Currently we have two different workflows: one that does a smoke test (runs an ephemeral testnet, checks that it didn't fall over), and another that does a deployment to `testnet-preview.penumbra.zone`.  These have basically overlapping functionality, since in each cases it's a fresh new testnet triggered by CI.

Also, it's not clear if the smoke test actually works, because we don't properly exit `pd` on error (cf #684), so the fact that it's still running might not be enough (not sure whether the smoke test does something more than this).

**Describe the solution you'd like**

We should figure out how to unify these workflows.

This could also be a good opportunity to script a sequence of actions that we think should work, and execute those.  For instance, we could:

- hardcode a seed phrase in the CI config (it's fine for this to be public, since it's only used for testing, but I guess we could also pass it in as an env secret or something, as long as we record it somewhere persistent like 1Password);
- add the corresponding address to the allocations, so it will have funds in every testnet;
- initialize a `pcli` wallet with the seed phrase;
- perform a sequence of actions that should all succeed, like:
  - send funds to itself
  - delegate to a validator
  - undelegate from a validator

PR #991, opened 2022-06-09
wip: integration testing

redshiftzero:
wip towards the integration test parts of #688


hdevalence:
Two comments:

1. Rather than having an environment variable, I think it would be better if we just put it directly into the source code as a constant. I realize that publishing a seed phrase in a git repo is a Worst Practice, but in our case, I think it's actually a good idea. Because we reuse whatever the current testnet allocations are for local devnets, it means that any time someone makes a local devnet, they can immediately run the integration tests, using the hardcoded seed phrase, which will get an allocation in their local devnet, with no extra steps.  In effect, this is declaring a common "dev account" shared with the entire world.  Sure, once a testnet is live, the tokens it has can be taken by anyone -- but that's not a big deal, it's just a spicy faucet.

2. Since the test actions are stateful (both on the `pcli` side as well as on the `pd` side), we might want them to be sequenced? Not sure what the best model for that is.

plaidfinch:
> Since the test actions are stateful (both on the pcli side as well as on the pd side), we might want them to be sequenced? Not sure what the best model for that is.

@hdevalence My first thought was a global static `FairMutex<usize>` in the test suite, used like below (I have only written this code, not compiled or run it):

```rust
use std::sync::Arc;
use parking_lot::FairMutex;

pub static TEST_NUMBER: Arc<FairMutex<usize>> = Arc::new(FairMutex::new(0));

pub fn test_number<T: Send + 'static>(n: usize, run_test: impl FnOnce() -> T + Send + 'static) -> T {
    std::thread::spawn(|| {
        let guard = loop {
            let guard = TEST_NUMBER.lock().unwrap();
            if *guard == n {
                break guard;
            }
        };
        let result = run_test();
        *guard += 1;
        result
    }).unwrap()
}
```

The reason to use a fair mutex is to prevent starvation when everyone is looping.

### Unfortunately...

After further thought, though, this relies on the assumption that the test runner will start every test concurrently, and I don't think that's true. You'd have to collapse all the tests into a single `#[test]` function, at which point the whole mutex dance is unnecessary because you could just... run them in order. So I'm just documenting this because I thought about it, not because we should do it.

hdevalence:
I think I'm missing something — why do we need a mutex at all? Cant we just write a bunch of functions and call them in sequence from one #[test] function?

plaidfinch:
> why do we need a mutex at all? Cant we just write a bunch of functions and call them in sequence from one #[test] function?

I think we should just do that, yeah :)

pcli/tests/test_integration.rs
```diff
@@ -0,0 +1,118 @@
...
+}
+
+#[ignore]
+#[test]
+fn transaction_send_from_addr_0_to_addr_1() {
+    let tmpdir = load_wallet_into_tmpdir();
+
+    // Send to self: tokens were distributed to `TEST_ADDRESS_0`, in our test
+    // we'll send `TEST_ASSET` to `TEST_ADDRESS_1` and then check our balance.
+    let server_host = option_env!("PENUMBRA_NODE").unwrap_or("testnet-preview.penumbra.zone");
```
hdevalence, line 61: Now that #992 is merged, we could skip setting the node, and rely on `pcli` to pick up the env variable.

pcli/tests/test_integration.rs
```diff
@@ -0,0 +1,118 @@
...
+//!
+//! These tests are marked with `#[ignore]`, but can be run with:
+//! `cargo test --package pcli -- --ignored`
+//!
+//! Tests against `testnet-preview.penumbra.zone` by default, override with
+//! environmental variable `PENUMBRA_NODE`.
+//!
+//! Tests assume that the initial state of the test account is after genesis,
+//! where no tokens have been delegated, and the address with index 0
+//! was distributed 1cube.
```
hdevalence, line 11: I think we'll need to allocate penumbra tokens to the test account, because otherwise we have no way to test (un)delegations...

.github/workflows/smoke.yml
```diff
@@ -21,12 +19,15 @@ jobs:
...
-          TESTNET_RUNTIME: 4m
+          TESTNET_RUNTIME: 5m
+
+      - name: Now start testnet again in the background so we can run integration tests.
+        run: docker-compose up --detach pd-node0
+
+      - name: Run integration tests against localhost.
+        run: cargo test --package pcli -- --ignored
+        env:
+          PENUMBRA_NODE: 192.167.10.10
```
hdevalence, line 33: Rather than restarting the smoke-test testnet and running the integration tests against it, could we just add the `cargo test` invocation as a post deployment step on the testnet deployment actions?

That way, we'd ensure that the deployment actions fail if the test fails, and we don't need a separate action run.

hdevalence: Oh, NVM, I see that this is changed to run on _every_ push, that seems better, and obsoletes my comment!

 I think it would still be good to add to the deployment actions, though.

hdevalence:
Looks like the smoke test is failing because it can't connect to the Tendermint node. `pcli` assumes that `pd` and `tendermint` are on the same host (we should keep this assumption). I tried setting it to 127.0.0.1 but no luck.

redshiftzero:
This is passing now so gonna merge this, but if running the smoke test on PRs is annoying (it's already marked as optional) then we can revise the frequency and go back to just running on pushes to `main`

PR #992, opened 2022-06-09
pcli: Read node and pviewd addresses from env vars 

hawkw:
Currently, users can configure `pcli` with the hostname of a Penumbra
node, the `pd` and Tendermint ports on that node, and the (optional)
address of a `pviewd` instance. However, these configurations must
currently be passed as CLI arguments to *every* `pcli` invocation. This
is a bit annoying: if I'm running `pviewd` on my machine, I'd probably
want to be able to configure `pcli` to always use my `pviewd` instance,
without having to pass the argument every time.

This commit adds support for reading the node and pviewd addresses from
environment variables. This way, if a user always wants to use a
particular `pviewd` instance (for example), they can add
```bash
export PENUMBRA_VIEW_ADDRESS="localhost:42069"
```
or something to their `.bashrc` or other shell configuration files.

In the future, we'd probably want to consider a config file as well, but
env variables are very easy to add support for.

Also, I added the `tracing` filter to the `clap` arguments as well, because
it seemed cute to do.

Depends on https://github.com/penumbra-zone/penumbra/pull/989.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -127,7 +127,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "db02d390bf6643fb404d3d22d31aee1c4bc4459600aef9113833d17e786c6e44"
 dependencies = [
  "quote 1.0.18",
- "syn 1.0.95",
+ "syn 1.0.96",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


hdevalence:
I'm happy to merge this if other people think it would be good -- personally, I don't like environment variables and would rather have to type everything out every time, but I also recognize that other people don't need to share my personal preferences.

PR #998, opened 2022-06-10
view: add support for tracking quarantined notes to view storage

### LINKED ISSUE #993: Extend View functionality to support tracking quarantine status of notes
Currently view service does not track quarantine status and we need to do that

View service/client should support `include_quarantined` for notes queries

**Additional context**
We will need to extend the NoteRecord to include more information from the CompactBlock relevant to quarantined status and update the local note storage entries appropriately on each commit to storage. The `NotesRequest` proto will need updated with `include_quarantined`. `pcli` will need to display quarantined values as a separate line item in `balance`, with a subcommand to break down the quarantined total (which could be added later outside of this scope, but could include estimates on when these notes will be available)

aubrika:
Closes #993


aubrika:
Todo:

`scan_block`:
- must also trial decrypt the quarantined_note_payloads (can zip together the note payload vecs, can abstract the trial decryption into helper function?)
- merge together quarantined and unquarantined nullifiers
- report rolled-back commitments and nullifiers as-is
- process the above on storage commit

proto/proto/view.proto
```diff
@@ -101,6 +101,8 @@ message NoteRecord {
     optional uint64 height_spent = 6;
     // The note position.
     uint64 position = 7;
+    // Note quarantine status
+    bool quarantined = 8;
```
hdevalence, line 105: This should record the unbonding epoch. Otherwise, there's no way for a client to know when the quarantined notes will be released.

hdevalence:
As discussed on Discord, we want to go in a different direction with the data modeling. The current `NoteRecord` represents a note that's been accepted by the chain, but quarantined notes haven't been.  The changes in this PR expand the `NoteRecord` to be either a note that's been accepted by the chain, or one that hasn't been, which means that all consumers have to handle both cases, even though there's no use case where someone wants to handle both.  It also makes the data organization somewhat unwieldy, because we now have a complex, inter-linked set of nullability requirements in the `notes` table:
```
    -- the nullifier for this note, used to detect when it is spent
    nullifier               BLOB, -- null iff quarantined
    -- the position of the note in the note commitment tree
    position                BIGINT, -- null iff quarantined
    -- the quarantine status of the note
    unbonding_epoch         BIGINT, -- not null iff quarantined
    identity_key            BLOB -- not null if quarantined, null if not quarantined and not spent
);
```
This is a sign that we're shoving two different kinds of data in one table, instead of just using two different tables for two different kinds of data.

Most of this PR is changes to support this data modeling, so I think the best option would be to start over in a different direction, using it as a reference as needed.

plaidfinch:
I think some of the changes here should be preserved in #1017, rather than dropped.

PR #1000, opened 2022-06-10
add IBCToken for representing IBC assets

### LINKED ISSUE #980: Create an IBC denomination type.
Following the example of the `DelegationToken`, we should add a refinement type for a `Denom` that holds an IBC asset. Internally, this would store the `Denom`, but it would also have methods to get the IBC path, or other information.

avahowell:
Closes #980 .


hdevalence:
I think we should probably avoid creating a new `ibc` crate and keep this with the other IBC code in the `components` crate, so that we don't go back to the model we had where there was an artificial distinction between "the X crate" and "the X component" (in another crate).

avahowell:
makes sense. i'll update this to do so

component/src/ibc/ibc_token.rs
```diff
@@ -0,0 +1,57 @@
...
+
+/// <https://github.com/cosmos/ibc-go/blob/main/docs/architecture/adr-001-coin-source-tracing.md>
+fn derive_ibc_denom(transfer_path: &str) -> String {
+    let mut hasher = Sha256::new();
+    hasher.update(transfer_path.as_bytes());
+
+    let denom_bytes = hasher.finalize();
+    let denom_hex = String::from_utf8(hex::encode_upper(denom_bytes).into()).unwrap();
+
+    format!("ibc/{}", denom_hex)
```
hdevalence, line 19: Do we want to use the `ibc/HASH` as the base denom, or the full transfer path?

Our asset IDs are already hashes of the full transfer path, so I'm wondering if our version of the "ibc denom" is already our asset ID.

If we store the transfer path as the denom, we can also have methods on the IBC denom that would do things like parse it to extract the chain of bearer assets, etc.

avahowell: This design follows the "coin source tracing" ADR as proposed in ibc-go:

https://github.com/cosmos/ibc-go/blob/main/docs/architecture/adr-001-coin-source-tracing.md

afaik, the two main problems that caused this designed to be reached were SDK-specific issues: without hashing, the denom can grow indefinitely as transfers are made across zones, which is a problem in the cosmos SDK due to the Coin type only supporting denoms of up to 64 character len, and the SDK Coin type applies some regex to the denom. 

Given that our asset IDs already hash the full transfer path, I think it makes sense to just have our IBC denom be the asset ID.

avahowell: ah, come to think of it, i think this approach would cause issues for our counterparties (SDK chains), for the reasons noted in that ADR: if we were to have our IBC denom be the penumbra asset ID (e.g., `ibc/passet194...`), then the denoms would be too large for the SDK.Coin type.

avahowell:
Updated this to use the full transfer path as the base denom. We don't need to follow the design in ADR-001, and can instead reuse our asset ID hashing scheme to uniquely address ibc transfer denoms.

component/src/ibc/ibc_token.rs
```diff
@@ -0,0 +1,58 @@
+use ibc::core::ics24_host::identifier::{ChannelId, PortId};
+use penumbra_crypto::asset;
+
+/// IBC token respresents a token that was created through IBC.
+pub struct IBCToken {
```
hdevalence, line 5: This should be at least `#[derive(Clone, Debug)]` and also probably `PartialEq`, `Eq`?

component/src/ibc/ibc_token.rs
```diff
@@ -0,0 +1,58 @@
...
+    transfer_path: String,
```
hdevalence, line 6: Along the lines of [parse, don't validate](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/), maybe instead of storing the encoded `transfer_path`, we should store
```
channel_id: ChannelId,
port_id: PortId,
origin_denom: String,
```

so that we can access the channel and port IDs without re-parsing?

component/src/ibc/mod.rs
```diff
@@ -7,6 +7,10 @@ mod client;
 mod component;
 mod connection;
 pub(crate) mod event;
+
+#[allow(dead_code)]
+pub(crate) mod ibc_token;
+
```
hdevalence, line 13: rather than having to do `allow(dead_code)`, if we did
```
mod ibc_token;
pub use ibc_token::IBCToken;
```
we'd make the type accessible and avoid the warning.

PR #1001, opened 2022-06-10
fix: break once the sync to latest_known_block_height is complete

### LINKED ISSUE #932: pcli error on first sync
**Describe the bug**
During the first sync of pcli,  the sync overshoots the expected block "denominator" and doesn't return the requested command, instead continuing the sync 
https://discord.com/channels/824484045370818580/962034051991097424/981258758535659520 

Kiling the sync and then re-running whatever command, correctly returns expected output. 

**To Reproduce**
1. Install pcli
2. Run a command that requires a sync for the first time e.g. `pcli balance`
3. See sync error (sync doesn't finalize)
4. Kill sync (control C)
5. Re-run `pcli balance`

**Expected behavior**
No hanging sync :) return expected output right when sync completes

**Screenshots**
Screen Shot 2022-05-31 at 4 56 44 PM

**Additional context**
N/A

redshiftzero:
Closes #932

I've turned down the error level in `view/src/worker.rs:216` to warn because when the `break` is hit here it was causing a spurious error reported to the user (i.e. when running without RUST_LOG)

This was tested against testnet-preview

### DIFF
--- a/view/src/service.rs
+++ b/view/src/service.rs
@@ -294,7 +294,7 @@ impl ViewProtocol for ViewService {
                     latest_known_block_height,
                     sync_height,
                 };
-                if sync_height == latest_known_block_height {
+                if sync_height >= latest_known_block_height {
                     break;
                 }
             }
--- a/view/src/worker.rs
+++ b/view/src/worker.rs
@@ -140,6 +140,7 @@ impl Worker {
         while let Some(block) = stream.message().await? {
             let block = CompactBlock::try_from(block)?;
             let height = block.height;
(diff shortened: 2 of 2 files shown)


view/src/service.rs
```diff
@@ -294,7 +294,7 @@ impl ViewProtocol for ViewService {
                     latest_known_block_height,
                     sync_height,
                 };
-                if sync_height == latest_known_block_height {
```
hdevalence, line 297: Do we know why this overran previously? Shouldn't we be receiving sequential heights?

Issue #1011, opened 2022-06-14
`pcli`: display unbonding notes separately with happy-path unbonding time


plaidfinch:
Blocked on #1009.

PR #1017, opened 2022-06-16
View quarantine v2

### LINKED ISSUE #1009: `pviewd`: data model changes, changes to `scan_block` when `Quarantined` changes to the `CompactBlock` occur
Blocked on #1008.

Subsumes #993.

aubrika:
Closes #1009 and #1011.

TODO:

- [x] Make `QuarantinedNoteRecord` proto type and domain type
- [x] Update scanning & sync logic to handle quarantining
- [x] Add RPC to view service that works like existing note RPC, but for quarantined notes
- [x] Use RPC to display quarantined notes in `pcli` (#1011)

### DIFF
--- a/pcli/src/command/balance.rs
+++ b/pcli/src/command/balance.rs
@@ -1,3 +1,5 @@
+use std::collections::BTreeMap;
+
 use anyhow::Result;
 use comfy_table::{presets, Table};
 use penumbra_crypto::{keys::DiversifierIndex, FullViewingKey, Value};
@@ -29,8 +31,12 @@ impl BalanceCmd {
 
         if self.by_address {
             let notes = view.unspent_notes_by_address_and_asset(fvk.hash()).await?;
+            let quarantined_notes = view
+                .quarantined_notes_by_address_and_asset(fvk.hash())
+                .await?;
 
-            let rows: Vec<(DiversifierIndex, Value)> = if self.by_note {
+            // `Option<u64>` indicates the unbonding epoch, if any, for a quarantined note
+            let rows: Vec<(DiversifierIndex, Value, Option<u64>)> = if self.by_note {
                 notes
                     .iter()
                     .flat_map(|(index, notes_by_asset)| {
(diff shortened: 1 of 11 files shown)


plaidfinch:
I think this should incorporate changes from https://github.com/penumbra-zone/penumbra/pull/998, rather than dropping all of that. In particular, some of the refactors related to how trial decryption happens in `scan_block`.

On another note, I also think we can clean up the database schema by having a single table for all the things all notes share, then joining with separate tables for the things only spent notes have vs. the things only quarantined notes have.

hdevalence:
> I think this should incorporate changes from https://github.com/penumbra-zone/penumbra/pull/998, rather than dropping all of that. In particular, some of the refactors related to how trial decryption happens in `scan_block`.

Yes, the idea was that the data modeling changes were sufficiently large that it would be faster and easier to just redo the changes using that branch as a reference, rather than trying to do git surgery. 

> On another note, I also think we can clean up the database schema by having a single table for all the things all notes share, then joining with separate tables for the things only spent notes have vs. the things only quarantined notes have.

We actually considered this and decided not to do it, because it makes all of the queries more complicated, since the relevant data is split across two tables instead of one, and it doesn't actually give any commonality, because there's no case where you want to query notes that are either accepted OR quarantined.

plaidfinch:
> We actually considered this and decided not to do it, because it makes all of the queries more complicated, since the relevant data is split across two tables instead of one, and it doesn't actually give any commonality, because there's no case where you want to query notes that are either accepted OR quarantined.

Aha, sounds good.

> Yes, the idea was that the data modeling changes were sufficiently large that it would be faster and easier to just redo the changes using that branch as a reference, rather than trying to do git surgery.

Yup, agreed, just want to make sure work doesn't get redone that doesn't need to be.

plaidfinch:
@hdevalence i think you forgot to push a file in [ebfab8b](https://github.com/penumbra-zone/penumbra/pull/1017/commits/ebfab8b9a17e82bd67194f50127ca27d34059ecf); looks like `quarantined_note_record.rs` is missing.

hdevalence:
> @hdevalence i think you forgot to push a file in [ebfab8b](https://github.com/penumbra-zone/penumbra/pull/1017/commits/ebfab8b9a17e82bd67194f50127ca27d34059ecf); looks like `quarantined_note_record.rs` is missing.

whoops! checking now

hdevalence:
@plaidfinch sorry about that, pushed it up as a fixup commit, feel free to rebase if you'd like!

plaidfinch:
no worries, thanks!

PR #1020, opened 2022-06-16
storage: make State a standalone type, not a typedef

hdevalence:
This opens the door to having an object cache.  The methods used to access the
state could now just as easily be methods on the `State`, rather than in the
`StateExt` trait (since the `State` is now a local type to the
`penumbra_storage` crate) -- except that then, every extension trait would no
longer be able to use any of those methods in provided implementations of the
trait methods, so it would no longer be possible to write an empty `impl`
block, and all the methods would have to be duplicated across both blocks.


hdevalence:
Actually doing the object cache is blocked on https://github.com/penumbra-zone/penumbra/issues/1021

hdevalence:
Actually doing this requires the ability to pass through `get_with_proof`, and that can't be done cleanly because of #1022.

Icing this for now.

hdevalence:
Closed in favor of a new design in #1454

PR #1023, opened 2022-06-17
Include original error messages and add more descriptive messages

### LINKED ISSUE #1019: Improve view service database error reporting
While working on pcli and galileo I've noticed that we frequently get errors printed to the log like so:

`2022-06-16T20:16:19.240877Z  WARN penumbra_view::worker: e=status: Unavailable, message: "database error", details: [], metadata: MetadataMap { header`

There's no additional debugging information and even with debug mode logging enabled all the surrounding SQL queries seem to be syntactically and semantically fine, and report returning result rows, so it's not clear what's going on.

The error handling should be made more verbose and bubble up the original error if possible.

zbuc:
Closes #1019

### DIFF
--- a/chain/src/view.rs
+++ b/chain/src/view.rs
@@ -92,7 +92,7 @@ pub trait View: StateExt {
         let chain_id = self
             .get_chain_id()
             .await
-            .map_err(|_| tonic::Status::unavailable("database error"))?;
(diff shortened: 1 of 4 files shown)


plaidfinch:
This looks great to me!

PR #1025, opened 2022-06-17
add IBC "app handlers", router for IBC events, ICS20 Transfer scaffold

avahowell:
This PR adds a scaffold that should allow us to cleanly implement ICS20 transfers, and any other future IBC applications (such as interchain accounts).

Originally, we had considered the ICS20 implementation to be through another component. However, this isn't correct, because IBC applications want to consume (validated) IBC channel events, not raw transactions.

Along the lines of https://github.com/cosmos/ibc/tree/master/spec/core/ics-026-routing-module and https://github.com/cosmos/ibc/tree/master/spec/core/ics-025-handler-interface, this PR adds an "App Handler" interface, which defines the interface that IBC applications use to consume events from the IBC core protocol, and adds an "App Router" that allows instantiations of IBC applications to bind to IBC Ports and receive events for that port. 

This PR implements the AppHandler interface in two distinct interfaces, which match our distinct "check" and "execute" approach for transaction execution.

### DIFF
--- a/component/src/ibc/component.rs
+++ b/component/src/ibc/component.rs
@@ -6,26 +6,25 @@
 mod channel;
 mod client;
 mod connection;
-mod ibc_handler;
 pub(crate) mod state_key;
-mod transfer;
 
+use crate::ibc::ibc_handler::AppRouter;
+use crate::ibc::transfer::ICS20Transfer;
 use crate::{Component, Context};
 use anyhow::Result;
 use async_trait::async_trait;
 use client::Ics2Client;
+use ibc::core::ics24_host::identifier::PortId;
 use penumbra_chain::{genesis, View as _};
 use penumbra_storage::State;
 use penumbra_transaction::Transaction;
 use tendermint::abci;
 use tracing::instrument;
 
-use self::ibc_handler::AppRouter;
-
 pub struct IBCComponent {
     client: client::Ics2Client,
     connection: connection::ConnectionComponent,
-    channel: channel::ICS4Channel<AppRouter>,
+    channel: channel::ICS4Channel,
 
     state: State,
 }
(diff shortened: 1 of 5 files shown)


component/src/ibc/ibc_handler.rs
```diff
@@ -0,0 +1,189 @@
...
+use ibc::core::ics04_channel::msgs::timeout::MsgTimeout;
+use ibc::core::ics24_host::identifier::PortId;
+use std::collections::HashMap;
+
+/// AppHandlerCheck defines the interface for an IBC application to consume IBC channel and packet
+/// events, and apply their validation logic. This validation logic is used for stateful validation
+/// only.
+#[async_trait]
+pub trait AppHandlerCheck: Send + Sync {
+    async fn chan_open_init_check(&self, msg: &MsgChannelOpenInit) -> Result<()>;
```
avahowell, line 28: should this also have a `ctx` so that we can report back IBC events from IBC applications? I think probably yes

component/src/ibc/ibc_handler.rs
```diff
@@ -0,0 +1,189 @@
...
+    }
+}
+
+#[async_trait]
+impl AppHandlerCheck for AppRouter {
+    async fn chan_open_init_check(&self, msg: &MsgChannelOpenInit) -> Result<()> {
+        if let Some(handler) = self.handlers.get(&msg.port_id) {
+            handler.chan_open_init_check(msg).await?;
+        }
+        Ok(())
```
avahowell, line 88: note that the way this is implemented currently means that if we get channel events for a port which we have no applications bound on, then we just no-op. This seems okay, but it might be worth thinking about if we should allow channels to be opened if we have no applications registered for their port

component/src/ibc/component.rs
```diff
@@ -31,7 +34,12 @@ impl IBCComponent {
     pub async fn new(state: State) -> Self {
         let client = Ics2Client::new(state.clone()).await;
         let connection = connection::ConnectionComponent::new(state.clone()).await;
-        let channel = channel::ICS4Channel::new(state.clone()).await;
+
+        let mut router = AppRouter::new();
+        let transfer = ICS20Transfer::new(state.clone());
+        router.bind(PortId::transfer(), Box::new(transfer));
```
avahowell, line 40: for future IBC apps, we would do 

`router.bind(PortId::interchain_accounts(), Box::new(interchain_acct_impl))`

hdevalence (approved):
LGTM

PR #1028, opened 2022-06-20
Use AllowList for metrics labels extracted from spans

### LINKED ISSUE #965: Use label `AllowList` for metrics labels
The library we're using for enriching metrics with labels derived from spans has an `AllowList` feature: https://docs.rs/metrics-tracing-context/latest/metrics_tracing_context/label_filter/struct.Allowlist.html

We should start using this to avoid having too many labels attached to our metrics. Specifically the `chain_id` and `worker` are most important to include right now.

zbuc:
Also renames the `worker` label as `role` and places it in an `app` span as per #965 

Closes #965

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2240,15 +2240,25 @@ dependencies = [
 ]
 
 [[package]]
-name = "metrics-exporter-prometheus"
-version = "0.8.0"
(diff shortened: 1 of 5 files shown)


hdevalence (approved):
LGTM!

PR #1031, opened 2022-06-20
pcli: don't wait for full confirmation of undelegations

hdevalence:
We won't detect change, because it's quarantined, but we also don't need to,
because it's not spendable.


plaidfinch:
lgtm!

Issue #1047, opened 2022-06-21
Define `Swap` and `SwapClaim` actions, based on the description in the protocol spec.


zbuc:
- [x] Domain type for Swap action
- [x] Domain type for SwapClaim action
- [x] Domain type for SwapPlaintext
- [x] Domain type for SwapCiphertext (should wrap a fixed-size byte array)
- [x] implementation of SwapPlaintext::encrypt(&self, ovk: &OutgoingViewingKey) -> SwapCiphertext, which should build a fixed-size binary encoding, then encrypt it
- [x] implementation of SwapCiphertext::decrypt(&self, ovk: &OutgoingViewingKey) -> Result<SwapPlaintext> that decrypts and parses the fixed-size encoding
- [x] #1126
- [x] a transparent SwapProof
- [x] a transparent SwapClaimProof

PR #1052, opened 2022-06-21
Swap mock & stub implementations

zbuc:


proto/proto/dex.proto
```diff
@@ -0,0 +1,49 @@
...
+    uint64 f = 3;
+    bytes bd = 4;
+    bytes pkd = 5;
+}
+
+message MockFlowCiphertext {
+    uint64 value = 1;
+}
+
+// Holds two asset IDs. Ordering doesn't reflect trading direction.
```
hdevalence, line 43: We should probably note that there is an ordering, though, in order to ensure that there's only one valid `TradingPair` for each pair of assets.

hdevalence:
I think it'd be helpful to pull out the mock flow ciphertexts into a separate PR that we can merge early, and to avoid adding the `Swap`, `SwapClaim` variants to `Transaction` until they're fleshed out.  That way, we can be merging changes to the `Swap`, `SwapClaim` without affecting what data people can post to the chain.

zbuc:
@hdevalence separated the MockFlowCiphertext (and proto types for Swap) into https://github.com/penumbra-zone/penumbra/pull/1055

zbuc:
@hdevalence I think this is at a good merge point now, and the rest of #1047 can be in a different PR

hdevalence:
> I'm not happy with how much I had to put in the `crypto` crate, but to add the `TradingPair` struct to `component` created a circular dependency, I had similar issues dealing with the `Fee` type in `transaction`.

I think this is probably OK. Actually implementing the swaps for real will require writing SNARK circuits that express the swap-related state transitions, and those circuit implementations should live in the `crypto` crate.  So the data types that the proofs are about could be there too.

crypto/src/lib.rs
```diff
@@ -18,6 +18,7 @@ mod note_payload;
 mod nullifier;
 mod prf;
 pub mod proofs;
+mod swap;
```
hdevalence, line 21: This could be `pub mod swap;`, without the `pub use` below, if we wanted to isolate the (more niche) swap functionality into one submodule.

crypto/src/swap.rs
```diff
@@ -0,0 +1,143 @@
...
+pub struct SwapPlaintext {
+    // Trading pair for the swap
+    pub trading_pair: TradingPair,
+    // Amount of asset 1
+    pub t1: u64,
+    // Amount of asset 2
+    pub t2: u64,
+    // Fee
+    // TODO: ideally would be transaction::Fee but this causes cyclic import issues
+    pub fee: u64,
```
hdevalence, line 21: Hmm, let's pull the `Fee` out into the `crypto` crate, then.

proto/proto/dex.proto
```diff
@@ -12,28 +12,66 @@ message Swap {
...
+  // A commitment to the fee's contribution to the transaction's value balance.
+  bytes cf = 4;
   // Swap NFT recording the user's contribution.
   crypto.NotePayload swap_nft = 5;
+  // Encrypted version of the original `Swap`, symmetrically encrypted w/ viewing key.
+  bytes swap_ciphertext = 6;
 }
 
 // XXX value fields need to have constant-length encoding
+// ? or probably not, since this is a plaintext
```
hdevalence, line 61: Yep, we can bundle the constant-length encoding into the encrypt/decrypt functions, since we'll only need to handle it there.

proto/proto/dex.proto
```diff
@@ -12,28 +12,66 @@ message Swap {
   MockFlowCiphertext enc_amount_1 = 2;
   // Encrypted amount of asset 2 of the trading pair.
   MockFlowCiphertext enc_amount_2 = 3;
+  // Encapsulates the authorized fields of the Swap transaction, used in signing.
+  SwapBody body = 4;
+  // The spend authorization signature is effecting data.
+  crypto.SpendAuthSignature auth_sig = 5;
```
hdevalence, line 18: I don't think we need a spend auth signature here, because the `Swap` action consumes the input funds from the transaction's value balance, rather than spending any notes itself.  So the `Swap` will have to be alongside a `Spend` action(s) that unlock the required input funds, and those actions will have the spend authorization signatures for each of those notes already.

proto/proto/dex.proto
```diff
@@ -12,28 +12,66 @@ message Swap {
...
+  bytes zkproof = 1;
+  // The nullifier for the Swap NFT to be consumed.
+  crypto.Nullifier nullifier = 2;
+  // The fee allows `SwapClaim` without an additional `Spend`.
+  transaction.Fee fee = 3;
+  // Note output for asset 1.
+  crypto.NotePayload output_1 = 4;
+  // Note output for asset 2.
+  crypto.NotePayload output_2 = 5;
+  // Root for the associated `Swap` action.
```
hdevalence, line 36: This should probably clarify that it's the *block* root, identifying the block the `Swap` was included in.

transaction/src/action/swap.rs
```diff
@@ -0,0 +1,105 @@
...
+use penumbra_crypto::NotePayload;
+use penumbra_crypto::{proofs::transparent::SpendProof, MockFlowCiphertext};
+use penumbra_crypto::{value, SwapCiphertext};
+use penumbra_proto::dex::TradingPair;
+use penumbra_proto::{dex as pb, Protobuf};
+
+#[derive(Clone, Debug)]
+pub struct Swap {
+    // A proof that this is a valid state change.
+    pub proof: SpendProof,
```
hdevalence, line 13: We shouldn't use a `SpendProof` here, because the swap proof statement is different.  Instead we should make a new `SwapProof` (it can be a placeholder at the moment), and fill in the swap-specific proof statements there later.

transaction/src/transaction.rs
```diff
@@ -78,6 +78,7 @@ impl Transaction {
         })
     }
 
+    // TODO: make sure payloads from Swap actions included
```
hdevalence, line 81: just checking, this is blocked on including the actions in the `Transaction` enum right?

zbuc: Yes

hdevalence (approved):
This looks good.  It'd be nice to clean up the warnings before merging, and also to have a `SwapProof` rather than a `SpendProof`, but those could also be done in later PRs.

PR #1054, opened 2022-06-22
pd cli updates: change --rocks-path to --home, add new testnet subcommand

### LINKED ISSUE #1045: Change `pd` command-line arguments

Following https://github.com/penumbra-zone/penumbra/pull/1043#issuecomment-1161973533, we should change the `pd` command-line arguments, and also adjust the tooling and scripts that use them.

- [x] Add a `--home` parameter to `pd start`, replacing the previous rocksdb path.  Then set the rocksdb path to `home/rocksdb` internally.  As a bonus, add a `PD_HOME` env var and a default that sets it to `$HOME/.penumbra/testnet_data/node0/`.
- [x] Split out a new `testnet` subcommand, with:
  - [x] a `testnet_dir` parameter, replacing the `output_dir` parameters on `generate-testnet`;
  - [x] `generate` subcommand, with the existing `generate-testnet` code;
  - [x] `join` subcommand, with the existing `join-testnet` code;
  - [x] `unsafe-reset-all`, which deletes the contents of the `testnet_dir`.
- [x] Update the existing deployment configs (`rg generate-testnet`, `rg start` would be helpful) to the new commands.

We could wait to clean up the command interface, but that just pushes the pain out to the future...

avahowell:
closes #1045

### DIFF
--- a/.github/workflows/deploy-preview.yml
+++ b/.github/workflows/deploy-preview.yml
@@ -78,7 +78,7 @@ jobs:
         envs: TESTNET_HOST
         script: |
           cd ~/penumbra
-          /root/.cargo/bin/cargo run --release --bin pd -- generate-testnet
+          /root/.cargo/bin/cargo run --release --bin pd -- testnet generate
           chmod -R 777 ~/.penumbra/testnet_data
     - name: Restart deployment
       uses: appleboy/ssh-action@master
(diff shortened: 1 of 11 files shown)


hdevalence (approved):
LGTM! I'll do another docs pass on the guide now that this is in place.

PR #1055, opened 2022-06-22
Add MockFlowCiphertext

### LINKED ISSUE #1033: Mock flow encryption
In advance of using a real implementation with ABCI++, it'd be helpful to have a mock implementation, along the lines of the existing transparent proofs.  This will be useful for swaps, for governance votes, and also for retrofitting into (un)delegations.

The important thing to mock here is that the values are in ciphertexts, so that we don't accidentally use plaintext values; modeling the DKG flow etc, or how clients will get public keys, is less important.

To start, we should add a `MockFlowCiphertext` proto and domain type.  The domain type should be a newtype wrapper for a `u64` that has a `mock_decrypt(&self) -> u64` operation to get the inner value, and an `Add` impl that does addition.

zbuc:
Closes #1033


crypto/src/flow.rs
```diff
@@ -0,0 +1,33 @@
...
+
+// Fake implementation for now, TODO: replace w/ additively homomorphic encryption impl
+impl MockFlowCiphertext {
+    pub fn mock_decrypt(&self) -> u64 {
+        self.0
+    }
+
+    pub fn add(&mut self, amount: u64) {
+        self.0 += amount
+    }
```
hdevalence, line 17: It'd be more ergonomic to have this be an `impl std::ops::Add`, but not essential to do right now

hdevalence (approved):

PR #1058, opened 2022-06-23
add `eddy` crate: flow encryption implementation scaffold

avahowell:
Co-Authored-By: Henry de Valence <hdevalence@penumbra.zone>


eddy/src/decryption_share.rs
```diff
@@ -2,18 +2,12 @@ use rand_core::{CryptoRng, RngCore};
...
-#[derive(Debug, Clone)]
-pub enum Verified {}
-#[derive(Debug, Clone)]
-pub enum Unverified {}
-
-pub trait VerificationState {}
-impl VerificationState for Verified {}
-impl VerificationState for Unverified {}
+pub struct Verified;
+pub struct Unverified;
```
hdevalence, line 6: These types don't have any connection to each other, other than both being used as type parameters for `DecryptionShare<S>`, and there's no constraint on the `DecryptionShare` type parameter that it be `Verified` or `Unverified`. This means that someone could construct a `DecryptionShare<u64>` or whatever else, and it's also a bit confusing, because looking at the `DecryptionShare` there's no real clue about what the parameter is supposed to represent.

To improve this, the `Verified` and `Unverified` structs could implement a `VerificationMarker` trait (naming: it could also be called `VerificationState`, but calling it `Marker` suggests that it's only a marker type, rather than any runtime state).  The `VerificationMarker` trait could use the [sealed trait pattern](https://rust-lang.github.io/api-guidelines/future-proofing.html#sealed-traits-protect-against-downstream-implementations-c-sealed) to prevent downstream implementations.

There's an example of this in the `decaf377-rdsa` crate: https://rustdoc.penumbra.zone/main/decaf377_rdsa/trait.Domain.html

eddy/src/ciphertext.rs
```diff
@@ -0,0 +1,52 @@
...
+            .await?
+            .ok_or_else(|| anyhow::anyhow!("could not find value in LUT"))?;
+        let value2 = table
+            .lookup(decryption2.compress().0)
+            .await?
+            .ok_or_else(|| anyhow::anyhow!("could not find value in LUT"))?;
+        let value3 = table
+            .lookup(decryption3.compress().0)
+            .await?
+            .ok_or_else(|| anyhow::anyhow!("could not find value in LUT"))?;
```
hdevalence, line 43: Mathematically, decryption can only fail if the value is not present in the lookup table, and while the point of the value integrity proofs is to guarantee that this cannot occur, someone could add too many ciphertexts at once, or similar, so it's important to model.

But in practice, the lookup table may not be entirely resident in memory, and might involve access to I/O, which is (1) async and (2) fallible.  So there are two kinds of errors: errors from the underlying storage, and a failed lookup.

The lookup trait returns a `Result<Option<u32>, anyhow::Error>`, with an outer `Result` allowing indication of an arbitrary I/O error, and the inner `Option` signaling whether or not the value was present.  This is a good API for external users to plug into, since they can put whatever I/O errors they want in the `Err` variant, and it's easy to signal missing values with `None`.

Here in decryption, though, we probably want a way to allow users to distinguish whether the error was due to underlying I/O issues, or whether a value went out of range.  If we use `anyhow!` to make a stringy error, we can't do this.

There are two options:

1. We could change the return type of decryption to `Result<Option<Value>, anyhow::Error>`, with the inner `Option` signaling whether the value was too large or not;
2. We could leave the return type as `Result<Value, anyhow::Error>`, but create a `LimbRangeError` struct, and use `thiserror` to derive an `Error` implementation for it, then do `.ok_or(LimbRangeError)?` here.

If we do the first option, all downstream users have to manually transform the inner `Option`. If we do the second option, downstream users that don't care can bubble up a generic error, and downstream users that do care can use `downcast_ref::<LimbRangeError>()` to handle it specially. 

So option (2) seems better to me.

eddy/src/transcript.rs
```diff
@@ -0,0 +1,45 @@
...
+        self.append_message(b"c1", &ciphertext.c1.compress().0);
+        self.append_message(b"c2", &ciphertext.c2.compress().0);
+    }
+    fn append_decryption_share_point(&mut self, point: &decaf377::Element) {
+        self.append_message(b"decryption-share-point", &point.compress().0);
+    }
+    fn append_blinding_commitment(&mut self, label: &'static [u8], point: &decaf377::Element) {
+        self.append_message(b"dom-sep", label);
+        self.append_message(b"blinding-commitment", &point.compress().0);
+    }
```
hdevalence, line 37: While `decaf377` gives us a prime-order group, the group still does have one small subgroup -- the identity element is a subgroup of size 1, and so I think we need to check that we're not processing a bunch of identity points.

One way we could do this is suggested in the [Merlin docs](https://merlin.cool/use/protocol.html#validation-logic) -- include validation in the transcript functions, making them return `Result`s and fail on identity points.

hdevalence:
Based on https://github.com/tendermint/tendermint/issues/9053, it seems like ABCI++ is not going to ship in its complete form (i.e., including `VoteExtension`s) in a deployable version of Tendermint before late this year at the earliest.  Unfortunately, this means that we will not be able to include flow encryption in the initial version of Penumbra, and we will have to add it in a later network upgrade some time in 2023.

As a result, we should pause this work, and just try to get it in a good enough state to merge in for now, and come back to it for v2.  I did a docs/API pass to that effect.

PR #1104, opened 2022-07-12
1047: Implement SwapPlaintext/SwapCiphertext encrypt/decrypt and transparent proofs

zbuc:
Uses the OVK as a chacha20poly1305 key to encrypt/decrypt Swap data.

NOTE: we discussed using an ephemeral key for `SwapPlaintext` encryption/decryption rather than the `ovk` directly. This will be done in a follow-up PR.

Also has an implementation of `Swap`/`SwapClaim` transparent proofs.


crypto/src/proofs/transparent.rs
```diff
@@ -397,21 +400,19 @@ pub struct SwapProof {
     pub g_d: decaf377::Element,
     // The transmission key for the address.
     pub pk_d: ka::Public,
-    // The value of the note.
-    pub value: Value,
+    // The value of the swap NFT.
+    pub swap_nft_value: Value,
```
hdevalence, line 404: Since the amount will always be 1, rather than recording an amount/asset id pair (`Value`), we could just record
```
pub swap_nft_id: asset::Id,
```

crypto/src/proofs/transparent.rs
```diff
@@ -422,15 +423,18 @@ impl SwapProof {
     pub fn verify(
         &self,
         anchor: tct::builder::block::Root,
-        value_commitment: value::Commitment,
+        swap_nft_value_commitment: value::Commitment,
```
hdevalence, line 426: A `SwapClaim`'s effect on the transaction value balance is just to contribute the pre-committed fees, right? So if we were going to rename this from `value_commitment`, it feels like `fee_commitment` would be a better name.

Or, we could leave it as `value_commitment` to match the other actions.

crypto/src/proofs/transparent.rs
```diff
@@ -422,15 +423,18 @@ impl SwapProof {
...
     ) -> anyhow::Result<(), Error> {
-        // Note commitment integrity.
+        // Swap NFT note commitment integrity.
         let s_component_transmission_key = Fq::from_bytes(self.pk_d.0);
         if let Ok(transmission_key_s) = s_component_transmission_key {
-            let note_commitment_test =
-                note::commitment(self.note_blinding, self.value, self.g_d, transmission_key_s);
+            let note_commitment_test = note::commitment(
+                self.note_blinding,
+                self.swap_nft_value,
```
hdevalence, line 434: Here this should hardcode `1` as the amount, and just include the asset ID in the witness data: https://github.com/penumbra-zone/penumbra/pull/1104/files#r925088722

crypto/src/proofs/transparent.rs
```diff
@@ -439,23 +443,18 @@ impl SwapProof {
             return Err(Error::TransmissionKeyMismatch);
         }
 
+        // TODO: do we need to check the swap NFT Asset ID is properly constructed here?
+
```
hdevalence, line 447: Yeah, we'll need to include (most?) of the fields of the `SwapPlaintext` in the `SwapClaimProof` (and decide exactly which ones), so that we can have the proof check that the asset ID in the note that's being spent is actually a commitment to the data that's used to form the output values.

crypto/src/proofs/transparent.rs
```diff
@@ -439,23 +443,18 @@ impl SwapProof {
...
         // Merkle path integrity.
         self.note_commitment_block_proof
             .verify(anchor)
             .map_err(|_| Error::MerkleRootMismatch)?;
 
         // Value commitment integrity.
-        if self.value.commit(self.v_blinding) != value_commitment {
+        if self.swap_nft_value.commit(self.v_blinding) != swap_nft_value_commitment {
```
hdevalence, line 454: The value commitment integrity check we want to do is different from this: we're going to be consuming the value recording the swap NFT and producing new output notes entirely _internally_ to the `SwapClaim` action, so the only effect on the transaction's value balance is the fee.

We want to check that the `value_commitment` is a commitment to `f` `penumbra` tokens, where `f` is the fee amount committed to by the swap nft's asset id.

crypto/src/proofs/transparent.rs
```diff
@@ -466,83 +465,63 @@ impl SwapProof {
...
-            return Err(Error::InvalidSpendAuthRandomizer);
-        }
-
-        // Diversified address integrity.
-        let fvk = keys::FullViewingKey::from_components(self.ak, self.nk);
-        let ivk = fvk.incoming();
-        if self.pk_d != ivk.diversified_public(&self.g_d) {
-            return Err(Error::InvalidDiversifiedAddress);
-        }
+        // TODO: is there an address check that needs to occur?
```
hdevalence, line 468: Yes, not just an address check but also the checks that are part of an `Output` proof.  We need to:

- create two output notes and do all the normal output checks;
- prove that one output is $\Lambda_1$ of asset $t_1$;
- prove that the other is $\Lambda_2$ of asset $t_2$;
- prove that both notes are sent to the transmission key committed to in the swap NFT's asset ID.

crypto/src/proofs/transparent.rs
```diff
@@ -587,6 +565,217 @@ impl TryFrom<transparent_proofs::SwapProof> for SwapProof {
...
+#[derive(Clone, Debug)]
+pub struct SwapProof {
+    // The diversified base for the destination address.
+    pub g_d: decaf377::Element,
+    // The transmission key for the destination address.
+    pub pk_d: ka::Public,
+    // The value of asset 1 in the swap.
+    pub value_t1: Value,
+    // The value of asset 2 in the swap.
+    pub value_t2: Value,
```
hdevalence, line 582: Q: since the types of asset 1 and 2 are public, do we need to have `Value`s in the proof, or could we just have the integer amounts, and pass the asset IDs (or even the value generators) into the verification method as public inputs to the proof?

The reason this might matter is that we want the transparent proof to be an exact replica of the witness data for ZK proofs, so everything in the proof check has to turn into circuit constraints, and if we pass in the value generators, we could avoid having to perform the hash-to-group operation inside of the circuit.

crypto/src/proofs/transparent.rs
```diff
@@ -587,6 +565,217 @@ impl TryFrom<transparent_proofs::SwapProof> for SwapProof {
...
+    // The fee value associated with the swap.
+    // TODO: should always be of penumbra token type, maybe this should
+    // be uint64 or a hardcoded value?
+    pub value_fee: Value,
```
hdevalence, line 586: Yeah, since the asset ID is fixed, this should just be a `u64`.  I think using `_amount` is a good convention for amounts, so this could be `fee_amount: u64`.

crypto/src/proofs/transparent.rs
```diff
@@ -587,6 +565,217 @@ impl TryFrom<transparent_proofs::SwapProof> for SwapProof {
...
+    // The asset ID of the Swap NFT.
+    pub swap_nft_asset_id: asset::Id,
+    // The blinding factor used for generating the value commitment.
+    // TODO: can this be shared amongst t1/t2/fee or is this unsafe?
+    pub v_blinding: Fr,
```
hdevalence, line 591: In general (and here), we need one blinding factor per revealed data -- since we have separate commitments to t1/t2/fee, we need one blinding factor for each.

crypto/src/proofs/transparent.rs
```diff
@@ -587,6 +565,217 @@ impl TryFrom<transparent_proofs::SwapProof> for SwapProof {
...
+                    // The swap NFT is always amount 1.
+                    amount: 1,
+                    asset_id: self.swap_nft_asset_id,
+                },
+                self.g_d,
+                transmission_key_s,
+            );
+
+            if note_commitment != note_commitment_test {
+                return Err(Error::NoteCommitmentMismatch);
```
hdevalence, line 631: note: we could just as well use `anyhow` here, if it's easier -- since the transparent proofs are supposed to model ZK proofs, we don't need to do fine-grained matching on the error cases, since all we'll get is an opaque "failed to verify" error anyways.

crypto/src/proofs/transparent.rs
hdevalence: We could leave out the amount correctness statements for now, because we don't currently have a way to model the price data that should be public inputs to the proof (that's #1126), and it would be good to merge a preliminary version of this first

crypto/src/swap.rs
```diff
@@ -1,13 +1,43 @@
...
+/// The nonce used for swap encryption.
+///
+/// The nonce will always be `[0u8; 12]` which is okay since we use a new
+/// ephemeral key each time.
+pub static SWAP_ENCRYPTION_NONCE: Lazy<[u8; 12]> = Lazy::new(|| [0u8; 12]);
+
+// Can add to this/make this an enum when we add additional types of swaps.
+// TODO: is this actually something we would do? suppose it doesn't hurt to build this
+// in early.
+pub const SWAP_TYPE: u8 = 0;
```
hdevalence, line 30: My feeling is that we probably don't want to build this in (at least, in this place), and that, in keeping with the "well-oiled joint principle", it'd be better to have that flexibility at the level of the transaction `Actions`, and handle changes by defining new actions. 

If we did it just at the level of the swap data, we'd only be able to make changes subject to the constraints that the new swap types fit exactly within the existing `Swap` action, which is already quite constraining.

So I think we should drop the type parameter here and just hardcode the binary format (then migrate off it if/when we need to by making a new action).

crypto/src/swap.rs
```diff
@@ -1,13 +1,43 @@
...
+pub const SWAP_CIPHERTEXT_BYTES: usize = 169;
+// Swap plaintext byte length
+pub const SWAP_LEN_BYTES: usize = 153;
+pub const OVK_WRAPPED_LEN_BYTES: usize = 80;
+
+/// The nonce used for swap encryption.
+///
+/// The nonce will always be `[0u8; 12]` which is okay since we use a new
+/// ephemeral key each time.
+pub static SWAP_ENCRYPTION_NONCE: Lazy<[u8; 12]> = Lazy::new(|| [0u8; 12]);
```
hdevalence, line 25: Hmm, is this OK? If the ephemeral key used for symmetric crypto is the same one used for the `NotePayload`, I think we need to make sure we have distinct nonces for encrypting the note payload, the memo, the swap plaintext, etc.

crypto/src/swap.rs
```diff
@@ -1,13 +1,43 @@
...
+
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error("Swap type unsupported")]
+    SwapTypeUnsupported,
+    #[error("Swap deserialization error")]
+    SwapDeserializationError,
+    #[error("Decryption error")]
+    DecryptionError,
+}
```
hdevalence, line 40: If we don't need to use these for control flow, we could drop the enum and just use `anyhow` to make ad-hoc errors.

crypto/src/swap.rs
```diff
@@ -26,10 +56,129 @@ pub struct SwapPlaintext {
...
+        epk: &ka::Public,
+    ) -> blake2b_simd::Hash {
+        let mut kdf_params = blake2b_simd::Params::new();
+        kdf_params.hash_length(32);
+        let mut kdf = kdf_params.to_state();
+        kdf.update(&shared_secret.0);
+        kdf.update(&epk.0);
+
+        kdf.finalize()
+    }
```
hdevalence, line 77: Can we share this code with the shared secret derivation used for encrypting the note plaintext that records the swap nft?

crypto/src/swap.rs
```diff
@@ -26,10 +56,129 @@ pub struct SwapPlaintext {
...
+        let mut kdf = kdf_params.to_state();
+        kdf.update(&ovk.0);
+        // TODO: should we be using the public fields e.g. t1, t2, trading_pair here?
+        // Note implementation uses value commitments...
+        // kdf.update(&cv_bytes);
+        // kdf.update(&cm_bytes);
+        kdf.update(&epk.0);
+
+        kdf.finalize()
+    }
```
hdevalence, line 103: We don't currently have a distinct ock type, do we?

Since the goal is to have the swap plaintexts be encrypted with the same key as the note plaintext (so that any entity that can see the note recording the swap NFT can also see the swap data it commits to, and vice versa), I think it would be good to find a way to share code between note encryption and swap plaintext encryption.

I wonder whether creating a common OCK type would give us a convenient place to put that code?

hdevalence: cc @redshiftzero

crypto/src/swap.rs
```diff
@@ -26,10 +56,129 @@ pub struct SwapPlaintext {
...
+
+        // Note: Here we use the same nonce as swap encryption, however the keys are different.
+        // For swap encryption we derive a symmetric key from the shared secret and epk.
+        // However, for encrypting the outgoing cipher key, we derive a symmetric key from the
+        // sender's OVK, and the epk. Since the keys are
+        // different, it is safe to use the same nonce.
+        //
+        // References:
+        // * Section 5.4.3 of the ZCash protocol spec
+        // * Section 2.3 RFC 7539
```
hdevalence, line 130: I think we want to use the same empheral symmetric key for encrypting both the note plaintext and the swap plaintext, but with different nonces.

crypto/src/swap.rs
```diff
@@ -82,14 +231,124 @@ impl From<SwapPlaintext> for pb::SwapPlaintext {
     }
 }
 
+impl From<&SwapPlaintext> for [u8; SWAP_LEN_BYTES] {
+    fn from(swap: &SwapPlaintext) -> [u8; SWAP_LEN_BYTES] {
+        let mut bytes = [0u8; SWAP_LEN_BYTES];
+        bytes[0] = SWAP_TYPE;
+        bytes[1..65].copy_from_slice(&swap.trading_pair.to_bytes());
```
hdevalence, line 238: Since it's already publicly identified in the `Swap`, do we need to encrypt the trading pair / include it in the `SwapPlaintext`?

Cons: takes extra space, redundant data
Pros: might mean that a client only needs to hold the `SwapPlaintext` to be able to claim the swap, and doesn't necessarily need a full transaction ?

That pro could be pretty good and worth the extra 64 bytes.

crypto/src/swap.rs
```diff
@@ -82,14 +231,124 @@ impl From<SwapPlaintext> for pb::SwapPlaintext {
...
+        let mut bytes = vec![SWAP_TYPE];
+        bytes.extend_from_slice(&swap.trading_pair.to_bytes());
+        bytes.extend_from_slice(&swap.t1.to_le_bytes());
+        bytes.extend_from_slice(&swap.t2.to_le_bytes());
+        bytes.extend_from_slice(&swap.fee.0.to_le_bytes());
+        bytes.extend_from_slice(&swap.pk_d.0);
+        bytes.extend_from_slice(&swap.b_d.compress().0);
+        bytes
+    }
+}
```
hdevalence, line 265: I think we could skip having this impl, to avoid having duplication with the array variant.

crypto/src/swap.rs
```diff
@@ -109,14 +368,38 @@ impl TryFrom<&[u8]> for SwapCiphertext {
     }
 }
 
-// TODO: ideally this would live in `component/dex/` or maybe a new subcrate including only
-// dex-related logic, but the former causes cyclic import issues
 #[derive(Debug, Clone)]
 pub struct TradingPair {
     pub asset_1: AssetId,
     pub asset_2: AssetId,
```
hdevalence, line 374: Maybe these should be private, so we can enforce the invariant that `asset_1 < asset_2` on decode / construction?

I think it would make sense to have
```rust
impl TradingPair {
    pub fn asset_1() -> asset::Id;
    pub fn asset_2() -> asset::Id;
    /// Checks that asset_1 < asset_2 or errors
    pub fn new(asset_1: asset::Id, asset_2: asset::Id) -> anyhow::Result<Self>;
    /// Constructs the canonical representation of the provided pair.
    pub fn canonical_order_for(pair: (asset::Id, asset::Id)) -> Self;
```

This way, we can enforce the ordering constraint, and provide people with a way to construct the canonical representative of the pair. By making the `canonical_order_for` constructor take a tuple, it's harder to accidentally mix it up.

crypto/src/swap.rs
```diff
@@ -109,14 +368,38 @@ impl TryFrom<&[u8]> for SwapCiphertext {
...
 }
 
+impl TradingPair {
+    /// Convert the trading pair to bytes.
+    pub fn to_bytes(&self) -> [u8; 64] {
+        let mut result: [u8; 64] = [0; 64];
+        result[0..32].copy_from_slice(&self.asset_1.0.to_bytes());
+        result[32..64].copy_from_slice(&self.asset_2.0.to_bytes());
+        result
+    }
```
hdevalence, line 384: ditto above in re: encapsulating fixed-size serialization, do we want this to be public?

crypto/src/swap.rs
redshiftzero: We could do something like (since all 0s is used for note encryption, 1u128 is used for memo encryption):

```rust
pub static SWAP_ENCRYPTION_NONCE: Lazy<[u8; 12]> = Lazy::new(|| { let nonce_bytes = 2u128.to_le_bytes(); nonce_bytes[0..12].try_into().expect("nonce fits in array") });
```

(Not for this PR but it's probably a good idea to put all our nonces together somewhere in the future)

crypto/src/swap.rs
redshiftzero: I think you can import `note::derive_symmetric_key` instead of adding it here (we also import `note::derive_symmetric_key` when doing the memo encryption)

crypto/src/swap.rs
redshiftzero: We do not have an Ock type, that seems like a good idea though (could file a followup for it if we wanted to out of scope for this PR). For this PR we do have the `note::derive_ock` function that could be shared - I think we can use the value and note commitment bytes of the swap note plaintext to derive the OCK key

crypto/src/swap.rs
zbuc: Is there any reason `new` shouldn't behave like `canonical_order_for`? Is there any reason we'd want to fail to construct a `TradingPair`?

hdevalence: Well, one reason would be if we're decoding a protobuf — I think we want to enforce that the serializations are canonical, since we may want to use the TradingPair as a lookup key in the database.  In general I think it makes sense to require that data we get from others be canonically encoded, and only use the order-independent constructor in special cases (like in pcli when we are trying to construct a TradingPair for some requested trade, or something).

hdevalence:
Merging this as-is so that it's possible to work on other parts of the swap/dex system in parallel.

@zbuc Could you make a list of issues tracking the remaining TODOs?

zbuc:
> Merging this as-is so that it's possible to work on other parts of the swap/dex system in parallel.
> 
> @zbuc Could you make a list of issues tracking the remaining TODOs?

https://github.com/penumbra-zone/penumbra/issues/1216

PR #1130, opened 2022-07-13
Implement block inclusion proof support in TCT

zbuc:
We should probably have proptests that the proofs work as intended, but this should be fine as-is for now.


plaidfinch (approved):
This looks good to me.

The property test would want to check the following property: If I create a block that contains some commitment, as well as arbitrary others (forgotten), then convert that block to a `block::Finalized` and grab its root, then insert it into an arbitrary tree (which may have had some operations occur to it before this), then witness the commitment in the entire tree, the `within_block` of that returned `Proof` must `verify` with respect to the block root.

PR #1162, opened 2022-07-14
pcli: add a `pcli q tx HASH` subcommand that prints full transaction data

hdevalence:
Along the way, this:

- Adds Serde support for Transaction
- Switches to colored JSON output


aubrika (approved):
LGTM!

PR #1165, opened 2022-07-15
ibc: allow nil version in ConnectionOpenInit, fix unwrap in ibc events

avahowell:
msg.version = nil is allowed in ConnectionOpenInit. This also fixes an issue where we could unwrap nil values in IBC events (some events necessarily have nil fields, depending on the protocol sequence).

### DIFF
--- a/component/src/ibc/event.rs
+++ b/component/src/ibc/event.rs
@@ -51,8 +51,11 @@ pub fn connection_open_init(
             )
                 .index(),
             (
-                "counterparty_conneciton_id",
-                counterparty.connection_id().unwrap().to_string(),
+                "counterparty_connection_id",
+                counterparty
+                    .connection_id()
+                    .map(|id| id.to_string())
+                    .unwrap_or_default(),
             )
                 .index(),
         ],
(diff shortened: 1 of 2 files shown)


component/src/ibc/event.rs
```diff
@@ -99,8 +105,8 @@ pub fn connection_open_ack(connection_id: &ConnectionId, connection_end: &Connec
                 connection_end
                     .counterparty()
                     .connection_id()
-                    .unwrap()
-                    .to_string(),
+                    .map(|id| id.to_string())
+                    .unwrap_or_default(),
```
hdevalence, line 109: What's the implication of using unwrap_or_default here? If the field is missing, is it safe to replace with ""? Should we be pre-screening these in the check methods?

avahowell: We should be (and are) doing validation of these optionals in the IBC state machine. There are situations where counterparty connection id is expected to be None, in these cases the cosmos SDK emits this as the empty string. In fact, the raw protos also encode it as an empty string, but `ibc-rs` converts it to an option (where None when "" and Some(val) otherwise)

PR #1166, opened 2022-07-15
pd: accelerate block streaming with concurrent state access

hdevalence:
We need to send block responses in order, but fetching the compact block
involves disk I/O, so we want to look ahead and start fetching compact blocks,
rather than waiting for each state query to complete sequentially.

To do this, we spawn a task that runs ahead and queues block fetches from the
state.  Each block fetch is also spawned as a new task, so they execute
independently, and those tasks' JoinHandles are sent back to this task using a
bounded channel.  The channel bound prevents the queueing task from running too
far ahead.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -635,15 +635,6 @@ checksum = "869119e97797867fd90f5e22af7d0bd274bd4635ebb9eb68c04f3f513ae6c412"
 
 [[package]]
 name = "cast"
-version = "0.2.7"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4c24dab4283a142afa2fdca129b80ad2c6284e073930f964c3a1293c225ee39a"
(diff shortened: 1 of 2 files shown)


hdevalence:
With this change I see >20k blocks/second served by a local `pd` instance:
```
    Finished release [optimized + debuginfo] target(s) in 6.23s
     Running `target/release/measure -n localhost stream-blocks`
[13s] ██████████████████████████████████████████████████  267867/267867  20145/s ETA: 0s
```

plaidfinch:
Oh, sweet! My suggestion here: since calls to RocksDB already internally use `spawn_blocking` to bridge each blocking call to the database, I think it might be more efficient here (in terms of server resource usage) to use `FuturesOrdered` rather than spawning a new task to wrap each block fetch.

plaidfinch (approved):

Issue #1022, opened 2022-06-16
Change `state_key` methods to return keys, not key hashes


hdevalence:
**Is your feature request related to a problem? Please describe.**

Currently, we have (or are in the process of having, as soon as #997) is done a `mod state_key` pattern, where all of the state keys are computed using `state_key::some_descriptive_name` functions in a standalone module, making it easy to see what all the state keys are.

Unfortunately, these methods all return `KeyHash`es, and the ICS23 proofs require having the full key, not just the key hash.  This messes up the state query interface, because we can't use the same code to query with or without a proof.  The resulting RPC implementation is a mess.

**Describe the solution you'd like**

Change the state key methods to return `String` or `&'static str` (as appropriate), using `.into()` to convert to `KeyHash`es as required.  Then clean up the `KeyValueRequest` to only take a `key`, not a `key_hash`, and to unconditionally do `get_with_proof` on that `key`, with the only difference being whether or not the proof is returned to the user.

plaidfinch:
Is doing `get_with_proof` more expensive than doing a `get` that doesn't return a proof?

hdevalence:
No, not really, they have to access the same data from the database, which is the expensive part (because in order to find the leaf, they have to traverse the tree from the root), so bundling it up into a proof is cheap.

PR #1176, opened 2022-07-18
refactor(component/stake): factor out explicit key formatting

### LINKED ISSUE #997: Use `state_key` pattern in staking component
Right now there are 19 instances of the string `format!("staking/` in the file `stake/component.rs`: it hasn't yet been refactored to use the `state_key` pattern we're adopted elsewhere. We should fix this to ensure that it is not brittle to modification of these paths. While so doing, we should also normalize the paths to add a trailing `/` separating the last path element from the interpolated value.

erwanor:
This PR fixes #997 (as a prereq to #1022 ). I've followed the naming convention set out in [shielded_pool::state_key](https://github.com/penumbra-zone/penumbra/blob/main/component/src/shielded_pool/state_key.rs#L16).

Happy to incorporate feedback if naming doesn't look right...

### DIFF
--- a/component/src/stake/component.rs
+++ b/component/src/stake/component.rs
@@ -30,6 +30,7 @@ use crate::stake::{
     rate::{BaseRateData, RateData},
     validator::{self, Validator},
     DelegationChanges, Uptime,
+    state_key,
 };
 
 // Max validator power is 1152921504606846975 (i64::MAX / 8)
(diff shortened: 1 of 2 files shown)


hdevalence:
I think we could probably merge it optimistically, especially since it will cause conflicts with any other changes to the `stake` code.

PR #1184, opened 2022-07-18
Define governance protos and domain types

### LINKED ISSUE #1168: First pass of protos for governance component
Ref #1138

### LINKED ISSUE #1169: Domain types for governance component
After #1168 is complete (and any design feedback is incorporated)

plaidfinch:
This resolves #1168 and resolves #1169.


proto/proto/governance.proto
```diff
@@ -0,0 +1,80 @@
...
+      // The name of the parameter.
+      string parameter = 1;
+      // Its new value, as a string (this will be parsed as appropriate for the parameter's type).
+      string value = 2;
+    }
+  }
+
+  // A DAO spend proposal describes zero or more transactions to execute on behalf of the DAO, with
+  // access to its funds, and zero or more scheduled transactions from previous passed proposals to
+  // cancel.
```
plaidfinch, line 64: @hdevalence I've modified the originally discussed plan for `DaoSpend` proposals to allow a single proposal to schedule an arbitrary number of transactions, or to cancel an arbitrary number of transactions previously scheduled. I don't think this will be difficult to implement, and I think it will make the use of the DAO substantially less onerous: in cases where a pattern of transactions is desired, such as a repeated set of payments or a sequence of LP positions, one vote can approve them all, but they can still be canceled later in reaction to changing circumstances by another proposal. What do you think about this design?

hdevalence: Hmm, that sounds useful!

proto/proto/governance.proto
```diff
@@ -0,0 +1,103 @@
...
+  message DaoSpend {
+    // The sequence of transactions to schedule for execution.
+    repeated ScheduledTransaction schedule_transactions = 1;
+
+    // A set of previously-scheduled transactions to cancel before they are executed.
+    repeated ScheduledTransaction cancel_transactions = 2;
```
hdevalence, line 70: How do we identify a transaction to be canceled? By checking equality of `ScheduledTransaction`s or `TransactionPlan`s?  I'm wondering if it would make sense for the `cancel_transactions` to just be a `repeated transaction.AuthHash` instead?  That data should uniquely identify the transaction plan.

proto/proto/governance.proto
```diff
@@ -0,0 +1,103 @@
...
+  message ParameterChange {
+    // The height at which the change should take effect.
+    uint64 effective_height = 1;
+
+    // The set of changes to chain parameters.
+    repeated SetParameter new_parameters = 2;
+
+    // A single change to an individual chain parameter.
+    message SetParameter {
+      // The name of the parameter.
```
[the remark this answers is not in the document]
plaidfinch: The reason I decided to go with stringly-typed rather than explicitly-enumerated was so that third-party software such as block explorers doesn't need to update in order to keep pace with changes to the set of possible parameters. Same justification for why the value is stringly-typed in the proto, but parsed (and validated) at proposal submission time.

If we look at how Cosmos Hub does it, they have a namespaced parameter universe: https://www.mintscan.io/cosmos/proposals/47. Ours is flat right now, but perhaps this indicates that we may want to have namespacing in the future, and we should migrate to that pattern now rather than later. What do you think?

proto/proto/governance.proto
plaidfinch: My intention was to check the scheduled transactions pending for block `execution_height` and delete the first that is equal to the listed `TransactionPlan`. Your approach would work too, I'll change to that.

The specific semantics of "delete the first that matches" rather than "delete all that match" mean that it is permissible to have duplicate transactions scheduled for the same block, and to have a vote that unschedules only one of them. (If the semantics said "delete all" then you'd have to specify an ordering between scheduling and canceling, so that canceling happens first, then scheduling, so that you could get "unschedule one of two" by first unscheduling all, then scheduling one -- this way, scheduling and canceling commute in effect.)

plaidfinch:
I've defined the authorization hashes for all the transparent governance items and left a TODO note where the auth hash for delegator votes will go, once those bodies are implemented. I think this is ready to merge now, but I would appreciate review on my implementation of authorization hashes (perhaps by @hdevalence?).

transaction/src/auth_hash.rs
```diff
@@ -124,6 +130,18 @@ impl TransactionPlan {
...
+                .personal(b"PAH:propose")
+                .hash(&payload.encode_to_vec());
+            state.update(auth_hash.as_bytes());
+        }
+        for payload in self.withdraw_proposals() {
+            let auth_hash = Params::default()
+                .personal(b"PAH:withdrawproposal")
+                .hash(&payload.encode_to_vec());
+            state.update(auth_hash.as_bytes());
+        }
```
plaidfinch, line 144: Is this right?

transaction/src/auth_hash.rs
```diff
@@ -223,6 +241,28 @@ impl Undelegate {
...
+        state.update(match self.vote {
+            // Manually choose a distinct byte for each vote type
+            Vote::Yes => b"Y",
+            Vote::No => b"N",
+            Vote::Abstain => b"A",
+            Vote::NoWithVeto => b"V",
+        });
+        state.update(&self.validator_identity.0.to_bytes());
+
+        state.finalize()
```
plaidfinch, line 262: Is this right?

transaction/src/transaction.rs
```diff
@@ -79,35 +79,27 @@ impl Transaction {
...
-            })
-            .collect()
+    pub fn note_payloads(&self) -> impl Iterator<Item = &NotePayload> {
+        self.actions().filter_map(|action| {
+            if let Action::Output(output) = action {
+                Some(&output.body.note_payload)
+            } else {
+                None
+            }
+        })
```
plaidfinch, line 89: This reduces the number of intermediate allocations by returning an iterator rather than a `Vec`; the consumer can `.collect()` it if desired.

transaction/src/transaction.rs
```diff
@@ -79,35 +79,27 @@ impl Transaction {
...
+    pub fn spent_nullifiers(&self) -> impl Iterator<Item = Nullifier> + '_ {
+        self.actions().filter_map(|action| {
+            // Note: adding future actions that include nullifiers
+            // will need to be matched here as well as Spends
+            if let Action::Spend(spend) = action {
+                Some(spend.body.nullifier)
+            } else {
+                None
+            }
+        })
```
plaidfinch, line 102: This reduces the number of intermediate allocations by returning an iterator rather than a `Vec`; the consumer can `.collect()` it if desired.

view/src/worker.rs
```diff
@@ -178,8 +177,7 @@ impl Worker {
             if inbound_transaction_ids.contains(&tx_id)
                 || transaction
                     .spent_nullifiers()
-                    .iter()
-                    .any(|nf| spent_nullifiers.contains(nf))
+                    .any(|nf| spent_nullifiers.contains(&nf))
```
plaidfinch, line 180: This is a fix because the nullifiers now are a copying iterator rather than a method returning a `Vec` of references.

transaction/src/auth_hash.rs
```diff
@@ -110,6 +110,12 @@ impl TransactionPlan {
         for undelegation in self.undelegations() {
             state.update(undelegation.auth_hash().as_bytes());
         }
+        for validator_vote in self.validator_votes() {
+            state.update(validator_vote.auth_hash().as_bytes());
+        }
+        for _delegator_vote in self.delegator_vote_plans() {
+            // TODO: get the authorization hash of the delegator vote body for each plan
+        }
```
plaidfinch, line 118: I think we should merge this PR and then make another one to implement governance **actions** rather than merely action plans as implemented in this one? Or should I keep pushing on this and implement them here?

plaidfinch:
This isn't yet ready, need to get it to a cleaner place.

plaidfinch:
I think this is ready for review now. I am particularly looking for review on the implementation of auth hashes.

PR #1188, opened 2022-07-19
bring IBC consensusStates into IBC key spec

avahowell:
in building the relayer, discovered that our state storage key for consensusStates was out of spec


plaidfinch:
Should we be using the state key pattern here?

hdevalence:
> Should we be using the state key pattern here?

Yeah, sounds good — that would be a good refactor to do separately from this issue!

PR #1189, opened 2022-07-19
add global penumbra app hash prefix for IBC proof compatibility

avahowell:
IBC counterparties verify inclusion proofs of IBC messages in penumbra's state. These inclusion proofs are verified up to penumbra's app hash, which prior to this PR is the root of the JMT. Unfortunately, IBC encodes a *prefix* [requirement](https://github.com/cosmos/ibc-go/blob/main/modules/light-clients/07-tendermint/types/client_state.go#L210), which is a *logical* prefix (i.e., a requirement of a supertree). This prefix [can't be empty](https://github.com/cosmos/ibc-go/blob/193e41480dc8a8994902d14b9f13d72a23df2734/modules/core/23-commitment/types/merkle.go#L124). Our previous AppHash implementation will not meet this requirement, since keys are randomized and there is no stable supertree prefix.

This PR adds a simple supertree: a single extra hash, implemented with a new AppHash type and a corresponding ICS proofspec. The AppHash is now `SHA-256("penumbra-root" || jmt.root())`. This allows us to create valid prefixed merkle proofs for IBC.


storage/src/app_hash.rs
```diff
@@ -0,0 +1,135 @@
...
+            empty_child: vec![],
+            min_prefix_length: 0,
+            max_prefix_length: 0,
+        }),
+        min_depth: 0,
+        max_depth: 2,
+    }
+}
+
+static APPHASH_DOMSEP: &str = "penumbra_root";
```
avahowell, line 43: not sure if this is the best label, it doesn't matter but would be cool to have something cool here :D

hdevalence (approved):
LGTM!

Leaving unmerged in case you want to change the domain separator -- no real opinions on what it should be.

avahowell:
Updated the domain separator to `PenumbraAppHash`

PR #1190, opened 2022-07-19
Instruct users to reset state before testnet join

elsehow:
This PR instructs users to reset the tesetnet state before running the `pd testnet join` command.

The reasoning for this change is this:
- Most testnet participants participate in multiple testnets; most followers of this guide will need to reset state.
- As the file was originally structured, a careless reader blindly running commands might reset the testnet state after running the join command, creating a confusing error condition.

### DIFF
--- a/docs/guide/src/pd/join-testnet.md
+++ b/docs/guide/src/pd/join-testnet.md
@@ -18,7 +18,14 @@ To join a testnet as a fullnode, check out the tag for the current testnet, run
 
 ### Generating configs
 
-To generate a set of configs for the current testnet, run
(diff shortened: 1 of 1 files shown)


hdevalence (changes requested):
This is a good change, but I think it was useful to have a separate "Resetting state" section, so it's clear that those commands are doing a separate thing.

Could we keep the header, so it's

```
### Resetting state

First, ...

### Generating configs

```
and the ordering is fixed, but it's still clear that they're doing different things?

elsehow:
Great suggestion @hdevalence - updated the docs to match.

PR #1193, opened 2022-07-19
feat: Add methods to handle ephemeral addresses 

### LINKED ISSUE #984: Add methods to handle ephemeral addresses
Following #983, we should have:

- methods to generate an ephemeral address;
- a `pcli addr ephemeral` command that generates and prints an ephemeral address;
- code in `pcli` that displays numeric address indices separately from ephemeral ones (which may be forced by #983 already).

redshiftzero:
Closes #984


hdevalence (approved):
LGTM!

PR #1204, opened 2022-07-21
view: couple transaction download with scanning

hdevalence:
The first version of this code had extended transaction details downloaded by a
separate task, decoupled from the main block scanner.  This was intended to
allow the block scanning to continue while waiting on transaction downloads.
However, it turned out to cause a lot of complexity, not in the code itself,
but in the resulting semantics: what does it mean if we error while downloading
the full transactions, but not while syncing? What if we cut out before
downloading the full transactions -- how can we resume or retry?  What if
someone asks the view service for details about a transaction that has been
detected but not yet downloaded?  All of this complexity goes away if we
require that the full transaction details be known when we record the scanned
block, so this commit changes the code to do that.


hdevalence:
I haven't tested this code yet.

PR #1207, opened 2022-07-22
stake: ban duplicate consensus keys

### LINKED ISSUE #1062: Add a consensus key signature to validator definitions to prevent rogue key attacks

Currently, the root of the validator identity is an `IdentityKey` (a decaf377-rdsa signing key), which is used to sign a `ValidatorDefinition` containing the consensus key.  This allows validators to rotate their consensus keys. However, although the validator definition signature establishes that the consensus key is authorized by the identity key, we don't currently require any authorization from the consensus key.

This opens the door to a rogue key attack, where someone can upload a validator definition with a consensus key they don't control.  It's not totally clear why they'd do this, but they could, for instance, "piggyback" on the behavior of another validator. 

To prevent this, the `ValidatorDefinition` should have a signature made by the consensus key on the identity key, proving that the creator of the validator definition has knowledge of the consensus secret key.

This signature could be on the message
```
b"penumbra_val_auth" || identity_key
```
Signing the identity key is enough to prove knowledge of the consensus secret key, and signing only the identity key means the signature doesn't have to be changed on every config change.

The signature could be added as a `bytes identity_key_sig` to the proto, and an `ed25519_consensus::Signature` on the domain type.

Building a validator definition now requires additional steps, because it needs to create a signature.  To handle this, the `template-definition` command should be extended with an optional parameter with the path to a Tendermint `priv_validator_key.json` file.  If the path is provided, the command should read the JSON file and load the secret key.  Otherwise, it should generate an ephemeral ed25519 secret key.  Then, it should produce a signature on the identity key, and write the public key and signature into the template.

The reason to make the parameter optional is so that if the ed25519 secret is not available (e.g., it's on an HSM, or whatever), it can be edited into the template later.

hdevalence:
Closes #1062 (albeit with a different mechanism).

This prevents a chain halt that can occur if two validator definitions are
uploaded with the same consensus key -- resulting in pd sending a validator
update to Tendermint with duplicate entries, causing a halt.

This change obviates the suggestion in #1062 to add a signature by the
consensus key -- now, it's impossible to create a validator definition with an
existing key, so no signature is needed.

This commit should fix the chain halt that killed the Eurydome testnet.

### DIFF
--- a/component/src/stake/component.rs
+++ b/component/src/stake/component.rs
@@ -1041,53 +1041,44 @@ impl Component for Staking {
             let v = validator::Definition::try_from(v.clone())
                 .context("supplied proto is not a valid definition")?;
 
-            let existing_v_by_id = self.state.validator(&v.validator.identity_key).await?;
-            let existing_v_by_ck = self
+            // Check whether we are redefining an existing validator.
+            if let Some(existing_v) = self.state.validator(&v.validator.identity_key).await? {
+                // Ensure that the highest existing sequence number is less than
+                // the new sequence number.
+                let current_seq = existing_v.sequence_number;
+                if v.validator.sequence_number <= current_seq {
+                    return Err(anyhow::anyhow!(
+                        "expected sequence numbers to be increasing: current sequence number is {}",
+                        current_seq
+                    ));
+                }
+            }
+
+            // Check whether the consensus key has already been used by another validator.
+            if let Some(existing_v) = self
                 .state
                 .validator_by_consensus_key(&v.validator.consensus_key)
-                .await?;
-
-            tracing::debug!(
-                ?v,
-                ?existing_v_by_id,
-                ?existing_v_by_ck,
-                "checking validator definition"
-            );
-
-            match (existing_v_by_id, existing_v_by_ck) {
-                // This is a redefinition of an existing validator.  Ensure that
-                // the highest existing sequence number is less than the new
-                // sequence number.
-                (Some(existing_v), _) => {
-                    let current_seq = existing_v.sequence_number;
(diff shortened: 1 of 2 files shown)


hdevalence:
This doesn't actually detect duplicate consensus keys, for some reason.

To test:
```
pd testnet generate
pd start --home ... # as in devnet quickstart
tendermint start --home ... # as in devnet quickstart
pcli validator fetch-definition -f exist.json VALIDATOR_ID
pcli validator template-definition -f dup.json
vim dup.json # paste ck from exist.json
pcli validator upload-definition -f dup.json # should fail, succeeds
```

hdevalence:
After fixing a latent bug that also affected tombstoning (cf #1209), we now reject duplicate consensus keys:
```
pre-checking transaction...
broadcasting transaction...
Error: Error submitting transaction: code 1, log: consensus key Ed25519(VerificationKey("cfe8f3f3b290d4a3c4edc585d123bb4e38d67d387b97df1c2a3aa7bc28c30d2f")) is already in use by validator penumbravalid1v6musy2yt6355atdttrflrcqqduxtrh7e7pzspg72du57kj43uqqm423ty
```

plaidfinch:
This looks good! I also notice that it seems like if you rotate your consensus key, this will ban anyone (including you) from using that consensus key ever again. This seems good, I think?

hdevalence:
It bans anyone else from using that key, but it's still possible for you to use it in a subsequent update — and actually, the current code is incorrect, because the check is only applied to new validators, not to updates to existing ones, so I can make an update to my validator that takes over someone else's consensus key.... I'll fix that.

hdevalence:
Should be fixed now.

PR #1212, opened 2022-07-22
Output validator custody file in pcli-compatible format

plaidfinch:
This changes the output of a generated validator's identity key to be compatible with the `pcli` `"custody.json"` format, by extracting `pcli`'s `Wallet` struct into the `wallet` crate and using that to output the spend key.


hdevalence:
I don't think we should call this "wallet", since it's only used by the SoftHSM code and doesn't contain any data that people would normally call wallet data.

hdevalence (approved):

PR #1213, opened 2022-07-23
Update priv_validator_key.json path

alterkahn:
small fix to the giude docs that changing the `priv_validator_key.json` path


hdevalence:
Thanks!

PR #1225, opened 2022-07-28
Await nullifiers in view service, making transaction confirmation robust when there is no change

### LINKED ISSUE #999: Add await-nullifier functionality to the view service
@hdevalence proposes:

```protobuf
rpc NullifierStatus(NullifierStatusRequest) returns (NullifierStatusResponse);

message NullifierStatusRequest {
  crypto.Nullifier nullifier =2;
  bool keep_alive = 3;
}

message NullifierStatusResponse {
  bool spent = 1;
}
```

> idea being that if keep_alive is present, the connection stays open until it becomes spent

With quarantining implemented, we may also want to add an option to watch for a nullifier being rolled back, or to watch for it only once it is finally confirmed, rather than provisionally added to the quarantined nullifier set.

plaidfinch:
Resolves #999.

### DIFF
--- a/proto/proto/view.proto
+++ b/proto/proto/view.proto
@@ -45,6 +45,9 @@ service ViewProtocol {
 
     // Query for a note by its note commitment, optionally waiting until the note is detected.
     rpc NoteByCommitment(NoteByCommitmentRequest) returns (NoteRecord);
+
+    // Query for whether a nullifier has been spent, optionally waiting until it is spent.
+    rpc NullifierStatus(NullifierStatusRequest) returns (NullifierStatusResponse);
 }
 
 message NoteByCommitmentRequest {
@@ -161,4 +164,14 @@ message QuarantinedNoteRecord {
 message QuarantinedNotesRequest {
     // Identifies the FVK for the notes to query.
     crypto.FullViewingKeyHash fvk_hash = 1;
+}
+
+message NullifierStatusRequest {
+  crypto.FullViewingKeyHash fvk_hash = 1;
+  crypto.Nullifier nullifier = 2;
+  bool await_detection = 3;
+}
+
+message NullifierStatusResponse {
+  bool spent = 1;
 }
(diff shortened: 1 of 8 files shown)


proto/proto/view.proto
```diff
@@ -45,6 +45,9 @@ service ViewProtocol {
 
     // Query for a note by its note commitment, optionally waiting until the note is detected.
     rpc NoteByCommitment(NoteByCommitmentRequest) returns (NoteRecord);
+
+    // Query for whether a nullifier has been spent, optionally waiting until it is spent.
+    rpc NullifierStatus(NullifierStatusRequest) returns (NullifierStatusResponse);
```
hdevalence, line 50: Rather than having this be a "nullifier status" that just returns a true/false, why not have
```
rpc NoteByNullifier(NoteByNullifierRequest) returns (NoteRecord);
```
with
```
message NoteByNullifierRequest {
  crypto.Nullifier nullifier = 2;
  // If set, waits until the note with this nullifier has been recorded as spent.
  bool await_spent = 3;
}
```

This is strictly more useful than the current RPC, because rather than just being useful for spend detection, it can also be used for general queries -- and if it is used for spend detection, a client gets info on the state of the note, and then they can choose to discard it, or not.

If we wanted to have a common path just for detecting spends that didn't make the user inspect the `NoteRecord`, we could implement that as a convenience method on the `ViewService` trait.

plaidfinch: I'd like to be able to do that -- but I think it isn't possible with the current separation of `NoteRecord` and `QuarantinedNoteRecord`: as implemented in this PR, the nullifier request checks both quarantined nullifiers and un-quarantined nullifiers, since when you undelegate, you should get a confirmation when the quarantined nullifier hits the chain, not all the way after it finally unbonds. To do this we'd have to either unify `NoteRecord` and `QuarantinedNoteRecord` (which we previously decided against) or have two separate kinds of nullifier request, which seems easy to mis-use.

What we could do, however, is return a record that just pairs the spent status of the nullifier with the commitment to which it pertains, so that then you can request the note record or quarantined note record as appropriate by using that commitment. Would that be desirable?

hdevalence: Yeah, OK, that makes sense, let's not do this suggestion then.

hdevalence (approved):

PR #1228, opened 2022-07-29
refactor(state_key): only use raw keys to build  `KeyValueRequest`s 

### LINKED ISSUE #1022: Change `state_key` methods to return keys, not key hashes

erwanor:
This PR fixes #1022, here's a breakdown of what it does:

1. [x] make all `state_key` methods return their respective raw path instead of `KeyHash`
2. [x] make downstream consumers of `state_key` modules meet the new API
3. [x] protobuf: remove the `key_hash` field from the `KeyValueRequest` message definition
4. [x] pd: modify how "specific" `Info` requests for `key_value` are executed (always use `get_with_proof`)
5. [x] pcli: stop using `KeyHash` in query commands for `ShieldedPool` (WIP - I have questions about this)
7. [x] cargo fmt
8. [x] testing

There are two chunks of code that require increased scrutiny from reviewers:
- [x] pd: check that the new logic for the key value req builder makes sense ([see diff here](https://github.com/penumbra-zone/penumbra/commit/9fbc0ef799bfcaec9c86dad389421e7832932d41))
- [ ] pcli: check that the simplified logic to exec `QueryCmd`s makes sense ([see diff here](https://github.com/penumbra-zone/penumbra/pull/1228/commits/2c762af1a4b54dabe3764eda576d41abcaa71dfc#diff-513818ec006aae6ecbe2ac3333af33dd6494d6d31c4cb18b179b11859dfdd8c2R31))

### DIFF
--- a/chain/src/state_key.rs
+++ b/chain/src/state_key.rs
@@ -1,13 +1,11 @@
-use jmt::KeyHash;
-
-pub fn chain_params() -> KeyHash {
-    "chain_params".to_string().into()
+pub fn chain_params() -> &'static str {
+    "chain_params"
 }
 
-pub fn block_height() -> KeyHash {
-    "block_height".to_string().into()
+pub fn block_height() -> &'static str {
+    "block_height"
 }
 
-pub fn block_timestamp() -> KeyHash {
-    "block_timestamp".to_string().into()
+pub fn block_timestamp() -> &'static str {
+    "block_timestamp"
 }
--- a/chain/src/view.rs
+++ b/chain/src/view.rs
@@ -23,7 +23,8 @@ pub trait View: StateExt {
 
     /// Writes the provided chain parameters to the JMT.
     async fn put_chain_params(&self, params: ChainParams) {
-        self.put_domain(state_key::chain_params().into(), params).await
+        self.put_domain(state_key::chain_params().into(), params)
+            .await
     }
 
     /// Gets the current epoch for the chain.
--- a/chain/src/view.rs
+++ b/chain/src/view.rs
@@ -16,14 +16,14 @@ use crate::{params::ChainParams, state_key, Epoch};
 pub trait View: StateExt {
     /// Gets the chain parameters from the JMT.
     async fn get_chain_params(&self) -> Result<ChainParams> {
-        self.get_domain(state_key::chain_params())
+        self.get_domain(state_key::chain_params().into())
             .await?
             .ok_or_else(|| anyhow!("Missing ChainParams"))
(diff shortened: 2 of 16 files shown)


erwanor:
Right now, `pcli` uses a `key_hash` to query `ShieldedPool` data. I'm trying to evaluate if using a raw key is OK. After taking a look at [`ShieldedPool`](https://github.com/penumbra-zone/penumbra/blob/main/pcli/src/command/query/shielded_pool.rs#L11), this seems fine but I need confirmation before I proceed! 

Once that's cleared, we will be able collapse this branch in [pcli/srccommand/query.rs](https://github.com/penumbra-zone/penumbra/blob/main/pcli/src/command/query.rs#L26) so that `QueryCmd::Key` and `QueryCmd::ShieldedPool` use the same logic to build their `KeyValueRequest`s.

hdevalence:
Querying with the key, rather than the key hash, seems fine (from a privacy point of view -- not 100% sure if that's the concern?), because the key hashes aren't actually hiding: if a full node wanted to monitor queries, they could do that just as easily with key hashes as full keys.

I looked briefly at

> check that the new logic for the key value req builder makes sense

and it seems fine -- it will compute a proof every time, but that doesn't seem terribly important relative to having simpler code.

erwanor:
I've done some cursory manual testing with the following queries, and they all return the same result as the main testnet branch:

```
cargo run --release --bin pcli -- q key staking/validator/penumbravalid142jc0tm6dkweww3z5t6rqdj4mkcsdw3qn329rys8vsvrp3fgvc9srtsvf6
```

```
cargo run --release --bin pcli -- q shielded-pool block-anchor 137359
```

hdevalence (approved):
LGTM!

hdevalence:
This is great, thanks!

Issue #1216, opened 2022-07-25
Swap TODO items


zbuc:
There were some in-progress TODO items from the Swap implementation that weren't completed prior to merge.

- [x] Record `swap_nft_value` as `asset::Id` rather than `Value` (since the amount is always `1`)
- [x] Rename `swap_nft_value_commitment` to `fee_commitment` to reflect that the `SwapClaim`'s only effect on the transaction value balance is contributing pre-committed fees
- [x] Include asset ID in witness data: https://github.com/penumbra-zone/penumbra/pull/1104/files#r925088722
- [x] Check format of Swap NFT asset ID (and other `SwapPlaintext` fields) in `SwapClaimProof` transparent proof
- [x] Fix `SwapClaimProof` value commitment integrity check: We want to check that the value_commitment is a commitment to f penumbra tokens, where f is the fee amount committed to by the swap nft's asset id.
- [x] #1266
- [x] Match transparent proof to ZK proof implementation by recording integer amounts for asset 1/asset 2 amounts rather than `Value` types
- [x] `value_fee: Value` -> `fee_amount: u64`
- [x] Need blinding factor for _each_ of t1/t2/fee
- [x] Remove `SWAP_TYPE`
- [x] Drop `swap::Error` enum and use `anyhow`
- [x] #1265
- [x] #1267
- [x] #1268
- [x] Remove `impl From<&SwapPlaintext> for Vec<u8> {`
- [x] Make `asset_1`/`asset_2` private to enforce `asset_1 < asset_2`
- [x] Make `to_bytes` private

More context here: https://github.com/penumbra-zone/penumbra/pull/1104

Issue #1126, opened 2022-07-13
Define how exactly we record clearing prices in `SwapClaim`s


hdevalence:
The protocol specification [currently describes](https://protocol.penumbra.zone/main/zswap/swap.html#batching-and-execution) how the results of a swap are computed using real-number prices as follows. The $i$-th user in the batch has submitted inputs $\Delta_1^i, \Delta_2^i$ and computes their outputs as
$$\Lambda_1^i = b (p_{t_1, t_2} \Delta_2^i) + (1 - b) \Delta_1^i$$
$$\Lambda_2^i = b (p_{t_2, t_1} \Delta_1^i) + (1 - b) \Delta_2^i$$
where $p_{t_1, t_2}$ and $p_{t_2, t_1}$ are the effective prices of $t_1$ in terms of $t_2$ and vice versa, and $b$ is a success bit set to $1$ if the swap succeeded and $0$ if it failed.  These formulas simplify to $\Lambda_1^i = p_{t_1, t_2} \Delta_2^i$, $\Lambda_2^i = p_{t_2, t_1} \Delta_1$ when the batch succeeds and $b = 1$, or to $\Lambda_1^i = \Delta_1^i$, $\Lambda_2^i = \Delta_2^i$ when the batch fails and $b = 0$.

This description gives the basic idea, but it's not workable to implement, because we need to specify how to compute the clearing prices using machine integers, not just real numbers, and we also need to specify how the clearing prices are written into the chain state.

In the staking computations, we used fixed-precision arithmetic with an implied, fixed denominator.  That approach is less ideal here, because using fixed-precision arithmetic means making an upfront choice of how to allocate precision, and it's not clear that there's a good way to make that choice. 

Instead, we could look at how we would actually compute the prices $p_{t_1, t_2}$.  These prices are _effective_ prices, inclusive of fees, so they're downstream of the actual AMM execution. If the combined batch input is
$$(\Delta_1, \Delta_2) = (\sum_i \Delta^i, \sum_i \Delta^i)$$
[...]

PR #1238, opened 2022-08-02
Swap fixes and cleanup

zbuc:
Addresses various items from https://github.com/penumbra-zone/penumbra/issues/1216:

* Record swap_nft_value as asset::Id rather than Value (since the amount is always 1)
* Rename swap_nft_value_commitment to fee_commitment to reflect that the SwapClaim's only effect on the transaction value balance is contributing pre-committed fees
* Include asset ID in witness data: https://github.com/penumbra-zone/penumbra/pull/1104/files#r925088722
* Check format of Swap NFT asset ID (and other SwapPlaintext fields) in SwapClaimProof transparent proof
* Fix SwapClaimProof value commitment integrity check: We want to check that the value_commitment is a commitment to f penumbra tokens, where f is the fee amount committed to by the swap nft's asset id.
* Match transparent proof to ZK proof implementation by recording integer amounts for asset 1/asset 2 amounts rather than Value types
* value_fee: Value -> fee_amount: u64
* Need blinding factor for each of t1/t2/fee
* Remove SWAP_TYPE
* Drop swap::Error enum and use anyhow
* Remove impl From<&SwapPlaintext> for Vec<u8> {
* Make asset_1/asset_2 private to enforce asset_1 < asset_2
* Make to_bytes private


crypto/src/dex/swap.rs
```diff
@@ -1,33 +1,101 @@
 mod ciphertext;
 mod plaintext;
+
+use anyhow::Result;
+use ark_ff::PrimeField;
 pub use ciphertext::SwapCiphertext;
+use decaf377::{Element, FieldExt, Fq};
+use decaf377_ka::Public;
```
hdevalence, line 8: nit: probably better not to interleave the re-exports defining the exposed api with the module's own use statements

crypto/src/dex/swap/plaintext.rs
```diff
@@ -163,27 +159,38 @@ impl Protobuf<pb::SwapPlaintext> for SwapPlaintext {}
...
+        //     .b_d
+        //     .try_into()
+        //     .map_err(|_| anyhow::anyhow!("invalid diversified basepoint in SwapPlaintext"))?;
+        // let b_d_encoding = decaf377::Encoding(b_d_bytes);
+        // let encoded_addr = format!(
+        //     "{}",
+        //     plaintext
+        //         .claim_address
+        //         .ok_or_else(|| anyhow::anyhow!("missing SwapPlaintext claim_address"))
+        // );
```
hdevalence, line 172: can we prune this commented section?
```suggestion
```

crypto/src/dex/swap/plaintext.rs
```diff
@@ -163,27 +159,38 @@ impl Protobuf<pb::SwapPlaintext> for SwapPlaintext {}
...
-                })?,
-            ),
+            // b_d: b_d_encoding.decompress().map_err(|_| {
+            //     anyhow::anyhow!("error decompressing diversified basepoint in SwapPlaintext")
+            // })?,
+            // pk_d: ka::Public(
+            //     plaintext.pk_d.try_into().map_err(|_| {
+            //         anyhow::anyhow!("invalid diversified publickey in SwapPlaintext")
+            //     })?,
+            // ),
```
hdevalence, line 193:
```suggestion
```

crypto/src/dex/trading_pair.rs
```diff
@@ -1,17 +1,34 @@
...
 }
 
 impl TradingPair {
+    pub fn new(asset_1: asset::Id, asset_2: asset::Id) -> Result<Self> {
+        if asset_2 < asset_1 {
+            return Err(anyhow!("asset_2 must be greater than asset_1"));
+        }
+
+        Ok(Self { asset_1, asset_2 })
+    }
```
hdevalence, line 20: It'd be handy to have a constructor that picks the correct ordering -- the test code for the LPNFTs constructs a `TradingPair`, for instance, and without having a way to get the ordering right, it would fail the test on deserialization if the ordering happens to be wrong.

crypto/src/dex/trading_pair.rs
```diff
@@ -1,17 +1,34 @@
...
+        self.asset_1
+    }
+
+    pub fn asset_2(&self) -> asset::Id {
+        self.asset_2
+    }
+
     /// Convert the trading pair to bytes.
-    pub fn to_bytes(&self) -> [u8; 64] {
+    pub(crate) fn to_bytes(&self) -> [u8; 64] {
```
hdevalence, line 31: do we just use this in the swap plaintext encoding? if so, could we fold it in there directly, rather than leaving it available for general use?

crypto/src/proofs/transparent.rs
```diff
@@ -418,57 +436,109 @@ impl SwapClaimProof {
...
+    /// * the pre-paid fee amount for the swap,
     pub fn verify(
         &self,
-        anchor: tct::builder::block::Root,
+        anchor: tct::Root,
         // Value commitment to the fees for the swap claim
         value_commitment: value::Commitment,
         nullifier: Nullifier,
         clearing_price_1: u64,
         clearing_price_2: u64,
```
hdevalence, line 447: Rather than taking `clearing_price_i`, we'll want the proof verification to take a `BatchSwapOutputData`, which will let us compute the pro rata share of the output amount exactly.

hdevalence: (This also means we don't have to add an explicit `success` bool below)

crypto/src/proofs/transparent.rs
```diff
@@ -418,57 +436,109 @@ impl SwapClaimProof {
...
         }
 
-        // TODO: is there an address check that needs to occur?
+        // TODO:
+        // The address should be the same for the Swap NFT and SwapClaim outputs
+        // Need output notes here, and to validate the amounts and addresses.
+        // let lambda_1 = success.into() * (clearing_price_1 * self.delta_2)
+        //     + (1 - success.into()) * self.delta_1;
+        // let lambda_2 = success.into() * (clearing_price_2 * self.delta_1)
+        //     + (1 - success.into()) * self.delta_2;
```
hdevalence, line 533: We should also tweak the description in the protocol spec to match the data defined in #1126 / #1214, so that the formulas in the spec are written in terms of a pro rata share of the batch, rather than in terms of a multiplication-by-price-factor (which is harder to do precisely).

proto/proto/dex.proto
```diff
@@ -21,33 +23,44 @@ message Swap {
...
-  // A commitment to the fee's contribution to the transaction's value balance.
+  // @exclude These will become commitments when flow encryption/ABCI++ are available
+  // The amount for asset 1.
+  uint64 delta_1 = 2;
+  // The amount for asset 2.
+  uint64 delta_2 = 3;
+  // @exclude // Commitment to the amount for asset 1 (delta 1).
+  // @exclude bytes delta_1_commitment = 2;
+  // @exclude // Commitment to the amount for asset 2 (delta 2).
+  // @exclude bytes delta_2_commitment = 3;
```
hdevalence, line 62: what does `@exclude` do?

hdevalence (approved):
This is great progress, let's merge it and keep iterating.

crypto/src/dex/trading_pair.rs
hdevalence: Also, it looks like the ordering check is only applied in `new`, rather than also being applied when we deserialize the proto, so we should add it there too.

proto/proto/dex.proto
zbuc: It is a keyword in the protobuf doc generator that excludes the comment from the docs

crypto/src/dex/trading_pair.rs
zbuc: @hdevalence what behavior should we have for `pub fn canonical_order_for(pair: (asset::Id, asset::Id)) -> Self;` if the same asset ID is passed twice? It seems like we might want to have a `Result` type returned, because an asset pair should consist of unique assets? Or should we just return a `TradingPair` with duplicated assets?

Issue #1232, opened 2022-07-30
gRPC proxy to Tendermint RPC


hdevalence:
**Is your feature request related to a problem? Please describe.**

Currently, Penumbra clients need to use two different types of RPC to use the network: Penumbra-specific gRPC via `pd`, and Tendermint RPC via Tendermint's JSON-RPC endpoint.  This situation causes a number of problems:

* Clients have to work with two different kinds of RPC, rather than just one.
* Clients have to speak to two different endpoints, rather than just one. Currently, we bake in the assumption that both Tendermint and pd are on the same host, but this might not always be the case.
* Clients have to work with the quirks of the Tendermint JSON encoding, which has various edge cases (e.g., some methods that expect data to be hex encoded when the request is sent with GET but base64 encoded when it's sent with POST).
* The Tendermint RPC isn't, generally speaking, safe to expose, making it risky for a full node to support clients, which isn't what we want — it should be an easy and safe default for full node operators to provide support to clients who want to use the network.

**Describe the solution you'd like**

We should create a gRPC version of the methods in the Tendermint RPC, then add a proxy implementation of that gRPC service to `pd` that converts each a gRPC request into a JSON-RPC request, forwards it to Tendermint, and converts the JSON response back into a proto to send to the caller.  

The Cosmos SDK provides similar functionality, and with a similar implementation strategy (defining a set of protos, and proxying requests), but it only supports a subset of methods (notably missing transaction broadcast and ABCI queries), and uses an in-process Tendermint instance, so the "rpc" is really a function call.

After defining and building a working gRPC version of the Tendermint RPC, we should attempt to upstream it to Tendermint itself. The initial implementation of this could copy the approach in the Cosmos SDK, translating requests with an in-process client. At that point, we could eliminate the JSON-RPC support in pd, and do a pure gRPC proxy.

Finally, once we have a pd proxy to Tendermint, we can configure rate limiting and request filtering (using Tower middleware) to protect the Tendermint instance against DoS, making it safer for node operators to support client RPC.

Concrete steps to complete this:

- [ ] Add a command line option to pd that gives the address of the tendermint instance to proxy requests to. (This is necessary, because currently, Tendermint connects to pd, and not the other way around).
- [ ] Vendor the protos in https://buf.build/cosmos/cosmos-sdk/docs/main:cosmos.base.tendermint.v1beta1 and extend them to cover all Tendermint RPC methods.
[...]

PR #1241, opened 2022-08-02
feat(pd): a gRPC proxy service for tendermint RPCs

erwanor:
This _draft_ PR implements #1232

### DIFF
--- a/pd/src/proxy/tendermint.rs
+++ b/pd/src/proxy/tendermint.rs
@@ -1,0 +1,97 @@
+use penumbra_proto::tendermint_proxy::tendermint_query_server::TendermintQuery;
+use penumbra_proto::tendermint_proxy::{
+    AbciQueryRequest, AbciQueryResponse, GetBlockByHeightRequest, GetBlockByHeightResponse,
+    GetLatestBlockRequest, GetLatestBlockResponse, GetLatestValidatorSetRequest,
+    GetLatestValidatorSetResponse, GetNodeInfoRequest, GetNodeInfoResponse, GetSyncingRequest,
+    GetSyncingResponse, GetValidatorSetByHeightRequest, GetValidatorSetByHeightResponse,
+};
+use tendermint_rpc::Client;
+use tonic;
+
+use tendermint_rpc;
+
+#[derive(Clone)]
+pub struct TendermintProxyStub {
+    pub tendermint_instance: String,
+}
+
+impl Default for TendermintProxyStub {
+    fn default() -> Self {
+        Self {
+            tendermint_instance: "127.0.0.1:26656".to_string(),
+        }
+    }
+}
+
+#[tonic::async_trait]
+impl TendermintQuery for TendermintProxyStub {
+    async fn get_node_info(
+        &self,
+        request: tonic::Request<GetNodeInfoRequest>,
+    ) -> Result<tonic::Response<GetNodeInfoResponse>, tonic::Status> {
+        let mut client =
+            tendermint_rpc::HttpClient::new(self.tendermint_instance.as_str()).unwrap();
+        let resp = client.status().await.unwrap();
+
+        // Find a way to get from RPC types to proto types.
+        // Enquiring with tendermint-rs devs, there was no clean way to do this in the code
+        // I have reviewed. Might be necessary to either:
+        //      - implement `TryFrom` / `TryInto` upstream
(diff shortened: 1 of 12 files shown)


pd/src/proxy/tendermint.rs
```diff
@@ -0,0 +1,97 @@
...
+        // Enquiring with tendermint-rs devs, there was no clean way to do this in the code
+        // I have reviewed. Might be necessary to either:
+        //      - implement `TryFrom` / `TryInto` upstream
+        //      - get the upstream repo to consolidate types
+
+        // Ok(tonic::Response::new(resp.into()));
+        unimplemented!()
+    }
+
+    #[doc = " GetSyncing queries node syncing."]
```
hdevalence, line 46: Where'd this `doc` attribute come from? Is this pulling comments from the protos somehow?

proto/ibc-go-vendor/cosmos/base/tendermint/v1beta1/query.proto
```diff
@@ -0,0 +1,208 @@
...
+import "tendermint/types/types.proto";
+import "cosmos/base/query/v1beta1/pagination.proto";
+import "cosmos/base/tendermint/v1beta1/types.proto";
+import "cosmos_proto/cosmos.proto";
+import "tendermint/types/block.proto";
+
+option go_package = "github.com/cosmos/cosmos-sdk/client/grpc/tmservice";
+
+// Service defines the gRPC querier service for tendermint queries.
+// TODO(erwan): rename this back to Service, investigate the ramifications of not doing so
```
hdevalence, line 17: What was the reason to rename it `TendermintQuery`? name conflicts with `tower::Service` or something?

proto/ibc-go-vendor/cosmos/base/tendermint/v1beta1/query.proto
```diff
@@ -0,0 +1,208 @@
...
+service TendermintQuery {
+  // GetNodeInfo queries the current node info.
+  rpc GetNodeInfo(GetNodeInfoRequest) returns (GetNodeInfoResponse) {
+    option (google.api.http).get = "/cosmos/base/tendermint/v1beta1/node_info";
```
hdevalence, line 21: No idea what the right answer is here, but should we be annotating these with `cosmos` paths, if we're not doing a cosmos-sdk thing, and thinking of this proto file as something we want to upstream? Or should we just call them `tendermint/blah`?

proto/ibc-go-vendor/cosmos/base/tendermint/v1beta1/query.proto
erwanor: Yes, this is just to reduce overhead during development

pd/src/proxy/tendermint.rs
erwanor: They're artifacts from the tonic codegen, I'll replace them with more descriptive doc comments!

proto/ibc-go-vendor/cosmos/base/tendermint/v1beta1/query.proto
erwanor: Agreed! `tendermint/query/node_info` would work. I'm not sure why the Cosmos SDK has different [names](https://docs.cosmos.network/v0.44/core/proto-docs.html#cosmos/base/tendermint/v1beta1/query.proto) for what I think (?) resolves to a [`Status`](https://docs.tendermint.com/master/spec/rpc/#status).

It'd be tempting to just stick to the spec described in the [Tendermint RPC](https://docs.tendermint.com/master/spec/). I'm still getting up-to-speed about the ecosystem, so I'll defer to you on this one

hdevalence:
Hey! As mentioned on Discord, sorry that this got overrun by the backport from Tendermint 0.35 to 0.34.  It seems like doing this well will require some changes to `tendermint-rs` to add in domain types. 

However, right now, we're not in a good place to "close the loop" on a change cycle with the `tendermint-rs`/`ibc-rs` crates, because the way we accomplished the 0.34 backport was to deeply fork the stack of crates depending on `tendermint-rs`/`ibc-rs`. So before we can start making changes and consuming them, we have to work with upstream to have semver releases that we can consume.

For that reason, I think it'd be better to continue to pause this work, to avoid the possibility of effort getting wasted on conflicting changes -- but thanks very much for starting it.

PR #1243, opened 2022-08-02
Fix consensus key change handling in tendermint updates

zbuc:
Track tendermint validator updates by consensus key, properly handle consensus key changes by setting the old key's voting power to 0

### DIFF
--- a/component/src/stake/component.rs
+++ b/component/src/stake/component.rs
@@ -47,6 +47,10 @@ pub struct Staking {
     /// List of changes to the tendermint validator set accumulated throughout
     /// this block, to be returned during `EndBlock`.
     tm_validator_updates: BTreeMap<PublicKey, u64>,
+    /// List of updated consensus keys tracked during this block. Used to filter
+    /// the `tm_validator_updates` to avoid reporting 0 voting power for new
+    /// consensus keys.
+    new_consensus_keys: Vec<PublicKey>,
 }
 
 impl Staking {
@@ -56,6 +60,7 @@ impl Staking {
             state,
             delegation_changes: Default::default(),
             tm_validator_updates: Default::default(),
+            new_consensus_keys: Default::default(),
         }
     }
 
@@ -559,6 +564,12 @@ impl Staking {
     pub async fn tm_validator_updates(&self) -> Result<Vec<ValidatorUpdate>> {
         let mut updates = Vec::new();
         for (consensus_key, power) in &self.tm_validator_updates {
+            if self.new_consensus_keys.contains(consensus_key) && *power == 0 {
+                // tendermint can't handle receiving 0 voting power for a new
+                // consensus key
+                continue;
+            }
+
             updates.push(ValidatorUpdate {
                 pub_key: *consensus_key,
                 power: (*power).try_into().unwrap(),
@@ -768,9 +779,13 @@ impl Staking {
             // Old consensus key gets 0 power so it drops from the active set.
             self.tm_validator_updates.insert(old_consensus_key, 0);
 
-            // Current consensus key shouldn't be reported to tendermint during the block
-            // it was changed.
-            self.tm_validator_updates.remove(&validator.consensus_key);
+            // Tendermint gets confused if it receives a 0 voting power for
+            // a consensus key it hasn't seen before.
+            //
+            // Since `end_epoch` occurs _after_ this, we need to track the fact
+            // that this validator experienced a consensus key change, to avoid
+            // reporting 0 voting power for a never-before-seen consensus key.
+            self.new_consensus_keys.push(validator.consensus_key);
 
             // Update the consensus key lookup, since the validator rotated their
             // consensus key.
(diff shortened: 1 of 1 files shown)


component/src/stake/component.rs
```diff
@@ -46,7 +46,11 @@ pub struct Staking {
     delegation_changes: DelegationChanges,
     /// List of changes to the tendermint validator set accumulated throughout
     /// this block, to be returned during `EndBlock`.
-    tm_validator_updates: BTreeMap<IdentityKey, u64>,
+    tm_validator_updates: BTreeMap<PublicKey, u64>,
+    /// List of updated consensus keys tracked during this block. Used to filter
+    /// the `tm_validator_updates` to avoid reporting 0 voting power for new
+    /// consensus keys.
+    new_consensus_keys: Vec<PublicKey>,
```
plaidfinch, line 53: I think a `BTreeSet` would be more efficient here, because we repeatedly query for inclusion. That would reduce time complexity for the query loop from quadratic to log-linear.

component/src/stake/component.rs
```diff
@@ -549,18 +562,16 @@ impl Staking {
...
+                // consensus key
+                continue;
+            }
+
             updates.push(ValidatorUpdate {
-                pub_key: validator.consensus_key,
+                pub_key: *consensus_key,
                 power: (*power).try_into().unwrap(),
             });
         }
```
plaidfinch, line 577: We can avoid the allocation here I think, by instead using a `filter_map` on an iterator over the updates, returning an `impl Iterator`.

plaidfinch (commented):
A couple small efficiency tweaks.

We can test if this would avoid the halt on Arche by syncing a node running this branch against it and seeing if it blows up. If it doesn't, I feel pretty good about this.

component/src/stake/component.rs
```diff
@@ -754,14 +765,49 @@ impl Staking {
...
-        self.state
-            .put_domain(
-                state_key::validator_id_by_consensus_key(&validator.consensus_key).into(),
-                validator.identity_key.clone(),
-            )
-            .await;
+        // If the consensus key changed, then the old consensus key's voting power drops
+        // to 0, and the new one isn't recorded (since a newly added validator can
+        // never be active, and Tendermint can't handle receiving a new validator with 0
+        // power).
```
hdevalence, line 771: I'm not sure this is true. What happens when someone rotates the consensus key of a currently active validator?

hdevalence:
I'm worried that this spreads out the logic about handling tendermint validator updates into multiple different places across the staking code. I think it would be better to continue to track validator updates by identity key, and then *also* track whatever information we need to be able to turn those updates into the format that Tendermint expects.

component/src/stake/component.rs
zbuc: The comment is inaccurate, what would happen in that case is that the new consensus key would get a voting power assigned during `end_epoch` (which would occur after the validator definitions are handled), so everything should work out correctly.

zbuc:
@hdevalence 

> I'm worried that this spreads out the logic about handling tendermint validator updates into multiple different places across the staking code.

Are you suggesting to handle this by handling changed validator consensus keys within `tm_validator_updates()` instead?

> I think it would be better to continue to track validator updates by identity key, and then also track whatever information we need to be able to turn those updates into the format that Tendermint expects.

I am not sure. Tendermint's primary key for validator power is the consensus key, and since these updates are formatted specifically for consumption by Tendermint (not for internal `pd` use otherwise), it makes some sense to track them by the consensus key. Otherwise we have to introduce additional complexity to track a history of consensus keys per identity key (i.e. what happens if a validator updates its consensus keys multiple times during a block? As implemented in this PR, that should be handled OK). I'm not sure what the upside would be of the additional complexity necessary to track tendermint validator power updates by identity key instead.

hdevalence:
> I am not sure. Tendermint's primary key for validator power is the consensus key, and since these updates are intended specifically for consumption by Tendermint (not for internal `pd` use otherwise), it makes some sense to track them by the consensus key. Otherwise we have to introduce additional complexity to track a history of consensus keys per identity key (i.e. what happens if a validator updates its consensus keys multiple times during a block? As implemented in this PR, that should be handled OK). I'm not sure what the upside would be of the additional complexity necessary to track tendermint validator power updates by identity key instead.

OK, fair point. Right now, though, various different methods now have to handle consensus keys, and track both identity *and* consensus keys.  For instance, `set_validator_state_inner` now has to fetch the consensus key, and pass it in to other methods, despite not really being concerned with the consensus key.  Also, we have duplicate logic, where there's a special case to handle 0-power updates both in `set_validator_state_inner` and in `update_validator`, and `update_validator` now has its own complex update logic.

It's hard to know what part of the code is responsible for what, and it's important that we keep the state transitions for the validator state machine as simple as possible, to prevent the type of edge cases this change is meant to address.  Perhaps we could encapsulate the tendermint validator updates, rather than just having various parts of the staking component edit them directly?

zbuc:
I like that idea. We could have a state machine struct tracking the tendermint validator powers, implemented with methods somewhat like a Staking sub-component.

PR #1245, opened 2022-08-03
docs: decaf377 spec fixes 

redshiftzero:
Fixes errors in our spec/docs noted during a security audit (report forthcoming)

This also uses an API change that was added in PR: https://github.com/penumbra-zone/decaf377/pull/31


docs/protocol/src/crypto/decaf377/decoding.md
```diff
@@ -1,24 +1,25 @@
...
 
-Decoding to a point works as follows:
+Decoding to a point works as follows where $a$ and $d$ are the curve parameters as described [here](../decaf377.md#curve-parameters).
 
-1. Decode `s_bytes` to a field element $s$, rejecting if the encoding is
-non-canonical.  (If the input is already a field element in the circuit case,
-skip this step).
+1. Decode `s_bytes` to a field element $s$. We interpret these bytes as unsigned
+little-endian bytes. We check if the length has 32 bytes, even though
+the top 3 bits of the last byte are not used. The 253 bits are verified to be canonical, and rejected if not (If the input is already a field element in the circuit case, skip this step).
```
hdevalence, line 7: We probably shouldn't do this, though, right? The new text is maybe a more accurate description of our current behavior, but shouldn't we enforce that the the entire encoding is canonical, and not leave three bits of malleability at the top?

hdevalence (commented):
This looks good except for the encoding point -- I think we probably should have totally canonical field element encodings, even if this requires layering behavior on top of arkworks?

docs/protocol/src/crypto/decaf377/decoding.md
redshiftzero: Ah, yeah this describes what we (well Arkworks) currently do. Should we be requiring that the last three bits are zero?

docs/protocol/src/crypto/decaf377.md
```diff
@@ -22,10 +22,26 @@ construction works the same way inside and outside of a circuit and imposes no
...
+
+The cofactor-4 Edwards curve defined over the BLS12-377 scalar field has the
+following parameters:
+
+* Base field: Integers mod prime $q=8444461749428370424248824938781546531375899335154063827935233455917409239041$
+* Elliptic curve equation: $ax^2 + y^2 = 1 + dx^2y^2$ with $a=-1$ and $d=3021$
+* Curve order: $4r$ where $r=2111115437357092606062206234695386632838870926408408195193685246394721360383$
+
+We use a conventional generator basepoint selected to have a convenient hex encoding: 0x080000000000000000000000000000000
+0000000000000000000000000000000. In affine coordinates this generator point has coordinates:
```
hdevalence, line 35: I think this will get mangled in markdown formatting, since it's split over two lines.  We could do a code block:
```
0x0800000000000000000000000000000000000000000000000000000000000000
```

docs/protocol/src/crypto/decaf377/decoding.md
hdevalence: Yeah, we should keep the written description, and change the implementation to match it, by checking that the top three bits are zero before passing in to Arkworks.

redshiftzero: Added this behavior in https://github.com/penumbra-zone/decaf377/pull/34/commits/c1db5ad55290f62584cd6cc1d5e74ae534cff127 and updated the text here

hdevalence:
LGTM!

Issue #1227, opened 2022-07-28
Rename `ChainParams` to `ChainParameters`


hdevalence:
The abbreviation is ad-hoc and inconsistent with other naming (API consumers have to remember that it's abbreviated by cutting out `eter` and not some other way).

PR #1248, opened 2022-08-05
renamed ChainParams to ChainParameters

borngraced:
I also simplified `if nullifiers.len() == 0` to `if nullifiers.is_empty()`

fixes: #1227

### DIFF
--- a/chain/src/genesis/app_state.rs
+++ b/chain/src/genesis/app_state.rs
@@ -2,14 +2,14 @@ use penumbra_proto::{chain as pb, stake as pb_stake, Protobuf};
 use serde::{Deserialize, Serialize};
 
 use super::Allocation;
-use crate::params::ChainParams;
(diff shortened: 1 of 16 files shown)


plaidfinch:
@borngraced There's a teensy merge conflict in `proto/build.rs`. Could you fix it? Thanks, otherwise this looks good to me!

hdevalence:
@plaidfinch I just rebased it, it should be good to go

plaidfinch:
The smoke test is failing because we merged changes to the JMT which altered its exterior interface, and we're not pinned to a version of the JMT.

hdevalence:
> The smoke test is failing because we merged changes to the JMT which altered its exterior interface, and we're not pinned to a version of the JMT.

Let's roll back the changes to the JMT? We were going to make a plan before actually doing them.

PR #1258, opened 2022-08-09
feat: create initial `FmdParameters` support

### LINKED ISSUE #1226: Add an `FmdParameters` structure that defines the current FMD parameters.

We want the FMD parameters to be auto-adjusted according to the empirical transaction rate on-chain.  Clients need to learn about these parameters somehow, so they can create clues with the correct FMD precision.  

We should add a new proto to `chain.proto` (and corresponding domain type to the `chain` crate):
```proto
message FmdParameters {
    uint32 precision_bits = 1;
}
```

We provide a way to record and query both the current and previous FMD parameters in the chain state:
```rust
// in penumbra_chain::state_key

fn fmd_parameters_current() -> KeyHash {
    "fmd_parameters/current".as_bytes().into()
}
fn fmd_parameters_previous() -> KeyHash {
    "fmd_parameters/previous".as_bytes().into()
}
```
Recording the current parameters will allow the node to check that a transaction was made with respect to the current parameters, and for clients to fetch the current parameters (if they didn't remember them from the compact block, for some reason).  Recording the previous parameters allows the node to have a "grace period" of a few blocks as the parameters change, so that a transaction prepared during the block right before a change isn't rejected because its submission was delayed.

Initially, we could set hardcoded current and previous parameters while processing `init_chain`, and punt on updating them for a later chunk of work. But this way, the chain state will have valid parameters for clients to query.

Next, we should add the `FmdParameters` as an optional field to the `CompactBlock`:
```
message CompactBlock {
    // ...
    FmdParameters fmd_parameters = 100;
    // ...
}
```
Since this field is optional, it should be represented as an `Option<FmdParameters>` in the domain type.

Finally, we need the view service to record the FMD parameters.  This could be done in a completely similar way as the current chain parameter storage.

The previous idea was to add this data to the `ChainParams` directly. But this would mix different kinds of parameters: currently, the `ChainParams` only has top-level, user-defined parameters about the chain, while the FMD parameters would be derived at every epoch.  Instead, we should keep the derived parameters separate from the specified ones.

**Additional context**

In the code above I used `FmdParameters` rather than `FmdParams` on purpose; this is inconsistent with `ChainParams`, but I think that the abbreviation there is a mistake and should be fixed in another issue, independent from this one:

- [x] #1227

redshiftzero:
This doesn't yet include logic for actually updating the parameters, just the initial support, and it adds them to the `CompactBlock`

Closes #1226

### DIFF
--- a/chain/src/params.rs
+++ b/chain/src/params.rs
@@ -157,6 +157,6 @@ impl From<FmdParameters> for pb::FmdParameters {
 
 impl Default for FmdParameters {
     fn default() -> Self {
-        Self { precision_bits: 1 }
+        Self { precision_bits: 0 }
     }
 }
--- a/chain/src/params.rs
+++ b/chain/src/params.rs
@@ -128,3 +128,35 @@ impl Default for ChainParameters {
         }
     }
 }
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(try_from = "pb::FmdParameters", into = "pb::FmdParameters")]
+pub struct FmdParameters {
+    pub precision_bits: u8,
+}
+
+impl Protobuf<pb::FmdParameters> for FmdParameters {}
+
+impl TryFrom<pb::FmdParameters> for FmdParameters {
+    type Error = anyhow::Error;
+
+    fn try_from(msg: pb::FmdParameters) -> Result<Self, Self::Error> {
+        Ok(FmdParameters {
+            precision_bits: msg.precision_bits.try_into()?,
+        })
+    }
+}
+
+impl From<FmdParameters> for pb::FmdParameters {
+    fn from(params: FmdParameters) -> Self {
+        pb::FmdParameters {
+            precision_bits: u32::from(params.precision_bits),
+        }
+    }
+}
+
+impl Default for FmdParameters {
+    fn default() -> Self {
+        Self { precision_bits: 1 }
+    }
+}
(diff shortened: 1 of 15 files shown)


chain/src/params.rs
```diff
@@ -128,3 +128,35 @@ impl Default for ChainParameters {
...
+    fn from(params: FmdParameters) -> Self {
+        pb::FmdParameters {
+            precision_bits: u32::from(params.precision_bits),
+        }
+    }
+}
+
+impl Default for FmdParameters {
+    fn default() -> Self {
+        Self { precision_bits: 1 }
```
hdevalence, line 160: Maybe this should be `0`, so that the `Default` impl for the domain type matches the `Default` impl for the proto type?

chain/src/sync.rs
```diff
@@ -115,6 +119,11 @@ impl From<CompactBlock> for pb::CompactBlock {
                 Some(cb.quarantined.into())
             },
             slashed: cb.slashed.into_iter().map(Into::into).collect(),
+            fmd_parameters: if cb.fmd_parameters.is_some() {
+                Some(cb.fmd_parameters.unwrap().into())
+            } else {
+                None
+            },
```
hdevalence, line 126: I think this could be something like:
```suggestion
            fmd_parameters: cb.fmd_parameters.map(Into::into),
```

view/src/sync.rs
```diff
@@ -50,10 +53,13 @@ pub async fn scan_block(
         epoch_root,
         quarantined,
         slashed,
+        fmd_parameters,
     }: CompactBlock,
     epoch_duration: u64,
     storage: &Storage,
 ) -> anyhow::Result<FilteredBlock> {
+    // TODO: Validation of FMD parameters?
+
```
hdevalence, line 62: What would we be validating?

hdevalence (approved):
LGTM!

PR #1261, opened 2022-08-09
Ics20 withdraw action

### LINKED ISSUE #982: Define a `Withdraw` transaction action
Inbound ICS20 transfers happen when relayers post IBC packets to the chain. Outbound transfers are initiated on Penumbra, and need to be integrated with the Penumbra transaction model.

To do this, we should add `Withdraw` transaction action that specifies:

- a destination chain;
- a transparent value, consisting of an amount + asset id (this implies the denom, which implies the channel/port/etc);
- a destination chain address;
- a "sender" penumbra address to use to return funds;
- a timeout height (on the source [Penumbra] chain ?);
- a timeout time.

This action should consume the value from the transaction's value balance and burn it.

We need to include the sender address to be able to return funds in the event of a transfer failure, but ideally this should be an uncommon case, and ideally outbound transfers should not disclose the user's Penumbra address.  To handle this, we should also

- [x] #983
- [x] #984

and then generate an ephemeral address to put in the `Withdraw` action. We have 11-byte address indices, so we have 2^88 - 2^64 ~ 2^88 possible ephemeral addresses; the birthday bound suggests we're likely to get a collision only after 2^44 IBC withdrawals, which seems fine.  However, from the point of view of the `Withdraw` action, the fact that the address _was_ actually generated this way is irrelevant (since the ephemeral addresses are publicly indistinguishable from normal addresses), so it doesn't affect the transaction action at all.

aubrika:
Closes #982

### DIFF
--- a/proto/proto/ibc.proto
+++ b/proto/proto/ibc.proto
@@ -38,6 +38,31 @@ message IBCAction {
   }
 }
 
+message ICS20Withdrawal {
+  // the chain ID of the destination chain for this ICS20 transfer
+  string destination_chain_id = 1;
+
+  // a transparent value consisting of an amount and an asset ID.
+  crypto.Value value = 2;
+
+  // the address on the destination chain to send the transfer to 
+  string destination_chain_address = 3;
+
+  // a "sender" penumbra address to use to return funds from this withdrawal.
+  // this should be an ephemeral address
+  crypto.Address sender = 5;
+
+  // the height (on Penumbra) at which this transfer expires (and funds are sent
+  // back to the sender address?). NOTE: if funds are sent back to the sender,
+  // we MUST verify a nonexistence proof before accepting the timeout, to
+  // prevent relayer censorship attacks. The core IBC implementation does this
+  // in its handling of validation of timeouts.
+  .ibc.core.client.v1.Height timeout_height = 6;
+
+  // the timestamp at which this transfer expires.
+  uint64 timeout_time = 7;
+}
+
 message ClientData {
   string clientID = 1;
   google.protobuf.Any clientState = 2; // NOTE: left as Any to allow us to add more client types later
--- a/proto/proto/ibc.proto
+++ b/proto/proto/ibc.proto
@@ -8,6 +8,7 @@ import "ibc/core/channel/v1/tx.proto";
 
 import "ibc/core/client/v1/tx.proto";
 import "ibc/core/client/v1/client.proto";
+import "crypto.proto";
 
 import "google/protobuf/any.proto";
 
@@ -50,14 +51,14 @@ message ICS20Withdrawal {
 
   // a "sender" penumbra address to use to return funds from this withdrawal.
   // this should be an ephemeral address
-  crypto.Address sender = 5;
+  crypto.Address return_address = 5;
 
   // the height (on Penumbra) at which this transfer expires (and funds are sent
   // back to the sender address?). NOTE: if funds are sent back to the sender,
   // we MUST verify a nonexistence proof before accepting the timeout, to
   // prevent relayer censorship attacks. The core IBC implementation does this
   // in its handling of validation of timeouts.
-  .ibc.core.client.v1.Height timeout_height = 6;
+  uint64 timeout_height = 6;
 
   // the timestamp at which this transfer expires.
   uint64 timeout_time = 7;
--- a/proto/proto/transaction.proto
+++ b/proto/proto/transaction.proto
@@ -54,16 +54,6 @@ message Action {
     stake.ValidatorDefinition validator_definition = 16;
     ibc.IBCAction ibc_action = 17;
 
-    /*
-    ICS20Withdrawal withdrawal = 18;
-    */
-
-    /*
-    PositionOpen position_open = 30;
-    PositionClose position_close = 31;
-    PositionWithdraw position_withdraw = 32;
-    PositionRewardClaim position_reward_claim = 34;
-    */
     // Governance:
     ProposalSubmit proposal_submit = 18;
     ProposalWithdraw proposal_withdraw = 19;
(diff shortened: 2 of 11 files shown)


avahowell:
LGTM pending those two comments

proto/proto/ibc.proto
```diff
@@ -38,6 +39,31 @@ message IBCAction {
...
+  // a "sender" penumbra address to use to return funds from this withdrawal.
+  // this should be an ephemeral address
+  crypto.Address sender = 5;
+
+  // the height (on Penumbra) at which this transfer expires (and funds are sent
+  // back to the sender address?). NOTE: if funds are sent back to the sender,
+  // we MUST verify a nonexistence proof before accepting the timeout, to
+  // prevent relayer censorship attacks. The core IBC implementation does this
+  // in its handling of validation of timeouts.
+  .ibc.core.client.v1.Height timeout_height = 6;
```
hdevalence, line 61: Elsewhere we use `uint64` heights, is there a reason to use the IBC proto type in our transaction action?

proto/proto/ibc.proto
```diff
@@ -38,6 +39,31 @@ message IBCAction {
...
+
+  // a transparent value consisting of an amount and an asset ID.
+  crypto.Value value = 2;
+
+  // the address on the destination chain to send the transfer to 
+  string destination_chain_address = 3;
+
+  // a "sender" penumbra address to use to return funds from this withdrawal.
+  // this should be an ephemeral address
+  crypto.Address sender = 5;
```
hdevalence, line 54: Since this has a different meaning than normal, we should rename it, perhaps to `return_address`, since that's what it's used for?

proto/proto/transaction.proto
```diff
@@ -52,6 +52,10 @@ message Action {
     stake.ValidatorDefinition validator_definition = 16;
     ibc.IBCAction ibc_action = 17;
 
+    /*
+    ICS20Withdrawal withdrawal = 18;
+    */
```
hdevalence, line 57: We should add in the action now, and reject it until the implementation is ready.  Otherwise we don't have a good way to incrementally implement more and more functionality, since the datatypes aren't there yet.

transaction/src/action.rs
```diff
@@ -44,6 +47,8 @@ pub enum Action {
     PositionClose(PositionClose),
     PositionWithdraw(PositionWithdraw),
     PositionRewardClaim(PositionRewardClaim),
+    // TODO: re-enable when Withdrawal is ready
+    //Withdrawal(ICS20Withdrawal),
```
hdevalence, line 51: ditto https://github.com/penumbra-zone/penumbra/pull/1261/files#r946167248

transaction/src/action.rs
```diff
@@ -169,4 +172,6 @@ impl TryFrom<pb::Action> for Action {
             }
         }
     }
+
+    type Error = anyhow::Error;
```
hdevalence, line 176: probably no particular reason to move this out of the usual place?

transaction/src/action/ibc.rs
```diff
@@ -0,0 +1,77 @@
...
+impl ICS20Withdrawal {
+    pub fn value_commitment(&self) -> value::Commitment {
+        // we need to compute a value commitment to the transparent value of this withdrawal
+        let withdrawal_value = self.value.commit(Fr::zero());
+
+        // consume the withdrawal value. this should be negative because it consumes the value
+        // balance of the transaction. when we compute the binding verification key, we sum all of
+        // the value commitments. thus, there should be a spend that ouputs a positive value
+        // balance to this tx prior to this withdrawal action.
+        -withdrawal_value
```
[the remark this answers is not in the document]
hdevalence: Yes, this is correct -- I'd tweak some of the wording in the comment though: it consumes *from* the transaction's value balance, rather than consuming the value balance itself.

hdevalence (changes requested):
We should add the new withdrawal action into the `Action` enum now, so that it becomes part of our data model.

Because transfers aren't implemented yet, we can reject them in the IBC component, using a similar method as we use to reject LP actions: https://github.com/penumbra-zone/penumbra/blob/main/component/src/dex/component.rs#L31-L44

transaction/src/action.rs
aubrika: I thought so too, however:

```error[E0046]: not all trait items implemented, missing: `Error`
   --> transaction/src/action.rs:136:1
    |
136 | impl TryFrom<pb::Action> for Action {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ missing `Error` in implementation
    |
    = help: implement the missing item: `type Error = Type;`

For more information about this error, try `rustc --explain E0046`.
error: could not compile `penumbra-transaction` due to previous error```

File: proto/proto/ibc.proto (hunk 1 above)
Comment (avahowell) [reply proto/proto/ibc.proto:61]: IBC heights are slightly different- they contain both an epoch and a height. We don't have epochs so the epoch is always 0. Using an IBC height makes it easier to interoperate with other chains, but if this is only going to be used for penumbra heights then we can just use a uint64

File: transaction/Cargo.toml [hunk 7]
```diff
@@ -14,7 +14,15 @@ penumbra-tct = { path = "../tct" }
...
 decaf377-rdsa = { git = "https://github.com/penumbra-zone/decaf377-rdsa" }
-poseidon377 = { git = "https://github.com/penumbra-zone/poseidon377", branch="main" }
+
+# poseidon377 = { git = "https://github.com/penumbra-zone/poseidon377", branch="main" }
+
+# Temporarily pinning poseidon377 until new parameters are audited and the other address-breaking
+# changes are ready to ship.
+poseidon377 = { git = "https://github.com/penumbra-zone/poseidon377", branch="oldparams" }
+ibc = { git = "https://github.com/penumbra-zone/ibc-rs.git", branch = "with-tendermintrs-24" }
+ibc-proto = { git = "https://github.com/penumbra-zone/ibc-rs.git", branch = "with-tendermintrs-24" }
```
hdevalence, line 24: This will need to be rebased on `main` before merging, so it doesn't touch the Poseidon parameters.

hdevalence (approved):
LGTM but these changes need to be rebased on `main` to deconflict the edits to the Poseidon crate.

PR #1263, opened 2022-08-10
Swap/SwapClaim implementations

zbuc:


crypto/src/dex/swap.rs
```diff
@@ -38,6 +38,7 @@ pub fn generate_swap_asset_id(
     delta_1: u64,
     delta_2: u64,
     fee: u64,
+    // TODO: try passing Address type here instead of the parts
```
hdevalence, line 41: We should probably sleep on this until after adding the clue key to the `Note`, so that a `Note` is literally just a typed `Value` plus the `Address` it was sent to.

crypto/src/proofs/transparent.rs
```diff
@@ -737,16 +735,23 @@ impl SwapProof {
...
 
-        if value_2_commitment != -self.value_t2.commit(self.delta_2_blinding) {
-            return Err(anyhow!("value commitment mismatch"));
-        }
+        // if value_2_commitment != -self.value_t2.commit(self.delta_2_blinding) {
+        //     return Err(anyhow!("value commitment mismatch"));
+        // }
 
-        if value_fee_commitment != -self.value_fee.commit(self.fee_blinding) {
+        let pen_denom = asset::REGISTRY.parse_denom("upenumbra").unwrap();
```
hdevalence, line 748: This should use `STAKING_TOKEN_ASSET_ID` instead, to avoid hardcoding the denom.

pcli/src/command/swap.rs
```diff
@@ -0,0 +1,69 @@
...
+        fee: u64,
+        /// Optional. Only spend funds originally received by the given address index.
+        #[clap(long)]
+        source: Option<u64>,
+    },
+    /// Submit a SwapClaim to the chain which will obtain the output amounts for a given Swap NFT.
+    SwapClaim {
+        /// The asset ID of the swap NFT to be claimed.
+        swap_nft_asset_id: String,
+    },
```
hdevalence, line 38: We shouldn't have a separate command for this. Instead, we should claim automatically, by (1) waiting for detection of the note with the swap nft and then (2) immediately building and submitting the swapclaim transaction.

This way, we ensure that the user never has to even know what a swap NFT is, it's purely an implementation detail.

pcli/src/command/swap.rs
```diff
@@ -0,0 +1,69 @@
...
+        asset_1_input_amount: String,
+        /// Asset ID of the second input asset.
+        asset_2_id: String,
+        /// The transaction fee (paid in upenumbra).
+        #[clap(long, default_value = "0")]
+        fee: u64,
+        /// Optional. Only spend funds originally received by the given address index.
+        #[clap(long)]
+        source: Option<u64>,
+    },
```
hdevalence, line 33: I think this would be better as `pcli tx swap`, i.e., a `tx` subcommand, like `send`, rather than a standalone subcommand.

If we do that, and claim automatically (see below), we won't need to implement a whole subcommand.

xref #1276

proto/build.rs
```diff
@@ -251,6 +258,26 @@ static FIELD_ATTRIBUTES: &[(&str, &str)] = &[
...
+        ".penumbra.transaction.SwapClaimPlan.note_blinding",
+        AS_HEX_FOR_BYTES,
+    ),
+    (
+        ".penumbra.transaction.SwapClaimPlan.value_blinding",
+        AS_HEX_FOR_BYTES,
+    ),
+    (".penumbra.transaction.SwapClaimPlan.esk", AS_HEX_FOR_BYTES),
+    // TODO: replace if we use UTF-8 memos
+    (".penumbra.transaction.SwapClaimPlan.memo", AS_HEX_FOR_BYTES),
```
hdevalence, line 280: Wait, why does a `SwapClaim` have a memo at all?

transaction/src/action.rs
```diff
@@ -55,9 +54,10 @@ impl Action {
             Action::Spend(spend) => spend.body.value_commitment,
             Action::Delegate(delegate) => delegate.value_commitment(),
             Action::Undelegate(undelegate) => undelegate.value_commitment(),
-            // TODO: re-enable when Swap/SwapClaim is ready
-            // Action::Swap(swap) => swap.value_commitment(),
-            // Action::SwapClaim(swap_claim) => swap_claim.value_commitment(),
+            Action::Swap(swap) => swap.value_commitment(),
+            // A SwapClaim action has no impact on the transaction's value balance as the
+            // outputs are handled internally to the action.
+            Action::SwapClaim(_) => value::Commitment::default(),
```
hdevalence, line 60: I don't think this is quite right -- the `SwapClaim` *contributes* the fee amount that was prepaid in the `Swap`, so that it can be consumed by the transaction's fee.

proto/build.rs
zbuc: this is outdated/copied from `Output` and needs removal, this PR is still a draft

transaction/src/action/swap_claim.rs
```diff
@@ -2,14 +2,14 @@ use penumbra_crypto::dex::BatchSwapOutputData;
...
+use penumbra_crypto::{proofs::transparent::SwapClaimProof, NotePayload};
 use penumbra_proto::{dex as pb, Protobuf};
 use penumbra_tct as tct;
 
 #[derive(Debug, Clone)]
 pub struct SwapClaim {
-    zkproof: OutputProof,
-    body: Body,
+    zkproof: SwapClaimProof,
+    pub body: Body,
```
hdevalence, line 12: We should make both of these `pub` to match the other transaction structures.

transaction/src/auth_hash.rs
```diff
@@ -213,6 +215,57 @@ impl spend::Body {
...
+    fn auth_hash(&self) -> Hash {
+        let mut state = blake2b_simd::Params::default()
+            .personal(b"PAH:swap_body")
+            .to_state();
+
+        // All of these fields are fixed-length, so we can just throw them
+        // in the hash one after the other.
+        // TODO: actually the trading pair isn't necessarily fixed-length
+        // right now, does this have implications?
+        state.update(&self.trading_pair.encode_to_vec());
```
hdevalence, line 228: The simplest thing here is probably to hash the assets in the pair in order.

transaction/src/auth_hash.rs
```diff
@@ -213,6 +215,57 @@ impl spend::Body {
...
+        state.update(&self.output_1.ephemeral_key.0);
+        state.update(&self.output_1.encrypted_note);
+        state.update(&self.output_2.note_commitment.0.to_bytes());
+        state.update(&self.output_2.ephemeral_key.0);
+        state.update(&self.output_2.encrypted_note);
+        state.update(&self.output_data.encode_to_vec());
+        state.update(&self.anchor.encode_to_vec());
+        // TODO: actually the trading pair isn't necessarily fixed-length
+        // right now, does this have implications?
+        state.update(&self.trading_pair.encode_to_vec());
```
hdevalence, line 263: ditto hashing asset ids in order.

hdevalence:
This is looking really good!

It's getting pretty big, so it'd probably be good to get it into an intermediate shape to merge in the next day or so.

One thing I'm wondering about is how the encryption of the swap plaintexts and note payloads interacts with the changes @redshiftzero made in #1269 to streamline and commonize the symmetric crypto -- should we rebase this on `main` to pick up those changes?

transaction/src/auth_hash.rs
zbuc: There's a `to_bytes` impl that results in a 64-byte fixed length encoding, that should be suitable? We could also add an `auth_hash` method that hashed that but I'm not sure why we'd want to.

zbuc: Ah but the `to_bytes` impl is only `pub(crate)` -- maybe it does make sense to add `auth_hash` then

Issue #1264, opened 2022-08-11
clean up and finalize symmetric crypto functionality


redshiftzero:
- [x] Ensure we use common code paths for deriving symmetric keys for memo, note, and swap encryption (I believe we do for memo and notes): https://github.com/penumbra-zone/penumbra/issues/1265
- [x] Add Ock type: `OutgoingCipherKey`: https://github.com/penumbra-zone/penumbra/issues/1267
- [x] Generate a better name for the return value of `Note::encrypt_key` and add that struct
- [x] Name the derived symmetric key e,g, `PayloadKey` (line 138 of `note.rs`)
- [x] Provide some sort of interface to get the `PayloadKey` - this is a more fine grained functionality than the viewing functionality since it would allow viewing a single action
- [x] Collecting the nonces together in one place (for swap, memo, note)
- [x] https://github.com/penumbra-zone/penumbra/issues/1268
- [x] Consider encrypting the shared secret directly instead of the transmission key and ephemeral secret
- [ ] Document how the transaction key hierarchy works in the spec

Note that this is closely related to the memo change (#1222)

redshiftzero:
The `SharedSecret` used to derive the `PayloadKey` (with the ephemeral public key) can be derived from the transmission key and the ephemeral private key. The `OvkWrappedKey` saves the transmission key (32 bytes) and the esk (32 bytes). We could instead encrypt directly (as the `OvkWrappedKey`) either the `SharedSecret` (32 bytes) or the `PayloadKey` (32 bytes), either option saves 32 bytes and a DH key exchange. I’ve tacked that change onto PR #1284.

PR #1269, opened 2022-08-12
crypto: add `PayloadKey` and use for notes, memos, and swaps

### LINKED ISSUE #1268: Use the same ephemeral symmetric key for note and swap encryption, with different nonces

redshiftzero:
This has two of the items towards #1264. It adds a `symmetric` module that contains a new `PayloadKey` to represent the symmetric key used to encrypt/decrypt notes, memos and swaps

Closes #1268

### DIFF
--- a/crypto/src/dex/swap/plaintext.rs
+++ b/crypto/src/dex/swap/plaintext.rs
@@ -8,12 +8,12 @@ use chacha20poly1305::{
 use penumbra_proto::{crypto as pb_crypto, dex as pb, Protobuf};
 
 use crate::dex::TradingPair;
-use crate::keys::OutgoingViewingKey;
-
-use super::{
-    SwapCiphertext, OVK_WRAPPED_LEN_BYTES, SWAP_CIPHERTEXT_BYTES, SWAP_ENCRYPTION_NONCE,
-    SWAP_LEN_BYTES,
+use crate::{
+    keys::OutgoingViewingKey,
+    symmetric::{PayloadKey, SWAP_ENCRYPTION_NONCE},
 };
+
+use super::{SwapCiphertext, OVK_WRAPPED_LEN_BYTES, SWAP_CIPHERTEXT_BYTES, SWAP_LEN_BYTES};
 
 #[derive(Clone)]
 pub struct SwapPlaintext {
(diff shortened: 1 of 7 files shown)


crypto/src/dex/swap/plaintext.rs
```diff
@@ -37,18 +37,6 @@ impl SwapPlaintext {
     // Theoretically, if a paranoid user did want to achieve forward secrecy, they could choose to encrypt
     // nonsense bytes as the swap plaintext as the swap ciphertext does not need to be valid for the
     // swap to succeed, however this is unsupported by the official client.
-    pub(super) fn derive_symmetric_key(
```
hdevalence, line 40: The comment should disappear along with the function, right?

hdevalence, line 40: This comment should disappear along with the function, right?

crypto/src/symmetric.rs
```diff
@@ -0,0 +1,59 @@
...
+pub static NOTE_ENCRYPTION_NONCE: Lazy<[u8; 12]> = Lazy::new(|| [0u8; 12]);
+
+/// The nonce used for memo encryption.
+pub static MEMO_ENCRYPTION_NONCE: Lazy<[u8; 12]> = Lazy::new(|| {
+    let nonce_bytes = 1u128.to_le_bytes();
+    nonce_bytes[0..12].try_into().expect("nonce fits in array")
+});
+
+/// The nonce used for swap encryption.
+pub static SWAP_ENCRYPTION_NONCE: Lazy<[u8; 12]> = Lazy::new(|| [9u8; 12]);
```
hdevalence, line 20: Two comments here:

1. Because we're already specifying static `[u8; 12]` values, we're not really getting much from using a `Lazy` here, just overhead.  We could skip it by making the constant be of type `[u8; 12]` directly and writing out the arrays.
2. The current encryption/decryption API takes the nonce as bytes. But since we know that we want to be using specific nonces for specific types of data, I'm wondering if we should reshape the API to avoid the byte arrays entirely?

This could look like an
```rust
enum PayloadKind {
    Note,
    Memo,
    Swap,
}
```
with
```rust
impl PayloadKind {
    fn nonce(&self) -> [u8; 12] {
        match self {
            Self::Note => [0,0,0,0,0,0,0,0,0,0,0,0],
            Self::Memo => [1,0,0,0,0,0,0,0,0,0,0,0],
            Self::Swap => [2,0,0,0,0,0,0,0,0,0,0,0],
        }
    }
}
```
so that the `PayloadKey` methods can take a `PayloadKind` and call `kind.nonce()` internally to get the bytes.  Then callers don't need to know anything about the bytes, or import a bunch of different constants, they just need to have the `PayloadKind`.  It also makes the safety rule more coherent: each `PayloadKey` can only be used to encrypt at most one object of each kind.

hdevalence (approved):
This is a great improvement, I think it could be even cleaner if we eliminated the nonce handling for the caller by factoring it out into a `PayloadKind` enum as in the comment, but this could be done in a followup PR.

crypto/src/symmetric.rs
redshiftzero: thanks for this suggestion! This is way better that what I was doing so I've added a `PayloadKind` into this PR

PR #1270, opened 2022-08-12
fmd: add `Vec<Clue>` to the `TransactionBody`

redshiftzero:
The next item towards #1039


decaf377-fmd/Cargo.toml
```diff
@@ -6,13 +6,18 @@ edition = "2018"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+penumbra-proto = { path = "../proto/" }
+
+anyhow = "1"
+bytes = "1"
```
hdevalence, line 12: Hmm, the idea/hope was that the FMD library would be more general than just Penumbra, so having it depend on `penumbra-proto` is less than ideal.  Since the reason to have the dependency is just to be able to implement the `Protobuf` trait, we could avoid it using the same method that we use for `decaf377-rdsa`: make the `penumbra-proto` crate depend on the foreign `decaf377-fmd` crate, and then put the `Protobuf` impls for the `decaf377-fmd` types alongside the `decaf377-rdsa` ones.

proto/proto/transaction.proto
```diff
@@ -22,6 +22,11 @@ message Transaction {
   crypto.MerkleRoot anchor = 3;
 }
 
+// A clue for use with Fuzzy Message Detection.
+message Clue {
+  bytes inner = 1;
+}
+
```
hdevalence, line 29: All the proto organization is a mess (let's clean it up after we get feature complete) but this probably belongs in `crypto.proto`, similar to the signature protos.

hdevalence (changes requested):
This looks good, but we should do some organizational changes before merging.

redshiftzero:
thanks for the feedback! I've addressed these changes

hdevalence (approved):
LGTM!

PR #1274, opened 2022-08-16
Skeleton for Swap/SwapClaim `pcli` commands

zbuc:
This PR adds the basic skeleton of the `pcli swap` subcommand.


hdevalence:
Looks like this was bundled into #1263? I think we should avoid a new command here, and just have `tx swap` that claims the swap outputs automatically.

zbuc:
It was small and self-contained so I cherry-picked it out into a new PR

PR #1279, opened 2022-08-17
crypto: add `OutgoingCipherKey`

### LINKED ISSUE #1265: Share `derive_symmetric_key` between notes and swaps

### LINKED ISSUE #1267: Create common `ock` type between notes and swaps

redshiftzero:
Closes #1265
Closes #1267
Towards #1264

### DIFF
--- a/crypto/src/dex/swap/plaintext.rs
+++ b/crypto/src/dex/swap/plaintext.rs
@@ -14,7 +14,7 @@ use crate::{
 
 use super::{SwapCiphertext, OVK_WRAPPED_LEN_BYTES, SWAP_CIPHERTEXT_BYTES, SWAP_LEN_BYTES};
 
-#[derive(Clone)]
+#[derive(Clone, Debug, PartialEq, Eq)]
 pub struct SwapPlaintext {
     // Trading pair for the swap
     pub trading_pair: TradingPair,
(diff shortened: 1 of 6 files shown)


crypto/src/dex/swap/plaintext.rs
```diff
@@ -38,57 +37,22 @@ impl SwapPlaintext {
...
-    /// Use Blake2b-256 to derive an encryption key `ock` from the OVK and public fields.
-    pub fn derive_ock(ovk: &OutgoingViewingKey, epk: &ka::Public) -> blake2b_simd::Hash {
-        // let cv_bytes: [u8; 32] = cv.into();
-        // let cm_bytes: [u8; 32] = cm.into();
-
-        let mut kdf_params = blake2b_simd::Params::new();
-        kdf_params.hash_length(32);
-        let mut kdf = kdf_params.to_state();
-        kdf.update(&ovk.0);
-        // TODO: should we be using the public fields e.g. t1, t2, trading_pair here?
```
redshiftzero, line 50: Note that by using the new shared `OutgoingCipherKey` logic instead of this method, I'm requiring a note and value commitment, which for swaps I'm assuming would come from the note corresponding to the swap - does this make sense to you @zbuc?

hdevalence: I think this fits with the discussions we've had about wanting to ensure that visibility for a `SwapPlaintext` is exactly the same as visibility for the note that records the swap NFT.

hdevalence (approved):

Issue #621, opened 2022-04-17
repo cleanup: move deployment-related files into a `deploy/` subdirectory


hdevalence:
**Is your feature request related to a problem? Please describe.**

Currently the repo root is quite "wide", and there are a lot of files sitting in the root. Some of these are related to deployments, and it'd be nice to bundle them into a `deploy/` subdirectory.

**Describe the solution you'd like**

(Using `/` to indicate the repo root, not the filesystem root)

- [ ] Create a `/deploy/` subdirectory;
- [ ] Move `/config` to `/deploy/config`;
- [ ] Move the `/Dockerfile*`, `/docker-compose*`, `/.dockerignore` out of the repo root into a new `/deploy/docker/` subfolder;
- [ ] Move `/scripts` to `/deploy/scripts`;
- [ ] Patch up files so that it's OK for them to not be in the repo root (does this mess up the existing Docker configs, or do we just have to use them differently?)

The moving stuff around part is pretty simple, and I'd otherwise just do it, but I'm not sure whether moving all of the docker-related stuff around will mess it up (cc @zbuc ?)

hdevalence:
The specific plan is probably out of date since last April, but it'd still make sense to clean up deployment and config files into a subdirectory.

PR #1281, opened 2022-08-17
Initial Directory Structure for testnet deployments.

danbryan:
Proposal for structuring deployments.

- Terraform modules in `deployments/terraform/modules/`
- Testnet deployment state in `deployments/networks/`


hdevalence:
xref https://github.com/penumbra-zone/penumbra/issues/621 -- we've wanted to do a bit of cleanup in this direction for a while, and this would have basically the same structure. Looks great!

PR #1282, opened 2022-08-17
feat: add full `Address` to `Note`

### LINKED ISSUE #1084: FMD: include clue key (i.e. full `Address`) in notes
Related to #1039, we should include a clue key in the `Note`.

redshiftzero:
Closes #1084

~Marked wip as this still needs the changes propagating to the view service to either add a column storing the clue key (probably I'll do this first option) or to store the full address somewhere~ (done, by having a single address column)

### DIFF
--- a/chain/src/genesis/allocation.rs
+++ b/chain/src/genesis/allocation.rs
@@ -61,8 +61,7 @@ impl Allocation {
     /// already public.
     pub fn note(&self) -> Result<Note, anyhow::Error> {
         Note::from_parts(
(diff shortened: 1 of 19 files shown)


hdevalence:
Hmm, it seems like it would be simpler to remove columns, and instead of having separate columns for each address component (diversifier / transmission key / clue key), to just have one column, storing the address itself.

hdevalence:
xref also to #1263 -- ideally we could get a version of #1263 merged in soon, in order to avoid further conflicts

PR #1296, opened 2022-08-18
add deterministic `Clue` generation and `CluePlan` [FMD part 1]

### LINKED ISSUE #1278: Add clues to the `TransactionPlan`
Before we can start enforcing the consensus rule in #1277, we need to add clues into the `TransactionPlan`.

There are two parts to this:

- [x] Change the `decaf377-fmd` crate to allow deterministic creation of clues;
- [x] Add "clue plans" to the `TransactionPlan`.

For deterministic creation of clues, we want to replace the existing [`create_clue`](https://rustdoc.penumbra.zone/main/decaf377_fmd/struct.ExpandedClueKey.html#method.create_clue) method with a new
```
pub fn create_clue_deterministic(
    &self,
    precision_bits: usize,
    rseed: [u8; 32],
) -> Result<Clue, Error>
```
and make the existing `create_clue` become a wrapper for this method that uses the `rng` to generate `rseed`.  Currently, the RNG is used to generate field elements `r` and `z`; instead, these should be generated by using `rseed` as the key for a BLAKE2 PRF (sort of like in the other key derivation steps, maybe we could hash the compact clue key in there and call it synthetic nonces like XEd25519).

One design question with the clue plans is whether we should try to distinguish dummy addresses, to allow the custodian to know with certainty that the dummy addresses really are dummy addresses.

One way to do this would be to make something like (sketch):
```
enum CluePlan {
    Real {
        address: Address,
        rseed: [u8; 32],
    },
    Dummy {
        dummy_spend_key: SpendKeyBytes, // used to derive address
        rseed: [u8; 32],
    },
}
```
but this doesn't really ensure that the `dummy_spend_key` isn't for some other address (although the stakes for sharing it are higher, I guess).

On the other hand, a malicious prover could compromise detection much more simply (they can just tweet out the transaction! or the FVK!). So my feeling is that this is probably overkill, and instead, we should just make a
```
struct CluePlan {
     address: Address,
     rseed: [u8; 32],
}
```

redshiftzero:
Closes #1278

### DIFF
--- a/decaf377-fmd/src/clue_key.rs
+++ b/decaf377-fmd/src/clue_key.rs
@@ -1,6 +1,6 @@
 use std::{cell::RefCell, convert::TryFrom};
 
-use ark_ff::{Field, UniformRand};
+use ark_ff::{Field, PrimeField, UniformRand};
 use bitvec::{array::BitArray, order};
 use decaf377::{FieldExt, Fr};
 use rand_core::{CryptoRng, RngCore};
@@ -74,7 +74,8 @@ impl ExpandedClueKey {
     }
 
     /// Create a [`Clue`] intended for the [`DetectionKey`](crate::DetectionKey)
-    /// corresponding to this clue key.
+    /// corresponding to this clue key, deterministically, using the provided
+    /// random seed.
     ///
     /// The clue will be detected by the intended detection key with probability
     /// 1, and by other detection keys with probability `2^{-precision_bits}`.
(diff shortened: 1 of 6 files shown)


decaf377-fmd/src/clue_key.rs
```diff
@@ -73,8 +73,9 @@ impl ExpandedClueKey {
         Ok(())
     }
 
-    /// Create a [`Clue`] intended for the [`DetectionKey`](crate::DetectionKey)
-    /// corresponding to this clue key.
+    /// Create a [`Clue`] deterministically. Intended for the
+    /// [`DetectionKey`](crate::DetectionKey) corresponding to this clue key based
+    /// on a random seed.
```
hdevalence, line 78: I think this would be clearer phrased as:

> Create a [`Clue`] intended for the [`DetectionKey`] corresponding to this clue key, deterministically, using the provided random seed.

hdevalence (approved):
LGTM modulo the wording tweak in the docs!

PR #1298, opened 2022-08-18
Group `pcli` command line options roughly by capability

### LINKED ISSUE #1276: Refactor `pcli` CLI design
The `pcli` design has gotten a little unwieldy and there's been some discussion about refactoring it for better usability. Initial thoughts based on conversation with @hdevalence:

We might consider separating subcommands not by functional area, but rather by capability, i.e.:

- `pcli q` queries public state (no key needed)
- `pcli view` queries private state (full viewing key needed)
- `pcli tx` executes transactions (spend key needed)

This should ideally cover most of the functionality we'd like, but there's some functionality we might want to be separate:

- `pcli validator` does validator-related things (validator key needed)
- `pcli wallet` manages keys (should this be renamed? "wallet" isn't great)

This work interacts with @zbuc's work on swap subcommands, so we should coordinate on a design here, implement it in a separate PR, and then rebase our respective work on top of that new structure.

plaidfinch:
Resolves #1276.


hdevalence:
This looks great, thanks!!

I did another edit pass, filled in docs, grouped commands by functionality (using `display_order`), and added aliases (`q`, `v`, `tx`).

hdevalence (approved):

PR #1299, opened 2022-08-18
add ics20 transfer packet processing

### LINKED ISSUE #978: Process transfer packets from counterparty chains
This chunk of work consists of adding functionality to the ICS20 component that receives transfer packets from counterparty chains, verifies them against the counterparty chain state, and acknowledges receipt.

avahowell:
Co-Authored-By: aubrey@penumbra.zone

Closes #978

### DIFF
--- a/component/src/ibc/transfer.rs
+++ b/component/src/ibc/transfer.rs
@@ -125,7 +125,32 @@ impl AppHandlerCheck for ICS20Transfer {
 
         Ok(())
     }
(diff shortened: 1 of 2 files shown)


hdevalence:
is this blocked on anything?

avahowell:
I don't think so, but for this to work end-to-end we need to decide on #1336. I think we can optimistically merge this and then build on top of it

hdevalence:
I don't think we need to decide on #1336 now (or even "soon"), since it's only a problem in practice for bridged ETH assets, bridged cosmos assets will be fine, so let's merge this to continue working on it.

PR #1305, opened 2022-08-22
Measure voting power in terms of staking token

### LINKED ISSUE #1280: staking: measure voting power in units of staking token
Per @hdevalence:

> the voting power is measured in units of "stake that has been bonded from genesis with 0 commission and no slashing" ... we should change the math so that we measure voting power just in units of "staking token"

Currently, voting power is calculated as:

https://github.com/penumbra-zone/penumbra/blob/0f099676441885eecc0fcba7bc3b71e00c7c403d/component/src/stake/rate.rs#L131-L138

I think what we want is:

```rust
 pub fn voting_power(&self, total_delegation_tokens: u64) -> u64 { 
     total_delegation_tokens
        .checked_mul(self.validator_exchange_rate)
        .expect("voting power in possible range")
 }
 ```

In other words, removing the denominator of `base_rate_data.base_exchange_rate` means that we would no longer be scaling all voting power by the (global) base exchange rate for a "perfect validator" who has existed since genesis.

In the above, the effect of `u64::checked_mul` is the same as casting up to `u128` and then checking on the downcast after the multiplication, since we don't have an intermediate division.

plaidfinch:
Resolves #1280.

### DIFF
--- a/component/src/stake/component.rs
+++ b/component/src/stake/component.rs
@@ -400,8 +400,7 @@ impl Staking {
                 .expect("delegation token should be known");
 
             // Calculate the voting power in the newly beginning epoch
-            let voting_power =
-                current_rate.voting_power(delegation_token_supply, &current_base_rate);
+            let voting_power = current_rate.voting_power(delegation_token_supply);
             tracing::debug!(?voting_power);
 
             // Update the state of the validator within the validator set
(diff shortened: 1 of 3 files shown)


redshiftzero:
After discussion in sprint planning, closing for now

Issue #1302, opened 2022-08-22
Current `main` has many warnings


hdevalence:
**Describe the bug**

We've accumulated a large number of warnings on current `main`.

This will be confusing for testnet users.

We should fix all the warnings before we release the Hermippe testnet tomorrow.

PR #1308, opened 2022-08-22
Support generating plans for Swap/SwapClaim in wallet

zbuc:
This adds initial support for generating `SwapPlan`s/`SwapClaimPlan`s in the wallet.

Only `SwapPlan` is fully implemented, as `SwapClaimPlan` requires the ability to retrieve clearing prices from the dex.

### DIFF
--- a/pcli/src/command/tx.rs
+++ b/pcli/src/command/tx.rs
@@ -169,10 +169,23 @@ impl TxCmd {
                     tokio::time::sleep(std::time::Duration::from_secs(6)).await;
                 }
             },
-            TxCmd::Swap { input, into, .. } => {
-                let _input = input.parse::<Value>()?;
-                let _into = asset::REGISTRY.parse_unit(into.as_str());
-                println!("Sorry, this command is not yet implemented");
+            TxCmd::Swap {
+                input,
+                into,
+                fee,
+                source,
+            } => {
+                let input = input.parse::<Value>()?;
+                // TODO: should we swap with specific units or the base denom only?
+                // This is unclear until the dex is further along.
+                let into = asset::REGISTRY.parse_unit(into.as_str()).base();
+
+                let plan =
+                    plan::swap(&app.fvk, &mut app.view, OsRng, input, into, *fee, *source).await?;
+
+                app.build_and_submit_transaction(plan).await?;
+
+                // TODO: wait until the swap is confirmed and then perform a SwapClaim automatically
             }
             TxCmd::Delegate {
                 to,
--- a/wallet/src/plan.rs
+++ b/wallet/src/plan.rs
@@ -1,3 +1,4 @@
+use rand_core::OsRng;
 use std::collections::{BTreeMap, HashMap};
 
 use anyhow::Result;
@@ -442,6 +443,9 @@ where
         return Ok(plan);
     }
 
+    // Use a random ephemeral address for claiming the swap.
+    let (claim_address, _dtk) = fvk.incoming().ephemeral_address(OsRng);
+
     // Add a `SwapPlan` action:
     plan.actions.push(
         SwapPlan::new(
@@ -450,9 +454,7 @@ where
             delta_1,
             delta_2,
             Fee(fee),
-            // The `fvk` is always the claim address.
-            // TODO: this should probably select a random address index.
-            fvk.incoming().payment_address(0u64.into()).0,
+            claim_address,
         )
         .into(),
     );
--- a/wallet/src/plan.rs
+++ b/wallet/src/plan.rs
@@ -4,11 +4,14 @@ use anyhow::Result;
 use penumbra_component::stake::rate::RateData;
 use penumbra_component::stake::validator;
 use penumbra_crypto::{
-    asset::Denom, keys::AddressIndex, memo::MemoPlaintext, transaction::Fee, Address,
-    DelegationToken, FullViewingKey, Value, STAKING_TOKEN_ASSET_ID, STAKING_TOKEN_DENOM,
+    asset::Denom, dex::TradingPair, keys::AddressIndex, memo::MemoPlaintext, transaction::Fee,
+    Address, DelegationToken, FullViewingKey, Note, Value, STAKING_TOKEN_ASSET_ID,
+    STAKING_TOKEN_DENOM,
 };
 use penumbra_proto::view::NotesRequest;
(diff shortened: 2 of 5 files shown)


wallet/src/plan.rs
```diff
@@ -1,14 +1,18 @@
...
-    asset::Denom, keys::AddressIndex, memo::MemoPlaintext, transaction::Fee, Address,
-    DelegationToken, FullViewingKey, Value, STAKING_TOKEN_ASSET_ID, STAKING_TOKEN_DENOM,
+    asset::Denom, dex::TradingPair, keys::AddressIndex, memo::MemoPlaintext, transaction::Fee,
+    Address, DelegationToken, FullViewingKey, Note, Value, STAKING_TOKEN_ASSET_ID,
+    STAKING_TOKEN_DENOM,
 };
 use penumbra_proto::view::NotesRequest;
-use penumbra_transaction::plan::{ActionPlan, OutputPlan, SpendPlan, TransactionPlan};
+use penumbra_transaction::plan::{
+    ActionPlan, OutputPlan, SpendPlan, SwapClaimPlan, SwapPlan, TransactionPlan,
```
redshiftzero, line 14: Nit: it would be good to try to resolve warnings at least before the next testnet (else we should do a pass e.g. like #1302 to avoid confusing testnet participants)

wallet/src/plan.rs
```diff
@@ -264,6 +268,272 @@ where
...
+                amount_to_spend: spend_amount,
+                include_spent: false,
+            })
+            .await?;
+        if notes_to_spend.is_empty() {
+            // Shouldn't happen because the other side checks this, but just in case...
+            return Err(anyhow::anyhow!("not enough notes to spend",));
+        }
+
+        let change_address_index: u64 = fvk
```
redshiftzero, line 348: Ah right, for the SwapClaim, we spend a note with 1 NFT _plus_ we need to spend enough notes to cover fees, which can result in change. I wonder if there's a nice way to factor out this change logic since it's largely the same afaict with the `swap` function (could be done in a followup / filed as an issue).

wallet/src/plan.rs
```diff
@@ -264,6 +268,272 @@ where
...
+    // Add a `SwapClaimPlan` action:
+    // plan.actions.push(
+    //     SwapClaimPlan::new(
+    //         &mut rng,
+    //         swap_nft_note,
+    //         swap_nft_position,
+    //         // The `fvk` is always the claim address.
+    //         fvk.address(),
+    //         fee,
+    //         output_data,
```
redshiftzero, line 310: Ah, so this is commented out because `output_data` is of type `BatchSwapOutputData` which we can't generate yet 👍

wallet/src/plan.rs
```diff
@@ -264,6 +268,272 @@ where
...
+    //         anchor,
+    //         trading_pair,
+    //     )
+    //     .into(),
+    // );
+
+    // The value we need to spend is 1 unit of the swap NFT, plus fees.
+    let mut value_to_spend: HashMap<Denom, u64> = HashMap::new();
+    // *value_to_spend.entry(swap_nft_note.denom()).or_default() += input_value;
```
redshiftzero, line 319: Commented out because `denom()` on the NFT note is not a thing yet?

redshiftzero (commented):
pretty cool! dropped some comments inline

wallet/src/plan.rs
```diff
@@ -264,6 +268,272 @@ where
...
+        0
+    };
+    let delta_2 = if trading_pair.asset_1() == input_value.asset_id {
+        0
+    } else {
+        input_value.amount
+    };
+
+    // If there is no input, then there is no swap.
+    if delta_1 == 0 && delta_1 == 0 {
```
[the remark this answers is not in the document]
zbuc: yes, good catch, thank you

wallet/src/plan.rs
zbuc: I'm going to push a second PR today/tomorrow to create a "failed" BatchSwapOutputData (basically representing that there wasn't enough liquidity available). I'll probably do so by creating the basic clearing price checking API.

redshiftzero: ah cool!

wallet/src/plan.rs
zbuc: Hrm, this isn't quite right for `SwapClaim`, because the fee is actually pre-paid during the `Swap` and the `SwapClaim` doesn't require a spend authorization (but wait, how do you spend the swap NFT without a spend auth?). @hdevalence

wallet/src/plan.rs
zbuc: Commented out because it wasn't clear how to implement `SwapClaim` here just yet

pcli/src/command/tx.rs
```diff
@@ -169,10 +169,23 @@ impl TxCmd {
...
+            TxCmd::Swap {
+                input,
+                into,
+                fee,
+                source,
+            } => {
+                let input = input.parse::<Value>()?;
+                // TODO: should we swap with specific units or the base denom only?
+                // This is unclear until the dex is further along.
+                let into = asset::REGISTRY.parse_unit(into.as_str()).base();
```
hdevalence, line 181: I think this change (grabbing the base denom) makes it have the correct behavior -- if I specify `--into penumbra`, for instance, I'll end up with a note recording `upenumbra`, the base denom, which is what I want.

pcli/src/command/tx.rs
```diff
@@ -169,10 +169,23 @@ impl TxCmd {
...
+
+                let plan =
+                    plan::swap(&app.fvk, &mut app.view, OsRng, input, into, *fee, *source).await?;
+
+                app.build_and_submit_transaction(plan).await?;
+
+                // TODO: wait until the swap is confirmed and then perform a SwapClaim automatically
```
hdevalence, line 188: The previous `App::submit_transaction` API would have been perfect for this, because it allowed the caller to await detection of an output note. Now it only allows the caller to await detection of a nullifier.  Restoring the output note detection functionality (another `Option`al argument?) would let this code grab hold of the swap NFT immediately.

wallet/src/plan.rs
hdevalence: spending the swap NFT is "internalized" into the `SwapClaim` action, so there shouldn't be anything to do other than construct a `SwapClaimPlan`.

PR #1312, opened 2022-08-24
build `Clue`s from `CluePlan`s [FMD part 2]

redshiftzero:
Towards #1039, adding FMD clues into transactions.
This is the next step after https://github.com/penumbra-zone/penumbra/pull/1296 which adds support deterministic clues and the `CluePlan` field in the transaction body.

### DIFF
--- a/transaction/src/plan/build.rs
+++ b/transaction/src/plan/build.rs
@@ -12,7 +12,8 @@ impl TransactionPlan {
     ///
     /// - `fvk`, the [`FullViewingKey`] for the source funds;
     /// - `auth_data`, the [`AuthorizationData`] authorizing the transaction;
-    /// - `witness_data`, the [`WitnessData`] used for proving.
+    /// - `witness_data`, the [`WitnessData`] used for proving;
+    /// - `precision_bits`, the current FMD precision.
     ///
     pub fn build<R: CryptoRng + RngCore>(
         self,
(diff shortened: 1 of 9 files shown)


transaction/src/plan/build.rs
```diff
@@ -12,14 +12,16 @@ impl TransactionPlan {
     ///
     /// - `fvk`, the [`FullViewingKey`] for the source funds;
     /// - `auth_data`, the [`AuthorizationData`] authorizing the transaction;
-    /// - `witness_data`, the [`WitnessData`] used for proving.
+    /// - `witness_data`, the [`WitnessData`] used for proving;
+    /// - `precision_bits`, the current FMD precision.
```
hdevalence, line 16: Hmm, should this be a parameter for `build`, or data that's included in the clue plan?

If the clues are part of the auth hash, the precision bits will need to be part of the plan, so that the custodian can compute the clues (to hash them) before signing.

But if transactions with the wrong FMD precision are rejected, the requirement that the precision be defined pre-signing is effectively placing a bound on how long a signer can take to authorize a transaction, so we should consider that timing when deciding how frequently to adjust the precision.

redshiftzero: Ah good point, I've moved the precision bits into the `CluePlan`, and added the `Clue`s to the authorizing data as part of this PR

PR #1314, opened 2022-08-25
Rebase swaps/swapclaims on current main branch changes, use address from notes, move NFT asset ID generation to SwapPlaintext

zbuc:
This simplifies a lot of the Swap/SwapClaim code and continues implementation stubbing in the client.

### DIFF
--- a/crypto/src/dex/swap/plaintext.rs
+++ b/crypto/src/dex/swap/plaintext.rs
@@ -37,7 +37,7 @@ impl SwapPlaintext {
     // Constructs the unique asset ID for a swap as a poseidon hash of the input data for the swap.
     //
     // https://protocol.penumbra.zone/main/zswap/swap.html#swap-actions
-    pub fn generate_swap_asset_id(&self) -> asset::Id {
+    pub fn asset_id(&self) -> asset::Id {
         let packed_values = {
             let mut bytes = [0u8; 24];
             bytes[0..8].copy_from_slice(&self.delta_1.to_le_bytes());
--- a/crypto/src/dex/swap/plaintext.rs
+++ b/crypto/src/dex/swap/plaintext.rs
@@ -55,7 +55,7 @@ impl SwapPlaintext {
                 self.claim_address
                     .diversified_generator()
                     .vartime_compress_to_field(),
-                *self.claim_address.cached_s(),
+                *self.claim_address.transmission_key_s(),
             ),
         );
 
--- a/crypto/src/dex/swap/plaintext.rs
+++ b/crypto/src/dex/swap/plaintext.rs
@@ -1,8 +1,11 @@
 use crate::symmetric::OutgoingCipherKey;
 use crate::transaction::Fee;
-use crate::{ka, Address};
+use crate::{asset, ka, Address};
 use anyhow::{anyhow, Error, Result};
+use ark_ff::PrimeField;
+use decaf377::Fq;
 use penumbra_proto::{crypto as pb_crypto, dex as pb, Protobuf};
+use poseidon377::hash_5;
 
 use crate::dex::TradingPair;
 use crate::{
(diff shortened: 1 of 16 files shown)


crypto/src/dex/swap/plaintext.rs
```diff
@@ -29,6 +34,34 @@ pub struct SwapPlaintext {
 }
 
 impl SwapPlaintext {
+    // Constructs the unique asset ID for a swap as a poseidon hash of the input data for the swap.
+    //
+    // https://protocol.penumbra.zone/main/zswap/swap.html#swap-actions
+    pub fn generate_swap_asset_id(&self) -> asset::Id {
```
zbuc, line 40: Maybe this is better renamed as `fn asset_id(...)`

crypto/src/proofs/transparent.rs
```diff
@@ -455,12 +455,12 @@ impl SwapClaimProof {
             amount: 1,
             asset_id: self.swap_nft_asset_id,
         };
-        let s_component_transmission_key = Fq::from_bytes(self.pk_d.0);
+        let s_component_transmission_key = Fq::from_bytes(self.claim_address.transmission_key().0);
```
hdevalence, line 458: could this be `cached_s()`?

proto/proto/dex.proto
```diff
@@ -39,7 +39,7 @@ message SwapClaimBody {
   // Note output for asset 2.
   crypto.NotePayload output_2 = 4;
   // Block root for the associated `Swap` action, identifying the block the `Swap` was included in.
-  crypto.MerkleRoot anchor = 5;
+  // crypto.MerkleRoot anchor = 5;
```
hdevalence, line 42: how do we identify the block the swap was included in?

transaction/src/plan/build.rs
```diff
@@ -64,6 +64,26 @@ impl TransactionPlan {
             actions.push(Action::Output(output_plan.output(fvk.outgoing())));
         }
 
+        // Build the transaction's swaps.
+        // TODO: figure this out
+        // for ((swap_plan, auth_sig), auth_path) in self
+        //     .swap_plans()
+        //     .zip(auth_data.spend_auths.into_iter())
+        //     .zip(witness_data.note_commitment_proofs.into_iter())
```
hdevalence, line 72: the swap plans shouldn't require pulling in spend auth signatures, because the `Swap` just draws from the transaction's value balance. but we'll need to figure out how we handle the authentication paths (note commitment proofs) -- in the existing `WitnessData` struct, there's just one for each spend, in order, because (now-broken assumption) the only place they were needed was for spends.

wondering if we should change the `WitnessData` domain type to store a `BTreeMap<note::Commitment, tct::Proof>`, and when deserializing from the proto, instead of just deserializing a list of proofs, call `.commitment()` on each one and use it as the key for the `BTreeMap`.

then instead of having to do `zip`s or whatever, or figure out how to maintain a brittle order dependency, we can just query for exactly the proof we want.

hdevalence:
This looks great, much simpler -- I think the proofs could also be potentially simplified in the same way, by replacing the address components with the address itself.

proto/proto/dex.proto
zbuc: This was based on the removal of the `anchor` from the `SwapClaimPlan`. The anchor is passed in to the `SwapClaimProof::verify` call: https://github.com/penumbra-zone/penumbra/blob/main/crypto/src/proofs/transparent.rs#L442

Is it possible to get the anchor from the block height? Otherwise I think we'll need to add an `anchor` argument to `SwapClaimPlan::swap_claim_proof` (https://github.com/penumbra-zone/penumbra/blob/main/transaction/src/plan/action/swap_claim.rs#L87). I don't _think_ it requires being in the `SwapClaimBody` for any reason but maybe I'm missing something.

PR #1327, opened 2022-08-26
Add pcli view query for transaction hashes by block

### LINKED ISSUE #1192: Add MVP of transaction viewing functionality to `pcli`

In #1116, we added code to the view service that downloads and records extended transaction info.  However, we don't currently expose that data in `pcli` anywhere.

To get started, we should add a new `pcli view` subcommand that will handle any queries to the view service.  Initially, this could just have a single subcommand, `pcli view list-transactions`, with optional `--start-height` and `--end-height` parameters.  To start, `list-transactions` should just print a table with the block height in one column and the transaction hash in the other.

This command would need to be backed by an RPC to the view service.  This could be
```proto
rpc Transactions(TransactionsRequest) returns (stream transaction.Transaction);
``` 
with
```proto
message TransactionsRequest {
    // Identifies the FVK for the notes to query.
    // If present, return only transactions after this height.
    uint64 start_height = 2;
    // If present, return only transactions before this height.
    uint64 end_height = 3;
}
```

The current implementation in `pcli` can just hash the returned transactions and discard them, but returning the full transaction data means we can do more stuff later.

We could also aim to make the `pcli` output machine-readable.  However, I think this is a mistake -- we're much better off telling people who want machine-readable data to just query the view service's sqlite database directly.

aubrika:
Closes #1192

### DIFF
--- a/pcli/src/command/view/transactions.rs
+++ b/pcli/src/command/view/transactions.rs
@@ -1,0 +1,42 @@
+use anyhow::Result;
+use comfy_table::{presets, Table};
+use penumbra_crypto::FullViewingKey;
+use penumbra_view::ViewClient;
+
+#[derive(Debug, clap::Args)]
+pub struct TransactionsCmd {
+    #[clap(short, long)]
+    pub start_height: Option<u64>,
+    #[clap(short, long)]
+    pub end_height: Option<u64>,
+}
+
+impl TransactionsCmd {
+    pub fn needs_sync(&self) -> bool {
+        true
+    }
+
+    pub async fn exec<V: ViewClient>(&self, fvk: &FullViewingKey, view: &mut V) -> Result<()> {
+        // Initialize the table
+
+        let mut table = Table::new();
+        table.load_preset(presets::NOTHING);
+
+        let txs = view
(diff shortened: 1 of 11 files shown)


pcli/src/command/view/transactions.rs
```diff
@@ -0,0 +1,42 @@
...
+        let txs = view
+            .transactions(fvk.hash(), self.start_height, self.end_height)
+            .await?;
+
+        table.set_header(vec!["Block Height", "Transaction Hash"]);
+
+        for tx in txs {
+            table.add_row(vec![
+                format!("{}", u128::from(tx.0)),
+                format!("{:?}", hex::encode(tx.1)),
```
hdevalence, line 34: This produces
```
 Block Height  Transaction Hash
 568           "ce94c042b9c3cc69a20704728bd73773d955ee5594e9e365cdfe9a976e725591"
```
but if we change to 
```suggestion
                format!("{}", hex::encode(tx.1)),
```
it will be
```
 Block Height  Transaction Hash
 568           ce94c042b9c3cc69a20704728bd73773d955ee5594e9e365cdfe9a976e725591
```

PR #1328, opened 2022-08-26
support multi-asset fees

### LINKED ISSUE #1223: Allow multi-asset fees in transactions

Currently, Penumbra encodes fees as an untyped integer, representing an amount of the staking token.  The requirement that fees are only paid in the staking token is motivated by privacy: fees are necessarily public, and allowing fees to be paid in multiple assets decreases privacy for users who use non-standard fee tokens.

However, in a cross-chain environment, this requirement imposes severe usability constraints, because users cannot move funds from another chain without first acquiring staking tokens by some other means (or worse, users could move funds into Penumbra, and then have them trapped, because they can't pay any fees to do anything).

For this reason, it seems preferable to allow fees to be paid in other assets at the protocol level, even if the UX *strongly discourages* users from doing so.

Change the `Fee` proto from
```proto
// Specifies fees paid by a transaction.
message Fee {
    uint64 amount = 1;
}
```
to
```proto
// Specifies fees paid by a transaction.
message Fee {
    // The amount of the token used to pay fees.
    uint64 amount = 1;
    // If present, the asset ID of the token used to pay fees.
    // If absent, specifies the staking token implicitly. 
    crypto.AssetId asset_id = 2;
}
```

Then:

- [x] Change the domain type to be `Fee(pub Value)` (a newtype wrapper for a typed value), and:
  - [x] In the conversion from proto to domain type, fill in `STAKING_TOKEN_ASSET_ID` if `asset_id` is `None`;
  - [x] In the conversion from domain to proto type, check if `asset_id` is `STAKING_TOKEN_ASSET_ID` and encode it to `None` if so;
- [x] Update the value commitment computations that use the existing `amount` field to form a value commitment to the correct asset type instead.

**Additional context**

Currently, we burn fees. This doesn't really make sense for arbitrary assets (e.g., IBC tokens), so, following work on governance, we should:

- [ ] #1224

redshiftzero:
Closes #1223. After discussion in Discord, we decided not to expose the multi-asset fees in pcli until we decide how we want to handle them economically.


crypto/src/dex/swap.rs
```diff
@@ -12,9 +12,9 @@ use once_cell::sync::Lazy;
 use penumbra_proto::{dex as pb, Protobuf};
 
 // Swap ciphertext byte length
-pub const SWAP_CIPHERTEXT_BYTES: usize = 184;
+pub const SWAP_CIPHERTEXT_BYTES: usize = 216;
 // Swap plaintext byte length
-pub const SWAP_LEN_BYTES: usize = 168;
+pub const SWAP_LEN_BYTES: usize = 200;
```
redshiftzero, line 17: cc @zbuc: the size of the swaps increases by 32 bytes to provide space for the fee asset ID

wallet/src/plan/planner.rs
```diff
@@ -68,14 +69,11 @@ impl<R: RngCore + CryptoRng> Planner<R> {
 
     /// Add a fee to the transaction plan.
     ///
-    /// Calling this function more than once will add to the fee, not replace it.
+    /// This function should be called once.
```
redshiftzero, line 72: We can't add to the existing fee, since there is a single fee field per transaction

(I could have changed this to allow addition when the fee asset IDs match, but that seems unnecessary and this is simpler)

hdevalence: Yeah, I don't think there's a good use case for incrementally adding fees.

PR #1329, opened 2022-08-29
Support awaiting detection of note commitments during pcli transaction submission

zbuc:


hdevalence (approved):
This seems good to merge -- but in the longer term, I'm wondering if we might want to move the logic about what to wait for out of `build_and_submit_transaction`, since that logic is pretty specific about the command, but now it's spread over both the command and the common function.

PR #1330, opened 2022-08-29
Automatically submit SwapClaim after Swap completes

zbuc:
#1329 should be merged first, this PR was branched off of that since it relies on the "waiting on commitments to sync to view" behavior introduced in #1329

### DIFF
--- a/pcli/src/command/tx.rs
+++ b/pcli/src/command/tx.rs
@@ -1,6 +1,8 @@
 use anyhow::{anyhow, Context, Result};
 use penumbra_component::stake::rate::RateData;
 use penumbra_crypto::{asset, DelegationToken, IdentityKey, Value, STAKING_TOKEN_ASSET_ID};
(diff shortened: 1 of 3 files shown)


pcli/src/command/tx.rs
```diff
@@ -178,13 +180,58 @@ impl TxCmd {
...
+                        ..Default::default()
+                    })
+                    .await?;
+
+                if swap_nft_notes.len() != 1 {
+                    return Err(anyhow::anyhow!(
+                        "expected 1 swap NFT note after submitting Swap, found {}",
+                        swap_nft_notes.len()
+                    ));
+                }
```
hdevalence, line 218: If, instead of putting the "wait for confirmation" logic inside of `build_and_submit_transaction`, we awaited detection of the swap nft by commitment here, we wouldn't need to build a query that tries to select it here, because the method that waits for detection returns a `SpendableNoteRecord`.

PR #1334, opened 2022-08-29
Update guide docs

pygmygoats:


docs/guide/src/pcli/transaction.md
```diff
@@ -0,0 +1,39 @@
...
+Find a validator to stake to:
+```
+cargo run --release --bin pcli view validator list
+```
+
+Copy and paste the address of one of the validators to stake to, then construct the staking tx:
+
+```
+cargo run --release --bin pcli tx delegate 10penumbra --to penumbrav1t...
+```
```
plaidfinch, line 36: I think this should instead say something like "Copy and paste the validator key for one of the validators to stake to..." and then specify the validator address as `penumbravalid1...`.

plaidfinch (changes requested):

plaidfinch (approved):

PR #1337, opened 2022-08-30
Implement swap_claim in wallet, implement batch_swap_output_data in view service

zbuc:


zbuc:
This doesn't include anything to populate the `batch_swap_output_data` table, that will come in a future PR.

hdevalence:
Do we want to record the `BatchSwapOutputData` in the view service? It doesn't have any private information, so we could just as well fetch it from a full node (by adding methods to the specific client query rpc).

zbuc:
> Do we want to record the `BatchSwapOutputData` in the view service? It doesn't have any private information, so we could just as well fetch it from a full node (by adding methods to the specific client query rpc).

We store some other non-private information (chain params, FMD params, assets) in the view service so it wasn't clear. Do we get anything by storing it locally? Probably not, it might be simpler to just fetch it like you suggest.

PR #1362, opened 2022-09-06
updates for 028-Harpalyke testnet

redshiftzero:
Towards #1333


redshiftzero:
With e1b02e05deedf0d347dba46a292b2670c6bf2c79 CI is now passing so merging. I could have unpinned both tokio and console-subscriber, but there are other API changes on our side, so am keeping them pinned for now and filed #1363 for followup.

PR #1371, opened 2022-09-07
protocol: lift memos to transaction level

### LINKED ISSUE #1222: lift `Memo` to transaction level
Currently we have memos on a per-Output basis. They are encrypted to a shared key between the sender and the recipient. 

One alternative is to have a per-transaction `Memo`. 

Pros:
* This makes transactions smaller: instead of N memos for N outputs, we'd have just one memo, plus some overhead due to the key wrapping we would need such that the recipient associated with each output can decrypt the memo. Memo ciphertexts are 528 bytes currently. 

Cons:
* If in a single transaction the user wanted to send a _different_ memo to two outputs, this would no longer be possible. Unclear how common this is. However they could send two transactions in that case. 
* In the case of multiple outputs, each recipient associated with an output learns the memo that is sent to all others in the transaction. How important is this?

redshiftzero:
Closes #1222

Followup:
- [ ] Add consensus rule to check we only have memos for transactions with outputs


avahowell:
LGTM

PR #1372, opened 2022-09-07
Live TCT visualization

plaidfinch:
This implements graphviz DOT rendering for the TCT, as well as an animated in-browser live-updated view for an arbitrary tree stored in a `tokio::watch`, which can be updated and queried using a REST API. This is used to make an interactive in-browser visualization demo, the binary `tct-live-edit`, which serves a single tree which can be simultaneously edited in the browser by any number of people. Some effort has been made to make this efficient for reasonably large numbers of viewers and controllers: rendering effort is shared between requests and cached, long polling is used to provide quick updates with very little extra traffic, etc.

Possible future improvements (not planned, just noting for future-me):

The graphviz rendering code is bespoke string munging, because I could not find a library that supported all the DOT attributes I needed to produce exactly the format needed to render the tree nicely. It could definitely be improved, both in terms of output file size (there is repetition that could be condensed) and code style of the rendering code itself. The browser/server interaction could also be further improved, in particular by tracking a single server-side incrementing version counter that includes interior mutations as version increments, and then using only the long-polling API, doing away with the need to use SSE to communicate extra interior mutation changes that can't be requested via the long-polling interface. This would reduce complexity on both the javascript client and rust sides.


hdevalence:
Is it possible to keep the frontend code in a different repository than this one? I'm not sure we have the capacity to maintain it as part of the monorepo.

plaidfinch:
> Is it possible to keep the frontend code in a different repository than this one? I'm not sure we have the capacity to maintain it as part of the monorepo.

We could do this, sure. There were some changes made in this PR in order to support the functionality, but it doesn't depend on internals of the TCT to work, so it can be pulled out. I'll do that, then merge this pull request to incorporate the changes that support the live view functionality, making a standalone repo for the live view itself.

plaidfinch:
Update: the code is extracted into a standalone crate, but still within the workspace. Once we solve why it refuses to compile when using the penumbra repo as a git dependency, we can pull it out into a separate repo, but I'm going to merge it now and we can figure this out later if it becomes an issue.

PR #1387, opened 2022-09-09
Enable Swap/SwapClaim plans in transaction build

zbuc:


transaction/src/auth_hash.rs
```diff
@@ -152,6 +152,11 @@ impl TransactionPlan {
         for swap_claim in self.swap_claim_plans() {
             state.update(swap_claim.swap_claim_body(fvk).auth_hash().as_bytes());
         }
+        let num_clues = self.clue_plans.len() as u32;
```
redshiftzero, line 155: My bad because I sent you astray in Discord earlier, but the two places that need to match with respect to the clue keys are: https://github.com/penumbra-zone/penumbra/blob/main/transaction/src/auth_hash.rs#L189 and https://github.com/penumbra-zone/penumbra/blob/main/transaction/src/auth_hash.rs#L92

I think they do match for clue keys, so the stanza here adding the clue keys into the auth hash should stay after adding all the actions

transaction/src/plan/build.rs
```diff
@@ -84,29 +84,23 @@ impl TransactionPlan {
...
         // Build the transaction's swap claims.
-        // for swap_claim_plan in self.swap_claim_plans().cloned() {
-        //     actions.push(Action::SwapClaim(swap_claim_plan.swap_claim(
-        //         fvk,
-        //         note_commitment_proof,
-        //         nk,
-        //         note_blinding,
-        //     )));
-        // }
+        for swap_claim_plan in self.swap_claim_plans().cloned() {
```
redshiftzero, line 93: In the transaction plan auth_hash, we're adding the swaps and swap_claims after the outputs, as we are here, so I think this makes sense as a resolution ✅

redshiftzero (changes requested):
I agree that your fix adding the swap and swap claims to `TransactionPlan::build` method should resolve the auth hash mismatch 🎉 , but if you agree with my comment re: the clue plans, that part should stay as it is on main.

PR #1401, opened 2022-09-12
provide dummy memo key such that `Transaction::auth_hash` does not panic

### LINKED ISSUE #1394: `TransactionPlan::auth_hash` can panic on missing memos
**Describe the bug**

The implementation of `TransactionPlan::auth_hash` can panic on a missing memo, because it has this code:
```rust
        for output in self.output_plans() {
            state.update(
                output
                    .output_body(fvk.outgoing(), &memo_key.clone().unwrap()) // PANIC
                    .auth_hash()
                    .as_bytes(),
            );
        }
 ```

**To Reproduce**

I encountered this when running `tx sweep`.

**Expected Behavior**

`TransactionPlan::auth_hash` should be infallible.

redshiftzero:
~Making this a little clearer for the future by getting the memo key with `expect`. Related to #1395~

Closes #1394
Flagging that any outputs without memos will be invalid transactions as of #1403

### DIFF
--- a/crypto/src/symmetric.rs
+++ b/crypto/src/symmetric.rs
@@ -98,6 +98,12 @@ impl TryFrom<Vec<u8>> for PayloadKey {
     }
 }
 
+impl From<[u8; 32]> for PayloadKey {
+    fn from(bytes: [u8; 32]) -> Self {
+        Self(*Key::from_slice(&bytes))
+    }
+}
+
 /// Represents a symmetric `ChaCha20Poly1305` key.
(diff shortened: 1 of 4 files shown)


hdevalence:
This is still going to panic if the TransactionPlan is missing a memo. I think that's not a good behavior, because TransactionPlans might come in eg over the network, and it shouldn't be possible to DoS the custodian.

hdevalence (approved):
LGTM

PR #1414, opened 2022-09-12
add `gm` and `gn` to asset registry, parse genesis allocations with `parse_unit`

### LINKED ISSUE #1398: Change `gm` and `gn` to be display denoms of a finer base denom.

Per #1397, it'd be convenient if `gm` and `gn` weren't atomic units, but could be subdivided.

- [ ] Add `gm` and `gn` to the asset registry, with a reasonable base denom (say 1e6, `ugm`, `ugn`, like `penumbra`).
- [ ] Patch the genesis processing logic to use `parse_unit` when processing genesis allocations, so that genesis allocations can be specified in display denoms as well as base denoms (i.e., if there's an allocation of `x` `"gm"`, parse `"gm"` with `parse_unit`, then multiply by `10.pow(unit.exponent().into())` to get the correct amount of the base token).

avahowell:
Closes #1398

### DIFF
--- a/crypto/src/value.rs
+++ b/crypto/src/value.rs
@@ -325,9 +325,6 @@ mod tests {
         assert_eq!(v1.format(&cache), "999upenumbra");
         assert_eq!(v2.format(&cache), "1mpenumbra");
         assert_eq!(v3.format(&cache), "4penumbra");
-        assert_eq!(
-            v4.format(&cache),
-            "1passet1ucz7d7n90gpw92s7lye2xl93syr3m08r38a2r49039g5sc940ygqhz4xa0"
-        );
+        assert_eq!(&v4.format(&cache)[..8], "1passet1");
     }
 }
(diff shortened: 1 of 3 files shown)


crypto/src/value.rs
```diff
@@ -298,7 +298,7 @@ mod tests {
     #[test]
     fn format_picks_best_unit() {
         let upenumbra_base_denom = asset::REGISTRY.parse_denom("upenumbra").unwrap();
-        let gm_base_denom = asset::REGISTRY.parse_denom("gm").unwrap();
+        let gm_base_denom = asset::REGISTRY.parse_denom("ugm").unwrap();
```
zbuc, line 301: it doesn't look like the `gm_base_denom` is ever used after this? should there be another test case? (not sure it'd really exercise anything different than the existing cases though)

zbuc (approved):

PR #1421, opened 2022-09-13
chore: use `tokio@1.21.1` and `console-subscriber@0.1.8`

### LINKED ISSUE #1363: Use tokio 1.21.0
We pinned to tokio 1.20.1 in https://github.com/penumbra-zone/penumbra/issues/1355 due to https://github.com/tokio-rs/console/issues/373. 

We should unpin and move to latest tokio (1.21.0 as of this ticket), making whatever API changes on our side we need. As part of this we will move to `console-subscriber` 0.1.8 which includes the fix for https://github.com/tokio-rs/console/issues/373.

erwanor:
This should fix #1363 , see the diff for more details but I went ahead and bumped a couple crates from `tokio@1` to `1.21.1`.

The main API change was in `penumbra-storage`: [`tokio::task::Builder::spawn_blocking`](https://docs.rs/tokio/latest/tokio/task/struct.Builder.html#method.spawn_blocking) is fallible. It returns an `io::Error` if the runtime fails to spawn a task. We're not handling those errors, but let me know if you think we should!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -856,9 +856,9 @@ dependencies = [
 
 [[package]]
 name = "console-subscriber"
-version = "0.1.7"
+version = "0.1.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e933c43a5db3779b3600cdab18856af2411ca2237e33ba8ab476d5d5b1a6c1e7"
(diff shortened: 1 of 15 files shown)


redshiftzero (approved):
thanks for the contribution!

PR #1422, opened 2022-09-13
view: add explanation for future devs about not panicing in view

### LINKED ISSUE #1086: view service: document no errors in view scanning
When we hit an error in the view service while scanning notes (e.g. #1085), we should ensure that we do not panic. When errors occur they should be gracefully handled (i.e. dropping malformed notes) such that the view service continues to run. This should be documented for developers.

redshiftzero:
Closes #1086

Is this an accurate description for future devs?


aubrika (approved):
Excellent add 👍🏻

Issue #1364, opened 2022-09-06
Add pcli view transaction command for single transactions by hash


aubrika:
**Is your feature request related to a problem? Please describe.**
The issue is that the pcli query tx command just displays the full raw transaction data - this is useful, but potentially more information than a user needs.

**Describe the solution you'd like**
We want a nice interface for summarizing the most relevant transaction details for a specific transaction, which we can do as `pcli v tx`, displaying a nice table broken down by address of the most important details of a specific transaction.

**Describe alternatives you've considered**
Using the `pcli q tx` command is very verbose and not easily digestible, so having a more succinct transaction details option seems useful.

Issue #1368, opened 2022-09-06
For each transaction action, determine which additional data if any is needed to compute net change to value balance


aubrika:
Each transaction action will not necessarily include the data required to compute and display its net effect on the account value balance - survey each action type to determine what additional data is needed as preliminary research for displaying this change.

- [x] `Output`
- [x] `Spend`
- [ ] `Delegate`
- [ ] `Undelegate`
- [ ] `ValidatorDefinition`
- [ ] `IBCAction`
- [x] `Swap`
- [x] `SwapClaim`
- [ ] `ProposalSubmit`
- [ ] `ProposalWithdraw`
- [ ] `DelegatorVote`
- [ ] `ValidatorVote`
- [ ] `PositionOpen`
- [ ] `PositionClose`
- [ ] `PositionWithdraw`
- [ ] `PositionRewardClaim`
- [ ] `ICS20Withdrawal`

aubrika:
During initial design work for #1426, @redshiftzero & I found that currently only Output, Spend, Swap, and SwapClaim Actions will need additional data in the context of a `TransactionView` in order to compute their net effects on account balance. These Action types will require their associated Notes and PayloadKeys to calculate their Amounts, and in the case of Spends, also their Nullifiers. Some of the other Action types above will require additional fields following the implementation of flow encryption, but proximally we should be able to use what we learned to implement the pcli command described in #1364.

aubrika:
Research complete, removing blocker on implementation of #1364, which is started in #1426

PR #1426, opened 2022-09-14
implement `TransactionView`

redshiftzero:
@aubrika and I started sketching out the `TransactionView` required for implementing `pcli v tx` as described in #1364


aubrika:
As mentioned in #1368, it looks like there are just four Action types that will require additional data not present in the Transaction object itself (for now). @hdevalence does this look like we are heading in the right direction here based on the stubbed decryption code?

transaction/src/action.rs
```diff
@@ -33,6 +35,10 @@ pub use vote::{DelegatorVote, ValidatorVote, ValidatorVoteBody, Vote};
 /// Common behavior between Penumbra actions.
 pub trait IsAction {
     fn balance_commitment(&self) -> balance::Commitment;
+    fn decrypt_with_perspective(
+        &self,
+        txp: &TransactionPerspective,
+    ) -> anyhow::Result<Option<ActionView>>;
```
hdevalence, line 41: What's the meaning of `Result<Option<ActionView>>`?

transaction/src/action/output.rs
```diff
@@ -22,6 +25,45 @@ impl IsAction for Output {
...
+    ) -> anyhow::Result<Option<ActionView>> {
+        // Get payload key for note commitment of note payload
+
+        let note_commitment = self.body.note_payload.note_commitment;
+
+        // Get payload key for note commitment of swap NFT.
+        let payload_key = txp
+            .payload_keys
+            .get(&note_commitment)
+            .ok_or_else(|| anyhow::anyhow!("corresponding payload key not found"))?;
```
hdevalence, line 41: This error message is pretty opaque, and I think it won't give much information if it's encountered. Perhaps
```
perspective is missing payload key for output note commitment {}
```
would be better, since it identifies exactly what's missing and in what context.

transaction/src/action/ibc.rs
```diff
@@ -31,6 +33,13 @@ impl IsAction for ICS20Withdrawal {
     fn balance_commitment(&self) -> penumbra_crypto::balance::Commitment {
         self.balance().commit(Fr::zero())
     }
+
+    fn decrypt_with_perspective(
+        &self,
+        txp: &TransactionPerspective,
+    ) -> anyhow::Result<Option<ActionView>> {
+        todo!()
```
hdevalence, line 41: What would this be filled in with, considering that the `ICS20Withdrawal` action doesn't have any shielded data?

transaction/src/lib.rs
```diff
@@ -8,10 +8,12 @@ mod witness_data;
 
 pub mod action;
 pub mod plan;
+pub mod transaction_view;
```
hdevalence, line 11: Since this is already in the `transaction` crate, it'd be better to call the module `view`, so that instead of `penumbra_transaction::transaction_view`, we have `penumbra_transaction::view`.

transaction/src/transaction_view/action_view.rs
```diff
@@ -0,0 +1,17 @@
...
+pub use swap_claim_view::SwapClaimView;
+pub use swap_view::SwapView;
+
+pub enum ActionView {
+    /// View of a Swap action
+    Swap(SwapView),
+    SwapClaim(SwapClaimView),
+    Output(OutputView),
+    Spend(SpendView),
+}
```
hdevalence, line 17: This only covers actions that have decryptable content. How does a consumer of a `TransactionView` understand a transaction, like an ICS20 withdrawal, that includes a mix of shielded and transparent actions?

transaction/src/transaction_view/transaction_perspective.rs
```diff
@@ -0,0 +1,34 @@
...
+    /// to decrypt the wrapped_memo_key, which will be used to decrypt the
+    /// memo in the transaction. This needs to be done only once, because
+    /// there is one memo shared between all outputs.
+    pub payload_keys: BTreeMap<note::Commitment, PayloadKey>,
+    /// Mapping of nullifiers spent in this transaction to notes.
+    pub spend_nullifiers: BTreeMap<Nullifier, Note>,
+
+    pub memo_cipher_text: Option<MemoCiphertext>,
+
+    pub memo_plaintext: Option<MemoPlaintext>,
```
hdevalence, line 31: I think I'm missing something -- why are the memo plaintext and ciphertext included in the `TransactionPerspective`?

First, why do we need to record both the ciphertext and the plaintext -- if we have the ciphertext, can't we decrypt it to obtain the plaintext?

Second, why do we need the ciphertext -- won't that be in the Transaction we're viewing with this Perspective?

transaction/src/action/output.rs
```diff
@@ -22,6 +25,45 @@ impl IsAction for Output {
...
+            .ok_or_else(|| anyhow::anyhow!("no memo present"))?
+            .to_owned();
+
+        let memo = MemoPlaintext::decrypt(memo_cipher_text, &decrypted_memo_key)?;
+
+        Ok(Some(ActionView::Output(OutputView {
+            decrypted_note,
+            memo,
+        })))
+    }
```
hdevalence, line 66: This method is never going to return `Ok(None)`, right?  Doesn't that mean that we'll error if a transaction has an output that's not viewable from our `Perspective`?

transaction/src/action.rs
hdevalence, line 41: What's the meaning of the `Result<Option<ActionView>>` here? What does an error indicate, vs. an inner `None` value?

Should there be an error at all? In what sense would the method be fallible?

Also, since this produces an `ActionView`, maybe we should call it `view_with_perspective` to be consistent with the metaphor?

hdevalence: On the fallibility front, the current behavior of the implementation for `Output`s is to error if the output is not viewable from this perspective, xref https://github.com/penumbra-zone/penumbra/pull/1426/files#r982852861 

But there are normal cases where the output is not viewable, like when someone is viewing an inbound transaction.  In that case, bubbling up an error doesn't seem right.  What's the intended way to handle opaque actions?

transaction/src/action/ibc.rs
aubrika: `None`, maybe, depending on where we fall on what to do when there is nothing to decrypt.

transaction/src/transaction_view/transaction_perspective.rs
aubrika: We don't, this is just a couple placeholder fields to avoid refactoring the IsAction decryption method or the TransactionView further, since the Transaction itself isn't available on the method that actually decrypts the Memo. It could be, but this was a question mark. I have no strong opinions about how it should be handled, but refactoring the Transaction-level decryption method was going to be more work than was justified to hold this up further.

transaction/src/action/output.rs
aubrika: I suppose so - signature can change accordingly.

transaction/src/action.rs
aubrika: Yes, this needs taken into account & refactored for the desired behavior for each specific case.

aubrika:
Following yesterday's design discussion, this work will be refactored to include an `ActionView` associated with each `Action`, with the inner type of the `ActionView`s of transparent/currently-unencrypted `Actions simply being the associated `Action` type. This will allow for `pcli v tx` to handle a Vec of `ActionView` rather than special-casing those transaction actions with encrypted contents.

proto/proto/penumbra/core/transaction/v1alpha1/transaction.proto
```diff
@@ -73,6 +73,54 @@ message Action {
...
+  // The chain this transaction is intended for.  Including this prevents
+  // replaying a transaction on one chain onto a different chain.
+  string chain_id = 3;
+  // The transaction fee.
+  crypto.v1alpha1.Fee fee = 4;
+  // A list of clues for use with Fuzzy Message Detection.
+  repeated crypto.v1alpha1.Clue fmd_clues = 5;
+  // An optional plaintext memo. It will only be populated if there are
+  // outputs in the actions of this transaction. 528 bytes.
+  optional bytes plaintext_memo = 6;
```
hdevalence, line 92: Let's make this `optional string memo = 6;`, so that API consumers don't have to handle byte parsing.

We can also delete the commend about "528 bytes", since that's the size of the ciphertext, but here I think it makes more sense to truncate trailing zero bytes in the plaintext so that consumers can just `println!("{}", view.memo)` or whatever.

proto/proto/penumbra/core/transaction/v1alpha1/transaction.proto
```diff
@@ -73,6 +73,54 @@ message Action {
...
+}
+
+// A view of a specific state change action performed by a transaction.
+message ActionView {
+  oneof action {
+    Spend spend = 1;
+    Output output = 2;
```
hdevalence, line 99: These ones (and the `Swap`, `SwapClaim` below) are the four that need replacing with view-specific protos, right?

transaction/src/action/delegate.rs
```diff
@@ -31,6 +33,10 @@ impl IsAction for Delegate {
     fn balance_commitment(&self) -> penumbra_crypto::balance::Commitment {
         self.balance().commit(Fr::zero())
     }
+
+    fn decrypt_with_perspective(&self, txp: &TransactionPerspective) -> anyhow::Result<ActionView> {
```
hdevalence, line 37: Now that we have the design decision to have an "interpreted" `ActionView`, I think that this method would really be better named `view_from_perspective`, rather than `decrypt` -- and we won't need a `Result`, since viewing is infallible.

transaction/src/action/ibc.rs
hdevalence: This is solved by the `ActionView` design

transaction/src/view.rs
```diff
@@ -0,0 +1,16 @@
...
+use penumbra_crypto::{memo::MemoPlaintext, transaction::Fee};
+pub use transaction_perspective::TransactionPerspective;
+
+pub struct TransactionView {
+    pub actions: Vec<ActionView>,
+    pub expiry_height: u64,
+    pub chain_id: String,
+    pub fee: Fee,
+    pub fmd_clues: Vec<Clue>,
+    pub memo: Option<MemoPlaintext>,
```
hdevalence, line 15: This could just be `Option<String>` instead; it's less code to parse the protos, and I think we should get rid of `MemoPlaintext` eventually anyways.

transaction/src/transaction_view/action_view.rs
aubrika: Latest design will associate each `Action` with a corresponding `ActionView` so that a consumer can treat shielded & transparent actions as the same type of thing.

transaction/src/action/delegate.rs
aubrika: I suggest we leave the signature for the moment & make an issue for next release to go over this work and make sure Results/Errors arising from composed function calls are handled in a way that's informative, because we can make `view_from_perspective` infallible, but there are component functions that may fail & I don't want to refactor around that thoughtlessly.

Issue #1369, opened 2022-09-06
Add a `MultiValue` utility type to the `crypto` crate alongside `Value`.


hdevalence:
**Is your feature request related to a problem? Please describe.**

Currently, we have a `Value` type recording a typed value (amount + asset id).  We can form commitments to `Value`s, and we can sum commitments even of different types, because of the way the value commitment mechanism works.

However, we can't directly represent the *opening* of one of those value commitments, because each `Value` can only record a single type of value.

It would be useful to have a type that could represent a vector of typed values.  This would let us implement tx viewing commands (e.g., in #1364) more easily, because we could implement one formatting method for a multi-value type, then construct instances of that type and format them, rather than having the code for representing an action's combined effects on value balance live in the UI in some ad-hoc way.

**Describe the solution you'd like**

Add a new type to the `crypto` module, alongside the existing `Value`:
```rust
pub struct MultiValue {
    inner: BTreeMap<asset::Id, u64>,
}
```
Then, fill in some useful utility methods:
- [x] conversion from `Value`s
- [x] a way to add `Value` to a `MultiValue`, accumulating into the correct entry
- [x] a way to iterate over the component `Value`s of a `MultiValue`
- [x] a way to create a `value::Commitment` given a blinding factor

We can test that the implementation is correct using property tests that use the homomorphic property of commitments (Commit(a + b) = Commit(a) + Commit(b)).

plaidfinch:
This is already fulfilled by https://github.com/penumbra-zone/penumbra/blob/main/wallet/src/plan/balance.rs#L16-L22. Perhaps we should move this type to a different place in the penumbra monorepo, so it can be used in more places?

Some things that could be added: `impl {Add, AddAssign, Sub, SubAssign}<Balance> for Value { type Output = Balance; }`, and the final checkbox above, a way to create a value commitment.

erwanor:
I am about to start work on implementing `commit` for `Balance`. This is what I have in mind:

For a blinding factor $b$, a vector of values $v_1, ..., v_n \in F$, and $G$, $H$ group elements:

$$bH + \sum_{i=1}^n (sign(v_i) (v_i G)$$

sign is negative if the respective `Balance` entry is `Required`, positive otherwise.

Does that make sense?

plaidfinch:
[...]

PR #1430, opened 2022-09-15
feat: implement `Balance::commit` and move crate to `penumbra-crypto`

erwanor:
Part of #1369 .

TODOs: 

- [x] lift `Balance` from `penumbra-wallet` to `penumbra-crypto`
- [ ] implement `Balance::commit`
- [ ] add tests for `Balance::commit`


erwanor:
I think it would be more useful to implement commit like this:

```rust
    pub fn commit(&self, blinding_factor: Fr) -> Commitment {
        let init = Commitment(Element::default());
        self.iter().fold(init, |acc, imbalance| {
            let (_sign, value) = imbalance.into_inner();
            let commitment = value.commit(blinding_factor)
            acc + commitment
        })
    }
}

```

because then we  can conveniently reconstruct a balance commitment by summing its individual parts:
```
B := (v_1, .., -v_i, v_i+1, ..., v_n)
commit(B) == commit(v_1)  + ... - commit(v_i) + commit(v_i+1) + ... commit(v_n)
```

however, my fear is that this makes the API misleading since the blinding factor adds up everytime.

What do you think? another @hdevalence @plaidfinch

PR #1431, opened 2022-09-15
Ignore `max_peer_block_height` for determining `latest_known_block_height`

### LINKED ISSUE #1427: pcli sync not terminating at current block height
**Describe the bug**
There's something going on with the sync. @zbuc noticed that the syncing seemed to slow down around 20k of the 56k to sync to, which I also see. The issue is that the reported height of the chain (via `pcli query chain info`) is only 20k, but the sync is continuing to run.

**To Reproduce**
1. `pcli view sync`

**Expected behavior**
Sync to current block height, and exit

**Screenshots**
Syncs quickly then slowly advances to a sync height that does not exist

zbuc:
Closes #1427

### DIFF
--- a/measure/src/main.rs
+++ b/measure/src/main.rs
@@ -127,24 +127,19 @@ impl Opt {
             .ok_or_else(|| anyhow::anyhow!("could not parse latest_block_height in JSON response"))?
             .parse()?;
 
-        let max_peer_block_height = sync_info
-            .get("max_peer_block_height")
-            .and_then(|c| c.as_str())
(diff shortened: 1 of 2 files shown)


zbuc:
Since `max_peer_block_height` is going to be dropped in future versions of Tendermint, I took the step of deprecating it completely here. It seems mostly useless to our clients, anyhow.

redshiftzero (approved):
I ran this against the current testnet and synced successfully :tada:. LGTM!

PR #1447, opened 2022-09-21
Simplify proofs by using Note/Address types rather than address components

### LINKED ISSUE #1315: Cleanup: use `Address` type within proofs
Now that the `Note` type contains the complete `Address` rather than several address components, the proof code can be simplified to work on `Address` types.

zbuc:
Closes #1315


redshiftzero (approved):
LGTM!

PR #1459, opened 2022-09-21
crypto: `Address` short form

### LINKED ISSUE #1367: Canonical short form for Penumbra addresses

Penumbra addresses are long:
```
penumbrav2t1f5u3qn64s0muy84pgqdwd2uew8j5cq9952hhq3p4we7xnz8065rufchf7lgl7l46p0gdy8hz86k0jesnj55v3yrj8neqrgwqc5s8fff8tnu7fwy6ccu4a6jm5fwex9saehwpgk
```

It would be helpful if we had some truncated short form that we could use in a UI. People are going to do this anyways, so it would be cool if we had a way to canonically truncate an address "safely".

Figure out whether the fact that we use F4Jumble makes it safe to define a canonical truncated short form of a Penumbra address.  If so, figure out what length it would need to be for security.  We could render it with an ellipsis, like
```
penumbrav2t1f5u3qn64s0muy84pgqdwd2uew8j5cq99…
```

redshiftzero:
Closes #1367

### DIFF
--- a/docs/protocol/src/protocol/addresses_keys/addresses.md
+++ b/docs/protocol/src/protocol/addresses_keys/addresses.md
@@ -83,6 +83,52 @@ human-readable prefixes:
 * `penumbra_tnXYZ_` for testnets, where XYZ is the current testnet number padded
   to three decimal places.
 
+### Short Address Form
+
+Addresses can be displayed in a short form. A short form of length $100$ bits (excludes the human-readable
+prefix) is recommended to mitigate address replacement attacks. The attacker's goal is to find a partially
+colliding prefix for any of $N$ addresses they have targeted. 
+
+#### Untargeted attack
+
+In an untargeted attack, the attacker wants to find two different addresses that have a colliding short form of length $M$ bits.
+
+A collision is found in $\sqrt(2^M)$ steps due to the birthday bound, where $M$ is the number of bits of the prefix.
+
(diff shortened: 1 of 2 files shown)


docs/protocol/src/protocol/addresses_keys/addresses.md
```diff
@@ -83,6 +83,52 @@ human-readable prefixes:
...
+* set of the short forms of the N targeted addresses. This set has $N$ elements (where each element has length $2^M$ bits).
+* set of the short forms of all addresses, which has size $2^M$ elements.
+
+A collision is found in $2^{M}/N$ steps.
+
+For example if the attacker has a target set of $N=2^{20} = 1,048,576$ addresses,
+she can find a collision after $2^{M/20}$ steps. Thus for a prefix length of $M$ bits,
+we get $M-20$ bits of security$. For a targeted security level of 100 bits,
+we need a prefix length of 100 bits, which corresponds to 20 characters of the Bech32m
+address, excluding the human-readable prefix and separator.
```
hdevalence, line 128: Maybe we should rephrase this to clarify that the security targets 80 bit security, and that we have to extend by the number of expected targets. We could be a little more conservative and pick 2^40 as a target set size, giving 120 bits total.

Also, if the target set size is `2^N` rather than `N = 2^...`, I think the notation becomes a little clearer, since then it's N+M.

hdevalence (approved):

PR #1464, opened 2022-09-22
Automated testnet - ingress for fullnodes

agouin:
Adds a kubernetes ingress for the fullnode services: tendermint rpc and penumbra grpc.

- Uses terraform to reserve a global static IP address for the ingress
- Ingress points to this IP name and uses the same ingress for both services.
- Requires 2 DNS `A` records to be setup, one for rpc and one for grpc.
- Google managed certificates will be provisioned once the A records are in place to the reserved IP
- Testnet teardown and replace will not destroy ingress. New pods will be included in the existing services

Ingress currently live at:
https://rpc.testnet.penumbra.strange.love/
https://grpc.testnet.penumbra.strange.love/


hdevalence:
FYI, we've just rolled back to Tendermint 0.34 in #1471; I don't think this will make any difference to the deployment, other than the version specifiers, but using TM 0.35 won't work any more.

agouin:
Updated to work with tm v0.34

hdevalence:
Is this good to merge?

danbryan (approved):

PR #1465, opened 2022-09-22
transaction: add IsAction trait

hdevalence:
This moves the existing balance commitment code into a trait common to all actions, which we can use as a place to add more common functionality for actions.


aubrika:
LGTM!

PR #1470, opened 2022-09-23
crypto: refactor transparent proof statements into gadgets

### LINKED ISSUE #1449: Refactor existing transparent proofs into gadgets
This is a stepping stone to moving to zk proofs

redshiftzero:
Closes #1449

### DIFF
--- a/crypto/src/proofs/transparent.rs
+++ b/crypto/src/proofs/transparent.rs
@@ -391,22 +391,15 @@ impl SwapClaimProof {
             self.note_commitment_proof.commitment(),
         )?;
 
-        // check the swap NFT Asset ID is properly constructed
-        let asset_id = self.swap_nft_asset_id;
-        let expected_plaintext = SwapPlaintext::from_parts(
+        // Check that the Swap NFT asset ID is properly constructed.
+        gadgets::asset_id_integrity(
+            self.swap_nft_asset_id,
             self.trading_pair.clone(),
             self.delta_1_i.into(),
             self.delta_2_i.into(),
             fee,
-            // This should ensure that the claim address matches the address
-            // used to construct the Swap NFT.
             self.claim_address,
-        )
-        .map_err(|_| anyhow!("error generating expected swap plaintext"))?;
-        let expected_asset_id = expected_plaintext.asset_id();
-        if expected_asset_id != asset_id {
-            return Err(anyhow!("improper swap NFT asset id"));
-        }
+        )?;
 
         // Validate the note commitment's height matches the output data's height.
         let position = self.note_commitment_proof.position();
--- a/crypto/src/proofs/transparent.rs
+++ b/crypto/src/proofs/transparent.rs
@@ -87,12 +87,7 @@ impl SpendProof {
 
         gadgets::rk_integrity(self.spend_auth_randomizer, rk, self.ak)?;
 
-        gadgets::diversified_address_integrity(
-            self.ak,
-            self.nk,
-            self.note.transmission_key().clone(),
-            self.note.diversified_generator(),
-        )?;
+        gadgets::diversified_address_integrity(self.ak, self.nk, self.note.clone())?;
 
         Ok(())
     }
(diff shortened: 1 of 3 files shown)


crypto/src/proofs/transparent.rs
```diff
@@ -56,63 +57,46 @@ impl SpendProof {
...
             self.note.diversified_generator(),
-            s_component_transmission_key,
-            self.note.clue_key(),
-        );
-
-        if self.note_commitment_proof.commitment() != note_commitment_test {
-            return Err(anyhow!("note commitment mismatch"));
-        }
+            self.note.transmission_key_s(),
+            self.note.clue_key().clone(),
```
hdevalence, line 65: Is there a reason for the `note_commitment_integrity` gadget not to take the `note` itself, rather than all of its components?

crypto/src/proofs/transparent.rs
```diff
@@ -56,63 +57,46 @@ impl SpendProof {
...
         // Merkle path integrity.
         self.note_commitment_proof
             .verify(anchor)
             .map_err(|_| anyhow!("merkle root mismatch"))?;
 
-        // Value commitment integrity.
-        if self.note.value().commit(self.v_blinding) != balance_commitment {
-            return Err(anyhow!("value commitment mismatch"));
-        }
+        gadgets::value_commitment_integrity(
```
hdevalence, line 74: Now that we've gone from "value commitment" to "balance commitment", should this gadget be `balance_commitment_integrity`?

crypto/src/proofs/transparent.rs
```diff
@@ -56,63 +57,46 @@ impl SpendProof {
...
+        )?;
 
-        // Spend authority.
-        let rk_bytes: [u8; 32] = rk.into();
-        let rk_test = self.ak.randomize(&self.spend_auth_randomizer);
-        let rk_test_bytes: [u8; 32] = rk_test.into();
-        if rk_bytes != rk_test_bytes {
-            return Err(anyhow!("invalid spend auth randomizer"));
-        }
+        gadgets::check_spend_authority(self.spend_auth_randomizer, rk, self.ak)?;
```
hdevalence, line 92: Naming nitpick: this isn't really checking that the prover has spend authority, right? The actual spend authority is checked by the auth sig, and this is proving that the provided randomized verification key is a correctly formed randomization of the right authorizing key.

Could we find a different name for this gadget that would capture this idea?

Not sure what the best name is, but what about something like `rk_integrity` or something more descriptive?

crypto/src/proofs/transparent.rs
```diff
@@ -397,39 +368,27 @@ impl SwapClaimProof {
...
+                asset_id: self.swap_nft_asset_id,
+            },
+            self.claim_address.diversified_generator().clone(),
+            self.claim_address.transmission_key_s().clone(),
+            self.claim_address.clue_key().clone(),
+            self.note_commitment_proof.commitment(),
+        )?;
+
+        // Check that the Swap NFT asset ID is properly constructed.
+        gadgets::asset_id_integrity(
```
hdevalence, line 384: Could this be `swap_nft_asset_id_integrity` instead, since it's only for swap NFTs and not general asset IDs? Or, we could factor the name into a module path, like `gadgets::swap::asset_id_integrity`?

crypto/src/proofs/transparent.rs
```diff
@@ -458,52 +417,56 @@ impl SwapClaimProof {
...
-            return Err(anyhow!("bad nullifier"));
-        }
+        gadgets::nullifier_integrity(
+            nullifier,
+            self.nk,
+            position,
+            self.note_commitment_proof.commitment(),
+        )?;
 
         let (lambda_1, lambda_2) = output_data.pro_rata_outputs((self.delta_1_i, self.delta_2_i));
```
hdevalence, line 427: This is (implicitly) one of the proof statements -- we'll need to have a gadget that checks the pro rata outputs are correct.  This doesn't quite fit the exact same form as the other gadgets, though. Maybe we should add `lambda_1` and `lambda_2` to the transparent proof?  In that case, we could delay adding that gadget to a future PR, so that this one can be a pure rearrangement of the existing checks.

PR #1471, opened 2022-09-23
roll back to tendermint 0.34

avahowell:
This PR updates all of our various dependencies to use 0.34-compatible versions.


avahowell:
This is building, but there seems to be an issue with tendermint 0.34 <> tower-abci. The app isn't receiving genesis and advancing state, it just sits idle.

hdevalence:
The issue with the app sitting idle was that `tower-abci` was using the 0.35 style length delimiters, rather than the 0.34 ones, and therefore sat waiting for the rest of the message. Fixed by https://github.com/penumbra-zone/tower-abci/commit/91b209f61ccc941683c16dfc34e4a3e3efae57fc

PR #1473, opened 2022-09-24
ci: disable protobuf docs

hdevalence:
These steps broke with the proto reorganization. We could update them, but we'd potentially have to update the steps every time we added new protos or changed the layout.  Now that we publish to buf.build, it seems like we might want to use that instead?


hdevalence:
I disabled the hosting with `firebase -P penumbra-protobuf hosting:disable`.

PR #1475, opened 2022-09-24
proto: commit generated code to repo

hdevalence:
This avoids the need to have a `protoc` during the build, at the cost that we now have to figure out how to ensure that the generated code is always up-to-date with the protos.


hdevalence:
#1476 is probably preferable to this, at least at the moment.

hdevalence:
It looks like `protobuf-src` sometimes tries to build from source: https://github.com/penumbra-zone/penumbra/actions/runs/3119805783/jobs/5059995187

Committing generated code seems like a cleaner option long-term.

PR #1485, opened 2022-09-26
update default config template with 0.34

### LINKED ISSUE #1482: Isonoe testnet produces blocks too fast
**Describe the bug**

The Isonoe testnet is running well on TM 0.34, but it's producing blocks much faster than the 5s block target.  Probably what's happened is that the config options required to get the behavior we want are different in 0.34 than 0.34.

I don't think this is critical to fix during the current Isonoe testnet, but we should figure out what the right config is and change the settings in the config template used by `pd testnet create` / `pd testnet join`.

avahowell:
closes #1482


avahowell:
yes it does, because it changes the timeout parameters to use underscores. the timeouts were set as defaults when we backported because 0.34 uses snake case instead of dash case, so our config parameters were just ignored

PR #1493, opened 2022-09-29
Create parallel `storage2` crate

### LINKED ISSUE #1488: Create a parallel `storage3` crate in the Penumbra workspace
To start #1454 #1487, we should create a parallel `storage3` crate in the workspace, and add it to the GH Actions workflow that renders docs.

zbuc:
Closes #1488


hdevalence:
Ah, looks like this broke the docker files, and hence the smoke test?

zbuc:
Oops, will fix

Issue #1490, opened 2022-09-28
Design how we will use the RocksDB API to maintain coherence between the JMT store and the key store


hdevalence:
Once we have a more exact definition for the new state API in #1489, we should plan what internal tracking we'll need to do to maintain coherence between the JMT state (which is internally versioned, but only has key hashes), and our table of keys that indexes it (which is not internally versioned).  To do this, we'll need to use the RocksDB snapshot APIs in some way.

zbuc:
There's a `create_checkpoint` method that will let us persist a RocksDB checkpoint to disk: https://docs.rs/rocksdb/latest/rocksdb/checkpoint/struct.Checkpoint.html#method.create_checkpoint

They say it's "lightweight" but in practice I'm not sure what that means. Judging from a random GitHub issue (lol) there seems to be a significant speed difference between snapshotting and checkpointing, with the obvious caveat that snapshots are only available in-memory: https://github.com/facebook/rocksdb/issues/8674

I am wondering how we could use the in-memory snapshots to maintain coherence across `pd` restarts. That is, there doesn't seem to be a way we could view the complete RocksDB state at a given historical block height once the snapshots have left memory, meaning the coherence between the indexing keys and JMT is broken. This makes me think that checkpoints are what we need.

In terms of maintaining coherence between JMT state and our table of indexing keys, it seems like what we want is for both of those to be associated with RocksDB checkpoints.

Naive proposal:

* During `end_block` after updates to the JMT and the indexing keys have been written, generate a checkpoint identifier (based on block height?)
* Save the checkpoint to disk
* For checking coherence at a given block height, load the checkpoint for the specified block height in a new `Storage` instance

hdevalence:
> I am wondering how we could use the in-memory snapshots to maintain coherence across `pd` restarts. That is, there doesn't seem to be a way we could view the complete RocksDB state at a given historical block height once the snapshots have left memory, meaning the coherence between the indexing keys and JMT is broken. This makes me think that checkpoints are what we need.

Well, one question is whether we want to maintain coherence across pd restarts. 

The primary reason to want coherence is so that we can have different parts of the program safely access state snapshots without the need to do locking. This allows us to, for instance, have the RPC operate independently of the consensus (or more importantly, ensure that the consensus is operating independently of any RPC requests, and won't block waiting to acquire a lock held by an RPC request).

But this motivation doesn't apply across restarts, because there's definitionally no sharing across a restart.

[...]

Issue #1454, opened 2022-09-21
State Architecture V3.0


hdevalence:
The chain state is the backbone of the Penumbra protocol, application logic, node implementation, client interfaces, etc.  We’re currently on the second major iteration of the state design. The first design was Postgres-based, and was only useful as an MVP. The second design moved all of the chain state into a JMT instance, and refactored all of the application logic into components that communicate by message-passing via the chain state.  This has worked reasonably well, but has a few major limitations that we’re now running into: 

1. We built the chain state to buffer writes for an entire block at once, and we built the `Component` trait to have three distinct phases: stateless verification, stateful verification, and execution. This design requires that all of the stateful verification be performed before any modifications are made to the chain state.  This means that we don’t need to abort partial writes from a transaction that failed partway through execution, but it means that transactions cannot contain sequential actions that modify chain state.  This design has a bunch of upsides – if stateful verification is separated from execution, it can be performed in advance and in parallel with the actual writes.  However, it means that we can’t handle IBC the way other chains would, e.g., it’s not possible to relay a counterparty chain’s state update AND a proof of inclusion of some data relative to the new state.  This would have to be spread over two blocks, which is sad.

2. We don’t have any way to iterate over the keys of the key/value store.  Instead, we make lookup k/v pairs where the value is a list of the keys we want to index (e.g., key whose value is a list of all of the validator identity keys), and writing updates to the index in this way has quadratic behavior.  This hasn’t been a big deal so far but it basically prevents us from handling situations with many keys (e.g., the DEX would like to have a list of positions).

3. We don’t actually have a consistent definition of what the keys actually are!  The JMT models keys as hashes, while our ICS23 proof specification models keys as strings (which are prehashed when used in the JMT).  Our internal logic sometimes uses strings and sometimes uses key hashes.

4.  We don’t have any way to delete entries from the key/value store.  So instead, we make a bunch of ad-hoc hacks to record tombstone values at the application level (e.g., `DelibleNoteSource`).  This adds a lot of work defining the ad-hoc tombstones, but doesn’t give us any consistent way to handle deletions.  Because the tombstones are actually existing in the Merkle tree, it also means that we can’t form proofs of nonexistence, and we can never reclaim space from the JMT as we delete data.

[...]

PR #1495, opened 2022-09-30
Add stub impls for Storage2 crate

### LINKED ISSUE #1489: Define a stub version of the State V3 API sketch with `todo!()` implementations
Following #1488, we should fill in the API sketch in #1454, and make whatever changes come up while filling it in.

The initial API sketch was as follows:

- `struct Storage` - as now, this a wrapper around the underlying RocksDB instance;
- `struct State` - this represents a lightweight copy-on-write fork of the chain state, implemented as a RYW cache over a pinned JMT version;
- `struct StateTransaction` - this represents a transactional set of changes to a `State` fork, implemented as a RYW cache over a `State`.
- `trait StateRead` - read access to the state, implemented for `State`, `StateTransaction`;
- `trait StateWrite` - write access to the state, implemented for `StateTransaction`.

The two main use flows are:

- Read access, e.g., for RPC:
  1. From a `Storage` instance, fork the current version of the state to obtain a `State` instance;
  2. Use the `State` as a `StateRead` to respond to queries;
- Write access:
  1. From a `Storage` instance, fork the current version of the state to obtain a `State` instance;
  2. For each transactional set of writes (e.g., a tx execution), create a new `StateTransaction`;
  3. Use the `StateTransaction` as a `StateRead + StateWrite` during tx execution;
  4. Commit the `StateTransaction` back to its parent `State`;
  5. Commit the `State` to the `Storage` (or, don't, in the case of the mempool);

The API should have methods that support those flows.

zbuc:
Closes #1489

### DIFF
--- a/storage2/src/lib.rs
+++ b/storage2/src/lib.rs
@@ -1,1 +1,15 @@
+// Required to ensure that Rust can infer a Send bound inside the TCT
+#![recursion_limit = "256"]
 
+use jmt::WriteOverlay;
+use std::sync::Arc;
+use tokio::sync::RwLock;
+
+mod metrics;
+mod state;
+mod storage;
+
+pub use crate::metrics::register_metrics;
+pub use storage::Storage;
+
+pub type State = Arc<RwLock<WriteOverlay<Storage>>>;
--- a/storage2/src/metrics.rs
+++ b/storage2/src/metrics.rs
@@ -1,0 +1,27 @@
+//! Crate-specific metrics functionality.
+//!
+//! This module re-exports the contents of the `metrics` crate.  This is
+//! effectively a way to monkey-patch the functions in this module into the
+//! `metrics` crate, at least from the point of view of the other code in this
+//! crate.
+//!
+//! Code in this crate that wants to use metrics should `use crate::metrics;`,
+//! so that this module shadows the `metrics` crate.
+//!
+//! This trick is probably good to avoid in general, because it could be
+//! confusing, but in this limited case, it seems like a clean option.
+
+pub use metrics::*;
+
+/// Registers all metrics used by this crate.
+pub fn register_metrics() {
+    register_gauge!(TCT_SIZE_BYTES);
+    describe_gauge!(
+        TCT_SIZE_BYTES,
+        Unit::Bytes,
+        "The size of the serialized TCT in bytes"
+    );
+}
+
+// TODO: rename after we remove the old storage crate
+pub const TCT_SIZE_BYTES: &str = "penumbra_storage2_tct_size_bytes";
--- a/storage2/src/state.rs
+++ b/storage2/src/state.rs
@@ -24,6 +24,10 @@ impl<'a, R: TreeReader + TreeWriter> State<'a, R> {
             jmt: todo!(),
         }
     }
+
+    pub fn begin_transaction(&mut self) -> StateTransaction<'a, R> {
+        StateTransaction::new()
+    }
 }
 
 impl<'a, R: TreeReader + TreeWriter> StateRead for State<'a, R> {
@@ -39,5 +43,5 @@ pub trait StateRead {
 
 pub trait StateWrite {
     /// Copy-on-write put
-    fn put(&mut self, key: jmt::KeyHash, value: jmt::OwnedValue) -> Self;
+    fn put(&mut self, key: jmt::KeyHash, value: Option<jmt::OwnedValue>) -> Self;
 }
--- a/storage2/src/state.rs
+++ b/storage2/src/state.rs
@@ -1,0 +1,43 @@
+use std::collections::HashMap;
+
+use jmt::{
+    storage::{TreeReader, TreeWriter},
+    JellyfishMerkleTree,
+};
+
+mod transaction;
+use transaction::Transaction as StateTransaction;
+
+/// State is a lightweight copy-on-write fork of the chain state,
+/// implemented as a RYW cache over a pinned JMT version.
+pub(crate) struct State<'a, R: TreeReader + TreeWriter> {
+    cache: HashMap<jmt::KeyHash, jmt::OwnedValue>,
+    jmt_version: jmt::Version,
+    jmt: &'a JellyfishMerkleTree<'a, R>,
+}
+
+impl<'a, R: TreeReader + TreeWriter> State<'a, R> {
+    pub fn new() -> Self {
+        Self {
+            cache: todo!(),
+            jmt_version: todo!(),
+            jmt: todo!(),
+        }
+    }
+}
+
+impl<'a, R: TreeReader + TreeWriter> StateRead for State<'a, R> {
+    fn get(&self, key: jmt::KeyHash) -> Option<&jmt::OwnedValue> {
+        todo!()
+    }
+}
+
+pub trait StateRead {
+    /// Get
+    fn get(&self, key: jmt::KeyHash) -> Option<&jmt::OwnedValue>;
+}
+
+pub trait StateWrite {
+    /// Copy-on-write put
+    fn put(&mut self, key: jmt::KeyHash, value: jmt::OwnedValue) -> Self;
+}
--- a/storage2/src/state/transaction.rs
+++ b/storage2/src/state/transaction.rs
@@ -1,0 +1,45 @@
+use std::collections::HashMap;
+
+use jmt::storage::{TreeReader, TreeWriter};
+
+use super::{State, StateWrite};
+
+/// Represents a transactional set of changes to a `State` fork,
+/// implemented as a RYW cache over a `State`.
+pub(crate) struct Transaction<'a, R: TreeReader + TreeWriter> {
+    // TODO: should higher-level types be used here instead of jmt types?
+    cache: HashMap<jmt::KeyHash, jmt::OwnedValue>,
+    unwritten_changes: Vec<(jmt::KeyHash, jmt::OwnedValue)>,
+    state: &'a State<'a, R>,
+}
+
+impl<'a, R: TreeReader + TreeWriter> Transaction<'a, R> {
+    pub fn new() -> Self {
+        Self {
+            cache: todo!(),
+            unwritten_changes: todo!(),
+            state: todo!(),
+        }
+    }
+
+    pub fn begin_transaction(&mut self) {
+        // Wipe the unwritten changes
+        todo!()
+    }
+
+    pub fn end_transaction(&mut self) {
+        // Write unwritten_changes to the state
+        todo!()
+    }
+}
+
+impl<'a, R: TreeReader + TreeWriter> StateWrite for Transaction<'a, R> {
+    fn put(&mut self, key: jmt::KeyHash, value: jmt::OwnedValue) -> Transaction<'a, R> {
+        todo!()
+    }
+}
+
+pub trait StateRead {
+    /// Get
+    fn get(&self, key: jmt::KeyHash) -> Option<&jmt::OwnedValue>;
+}
--- a/storage2/src/state/transaction.rs
+++ b/storage2/src/state/transaction.rs
@@ -1,5 +1,7 @@
 use std::collections::HashMap;
 
+use anyhow::Result;
+
 use jmt::storage::{TreeReader, TreeWriter};
 
 use super::{State, StateWrite};
@@ -22,19 +24,15 @@ impl<'a, R: TreeReader + TreeWriter> Transaction<'a, R> {
         }
     }
 
-    pub fn begin_transaction(&mut self) {
-        // Wipe the unwritten changes
-        todo!()
-    }
-
-    pub fn end_transaction(&mut self) {
+    pub fn commit(self) -> Result<()> {
         // Write unwritten_changes to the state
+        // `self` will be consumed afterwards
         todo!()
     }
 }
 
 impl<'a, R: TreeReader + TreeWriter> StateWrite for Transaction<'a, R> {
-    fn put(&mut self, key: jmt::KeyHash, value: jmt::OwnedValue) -> Transaction<'a, R> {
+    fn put(&mut self, key: jmt::KeyHash, value: Option<jmt::OwnedValue>) -> Transaction<'a, R> {
         todo!()
     }
 }
--- a/storage2/src/storage.rs
+++ b/storage2/src/storage.rs
@@ -1,0 +1,227 @@
+use std::{path::PathBuf, sync::Arc};
+
+use ::metrics::gauge;
+use anyhow::Result;
+use futures::future::BoxFuture;
+use jmt::{
+    storage::{Node, NodeBatch, NodeKey, TreeReader, TreeWriter},
+    WriteOverlay,
+};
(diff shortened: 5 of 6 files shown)


storage2/src/state/transaction.rs
```diff
@@ -0,0 +1,45 @@
...
+
+    pub fn begin_transaction(&mut self) {
+        // Wipe the unwritten changes
+        todo!()
+    }
+
+    pub fn end_transaction(&mut self) {
+        // Write unwritten_changes to the state
+        todo!()
+    }
```
plaidfinch, line 33: Why are these two methods that take `Transaction` as `&mut self`? It seems to me that it would make more sense to model a transaction as a separate object whose life ends when it is committed, rather than something reusable.

storage2/src/state/transaction.rs
```diff
@@ -0,0 +1,45 @@
...
+}
+
+impl<'a, R: TreeReader + TreeWriter> StateWrite for Transaction<'a, R> {
+    fn put(&mut self, key: jmt::KeyHash, value: jmt::OwnedValue) -> Transaction<'a, R> {
```
plaidfinch, line 37: When deletion happens, this will probably be `Option<jmt::OwnedValue>`.

storage2/src/state.rs
```diff
@@ -0,0 +1,43 @@
...
+}
+
+pub trait StateRead {
+    /// Get
+    fn get(&self, key: jmt::KeyHash) -> Option<&jmt::OwnedValue>;
+}
+
+pub trait StateWrite {
+    /// Copy-on-write put
+    fn put(&mut self, key: jmt::KeyHash, value: jmt::OwnedValue) -> Self;
```
plaidfinch, line 42: I don't think we will ever want to call `put` outside a transaction?

storage2/src/state/transaction.rs
plaidfinch: It seems to me that the API ought to look more like what `sqlx` does to model transactions. See [how a transaction is created with `Pool::begin`](https://docs.rs/sqlx/latest/sqlx/struct.Pool.html#method.begin) and [how a transaction is ended with `Transaction::commit`](https://docs.rs/sqlx/latest/sqlx/struct.Transaction.html#method.commit).

storage2/src/state.rs
zbuc: @plaidfinch are you suggesting to implement it directly on the `Transaction` type rather than in a trait? I think we might need a `StateWrite` implementation for `State` when the transaction gets applied, or maybe @hdevalence had something else in mind in https://github.com/penumbra-zone/penumbra/issues/1489

plaidfinch: I don't think we should have a `StateWrite` implementation for `State`; while it would be possible to provide one, that would mean you could put a raw `State` in places where there should always be a transaction. Instead, `State` could have an inherent method of the same signature, if it's needed to implement transaction application?

storage2/src/metrics.rs
```diff
@@ -0,0 +1,27 @@
...
+    register_gauge!(TCT_SIZE_BYTES);
+    describe_gauge!(
+        TCT_SIZE_BYTES,
+        Unit::Bytes,
+        "The size of the serialized TCT in bytes"
+    );
+}
+
+// TODO: rename after we remove the old storage crate
+pub const TCT_SIZE_BYTES: &str = "penumbra_storage2_tct_size_bytes";
```
hdevalence, line 27: We shouldn't carry over this code, because it's related to the details of the existing implementation, not the sketch of the new API.

storage2/src/state.rs
```diff
@@ -0,0 +1,47 @@
...
+
+mod transaction;
+use transaction::Transaction as StateTransaction;
+
+/// State is a lightweight copy-on-write fork of the chain state,
+/// implemented as a RYW cache over a pinned JMT version.
+pub(crate) struct State<'a, R: TreeReader + TreeWriter> {
+    cache: HashMap<jmt::KeyHash, jmt::OwnedValue>,
+    jmt_version: jmt::Version,
+    jmt: &'a JellyfishMerkleTree<'a, R>,
```
hdevalence, line 16: We can't know what data the `State` will need to maintain until we design how it works in #1490, so we should not fill it in with fields that might be wrong.

storage2/src/state.rs
```diff
@@ -0,0 +1,47 @@
...
+}
+
+impl<'a, R: TreeReader + TreeWriter> State<'a, R> {
```
hdevalence, line 19: We can't have the `State` include a lifetime bound; that will make using it unworkable, since now every downstream consumer has to manage another object's lifetime and the relation between that object and the `State`.

storage2/src/state/transaction.rs
```diff
@@ -0,0 +1,43 @@
...
+
+use super::{State, StateWrite};
+
+/// Represents a transactional set of changes to a `State` fork,
+/// implemented as a RYW cache over a `State`.
+pub(crate) struct Transaction<'a, R: TreeReader + TreeWriter> {
+    // TODO: should higher-level types be used here instead of jmt types?
+    cache: HashMap<jmt::KeyHash, jmt::OwnedValue>,
+    unwritten_changes: Vec<(jmt::KeyHash, jmt::OwnedValue)>,
+    state: &'a State<'a, R>,
```
hdevalence, line 15: We shouldn't fill in the internals here, because we haven't designed how it should work, and we can't do that (#1490) until we have an API sketch.

The storage `Transaction` shouldn't take a generic `TreeReader + TreeWriter`, since we aren't plugging in an arbitrary storage backend.

Having a bound on the lifetime of the `State` the transaction is modifying seems reasonable -- but perhaps we should be holding a `&'a mut State`, to enforce that there can only be one live `Transaction` at a time?

storage2/src/state/transaction.rs
```diff
@@ -0,0 +1,43 @@
...
+}
+
+impl<'a, R: TreeReader + TreeWriter> Transaction<'a, R> {
+    pub fn new() -> Self {
+        Self {
+            cache: todo!(),
+            unwritten_changes: todo!(),
+            state: todo!(),
+        }
+    }
```
hdevalence, line 25: We can't have a `Transaction::new`, because the transaction has to be created with reference to a `State` instance.  We should just have one method on the `State` that produces a `Transaction` (beginning its lifecycle) and a `commit()` method that consumes it (ending its lifecycle).

storage2/src/state/transaction.rs
```diff
@@ -0,0 +1,43 @@
...
+
+    pub fn commit(self) -> Result<()> {
+        // Write unwritten_changes to the state
+        // `self` will be consumed afterwards
+        todo!()
+    }
```
hdevalence, line 31: Why does this return a `Result<()>`? Since we're writing changes to the in-memory `State` fork, where does the fallibility come from?

Also, should we be calling this `commit`, if we're not yet writing data to durable storage? Is it potentially confusing to have both `Transaction::commit` and `State::commit`, with different semantics about durability?

Should we call this `Transaction::finish` instead? Or would having two different names be more confusing?

storage2/src/storage.rs
```diff
@@ -0,0 +1,227 @@
...
+                    let mut opts = Options::default();
+                    opts.create_if_missing(true);
+                    opts.create_missing_column_families(true);
+
+                    Ok(Self(Arc::new(DB::open_cf(&opts, path, ["jmt", "nct"])?)))
+                })
+            })
+            .unwrap()
+            .await
+            .unwrap()
```
hdevalence, line 38: We should snip out this code until we fill in the implementation, so we can focus on the API structure. For instance, we won't want to have just those column families, so the existing code doesn't actually fit.

storage2/src/storage.rs
```diff
@@ -0,0 +1,227 @@
...
+    }
+
+    /// Returns the latest version (block height) of the tree recorded by the
+    /// `Storage`, or `None` if the tree is empty.
+    pub async fn latest_version(&self) -> Result<Option<jmt::Version>> {
+        Ok(self
+            .get_rightmost_leaf()
+            .await?
+            .map(|(node_key, _)| node_key.version()))
+    }
```
hdevalence, line 48: Assuming we want to maintain the `latest_version` method (why? what do we need it for?), we shouldn't fill in an implementation yet.

storage2/src/storage.rs
```diff
@@ -0,0 +1,227 @@
...
+        let version = self
+            .latest_version()
+            .await?
+            .unwrap_or(WriteOverlay::<Storage>::PRE_GENESIS_VERSION);
+
+        tracing::debug!("creating state for version {}", version);
+        Ok(Arc::new(RwLock::new(WriteOverlay::new(
+            self.clone(),
+            version,
+        ))))
```
hdevalence, line 63: This implementation isn't correct, because a `State` isn't an Arc<... any more, so we should delete it until we define what a `State` actually is and what data it consists of.

storage2/src/storage.rs
```diff
@@ -0,0 +1,227 @@
...
+                    if let Some(tct_bytes) = db.get_cf(nct_cf, "tct")? {
+                        Ok(bincode::deserialize(&tct_bytes)?)
+                    } else {
+                        Ok(tct::Tree::new())
+                    }
+                })
+            })
+            .unwrap()
+            .await?
+    }
```
hdevalence, line 117: These methods violate the design principles of the rest of the API (e.g., they aren't versioned, and they can screw up someone else's view of the `State`), so we should delete them, and figure out how to record the NCT through the new API we're designing.

storage2/src/storage.rs
```diff
@@ -0,0 +1,227 @@
...
+                        }
+                        Ok(ret)
+                    })
+                })
+                .unwrap()
+                .await
+                .unwrap()
+        })
+    }
+}
```
hdevalence, line 227: We shouldn't implement this trait on the `Storage` type, and in any case the trait will have to change, because we'll be using the non-`async` JMT API, so we should delete this implementation.

storage2/src/lib.rs
```diff
@@ -1 +1,15 @@
...
+use tokio::sync::RwLock;
+
+mod metrics;
+mod state;
+mod storage;
+
+pub use crate::metrics::register_metrics;
+pub use storage::Storage;
+
+pub type State = Arc<RwLock<WriteOverlay<Storage>>>;
```
hdevalence, line 15: We don't want a `State` to be an `Arc<...<`, we want it to be a new struct, whose contents are TBD based on the API we design in this PR.

storage2/src/state.rs
```diff
@@ -0,0 +1,47 @@
...
+    storage::{TreeReader, TreeWriter},
+    JellyfishMerkleTree,
+};
+
+mod transaction;
+use transaction::Transaction as StateTransaction;
+
+/// State is a lightweight copy-on-write fork of the chain state,
+/// implemented as a RYW cache over a pinned JMT version.
+pub(crate) struct State<'a, R: TreeReader + TreeWriter> {
```
hdevalence, line 13: This struct shouldn't be `pub(crate)`, since it's the public API we want people to use. We also want it to have a `'static` lifetime (i.e., no lifetime bounds), because it should be possible for it to live as long as the user wants (and maintain a consistent view of the pinned snapshot in that time), and we don't want it to have any trait bounds required (we're not generic over the storage, we are the storage).

storage2/src/state.rs
```diff
@@ -0,0 +1,47 @@
...
+impl<'a, R: TreeReader + TreeWriter> StateRead for State<'a, R> {
+    fn get(&self, key: jmt::KeyHash) -> Option<&jmt::OwnedValue> {
+        todo!()
+    }
+}
+
+pub trait StateRead {
+    /// Get
+    fn get(&self, key: jmt::KeyHash) -> Option<&jmt::OwnedValue>;
+}
```
hdevalence, line 42: Per #1454, a goal of the redesign is that we want to be operating on string keys, so we shouldn't be taking a `jmt::KeyHash` here, we should be taking a string key. Since we don't need to hold on to the key past the length of the query, we don't need to take ownership, so that could be a `key: &str`.

We don't want to return a `jmt::OwnedValue`, because we don't want API users to have to reference the JMT API. We could return a `Vec<u8>` instead.

hdevalence: Also, do we want the get method to return a byte vector at all? That's in some sense the "underlying" operation, but it's also the least frequently used. We mostly want domain types, or sometimes proto types.

What if instead, we renamed the byte-returning method as get_raw (a required method), and then had provided methods get_proto (returning a proto type) and get (returning a domain type).

Then the trait implementor would only have to implement get_raw, but the user could call get and work with domain types.

hdevalence: Finally, I think that this method can't be a synchronous read, but will have to be an async one. We could either use the async_trait macro, or return an associated type. Using async_trait seems easier to handle the provided methods.

storage2/src/lib.rs
```diff
@@ -1 +1,15 @@
+// Required to ensure that Rust can infer a Send bound inside the TCT
```
hdevalence, line 1: We don't need this in this crate, I think, because we're not using the TCT here.

storage2/src/state/transaction.rs
```diff
@@ -0,0 +1,43 @@
...
+}
+
+impl<'a, R: TreeReader + TreeWriter> StateWrite for Transaction<'a, R> {
+    fn put(&mut self, key: jmt::KeyHash, value: Option<jmt::OwnedValue>) -> Transaction<'a, R> {
+        todo!()
```
hdevalence, line 36: As with StateRead, we should be taking string keys, which will have to be owned values (hence String).

We'll also want a similar API, with put, put_proto, and put_raw.

Taking the value as an Option is a bad API for general purpose use like this, because it forces every caller to Option-wrap their data, and the reader of every call site to understand that put(Some) means "actually put" while put(None) means "delete". Instead, we should just have a distinct delete() method.

hdevalence: We also shouldn't return the transaction. Since we're modifying it with &mut self, the caller doesn't need a return value, and the Transaction isn't cloneable.

hdevalence:
I think we still have some work to do on this sketch, so I reopened #1489

storage2/src/state/transaction.rs
zbuc: > Should we call this Transaction::finish instead? Or would having two different names be more confusing?

I don't have a strong opinion here. Reflexively we're used to "committing" database transactions, but you're right that there are different semantics to this state transaction vs a DB transaction.

hdevalence: Cool, let's keep it as `commit` then, I think it's a reasonable choice, we should just be clear that we're committing to our parent in-memory state fork, rather than committing that fork to the storage.

storage2/src/storage.rs
zbuc: @hdevalence why shouldn't we implement the `TreeReader` trait on `Storage`? We still have need to read the JMT in `Storage`, i.e. to get the JMT version

PR #1498, opened 2022-10-03
Use pro-rata output in constructing SwapClaimProof

### LINKED ISSUE #1497: Swaps for non-existent liquidity pools irrecoverably fail
**Describe the bug**
If you perform a swap for a non-existent liquidity pool, e.g. pizza:penumbra, you receive a swap NFT that can never be redeemed. 

**To Reproduce**
1. Swap pizza for penumbra
2. Validate that you have a swap NFT in your balance
3. Validate that tx sweep doesn't redeem the swap NFT either

**Expected behavior**
During end_block for the block in which the swap took place, swaps for non-existent liquidity pools should have a "failed" status batch swap output created so the individual swaps can retrieve their input assets.

zbuc:
Fixes #1497

### DIFF
--- a/crypto/src/proofs/transparent.rs
+++ b/crypto/src/proofs/transparent.rs
@@ -311,8 +311,8 @@ pub struct SwapClaimProof {
     pub delta_2_i: u64,
 
     // Describes output amounts
-    pub lambda_1: u64,
-    pub lambda_2: u64,
+    pub lambda_1_i: u64,
+    pub lambda_2_i: u64,
 
     // Describes first output note (lambda 1)
     pub note_blinding_1: Fq,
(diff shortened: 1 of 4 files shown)


plaidfinch:
@zbuc should we also have a guard on tx submission that prevents swaps from being submitted when there's no liquidity to begin with? you could fail the transaction entirely, in `check_tx_stateful`, rather than turning it into a swap NFT that needs to be redeemed.

zbuc:
> @zbuc should we also have a guard on tx submission that prevents swaps from being submitted when there's no liquidity to begin with? you could fail the transaction entirely, in `check_tx_stateful`, rather than turning it into a swap NFT that needs to be redeemed.

@plaidfinch I considered that but decided against it because the available liquidity could change during the block the swap is submitted, i.e. there's no way to tell with certainty whether a swap will succeed or fail when you submit it. But maybe there could be a prompt instead? "There doesn't seem to be liquidity for this trading pair and the swap may fail as a result, are you sure you wish to submit?"

plaidfinch:
Oh, that's a good point. I think that we should not do interactivity here, per previous conversations with @hdevalence where we decided that interactivity in pcli is probably not a good thing. Let's just have the swap generate a failed NFT that can be immediately claimed back, like you already did :)

PR #1504, opened 2022-10-04
Storage stubs continued

### LINKED ISSUE #1489: Define a stub version of the State V3 API sketch with `todo!()` implementations

zbuc:
Incorporates @hdevalence feedback on #1495 

Closes #1489

### DIFF
--- a/storage2/src/state.rs
+++ b/storage2/src/state.rs
@@ -1,11 +1,11 @@
-use std::fmt::Debug;
-
 use async_trait::async_trait;
 
-use penumbra_proto::{Message, Protobuf};
-
+mod read;
 mod transaction;
+mod write;
+pub use read::StateRead;
 pub use transaction::Transaction as StateTransaction;
+pub use write::StateWrite;
 
 /// State is a lightweight copy-on-write fork of the chain state,
 /// implemented as a RYW cache over a pinned JMT version.
@@ -39,70 +39,4 @@ impl StateRead for State {
     fn get_raw(&self, key: String) -> Option<Vec<u8>> {
         todo!()
     }
-}
-
-#[async_trait]
-pub trait StateRead {
-    /// Get
-    fn get_raw(&self, key: String) -> Option<Vec<u8>>;
-
-    /// Gets a domain type from the State.
-    fn get<D, P>(&self, key: String) -> Option<D>
-    where
-        D: Protobuf<P>,
-        // TODO: does this get less awful if P is an associated type of D?
-        P: Message + Default,
-        P: From<D>,
-        D: TryFrom<P> + Clone + Debug,
-        <D as TryFrom<P>>::Error: Into<anyhow::Error>,
-    {
-        todo!()
-    }
-
-    /// Gets a proto type from the State.
-    fn get_proto<D, P>(&self, key: String) -> Option<P>
-    where
-        D: Protobuf<P>,
-        // TODO: does this get less awful if P is an associated type of D?
-        P: Message + Default,
-        P: From<D>,
-        D: TryFrom<P> + Clone + Debug,
-        <D as TryFrom<P>>::Error: Into<anyhow::Error>,
-    {
-        todo!()
-    }
-}
-
-pub trait StateWrite {
-    /// Copy-on-write put
-    fn put_raw(&mut self, key: String, value: Vec<u8>);
-
-    /// Sets a domain type from the State.
-    fn put<D, P>(&self, key: String, value: D)
-    where
-        D: Protobuf<P>,
-        // TODO: does this get less awful if P is an associated type of D?
-        P: Message + Default,
-        P: From<D>,
-        D: TryFrom<P> + Clone + Debug,
-        <D as TryFrom<P>>::Error: Into<anyhow::Error>,
-    {
-        todo!()
-    }
-
-    /// Puts a proto type on the State.
-    fn put_proto<D, P>(&self, key: String, value: P)
-    where
-        D: Protobuf<P>,
-        // TODO: does this get less awful if P is an associated type of D?
-        P: Message + Default,
-        P: From<D>,
-        D: TryFrom<P> + Clone + Debug,
-        <D as TryFrom<P>>::Error: Into<anyhow::Error>,
-    {
-        todo!()
-    }
-
-    /// Delete a key from state.
-    fn delete(&mut self, key: String);
 }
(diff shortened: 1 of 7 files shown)


storage2/src/state.rs
```diff
@@ -1,47 +1,82 @@
...
+    pub fn get_proto<D, P>(&self, key: String) -> Option<P>
+    where
+        D: Protobuf<P>,
+        // TODO: does this get less awful if P is an associated type of D?
+        P: Message + Default,
+        P: From<D>,
+        D: TryFrom<P> + Clone + Debug,
+        <D as TryFrom<P>>::Error: Into<anyhow::Error>,
+    {
+        todo!()
```
hdevalence, line 59: Let's put these as [provided methods](https://doc.rust-lang.org/book/ch10-02-traits.html#default-implementations) on the `StateRead` trait, implemented in terms of the required `get_raw` method, so that they're available for every `StateRead` implementation.

storage2/src/state.rs
```diff
@@ -1,47 +1,82 @@
...
 pub trait StateRead {
     /// Get
-    fn get(&self, key: jmt::KeyHash) -> Option<&jmt::OwnedValue>;
+    fn get_raw(&self, key: String) -> Option<Vec<u8>>;
 }
 
 pub trait StateWrite {
     /// Copy-on-write put
-    fn put(&mut self, key: jmt::KeyHash, value: Option<jmt::OwnedValue>) -> Self;
+    fn put(&mut self, key: String, value: jmt::OwnedValue);
```
hdevalence, line 78: Let's make this `put_raw`, have it take a `Vec<u8>`, and then add provided `put_proto`, `put` methods (analogous to `get`, `get_proto` on `StateRead`) to the `StateWrite` trait.

storage2/src/state/transaction.rs
```diff
@@ -1,43 +1,33 @@
...
-pub(crate) struct Transaction<'a, R: TreeReader + TreeWriter> {
-    // TODO: should higher-level types be used here instead of jmt types?
-    cache: HashMap<jmt::KeyHash, jmt::OwnedValue>,
-    unwritten_changes: Vec<(jmt::KeyHash, jmt::OwnedValue)>,
-    state: &'a State<'a, R>,
+pub struct Transaction {
+    // TODO determine which fields to include (#1490)
+    // cache: HashMap<jmt::KeyHash, jmt::OwnedValue>,
+    // unwritten_changes: Vec<(jmt::KeyHash, jmt::OwnedValue)>,
+    // state: &'a mut State<'a>,
```
hdevalence, line 9: One way we could enforce that only one `Transaction` is live for each `State` instance at a time would be to add a `&mut State` reference to the `Transaction`, so that it becomes
```
pub struct Transaction<'s> {
    state: &'s mut State,
    // ...
}
```

Question: do we want to do this?

The advantage is that we can guarantee that there's only one `Transaction` at a time, and also that in `Transaction::commit`, we can directly edit the `State` (by having `pub(super) State::foo(&mut self, ...)` methods and calling them in `Transaction::commit`).

The downside is that now, if we pass around a `Transaction`, we're also passing around a lifetime bound. This might be inconvenient if we wanted to, e.g., spawn multiple tasks using the same `Transaction`, but we probably don't want to do that, because of concurrency reasons.

On balance, it seems like this could be a good API choice?

storage2/src/state/transaction.rs
```diff
@@ -1,43 +1,33 @@
...
+    fn delete(&mut self, key: String) {
         todo!()
     }
 }
 
 pub trait StateRead {
     /// Get
-    fn get(&self, key: jmt::KeyHash) -> Option<&jmt::OwnedValue>;
+    fn get(&self, key: String) -> Option<&jmt::OwnedValue>;
 }
```
hdevalence, line 33: Wait, do we have multiple `StateRead` definitions?

I think it'd be good to split out the trait definitions into `state/read.rs` and `state/write.rs`, then re-exporting the traits up to the crate root.

storage2/src/state/transaction.rs
zbuc: I am curious what the general usage pattern in our component code will look like.

i.e. are we generally going to create and commit a transaction within the same method, or would a component ever create a transaction in `begin_block` and only commit it during `end_block`? That would make using the `&mut State` reference difficult.

hdevalence: The overall goal is to be able to combine verification and execution, by starting to execute a (blockchain) transaction, but discard any writes it made if it errors during the execution.  The (storage) transaction is a way to accomplish this, by starting a new (storage) transaction when beginning execution of a (blockchain) transaction, then committing it only when the execution completes successfully.

So I don't think we'll ever want to have a component create a `Transaction` itself, since it's only going to be one part of the whole execution.  We'll probably want to have the `App` construct a `Transaction` when starting to process a (blockchain) transaction, then pass it in to all the components.  For `begin_block` and `end_block`, the fallibility / "transactionality" is less important, because that functionality should be infallible anyways, but I think it would make sense to have the `App` construct a `Transaction` at the start of `begin_block`/`end_block` and commit it at the end of the same method.

zbuc: If the `App` owns the `Transaction`, then I think it makes sense to enforce that only one can exist at a time. The fact that it dissuades us from using transactions concurrently sounds like a positive thing IMO.

hdevalence: Cool, sounds good to me.

PR #1505, opened 2022-10-04
Refactor view_from_perspective to infallibly return an ActionView

### LINKED ISSUE #1496: TransactionView enhancements
Various to-do/enhancement/nice-to-have items around `TransactionView`:

- [ ] Refactor to make `view_from_perspective` infallibly return an `ActionView`
- [ ] Refactor to simply use a `String` instead of a `MemoPlaintext`

aubrika:
Closes #1496

### DIFF
--- a/transaction/src/view/action_view/output_view.rs
+++ b/transaction/src/view/action_view/output_view.rs
@@ -1,6 +1,14 @@
 use penumbra_crypto::{Note, PayloadKey};
 
-pub struct OutputView {
-    pub decrypted_note: Note,
-    pub decrypted_memo_key: PayloadKey,
+use crate::action::Output;
+#[allow(clippy::large_enum_variant)]
+pub enum OutputView {
+    Visible {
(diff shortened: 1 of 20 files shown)


transaction/src/view/action_view/output_view.rs
```diff
@@ -1,6 +1,14 @@
...
+#[allow(clippy::large_enum_variant)]
+pub enum OutputView {
+    Visible {
+        output: Output,
+        decrypted_note: Note,
+        decrypted_memo_key: PayloadKey,
+    },
+    Opaque {
+        output: Output,
+    },
```
hdevalence, line 13: These aren't connected to the protos yet, right?

aubrika: They aren't, and I'm not sufficiently familiar with the proto syntax to know how this kind of enum can best be represented in the proto crate - do you have an example?

PR #1506, opened 2022-10-05
Start implementing Snapshot

zbuc:


storage2/src/snapshot.rs
```diff
@@ -0,0 +1,29 @@
+/// Snapshots maintain a point-in-time view of the underlying storage, suitable
+/// for read-only access by multiple threads, i.e. RPC calls.
+///
+/// This is implemented as a wrapper around a [RocksDB snapshot](https://github.com/facebook/rocksdb/wiki/Snapshot)
+/// with an associated JMT version number for the snapshot.
+pub(crate) struct Snapshot {
+    rocksdb_snapshot: rocksdb::Snapshot<'static>,
```
hdevalence, line 7: Could we add a `TODO` comment highlighting that the `'static` is a temporary hack, and that we need to find a workaround as a separate issue?

storage2/src/state.rs
```diff
@@ -7,36 +9,37 @@ pub use read::StateRead;
...
         }
     }
 
     pub fn begin_transaction(&mut self) -> StateTransaction {
-        StateTransaction{
-            // cache: todo!(),
-            // unwritten_changes: todo!(),
-            // state: self,
-        }
+        StateTransaction {}
```
hdevalence, line 30: Should we be passing `&mut self` into the `StateTransaction`?

storage2/src/storage.rs
```diff
@@ -3,10 +3,14 @@ use std::{path::PathBuf, sync::Arc};
...
 use rocksdb::DB;
 
+use crate::snapshot::Snapshot;
 use crate::State;
 
-#[derive(Clone, Debug)]
-pub struct Storage(Arc<DB>);
+#[derive(Clone)]
+pub struct Storage {
+    snapshot: Arc<Snapshot>,
```
hdevalence, line 11: Maybe this should be called `latest_snapshot`, for clarity?  Also, will we need to wrap this in a `parking_lot::RwLock` so that we can insert the new latest snapshot?

storage2/src/storage.rs
```diff
@@ -21,7 +25,7 @@ impl Storage {
 
     /// Returns a new [`State`] on top of the latest version of the tree.
     pub async fn state(&self) -> Result<State> {
-        todo!()
+        Ok(State::new(self.snapshot.clone()))
```
hdevalence, line 28: Since this isn't fallible, could we return a `State` infallibly here?

PR #1511, opened 2022-10-06
ICS20 withdrawal implementation

### LINKED ISSUE #985: Process `Withdraw` actions in the ICS20 component
To handle a withdraw action, we need to:

- [ ] Create an appropriate `FungibleTokenPacketData` from the action contents;
- [ ] Add the packet to the chain state in a way that's legible to the counterparty chain;
- [ ] Somehow register a hook to process either an acknowledgement or a timeout;
- [ ] Add logic to return funds in the event of a timeout.

Spec link: https://github.com/cosmos/ibc/tree/master/spec/app/ics-020-fungible-token-transfer#packet-relay

avahowell:
this PR implements ICS20 withdrawals, allowing us to commit withdrawal packets to penumbra's state.

Closes #985

### DIFF
--- a/component/src/ibc/component/state_key.rs
+++ b/component/src/ibc/component/state_key.rs
@@ -92,5 +92,5 @@ pub fn packet_commitment_by_port(
 }
 
 pub fn ics20_value_balance(channel_id: &ChannelId, asset_id: &asset::Id) -> String {
-    format!("ics2-value-balance/{}/{}", channel_id, asset_id)
+    format!("ics20-value-balance/{}/{}", channel_id, asset_id)
 }
--- a/component/src/ibc/component/state_key.rs
+++ b/component/src/ibc/component/state_key.rs
@@ -6,6 +6,8 @@ use ibc::{
     Height,
 };
 
+use penumbra_crypto::asset;
+
 use std::string::String;
 
 pub fn client_type(client_id: &ClientId) -> String {
@@ -87,4 +89,8 @@ pub fn packet_commitment_by_port(
         "commitments/ports/{}/channels/{}/packets/{}",
         port_id, channel_id, sequence
     )
+}
+
+pub fn ics20_value_balance(channel_id: &ChannelId, asset_id: &asset::Id) -> String {
+    format!("ics2-value-balance/{}/{}", channel_id, asset_id)
 }
--- a/component/src/ibc/packet.rs
+++ b/component/src/ibc/packet.rs
@@ -12,32 +12,69 @@ use ibc::core::ics24_host::identifier::ChannelId;
 use ibc::core::ics24_host::identifier::PortId;
 use ibc::timestamp::Timestamp;
 use penumbra_storage::StateExt;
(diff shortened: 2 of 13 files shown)


component/src/ibc/packet.rs
```diff
@@ -0,0 +1,165 @@
...
+
+mod private {
+    use super::*;
+
+    pub trait Sealed {}
+    impl Sealed for Checked {}
+    impl Sealed for Unchecked {}
+}
+
+pub struct IBCPacket<S: CheckStatus> {
```
avahowell, line 34: I thought this typestate programming would be nice for correctness, but it's kind of awkward right now since there's not an easy way to persist data across a tx execution cycle (from check <> execute). So right now this reruns the check in execute, which drops some performance on the floor

hdevalence: Hmm, let's make a note to come back to this once we land the state refactoring?

component/src/ibc/component/state_key.rs
```diff
@@ -88,3 +90,7 @@ pub fn packet_commitment_by_port(
         port_id, channel_id, sequence
     )
 }
+
+pub fn ics20_value_balance(channel_id: &ChannelId, asset_id: &asset::Id) -> String {
+    format!("ics2-value-balance/{}/{}", channel_id, asset_id)
```
avahowell, line 95: should be ics20

component/src/ibc/transfer.rs
```diff
@@ -28,6 +35,71 @@ impl ICS20Transfer {
...
+        // send packet
+        use crate::ibc::packet::SendPacket;
+        self.state.send_packet_check(ctx.clone(), packet).await?;
+
+        Ok(())
+    }
+
+    pub async fn withdrawal_execute(&mut self, ctx: Context, withdrawal: &ICS20Withdrawal) {
+        // create packet
+        let packet: IBCPacket<Unchecked> = withdrawal.clone().into();
```
avahowell, line 52: note: to avoid rechecking here or needing to communicate the withdrawal state from `withdrawal_check`, implement a `.assume_checked()` fn on IBCPacket that allows us to transform it to a Checked ibc packet without re-running all of the checks (which were already run in `withdrawal_check`) (thx @plaidfinch)

component/src/ibc/transfer.rs
```diff
@@ -28,6 +35,71 @@ impl ICS20Transfer {
...
+            .state
+            .send_packet_check(ctx.clone(), packet)
+            .await
+            .unwrap();
+
+        // TODO: we should probably encapsulate this in a type
+        let prefix = format!("{}/{}/", withdrawal.source_port, withdrawal.source_channel);
+        // we are the source if the denomination is not prefixed with the source port and source
+        // channel
+        let is_source = !withdrawal.denom.starts_with(&prefix);
```
avahowell, line 65: We should probably encapsulate this checking logic in a type. We could also use Penumbra's chain state to record a list of all IBC assets and reference that list, instead of doing string parsing, but that might be redundant

Issue #1516, opened 2022-10-06
Failure to acquire permissions to wallet directory on macOS


plaidfinch:
**Describe the bug**

On macOS 12.6 with SIP enabled, @mikayla-maki observes that a fresh installation of `pcli` results in a permissions error `Error: No such file or directory (os error 2)` when getting started with `cargo run --quiet --release --bin pcli keys generate`. Upon manually creating the directory `~/Library/Application\ Support/zone.penumbra.pcli/`, the error persists, but as a panic: `thread 'main' panicked at 'can create penumbra wallet archive directory: Os { code: 13, kind: PermissionDenied, message: "Permission denied" }', pcli/src/command/keys.rs:57:14`.

Permissions on that folder are:

```
drwxr-xr-x   3 mikaylamaki  staff    96B Oct  6 14:27 zone.penumbra.pcli
```

**To Reproduce**

Summary in @mikayla-maki's words:

> So, here's the full set of things I did:
> 1. go to: https://guide.penumbra.zone/main/pcli/install.html
> 2.1: I already have xcode command line tools installed, from the xcode application
> 2. brew install protobuf
> 3. git clone https://github.com/penumbra-zone/penumbra
> 4. cd penumbra && git fetch && git checkout 031-autonoe && cargo update
> 5. cargo build --release --bin pcli
> 6. cargo run --quiet --release --bin pcli keys generate
>
> At this step, I get an output of:
> - An unused import warning
> - The big fire warning
> - My seed phrase
> - the text Error: No such file or directory (os error 2)
>
> If I then execute:
>
> 7. mkdir ~/Library/Application\ Support/zone.penumbra.pcli/
> I get the same as the above, except with the error:
> thread 'main' panicked at 'can create penumbra wallet archive directory: Os { code: 13, kind: PermissionDenied, message: "Permission denied" }', pcli/src/command/keys.rs:57:14
>
> Additional note: if I run with sudo (or some other permissions nuke) from the state of the filesystem at step 7 everything works successfully from there on.

**Expected behavior**

The wallet should be generated and no errors should be thrown.

mikayla-maki:
Given the successful second fresh-install attempt, it seems that the additional note about sudo was incorrect. Another error we encountered trying to recreate the problem was caused by the existence of the archive: `/Users/mikaylamaki/Library/Application Support/zone.penumbra.penumbra-testnet-archive/1798c88f1bba05b3/custody.json`. The first error, that started this saga, is still present.

PR #1517, opened 2022-10-06
Added call to ensure that the opt.data_dir exists before pcli runs

mikayla-maki:
This fixes https://github.com/penumbra-zone/penumbra/issues/1516, thanks to @plaidfinch for walking me through my first PR!


plaidfinch:
Looks great! I'll merge this when it passes CI.

Issue #1507, opened 2022-10-05
Duplicate note commitment causing failure when syncing view service


hdevalence:
**Describe the bug**

When running `pcli view balance`, I get an error:

```
~/c/p/penumbra ((031-autonoe)|✔) $ cargo run --release --bin pcli -- view balance
Scanning blocks from last sync height 13544 to latest height 27114
[0s] ██████████████████████████████████████████████████   13570/13570   14873/s ETA: 0s
Error: status: Internal, message: "Worker failed: error returned from database: UNIQUE constraint failed: notes.note_commitment", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc"} }
```

There are two separate problems here:

1. Something is creating duplicate notes, which indicates a bug -- the blinding factor in a note is intended to ensure that note commitments are globally unique, and that's not happening here.
2. The view service should not fail on duplicate note commitments.  Although note commitments are verified, in the sense that they must be created from the actual note contents, they're still attacker-controlled (an attacker could deliberately craft two identical notes with identical blinding factors, to obtain duplicate note commitments), so the view service cannot assume that notes are constructed with unique blinding factors.

**This is actually a security bug**, because it violates infallibility of scanning -- an attacker could use this to take a targeted address's view service offline, and thus learn about that user's network location (by watching which client goes offline).

**To Reproduce**

My testnet FVK is

```
penumbrafullviewingkey10meak7pngg580r57fpvclrrzvaf8da56nayejxssv50w85anngfy2sct8myyjy8fgxxs7whrrpew5ncyzma76ern06ueevump8ae7zqd2wd7w
```

and this can probably be used to reproduce the problem.

hdevalence:
To fix this, it seems like we'll either have to decide to:

1. Relax the `UNIQUE` constraint on the notes table, sacrificing uniqueness of note-by-commitment lookups. This relaxation would have to propagate through the whole API, because now a note commitment corresponds to a set of `NoteRecord`s.
2. Reject duplicate notes during scanning, so that duplicated notes are not accepted by the user (and thus become unspendable without a special software implementation). This acts as a forcing function to ensure that client software doesn't create duplicate notes (because the recipients won't accept them).

hdevalence:
My sense is that (2) is probably a better option, since duplicate note commitments "shouldn't" happen, and that way we avoid contorting our API to accomodate malformed data.

aubrika:
[...]

PR #1519, opened 2022-10-11
test: note commitment bug regression

redshiftzero:
Attempt at a regression test for #1507 (downside of doing this through the smoke tests is that it will be slow)


redshiftzero:
ok confirmed: after waiting for an epoch duration, in [this build](https://github.com/penumbra-zone/penumbra/actions/runs/3228324840/jobs/5284272082) we can see that we hit the unique constraint on the note commitment as in #1507:

```
status: Internal, message: \"Worker failed: error returned from database: UNIQUE constraint failed: notes.note_commitment\"
```

PR #1523, opened 2022-10-13
Further implementation of storage2

zbuc:


storage2/src/storage.rs
```diff
@@ -29,16 +64,129 @@ impl Storage {
...
+
+                    // 4. update the snapshot
+                    // Now that we've successfully written the new nodes, update the version.
+                    let jmt_version = new_version;
+                    let snapshot = db.snapshot();
+                    // Obtain the write-lock for the latest snapshot, and replace it with the new snapshot.
+                    let mut guard = self.latest_snapshot.write();
+                    *guard = Arc::new(Snapshot::new(snapshot, jmt_version, db));
+                    // Drop the write-lock (this will happen implicitly anyways, but it's good to be explicit).
+                    drop(guard);
```
plaidfinch, line 126: I think you could rephrase this as:

```rust
*self.latest_snapshot.write() = Arc::new(Snapshot::new(db.snapshot(), new_version, db));
```

storage2/src/storage.rs
```diff
@@ -29,16 +64,129 @@ impl Storage {
...
+                span.in_scope(|| {
+                    let snap = self.latest_snapshot.read().clone();
+                    let jmt = JellyfishMerkleTree::new(snap.as_ref());
+                    // Write the unwritten changes from the state to the JMT.
+                    let (jmt_root_hash, batch) = jmt.put_value_set(
+                        state
+                            .unwritten_changes
+                            .into_iter()
+                            .map(|x| (KeyHash::from(x.0), x.1))
+                            .collect(),
```
plaidfinch, line 107: This `.collect()` is now unnecessary (and I think will lead to a type ambiguity error) due to the change I made to the `put_value_set` function: now it takes an `IntoIterator` (and since `Iterator`s implement `IntoIterator`, this will Just Work:tm:).

storage2/src/storage.rs
```diff
@@ -29,16 +64,129 @@ impl Storage {
...
+
+        let db = self.db;
+
+        let span = Span::current();
+        tokio::task::Builder::new()
+            .name("put_nct")
+            .spawn_blocking(move || {
+                span.in_scope(|| {
+                    let nct_cf = db.cf_handle("nct").expect("nct column family not found");
+                    db.put_cf(nct_cf, "tct", &tct_data)
```
plaidfinch, line 82: Should we call this key "tct" or "nct"? Our only NCT is the TCT, so I'm tempted to say "nct".

storage2/src/storage.rs
```diff
@@ -29,16 +64,129 @@ impl Storage {
...
-        // 2. update the snapshot
-        // TODO: set jmt_version correctly
-        let jmt_version = 0;
-        let snapshot = self.db.snapshot();
-        // Obtain the write-lock for the latest snapshot, and replace it with the new snapshot.
-        let mut guard = self.latest_snapshot.write();
-        *guard = Arc::new(Snapshot::new(snapshot, jmt_version));
-        // Drop the write-lock (this will happen implicitly anyways, but it's good to be explicit).
-        drop(guard);
+    pub async fn apply(&'static mut self, state: State, nct: &Tree) -> Result<()> {
```
plaidfinch, line 67: Does this need to take `self` as `&'static mut`, or can it just be `&mut`?

storage2/src/storage.rs
```diff
@@ -29,16 +64,129 @@ impl Storage {
...
+    /// Writes a node batch into storage.
+    //TODO: Change JMT traits to accept owned NodeBatch
+    fn write_node_batch(&self, node_batch: &NodeBatch) -> Result<()> {
+        let db = self.db;
+        let node_batch = node_batch.clone();
+
+        // The writes have to happen on a separate spawn_blocking task, but we
+        // want tracing events to occur in the context of the current span, so
+        // propagate it explicitly:
+        let span = Span::current();
```
plaidfinch, line 148: I don't see this span getting used. Is this an old piece of code and comment, or is the span not getting propagated as it is intended to be?

storage2/src/storage.rs
```diff
@@ -29,16 +64,129 @@ impl Storage {
...
+
+        Ok(())
+    }
+}
+
+// TODO: maybe these should live elsewhere?
+fn get_rightmost_leaf(db: &DB) -> Result<Option<(NodeKey, LeafNode)>> {
+    let span = Span::current();
+
+    // TODO: should this be in a tokio task like before?
```
plaidfinch, line 167: This should not be in a tokio task, because `get_rightmost_leaf` is not an async function.

storage2/src/storage.rs
```diff
@@ -29,16 +64,129 @@ impl Storage {
...
+            db.put_cf(jmt_cf, key_bytes, &value_bytes)?;
+        }
+
+        Ok(())
+    }
+}
+
+// TODO: maybe these should live elsewhere?
+fn get_rightmost_leaf(db: &DB) -> Result<Option<(NodeKey, LeafNode)>> {
+    let span = Span::current();
```
plaidfinch, line 165: Is this span meant to be used somewhere? I don't see it being used.

storage2/src/storage.rs
```diff
@@ -29,16 +64,129 @@ impl Storage {
...
+
+            let jmt_cf = db.cf_handle("jmt").expect("jmt column family not found");
+            db.put_cf(jmt_cf, key_bytes, &value_bytes)?;
+        }
+
+        Ok(())
+    }
+}
+
+// TODO: maybe these should live elsewhere?
```
plaidfinch, line 163: Shouldn't these just be directly in the implementation of `TreeReader`?

storage2/src/state/transaction.rs
```diff
@@ -5,22 +5,16 @@ use super::StateWrite;
 /// Represents a transactional set of changes to a `State` fork,
 /// implemented as a RYW cache over a `State`.
 pub struct Transaction<'a> {
-    // TODO determine which fields to include (#1490)
-    // cache: HashMap<jmt::KeyHash, jmt::OwnedValue>,
-    // unwritten_changes: Vec<(jmt::KeyHash, jmt::OwnedValue)>,
+    pub(crate) unwritten_changes: Vec<(String, Option<Vec<u8>>)>,
```
plaidfinch, line 8: I think this should be a `HashMap<String, Option<Vec<u8>>>`, not a `Vec`. That way if you overwrite a key multiple times, or delete it after inserting it in the same unwritten set, you don't have extra memory used to represent those useless operations, and they don't have to be processed by `put_value_set`. This works now that `put_value_set` takes any `IntoIterator<Item = (KeyHash, Option<OwnedValue>)>` as an argument, rather than just a `Vec`.

storage2/src/state.rs
```diff
@@ -29,14 +34,22 @@ impl State {
     pub fn begin_transaction(&mut self) -> StateTransaction {
         StateTransaction::new(self)
     }
+
+    // Apply the unwritten changes of a transaction to this state fork.
+    pub fn apply_transaction(&mut self, transaction: StateTransaction) {
+        for (key, value) in transaction.unwritten_changes.iter() {
+            self.unwritten_changes.insert(key.clone(), value.clone());
+        }
```
plaidfinch, line 42: This could be rewritten as:

```rust
self.unwritten_changes.extend(transaction.unwritten_changes)
```

This has the advantage of avoiding cloning, which you don't need to do because `StateTransaction` is taken by value.

storage2/src/snapshot.rs
```diff
@@ -28,3 +37,50 @@ impl Snapshot {
...
+        Ok(value)
+    }
+
+    fn get_rightmost_leaf(&self) -> Result<Option<(NodeKey, LeafNode)>> {
+        let jmt_cf = self
+            .db
+            .cf_handle("jmt")
+            .expect("jmt column family not found");
+        let mut iter = self.rocksdb_snapshot.raw_iterator_cf(jmt_cf);
+        let mut ret = None;
```
plaidfinch, line 70: I think this would be clearer with multiple `return`s rather than a single return value that is initialized and then mutated.

storage2/src/snapshot.rs
```diff
@@ -28,3 +37,50 @@ impl Snapshot {
...
+            .db
+            .cf_handle("jmt")
+            .expect("jmt column family not found");
+        let value = self
+            .rocksdb_snapshot
+            .get_cf(jmt_cf, &node_key.encode()?)?
+            .map(|db_slice| Node::decode(&db_slice))
+            .transpose()?;
+
+        tracing::trace!(?node_key, ?value);
```
plaidfinch, line 60: Do we want to trace after the possible errors could be thrown, or before? I think before, so that the trace shows up even if an error is thrown.

storage2/src/snapshot.rs
```diff
@@ -28,3 +37,50 @@ impl Snapshot {
...
 }
+
+/// A reader interface for rocksdb. NOTE: it is up to the caller to ensure consistency between the
+/// rocksdb::DB handle and any write batches that may be applied through the writer interface.
+impl TreeReader for Snapshot {
+    /// Gets node given a node key. Returns `None` if the node does not exist.
+    fn get_node_option(&self, node_key: &NodeKey) -> Result<Option<Node>> {
+        let node_key = node_key.clone();
+
+        let span = Span::current();
```
plaidfinch, line 48: This span isn't used. Should it be? What did the previous code do with this (I assume that's why it's here now?)?

storage2/src/snapshot.rs
```diff
@@ -28,3 +37,50 @@ impl Snapshot {
         self.jmt_version
     }
 }
+
+/// A reader interface for rocksdb. NOTE: it is up to the caller to ensure consistency between the
+/// rocksdb::DB handle and any write batches that may be applied through the writer interface.
+impl TreeReader for Snapshot {
+    /// Gets node given a node key. Returns `None` if the node does not exist.
+    fn get_node_option(&self, node_key: &NodeKey) -> Result<Option<Node>> {
+        let node_key = node_key.clone();
```
plaidfinch, line 46: I don't think the node key needs to be cloned. It's only used by reference below.

storage2/src/storage.rs
zbuc: We were using both in the original code, I started switching over to just nct for the reason you mentioned but missed this :)

storage2/src/storage.rs
zbuc: But whatever calls it should be in a Tokio task because it is calling out to the blocking rocksdb code?

storage2/src/storage.rs
zbuc: Yeah this is an artifact from when this was inside of a Tokio task

storage2/src/snapshot.rs
zbuc: Another Tokio task artifact

storage2/src/storage.rs
plaidfinch: At the outermost point possible, it should be in a `spawn_blocking` call, yes. But in order to make things most efficient, we should push that boundary as far out as possible, because each `spawn_blocking` corresponds to a thread allocated on the blocking thread pool.

zbuc: Though I guess this storage code will ultimately be abstracted away behind another layer of API and the higher-level `pd` code shouldn't have to worry about spawning threads when DB access is performed.

storage2/src/lib.rs
```diff
@@ -1,4 +1,5 @@
-use jmt::WriteOverlay;
+#![recursion_limit = "256"] // required for TCT
+
```
hdevalence, line 2: We should remove this, and the dependency of the `storage2` crate on the `tct` crate.

storage2/src/metrics.rs
```diff
@@ -14,4 +14,13 @@
...
 
 /// Registers all metrics used by this crate.
-pub fn register_metrics() {}
+pub fn register_metrics() {
+    register_gauge!(NCT_SIZE_BYTES);
+    describe_gauge!(
+        NCT_SIZE_BYTES,
+        Unit::Bytes,
+        "The size of the serialized NCT in bytes"
+    );
```
hdevalence, line 23: I think we should drop this metric; it gives no useful information, because the NCT size is always in a bounded range.

storage2/src/storage.rs
plaidfinch: That's right: the storage API exposed to `pd` will be an `async` API, but we want to only spawn one thread for each call to a function on `Storage`, rather than many. The `Storage` interface will end up wrapping each method's body in a `spawn_blocking`, I think.

storage2/src/state.rs
```diff
@@ -29,14 +34,22 @@ impl State {
...
 }
 
 #[async_trait]
 impl StateRead for State {
     fn get_raw(&self, key: String) -> Option<Vec<u8>> {
         // If the key is available in the unwritten_changes cache, return it.
+        // A `None` value represents that the key has been deleted.
         if let Some(value) = self.unwritten_changes.get(&key) {
-            return Some(value.clone());
+            return value.clone();
```
hdevalence, line 52: I got confused reading this, because I thought that
```
// A `None` value ...
```
was referring to the return value of `.get(&key)`, not to the *inner* value.  It might be clearer to say something like

> `.get()` returns an `Option<Option<Vec<u8>>`; the outer `Option` signals whether the key is in the `unwritten_changes` cache, and the inner `Option` signals whether the value is present or deleted

plaidfinch: You could also write:

```rust
fn get_raw(&self, key: String) -> Option<Vec<u8>> {
    self.unwritten_changes.get(&key).or_else(|| self.snapshot.get_raw(key))
}
```

This makes it more evident to me what's going on here.

storage2/src/snapshot.rs
```diff
@@ -28,3 +37,50 @@ impl Snapshot {
         self.jmt_version
```
plaidfinch, line 37: I think `get_raw` ought to be able to take a `&str`, not a `String`, right?

storage2/src/storage.rs
hdevalence, line 67: I think it would be more extensible for the future if this method only took a `State`, not a `Tree`, so that we're not special-casing the NCT state.  Instead, we would have methods on `State`/`StateTransaction` for writing key/value pairs that aren't part of the consensus state.

plaidfinch: Yeah, that makes sense to me. Probably the `State` should have something like a `HashMap<&'static str, HashMap<String, Option<Vec<u8>>>>>` to track changes to column families (named by `&'static str`) that are not the JMT?

storage2/src/storage.rs
```diff
@@ -29,16 +64,129 @@ impl Storage {
...
+            })
+            .unwrap()
+            .await
+            .unwrap();
+
+        Ok(())
     }
 }
+
+impl TreeWriter for Storage {
```
hdevalence, line 138: We shouldn't implement `TreeWriter` for `Storage`, because we don't want the JMT traits to be part of the public API, but trait implementations don't have separate visibility from the types that implement them, and the `Storage` type is public.

We could perhaps create a private wrapper type around a reference instead, and implement the trait on that type.

storage2/src/storage.rs
hdevalence: In order to abstract away the storage code, though, we'll need to implement the trait differently: https://github.com/penumbra-zone/penumbra/pull/1523/files#r995878288

Otherwise, it will leak into the public API.

storage2/src/state.rs
```diff
@@ -15,7 +19,8 @@ use crate::snapshot::Snapshot;
 /// implemented as a RYW cache over a pinned JMT version.
 pub struct State {
     snapshot: Arc<Snapshot>,
-    unwritten_changes: HashMap<String, Vec<u8>>,
+    // A `None` value represents deletion.
+    pub(crate) unwritten_changes: HashMap<String, Option<Vec<u8>>>,
```
hdevalence, line 23: Rather than a `HashMap`, I think we should use a `BTreeMap` here, so that in the future, we'll be able to iterate over keys.

storage2/src/state/transaction.rs
hdevalence: Rather than a `HashMap`, it'd be better to use a `BTreeMap`, so that we'll be able to iterate over keys in order.

storage2/src/storage.rs
zbuc: This isn't working right for some reason:

```rust
mismatched types
expected struct `std::vec::Vec<(jmt::KeyHash, std::option::Option<std::vec::Vec<u8>>)>`
   found struct `std::iter::Map<std::collections::btree_map::IntoIter<std::string::String, std::option::Option<std::vec::Vec<u8>>>, [closure@storage2/src/storage.rs:106:34: 106:63]>`
```

zbuc: despite rust-analyzer saying that it takes an `IntoIterator`... maybe something's cached on my local

plaidfinch: Try doing `cargo update` and see if that fixes it.

zbuc: I did, it's weirdly only happening in VSCode, on the command-line when I `cargo build` it's fine

storage2/src/storage.rs
zbuc: If I remove the lifetime, the build breaks

storage2/src/storage.rs
zbuc: We can't do that because we need to use this in the constructor of `Storage` (`Storage::load`), prior to `Storage` being instantiated

zbuc: Or maybe we can now that we have the separate `Inner` type

storage2/src/storage.rs
plaidfinch: Have you restarted rust-analyzer since you did the cargo update?

storage2/src/storage.rs
plaidfinch: What's the error?

storage2/src/storage.rs
zbuc: Hm, no, because the DB reference and Snapshot need to live alongside each other, and the Snapshot needs the version... maybe there's some type tetris we can do, not sure it's worth it

storage2/src/storage.rs
zbuc: Yeah, I wound up working around it by manually specifying the revision in the `Cargo.toml` 🙃

storage2/src/storage.rs
zbuc:
```
borrowed data escapes outside of associated function
`self` escapes the associated function body here
```

referring to the tokio task

storage2/src/storage.rs
plaidfinch: I don't think that should be necessary. I would try `cargo clean && cargo update`. Something is cached that shouldn't be.

storage2/src/storage.rs
plaidfinch: Oh I see, yes, that makes sense. Tokio tasks can only take `'static` things, because they might outlive the scope in which they are spawned.

Issue #1044, opened 2022-06-21
`pcli` does not work offline following refactoring


hdevalence:
**Describe the bug**

The refactoring @hawkw did in #1003 made the code much cleaner, splitting the role of the `Opt` struct (now just representing the arguments themselves) and the `App` struct (now representing the configured and constructed application data.

However, as a side effect, it also made `pcli` stop working offline, for subtle reasons.  Previously, the `needs_sync` method on the command controlled both (1) whether `pcli` would try to synchronize before doing the command and (2) whether `pcli` would connect to the network at all.  Now, `needs_sync` controls whether we wait to sync before running the command, but we unconditionally connect to the remote node in `Opt::view_client`.  This means that if the node is offline, we won't run the command at all.

I'm not sure what the best way to fix this is, because we can't easily get rid of the requirement to construct a view service in the `App`.  It's not essential to have `pcli` work solely offline, at least for now, so this isn't urgent, but it would be ideal to be able to run it offline.

hawkw:
Agh, whoops! I can try to fix this.

PR #1526, opened 2022-10-14
Allow `pcli` to be run offline 

mikayla-maki:
This fixes  https://github.com/penumbra-zone/penumbra/issues/1044

This PR contains 3 changes:

1. Removes an `fs::create_dir_all` call that was rendered extraneous by https://github.com/penumbra-zone/penumbra/pull/1517
2. Renames 'needs_sync' to 'offline' and inverts the related values
3. Only instantiates a view if the command requires network access.

Note that the `app.view.as_mut().unwrap()` calls cannot be turned into `app.view()` calls, as pcli makes heavy use of disjoint borrows to 'disassemble' a `&mut app` into a `&fvk` and a `&mut view`. Calling `app.view()` borrows the entire struct and makes it impossible to get a reference to the full viewing key.

TODO:
- [x] Write some tests to make sure that I did this correctly.


mikayla-maki:
The failing test was caused by a `pcli tx proposal template`being marked as 'offline' even though it uses the view. I followed this up by manually looking through the other offline implementations for if they use the view, then manually building and running every command that is marked as 'offline'. I found 2 other inconsistencies, one for `pcli view address` being marked as offline but taking a view because of some type casting problems, and a special case for `pcli view balance --offline` which specifies that it runs an unsynced view.

pcli/src/command/view.rs
```diff
@@ -63,8 +63,8 @@ impl ViewCmd {
             ViewCmd::Reset(_reset) => {
                 // The wallet has already been reset by a short-circuiting path.
             }
-            ViewCmd::Address(address_cmd) => {
-                address_cmd.exec(full_viewing_key)?;
+            ViewCmd::Address(_address) => {
+                // Already done previously
```
hdevalence, line 67: Wait, what's happening here? It doesn't look like anything is executing the address command...

pcli/src/command/view/balance.rs
```diff
@@ -18,8 +18,8 @@ pub struct BalanceCmd {
 }
 
 impl BalanceCmd {
-    pub fn needs_sync(&self) -> bool {
-        !self.offline
+    pub fn offline(&self) -> bool {
+        self.offline
```
hdevalence, line 22: This doesn't work any more, because now we're combining "does the command require a view service" with "does the command work offline". Now, the offline option for balance won't work, because the view service won't be present.

To fix this, we can remove the option entirely, and have the balance command always require a sync.

pcli/src/main.rs
```diff
@@ -32,7 +32,7 @@ const VIEW_FILE_NAME: &str = "pcli-view.sqlite";
 
 #[derive(Debug)]
 pub struct App {
-    pub view: ViewProtocolClient<BoxGrpcService>,
+    pub view: Option<ViewProtocolClient<BoxGrpcService>>,
```
hdevalence, line 35: It'd be helpful to add a comment here explaining the meaning of the Option: that it's None for commands that can run offline, and Some for commands that can't.

pcli/src/main.rs
```diff
@@ -127,11 +128,15 @@ async fn main() -> Result<()> {
...
-                .await?
+            if let ViewCmd::Address(address_cmd) = view_cmd {
+                address_cmd.exec(&app.fvk)?;
+            } else {
+                let mut oblivious_client = app.oblivious_client().await?;
+
+                view_cmd
+                    .exec(&app.fvk, app.view.as_mut().unwrap(), &mut oblivious_client)
+                    .await?
+            }
```
hdevalence, line 139: Ahhh, this is why the address command is "already handled"? It's a little confusing to have non-local logic, maybe this could be pushed into the view command's exec() method itself, but pass an Option to exec() to be able to handle the offline behavior of view addr ?

pcli/src/opt.rs
```diff
@@ -88,7 +86,13 @@ impl Opt {
...
 
         // ...and the view service...
-        let view = self.view_client(&fvk).await?;
+        // There is one command that is both offline and *un*synced:
+        // `pcli view balance --offline`. Hence, this check.
+        let view = if !self.cmd.offline() || matches!(self.cmd, Command::View(Balance(_))) {
+            Some(self.view_client(&fvk).await?)
+        } else {
+            None
+        };
```
hdevalence, line 95: As above, I think the simpler thing here would just be to remove offline balance support. As noted in your comment here, it's a bit of a weird outlier.

mikayla-maki:
Implemented those updates!

PR #1533, opened 2022-10-18
zk: wip using ark-groth16 for the OutputProof

### LINKED ISSUE #1094: Groth16 prototyping
Related to #2

We've prototyped with PLONK (#666, #665) for the `OutputProof`, and we should perform that same prototyping with some of the groth16 libraries we might use, e.g. https://github.com/arkworks-rs/groth16 is a good starting point. 

The goal of this is to get concrete numbers for prover time, bearing in mind that we can parallelize proofs (#717). While PLONK can be optimized to be faster than what we saw from our prototyping (approx ~10s per action), if we can get better out of the box performance from Groth16, we may proceed with that for some of the more fixed functionality for the initial mainnet launch, i.e. for spends and outputs. We can then migrate away from Groth16 in a future network upgrade once our PLONK performance is better in terms of prover time.

redshiftzero:
Closes #1094

This adds an `OutputCircuit`, a struct used by the prover that has the witnesses and public inputs as members. Implemented on the `OutputCircuit` is `ConstraintSynthesizer<Fq>` which has the logic for generating constraints (this logic is what will later need to get broken up into gadgets, not sure how best to do that yet though). Currently the `ConstraintSynthesizer<Fq>` is just adding constraints for performing the diversified basepoint check. The `OutputCircuit` is used to generate the (zk) `OutputProof` which can later be verified given the public inputs and the verifying key from the circuit-specific setup. 

TODO:
- [ ] Implement more of the circuit logic to match #665 for getting comparable concrete prover time numbers


redshiftzero:
This was mostly for prototyping, closing in favor of #1640

PR #1536, opened 2022-10-18
Implement `pcli v tx` functionality for viewing transactions

### LINKED ISSUE #1364: Add pcli view transaction command for single transactions by hash

aubrika:
Closes #1364

### DIFF
--- a/pcli/src/command/view/tx.rs
+++ b/pcli/src/command/view/tx.rs
@@ -1,7 +1,8 @@
 use anyhow::Result;
 use comfy_table::{presets, Table};
-use penumbra_crypto::FullViewingKey;
+use penumbra_crypto::{asset::Cache, dex::swap::SwapPlaintext, FullViewingKey, Note, Value};
 use penumbra_transaction::{
+    action::{Swap, SwapClaim},
     view::action_view::{OutputView, SpendView, SwapClaimView, SwapView},
     Transaction,
 };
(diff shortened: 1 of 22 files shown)


pcli/src/command/view/tx.rs
```diff
@@ -0,0 +1,97 @@
...
+                penumbra_transaction::ActionView::SwapClaim(swap_claim_view) => {
+                    match swap_claim_view {
+                        SwapClaimView::Visible {
+                            swap_claim,
+                            decrypted_note_1,
+                            decrypted_note_2,
+                        } => format!("{:?}", "SwapClaim"),
+                        SwapClaimView::Opaque { swap_claim } => format!("{:?}", "SwapClaim"),
+                    }
+                }
```
hdevalence, line 73: Rather than having nested match statements, I think it'd be cleaner to combine patterns to have one match arm per distinct case:

```suggestion
                ActionView::SwapClaim(SwapClaimView::Visible {
                    swap_claim,
                    decrypted_note_1,
                    decrypted_note_2,
                }) => {
                    //format!("{:?}", "SwapClaim with contents"),
                }
                ActionView::SwapClaim(SwapClaimView::Opaque { swap_claim }) => {
                    //format!("{:?}", "SwapClaim without contents"),
                }
```

view/src/storage.rs
```diff
@@ -739,47 +836,8 @@ impl Storage {
             let position = (u64::from(note_record.position)) as i64;
             let source = note_record.source.to_bytes().to_vec();
 
-            // If this note corresponded to a previously quarantined note, delete it from quarantine
```
zbuc, line 742: Why is all this code removed in this PR? It seems to break the smoke tests

redshiftzero: It looks like it reverts the note commitment fix added in #1520 so could just be due to a rebase, I think we do want that logic (the smoke test in question is failing as it is a regression test for the bug in #1520)

proto/proto/penumbra/view/v1alpha1/view.proto
```diff
@@ -55,8 +55,21 @@ service ViewProtocol {
...
+}
+
+message PerspectiveRequest {
+    bytes tx_hash = 1;
+}
+
+message PerspectiveResponse {
+    core.transaction.v1alpha1.TransactionPerspective txp = 1;
+    core.transaction.v1alpha1.Transaction tx = 2;
 }
```
plaidfinch, line 73: I would rename these to `TransactionPerspective{Request, Response}`, even though it's longer. I think this is more clear what they're doing.

transaction/src/transaction.rs
```diff
@@ -42,6 +46,49 @@ pub struct Transaction {
...
+                    let epk = &output.body.note_payload.ephemeral_key;
+                    let shared_secret = fvk.incoming().key_agreement_with(epk)?;
+                    let payload_key = PayloadKey::derive(&shared_secret, epk);
+                    let commitment = output.body.note_payload.note_commitment;
+
+                    result.insert(commitment, payload_key);
+                }
+                _ => {}
+            }
+        }
```
plaidfinch, line 88: Are these the only actions that require payload keys? Will there be others in the future?

The wildcard match here makes me a little uncomfortable because I worry we'll forget to update this function. One way to ensure we don't forget to update it would be to delegate the extraction of payload keys to a separate function on each action type, and invoke each (many of them would not do anything). This would ensure that if we add new actions, we won't forget to update this function.

view/src/client.rs
```diff
@@ -96,6 +97,19 @@ pub trait ViewClient {
...
 
+    /// Queries for a transaction hashes by its transaction hash
+
+    async fn transaction_by_hash(
+        &mut self,
+        tx_hash: abci::transaction::Hash,
+    ) -> Result<Option<Transaction>>;
+
+    /// Generates a full perspective for a selected transaction using a full viewing key
+    async fn perspective(
```
plaidfinch, line 108: I would rename this to `transaction_perspective`, for clarity.

view/src/client.rs
```diff
@@ -438,6 +452,40 @@ where
...
+            }),
+        )
+        .await?
+        .into_inner()
+        .tx
+        .map(|tx| tx.try_into())
+        .map_or(Ok(None), |v| v.map(Some))
+    }
+
+    async fn perspective(
```
plaidfinch, line 472: Similarly, here, I think `transaction_perspective` is more precise.

view/src/service.rs
```diff
@@ -239,6 +240,65 @@ impl ViewProtocol for ViewService {
...
+        let tx = self
+            .storage
+            .transaction_by_hash(&request.tx_hash)
+            .await
+            .map_err(|_| {
+                tonic::Status::failed_precondition("Error retrieving transaction by hash")
+            })?
+            .ok_or_else(|| {
+                tonic::Status::failed_precondition("No transaction found with this hash")
+            })?;
```
plaidfinch, line 265: Perhaps should we include the hash in the error message?

plaidfinch (commented):
This looks good to me! Just a few nits, mostly. I have not run it yet; I'll try that now.

plaidfinch:
I tried running this:

1. Requested a transaction from the faucet (address: `penumbrav2t12l90dw58y0mnjslwn7m3wxqt5fw2egv0u6aftc2p2ptfue3xkc4flhqqkdsxfnvr8p5223w8lfe9zh6jj6q3z96pw59kzvdn4rnq720vrnd4taztlg0hu70whl624k73cf3jcl`).
2. Retrieved the transaction hash using `pcli view list-tx-hashes` (hash: `e2620f098869f5813208cc4947871b77eb0dc0db59644303dc3eb489f3a62dd9`).
3. Attempted to view transaction contents, but was met with an error:
```
Error: status: FailedPrecondition, message: "Error retrieving note by nullifier: Note commitment for nullifier Nullifier(\"f69656c0a4916fee01efec66ed48622b88cea07d62ed647ad596df7b29336209\") not found", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc"} }
```

PR #1541, opened 2022-10-21
Asyncify storage code

zbuc:
I'm not sure about the handling of the `RwLock` + `Arc` in `storage::Inner`

### DIFF
--- a/storage2/src/state.rs
+++ b/storage2/src/state.rs
@@ -52,7 +52,6 @@ impl State {
 }
 
 #[async_trait]
-// TODO: @hdevalence does this really need to be a trait?
 impl StateRead for State {
     async fn get_raw(&self, key: &str) -> Result<Option<Vec<u8>>> {
         // If the key is available in the unwritten_changes cache, return it.
--- a/storage2/src/state.rs
+++ b/storage2/src/state.rs
@@ -19,7 +19,7 @@ pub struct State {
     // A `None` value represents deletion.
     pub(crate) unwritten_changes: BTreeMap<String, Option<Vec<u8>>>,
     // A `None` value represents deletion.
-    pub(crate) sidecar_changes: BTreeMap<Vec<u8>, Option<Vec<u8>>>,
+    pub(crate) nonconsensus_changes: BTreeMap<Vec<u8>, Option<Vec<u8>>>,
 }
 
 impl State {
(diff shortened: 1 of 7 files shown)


storage2/src/state.rs
```diff
@@ -19,42 +19,58 @@ pub struct State {
...
+        // Write the unwritten nonconsensus changes to the state:
+        self.nonconsensus_changes
+            .extend(transaction.nonconsensus_changes);
+
+        Ok(())
+    }
 }
 
 #[async_trait]
+// TODO: @hdevalence does this really need to be a trait?
```
plaidfinch, line 55: This was resolved in Discord, right? Might be worth putting a comment here about why: I actually forget why it needs to be a trait myself, again, but I know we reached consensus about it.

storage2/src/storage.rs
```diff
@@ -133,35 +136,33 @@ impl Storage {
...
                     // 4. update the snapshot
                     // Now that we've successfully written the new nodes, update the version.
                     let jmt_version = new_version;
-                    let snapshot = db.snapshot();
+                    let snapshot = Arc::new(db.snapshot());
                     // Obtain the write-lock for the latest snapshot, and replace it with the new snapshot.
                     let mut guard = self.0.latest_snapshot.write();
                     *guard = Arc::new(Snapshot::new(snapshot, jmt_version, db));
                     // Drop the write-lock (this will happen implicitly anyways, but it's good to be explicit).
                     drop(guard);
```
plaidfinch, line 161: I think you can do this like:

```rust
*self.0.latest_snapshot.write() = Arc::new(Snapshot::new(Arc::new(db.snapshot()), jmt_version, db));
```

Whether that's more readable is debatable, though :upside_down_face:

plaidfinch: Thinking about why we need the double-`Arc`: the outer one is because we may have many references to one of our `Snapshot`s, but why does `Snapshot` itself need to wrap the `rocksdb::Snapshot` in an `Arc`? Could we elide that? What gets in the way?

plaidfinch (commented):
This looks solid to me, seems like it's coming together nicely :) Just a few questions about `Arc` handling.

storage2/src/storage.rs
zbuc: Without having the `rocksdb::Snapshot` wrapped in an `Arc`, the tokio tasks that need to access it don't work out.

plaidfinch: Ah, that makes sense. I think we should leave it as is; an extra indirection is fine.

hdevalence: > the outer one is because we may have many references to one of our `Snapshot`s

Why does that mean we need an `Arc`-wrapper? What if we just make `Snapshot` `Clone`?

PR #1546, opened 2022-10-21
Ranged storage access

zbuc:


storage2/src/snapshot.rs
```diff
@@ -8,58 +14,121 @@ use crate::state::StateRead;
...
+        prefix: &str,
+    ) -> Result<Pin<Box<dyn Stream<Item = (String, Box<[u8]>)> + Send + '_>>> {
+        let span = Span::current();
+        let db = self.0.db;
+        let rocksdb_snapshot = self.0.rocksdb_snapshot.clone();
+        let mut options = rocksdb::ReadOptions::default();
+        options.set_iterate_range(rocksdb::PrefixRange(prefix.as_bytes()));
+        let mode = rocksdb::IteratorMode::Start;
+
+        let (tx, rx) = mpsc::channel(100);
```
plaidfinch, line 96: I wonder what the appropriate value to set here is. Should this be configurable, so we can tune and benchmark it?

storage2/src/state.rs
```diff
@@ -38,23 +41,74 @@ impl State {
...
+        let mut snapshotted_stream = self.snapshot.prefix_raw(prefix).await?;
+        let mut snapshotted_match = snapshotted_stream.next().await;
+        for (key, value) in self.unwritten_changes.iter() {
+            // Iterate the unwritten_changes cache (sorted by key) until we reach the keys
+            // that match the prefix.
+            //
+            // If value is `None`, then the key has been deleted, and we should skip it.
+            if !key.starts_with(prefix) || value.is_none() {
+                continue;
+            }
```
plaidfinch, line 80: You can use the [`BTreeMap::range`](https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.range) method here to skip right to the key you want to start with, which means you don't have to scan all the changes, just the relevant ones.

storage2/src/state.rs
```diff
@@ -38,23 +41,74 @@ impl State {
...
+        }
+
+        // Send any remaining data from the snapshot stream.
+        while let Some((snapshotted_key, snapshotted_value)) = snapshotted_match {
+            tx.send((snapshotted_key, snapshotted_value)).await?;
+            // Advance the snapshot stream to the next match.
+            snapshotted_match = snapshotted_stream.next().await;
+        }
+
+        Ok(Box::pin(tokio_stream::wrappers::ReceiverStream::new(rx)))
```
plaidfinch, line 112: I think this loop could be rewritten more simply using `BTreeMap::range` as noted above and [`IterTools::merge_by`](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.merge_by) to merge the two iterators in order. Make sure to cut off the `BTreeMap::range` when it exceeds the prefix, perhaps using `.take_while()`?

plaidfinch (commented):
This is looking good! I put some commentary in regarding refactoring the interleaving of iterators; I think this can be clearer and also faster by using some lesser-known facilities of `BTreeMap`. What do you think?

storage2/src/snapshot.rs
zbuc: Configurable in what sense, as a command-line flag?

plaidfinch: I don't think we should necessarily have a command line flag, but it might be good to (as a maintenance task) ferret out such "magic constants" and perhaps have a configuration file that defines them, with some sensible defaults. I have no idea if 100 is the best buffer size here, and similarly for other places in the codebase where we arbitrarily pick one; it might be worth experimentally finding out.

hdevalence: 10 is a good number, it's good to avoid buffer bloat. I don't think we'll want a configuration file, we can just pick something now and do performance tuning later/never.

zbuc:
> This is looking good! I put some commentary in regarding refactoring the interleaving of iterators; I think this can be clearer and also faster by using some lesser-known facilities of `BTreeMap`. What do you think?

Sounds good, I had no idea those methods existed :)

storage2/src/state.rs
zbuc: Turns out we can't use `IterTools::merge_by` because we don't have two iterators, and it's not clear that re-implementing that logic for `Stream` is worthwhile

Issue #1510, opened 2022-10-05
Refactor MemoPlaintext to be a fixed-length String


aubrika:
This touches enough of the codebase to deserve its own tracking issue: MemoPlaintext is used in several places but we would be better served by a fixed-length string, though swapping it out may take some thinking about how best to refactor the functions on the implementation.

conorsch:
> we would be better served by a fixed-length string

Can someone elaborate on the utility of switching to a string and ditching the struct entirely? I can see that a string would be more pleasant to work with than a vec, but having the methods bundled to the struct for encrypt/decrypt actions strikes me as a pleasant convenience. If anything, I'm biased toward leveraging the `impl` further: we currently validate that the memo text is less than `MEMO_LEN_BYTES` in several places, but it might be safer to move that validation logic into a constructor, so that future maintainers don't overlook it when using memos elsewhere.

zbuc:
The intent is to make sure the memo is UTF-8 encoded, so really what we are looking for is a fixed-length UTF-8 string

hdevalence:
> I can see that a string would be more pleasant to work with than a vec, but having the methods bundled to the struct for encrypt/decrypt actions strikes me as a pleasant convenience.

One point here is that, ideally, users of these APIs won't be calling encrypt/decrypt directly: when creating a transaction, they would construct a `TransactionPlan` with the `Planner` API, then call `build` (which would do encryption internally), and when viewing a transaction, they'd use a `TransactionPerspective` to create a `TransactionView`, which provides an already decrypted and "interpreted" view of the transaction.

In either of those cases, an API user will just be operating on the plaintext of the transaction, and at that point, having a `String` is much more ergonomic.

We also don't necessarily need or want the string to be fixed-length.  For instance, if I create a transaction with the memo field `"for pizza"`, my 9-byte memo will be padded with 0s to get a fixed-length memo ciphertext. But when I look at the transaction later, I don't really want to see `"for pizza\0\0\0...(503 zeros)...\0"`, I just want to see `"for pizza"`.  Or, if I'm reviewing a `TransactionPlan` before signing, seeing the zero-padded memo plaintext isn't particularly helpful either.

So I think it's simplest just to treat the memo plaintext as an ordinary `String`.

Now we have two decisions to make:

1. How do we handle the conversion from variable-length `String`s to 512-byte plaintexts when building a transaction?
[...]

PR #1548, opened 2022-10-25
Convert MemoPlaintext to String type

conorsch:
We need a fixed length of 512 bytes for passing the memo plaintext to cryptographic functions. Everywhere else, though, a String is most convenient. Here we add some conversion traits to make working with MemoPlaintext a bit more ergonomic. Ideally we'd perform validation such that a too-long String cannot be used to construct a MemoPlaintext.

Opening early for feedback on approach and style. The `strip_padding` function should be smarter, will address its TODO before declaring final. I'm tempted to create a constructor for MemoPlaintext so that we could validate string length just once and be done with it.

### DIFF
--- a/crypto/src/memo.rs
+++ b/crypto/src/memo.rs
@@ -1,6 +1,6 @@
 use std::{
     convert::{TryFrom, TryInto},
-    fmt::{self, Debug, Formatter},
+    fmt::Debug,
 };
 
 use anyhow::anyhow;
@@ -18,76 +18,64 @@ pub const MEMO_CIPHERTEXT_LEN_BYTES: usize = 528;
 // This is the `MEMO_CIPHERTEXT_LEN_BYTES` - MAC size (16 bytes).
 pub const MEMO_LEN_BYTES: usize = 512;
 
-// The memo is stored separately from the `Note`.
-// TODO: MemoPlaintext should just be a fixed-length string, drop this type entirely
-#[derive(Clone, PartialEq, Eq)]
-pub struct MemoPlaintext(pub [u8; MEMO_LEN_BYTES]);
-
-impl Debug for MemoPlaintext {
-    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
-        write!(f, "MemoPlaintext({})", hex::encode(self.0))
-    }
-}
-
-impl Default for MemoPlaintext {
-    fn default() -> MemoPlaintext {
-        MemoPlaintext([0u8; MEMO_LEN_BYTES])
-    }
-}
-
-impl TryFrom<&[u8]> for MemoPlaintext {
-    type Error = anyhow::Error;
-
-    fn try_from(input: &[u8]) -> Result<MemoPlaintext, Self::Error> {
-        if input.len() > MEMO_LEN_BYTES {
-            return Err(anyhow::anyhow!("provided memo exceeds maximum memo size"));
-        }
-        let mut mp = [0u8; MEMO_LEN_BYTES];
-        mp[..input.len()].copy_from_slice(input);
-
-        Ok(MemoPlaintext(mp))
-    }
-}
-
 #[derive(Clone, Debug)]
 pub struct MemoCiphertext(pub [u8; MEMO_CIPHERTEXT_LEN_BYTES]);
 
-impl MemoPlaintext {
+impl MemoCiphertext {
     /// Encrypt a memo, returning its ciphertext.
-    pub fn encrypt(&self, memo_key: PayloadKey) -> MemoCiphertext {
-        let encryption_result = memo_key.encrypt(self.0.to_vec(), PayloadKind::Memo);
+    pub fn encrypt(memo_key: PayloadKey, memo: &str) -> Result<MemoCiphertext, anyhow::Error> {
+        let memo_len = memo.as_bytes().len();
+        if memo_len > MEMO_LEN_BYTES {
+            return Err(anyhow::anyhow!(
+                "provided memo plaintext of length {memo_len} exceeds maximum memo length of {MEMO_LEN_BYTES}"
+            ));
+        }
+        let mut m = [0u8; MEMO_LEN_BYTES];
+        m[..memo_len].copy_from_slice(memo.as_bytes());
+
+        let encryption_result = memo_key.encrypt(m.to_vec(), PayloadKind::Memo);
         let ciphertext: [u8; MEMO_CIPHERTEXT_LEN_BYTES] = encryption_result
             .try_into()
             .expect("memo encryption result fits in ciphertext len");
 
-        MemoCiphertext(ciphertext)
+        Ok(MemoCiphertext(ciphertext))
     }
 
-    /// Decrypt a `MemoCiphertext` to generate a plaintext `Memo`.
+    /// Decrypt a [`MemoCiphertext`] to generate a plaintext [`String`].
     pub fn decrypt(
-        ciphertext: MemoCiphertext,
         memo_key: &PayloadKey,
-    ) -> Result<MemoPlaintext, anyhow::Error> {
-        let encryption_result = memo_key
+        ciphertext: MemoCiphertext,
+    ) -> Result<String, anyhow::Error> {
+        let plaintext_bytes = MemoCiphertext::decrypt_bytes(memo_key, ciphertext)?;
+        Ok(String::from_utf8_lossy(&plaintext_bytes)
+            .trim_end_matches(0u8 as char)
+            .to_string())
+    }
+
+    /// Decrypt a [`MemoCiphertext`] to generate a fixed-length slice of bytes.
+    pub fn decrypt_bytes(
+        memo_key: &PayloadKey,
+        ciphertext: MemoCiphertext,
+    ) -> Result<[u8; MEMO_LEN_BYTES], anyhow::Error> {
+        let decryption_result = memo_key
             .decrypt(ciphertext.0.to_vec(), PayloadKind::Memo)
             .map_err(|_| anyhow!("decryption error"))?;
-        let plaintext_bytes: [u8; MEMO_LEN_BYTES] = encryption_result
-            .try_into()
-            .map_err(|_| anyhow!("could not fit plaintext into memo size"))?;
-
-        Ok(MemoPlaintext(plaintext_bytes))
+        let plaintext_bytes: [u8; MEMO_LEN_BYTES] = decryption_result.try_into().map_err(|_| {
+            anyhow!("post-decryption, could not fit plaintext into memo size {MEMO_LEN_BYTES}")
+        })?;
+        Ok(plaintext_bytes)
     }
 
-    /// Decrypt a `MemoCiphertext` using the wrapped OVK to generate a plaintext `Memo`.
+    /// Decrypt a [`MemoCiphertext`] using the wrapped OVK to generate a plaintext [`Memo`].
     pub fn decrypt_outgoing(
-        ciphertext: MemoCiphertext,
+        wrapped_memo_key: &WrappedMemoKey,
         wrapped_ovk: OvkWrappedKey,
         cm: note::Commitment,
         cv: balance::Commitment,
         ovk: &OutgoingViewingKey,
         epk: &ka::Public,
-        wrapped_memo_key: &WrappedMemoKey,
-    ) -> Result<MemoPlaintext, anyhow::Error> {
+        ciphertext: MemoCiphertext,
+    ) -> Result<String, anyhow::Error> {
         let shared_secret = Note::decrypt_key(wrapped_ovk, cm, cv, ovk, epk)
             .map_err(|_| anyhow!("key decryption error"))?;
 
@@ -100,11 +88,13 @@ impl MemoPlaintext {
             .decrypt(ciphertext.0.to_vec(), PayloadKind::Memo)
             .map_err(|_| anyhow!("decryption error"))?;
 
-        let plaintext_bytes: [u8; MEMO_LEN_BYTES] = plaintext
-            .try_into()
-            .map_err(|_| anyhow!("could not fit plaintext into memo size"))?;
+        let plaintext_bytes: [u8; MEMO_LEN_BYTES] = plaintext.try_into().map_err(|_| {
+            anyhow!("post-decryption, could not fit plaintext into memo size {MEMO_LEN_BYTES}")
+        })?;
 
-        Ok(MemoPlaintext(plaintext_bytes))
+        Ok(String::from_utf8_lossy(&plaintext_bytes)
+            .trim_end_matches(0u8 as char)
+            .to_string())
     }
 }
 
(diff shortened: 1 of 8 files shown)


crypto/src/memo.rs
```diff
@@ -42,10 +42,17 @@ impl TryFrom<&[u8]> for MemoPlaintext {
...
+        Ok(MemoPlaintext(s))
+    }
+}
 
-        Ok(MemoPlaintext(mp))
+impl From<[u8; MEMO_LEN_BYTES]> for MemoPlaintext {
+    fn from(input: [u8; MEMO_LEN_BYTES]) -> MemoPlaintext {
+        let no_pad = MemoPlaintext::strip_padding(input);
+        let s: String = String::from_utf8_lossy(&no_pad).to_string();
+        println!("Behold the string: {:?}", s);
```
plaidfinch, line 54: I think you mean to delete this line eventually :)

plaidfinch (commented):
I have a suggestion for an alternative approach, let me know what you think:

In the spirit of [parse, don't validate](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/), what if we kept the internal representation of `MemoPlaintext` as a `[u8; MEMO_LEN_BYTES]` but made its contents encapsulated, giving it a (fallible) constructor that takes a string and pads the string with `\0`, or errors if the string has too many bytes? That way we enforce the correct representation at the point of construction, and it's carried through in the type system? On the other end, the `Display` impl for `MemoPlaintext` would strip the trailing `\0` bytes, and decode to UTF-8 lossily (which means that the internal bytes would still be preserved, inspectable by implementing `AsRef<[u8; MEMO_LEN_BYTES]>` and `AsRef<[u8]>` for `MemoPlaintext`).

This line of thinking leads me to realize that we actually can't strip trailing `\0` bytes naively, because a multi-byte UTF-8 character might have a final byte of `\0`, which means if we strip all trailing `\0`, then decode lossily from UTF-8, we end up truncating (and thereby corrupting) the display of the final UTF-8 character if it's a multi-byte character ending in the zero byte. Instead, I think we should pad with *space characters* (`U+0020`), and then strip trailing spaces in the `Display` implementation *after* decoding lossily from UTF-8. This does mean that it is not possible to disambiguate memos with trailing spaces inserted by the user from those where such padding was inserted automatically, but I think this is fine.

An alternate solution to this problem would be to pad with `\0` as per this PR, but decode by first decoding lossily from UTF-8, then stripping trailing [UTF-8 replacement characters](https://doc.rust-lang.org/std/char/constant.REPLACEMENT_CHARACTER.html), which preserves the ability for trailing spaces inserted by the user to be represented faithfully in the `Display` impl.

plaidfinch:
Some more musings:

What's kind of funny about "parse, don't validate" here is that "parsing" usually would imply going *from* `[u8]` *to* `String`, because `String` seems like the more "structured", "high level" type here. But that can be viewed as backwards, I think, because what's being _parsed_ (and thereby validated) is that the `String` does not exceed the byte limit, since the real representation we have to deal with is `[u8; MEMO_LEN_BYTES]` when receiving arbitrary transactions which may or may not have been constructed by our own client software. In the reverse direction, going back from a `[u8; MEMO_LEN_BYTES]` to a `String` is _also_ a kind of parsing, albeit one we're choosing to make into a total function by converting invalid bytes to the replacement character.

Another reason why it's preferable to store `[u8; MEMO_LEN_BYTES]` rather than `String` is it prevents the problem of having a `String` which corresponds to `[u8; MEMO_LEN_BYTES]`, but which itself is actually more than `MEMO_LEN_BYTES` long, because the UTF-8 replacement character, `U+FFFD`, is 3 bytes long. So in the situation where `MemoPlaintext` stores a `String`, if you received a `MemoCiphertext` which decrypted to `[b'\7'; MEMO_LEN_BYTES]` (512 non-printable bell characters), `String::from_utf8_lossy` would convert this into a `String` of length 1536 bytes (512 replacement characters, each 3 bytes). Because `MemoPlaintext` is meant to represent something which can be infallibly encoded into `MemoCiphertext`, client software which tried to re-use this existing `MemoPlaintext` for an outgoing transaction (something we don't do now, but which someone might do) could well panic at the invalid length, since it would be assuming the length is okay. We want to avoid panics which could be induced by attacker-controlled inputs, since this can lead to network de-anonymization attacks, so this feels like a spot where we could very easily write the wrong code. It also means that "receive a `MemoPlaintext` and then send it out in another transaction" does not guarantee that the outgoing transaction has the same `MemoPlaintext` as the original sender used; this is only true if the original memo was UTF-8.

hdevalence:
Per https://github.com/penumbra-zone/penumbra/issues/1510#issuecomment-1287308218 , the reason to prefer `String` is that it's much simpler for downstream API consumers to work with and reason about.  Pretty much every user of the API is going to convert to a `String` to do anything with the data anyways, so we might as well do it for them. Moreover, the users of the `TransactionView`s may not even be in Rust, and may not have access to our custom types.

crypto/src/memo.rs
```diff
@@ -21,17 +21,17 @@ pub const MEMO_LEN_BYTES: usize = 512;
 // The memo is stored separately from the `Note`.
 // TODO: MemoPlaintext should just be a fixed-length string, drop this type entirely
 #[derive(Clone, PartialEq, Eq)]
-pub struct MemoPlaintext(pub [u8; MEMO_LEN_BYTES]);
+pub struct MemoPlaintext(pub String);
```
hdevalence, line 24: Let's eliminate this type entirely, and just use a plain `String` in its place.

crypto/src/memo.rs
```diff
@@ -55,7 +62,7 @@ pub struct MemoCiphertext(pub [u8; MEMO_CIPHERTEXT_LEN_BYTES]);
 impl MemoPlaintext {
     /// Encrypt a memo, returning its ciphertext.
     pub fn encrypt(&self, memo_key: PayloadKey) -> MemoCiphertext {
```
hdevalence, line 64: If we eliminate the `MemoPlaintext` entirely, we won't have a struct to implement this on, so it'll need to be a method on `MemoCiphertext`, and we'll need it to be fallible, because we want to error when the string is too long:
```rust
impl MemoCiphertext {
    pub fn encrypt(memo: &str, memo_key: PayloadKey) -> Result<Self, anyhow::Error> {
        // ..
    }
}
```

crypto/src/memo.rs
```diff
@@ -75,7 +82,17 @@ impl MemoPlaintext {
...
+        let no_pad: Vec<u8> = MemoPlaintext::strip_padding(plaintext_bytes);
+        let plaintext_str = String::from_utf8_lossy(no_pad.as_slice());
+        Ok(MemoPlaintext(plaintext_str.to_string()))
+    }
+
+    /// Removes zero-byte padding from a slice of bytes
+    pub fn strip_padding(input: [u8; MEMO_LEN_BYTES]) -> Vec<u8> {
+        // TODO: Make this less naive; should only strip trailing contiguous zero-bytes.
+        let result: Vec<u8> = input.into_iter().filter(|&x| x != 0u8).collect();
+        result
```
hdevalence, line 95: `trim_end_matches` looks like it might be useful here? https://doc.rust-lang.org/stable/std/primitive.str.html#method.trim_end_matches

plaidfinch:
@hdevalence's alternate approach seems good to me. Perhaps we should provide:

```rust
impl MemoCiphertext {
    fn decrypt(memo: &self, memo_key: PayloadKey) -> String { /* ... */ }
    
    fn decrypt_bytes(memo: &self, memo_key: PayloadKey) -> [u8; MEMO_LEN_BYTES] { /* ... */ }
}
```

This means that clients who care about the exact bytes could ask for them.

conorsch:
Updated based on feedback. Please take another look, @plaidfinch!

crypto/src/memo.rs
```diff
@@ -18,64 +18,51 @@ pub const MEMO_CIPHERTEXT_LEN_BYTES: usize = 528;
...
     pub fn decrypt(
         ciphertext: MemoCiphertext,
         memo_key: &PayloadKey,
-    ) -> Result<MemoPlaintext, anyhow::Error> {
-        let encryption_result = memo_key
+    ) -> Result<String, anyhow::Error> {
+        let plaintext_bytes = MemoCiphertext::decrypt_bytes(ciphertext, memo_key)?;
+        Ok(String::from_utf8_lossy(&plaintext_bytes)
+            .trim_end_matches(0u8 as char)
+            .to_string())
```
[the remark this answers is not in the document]
plaidfinch: Though, if [this test](https://github.com/penumbra-zone/penumbra/pull/1548/files#diff-40dc1380c8a207bd49c7d1b4b0830305a8a5f9c8cf9559797b4d3984f8a15f8cR130) passes, then I suppose it _definitely_ isn't converting it, so maybe you're right!

crypto/src/memo.rs
```diff
@@ -140,23 +129,20 @@ mod tests {
     #[test]
     fn test_memo_encryption_and_decryption() {
```
plaidfinch, line 130: This is a good test! I think it should be pretty easy to convert it to a `proptest` that checks the condition:

> For any string, if its length in bytes is less than or equal to `MEMO_LEN_BYTES`, `string == decrypt(encrypt(string))` and if its length in bytes is greater than `MEMO_LEN_BYTES`, `encrypt(string)` returns an error.

That generalizes this test and the below, and might ferret out any corner cases we could be missing.

plaidfinch: If you haven't used `proptest` before, check out [The Proptest Book](https://altsysrq.github.io/proptest-book/intro.html).

plaidfinch: I feel relatively confident that if these two tests you already have pass, then things probably work, but there could be corner cases still, and I'm mostly commenting to point out the underappreciated fact that writing proptests is often little harder than writing single pointwise unit tests, and provides a huge amount of more confidence in correctness. So, when it's easy (and sometimes when it's hard) I use proptest all the time.

transaction/src/plan/memo.rs
```diff
@@ -9,20 +9,30 @@ use rand::{CryptoRng, RngCore};
 
 #[derive(Clone, Debug)]
 pub struct MemoPlan {
-    pub plaintext: MemoPlaintext,
+    pub plaintext: [u8; MEMO_LEN_BYTES],
```
plaidfinch, line 12: The `MemoPlan` is meant to be human-readable, which means that, I think, this should be a `String` here.

transaction/src/plan/memo.rs
```diff
@@ -9,20 +9,30 @@ use rand::{CryptoRng, RngCore};
...
+        if plaintext_str.as_bytes().len() > MEMO_LEN_BYTES {
+            return Err(anyhow::anyhow!(
+                "provided plaintext for memoplan exceeds maximum memo size"
+            ));
+        }
         let key = PayloadKey::random_key(rng);
-        MemoPlan { plaintext, key }
+        let mut plaintext = [0u8; MEMO_LEN_BYTES];
+        plaintext[..plaintext_str.as_bytes().len()].copy_from_slice(plaintext_str.as_bytes());
+        Ok(MemoPlan { plaintext, key })
```
plaidfinch, line 30: This logic is replicated from the logic in `MemoCiphertext::encrypt`, which feels like a small code smell. If we do indeed store a `String` in the `MemoPlan` rather than an array of bytes (as I recommend above), then we don't need this check at all, since it will be performed when we try to encrypt the memo in `MemoPlan::memo`.

transaction/src/plan/memo.rs
```diff
@@ -9,20 +9,30 @@ use rand::{CryptoRng, RngCore};
...
     }
 
     /// Create a [`MemoCiphertext`] from the [`MemoPlan`].
-    pub fn memo(&self) -> MemoCiphertext {
-        self.plaintext.encrypt(self.key.clone())
+    pub fn memo(&self) -> Result<MemoCiphertext, anyhow::Error> {
+        MemoCiphertext::encrypt(&String::from_utf8_lossy(&self.plaintext), self.key.clone())
```
plaidfinch, line 35: As above, if we make the `MemoPlan` contain a `String`, then we don't need to do a conversion from UTF-8 here, which would be internally converted right back again in `MemoCiphertext::encrypt`.

crypto/src/memo.rs
```diff
@@ -18,64 +18,51 @@ pub const MEMO_CIPHERTEXT_LEN_BYTES: usize = 528;
...
+                "provided memo plaintext exceeds maximum memo size"
+            ));
+        }
+        let mut m = [0u8; MEMO_LEN_BYTES];
+        m[..memo.as_bytes().len()].copy_from_slice(memo.as_bytes());
+
+        let encryption_result = memo_key.encrypt(m.to_vec(), PayloadKind::Memo);
         let ciphertext: [u8; MEMO_CIPHERTEXT_LEN_BYTES] = encryption_result
             .try_into()
             .expect("memo encryption result fits in ciphertext len");
```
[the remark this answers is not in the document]
plaidfinch: We don't have to resolve this to merge this PR, it's more of a "this would be nice" kind of thing.

plaidfinch (commented):
This looks pretty good to me! I have a couple of nits, one or two possible enhancements, and one "pretty sure we should do it slightly differently" that should be small to fix.

Aside from the individual comments, a small thing you can do to produce better documentation:
```
/// ... [`MemoCiphertext`] ...
```
...will generate a link in the docs to that item. It's good practice to do this for every linkable thing you mention in a rustdoc string :)

crypto/src/memo.rs
plaidfinch: So this is more of a nice opportunity for me to plug my favorite software assurance power tool, which I've used to great effect in a lot of places at Penumbra and elsewhere :)

conorsch:
Great comments, thanks for the rapid response. Regarding the changes to `MemoPlan`, I tried that as a String, but the smoke tests (i.e. pcli integration tests were failing), even though `cargo test` was happy. So there's definitely some more parsing to be done around MemoPlan, which I'll work on. The pointer to `proptest` may be helpful there.

The rest of the comments were super straightforward. I'll work on the changes, then ping you again when it's ready for attention.

crypto/src/memo.rs
```diff
@@ -18,64 +18,51 @@ pub const MEMO_CIPHERTEXT_LEN_BYTES: usize = 528;
...
-    pub fn encrypt(&self, memo_key: PayloadKey) -> MemoCiphertext {
-        let encryption_result = memo_key.encrypt(self.0.to_vec(), PayloadKind::Memo);
+    pub fn encrypt(memo: &str, memo_key: PayloadKey) -> Result<MemoCiphertext, anyhow::Error> {
+        if memo.as_bytes().len() > MEMO_LEN_BYTES {
+            return Err(anyhow::anyhow!(
+                "provided memo plaintext exceeds maximum memo size"
+            ));
+        }
+        let mut m = [0u8; MEMO_LEN_BYTES];
+        m[..memo.as_bytes().len()].copy_from_slice(memo.as_bytes());
```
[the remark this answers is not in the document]
conorsch: Necessary! Otherwise:

```
---- memo::tests::test_memo_encryption_and_decryption stdout ----
thread 'memo::tests::test_memo_encryption_and_decryption' panicked at 'source slice length (2) does not match destination slice length (512)', crypto/src/memo.rs:34:15
```

However, I can still make it terser, by reusing a var for the `memo_len` value, since the error message needs that anyway.

crypto/src/memo.rs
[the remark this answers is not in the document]
hdevalence: Yep, `0u8` is valid UTF-8, encoding `U+0000`, so it should be preserved: https://en.wikipedia.org/wiki/UTF-8#Encoding

conorsch:
Updated based on latest feedback. I've not implemented a proptest, as suggested, but I'm willing to follow up with a smaller PR for that. We're bumping up against next testnet, so I consider this PR ready for final review, @plaidfinch.

conorsch:
> I've not implemented a proptest, as suggested, but I'm willing to follow up with a smaller PR for that

Just kidding, appended here. =) Used a separate commit to make it easier to snip out if it complicates review.

wallet/src/plan/planner.rs
```diff
@@ -77,8 +76,8 @@ impl<R: RngCore + CryptoRng> Planner<R> {
 
     /// Set a memo for this transaction plan.
     #[instrument(skip(self))]
-    pub fn memo(&mut self, memo: MemoPlaintext) -> &mut Self {
-        self.plan.memo_plan = Some(MemoPlan::new(&mut self.rng, memo));
+    pub fn memo(&mut self, memo: String) -> &mut Self {
+        self.plan.memo_plan = Some(MemoPlan::new(&mut self.rng, memo).unwrap());
```
hdevalence, line 80: I don't think we want to `unwrap()` here, because then we'll panic if an over-length memo is ever supplied. Instead we should return a `Result`, so that we just error out if the memo is too long.

crypto/src/memo.rs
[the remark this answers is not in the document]
hdevalence: I believe this is an artifact of the underlying symmetric encryption API.

wallet/src/plan/planner.rs
hdevalence: Fixed in https://github.com/penumbra-zone/penumbra/pull/1548/commits/1da961ff0b048af60f4f8320794490199fbdf209 rather than waiting for another review round-trip, since this looks great otherwise

PR #1550, opened 2022-10-25
crypto: remove ad-hoc use of signs in balance commitments

### LINKED ISSUE #1483: Eliminate signedness for balance commitments
> We probably don't want to do this in this refactor, but now that we have a notion of a "balance commitment", I think we should go through and make all the balance commitment handling consistent (no different signs for different actions).

_Originally posted by @hdevalence in https://github.com/penumbra-zone/penumbra/pull/1470#discussion_r980147187_

erwanor:
Fixes #1483 

[`Balance`](https://rustdoc.penumbra.zone/main/penumbra_crypto/balance/struct.Balance.html) keeps track of different `Value`s as well as their signs, and incorporates it to create [commitments](https://github.com/penumbra-zone/penumbra/blob/facc120686043aa0e03b7a00ea62b22ee7814144/crypto/src/balance.rs#L80).

A commitment to a negative value used to be the negation of the value commitment i.e. for a value $v_o$, a commitment to $-v_o$ with randomness $rcv_o$ was defined as: $cv_o = [-v_o] G_v + [-rcv_o] G_{rcv}$.

With the introduction of `Balance::commitment`, we now define the same commitment to a negative value as: $cv_o = [-v_o] G_v + [rcv_o] G_{rcv}$.

This PR eliminate the use of negated value commitments, and replace them with the `Balance::commit` implementation outlined above.

### DIFF
--- a/crypto/src/proofs/transparent.rs
+++ b/crypto/src/proofs/transparent.rs
@@ -779,10 +779,11 @@ mod tests {
     use super::*;
     use crate::{
         keys::{SeedPhrase, SpendKey},
-        note, Note, Value, Balance,
+        note, Balance, Note, Value,
     };
 
     #[test]
+    /// Check that the `OutputProof` verification suceeds.
     fn test_output_proof_verification_success() {
         let mut rng = OsRng;
 
@@ -816,6 +817,8 @@ mod tests {
     }
 
     #[test]
+    /// Check that the `OutputProof` verification fails when using an incorrect 
+    /// note commitment.
     fn test_output_proof_verification_note_commitment_integrity_failure() {
         let mut rng = OsRng;
 
@@ -829,6 +832,9 @@ mod tests {
             amount: 10u64.into(),
             asset_id: asset::REGISTRY.parse_denom("upenumbra").unwrap().id(),
         };
+
+        let balance_to_send = -Balance::from(value_to_send);
+
         let v_blinding = Fr::rand(&mut rng);
         let note = Note::generate(&mut rng, &dest, value_to_send);
         let esk = ka::Secret::new(&mut rng);
@@ -850,7 +856,7 @@ mod tests {
 
         assert!(proof
             .verify(
-                -value_to_send.commit(v_blinding),
+                balance_to_send.commit(v_blinding),
                 incorrect_note_commitment,
                 epk
             )
@@ -858,6 +864,8 @@ mod tests {
     }
 
     #[test]
+    /// Check that the `OutputProof` verification fails when using an incorrect 
+    /// balance commitment.
     fn test_output_proof_verification_balance_commitment_integrity_failure() {
         let mut rng = OsRng;
 
@@ -872,6 +880,10 @@ mod tests {
             asset_id: asset::REGISTRY.parse_denom("upenumbra").unwrap().id(),
         };
         let v_blinding = Fr::rand(&mut rng);
+
+        let bad_balance = Balance::from(value_to_send);
+        let incorrect_balance_commitment = bad_balance.commit(Fr::rand(&mut rng));
+
         let note = Note::generate(&mut rng, &dest, value_to_send);
         let esk = ka::Secret::new(&mut rng);
         let correct_epk = esk.diversified_public(&note.diversified_generator());
@@ -881,7 +893,6 @@ mod tests {
             v_blinding,
             esk,
         };
-        let incorrect_balance_commitment = value_to_send.commit(Fr::rand(&mut rng));
 
         assert!(proof
             .verify(incorrect_balance_commitment, note.commit(), correct_epk)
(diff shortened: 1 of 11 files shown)


crypto/src/proofs/transparent.rs
```diff
@@ -809,7 +812,7 @@ mod tests {
         };
 
         assert!(proof
-            .verify(-value_to_send.commit(v_blinding), note.commit(), epk)
```
redshiftzero, line 812: minor: there are two other places in these transparent proof tests where you can remove the signs on value just to prevent confusion in the future (lines 851, 917)

crypto/src/proofs/transparent.rs
```diff
@@ -797,6 +797,9 @@ mod tests {
             amount: 10u64.into(),
             asset_id: asset::REGISTRY.parse_denom("upenumbra").unwrap().id(),
         };
+
+        let balance = -crate::Balance::from(value_to_send);
```
redshiftzero, line 801: I think the test failure is due to mismatching signs, I think we can remove the conversion to Balance and call commit directly on `value_to_send` below (i.e. not `-value_to_send`)

redshiftzero (commented):
Just scanned through and I think this covers all the places to update. When tests pass (one minor thing in comments re: the test failure), LGTM

crypto/src/proofs/transparent.rs
```diff
@@ -122,7 +122,7 @@ impl OutputProof {
         gadgets::note_commitment_integrity(self.note.clone(), note_commitment)?;
 
         gadgets::balance_commitment_integrity(
-            -balance_commitment,
+            balance_commitment,
```
erwanor, line 125: @redshiftzero is it actually correct to remove the minus here?

`OutputProof::verify` expects a balance commitment that's negative (or, does it?), and the gadget will check integrity against the note commitment (positive, afaict). I could make `test_output_proof_verification_success` pass by feeding `verify` with a positive balance commitment, but since it's an `OutputProof` shouldn't it be negative?

crypto/src/proofs/transparent.rs
erwanor: Thanks!

crypto/src/proofs/transparent.rs
redshiftzero: good question: the output proof doesn't check the signs, it's only checking that the commitment has been formed properly, i.e. that the value (should always be positive even for something that will ultimately subtract from the value balance in the transaction) and the blinding factor are an opening of the public value/balance commitment. Later on when we sum the commitments (the verifier does this) the appropriate sign will be used to ensure that the action has the appropriate contribution to the value balance, and that is checked cryptographically using the binding signature at the transaction level.

erwanor: got it, but the balance commitment won't open to the note value commitment because they're different:

the balance commitment:

$cv_o = [-v_o] G_v + [rcv_o] G_{rcv}$

the note commitment, in the happy path:

$cv_o = [v_o] G_v + [rcv_o] G_{rcv}$

This is how $cv_o$ is computed in `OutputPlan::output_body`:
```rust
     let balance_commitment = self.balance().commit(self.value_blinding); // balance is negated
```

on the other hand, this works well for `SpendProof`, so maybe we could modify `gadgets::balance_commitment_integrity` to take balances instead of values:

```rust
/// Check the integrity of the balance (previously value) commitment.
pub(crate) fn balance_commitment_integrity(
    balance_commitment: balance::Commitment,
    balance_blinding: Fr,
    balance: Balance,
) -> Result<()> {
    if balance_commitment != balance.commit(value_blinding) {
        Err(anyhow!("balance commitment mismatch"))
    } else {
        Ok(())
    }
}
```

where `balance` is built from the note value, and `blinding` is specified within the proof. In the case of `OutputProof` we would negate the balance, and in the case of `SpendProof` we wouldn't. This way we get closer to eliminating `Value::commit` and the gadget  logic doesn't change.

I think I would also add a couple tests in `OutputPlan` and `SpendPlan` to anchor the way we prepare / verify commitments, since `penumbra_transaction` imports `penumbra_crypto`.

redshiftzero: ahhh gotcha, so since the change in your PR here now uses `Balance` in `OutputPlan::output_body`, we'd also need to flip the sign of the blinding factor for all else to stay the same. Instead I think your suggestion to modify the balance commitment integrity check to take a balance could be good. Tests to anchor this behavior in penumbra_transaction would also be great!

crypto/src/proofs/transparent.rs
erwanor: I've kept the negated `Balance` following what we talked about (preparing to remove `Value::commit`)

redshiftzero (approved):
LGTM!

PR #1554, opened 2022-10-26
spec: add example on binding signature

redshiftzero:
This is an example to describe how the binding signature / value balance trick works. Is this helpful? Are there issues with this explanation?

### DIFF
--- a/docs/protocol/src/crypto/decaf377-rdsa.md
+++ b/docs/protocol/src/crypto/decaf377-rdsa.md
@@ -76,12 +76,9 @@ Randomizing a signing key and then deriving the verification key associated to t
 The second signature domain used in Penumbra is for binding signatures.  The
 basepoint $B_{\mathsf{Binding}}$ is the result of converting
 `blake2b(b"decaf377-rdsa-binding")` to an $\mathbb F_r$ element and applying
-`decaf377`'s CDH map-to-group method.
+`decaf377`'s CDH encode-to-curve method.
 
 Since the verification key corresponding to the signing key $a \in \mathbb F_r$ is $A = [a]B_D$, adding and subtracting signing and verification keys commutes with derivation of the verification key, as desired.
-
-**WARNING**: the `decaf377` CDH map is unstable, so it may change in the future and change binding signatures with it.
-
 
 [^1]: This situation is a good example of why it's better to avoid the terms
 "public key" and "private key", and prefer more precise terminology that names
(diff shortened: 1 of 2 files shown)


docs/protocol/src/crypto/decaf377-rdsa.md
```diff
@@ -76,15 +76,51 @@ Randomizing a signing key and then deriving the verification key associated to t
...
 attribute of how they're commonly used. In this example, the verification key
 should not be public, since it could link different transactions.
+
+### Simple example: Binding signature
+
+Let’s say we have two actions in a transaction: one spend (indicated with subscript $s$) and one output (indicated with subscript $o$).
+
+The balance commitments for those actions are:
+
+$cv_o = [v_o] G_v + [rcv_o] G_{rcv}$
```
erwanor, line 95: How would you feel about making $v_o$ a positive value, and show the negative sign like this:

$cv_o = - [v_o] G_v + [rcv_o] G_{rcv}$

so that we can write the verification key:

$bvk = cv_s + cv_o = [v_s - v_o] G_v + [rcv_s + rcv_o] G_{rcv}$

it's the same, but I think it matches how we do it in practice:
```rust
for action in &self.transaction_body.actions {
    balance_commitments += action.balance_commitment().0;
}
```

erwanor:
That example made a bunch of things click, thanks for writing it!

docs/protocol/src/crypto/decaf377-rdsa.md
redshiftzero: You are correct that we are adding all the balance commitments, however the balance commitment for e.g. an output is computed as $-cv_o$ where $cv_o = [v_o] G_v + [rcv_o] G_{rcv}$ (this happens when constructing `Output::Body`), so the balance commitment being added for an output is: 

$-[v_o] G_v - [rcv_o] G_{rcv}$

i.e. we also have a negative sign in front of the blinding factor. So I'm on the fence because we're not committing to a negative value, we're putting a negative sign in front of the full commitment. What do you think? I started re-defining $cv_o$ as the $-[v_o] G_v - [rcv_o] G_{rcv}$ but it's a little non-obvious at first where the sign came from until we get to explaining that we are adding or subtracting the commitments based on their contribution to the transaction balance. 

Related: one other thing I think is a little confusing relative to the code is that here in the example we are committing to individual values using individual blinding factors for each value using `Value::commit`, and then adding or subtracting the resulting commitments. But if one looks at the `Balance::commit` method however, a single blinding factor is used (since this was to be used for committing to multiple values in one shot). The two methods are equivalent if one uses a blinding factor in `Balance::commit` that is the sum of the individual blinding factors with the appropriate signs. For a concrete example, I've added [this bit of a test](https://github.com/penumbra-zone/penumbra/pull/1554/commits/3ad5160a0f099be791183093798a2d2e4a51eb14).

PR #1558, opened 2022-10-27
Integrate new storage code

zbuc:
Integrates the `storage2` crate and overhauls the `Component` implementations to the new execution model.


hdevalence:
The current state of this branch is that all of the component implementations (except the IBC component, which has been commented out) and the `pd` code that uses them have been updated to use the new APIs provided by the `storage2` crate.

The smoke tests should pass, and at that point, I think we're good to squash and merge this PR.  We can update the IBC component and do other improvements afterwards, but this completes the "stop the world" integration work.

PR #1559, opened 2022-10-27
Document protobuf build process

### LINKED ISSUE #1530: update dev guide to explain proto build process
We should add a section to the dev guide about generating protos explaining:

1. The instructions in https://github.com/penumbra-zone/penumbra/tree/main/tools/proto-compiler about how to compile the protos
2. That the version of protoc in apt is pretty old and to follow these instructions if you're not on mac: https://grpc.io/docs/protoc-installation/#install-pre-compiled-binaries-any-os 

See PR #1475 which has helpful text for the guide

conorsch:
Updated the dev guide to include instructions on rebuilding the protobuf spec files. While I was in the guide, made some small touch-ups to URLs throughout. Also tacked on a new CI check, so that forgetting to rebuild and commit the protobuf specs will now result in failing builds, so we can catch that mistake prior to merging into main. You can see a failing CI check based on a temporary (now removed) commit [here](https://github.com/penumbra-zone/penumbra/actions/runs/3341262009/jobs/5532290154).

Closes #1530.

### DIFF
--- a/docs/guide/src/dev/protobuf.md
+++ b/docs/guide/src/dev/protobuf.md
@@ -1,0 +1,50 @@
+# Maintaining protobuf specs
+
+The Penumbra project dynamically generates code for interfacing
+with [gRPC]. The following locations within the repository
+are relevant:
+
+  * `proto/proto/penumbra/**/*.proto`, the developer-authored spec files
+  * `proto/src/gen/*.rs`, the generated Rust code files
+  * `tools/proto-compiler/`, the build logic for generated the Rust code files
+
+We use [buf] to auto-publish the protobuf schemas at
+[buf.build/penumbra-zone/penumbra][protobuf], and to generate Go and Typescript packages.
+The Rust code files are generated with our own tooling, located at `tools/proto-compiler`.
+
+Our custom tooling for generating the Rust files will also shape the Serde implementations
+of the derived Rust types to have more favorable JSON output (such as rendering
+addresses as [Bech32]-encoded strings).
+
+## Installing protobuf
+
+Obtain the most recent pre-compiled binary from the [`protoc` website].
+After installing, run `protoc --version` and confirm you're running
+at least `3.21.8` (or newer). Don't install `protoc` from package managers
+such as `apt`, as those versions are often outdated, and will not work
+with Penumbra.
+
+## Building protobuf
+
+Switch to the [proto-compiler] directory and run the tool:
+
+```shell
+cd tools/proto-compiler
+cargo run
+```
+
+Then run `git status` to determine whether any changes were made.
+The build process is deterministic, so regenerating multiple times
+from the same source files should not change the output.
+A possible exception to this rule is if `prost` makes a superficial
+change to the output that isn't substantive.
+
+If the generated output would change in any way, CI will
+fail, prompting the developer to commit the changes.
+
+[`protoc` website]: https://grpc.io/docs/protoc-installation/#install-pre-compiled-binaries-any-os
+[proto-compiler]: https://github.com/penumbra-zone/penumbra/tree/main/tools/proto-compiler
+[gRPC]: https://grpc.io/
+[protobuf]: https://buf.build/penumbra-zone/penumbra
+[buf]: https://buf.build/
+[Bech32]: https://en.bitcoin.it/wiki/Bech32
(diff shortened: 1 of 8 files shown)


plaidfinch:
This looks great :D

plaidfinch:
Feel free to merge this in yourself as soon as CI completes :)

.github/workflows/protobuf.yml
```diff
@@ -0,0 +1,42 @@
...
+        run: |
+          cd tools/proto-compiler
+          cargo run
+          if ! git diff --quiet; then
+              echo "ERROR: protobuf files must be regenerated and committed:"
+              git diff --stat
+              exit 1
+          else
+              echo "OK: no changes required to protobuf specs"
+          fi
```
hdevalence, line 42: This will fail not just if the protos haven't been regenerated, but also if there's any changes to the output produced by `prost`, regardless if that output is meaningfully different (e.g., if a new minor version changed the whitespace).  This could lead to a bad situation where the build fails because of slightly different versions between the PR author and the CI server.

I'm not sure how likely this is, or, even if it happened, how annoying it would be, but it's something that we should watch out for.  (It's definitely good to try to catch the more common case, where someone forgets to rerun the proto compiler after editing protos).

.github/workflows/protobuf.yml
```diff
@@ -0,0 +1,42 @@
+---
+name: Protobuf specs
+on: [pull_request]
+jobs:
+  protobuf-fresh:
```
hdevalence, line 5: We already have a file defining GH Actions for protobufs -- could we put this action in the same file as those, so we have one place for all of the proto-related actions?

docs/guide/src/dev/protobuf.md
```diff
@@ -0,0 +1,38 @@
+# Maintaining protobuf specs
+
+The Penumbra project dynamically generates code for interfacing
+with [gRPC]. The following locations within the repository
+are relevant:
+
+  * `proto/proto/penumbra/**/*.proto`, the developer-authored spec files
+  * `proto/src/gen/*.rs`, the generated output files
+  * `tools/proto-compiler/`, the build logic for the generation
```
hdevalence, line 9: In this section it might be helpful to link to https://buf.build/penumbra-zone/penumbra and mention that we auto-publish our protobuf schemas with Buf, and rely on Buf for generating Go and Typescript packages, but generate the Rust code with our own tooling.

The Buf pages render fairly nice docs, so that's worth calling out.

docs/guide/src/dev/protobuf.md
```diff
@@ -0,0 +1,38 @@
...
+```shell
+cd tools/proto-compiler
+cargo run
+```
+
+Then run `git status` to determine whether any changes were made.
+The build process is deterministic, so regenerating multiple times
+from the same source files will not change the output.
+
+If changes were made, commit those changes to the repository and PR in.
```
hdevalence, line 34: It might be worth mentioning that the code in `proto-compiler` crate also "shapes" the Serde implementations of the derived Rust types to have more favorable JSON output (e.g., rendering addresses as Bech32-encoded strings, etc).

hdevalence (approved):
Awesome!

.github/workflows/protobuf.yml
conorsch: Great point, I hadn't realized that. Will add a descriptive comment in the workflow declaration explaining the trade-offs.

plaidfinch: The exact version of prost should be the same between CI and dev machine, because we have a `Cargo.lock` file, right? I suspect that we might possibly see a CI failure on a minor version bump of prost, but I don't think it will happen spuriously, it should happen when the `Cargo.lock` is updated by `cargo update`.

Conjecture: I also suspect that prost's code generation isn't likely to change spacing, because if it's printing a Rust syntax tree, that should be more stable than prost itself, since that would come from a crate like `syn` that provides parsing/printing of the Rust AST.

If it turns out that this produces a lot of false positives, we can revisit the check, but until then, I think it's very good to have.

One alternative which could be better, if there are lots of false positives: could we run the same check as a precommit hook?

conorsch: In my experience precommit hooks help individual developers to maintain baseline conformance, but aren't as useful for teams, since they're opt-in. I'd very much prefer the check to live in CI, where no one need think of it, until it helpfully nags. Should it begin to nag unhelpfully, then I hereby volunteer as tribute to take a closer look.

plaidfinch: sgtm!

conorsch:
Great feedback! All comments addressed. Will merge after CI is pleased.

PR #1560, opened 2022-10-28
Add a `--raw` option to `pcli view tx`

hdevalence:
Most of this work consisted of connecting the proto and domain types.


hdevalence:
The resulting output is a bit messy, but we can refine the JSON at any point.

Issue #1393, opened 2022-09-11
`pcli` doesn't handle ephemeral address indexes correctly


hdevalence:
**Describe the bug**

`pcli` gets kind of borked when the view service has a note that was sent to an ephemeral address, because there are various places in the code where it's assumed that address indexes are always numeric.

Examples:

- `pcli view balance --by-address` will display 128-bit address indexes, as if they were integers, but those integers can't be plugged in as a source address
- When swap claims were output to an ephemeral address, they broke the spend planner, because it barfed on seeing a note with a non-numeric address index.

We should figure out how we want to deal with this.  Currently, pcli's `source_address` feature is a way to use addresses as pseudo-accounts. This seems useful, because it allows having multiple accounts (for accounting purposes) without having to do extra scanning work. Do we want to have any way for the CLI to distinguish between different ephemeral addresses, or should it treat "all ephemeral addresses" as a single pseudo-account?

plaidfinch:
The second bullet point is outdated, right?

mikayla-maki:
Talked with @plaidfinch about tackling this today and evaluated the two solutions:
- Collapsing down all ephemeral transactions to one ephemeral account. This might have a usability issue as users might want to know the actual address for each ephemeral account in order to confirm that they received a trade properly.
- Expanding the table to show the associated index, with a placeholder for ephemeral addresses, and an addresses prefix. This might also have usability problems for people who have a large number of external transactions. Most of the addresses would just be noise.

hdevalence:
What's the concern with collapsing all of the ephemeral addresses into one row? We'll still be able to list off individual transactions, or display individual note balances, so a user can confirm specific transactions if they want.

I think we want to logically group all of the ephemeral addresses, rather than support their being treated separately, because we already have a mechanism to allow users to track activity on a per-address basis, namely, by using indexed addresses.

mikayla-maki:
Good point on how displaying transactions covers that use case. I’ll focus on treating all ephemeral addresses as one address then :D

PR #1566, opened 2022-10-28
 Rework pcli view balance code to hide and combine ephemeral addresses

mikayla-maki:
This fixes https://github.com/penumbra-zone/penumbra/issues/1393

While implementing this, I attempted to refactor this command's code to reduce duplication. I normalized the row representations into a single type and then moved the table generation to the end of the function. This required  unwrapping the value in the `Addr Index` column if the `--by-address` flag is set as the notes generated for `pcli view balance` don't have any address information. End users should never see the except case on `pcli/src/command/view/balance.rs:171`.

Other than those improvements, the code is still a bit of a mess. The fundamental problem is that `pcli view balance` is manually implementing a  `SELECT SUM(note.amount) GROUP BY asset` query on the note data. `pcli view balance` needs just the list by `Asset -> Note` but doesn't have a way of asking for it from the view, and so has to parse a `Asset -> AddressIndex -> Note`. Meanwhile `pcli view balance --by-address`, needs to do this same group-by-asset operation, but only after it's been grouped by the address index: `AddressIndex -> Asset -> Note`. These types are too different to be generic over, so all of these queries should probably just be implemented in SQL as queries on the view.

For testing, I did a smoke test using my personal notes in the Eirene testnet, here's the test data I had available:

 `pcli view address --by-address` on main produces:

```
 Addr Index                               Amount
 0                                        1800penumbra
 221785008525830479505245184476247484031  200penumbra
 271924351478716824407855945686860083103  10gm
 271924351478716824407855945686860083103  10pizza
 271924351478716824407855945686860083103  100penumbra
 271924351478716824407855945686860083103  10gn
```

Running `pcli view address --by-address` on this fork produces:

```
 Addr Index  Amount
 0           1800penumbra
 Ephemeral   10gm
 Ephemeral   10pizza
 Ephemeral   300penumbra
 Ephemeral   10gn
```

So, it looks like it works 👀

### DIFF
--- a/pcli/src/command/view/balance.rs
+++ b/pcli/src/command/view/balance.rs
@@ -2,7 +2,7 @@ use std::collections::BTreeMap;
 
 use anyhow::Result;
 use comfy_table::{presets, Table};
-use penumbra_crypto::{keys::AddressIndex, FullViewingKey, Value};
+use penumbra_crypto::{asset::Cache, keys::AddressIndex, FullViewingKey, Value};
 use penumbra_view::ViewClient;
 #[derive(Debug, clap::Args)]
 pub struct BalanceCmd {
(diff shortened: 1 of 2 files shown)


hdevalence (approved):
Awesome, thanks!

PR #1567, opened 2022-10-29
storage2: add a RocksDbSnapshot wrapper with internal Arc

### LINKED ISSUE #1512: Find/Implement workaround for 'static lifetime of Snapshot instances
Currently we've temporarily implemented our `Snapshot` struct to wrap a `rocksdb::Snapshot` with a `'static` lifetime:

```rust
pub(crate) struct Snapshot {
    // TODO: the `'static` lifetime is a temporary hack and we'll need to find a workaround separately
    rocksdb_snapshot: rocksdb::Snapshot<'static>,
    jmt_version: jmt::Version,
}
```

This was the quickest way to get an implementation underway, since `rocksdb::Snapshot` holds a `&db: DBAccess` which requires a lifetime constraint. However, in the real world, it's unlikely that a `'static` lifetime is actually feasible for us to use here.

We'll need to find some kind of workaround.

Ideas:

1. Fork the Rust rocksdb library and provide a better-shaped API
2. Perform some unsafe pointer wrangling to reborrow a pointer dereference and avoid lifetime checking
3. Use [std::mem::transmute](https://doc.rust-lang.org/std/index.html)::[mem](https://doc.rust-lang.org/std/mem/index.html)::[transmute](https://doc.rust-lang.org/std/mem/fn.transmute.html#) to get a different lifetime
4. At `pd` start, `Box` the db instance (rather than `Arc<...>`) and leak the box so we have a `'static` reference valid for the lifetime of the program

hdevalence:
This change works around a limitation of the `rocksdb` API: the `rocksdb::Snapshot` can only take a borrowed database handle, not an `Arc`'d one, so the lifetime of the `rocksdb::Snapshot` is bound to the lifetime of the borrowed handle.  This isn't what we want, because we're using all of these snapshots in an async context, where objects need to be able to live arbitrarily long (so that they can be included in futures that are spawned as tasks, and executed whenever).

To fix this, we add a wrapper type that bundles an `Arc`'d handle together with the `rocksdb::Snapshot`, so that the database is guaranteed to live at least as long as any snapshot of it.

This requires unsafe code, which is encapsulated in the single-purpose `RocksDbSnapshot` wrapper type, so that it's self-contained and it's possible to reason about the unsafe code locally.

Closes #1512

### DIFF
--- a/storage2/src/snapshot.rs
+++ b/storage2/src/snapshot.rs
@@ -9,38 +9,38 @@ use tracing::Span;
 
 use crate::state::StateRead;
 
-/// Snapshots maintain a point-in-time view of the underlying storage, suitable
-/// for read-only access by multiple threads, i.e. RPC calls.
+mod rocks_wrapper;
+use rocks_wrapper::RocksDbSnapshot;
+
+/// A snapshot of the underlying storage at a specific state version, suitable
+/// for read-only access by multiple threads, e.g., RPC calls.
 ///
-/// This is implemented as a wrapper around a [RocksDB snapshot](https://github.com/facebook/rocksdb/wiki/Snapshot)
-/// with an associated JMT version number for the snapshot.
+/// Snapshots are cheap to create and clone.  Internally, they're implemented as
+/// a wrapper around a [RocksDB
+/// snapshot](https://github.com/facebook/rocksdb/wiki/Snapshot) with a pinned
+/// JMT version number for the snapshot.
 #[derive(Clone)]
-pub(crate) struct Snapshot(pub(crate) Inner);
+pub struct Snapshot(Arc<Inner>);
 
 // We don't want to expose the `TreeReader` implementation outside of this crate.
-#[derive(Clone)]
-pub(crate) struct Inner {
-    // TODO: the `'static` lifetime is a temporary hack and we'll need to find a workaround separately (tracked in #1512)
-    rocksdb_snapshot: Arc<rocksdb::Snapshot<'static>>,
-    db: &'static rocksdb::DB,
-    jmt_version: jmt::Version,
+struct Inner {
+    snapshot: RocksDbSnapshot,
+    version: jmt::Version,
+    // Used to retrieve column family handles.
+    db: Arc<rocksdb::DB>,
 }
 
 impl Snapshot {
-    pub(crate) fn new(
-        rocksdb_snapshot: Arc<rocksdb::Snapshot<'static>>,
-        jmt_version: jmt::Version,
-        db: &'static rocksdb::DB,
-    ) -> Self {
-        Self(Inner {
-            rocksdb_snapshot,
-            jmt_version,
+    pub(crate) fn new(db: Arc<rocksdb::DB>, version: jmt::Version) -> Self {
+        Self(Arc::new(Inner {
+            snapshot: RocksDbSnapshot::new(db.clone()),
+            version,
             db,
-        })
+        }))
     }
 
-    pub fn jmt_version(&self) -> jmt::Version {
-        self.0.jmt_version
+    pub fn version(&self) -> jmt::Version {
+        self.0.version
     }
 }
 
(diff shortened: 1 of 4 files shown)


storage2/src/snapshot.rs
```diff
@@ -9,38 +9,38 @@ use tracing::Span;
...
 ///
-/// This is implemented as a wrapper around a [RocksDB snapshot](https://github.com/facebook/rocksdb/wiki/Snapshot)
-/// with an associated JMT version number for the snapshot.
+/// Snapshots are cheap to create and clone.  Internally, they're implemented as
+/// a wrapper around a [RocksDB
+/// snapshot](https://github.com/facebook/rocksdb/wiki/Snapshot) with a pinned
+/// JMT version number for the snapshot.
 #[derive(Clone)]
-pub(crate) struct Snapshot(pub(crate) Inner);
+pub struct Snapshot(Arc<Inner>);
```
hdevalence, line 23: Moving to an internal refcount simplifies the implementation, because when using `spawn_blocking`, we effectively need a way to access the same `Snapshot` from multiple threads.  The `Arc<Inner>` pattern means we just have to increment a refcount to pass a copy of ourself to the task we spawn.

storage2/src/snapshot.rs
```diff
@@ -9,38 +9,38 @@ use tracing::Span;
...
-pub(crate) struct Inner {
-    // TODO: the `'static` lifetime is a temporary hack and we'll need to find a workaround separately (tracked in #1512)
-    rocksdb_snapshot: Arc<rocksdb::Snapshot<'static>>,
-    db: &'static rocksdb::DB,
-    jmt_version: jmt::Version,
+struct Inner {
+    snapshot: RocksDbSnapshot,
+    version: jmt::Version,
+    // Used to retrieve column family handles.
+    db: Arc<rocksdb::DB>,
```
hdevalence, line 30: It's slightly dissatisfying that we have to keep the `db` handle around for column family lookups-by-string, rather than storing `jmt_cf`, ... etc, `ColumnFamily` instances. The reason we have to do this is that the `ColumnFamily` type from `rocksdb` is a wrapper around an FFI'd column family pointer, so it's neither `Copy` nor `Clone`, and the `DB` only gives a borrowed `&ColumnFamily` from a hashmap that it stores internally.  So avoiding the hashmap would mean building something similar to the `RocksDbSnapshot` "internal Arc" wrapper but for column families, which doesn't seem worthwhile.

storage2/src/snapshot.rs
hdevalence, line 23: This commit simplifies visibility from `pub(crate) Struct(pub(crate) Inner)` to `pub Struct(Inner)`, so that visibility of the outer struct is controlled by whether or not the parent module re-exports it, and the inner type is completely encapsulated.  I think this is generally a bit cleaner.

storage2/src/snapshot.rs
```diff
@@ -130,18 +138,20 @@ impl StateRead for Snapshot {
 
 /// A reader interface for rocksdb. NOTE: it is up to the caller to ensure consistency between the
 /// rocksdb::DB handle and any write batches that may be applied through the writer interface.
-impl TreeReader for Inner {
+impl TreeReader for Snapshot {
```
hdevalence, line 141: The `TreeReader` implementation is now on the `Snapshot`, since the `Inner` type is hidden from the rest of the crate.  But we don't expose it outside of the crate anyways.

storage2/src/storage.rs
```diff
@@ -12,13 +12,15 @@ use tracing::Span;
 use crate::snapshot::Snapshot;
 use crate::State;
 
+/// A handle for a storage instance, wrapping a RocksDB instance.
+#[derive(Clone)]
+pub struct Storage(Arc<Inner>);
```
hdevalence, line 17: The same logic as https://github.com/penumbra-zone/penumbra/pull/1567/files#r1008648947 simplifies the implementation of committing new data to storage, and also gives an easy answer for how different threads access the `Storage` handle they use to create `State` forks.

storage2/src/storage.rs
```diff
@@ -59,15 +57,16 @@ impl Storage {
     /// `Storage`, or `None` if the tree is empty.
     pub async fn latest_version(&self) -> Result<Option<jmt::Version>> {
         // TODO: do better
-        Ok(latest_version(self.0.db).unwrap())
+        latest_version(&self.0.db)
```
hdevalence, line 60: I wasn't sure what `do better` meant here, maybe it's about doing the read asynchronously?  One idea would be to read it out of the `latest_snapshot`, but I wasn't 100% clear that the existing code correctly handles the case where the tree is empty.

storage2/src/storage.rs
```diff
@@ -59,15 +57,16 @@ impl Storage {
...
     }
 
     /// Returns a new [`State`] on top of the latest version of the tree.
-    pub async fn state(&self) -> State {
+    pub fn state(&self) -> State {
```
hdevalence, line 64: This can be non-`async`, since we're just cloning the snapshot handle we already have.

storage2/src/storage.rs
```diff
@@ -59,15 +57,16 @@ impl Storage {
...
         State::new(self.0.latest_snapshot.read().clone())
     }
 
-    pub async fn apply(&'static mut self, state: State) -> Result<()> {
+    pub async fn apply(&self, state: State) -> Result<()> {
```
hdevalence, line 68: This takes `&self`, not `&mut self`, because it's handling synchronization internally, so it doesn't need an exclusive reference (cf [this distinction](https://docs.rs/dtolnay/latest/dtolnay/macro._02__reference_types.html))

hdevalence:
Merging this optimistically since it changes the API for the `Storage` type and prevents incorrect `'static` bounds from propagating through the codebase, but happy to continue review in future PRs.

Issue #1555, opened 2022-10-26
Change `ViewProtocol`'s `WitnessRequest` to take a `TransactionPlan`


hdevalence:
**Is your feature request related to a problem? Please describe.**

Transaction construction has three steps:

1. Send an authorization request to the custody service to get auth data;
2. Send a witness request to the view service to get witness data;
3. Combine the auth data, the witness data, and the transaction plan to build the transaction. 

Currently, a transaction builder has to do a somewhat complex dance to construct the `WitnessRequest` and the witness data, because of handling of dummy notes: https://github.com/penumbra-zone/penumbra/blob/main/wallet/src/build.rs#L30-L56

This is necessary because the view service *should* error when someone requests witness data that it doesn't know about, but it can't do this if dummy note commitments are included, because it would need a way to know which note commitments are dummy notes.

**Describe the solution you'd like**

If we change the `ViewProtocol`'s `WitnessRequest` to take a `TransactionPlan` instead, we can move the logic in https://github.com/penumbra-zone/penumbra/blob/main/wallet/src/build.rs#L30-L56 into the view service itself. This way the caller doesn't have to do anything special to construct the request, and we're future-proof against future changes/extensions to witness data.

PR #1568, opened 2022-10-29
client: change ViewTransaction::witness() to take TransactionPlan

therealyingtong:
Closes https://github.com/penumbra-zone/penumbra/issues/1555


hdevalence:
Hey! Sorry for the delayed followup. This looks great. In the future, I think it'd be good to push the change all the way out into the protobufs, but I'm not sure it matters at the moment because the only clients that would be requesting witness data for proving are using the Rust code to build the transaction.

PR #1569, opened 2022-10-29
[WIP] storage2: add an end-to-end test of storage/state flow and do a docs/API pass

hdevalence:
Before starting the replacement of `storage` with `storage2`, it seems like a good idea to have an end-to-end integration test to check that the external API is usable on its own and that it appears to work correctly.

This PR does a docs/API pass on the external API, then adds the skeleton of a simple end-to-end test. The test is verbose but very explicit. (It does not currently pass).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3405,6 +3405,7 @@ dependencies = [
  "tokio-stream",
  "tonic",
  "tracing",
+ "tracing-subscriber 0.3.16",
 ]
 
 [[package]]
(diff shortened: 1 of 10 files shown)


hdevalence:
Looks like there's a recursion leading to a stack overflow in the current implementation of `StateTransaction::prefix_raw`: it calls `prefix_raw_with_cache`, which itself calls `prefix_raw`.

I think there's also an issue with reading data from the JMT, but I'm unsure.

zbuc:
The infinite recursion bug was because we were passing the wrong thing to the `prefix_raw_with_cache` calls; it needed the underlying storage, not `self`.

Issue #1378, opened 2022-09-09
Tracking Issue: upgrade to 128-bit amount values in notes


avahowell:
Following #1336, where we have decided to move to 128 bit amounts in our notes, we need to do the implementation. A few steps are as follows

- [ ] Update the spec to specify how we handle bridged assets, where we will return an error if we can't fit an asset in a u128, and the exact encoding for the u128
- [x] Implement a new `Amount` domain type that encapsulates a u128, and provides functionality for encoding the u128 as two u64s for use in our protobufs.
- [x] Propagate the new `Amount` type throughout the codebase, replacing u64 where it was used previously.

hdevalence:
@avahowell What's the current status of this issue?

avahowell:
We have the new amount type now, but it is still using a u64 internally

hdevalence:
Following #1819, what's the status of this on the client side? We at least need to be able to record 128-bit values in sqlite in the rust view server (cc @aubrika).

conorsch:
@aubrika Could you summarize your opinions on our options here? I believe there was some relevant discussion in Discord recently that you may want to cross-link.

aubrika:
> @aubrika Could you summarize your opinions on our options here? I believe there was some relevant discussion in Discord recently that you may want to cross-link.

I think storing the u128 as a `BLOB` in SQLite will make for easier type conversions than using two i64 columns & will still allow us to do sorting & range selects on the `BLOB` values. I don't think there's any issue yet for the needed storage changes but the needed changes shouldn't be too complicated.

conorsch:
@erwanor Can you take another look and just comment on current status? I suspect this work will need to be tracked on the next testnet, given @aubrika's comment above, but I defer to you on that.

redshiftzero:
One idea mentioned in sprint planning was to add another token to genesis with a high level of precision similar to eth (e18) to see what breaks. This could be done with a single note.

erwanor:
+ We'll also need to change the signature to `Amount::value -> u128` and integrate that with the rest of the codebase

hdevalence:
Sounds good, I think we'll generally want to avoid converting to u64s, and migrating our existing uses to use Amounts

hdevalence:
This has been done for a while.

PR #1571, opened 2022-10-31
feat(crypto): make `Amount` use 128 bits internally

erwanor:
Partial implementation of #1378, this PR:

- make `Amount` use a `u128` internally
- provide a getter `Amount::value` returning a `u64`
- adds the necessary conversion traits
- implements encoding a `u128` into two `u64`s, and vice-versa


crypto/src/asset/amount.rs
```diff
@@ -1,26 +1,120 @@
...
+            for i in 0..16 {
+                if i < 8 {
+                    lo_bytes[i] = bytes[i];
+                } else {
+                    hi_bytes[i % 8] = bytes[i];
+                }
+            }
+
+            let lo = u64::from_le_bytes(lo_bytes);
+            let hi = u64::from_be_bytes(hi_bytes);
```
avahowell, line 41: why is `hi` big endian and `lo` little endian?

erwanor: Mostly aesthetic: this lets us hide the byte shifting in the decoding step:

```rust
        let inner = u128::from_be(hi) + lo;
```

instead of:

```rust
        let inner = (u128::from_le(hi) << 64) + lo;
```

but I can change it to the latter, no strong opinion on this

erwanor:
Even though `Amount` uses `u128` internally, the `Amount::value` method casts it down to `u64` unconditionally. This is so the PR can be used right away without propagating changes through out the component stack that we're reworking. However, this also means that we do not actually support bigger `Amount`s yet.

crypto/src/asset/amount.rs
avahowell: this makes sense to me just wanted to check

avahowell:
this LGTM, I think using all little endian and shifting when decoding is clearer

hdevalence:
Let's hold off on merging this until we land the in-flight component changes? I think it will be easier and cleaner to rebase this on top of that than the other way around.

crypto/src/asset/amount.rs
```diff
@@ -1,26 +1,54 @@
...
+                    lo_bytes[i] = bytes[i];
+                } else {
+                    hi_bytes[i % 8] = bytes[i];
+                }
+            }
+
+            let lo = u64::from_le_bytes(lo_bytes);
+            let hi = u64::from_le_bytes(hi_bytes);
+            pb::Amount { lo, hi }
+        }
```
hdevalence, line 43: This seems a bit complicated. I think it would be simpler to do something like


let lo = value as u64;
let hi = (value >> 64) as u64;

to get the high and low 64 bit parts directly.

crypto/src/asset/amount.rs
```diff
@@ -29,7 +57,77 @@ impl TryFrom<pb::Amount> for Amount {
     type Error = anyhow::Error;
 
     fn try_from(amount: pb::Amount) -> Result<Self, Self::Error> {
-        Ok(Amount { inner: amount.lo })
+        let lo = amount.lo as u128;
+        let hi = amount.hi as u128;
+        // `hi` and `lo` represent the high/low order bytes respectively.
+        //
+        // We want to encode `hi` and `lo` into a single `u128` of the form:
```
hdevalence, line 64: This is decoding rather than encoding, right?

crypto/src/asset/amount.rs
erwanor: oh, this is nice!

redshiftzero:
From discussion in a meeting, after this is merged the changes will need to be propagated through the staking rate calculations. This can be rebased and merged

erwanor:
Switched to #1819

PR #1580, opened 2022-11-03
Make TCT internally clone-on-write for cheap in-memory checkpoints

plaidfinch:
This is an exploratory PR which, although it is essentially complete, was an attempt to assess "does this work" in a time-boxed couple of hours, to prepare for writing up a design for the new quarantining system. We could merge it now, but I'm marking as draft for now because it could use some more testing, perhaps, and maybe even benchmarking against previous performance.

## Why do we want this (maybe)?

Cheap cloning for the TCT means we can perform any sequence of actions on a clone of the tree at only a slight slowdown as internal nodes are copied on demand. In particular, this would set the stage for us to be able to "rewind" the in-memory TCT to any previous position, at _constant cost_ regardless of the difference between the current position and the rewound position, and regardless of how many operations have occurred in between those two moments.

Why do we potentially want a method like the following?

```rust
impl Tree {
    pub fn rewind_to(&self, position: Position) -> Option<Self> { /* ... */ }
}
```

(Note that this would only work if the `position` given is witnessed.)

Well, this would allow us to prove inclusion of any note commitment relative to *any past root*, in constant time. It does not require reconstructing the entire tree by incrementally re-de-serializing up to a given position from disk, which costs in the size of the tree up to that position.

We would be able to use this to implement the new quarantining mechanism, so that to prove you're allowed to redeem a "quarantine token", you need to prove you had it _as of the most recent epoch that unbonded_.

## What's in the ~~`Box`~~`Arc`?

This includes several changes that play together:

1. The `Box` in the children of complete nodes is now an `Arc`, whose contents are cloned on modification.
2. There is a new point of indirection introduced in frontier nodes, where the focus of a frontier node is now an `Arc`, whose contents are cloned on modification.
3. The index for the tree is now an `im::HashMap`, which provides the same cheap cloning as the above changes, so that the whole structure can be quickly cloned.
4. The entirety of `Tree`, `epoch::Builder`, and `block::Builder`, are wrapped in an `Arc` each, which means that they can be cloned essentially for free, with deeper cloning happening on demand.

## Happy coincidences

- As a side effect of `Arc`-wrapping things, this means the viral `#![recursion_limit = "256]` bound that needed to be placed in every dependent crate is now gone.
- Making the TCT clone-on-write means we can speculatively modify it cheaply, which means that in `pd` we can get transactional modification of the NCT in the shielded pool without expensive cloning (cc: @hdevalence).

## Drawbacks

The data structure will be somewhat less cache-friendly due to the indirection, and use somewhat more memory, due to the use of `Arc` over `Box` and the introduction of several new `Arc`s. I don't think these are very big concerns, given that it's a very sparse structure in all use cases, and the bottleneck is always going to be hashing and (de)serialization, not cache/memory performance.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -560,6 +560,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
 [[package]]
+name = "bitmaps"
+version = "2.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "031043d04099746d8db04daf1fa424b2bc8bd69d92b25962dcde24da39ab64a2"
+dependencies = [
+ "typenum",
+]
+
+[[package]]
 name = "bitvec"
 version = "0.22.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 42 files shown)


plaidfinch:
Superseded by #1599.

PR #1598, opened 2022-11-09
feat(snapshot_cache): implement `SnapshotCache` 

### LINKED ISSUE #1590: Add a snapshot cache to the `Storage` to allow querying specific state versions.

Currently, the `Storage` structure only allows getting the *most recent* state snapshot.  This means that we cannot properly support queries against specific state versions -- even if they are recent and could still be in-memory -- because we have no way to obtain a snapshot for that height.

This is a problem for implementing some IBC-relayer-related queries, which would like to specify the exact height they're querying about, in order to not have inconsistent data.  (For instance, imagine a relayer querying the state in order to build packets to relay, but in between different queries while building up their request, a new block arrives, and some queries are made with respect to the new state instead of the old one).

We can't reach arbitrarily deep into the past, because of our decision to maintain an index of JMT keys, and that index only being accessible through rocksdb snapshots (which are in-memory only).  But we could maintain a cache of recent snapshots, and allow users of a `Storage` instance to (fallibly) request snapshots of specific versions.

At a high level, we want to replace the `storage::Inner`'s `latest_snapshot: RwLock<Snapshot>` with a `snapshot_cache: RwLock<SnapshotCache>`, for a new `SnapshotCache` type we'll define in a new `::snapshot_cache` module in the crate root.

The `SnapshotCache` should support the following operations:
```rust
impl SnapshotCache {
    pub fn new(initial: Snapshot, max_size: usize) -> Self { ... }
    pub fn push(&mut self, snapshot: Snapshot) { ... }
    pub fn latest(&self) -> Snapshot { ... }
    pub fn get(&self, version: jmt::Version) -> Option<Snapshot> { ... }
}
```

Internally, this could be backed by a `VecDeque`. The `push` method would add a new snapshot with `push_front`, check if the new size was beyond `max_size`, and drop the oldest one if so.  The only slightly tricky part is implementing `get`, which should look at the version of the latest element to compute the correct offset into the snapshot list.

After replacing the current `RwLock<Snapshot>` with an `RwLock<SnapshotCache>` (the number `10` seems like a reasonable hardcoded value of `max_size` to use when constructing it), we should:

- rename `Storage::state()` to `Storage::latest_state()` (using rust-analyzer)
- add a new `Storage::state(&self, version: jmt::Version) -> Option<State>` that requests a state snapshot by version.

erwanor:
Fixes #1590 

- [ ] implement a `SnapshotCache`
- [ ] integrate with `penumbra_storage2`
- [ ] tests / docs / fmt


zbuc:
LGTM 👍

zbuc (approved):

hdevalence (changes requested):
It looks like there are test failures because the `state` -> `latest_state` rename wasn't performed consistently. I checked the commit history but it looks like the renaming happened as part of the implementation work, rather than as a separate step, so it's hard to track where the renaming happened.  The reason I suggested doing the renaming as a separate step is so that we can keep track of it as a separate group of changes.  I'm not sure what the least painful way to manage it at this stage is.  Maybe we can start with `main`, use rust-analyzer to do the renaming as an atomic commit, and then rebase these changes onto that?

storage/src/snapshot_cache.rs
```diff
@@ -0,0 +1,268 @@
+use crate::Snapshot;
+use std::collections::VecDeque;
+
+use anyhow::anyhow;
+
+/// The number of snapshots kept in the [`SnapshotCache`].
+pub const DEFAULT_SNAPSHOT_CACHE_SIZE: usize = 10;
```
hdevalence, line 7: Since the actual size is set in the constructor, it's better not to have a constant here, because having the constant as part of the public API suggests that it does something, when in fact it's not used at all internally, it's just a value that an API consumer *could* put into the constructor.

storage/src/snapshot_cache.rs
```diff
@@ -0,0 +1,268 @@
...
+    /// Creates a [`SnapshotCache`] with `max_size` capacity, and inserts an initial `Snapshot` in
+    /// it.
+    ///
+    /// ## Panics
+    ///
+    /// This function panics if the specified capacity is zero.
+    pub fn new(initial: Snapshot, max_size: usize) -> Self {
+        if max_size == 0 {
+            panic!("snapshot_cache: capacity must be greater than 0.");
+        }
```
hdevalence, line 43: Why panic here? Do we really want to kill the program over that? Why not do
```suggestion
    pub fn new(initial: Snapshot, max_size: usize) -> Self {
        let max_size = std::cmp::max(max_size, 1);
```
instead?

storage/src/snapshot_cache.rs
```diff
@@ -0,0 +1,268 @@
...
+    /// ## Panics
+    ///
+    /// This function panics if the supplied snapshot's version is:
+    ///
+    /// - stale i.e. older than the most recently inserted snapshot
+    ///
+    /// - skipping a version i.e. the difference between the version numbers is greater than 1.
+    ///
+    /// For a non-faillible version, see [`try_push`].
+    pub fn push(&mut self, snapshot: Snapshot) {
```
hdevalence, line 102: Since this is just wrapping `try_push` with an `expect`, I think we could delete it and just have the caller use `expect` if they don't want to deal with an error.

storage/src/snapshot_cache.rs
```diff
@@ -0,0 +1,268 @@
...
+        self.try_push(snapshot)
+            .expect("failed to push snapshot in cache");
```
hdevalence, line 104: It's a moot point in this case, but ideally the `expect` message would specify the invariant that was not satisfied, rather than just saying that a failure occurred.

storage/src/snapshot_cache.rs
```diff
@@ -0,0 +1,268 @@
...
+    }
+
+    /// Attempts to fetch a [`Snapshot`] with a matching `jmt::Version`, and returns `None` if none
+    /// was found.
+    pub fn get(&self, version: jmt::Version) -> Option<Snapshot> {
+        let latest_version = self.latest().version();
+
+        // Special case: the first ever snapshot inserted is allowed to have a "pre-genesis"
+        // `jmt::Version` version (`u64::MAX`). In this case, it is guaranteed to always be
+        // at the back of the queue.
```
hdevalence, line 127: Do we need to special-case this?  Could we instead just compute the difference between the two versions with wrapping arithmetic, and use that as the offset?

storage/src/snapshot_cache.rs
```diff
@@ -0,0 +1,268 @@
...
+    fn create_storage_instance() -> Storage {
+        use tempfile::tempdir;
+        use tokio::runtime::Runtime;
+        let rt = Runtime::new().unwrap();
+        // create a storage backend for testing
+        let dir = tempdir().unwrap();
+        let file_path = dir.path().join("snapshot-cache-testing.db");
+
+        rt.block_on(Storage::load(file_path)).unwrap()
+    }
```
hdevalence, line 164: The `SnapshotCache` API is synchronous so it doesn't matter in this case, but rather than manually constructing a runtime every time, we could use the `#[tokio::test]` to write the tests as `async fn`.

storage/src/snapshot_cache.rs
hdevalence: Then we could also get rid of the `oldest` method entirely, too.

hdevalence:
Maybe we could rebase this on top of dbd5324e in the `renamed-state-method` branch?

storage/src/snapshot_cache.rs
erwanor: Good point, it's not a special case if we remove the `if version > latest_version` return early logic, cleaner indeed!

storage/src/snapshot_cache.rs
erwanor: Thanks, this makes sense

storage/src/storage.rs
```diff
@@ -67,13 +68,16 @@ impl Storage {
                         .unwrap_or(u64::MAX);
 
                     let latest_snapshot = RwLock::new(Snapshot::new(db.clone(), jmt_version));
+                    let latest_snapshot_tmp = Snapshot::new(db.clone(), jmt_version);
 
                     // We discard the receiver here, because we'll construct new ones in subscribe()
                     let (snapshot_tx, _) =
                         watch::channel(StateNotification(latest_snapshot.read().clone()));
 
+                    let snapshots = RwLock::new(SnapshotCache::new(latest_snapshot_tmp, 10));
```
hdevalence, line 77: Wait, why are we constructing an `RwLock<Snapshot>` here? Shouldn't this be something like
```rust
let snapshots = RwLock::new(SnapshotCache::new(
    Snapshot::new(db.clone(), jmt_version),
    10
    ));

let (snapshot_tx, _) = watch::channel(StateNotification(snapshots.read().latest()));
```
since we're not using an `RwLock`'d snapshot any more?

hdevalence (approved):
LGTM modulo comments!

Issue #1591, opened 2022-11-08
Clean up handling of Merkle proof queries


hdevalence:
**Is your feature request related to a problem? Please describe.**

Currently, our Merkle proof queries are a mess. We currently have at least two different ways of returning Merkle proofs, with two different root hashes. The situation is as follows.  

First, ideally our app hash would be the root hash of the JMT.  Unfortunately this is not possible, because our IBC counterparty chains, who use the Cosmos SDK, require that IBC proofs have two nontrivial ICS23 proofs, an artefact of the SDK's use of an IBC-specific IAVL subtree (specifically, ibc-go requires that the tree has a [stable prefix](https://github.com/cosmos/ibc-go/blob/c9b806485e1d28f5fed46f323cf6236ccfbd45bd/modules/core/03-connection/keeper/verify.go#L39), something that isn't achievable in a sparsed prehashed tree like the JMT).  To achieve interoperability, we introduce a stable prefix by making our app hash be a hash of the JMT root hash, so we can treat it as a depth-1 Merkle tree, and supply a chain of two proofs.  There are thus two different Merkle roots at play: the "real" JMT root, and the "workaround" apphash.  This was introduced in https://github.com/penumbra-zone/penumbra/pull/1189

Second, there are two kinds of Merkle proofs in use, Tendermint Merkle proofs and ICS23 Merkle proofs. These formats are similar but not identical.  ABCI queries from Tendermint need to be responded to with Tendermint Merkle proofs.  This way, a relayer which queries Tendermint's RPC can have their request forwarded to the ABCI app, and get a Tendermint Merkle proof back, which they then immediately convert to an ICS23 Merkle proof (🫡).

**Describe the solution you'd like**

As in #1586 #1588, ideally the `storage2` crate is general-purpose, and independent of Penumbra.  So the `get_with_proof` method in `storage2::State` should return an ICS23 proof up to the JMT root.

We can handle the Penumbra-specific interop workaround with a new extension trait in the `chain` crate, providing methods like `get_with_proof_to_apphash` or `get_with_proof_to_apphash_tendermint` returning ICS23 or Tendermint proofs. In the case of ICS23, it should return an `ibc::core::ics23_commitment::merkle::MerkleProof`. In the case of tendermint, it should return a `tendermint::merkle::proof::Proof`.

conorsch:
Confirmed during weekly wrap-up today that this work is indeed complete.

Issue #1588, opened 2022-11-08
Move `AppHash` from `storage2` to `chain`


hdevalence:
**Is your feature request related to a problem? Please describe.**

Currently, the `storage2` crate contains the implementation of our custom `AppHash` type.  This type exists to do extra hashing to a `jmt::RootHash`, so that when preparing ICS23 merkle proofs to send to Cosmos SDK counterparties, we can supply two non-trivial proofs, which is an undocumented requirement of that implementation.  (In other words, we define Penumbra's merkle tree to be a tier of two merkle trees; a dummy depth-1 merkle tree with a single leaf node, and then the real JMT structure).

However, this `AppHash` type is Penumbra-specific, and so it should ideally not live in the (ideally) general-purpose `storage2` crate.

**Describe the solution you'd like**

Move the `AppHash` type from `storage2` to `chain`.

PR #1605, opened 2022-11-11
clean up merkle proof API: move apphash calculation to penumbra_chain

avahowell:
#1591 #1588


hdevalence:
Rebased onto current `main` to try rerunning smoke tests with NCT divergence flag enabled.

PR #1609, opened 2022-11-14
refactor: create `penumbra_proto::{StateReadProto, StateWriteProto}`

### LINKED ISSUE #1586: Invert dependency relationship between `storage2` and `proto` crates

Currently, the `storage2` crate depends on the `proto` crate, so that it can offer proto/domain-typed methods on top of the stored data: after implementing, e.g., `StateRead::get_raw`, implementations of `StateRead::get_proto<P>` and `StateRead::get<D, P>` become available automatically.

This is a very convenient design, and we should keep it.  Unfortunately, however, it makes the `storage2` crate less generally useful than it could be, because it forces a Penumbra-specific dependency on what would otherwise be a general-purpose crate.

Invert the dependency relationship between `storage2` and `proto` crates:

- [ ] Remove the dependency of `storage2` on the `proto` crate
- [ ] Make the `proto` crate depend on `storage2`, as an optional dependency
- [ ] Remove the proto and domain type methods from the `StateRead` and `StateWrite` traits in the `storage2` crate
- [ ] Add new `StateReadProto` and `StateWriteProto` extension traits in the `proto` crate, in their own submodule that's `cfg`-gated on the optional dependency, and re-export them from the crate root.
- [ ] Update users of `StateRead` and `StateWrite` to also use `StateReadProto` and `StateWriteProto`.

Making the extension traits gated on an optional feature means that users of the `proto` crate who don't care about `storage2` integration don't transitively depend on all of `rocksdb`.

**Additional context**

This probably makes sense to do after renaming `storage2` to `storage` and deleting the old implementation, just so we don't have to propagate the renaming even further.

erwanor:
Fixes #1586 .

Invert the dependency relationship between `storage` and `proto` crates:

- [x]  Remove the dependency of `storage` on the proto crate
- [x]  Make the proto crate depend on `storage`, as an optional dependency
- [x]  Remove the proto and domain type methods from the StateRead and StateWrite traits in the storage2 crate
- [x]  Add new `StateReadProto` and `StateWriteProto` extension traits in the proto crate, in their own submodule that's cfg-gated on the optional dependency, and re-export them from the crate root.
- [x]  Update users of StateRead and StateWrite to also use `StateReadProto` and `StateWriteProto`.


proto/src/state.rs
```diff
@@ -0,0 +1,5 @@
+#![cfg(feature = "penumbra-storage")]
```
zbuc, line 1: Will this only gate the `read` mod behind the feature flag?

erwanor: I _think_ the shebang should make it crate wide but let me double check that

chain/src/view.rs
```diff
@@ -17,7 +18,7 @@ use crate::{
 /// initialized, so they will error on an empty state.
 //#[async_trait(?Send)]
 #[async_trait]
-pub trait StateReadExt: StateRead {
+pub trait StateReadExt: StateReadProto {
```
hdevalence, line 21: Do all these bounds have to change? Since `StateReadProto` has a blanket implementation (or should), it should be present for any `StateRead`.

proto/src/state.rs
zbuc: Yeah I think you're right: https://doc.rust-lang.org/reference/attributes.html

chain/src/view.rs
erwanor: good catch thanks, yeah they have blanket impls!

proto/src/state/write.rs
```diff
@@ -0,0 +1,29 @@
+use crate::{Message, Protobuf};
+
+use std::fmt::Debug;
+
+use penumbra_storage::StateWrite;
+
+pub trait StateWriteProto: StateWrite + Send + Sync {
```
erwanor, line 7: @hdevalence maybe we would rather want to be consistent with the relationship between `StateRead` and `StateWrite`

```rust
pub trait StateWriteProto: StateWrite + StateReadProto + Send + Sync
```

thoughts?

hdevalence: I don't think we need to put this bound in the trait definition, since it's covered by the blanket impls?

chain/src/view.rs
```diff
@@ -118,7 +119,7 @@ pub trait StateReadExt: StateRead {
     }
 }
 
-impl<T: StateRead + ?Sized> StateReadExt for T {}
+impl<T: StateRead + StateReadProto + ?Sized> StateReadExt for T {}
```
hdevalence, line 122: I don't think this change should be necessary, the `StateRead + ?Sized` bound should be enough.

proto/src/state.rs
hdevalence: Since there are `cfg` directives in the outer module, however, I think it would be cleaner to move this to `lib.rs` and have it attach to `mod state;` there, so that all of the `cfg` directives are in one place.

hdevalence (approved):
LGTM modulo comments

PR #1613, opened 2022-11-14
Rename storage trait methods

### LINKED ISSUE #1585: Rename `storage2` trait methods

Some of the trait methods in the `storage2` interfaces don't have the optimal names, and it'd be better to rename them.  This is a mechanical change, but we should wait to do this until we don't have any large, in-flight work (like #1583).

Rename:

- [ ] `StateRead::get_nonconsensus` -> `StateRead::nonconsensus_get_raw`
- [ ] `StateRead::get_ephemeral` -> `StateRead::object_get`
- [ ] `StateWrite::put_nonconsensus` -> `StateWrite::nonconsensus_put_raw`
- [ ] `StateWrite::delete_nonconsensus` -> `StateWrite::nonconsensus_delete`
- [ ] `StateWrite::put_ephemeral` -> `StateWrite::object_put`
- [ ] `StateWrite::delete_ephemeral` -> `StateWrite::object_delete`

The motivation for "ephemeral" vs "object" is that "(in-memory) object store" is a better name; "ephemeral" causes confusion because, ephemeral on what time span?

The motivation for the category prefixes is that it ensures that different categories of methods are grouped together when sorted alphabetically, provididng a more consistent API. It also allows us to extend the nonconsensus k/v store with typed wrapper methods in the future without having to make more API changes.

zbuc:
Closes #1585

### DIFF
--- a/component/src/dex/component.rs
+++ b/component/src/dex/component.rs
@@ -274,7 +274,7 @@ pub trait StateReadExt: StateRead {
     }
 
     fn swap_flows(&self) -> BTreeMap<TradingPair, SwapFlow> {
-        self.get_ephemeral::<BTreeMap<TradingPair, SwapFlow>>(state_key::swap_flows())
+        self.object_get::<BTreeMap<TradingPair, SwapFlow>>(state_key::swap_flows())
             .cloned()
             .unwrap_or_default()
     }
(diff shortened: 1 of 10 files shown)


hdevalence (approved):
LGTM!

PR #1617, opened 2022-11-15
fix testnet update.md

SGTstake:


erwanor:
good catch, thanks @SGTstake

PR #1618, opened 2022-11-15
ActionHandler impls

### LINKED ISSUE #1592: Move action handling logic out of the `Component` trait into a new `ActionHandler` trait

As a side effect of our initial component design, we ended up with a "breadth-first" processing of transaction actions, where each `Component` has methods related to transaction checking and execution, and processing a transaction involves passing the transaction through each `Component`'s checks and execution methods.

This is error-prone and difficult to understand, because it's not clear what steps are actually involved in processing any specific action, or whether they're duplicated between components, etc.

While rewriting the `Component` implementations as part of #1558, we discussed a better alternative, which would be to conceptually separate the *chain-wide* execution (currently and to remain in the `Component` trait) from the *per-transaction* execution (which moves to a new `ActionHandler` trait).

This is now possible to do, because as a result of #1558, there is no such thing as "instantiating" a `Component`, as a `Component` has no state anywhere other than in the `State` itself, and all of the meat of the implementation is defined in extension traits that add behavior to a `StateRead` or `StateWrite`.  So, for instance, the shielded pool's `NoteManager` trait adds in `mint_note`, `add_note`, etc methods onto a `StateWrite` implementation.

- [x] Create a new module subtree in the `component` crate as follows:
  - `component/src/action_handler.rs`: will have trait definition
  - `component/src/action_handler/transaction.rs`: will have top-level impl for `Transaction`
  - `component/src/action_handler/actions.rs`: forwarding impl for `Action`
  - `component/src/action_handler/actions/spend.rs`: impl for `Spend`
  - `component/src/action_handler/actions/output.rs`: impl for `Output`
  - `component/src/action_handler/actions/...`: ... one file for each action type
- [x] Move the (commented out) stub definition of the `ActionHandler` trait out of `component/src/lib.rs` into the `action_handler` module
- [x] Add an empty impl of the `ActionHandler` trait for each action with a domain type in the appropriate `actions/foo.rs` submodule
  - Some actions don't have domain types to implement the trait on, so skip these
- [x] Add an `impl ActionHandler for Action` in `actions.rs` using a match statement to forward to the inner action type
  - Actions without a domain type are handled in this impl, initially with an empty body
- [x] Add an `impl ActionHandler for Transaction` in `transaction.rs` that iterates over the actions and calls the handler method on each one.

At this point, there's a full scaffold of empty `ActionHandler` impls to slot code into.

- [x] Remove the `check_tx_stateless`, `check_tx_stateful`, and `execute_tx` methods from the `Component` trait.  Now the `Component` impls will all error because they have methods that aren't defined in the trait, highlighting all the logic that has to move.  Go through the erroring methods, and move the per-action logic to the `ActionHandler` for that action.  The component's extension traits can be brought into scope to access any required behavior on the state.

In some places (e.g., parts of the shielded pool), we've moved towards an idiom for highlighting explicit consensus rules by factoring them out into descriptively named functions, e.g. `consensus_rules::stateless::num_clues_equal_to_num_outputs`.  This is a good idea, and if possible, it should be extended during refactoring. Making entirely separate files is probably overkill, but, e.g., the `spend` module could have
```rust
impl ActionHandler for Spend {
    // ...
}

mod consensus_rules {
    use super::*;

    mod stateless {
        fn foo(...) { ... }
    }
}
```

This way, we'll be able to obtain a list of all per-transaction consensus rules by running `rg consensus_rules::`, or some other query.

**Additional context**

Once this is done, we'll be in a much better position to clean up the component implementations, because we'll be able to see exactly what the "external" interfaces to the component logic is -- it's all the methods that are used in the `ActionHandler` implementations.

zbuc:
Closes #1592


component/src/action_handler/actions/ics20/withdrawal.rs
```diff
@@ -3,21 +3,31 @@ use std::sync::Arc;
...
     }
 
-    async fn check_tx_stateful(&self, _state: Arc<State>) -> Result<()> {
-        todo!()
+    #[instrument(name = "ics20_withdrawal", skip(self, state, _context))]
+    async fn check_stateful(&self, state: Arc<State>, _context: Arc<Transaction>) -> Result<()> {
+        <penumbra_storage::State as crate::ibc::transfer::Ics20TransferReadExt>::withdrawal_check(
+            state.clone(),
+            self,
+        )
```
hdevalence, line 23: Hmm, this is a little weird, why doesn't `state.withdrawal_check(self)` work here?

hdevalence:
@avahowell do you have any insight into the test failure here?

zbuc:
> @avahowell do you have any insight into the test failure here?

@hdevalence it's because we're now running _all_ the validations rather than just the ones relating to IBC. Let me see if I can call into the IBC validations directly

component/src/action_handler/actions/ics20/withdrawal.rs
zbuc: Ah, the implementation of `withdrawal_check` doesn't take a `self` parameter currently (due to copy-pasta) and rust-analyzer suggested this and I went with it. Should be easy to fix.

hdevalence:
Since we're ideally not going to keep quarantining around, I think it's fine to use whatever organization now, and then realign it after we remove quarantining.

Issue #1210, opened 2022-07-22
Add validator definition with duplicate consensus key to smoke test


hdevalence:
**Describe the solution you'd like**

We should check that uploading a validator definition with a duplicate consensus key fails.  To do this, we can grab the consensus key of an existing validator, insert it into a default validator definition, and upload it.

**Additional context**
#1207

hdevalence:
We should also try *updating* a validator definition to include someone else's consensus key

PR #1631, opened 2022-11-17
test: uploading duplicate consensus key fails

conorsch:
Towards #1210. Adds the first of two tests required by that ticket. Currently passing smoke tests, so considering ready for review. I'll follow up separately with another test, at which point we can consider 1210 complete.

### DIFF
--- a/pcli/tests/network_integration.rs
+++ b/pcli/tests/network_integration.rs
@@ -14,7 +14,8 @@ use std::{thread, time};
 use assert_cmd::Command;
 use predicates::prelude::*;
 use regex::Regex;
-use tempfile::{tempdir, TempDir};
+use serde_json::Value;
+use tempfile::{tempdir, NamedTempFile, TempDir};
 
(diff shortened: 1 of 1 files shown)


pcli/src/main.rs
```diff
@@ -58,7 +58,7 @@ impl App {
             .transpose()?
             .ok_or_else(|| anyhow::anyhow!("view service did not report sync status"))?;
 
-        println!(
+        eprintln!(
```
hdevalence, line 61: All our other output goes to stdout, do we want this one specifically to go to stderr?

conorsch: Fair point, not strictly necessary here. I'll remove that change from this PR, and open a separate issue to discuss the stderr/stdout inconsistency.

hdevalence: No worries, I wasn't concerned about it not being necessary here, just wanting to make sure if we do that change, we do it completely and not just here.

PR #1643, opened 2022-11-20
test: validator malformed definition update fails

### LINKED ISSUE #1210: Add validator definition with duplicate consensus key to smoke test

conorsch:
Builds on #1631 and closes #1210. The new test confirms that trying to update a validator with a mismatched consensus key fails. Ideally we'd create a new validator identity from scratch in this test, but instead I'm reusing the validator created in the previous test. This is suitable for how we execute the smoketest suite, but does assume strict ordering of the tests. At some point in the future, we may wish to generate multiple wallets (and therefore multiple validator identities), but for now, this is Progress™.

### DIFF
--- a/pcli/tests/network_integration.rs
+++ b/pcli/tests/network_integration.rs
@@ -9,9 +9,11 @@
 //! where no tokens have been delegated, and the address with index 0
 //! was distributed 1cube.
 
+use std::path::PathBuf;
 use std::{thread, time};
 
 use assert_cmd::Command;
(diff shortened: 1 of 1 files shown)


conorsch:
Rebased on #1649 to keep CI passing; will update and mark RFR once that's in.

PR #1653, opened 2022-11-22
docs: no 'cargo update' during testnet upgrades

conorsch:
Running `cargo update` as currently recommended in the docs will cause local dependencies to get out of sync with tagged testnet version. We commit the `Cargo.lock` file in order to version lock dependencies; `cargo update` breaks that strict version locking. Worse, `cargo update` will result in a dirty git working tree, requiring manual cleanup (or `git stash`) on subsequent testnet upgrades. We don't document that manual cleanup, nor should we: we should expect that all nodes run exactly the versions of dependencies that we specify in Cargo.lock.

### DIFF
--- a/docs/guide/src/pcli/install.md
+++ b/docs/guide/src/pcli/install.md
@@ -56,7 +56,7 @@ latest tag for the current
 [testnet](https://github.com/penumbra-zone/penumbra/releases):
 
 ```bash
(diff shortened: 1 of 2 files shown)


hdevalence:
This is a good change. I'd prefer to remove the `Cargo.lock` entirely; I believe strict version locking isn't actually desirable.

PR #1654, opened 2022-11-22
Clean up IBC connection tracking

### LINKED ISSUE #1636: Port IBC component's Client Connections to use range queries
Part of #1593

zbuc:
Closes #1636 

@avahowell can you please comment on if we need the association of connections w/ client IDs at all? It looks like they weren't actually being used, but I'm not sure if it's needed in the future.


conorsch:
@zbuc This needs a rebase, to resolve conflicts, before it's reviewable.

zbuc:
@conorsch rebased!

hdevalence (approved):

avahowell:
The reverse lookup (client id -> connections) is a convenience function for relayers. It's called for in the ICS-3 spec https://github.com/cosmos/ibc/tree/main/spec/core/ics-003-connection-semantics. Presumably this is so a relayer can quickly fetch all of the connections associated with a client ID and track/update their state.

component/src/ibc/component/state_key.rs
```diff
@@ -29,16 +29,32 @@ pub fn client_processed_times(client_id: &ClientId, height: &Height) -> String {
...
 
-pub fn connection(connection_id: &ConnectionId) -> String {
-    format!("connections/{}", connection_id.as_str())
-}
+    use std::string::String;
+
+    // TODO: is this going to be needed in the future? When refactoring,
+    // I noticed the previous `client_connections` was not used anywhere.
+    pub fn by_client_id_list(client_id: &ClientId) -> String {
+        format!("clients/{}/connections/", client_id)
```
avahowell, line 40: this should be `clients/{}/connections` instead (no trailing /)

zbuc: @avahowell we are using the trailing `/` to denote lists accessed by prefix now, to prevent unexpected collision if say there was another hypothetical key like `clients/{}/connections_historical/{}` added

zbuc: that is, when we access this key we would use a prefix query that would return every key/value pair beginning with the key as a prefix.

For example from the staking component:

```rust

    async fn validator_list(&self) -> Result<Vec<Validator>> {
        self.prefix(state_key::validators::list())
            .next()
            .await
            .into_iter()
            // The prefix stream returns keys and values, but we only want the values.
            .map(|r| Ok(r?.1))
            .collect()
    }
```

And the associated keys:

```rust
pub mod validators {
    use penumbra_crypto::IdentityKey;
    pub fn list() -> &'static str {
        "staking/validator/"
    }

    pub fn by_id(id: &IdentityKey) -> String {
        format!("staking/validator/{}", id)
    }
}
```

PR #1655, opened 2022-11-22
testnet: new release 036-iocaste.2

erwanor:


conorsch:
We had some prompt reports in the Discord `#testnet-feedback` channel about the misconfiguration. Thanks to all who reported it! Quick fix, @erwanor, approving. Will let you handle the follow-up tag.

conorsch:
> approving

Can't actually do that until CI is done with it. I don't expect any CI hiccups, but waiting ~20m to be sure.

erwanor:
I'm taking a second to make sure this actually fixes the testnet.

PR #1657, opened 2022-11-22
toolchain: specify rust@1.65.0

erwanor:


hdevalence:
I'm not sure we want to pin Rust 1.65 specifically -- when are we going to update this? who will be responsible for doing that? how do we weigh missing out on future performance improvements while we don't update?

conorsch:
Some context: on the most recent testnet, we added a "while let" line of code that was flagged as nightly-only on older versions of Rust. This caused some folks running full nodes and validators to encounter a build error. We haven't been careful about instructing people to keep Rust updated, but we totally should. 

Longer term, once we have prebuilt container images that people can easily run, the ergonomics of staying in sync are a bit easier. But we'll always need to be clean about build requirements for those who want to rebuild from source. 

So it sounds to me like the solution here is "clarify in the guide that one should run `rustup update`" which should cover our bases. However, it'd be nice to say `>=1.56` so we can provide more contextualized errors. I'm not sure whether core cargo tooling permits that, though.

hdevalence:
It's possible to set an MSRV with cargo, and I think we should do that: https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field

I think this will get us the behavior we want: any older versions of rust will cause an immediate, more intuitive error, but unlike the toolchain approach, we're not pinning to a specific version that will become obsolete without active maintenance.

hdevalence:
From looking at the docs on that field, it looks like something we have to add to a specific package (crate), rather than the workspace. If so, I'd suggest that we just add it to `pd`, rather than every crate, so that whenever we want to change it, we only have to edit one place.

PR #1661, opened 2022-11-23
Convert governance voting validators to use range query, introduce ne…

zbuc:
…w `prefix_domain` for specific query API


hdevalence (approved):
LGTM!

PR #1662, opened 2022-11-23
docs: add talks & presentations to resources page

conorsch:
Simple change, trying to collect a few useful talks for folks who want to understand the platform better. There's an argument that this content should go on https://penumbra.zone, rather than https://guide.penumbra.zone, but we can always cross link later.

Things to check as part of review:

1. Are these talks reasonable ones to call attention to?
2. Are the links and titles accurate?
3. Is the guide a good enough place to link to these talks for now?


conorsch:
Self-merging, so we have a stable URL to share in Discord. Will add more resources over time, as inspiration strikes.

Issue #1581, opened 2022-11-03
Tokenize Quarantining


plaidfinch:
## The situation

Right now, we have a bespoke implementation of quarantining for undelegations which has a number of downsides:

1. It's complex, both on the node side (which would be fine, except...) and also this complexity bleeds over into client interface implementations. Clients need to track a whole separate Kind Of Thing which is an **un-**positioned quarantined `Note`, and correctly model the state change when a note becomes unquarantined. It's easy to get wrong (we have, several times, notably a write ordering causing #1507), and we don't want to impose this burden forever to penumbra client interface users.
2. It requires special guardrails to avoid accidentally making a randomly selected chunk of your notes temporarily illiquid: see #1010 for the solution to that. Every penumbra client would need to make sure they do this, or else risk a bug where they lock up unrelated user funds of unbounded amount until the unbonding period expires.
3. It works differently than the rest of the "zk-async" penumbra state model. We now know how to model things Of This Sort better, as we've done with swap/swap-claim: make a token representing the "future" you want to redeem, then make a second transaction to redeem it. This works for swaps, and it should work for quarantining (as well as governance proposal deposit refunds, to be discussed separately).

## A solution

Instead of our bespoke quarantining system, we should make new assets, here called **Unbonding Tokens** (bikeshed in the comments). Just as every validator is associated with precisely one Delegation Token asset, they would also be associated with a family of unbonding token denominations, parameterized by epoch. For example, the denomination `unbonding_epoch_12_until_15_penumbravalid1ymlgthzee7nzn7nhrz38u4p5a580uzdupu6dzckr56v63v4tguysnpz7fm` represents tokens which *began* unbonding in epoch 12 and will finish unbonding in epoch 15 (in this hypothetical, the `unbonding_epochs` chain parameter is set to 3).

This solution also changes the meaning of the `Undelegate` action, and adds a new `UndelegateClaim` action. Their meanings are as follows:

[...]

PR #1663, opened 2022-11-24
Add blinding factor to Swap plaintext

### LINKED ISSUE #1622: Add a blinding factor to swap NFTs and make swap outputs deterministically generated from it

#1544 flags an issue with the current implementation of the swap/swapclaim mechanism: because the output note plaintexts are generated by the claimer, and not verified as part of the proof, a malicious claimer (in possession of only a full viewing key) can burn output funds by correctly constructing output notes and proving correctness of their note commitments, but not correctly encrypting the notes themselves to the recipient.

Fixing this has two parts. In the first part, this issue, we change the swap/swapclaim mechanism so that the output notes are deterministically derived only from data in the original swap, so that the entity that signed the swap can always derive the output notes.  In the second part, later, we would like to eliminate the deterministically-derived notes from the swapclaim, to save space on chain and work for other clients. 

- [x] Add a swap blinding factor (an `Fq` element) to the swap plaintext
  - [x] in the proto type
  - [x] the domain type
  - [x] in the fixed-length encoding we use to construct the swap ciphertext
- [x] Extend the `Planner` API to choose a random blinding factor when constructing the swap plan, similar to other blinding factors (e.g., the note blinding in an output plan, etc)
- [x] Change the derivation of the swap nft asset id (in `SwapPlaintext::asset_id`) to compute the asset id as `hash_7(DOM_SEP, (swap_blinding, amount, ...))` instead of `hash_6(DOM_SEP, (amount, ...))`.
- [x] Add the swap blinding factor to the `SwapProof`. Currently, there's no asset id integrity check in the `SwapProof` to adjust -- this is actually a security-critical soundness bug. To fix it:
  - [x] Remove the `swap_nft_asset_id` field from the `SwapProof`. Instead, *recompute* what it *should be* inside of the proof verification, by using the data in the `SwapProof` to construct a `SwapPlaintext`, and compute the expected asset ID by calling `SwapPlaintext::asset_id`, rather than using the attacker-controlled asset ID.
- [x] Add the swap blinding factor to the `SwapClaimProof`, and adjust its checks too. Currently, the swap claim proof checks the asset ID using a `swap_nft_asset_id_integrity` gadget, which constructs a `SwapPlaintext` out of the data in the `SwapClaimProof`, recomputes the expected asset ID, and checks that the provided one matches.  Instead, we should also remove the `swap_nft_asset_id` from the `SwapClaimProof`, and rather than checking it against the expected value we recompute, we should just use the recomputed value directly.
- [x] Change the `SwapClaim` to deterministically generate note blinding factors from the original randomness in the swap blinding:
  - [x] Add a `pub SwapPlaintext::output_blinding_factors(&self) -> (Fq, Fq)` method that generates the blinding factors for the two output notes, with something like the following (see also the `SwapPlaintext::asset_id()` implementation): 
      ```rust
    pub static OUTPUT_1_BLINDING_DOMAIN_SEPARATOR: Lazy<Fq> =
        Lazy::new(|| Fq::from_le_bytes_mod_order(blake2b_simd::blake2b(b"penumbra.swapclaim.output1.blinding").as_bytes()));
    pub static OUTPUT_2_BLINDING_DOMAIN_SEPARATOR: Lazy<Fq> =
        Lazy::new(|| Fq::from_le_bytes_mod_order(blake2b_simd::blake2b(b"penumbra.swapclaim.output2.blinding").as_bytes()));

    impl SwapPlaintext {
        pub fn output_blinding_factors(&self) -> (Fq, Fq) {
            (
                hash_1(OUTPUT_1_BLINDING_DOMAIN_SEPARATOR, self.swap_blinding),
                hash_1(OUTPUT_2_BLINDING_DOMAIN_SEPARATOR, self.swap_blinding),
            )
        }
    }
    ```
  - [x] Remove the `output_1_blinding` and `output_2_blinding` fields from the `SwapClaimPlan`.  Instead, when constructing the swap claim in `swap_claim_body()`, use `SwapPlaintext::output_blinding_factors()` to generate them from `self.swap_plaintext`.
  - [x] Remove the `note_blinding_1` and `note_blinding_2` fields from the `SwapClaimProof`. Instead, recompute them inside the proof verification by calling `output_blinding_factors()` on the same `SwapPlaintext` used to recompute the asset ID.

This shouldn't require any changes to the view service, or downstream users of the planner API, because it only changes how the blinding factors are derived (deterministically, rather than internal randomness), and because it only changes the format of the `SwapPlaintext`, which is stored in the view service as a blob.

aubrika:
Closes #1622

### DIFF
--- a/component/src/dex/tests.rs
+++ b/component/src/dex/tests.rs
@@ -1,0 +1,100 @@
+use std::sync::Arc;
+
+use penumbra_chain::{test_keys, StateReadExt, StateWriteExt};
+use penumbra_crypto::{
+    asset,
+    dex::{swap::SwapPlaintext, TradingPair},
+    transaction::Fee,
+    Address, Amount,
+};
+use penumbra_storage::{ArcStateExt, TempStorage};
+use penumbra_transaction::{
+    plan::{SwapClaimPlan, SwapPlan},
+    Transaction,
+};
+use rand_core::SeedableRng;
+use tendermint::abci;
+
+use crate::{shielded_pool::ShieldedPool, ActionHandler, Component, TempStorageExt};
+
+use super::{Dex, StateReadExt as _};
+
+#[tokio::test]
+async fn swap_and_swap_claim() -> anyhow::Result<()> {
+    let mut rng = rand_chacha::ChaChaRng::seed_from_u64(1312);
+
+    let storage = TempStorage::new().await?.apply_default_genesis().await?;
+    let mut state = Arc::new(storage.latest_state());
+
+    let height = 1;
+
+    // 1. Simulate BeginBlock
+
+    let mut state_tx = state.try_begin_transaction().unwrap();
+    state_tx.put_block_height(height);
+    state_tx.apply();
+
+    // 2. Create a Swap action
+
+    let gm = asset::REGISTRY.parse_unit("gm");
+    let gn = asset::REGISTRY.parse_unit("gn");
+    let trading_pair = TradingPair::new(gm.id(), gn.id()).expect("distinct assets");
+
+    let delta_1 = Amount::from(100_000u64);
+    let delta_2 = Amount::from(0u64);
+    let fee = Fee::default();
+    let claim_address: Address = test_keys::ADDRESS_0.clone();
+
+    let plaintext =
+        SwapPlaintext::new(&mut rng, trading_pair, delta_1, delta_2, fee, claim_address);
+
+    let swap_plan = SwapPlan::new(&mut rng, plaintext);
+    let swap = swap_plan.swap(&*test_keys::FULL_VIEWING_KEY);
+
+    // 3. Simulate execution of the Swap action
+
+    // The Swap ActionHandler doesn't use the context, so pass a dummy one.
+    let context = Arc::new(Transaction::default());
+
+    swap.check_stateless(context).await?;
+    swap.check_stateful(state.clone()).await?;
+    let mut state_tx = state.try_begin_transaction().unwrap();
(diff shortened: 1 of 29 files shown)


component/src/dex/tests.rs
```diff
@@ -0,0 +1,143 @@
...
+    // that records the note payload (for any action) is in the transaction-wide
+    // ActionHandler impl.  Why? because with the current quarantine system, we
+    // have to quarantine all outputs from any action if a quarantined output is
+    // present, so we can't do the correct and simple thing of executing each
+    // action in its own action handler. After tokenized unbonding, we should
+    // fix this!
+    //
+    // Because of this, we have to construct a whole dummy tx that we can
+    // .execute().  Otherwise, we won't actually record the swap nft for use in
+    // testing the SwapClaim logic.
```
hdevalence, line 68: cc @plaidfinch  / #1581

component/src/mock_client.rs
```diff
@@ -0,0 +1,97 @@
...
+use crate::shielded_pool::StateReadExt as _;
+
+/// A bare-bones mock client for use exercising the state machine.
+pub struct MockClient {
+    latest_height: u64,
+    epoch_duration: u64,
+    fvk: FullViewingKey,
+    notes: BTreeMap<note::Commitment, Note>,
+    nct: penumbra_tct::Tree,
+}
```
hdevalence, line 16: @penumbra-zone/penumbra-labs we can use this for writing other state machine integration tests, too, like spend/output tests.

hdevalence (approved):

PR #1672, opened 2022-11-28
pd: add extra peers to configs in testnet join

hdevalence:
This commit changes the `testnet join` logic to also scrape the `/net_info` RPC of the Tendermint node it attempts to clone, in order to get a list of connected peers, and optimistically adds all connected peers as seed nodes to the generated configs.

The hope is that this addresses a case where the initial bootstrap node has hit its connection limit and doesn't accept new connections, by allowing new nodes to attempt to connect to other nodes in the network.


conorsch:
Works great. After running `pd testnet join` with these changes, I see 50 peers added to `~/.penumbra/testnet_data/node0/tendermint/config/config.toml`.

PR #1673, opened 2022-11-28
pd: avoid two levels of defaults in ChainParameters

hdevalence:
This commit changes the Clap arguments for `pd testnet generate` to not include a second set of default values for chain parameters beyond those already specified in `ChainParameters::default()`, so we only have one place to look.

Also increase the default validator limit to 80.


conorsch:
Effective limit has been increased 32 -> 80.

Issue #1664, opened 2022-11-25
Add infrastructure for writing unit tests on state machine components


hdevalence:
**Is your feature request related to a problem? Please describe.**

We want to be able to write unit tests that check parts of the state machine in isolation. For instance, we'd like to be able to check individual consensus rules (does this consensus rule reject this input?), or to be able to check pre/post conditions for specific parts of the state machine (do we maintain the desired invariants after processing this input? do we write the data we expect?).

Currently, we don't have any meaningful scaffolding for writing these kinds of tests, which means that we rely solely on end-to-end integration tests.  But while end-to-end integration tests are useful to have, they're more complicated and expensive to write and run, and shouldn't be the only option for writing tests.  **We should aim that the majority of the test suite is runnable by `cargo test`, so that most assurance can be accessed through existing Rust tooling**.  Not only is this a simpler and easier development workflow -- it only requires `cargo`, and not other tooling -- it can also run many tests in parallel, and faster.

As a prototypical example, consider #1631, which added a test to check the consensus rule that validator definitions with duplicate consensus keys are rejected. This is a good test to have, but we don't get any assurance benefit out of having it be a full, end-to-end network integration test that:
- runs `pcli` in a subprocess
- makes network requests to a running `pd` instance
- parses the JSON output of `pcli`
- edits a new validator definition
- assembles it into a full, valid transaction
- sends that transaction over the network to a running `tendermint` instance
- has that transaction sent over ABCI to `pd`
- has the mempool check every other consensus rule on the transaction
- has the mempool check the desired consensus rule (if it didn't already reject it)
- checks that the `pcli` subprocess that sent the transaction errored out

instead of just checking the specific consensus rule we care about directly.  Not only is this process more expensive, if what we care about is checking that duplicate consensus keys are rejected, it's actually **worse** for assurance than a unit test, because we don't know that the reason the transaction was rejected was the one we're trying to test, rather than some other reason.

This isn't meant as a criticism of #1631, to be clear, and doing it as an integration test was my suggestion in #1210; the point is that at the moment, we don't have any other alternative structure to use to write this kind of test, but now that we've completed #1454, we can add one. This also isn't meant as a claim that we shouldn't have integration tests, either. Doing the full, end-to-end sequence of steps is important to ensure that none of the plumbing connecting the different parts of the system is broken.  **But the goal of the network integration tests should be to test the plumbing, not all of the nooks and crannies of the state machine.**

So, how should we test the state machine?

[...]

PR #1675, opened 2022-11-29
component: better test infrastructure

### LINKED ISSUE #853: Create a `Storage::ephemeral_test_instance()` method
I think the test in question is using tmpdirs to make a testing `Storage` instance. This is a side note, but along the way to fixing it, I think it'd be useful to move that code into the `storage` crate as a `Storage::ephemeral_test_instance()` method that returns `(Self, Handle)`, where `Handle` is the RAII guard you get from the crate that makes temporary directories and cleans up the directory on drop.

_Originally posted by @hdevalence in https://github.com/penumbra-zone/penumbra/issues/797#issuecomment-1124001693_

hdevalence:
Work towards #1664

Closes #853

This commit cleans up the IBC client test as an example case of the new API.

Along the way, we also clean up some other ergonomic pain points, notably adding an `Arc<State>::try_begin_transaction()` method.

### DIFF
--- a/component/src/app/mod.rs
+++ b/component/src/app/mod.rs
@@ -4,7 +4,7 @@ use anyhow::Result;
 use penumbra_chain::params::FmdParameters;
 use penumbra_chain::{genesis, AppHash, StateWriteExt as _};
 use penumbra_proto::{Protobuf, StateWriteProto};
-use penumbra_storage::{State, Storage};
+use penumbra_storage::{ArcStateExt, State, Storage};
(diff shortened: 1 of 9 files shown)


hdevalence:
@avahowell Could you look at the content of the IBC test and check that we didn't clobber anything?

hdevalence:
Merging optimistically to be able to do some changes to the `ActionHandler` trait.

avahowell:
the IBC test looks good to me, I like this test infrastructure approach

PR #1680, opened 2022-11-29
Vendor cosmos SDK proto files and compile Tendermint proto service de…

zbuc:
…finition


pd/src/main.rs
```diff
@@ -57,9 +57,20 @@ enum RootCommand {
...
         #[clap(short, long, default_value = "9000")]
         metrics_port: u16,
+        /// Proxy Tendermint requests against the gRPC server to this host.
+        #[clap(
+            short,
+            long,
+            default_value = "testnet.penumbra.zone",
+            parse(try_from_str = url::Host::parse)
+        )]
+        tendermint_host: url::Host,
```
hdevalence, line 70: do we want this to default to `testnet.penumbra.zone`? since the `pd` app will be driven by a local tendermint instance, there will be one available, and it seems better to use that one.

if we make the default be `127.0.0.1`, then the Quickstart workflow will work without extra arguments.

or, do we actually want this to be two seperate arguments? what if it just took a `std::net::SocketAddr`?

pd/src/main.rs
```diff
@@ -57,11 +61,16 @@ enum RootCommand {
         /// Bind the gRPC server to this port.
         #[clap(short, long, default_value = "8080")]
         grpc_port: u16,
-        /// Bind the metrics endpoint to this port.
+        /// bind the metrics endpoint to this port.
         #[clap(short, long, default_value = "9000")]
         metrics_port: u16,
+        /// bind the tendermint proxy server to this port.
+        #[clap(short, long, default_value = "36657")]
+        tendermint_proxy_port: u16,
```
hdevalence, line 69: do we need a separate port for this? ideally we'd just add it int o the main GRPC port and let the router forward requests appropriately, since the goal is to have just one endpoint.

pd/src/info.rs
```diff
@@ -15,19 +15,32 @@ use crate::RequestExt;
...
+/// The fields of this struct are the configuration and data
+/// necessary to the gRPC services.
 #[derive(Clone, Debug)]
 pub struct Info {
+    /// Storage interface for retrieving chain state.
     storage: Storage,
+    /// Address of upstream Tendermint server to proxy requests to.
+    /// TODO: this is really only needed by the `tendermint_proxy::TendermintService` trait's methods,
+    /// should there be a second `Info` struct to implement that trait?
+    tendermint_url: url::Url,
```
hdevalence, line 34: Maybe we should make the tendermint proxy be its own completely separate thing?

pd/src/info/tendermint_proxy.rs
```diff
@@ -0,0 +1,359 @@
...
+        req: tonic::Request<GetSyncingRequest>,
+    ) -> Result<tonic::Response<GetSyncingResponse>, Status> {
+        todo!()
+    }
+
+    async fn get_latest_block(
+        &self,
+        req: tonic::Request<GetLatestBlockRequest>,
+    ) -> Result<tonic::Response<GetLatestBlockResponse>, Status> {
+        todo!()
```
hdevalence, line 126: We could replace some of these `todo!()`s with `tonic::Status::Unimplemented` and just only support the subset of methods that we actually use from `pcli`.

pd/src/info/tendermint_proxy.rs
```diff
@@ -0,0 +1,359 @@
...
+// We need to use the tracing-futures version of Instrument,
+// because we want to instrument a Stream, and the Stream trait
+// isn't in std, and the tracing::Instrument trait only works with
+// (stable) std types.
+//use tracing_futures::Instrument;
+
+use super::Info;
+
+#[tonic::async_trait]
+trait TendermintServiceExt {}
```
hdevalence, line 37: What's the goal of the trait extension here?

pd/src/info/tendermint_proxy.rs
zbuc: So it turns out that we don't actually use _any_ of the methods defined in the trait from pcli.

hdevalence: I think we do full block downloads in the view service?

zbuc: That's right, sorry. But we also call the "get status" call which isn't represented in the cosmos tendermint service definition

view/src/worker.rs
```diff
@@ -33,7 +36,8 @@ pub struct Worker {
     fvk: FullViewingKey, // TODO: notifications (see TODOs on ViewService)
     error_slot: Arc<Mutex<Option<anyhow::Error>>>,
     sync_height_tx: watch::Sender<u64>,
-    tm_client: tendermint_rpc::HttpClient,
+    tm_client: TendermintServiceClient<Channel>,
+    tm_proxy_client: TendermintProxyServiceClient<Channel>,
```
hdevalence, line 40: Could be later, but rather than maintaining a bunch of different clients, maybe we should just maintain one `Channel`, then whenever we need an XYZClient, we would clone the `Channel` and build the client we needed? https://docs.rs/tonic/latest/tonic/transport/struct.Channel.html#multiplexing-requests

view/src/worker.rs
```diff
@@ -77,9 +81,12 @@ impl Worker {
...
 
-        let tm_client = tendermint_rpc::HttpClient::new(
-            format!("http://{}:{}", node, tendermint_port).as_str(),
-        )?;
+        let tm_client =
+            TendermintServiceClient::connect(format!("http://{}:{}", node, tendermint_port))
+                .await?;
+        let tm_proxy_client =
+            TendermintProxyServiceClient::connect(format!("http://{}:{}", node, tendermint_port))
+                .await?;
```
hdevalence, line 89: these should be using `pd_port` now, right, since they can all be on the same endpoint?

proto/proto/penumbra/client/v1alpha1/client.proto
```diff
@@ -195,3 +206,74 @@ message PrefixValueResponse {
...
+  // Broadcast a transaction asynchronously.
+  rpc BroadcastTxAsync(BroadcastTxAsyncRequest) returns (BroadcastTxAsyncResponse) {
+    option (google.api.http).get = "/broadcast_tx_async";
+  }
+
+  // Broadcast a transaction synchronously.
+  rpc BroadcastTxSync(BroadcastTxSyncRequest) returns (BroadcastTxSyncResponse) {
+    option (google.api.http).get = "/broadcast_tx_sync";
+  }
+}
```
hdevalence, line 226: Did we make this RPC up, or did we vendor / adapt it from somewhere? (Not sure it really matters, just thought it might be good to note which case we were in)

zbuc: This is one that we made up, since the existing cosmos tendermint gRPC client didn't have these methods

proto/src/lib.rs
```diff
@@ -33,177 +33,224 @@ pub use state::StateReadProto;
...
 
-/// Core protocol structures.
-pub mod core {
-    /// Crypto structures.
-    pub mod crypto {
-        pub mod v1alpha1 {
-            include!("gen/penumbra.core.crypto.v1alpha1.rs");
+pub use penumbra::*;
+
+pub mod penumbra {
```
hdevalence, line 38: Not sure how much we get out of putting all the penumbra code in a submodule that we re-export, but i guess it's not the worst to be consistent with the other vendored protos.

view/src/service.rs
```diff
@@ -145,35 +148,23 @@ impl ViewService {
...
-                r#"http://{}:{}/status"#,
-                self.node, self.tendermint_port
-            ))
-            .send()
-            .await?
-            .json()
-            .await?;
+        let mut client = TendermintProxyServiceClient::connect(format!(
+            "http://{}:{}",
+            self.node, self.tendermint_port
```
hdevalence, line 153: This should be using `pd_port` to use just one endpoint, right?

hdevalence (approved):
This LGTM modulo collapsing `pd_port` and `tendermint_port` in the client code and propagating the removal of `tendermint_port` up through to the command line arguments.

pd/Cargo.toml
```diff
@@ -87,6 +88,7 @@ ed25519-consensus = "2"
 ed25519-dalek = "1.0.1"
 
 async-trait = "0.1.52"
+tendermint-rpc = { version = "0.26.0", features = ["http-client"] }
```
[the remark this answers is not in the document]
hdevalence: No, let's skip it.

pcli/src/command/query/tx.rs
```diff
@@ -18,7 +18,7 @@ impl Tx {
 
         // generic bounds on HttpClient::new are not well-constructed, so we have to
         // render the URL as a String, then borrow it, then re-parse the borrowed &str
-        let client = HttpClient::new(app.tendermint_url.to_string().as_ref()).unwrap();
+        let client = HttpClient::new(app.pd_url.to_string().as_ref()).unwrap();
```
hdevalence, line 21: Hmm, this is using the `tendermint_rpc` client to talk to the `pd_url`, don't we want to be using the new grpc proxy instead?

zbuc: good catch

PR #1681, opened 2022-11-30
Track jmt preimage

plaidfinch:


plaidfinch:
@zbuc @aubrika this has been superseded, right?

zbuc:
> @zbuc @aubrika this has been superseded, right?

We don't have the inverse lookup you were implementing here but I'm not sure what it was needed for

Issue #1556, opened 2022-10-26
Transport security and client authentication for View and Custody services


redshiftzero:
Currently, we don't have a meaningful transport security or client authentication story for the view or custody services.  This was fine when we were using the view service in-process, or as a local daemon on the same machine for testing, but is not fine as we get closer to the end goal of having the view and custody services act as a "personal RPC".

There are two aspects to this problem, transport security and client authentication:

* The transport security problem is to establish a secure channel to the view server, and authenticate the server identity.
* The client authentication problem (for the view service) is to have the client demonstrate to the view server that it has the capability to view the data it's requesting.
* The client authentication problem (for the custody service) is to have the client demonstrate to the custody service that it has the capability to *request* authorization of a transaction. [Note that this is weaker than the capability to *authorize* a transaction, which remains solely in the hands of the custody service].

In the current iteration of the view protocol, we do not attempt to solve transport security at all, and grant access based just on having the client supply the `AccountId` (a hash of the full viewing key). This isn't a good solution on either front.

### Transport Security

Since we're using GRPC, we should solve transport security using TLS.  The only significant design question here is how to authenticate the view server.  There are two options here.

**Public certificates**: one option is to use a DNS name to obtain a publicly verifiable certificate issued by a certificate authority.

**Pinned certificates**: another option is to have the view server create its own certificate, and for the client to pin it. Probably the most convenient way to handle this would be similar to the way that Tendermint Node IDs work: we'd hash the view server's certificate to `VIEW_SERVER_ID`, specify the address the client connects to as `VIEW_SERVER_ID@123.234.12.34`, and have the client validate the server's certificate against the pinned `VIEW_SERVER_ID`.

In either case, we would require the use of TLS to connect to the view service, and remove the current capability for non-encrypted connections.

Public certificates have the advantage that they can be used from web contexts, where transport security is handled and controlled by the browser. Pinned certificates have the advantage that they don't require a public DNS name, and can be used on local networks. 

On the client side, public certificates can be validated in a standard way, while pinned certificates would require hooking into the certificate validation logic of the TLS stack. On the server side, public certificates require management of the entire certificate lifecycle, while pinned certificates just require generating a one-off certificate.

### Client Authentication (View Service)

[...]

PR #1692, opened 2022-12-02
  proto(view): add bearer token field

### LINKED ISSUE #1682: Add client authorization to view protocol requests.

See #1556.

Add bearer tokens to the view protocol RPC, as described in that issue:

```
package penumbra.view.v1alpha1;

message ViewAuthToken {
  bytes inner = 1;
}

message ViewAuthRequest {
 core.crypto.v1alpha1.FullViewingKey fvk = 1;
}

message ViewAuthResponse {
 ViewAuthToken token = 1;
}

service ViewAuthService {
  rpc ViewAuth(ViewAuthRequest) returns (ViewAuthResponse);
}
```
Then replace the existing fvk hash / account id with the bearer tokens:
```
// ... in messages for the main view protocol ...

message NotesRequest {
  // Authorizes the request.
  ViewAuthToken token = 1;
  // If set, return spent notes as well as unspent notes.
  bool include_spent = 2;
  // ...
}
```

The existing `pviewd` code has logic to check the existing account ID.  We'll want to delete that logic entirely, but in this issue it's not necessary to replace -- it's more important to get the protos we use for data interchange correct, and we can add an implementation in a later PR.

conorsch:
We plan to implement bearer-token authentication. This change adds an optional field "token" to the view service protocol. We preserve the use of account_id, since it's used for filtering events in e.g. the custody service. The field numbers for both "account_id" and "token" have been changed to 14 and 15 respectively, in an attempt to stabilize the proto, so subsequent implementing token auth updates don't need to make proto-breaking changes to field numbers.
    
Closes #1682. Towards #1556.

### DIFF
--- a/view/src/client.rs
+++ b/view/src/client.rs
@@ -197,6 +197,7 @@ pub trait ViewClient {
         let notes = self
             .quarantined_notes(pb::QuarantinedNotesRequest {
                 account_id: Some(account_id.into()),
+                ..Default::default()
             })
             .await?;
         tracing::trace!(?notes);
@@ -225,6 +226,7 @@ pub trait ViewClient {
         let notes = self
             .quarantined_notes(pb::QuarantinedNotesRequest {
                 account_id: Some(account_id.into()),
+                ..Default::default()
             })
             .await?;
         tracing::trace!(?notes);
@@ -263,6 +265,7 @@ where
         let status = self
             .status(tonic::Request::new(pb::StatusRequest {
                 account_id: Some(account_id.into()),
+                ..Default::default()
             }))
             .await?
             .into_inner();
@@ -277,6 +280,7 @@ where
         let stream = self
             .status_stream(tonic::Request::new(pb::StatusStreamRequest {
                 account_id: Some(account_id.into()),
+                ..Default::default()
             }))
             .await?
             .into_inner();
@@ -376,6 +380,7 @@ where
                 account_id: Some(account_id.into()),
                 note_commitment: Some(note_commitment.into()),
                 await_detection: false,
+                ..Default::default()
             }),
         )
         .await?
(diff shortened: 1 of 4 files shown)


hdevalence:
We should remove the old fields in one go, since the goal is to reduce churn for external consumers of the protocol.

hdevalence (changes requested):
We should remove the account ID fields in this PR, since the goal is to reduce the amount of churn in our interfaces, rather than increasing it.

conorsch:
I'll work on snipping out the `account_id` logic. Had some trouble the first time around, handling the request logic for the custody soft kms. Will take another crack at it.

view/src/client.rs
```diff
@@ -196,6 +196,7 @@ pub trait ViewClient {
     ) -> Result<BTreeMap<AddressIndex, BTreeMap<asset::Id, Vec<QuarantinedNoteRecord>>>> {
         let notes = self
             .quarantined_notes(pb::QuarantinedNotesRequest {
+                token: Option::None,
                 account_id: Some(account_id.into()),
             })
```
hdevalence, line 201: Rather than explicitly setting `token: Option::None`, I think it'd be cleaner to do
```suggestion
            .quarantined_notes(pb::QuarantinedNotesRequest {
                account_id: Some(account_id.into()),
                ..Default::default()
            })
```
The advantage of this is that it's more clear to the reader that the `token` value is being left unset-as-a-default, rather than that we're explicitly setting it to `None` for some specific reason.

conorsch:
This is going to slip off of testnet 037 (https://github.com/penumbra-zone/penumbra/issues/1699); will still land the changes as described, but they'll ship in 038.

hdevalence (approved):

PR #1704, opened 2022-12-05
stake: fix bug in `validator_list()`

hdevalence:
The implementation was changed here:

https://github.com/penumbra-zone/penumbra/commit/03de128b249e7b3f2e5913c870f5cc5bdcb42755#diff-6e0042a0b016aaac378df25ae205439acf4f2e438704798a9870337b3ba35f25L1023-R1029

This implementation only calls `next()` on the _stream_ once, and then iterates over "maybe the first value of the stream", and maps that iterator into the desired shape, so while it typechecks, it only ever returns the first validator in the list. Instead, we want to use the `TryStreamExt` methods to map on the _stream_, and then `try_collect` the whole thing.

Testing this locally shows that it fixes the bug causing the chain halt on `037-megaclite`.

This bug isn't particularly interesting, which is in a way reassuring -- it's not a subtle logic error, instead it's the kind of thing that should ideally be caught by the test suite, so that we can do mechanical refactors like in the linked commit without having to worry too much. We should try to figure out why the test suite didn't catch it, and how we could improve tests to catch things like this in the future.

### DIFF
--- a/component/src/stake/component.rs
+++ b/component/src/stake/component.rs
@@ -5,7 +5,7 @@ use crate::Component;
 use ::metrics::{decrement_gauge, gauge, increment_gauge};
 use anyhow::{anyhow, Result};
 use async_trait::async_trait;
-use futures::StreamExt;
+use futures::{StreamExt, TryStreamExt};
 use penumbra_chain::quarantined::Slashed;
 use penumbra_chain::{genesis, Epoch, NoteSource, StateReadExt as _};
 use penumbra_crypto::{DelegationToken, IdentityKey, Value, STAKING_TOKEN_ASSET_ID};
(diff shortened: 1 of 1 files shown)


conorsch (approved):
Approving, since I watched @hdevalence uncover and fix this bug. Early results from running a patched `pd` binary against the halted chain state indicate this is a solid fix. 

Waiting for CI to pass, then will merge and proceed with `037-megaclite.1` hotfix.

zbuc:
The [network integration tests](https://github.com/penumbra-zone/penumbra/blob/main/pcli/tests/network_integration.rs) currently only call `pcli validator list` once, and only fetch the first validator from the list.

Issue #1659, opened 2022-11-23
Replace current `testnet-preview` deployment with new k8s deployment


hdevalence:
**Is your feature request related to a problem? Please describe.**

We should try to move over to the new k8s deployment system built by Strangelove, and start with replacing `testnet-preview`.  The goal of `testnet-preview` is that it should be an exact preview of what would be deployed if the current state of the `main` branch were tagged as a release.  This ensures that there are no deployment surprises when tagging a release, and allows testing client protocols against the current state of the `main` branch.

The only difference between `testnet-preview.` and `testnet.` should be that when deploying `testnet.`, we pass the `--preserve-chain-id` parameter to `pd testnet generate` to avoid randomizing the chain ID (since there should only be one deployment per tag).

**Describe the solution you'd like**

- [x] Update k8s helm chart logic to support side-by-side deploys - https://github.com/penumbra-zone/penumbra/pull/1707
- [x] Write new "deploy-preview-new" workflow for testnet-preview on k8s - https://github.com/penumbra-zone/penumbra/pull/1719
- [x] Ensure the action is triggered on pushes to `main` and uses the latest container images (does it need to wait for them to be built?) - https://github.com/penumbra-zone/penumbra/pull/1730
- [x] Merge new workflows into main, use GHA interface to trigger them
- [x] Compare `/status` endpoints between e.g. http://testnet-preview.penumbra.zone:26657/status & http://fullnode.testnet-preview.penumbra.zone:26657/status
- [ ] Ask for team testing against fullnode.testnet-preview.penumbra.zone 
- [x] Update DNS for cut-over on "testnet-preview" (can happen whenever, mostly internal use only)
- [x] Update DNS for cut-over (will need to be coordinated with next testnet deploy, otherwise we lose state)

hdevalence:
The current k8s deployment provides TLS access to the Tendermint RPC endpoint (load-balancing over fullnodes).  We should provide an additional endpoint that gives TLS access to the `pd` GRPC endpoint.  

We are not in a position to use a TLS endpoint from `pcli` for kind of boring reasons (we hardcode "http" in a bunch of places, and assume that we have one host for both `tendermint` + `pd` with endpoints on different ports), but exposing a TLS `pd` endpoint is important to do now because we're trying to use `grpc-web` to access it, and without TLS, this is not really possible because of mixed content rules.

conorsch:
Took a look at what's required here for cut-over. We currently run two discrete testnets:

[...]

PR #1707, opened 2022-12-06
ci: refactor helm logic for unique resources

conorsch:
Towards #1659.

## Overview

We want to run two, unrelated testnets, concurrently:
    
* testnet.penumbra.zone
* testnet-preview.penumbra.zone
    
These changes make the helm chart more explicit about labeling resources according to a "release" name, allowing   multiple deploys of the same logic to provision separate resources. It also adds "labels" to the resource, so that they can be selected by group, e.g.:
    
      kubectl get pods -l app.kubernetes.io/name=penumbra # shows all pods
      kubectl get pods -l app.kubernetes.io/instance=penumbra-testnet # shows just testnet pods

## Still to come    
No config yet exists for "testnet-preview", but this change allows that configuration to be added non-destructively. I'll follow up with more changes to add a new workflow. 

## How to review
The workflow logic in this PR is substantially different from that already in main; as a result, the two deploys will conflict with one another. That's fine for now, since post-merge I'll manually clean up the discrepancies. Only the "testnet" logic is targeted here, so it won't run automatically until another tag is pushed (but can still be run ad-hoc at any time). 

Happy to pair with anyone who is curious to know more. @agouin is currently AFK, but I'll brief them on the changes when they're back. The entirety of this diff affects **only** the filepaths `deployments/` and `.github/workflows/`; no source code for the project has changed.


conorsch:
Marking as ready for review. This PR doesn't add a "testnet-preview" deployment to k8s, despite the branch name. It does move things around such that a side-by-side deploy will be possible. There are still some kinks to work out in the multi-network deployment strategy, though: https://github.com/penumbra-zone/penumbra/issues/1659#issuecomment-1340143282

conorsch:
Pushed up one more commit, adding `--preserve-chain-id` to the k8s testnet deploy. Comparing the two currently active endpoints:

* http://testnet.penumbra.zone:26657/status
* http://fullnode.testnet.penumbra.zone:26657/status

made it clear that `node_info.network` needed to be changed to match. If anyone can spot other divergences, please mention them on #1659.

PR #1710, opened 2022-12-07
Add support for out-of-band note distribution

### LINKED ISSUE #1549: Native support for out-of-band note distribution.
Currently, the `CompactBlock` contains a list of annotated note payloads, each of which contains a note commitment, an encryption of the note plaintext, and the source of the note (usually a transaction hash).

This data allows a client to trial decrypt each note to discover which notes are addressed to it.  The client then has to insert *every* note commitment into the commitment tree, marking the ones it cares about for retention and marking the others for forgession. (The TCT accelerates this process by allowing a client to insert a precomputed block root when all note commitments in the block are irrelevant, but this decision can only be made after trial decryption).  

This is **in-band** note distribution: the note contents are distributed via the blockchain.  However, the chain only actually needs to see the note *commitments* to verify proofs.  The chain does not check the note ciphertexts and does not need to. Notes can also be distributed **out-of-band**, without using the chain state.

We'd like to change the semantics of the `CompactBlock` to have native support for out-of-band note distribution. There are two motivations for this:

1. Scalability: if a note is distributed out-of-band, every *other* user of Penumbra can skip scanning it.
2. Security: when performing asynchronous computation like a swap, we'd like later phases of execution to be fully deterministically derived from data signed in the original transaction (as in #1544), and at that point there's no need to be creating encrypted notes at all.

In effect, this allows any Penumbra user to create their own personal rollup, maintaining their own state off-chain and only posting state commitments, nullifiers, and proofs to the L1.  This seems particularly useful for active market makers, who can cheaply perform frequent updates to their own state (e.g., updating their liquidity positions potentially every block), while saving on gas and saving other users from scanning irrelevant state updates. 

There are other important problems to solve to enable this use case, notably, maintaining a fault-tolerant record of the state that has been moved off-chain, but those are out-of-scope of the core Penumbra protocol.

### Redefining `NotePayload` for out-of-band distribution

EDIT: **we decided not to do the below design, and instead to make a first-class `StatePayload` enum, defined as follows:**.
```rust
pub enum StatePayload {
    RolledUp(note::Commitment),
    Note {
        source: NoteSource,
        note: EncryptedNote,
    },
    Swap {
        source: NoteSource,
        swap: SwapPayload,
    },
}
```

Original:

The `NotePayload` domain and proto types are currently defined as follows:
```rust
pub struct NotePayload {
    pub note_commitment: note::Commitment,
    pub ephemeral_key: ka::Public,
    pub encrypted_note: [u8; note::NOTE_CIPHERTEXT_BYTES],
}
```
```proto
// The body of an output description, including only the minimal
// data required to scan and process the output.
message NotePayload {
  // The note commitment for the output note.
  NoteCommitment note_commitment = 1;
  // The encoding of an ephemeral public key.
  bytes ephemeral_key = 2;
  // An encryption of the newly created note.
  bytes encrypted_note = 3;
}
```

Since proto fields are optional, we'd leave the proto unchanged, but modify the domain type to be
```rust
pub struct EncryptedNote {
    pub note_commitment: note::Commitment,
    pub ephemeral_key: ka::Public,
    pub encrypted_note: [u8; note::NOTE_CIPHERTEXT_BYTES],
}

pub enum NotePayload {
    InBand(EncryptedNote),
    OutOfBand(note::Commitment),
}
```
The existing `NotePayload::trial_decrypt()` becomes `EncryptedNote::trial_decrypt`.  We also need to update the `auth_hash` implementation to encode the enum.

For the `AnnotatedNotePayload`, we don't care about the source if the note was sent out-of-band (the reason it's in the compact block in the first place is to allow querying for full transaction data), so we can again leave the proto unchanged, but change the domain type to
```rust
pub enum AnnotatedNotePayload {
    InBand {
        payload: EncryptedNote,
        source: NoteSource,
    },
    OutOfBand(note::Commitment),
}
```

### Nullifiers

**EDIT: we decided not to do this design, and instead to have distinct actions for rolled-up spends and outputs.**

Out-of-band note commitments skip most of the data currently an `Output` contributes to the `CompactBlock`, dropping from `32+32+168+32=264` to `32` bytes (not counting protobuf padding).  To complete the spend-output cycle, however, it would be useful to also support out-of-band nullifiers, dropping the contribution of a `Spend` from `32` to `0` bytes.

Unlike note commitments, which must be broadcast to every user to allow them to synchronize their local instance of the note commitment tree, nullifiers can be freely dropped from the `CompactBlock`, without affecting any user other than the one whose nullifier it was.

To support this, we should add a `bool out_of_band = 5;` to the `SpendBody` proto (and corresponding changes to the domain type).  If `out_of_band` is `true`, the shielded pool component records the nullifier as spent in the chain state as normal, but does *not* add it to the current `CompactBlock`.

### Required Changes

After

- [ ] redefining the `NotePayload` and `AnnotatedNotePayload` types
- [ ] adding `out_of_band` to `SpendBody`

we'll need to propagate those semantics through the rest of the node:

- [ ] update components to handle in-band and out-of-band note payloads;
- [ ] change the memo presence consensus rule so that the "number of outputs" only includes outputs with *in-band* notes, to allow OOB outputs to skip a memo field;
- [ ] update shielded pool component to skip adding `out_of_band` spends;

and then update the data types for the full lifecycle of pre- and post- construction transaction handling:

- [ ] update `TransactionPlan` structures to reflect out-of-band intent while planning;
- [ ] propagate those changes into the `Planner`, but without yet giving the possibility to create OOB spends or outputs;
- [ ] update `TransactionPerspective` to allow the possibility of including the openings of "rolled-up" `Output`s, similar to the way we already handle `Spend`s;
- [ ] update `TransactionView` structures to record both whether outputs were out-of-band and whether we have visibility into the OOB data;

and then update the view server to handle OOB data:

- [x] move the `height_created`, `address_index`, `source` columns of the `notes` table into the `spendable_notes` table, so that the `notes` table is purely a preimage lookup from note commitments to notes;
- [x] add `async fn Storage::record_note(&self, note: Note)` that computes the note commitment, then records the `Note` in the `notes` table (without making any assumptions that it's spendable, just allowing lookup by note commitment), so that we have a way to feed "advice" to the view server about known notes;
- [x] add `async fn Storage::match_notes(&self, commitments: impl Iterator<Item=&note::Commitment>) -> BTreeMap<note::Commitment, Note>` that matches a list of note commitments against the `notes` table and returns the known ones;
- [x] change the scanning logic to call `match_notes` against the note commitments in the `CompactBlock` to learn which ones were known out-of-band, and feeds them in as if they were decryption results;

and then finally prevent users from creating OOB transactions using the default signing mechanism, which doesn't maintain a fault tolerant log and therefore isn't safe to use with out-of-band data (what happens if you drop your database?):

- [ ] Add a policy check to `custody::SoftKms::sign` that rejects any `TransactionPlan`s with OOB spends or outputs.

### LINKED ISSUE #1711: Minimal out-of-band note distribution

#1549 describes and motivates out-of-band note distribution. Since we also want this for #1679, we'd like to start by implementing a minimal version.

Rather than doing a full implementation of out-of-band note distribution, we'll do the minimal set of changes needed to support #1679.

First, rename the existing  `NotePayload` to `EncryptedNote`, in both proto and domain types:
```rust
// Represents an encrypted note.
pub struct EncryptedNote {
    pub note_commitment: note::Commitment,
    pub ephemeral_key: ka::Public,
    pub encrypted_note: [u8; note::NOTE_CIPHERTEXT_BYTES],
}
```
Next, replace the existing `AnnotatedNotePayload` with a new `StatePayload`, defined as follows:
```rust
pub enum StatePayload {
    RolledUp(note::Commitment),
    Note {
        source: NoteSource,
        encrypted_note: EncryptedNote,
    }
}

impl StatePayload {
    pub fn commitment(&self) -> &note::Commitment {
        match self {
            NotePayload::InBand { encrypted_note, .. } => &encrypted_note.note_commitment,
            NotePayload::OutOfBand(note_commitment) => &note_commitment,
        }
    }
}
```
[The naming changed from the original suggestion in #1549 to be more reflective of what we're aiming for in general].

Then, change the `CompactBlock` to contain a list of `StatePayload`s.  Finally, propagate these changes through the codebase, in general replacing `AnnotatedNotePayload` with `StatePayload` and `NotePayload` with `EncryptedNote`, until it all compiles again.

aubrika:
Closes #1711


hdevalence:
Rebased on top of #1721

PR #1719, opened 2022-12-08
ci: adds testnet-preview k8s deploy workflow

conorsch:
Depends on #1707, so would prefer to land that first. Then this diff becomes pretty small.

Duplicates the "testnet" logic for "testnet-preview", targeting k8s cluster. Does not spin up a discrete cluster; we reuse a single cluster, with carefully labelled resources, and manage the configs via helm.


conorsch:
As with #1707, these changes have been manually applied to the cluster, but they'll break: every merge into `main` prior to this PR landing will result in a broken config. Once this is merged, I'll manually reset the cluster state once, then re-deploy via the GitHub Actions interview, and monitor for results.

PR #1721, opened 2022-12-08
Implement tokenized unbonding

hdevalence:
Work towards #1581


proto/proto/penumbra/core/transaction/v1alpha1/transaction.proto
```diff
@@ -156,6 +157,13 @@ message ActionView {
     dex.v1alpha1.PositionWithdraw position_withdraw = 32;
     dex.v1alpha1.PositionRewardClaim position_reward_claim = 34;
 
+    stake.v1alpha1.Delegate delegate = 41;
+    stake.v1alpha1.Undelegate undelegate = 42;
+    // TODO: we have no way to recover the opening of the undelegate_claim's
+    // balance commitment, and can only infer the value from looking at the rest
+    // of the transaction. is that fine?
```
hdevalence, line 164: For instance, in a transaction that splits the undelegate claim into two sub-actions (not sure why they'd do that, but nothing stops them), we can't see how much was in each, we can only infer the net amount. 

We also can't recover the full opening of the balance commitment in any case, because we don't retain the blinding factor, and commitments are perfectly hiding.

This seems like it's probably fine, or that to the extent it's not, it's not worth the cost of "fixing".

chain/src/note_source.rs
```diff
@@ -6,17 +6,25 @@ use serde::{Deserialize, Serialize};
...
     Genesis,
     FundingStreamReward { epoch_index: u64 },
     ProposalDepositRefund { proposal_id: u64 },
 }
 
+impl Default for NoteSource {
+    fn default() -> Self {
+        Self::Unknown
+    }
+}
```
hdevalence, line 19: This lets component implementations fill in a Default note source if one is not set, for instance when we're doing a unit test.

component/src/action_handler/actions/output.rs
```diff
@@ -2,43 +2,44 @@ use std::sync::Arc;
...
+    #[instrument(name = "output", skip(self, state))]
+    async fn execute(&self, state: &mut StateTransaction) -> Result<()> {
+        let source = state.object_get("source").cloned().unwrap_or_default();
+
+        state
+            .add_note(AnnotatedNotePayload {
+                source,
+                payload: self.body.note_payload.clone(),
+            })
+            .await;
```
hdevalence, line 41: Now that we don't have quarantining logic, we can process notes in each `ActionHandler` impl.

component/src/action_handler/actions/spend.rs
```diff
@@ -44,9 +47,14 @@ impl ActionHandler for Spend {
...
-    async fn execute(&self, _state: &mut StateTransaction) -> Result<()> {
-        // Handled in [`crate::action_handler::transaction::Transaction`].
+    #[instrument(name = "spend", skip(self, state))]
+    async fn execute(&self, state: &mut StateTransaction) -> Result<()> {
+        let source = state.object_get("source").cloned().unwrap_or_default();
+
+        state.spend_nullifier(self.body.nullifier, source).await;
+        // TODO: why do we manage event emission separately up at the top level
+        // instead of integrated into state machine?
+        state.record(shielded_pool::event::spend(self.body.nullifier));
```
hdevalence, line 57: I think this is a holdover from before we combined the state system and the events system; it seems like it would be better to fold the event handling into the component logic.

component/src/action_handler/actions/undelegate.rs
```diff
@@ -78,6 +86,11 @@ impl ActionHandler for Undelegate {
     async fn execute(&self, state: &mut StateTransaction) -> Result<()> {
         tracing::debug!(?self, "queuing undelegation for next epoch");
         state.stub_push_undelegation(self.clone());
+        // Register the undelegation's denom, so we clients can look it up later.
+        state
+            .register_denom(&self.unbonding_token().denom())
+            .await?;
+        // TODO: should we be tracking changes to token supply here or in end_epoch?
```
hdevalence, line 93: As a followup to this, we should nail down how we track changes to the token supply, and if we can do so -- the undelegate action reveals the amount, for now, but the undelegateclaim doesn't, so we may not be able to precisely track the supply of all the tokens precisely.

component/src/governance/execute.rs
```diff
@@ -66,6 +69,12 @@ pub async fn proposal_submit(
         .await;
     state.put_proposal_voting_end(proposal_id, voting_end).await;
 
+    // If there was a proposal submitted, ensure we track this so that clients
+    // can retain state needed to vote as delegators
+    let mut compact_block = state.stub_compact_block();
+    compact_block.proposal_started = true;
+    state.stub_put_compact_block(compact_block);
+
```
hdevalence, line 77: this moved out of the `Transaction` actionhandler, which now just calls into the options

component/src/governance/state_key.rs
```diff
@@ -1,4 +1,4 @@
-use penumbra_crypto::IdentityKey;
+use penumbra_crypto::stake::IdentityKey;
```
hdevalence, line 1: I rearranged the staking-related contents of the `crypto` crate into a `stake` module.  The crate/module organization continues to be a mess.

component/src/stake/component.rs
```diff
@@ -240,13 +244,10 @@ pub(crate) trait StakingImpl: StateWriteExt {
...
-                let penalty = self
-                    .get_chain_params()
-                    .await?
-                    .slashing_penalty_misbehavior_bps;
+                let penalty = self.get_chain_params().await?.slashing_penalty_misbehavior;
 
-                // Apply the penalty to the validator's current exchange rate.
-                self.apply_slashing_penalty(identity_key, penalty).await?;
+                // Record the slashing penalty on this validator.
+                self.record_slashing_penalty(identity_key, penalty).await?;
```
hdevalence, line 250: We no longer apply the slashing penalty to the validator's exchange rate immediately. Instead, we record the penalty when it happens, and apply the compounded penalty at the epoch boundary.

component/src/stake/component.rs
```diff
@@ -229,7 +231,9 @@ pub(crate) trait StakingImpl: StateWriteExt {
                 self.set_validator_bonding_state(
                     identity_key,
                     Unbonding {
-                        unbonding_epoch: self.current_unbonding_end_epoch().await?,
+                        unbonding_epoch: self
+                            .current_unbonding_end_epoch_for(&identity_key)
```
hdevalence, line 235: `current_unbonding_end_epoch` is now `current_unbonding_end_epoch_for(validator)` , so that it's not possible to use a generic unbonding period rather than the validator-specific one (which may be shorter, if the validator is already partly through unbonding).

component/src/stake/component.rs
```diff
@@ -323,13 +324,19 @@ pub(crate) trait StakingImpl: StateWriteExt {
...
                 .await?
                 .ok_or_else(|| {
                     anyhow::anyhow!("validator had ID in validator_list but rate not found in JMT")
                 })?;
+            // ... as soon as we apply any penalties recorded in the previous epoch.
+            let penalty = self
+                .penalty_in_epoch(&validator.identity_key, epoch_to_end.index)
+                .await?
+                .unwrap_or_default();
+            let current_rate = old_next_rate.slash(penalty);
```
hdevalence, line 339: Here we apply the slashing penalty we recorded earlier.

component/src/stake/component.rs
```diff
@@ -911,6 +918,39 @@ pub trait StateReadExt: StateRead {
...
+            .try_collect::<BTreeMap<String, Penalty>>()
+            .await?;
+
+        let start_key = state_key::penalty_in_epoch(id, start);
+        let end_key = state_key::penalty_in_epoch(id, end);
+
+        let mut compounded = Penalty::default();
+        for (_key, penalty) in all_penalties.range(start_key..end_key) {
+            compounded = compounded.compound(*penalty);
+        }
```
hdevalence, line 949: This pulls more data than it could, but we only do prefix queries, not range queries, and also efficiency doesn't matter here because there shouldn't be very many slashings.

crypto/src/stake/penalty.rs
```diff
@@ -0,0 +1,108 @@
...
+use serde::{Deserialize, Serialize};
+
+use crate::{asset, Amount, Balance, Value, STAKING_TOKEN_ASSET_ID};
+
+/// Tracks slashing penalties applied to a validator in some epoch.
+///
+/// The penalty is represented as a fixed-point integer in bps^2 (denominator 10^8).
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(try_from = "pbs::Penalty", into = "pbs::Penalty")]
+pub struct Penalty(pub u64);
```
hdevalence, line 13: The new `Penalty` tracks slashing penalties, with implementations of slashing and compounding penalties.

crypto/src/stake/penalty.rs
```diff
@@ -0,0 +1,108 @@
...
+
+impl Default for Penalty {
+    fn default() -> Self {
+        Penalty(0)
+    }
+}
```
hdevalence, line 19: This Default impl lets us unconditionally pull Penalties out of the state without having to actually record them in the common case (no slashing).

Issue #1712, opened 2022-12-07
Define a way to derive a nullifier for a positioned swap


hdevalence:
As part of #1679, we need a way to derive nullifiers for positioned swaps, analogous to nullifiers for positioned notes.

PR #1723, opened 2022-12-08
refactor: add `swap_commitment` method and use inner value as asset ID

### LINKED ISSUE #1715: Rename the `swap_nft_asset_id()` methods to `swap_commitment()`
As an intermediate step, we can use the inner value of the `SwapCommitment` where the `swap_nft_asset_id` is currently used.

redshiftzero:
Closes #1715 (required for #1712)

### DIFF
--- a/crypto/src/dex.rs
+++ b/crypto/src/dex.rs
@@ -1,7 +1,6 @@
 pub mod lp;
 pub mod swap;
 pub use swap::BatchSwapOutputData;
-pub use swap::SwapCommitment;
 
 mod trading_pair;
 pub use trading_pair::TradingPair;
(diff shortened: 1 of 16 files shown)


redshiftzero:
After discussion in Discord I've modified this to remove the `SwapCommitment` type in favor of using the existing `tct::Commitment` type both for note commitments (existing behavior) and swaps. 
The `NoteCommitment` proto is renamed to `StateCommitment` to avoid confusion, though there's other renaming that should probably occur at a later time, e.g. note commitment proofs should become state commitment proofs etc.

hdevalence:
Rebased on top of #1721

Issue #1396, opened 2022-09-12
Refactor wallet code to eliminate most `plan::foo` methods in favor of the Planner API


hdevalence:
**Is your feature request related to a problem? Please describe.**

Now that we have the Planner API, we have a bunch of methods like
```rust
/// Generate a new transaction plan delegating stake
#[instrument(skip(fvk, view, rng, rate_data, unbonded_amount, fee, source_address))]
pub async fn delegate<V, R>(
    fvk: &FullViewingKey,
    view: &mut V,
    rng: R,
    rate_data: RateData,
    unbonded_amount: u64,
    fee: Fee,
    source_address: Option<u64>,
) -> Result<TransactionPlan>
where
    V: ViewClient,
    R: RngCore + CryptoRng,
{
    Planner::new(rng)
        .fee(fee)
        .delegate(unbonded_amount, rate_data)
        .plan(view, fvk, source_address.map(Into::into))
        .await
        .context("can't build delegate plan")
}
```
which is just a worse wrapper around the Planner API.

**Describe the solution you'd like**

All of these methods, with the exception of `sweep`, should be eliminated, making sure that the planner API is at least as ergonomic as the method it's eliminating.  Alternatively, if there's some reason we can't eliminate one, we should have a reason why it doesn't fit the planner API. (For instance, we can't eliminate `sweep`, because it makes a list of transaction plans, while the planner API does a single transaction, so that's a differently shaped problem).

hdevalence:
When we do this, we should eliminate the `wallet` crate entirely. It's a vestigial holdover that currently serves no purpose.

PR #1736, opened 2022-12-09
Bind swaps to height

### LINKED ISSUE #1679: Swap NFT claim is not bound to height properly
**Describe the bug**

Original discussion in Discord:

> We prove that the swap claim uses the clearing price from the block that created the note recording the swap nft
.... but ... you can send the swap nft to yourself (thus including it in a different block height).
>
> So it seems like if there's a price later that you like better, you can send the swap nft to yourself and then get the new price.
> 
> The tricky thing is that we can't easily bind the height when creating the swap nft, because we don't know what block it will end up in.
> 
> instead of having the swap action produce an output note, it publishes the swap nft asset id -- which will be random, because it will have a blinding factor -- and then the chain creates a new "executed swap nft" (unit 1 of asset id hash(height, swap_nft)) and puts it into the note commitment tree.
> 
> then in swap claim, you could open the "executed swap nft" inside the proof to get an authenticated height?
> idk
> we should wait to fix this until after doing the swap blinding stuff
> the problem / lesson to learn is that when we resume execution, we can only resume execution of data actually committed to by the state nft

After further discussion, we realized that the best fix isn't the one above. Instead, a better fix has an overlap with #1549, which changes the semantics of the `CompactBlock` to allow us to add commitments into the TCT without having to have an encrypted note to go with the commitment.

Currently, we record the user's participation in a batch swap roughly as follows:

1. Create a `SwapPlaintext` recording the user's contribution to the swap (including their final claim address).
2. Poseidon-hash the `SwapPlaintext` to an `Fq` element, which we use as the "swap NFT" asset ID.
3. Create a note recording unit 1 of the swap NFT asset ID, controlled by the same address as the claim address.
4. Later, in the `SwapClaim`, do a special kind of spend action that skips spend authorization but proves that the spend is of 1 unit of an asset whose ID is the Poseidon hash of some `SwapPlaintext`, and proves that the output notes have the correct amounts for the batch, that they're controlled by the claim address, etc.

The first key point here is that the swap NFT asset ID acts as a commitment to the `SwapPlaintext`, and that opening that commitment inside the `SwapClaim` proof allows resuming the computation. But there's a bunch of extra steps that aren't actually necessary, and in this case, led to a bug. In particular, what is a note? A note is a typed quantity of value, plus a capability to spend that value. But here we're disregarding that capability, and saying that we have a different notion of what it means to consume the "value" -- namely, resuming the computation.

The second key point is that the swap commitment is also an `Fq` element, just like a note commitment. From the point of view of the TCT, all the leaves are just `Fq` elements; the fact that they're formed as note commitments is irrelevant.

### Solution

Once we have #1622 and #1549, all the ingredients for a much simpler solution are in place: we recast the note commitment tree as a _state commitment_ tree, define both note commitments and swap commitments as different kinds of state commitments, and insert the swap commitments directly into the state commitment tree rather than creating "swap NFTs". Then, to perform a swap claim, we simply witness a TCT inclusion path from the swap commitment up to the anchor, and prove that the path's block height matches the supplied `BatchSwapOutputData`. There's no need to make a fake, special-case spend proof, and the swap commitment is inserted in the TCT by the chain, so there's no more malleability issue.

Moreover, on the client side, we also have an opportunity to clean up the `CompactBlock` a bit, and prevent the current situation, where clients are forced to download additional data from a fullnode to understand their own notes. We'd do this by including the `SwapCiphertext`s and the `BatchSwapOutputData` in the `CompactBlock`.

- [x] #1713
- [x] #1714
- [x] #1715
- [ ] Rename the new `NotePayload` enum introduced in #1549 to `StatePayload`, with variants `RolledUp`, `Note` (containing an `EncryptedNote`), `Swap` (containing a `SwapCiphertext`).
- [ ] Change the `Swap` action to contain only the `SwapCommitment` and `SwapCiphertext`, rather than a general purpose note that happens to contain a swap NFT.
- [ ] Change the DEX component to record the `SwapCommitment`s in the TCT and to record the `StatePayload::Swap` in the `CompactBlock`.
- [ ] Update the `SwapProof` to reflect the simplification.
- [x] #1712
- [ ] Change the `SwapClaim` so that it:
  - [ ] reveals the swap nullifier
  - [ ] produces two note commitments for output notes
  - [ ] contains an amended `SwapClaimProof`, which proves:
    - [ ] knowledge of the swap plaintext corresponding to a swap commitment
    - [ ] knowledge of an inclusion proof for that swap commitment in the state commitment tree
    - [ ] consistency between the block height implied by the commitment position and the `BatchSwapOutputData`
    - [ ] that the swap nullifier was well-formed for the positioned swap commitment
    - [ ] that the output notes are of the correct amounts, and controlled by the address in the swap plaintext.

hdevalence:
Closes #1679

### DIFF
--- a/chain/src/sync/compact_block.rs
+++ b/chain/src/sync/compact_block.rs
@@ -1,7 +1,10 @@
-use std::convert::TryFrom;
+use std::{collections::BTreeMap, convert::TryFrom};
 
 use anyhow::Result;
-use penumbra_crypto::Nullifier;
+use penumbra_crypto::{
+    dex::{BatchSwapOutputData, TradingPair},
+    Nullifier,
+};
 use penumbra_proto::{
     client::v1alpha1::CompactBlockRangeResponse, core::chain::v1alpha1 as pb, Protobuf,
 };
@@ -18,18 +21,20 @@ use super::StatePayload;
 #[serde(try_from = "pb::CompactBlock", into = "pb::CompactBlock")]
 pub struct CompactBlock {
     pub height: u64,
-    // State payloads describing new state fragments.
+    /// State payloads describing new state fragments.
     pub state_payloads: Vec<StatePayload>,
-    // Nullifiers identifying spent notes.
+    /// Nullifiers identifying spent notes.
     pub nullifiers: Vec<Nullifier>,
-    // The block root of this block.
+    /// The block root of this block.
     pub block_root: block::Root,
-    // The epoch root of this epoch, if this block ends an epoch (`None` otherwise).
+    /// The epoch root of this epoch, if this block ends an epoch (`None` otherwise).
     pub epoch_root: Option<epoch::Root>,
-    // Latest FMD parameters. `None` if unchanged.
+    /// Latest FMD parameters. `None` if unchanged.
     pub fmd_parameters: Option<FmdParameters>,
-    // If the block indicated a proposal was being started.
+    /// If the block indicated a proposal was being started.
     pub proposal_started: bool,
+    /// Output prices for batch swaps occurring in this block.
+    pub swap_outputs: BTreeMap<TradingPair, BatchSwapOutputData>,
     // **IMPORTANT NOTE FOR FUTURE HUMANS**: if you want to add new fields to the `CompactBlock`,
     // you must update `CompactBlock::requires_scanning` to check for the emptiness of those fields,
     // because the client will skip processing any compact block that is marked as not requiring
@@ -46,6 +51,7 @@ impl Default for CompactBlock {
             epoch_root: None,
             fmd_parameters: None,
             proposal_started: false,
+            swap_outputs: BTreeMap::new(),
         }
     }
 }
@@ -77,6 +83,7 @@ impl From<CompactBlock> for pb::CompactBlock {
             epoch_root: cb.epoch_root.map(Into::into),
             fmd_parameters: cb.fmd_parameters.map(Into::into),
             proposal_started: cb.proposal_started,
+            swap_outputs: cb.swap_outputs.into_values().map(Into::into).collect(),
         }
     }
 }
@@ -92,6 +99,12 @@ impl TryFrom<pb::CompactBlock> for CompactBlock {
                 .into_iter()
                 .map(StatePayload::try_from)
                 .collect::<Result<Vec<StatePayload>>>()?,
+            swap_outputs: value
+                .swap_outputs
+                .into_iter()
+                .map(BatchSwapOutputData::try_from)
+                .map(|s| s.map(|swap_output| (swap_output.trading_pair, swap_output)))
+                .collect::<Result<BTreeMap<TradingPair, BatchSwapOutputData>>>()?,
             nullifiers: value
                 .nullifiers
                 .into_iter()
--- a/chain/src/sync/state_payload.rs
+++ b/chain/src/sync/state_payload.rs
@@ -1,7 +1,7 @@
 use std::convert::TryFrom;
 
 use anyhow::Result;
-use penumbra_crypto::{note, EncryptedNote};
+use penumbra_crypto::{dex::swap::SwapPayload, note, EncryptedNote};
 use penumbra_proto::core::chain::v1alpha1::{self as pb};
 
 use serde::{Deserialize, Serialize};
@@ -17,6 +17,10 @@ pub enum StatePayload {
         source: NoteSource,
         note: EncryptedNote,
     },
+    Swap {
+        source: NoteSource,
+        swap: SwapPayload,
+    },
 }
 
 impl StatePayload {
@@ -24,6 +28,7 @@ impl StatePayload {
         match self {
             Self::RolledUp(commitment) => commitment,
             Self::Note { note, .. } => &note.note_commitment,
+            Self::Swap { swap, .. } => &swap.commitment,
         }
     }
 
@@ -31,6 +36,7 @@ impl StatePayload {
         match self {
             Self::RolledUp(_) => None,
             Self::Note { source, .. } => Some(source),
+            Self::Swap { source, .. } => Some(source),
         }
     }
 }
@@ -50,6 +56,14 @@ impl From<StatePayload> for pb::StatePayload {
                     pb::state_payload::Note {
                         source: Some(source.into()),
                         note: Some(note.into()),
+                    },
+                )),
+            },
+            StatePayload::Swap { source, swap } => pb::StatePayload {
+                state_payload: Some(pb::state_payload::StatePayload::Swap(
+                    pb::state_payload::Swap {
+                        source: Some(source.into()),
+                        swap: Some(swap.into()),
                     },
                 )),
             },
@@ -74,6 +88,17 @@ impl TryFrom<pb::StatePayload> for StatePayload {
             })) => Ok(StatePayload::Note {
                 note: note
                     .ok_or_else(|| anyhow::anyhow!("missing note"))?
+                    .try_into()?,
+                source: source
+                    .ok_or_else(|| anyhow::anyhow!("missing source"))?
+                    .try_into()?,
+            }),
+            Some(pb::state_payload::StatePayload::Swap(pb::state_payload::Swap {
+                source,
+                swap,
+            })) => Ok(StatePayload::Swap {
+                swap: swap
+                    .ok_or_else(|| anyhow::anyhow!("missing swap"))?
                     .try_into()?,
                 source: source
                     .ok_or_else(|| anyhow::anyhow!("missing source"))?
--- a/component/src/action_handler/actions/swap.rs
+++ b/component/src/action_handler/actions/swap.rs
@@ -51,8 +51,12 @@ impl ActionHandler for Swap {
         state.put_swap_flow(&swap.body.trading_pair, swap_flow);
 
         // Record the swap commitment in the state.
+        let source = state.object_get("source").cloned().unwrap_or_default();
         state
-            .add_state_payload(StatePayload::RolledUp(swap.body.payload.commitment))
+            .add_state_payload(StatePayload::Swap {
+                source,
+                swap: self.body.payload.clone(),
+            })
             .await;
 
         Ok(())
--- a/component/src/action_handler/actions/swap.rs
+++ b/component/src/action_handler/actions/swap.rs
@@ -11,7 +11,7 @@ use decaf377::Fr;
 use penumbra_chain::sync::StatePayload;
 use penumbra_crypto::{MockFlowCiphertext, Value, STAKING_TOKEN_ASSET_ID};
 use penumbra_storage::{State, StateRead, StateTransaction};
-use penumbra_transaction::{action::Swap, Transaction};
+use penumbra_transaction::{action::Swap, IsAction, Transaction};
 use tracing::instrument;
 
 use crate::action_handler::ActionHandler;
@@ -20,39 +20,17 @@ use crate::action_handler::ActionHandler;
 impl ActionHandler for Swap {
     #[instrument(name = "swap", skip(self, _context))]
     async fn check_stateless(&self, _context: Arc<Transaction>) -> Result<()> {
-        let swap = self;
-
-        // Check swap proof
-        swap.proof
-            .verify(
-                // TODO: no value commitments until flow encryption is available
-                // so we pass placeholder values here, the proof doesn't check these right now
-                // and will fail when checking is re-enabled.
-                Value {
-                    amount: 0u64.into(),
-                    asset_id: *STAKING_TOKEN_ASSET_ID,
-                }
-                .commit(Fr::zero()),
-                Value {
-                    amount: 0u64.into(),
-                    asset_id: *STAKING_TOKEN_ASSET_ID,
-                }
-                .commit(Fr::zero()),
-                swap.body.fee_commitment,
-                swap.body.swap_nft.note_commitment,
-                swap.body.swap_nft.ephemeral_key,
-            )
-            .context("A swap proof did not verify")?;
-
-        // TODO: are any other checks necessary?
+        self.proof.verify(
+            self.body.fee_commitment,
+            self.body.payload.commitment,
+            self.balance_commitment(),
+        )?;
 
         Ok(())
     }
 
     #[instrument(name = "swap", skip(self, _state))]
     async fn check_stateful(&self, _state: Arc<State>) -> Result<()> {
-        // TODO: are any other checks necessary?
-
         Ok(())
     }
 
@@ -72,13 +50,9 @@ impl ActionHandler for Swap {
         // Set the batch swap flow for the trading pair.
         state.put_swap_flow(&swap.body.trading_pair, swap_flow);
 
-        // Record the Swap NFT in the state.
-        let source = state.object_get("source").cloned().unwrap_or_default();
+        // Record the swap commitment in the state.
         state
-            .add_state_payload(StatePayload::Note {
-                source,
-                note: self.body.swap_nft.clone(),
-            })
+            .add_state_payload(StatePayload::RolledUp(swap.body.payload.commitment))
             .await;
 
         Ok(())
--- a/component/src/action_handler/actions/swap.rs
+++ b/component/src/action_handler/actions/swap.rs
@@ -4,12 +4,12 @@ use crate::{
     dex::{StateReadExt as _, StateWriteExt as _},
     shielded_pool::NoteManager,
 };
-use anyhow::{Context, Result};
-use ark_ff::Zero;
+use anyhow::Result;
+
 use async_trait::async_trait;
-use decaf377::Fr;
+
 use penumbra_chain::sync::StatePayload;
-use penumbra_crypto::{MockFlowCiphertext, Value, STAKING_TOKEN_ASSET_ID};
+use penumbra_crypto::MockFlowCiphertext;
 use penumbra_storage::{State, StateRead, StateTransaction};
 use penumbra_transaction::{action::Swap, IsAction, Transaction};
 use tracing::instrument;
--- a/component/src/action_handler/actions/swap_claim.rs
+++ b/component/src/action_handler/actions/swap_claim.rs
@@ -22,7 +22,7 @@ impl ActionHandler for SwapClaim {
                 context.anchor,
                 self.body.nullifier,
                 self.body.output_data,
-                self.body.epoch_duration,
+                self.epoch_duration,
                 self.body.output_1_commitment,
                 self.body.output_2_commitment,
                 self.body.fee.clone(),
@@ -39,7 +39,7 @@ impl ActionHandler for SwapClaim {
         // 1. Validate the epoch duration passed in the swap claim matches
         // what we know.
         let epoch_duration = state.get_epoch_duration().await?;
-        let provided_epoch_duration = swap_claim.body.epoch_duration;
+        let provided_epoch_duration = swap_claim.epoch_duration;
         if epoch_duration != provided_epoch_duration {
             return Err(anyhow::anyhow!(
                 "provided epoch duration does not match chain epoch duration"
--- a/component/src/action_handler/actions/swap_claim.rs
+++ b/component/src/action_handler/actions/swap_claim.rs
@@ -17,28 +17,17 @@ use crate::{
 impl ActionHandler for SwapClaim {
     #[instrument(name = "swap_claim", skip(self))]
     async fn check_stateless(&self, context: Arc<Transaction>) -> Result<()> {
-        let swap_claim = self;
-
-        let fee = swap_claim.body.fee.clone();
-
-        // Check swap claim proof
-        let anchor = context.anchor;
-        swap_claim
-            .proof
+        self.proof
             .verify(
-                anchor,
-                swap_claim.body.nullifier,
-                swap_claim.body.output_data,
-                swap_claim.body.epoch_duration,
-                swap_claim.body.output_1.note_commitment,
-                swap_claim.body.output_2.note_commitment,
-                fee,
-                swap_claim.body.output_1.ephemeral_key,
-                swap_claim.body.output_2.ephemeral_key,
+                context.anchor,
+                self.body.nullifier,
+                self.body.output_data,
+                self.body.epoch_duration,
+                self.body.output_1_commitment,
+                self.body.output_2_commitment,
+                self.body.fee.clone(),
             )
             .context("a swap claim proof did not verify")?;
-
-        // TODO: any other stateless checks?
 
         Ok(())
     }
@@ -87,16 +76,14 @@ impl ActionHandler for SwapClaim {
         let source = state.object_get("source").cloned().unwrap_or_default();
 
         state
-            .add_state_payload(StatePayload::Note {
-                source,
-                note: self.body.output_1.clone(),
-            })
+            .add_state_payload(StatePayload::RolledUp(
+                self.body.output_1_commitment.clone(),
+            ))
             .await;
         state
-            .add_state_payload(StatePayload::Note {
-                source,
-                note: self.body.output_2.clone(),
-            })
+            .add_state_payload(StatePayload::RolledUp(
+                self.body.output_2_commitment.clone(),
+            ))
             .await;
 
         state.spend_nullifier(self.body.nullifier, source).await;
--- a/component/src/dex/component.rs
+++ b/component/src/dex/component.rs
@@ -1,5 +1,6 @@
 use std::collections::BTreeMap;
 
+use crate::shielded_pool::{StateReadExt as _, StateWriteExt as _};
 use crate::Component;
 use anyhow::Result;
 use async_trait::async_trait;
@@ -126,9 +127,14 @@ impl<T: StateRead> StateReadExt for T {}
 #[async_trait]
 pub trait StateWriteExt: StateWrite + StateReadExt {
     fn set_output_data(&mut self, output_data: BatchSwapOutputData) {
+        // Write the output data to the state under a known key, for querying, ...
         let height = output_data.height;
         let trading_pair = output_data.trading_pair;
         self.put(state_key::output_data(height, trading_pair), output_data);
+        // ... and also add it to the compact block to be pushed out to clients.
+        let mut compact_block = self.stub_compact_block();
+        compact_block.swap_outputs.insert(trading_pair, output_data);
+        self.stub_put_compact_block(compact_block);
     }
 
     fn set_stub_cpmm_reserves(&mut self, trading_pair: &TradingPair, reserves: Reserves) {
--- a/component/src/dex/tests.rs
+++ b/component/src/dex/tests.rs
@@ -10,7 +10,7 @@ use penumbra_crypto::{
 use penumbra_storage::{ArcStateExt, TempStorage};
 use penumbra_transaction::{
     plan::{SwapClaimPlan, SwapPlan},
-    Action, Transaction, TransactionBody,
+    Transaction,
 };
 use rand_core::SeedableRng;
 use tendermint::abci;
--- a/component/src/dex/tests.rs
+++ b/component/src/dex/tests.rs
@@ -86,23 +86,18 @@ async fn swap_and_swap_claim() -> anyhow::Result<()> {
 
     let output_data = state.output_data(height, trading_pair).await?.unwrap();
 
-    let swap_nft_note = client
-        .note_by_commitment(&swap.body.swap_nft.note_commitment)
-        .expect("client should have detected the swap nft note");
+    let commitment = swap.body.payload.commitment;
+    let swap_auth_path = client.witness(commitment).unwrap();
+    let detected_plaintext = client.swap_by_commitment(&commitment).unwrap();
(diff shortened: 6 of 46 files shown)


hdevalence:
Hmm, running
```
cargo run --release --bin pcli -- -n localhost tx swap 10gn --into gm
```
gives a `    balance commitment mismatch` in pre-submission checks (stateless), but
```
cargo run --release --bin pcli -- -n localhost tx swap 10gm --into gn
```
works fine, maybe there's some order dependence going on...

hdevalence:
> works fine, maybe there's some order dependence going on...

Indeed, I typoed the index (fixed in https://github.com/penumbra-zone/penumbra/pull/1736/commits/ed3436a97037a40a8895ee96ade4aa134c5976b5) and it wasn't detected because the trading pair we use in the unit test happens to have a particular order.  We should extend it to swap back, or add another test that runs multiple swaps at once.

chain/src/sync/compact_block.rs
```diff
@@ -18,18 +21,20 @@ use super::StatePayload;
...
+    /// The epoch root of this epoch, if this block ends an epoch (`None` otherwise).
     pub epoch_root: Option<epoch::Root>,
-    // Latest FMD parameters. `None` if unchanged.
+    /// Latest FMD parameters. `None` if unchanged.
     pub fmd_parameters: Option<FmdParameters>,
-    // If the block indicated a proposal was being started.
+    /// If the block indicated a proposal was being started.
     pub proposal_started: bool,
+    /// Output prices for batch swaps occurring in this block.
+    pub swap_outputs: BTreeMap<TradingPair, BatchSwapOutputData>,
```
hdevalence, line 37: This PR adds the batch swap outputs into the compact block, so that the view service can determine the corresponding notes purely from the contents of the `CompactBlock`. I think this should be a design rule going forward: all of a user's note state should be recoverable purely from the contents of the `CompactBlock`, even if the *interpretation* of that state, or further action on it, requires additional information (e.g., looking up asset IDs, or looking up validator rates to make a delegate/undelegate/etc).

chain/src/sync/state_payload.rs
```diff
@@ -17,20 +17,26 @@ pub enum StatePayload {
         source: NoteSource,
         note: EncryptedNote,
     },
+    Swap {
+        source: NoteSource,
+        swap: SwapPayload,
+    },
```
hdevalence, line 23: We add a new `Swap` variant, which also means that the `SwapPayload`s (encrypted `SwapPlaintext`s) go in the compact block and don't need to be downloaded seperately.

component/src/action_handler/actions/swap.rs
```diff
@@ -20,39 +20,17 @@ use crate::action_handler::ActionHandler;
...
-                swap.body.swap_nft.ephemeral_key,
-            )
-            .context("A swap proof did not verify")?;
-
-        // TODO: are any other checks necessary?
+        self.proof.verify(
+            self.body.fee_commitment,
+            self.body.payload.commitment,
+            self.balance_commitment(),
+        )?;
```
hdevalence, line 27: I streamlined the `Swap`/`SwapClaim` proofs, but there's still some followup work there (will note points)

component/src/action_handler/actions/swap_claim.rs
```diff
@@ -50,7 +39,7 @@ impl ActionHandler for SwapClaim {
         // 1. Validate the epoch duration passed in the swap claim matches
         // what we know.
         let epoch_duration = state.get_epoch_duration().await?;
-        let provided_epoch_duration = swap_claim.body.epoch_duration;
+        let provided_epoch_duration = swap_claim.epoch_duration;
```
hdevalence, line 42: I moved the `epoch_duration` out of the body of the swap claim, because it should be considered as a public input to the proof which the user happens to supply in the transaction, like the commitment tree anchor, and not something that should be signed over.

component/src/action_handler/actions/swap_claim.rs
```diff
@@ -87,16 +76,14 @@ impl ActionHandler for SwapClaim {
...
+            ))
             .await;
         state
-            .add_state_payload(StatePayload::Note {
-                source,
-                note: self.body.output_2.clone(),
-            })
+            .add_state_payload(StatePayload::RolledUp(
+                self.body.output_2_commitment.clone(),
+            ))
```
hdevalence, line 86: Now that we have support for bare note commitments, users don't have to expend any work scanning other users' swap outputs (which aren't directed at them anyways).

component/src/dex/component.rs
```diff
@@ -126,9 +127,14 @@ impl<T: StateRead> StateReadExt for T {}
...
 pub trait StateWriteExt: StateWrite + StateReadExt {
     fn set_output_data(&mut self, output_data: BatchSwapOutputData) {
+        // Write the output data to the state under a known key, for querying, ...
         let height = output_data.height;
         let trading_pair = output_data.trading_pair;
         self.put(state_key::output_data(height, trading_pair), output_data);
+        // ... and also add it to the compact block to be pushed out to clients.
+        let mut compact_block = self.stub_compact_block();
+        compact_block.swap_outputs.insert(trading_pair, output_data);
+        self.stub_put_compact_block(compact_block);
```
hdevalence, line 137: This is where we add the output data into the compact block.

component/src/dex/tests.rs
```diff
@@ -86,23 +86,18 @@ async fn swap_and_swap_claim() -> anyhow::Result<()> {
...
-        swap_nft_note_auth_path.position(),
-        epoch_duration,
+    let claim_plan = SwapClaimPlan {
+        swap_plaintext: plaintext,
+        position: swap_auth_path.position(),
         output_data,
-    );
-    let claim = claim_plan.swap_claim(&test_keys::FULL_VIEWING_KEY, &swap_nft_note_auth_path);
+        epoch_duration,
+    };
```
hdevalence, line 99: Now that the `SwapClaim` is fully determined by the (positioned) swap and the output data, it doesn't really make sense to have `SwapClaimPlan::new`, since it's not saving anything over using the bare struct (but does lose the nice struct initialization syntax).

component/src/mock_client.rs
```diff
@@ -75,8 +78,39 @@ impl MockClient {
...
+                            // Pre-insert the output notes into our notes table, so that
+                            // we can notice them when we scan the block where they are claimed.
+                            self.notes.insert(output_1.commit(), output_1);
+                            self.notes.insert(output_2.commit(), output_2);
+                        }
+                        None => {
+                            self.nct.insert(Forget, payload.commitment)?;
+                        }
+                    }
+                }
```
hdevalence, line 105: The `MockClient` does double duty, also serving as a worked example of the scanning logic without any fancy optimizations, and here is the logic for scanning `Swap`s: after trial decryption, grab the output data that's bundled into the compact block, and use it to pre-insert the notes that will be generated in the claim, ...

component/src/mock_client.rs
```diff
@@ -75,8 +78,39 @@ impl MockClient {
...
                 StatePayload::RolledUp(commitment) => {
-                    self.nct.insert(Forget, commitment)?;
+                    if self.notes.contains_key(&commitment) {
+                        // This is a note we anticipated, so retain its auth path.
+                        self.nct.insert(Keep, commitment)?;
+                    } else {
+                        // This is someone else's note.
+                        self.nct.insert(Forget, commitment)?;
+                    }
```
hdevalence, line 113: ... so that when the `SwapClaim` generates a `RolledUp(commitment)`, we can recognize it as the output notes we already precomputed.

hdevalence: (It's a moot point for the mock client, but we don't want the sync process to automatically generate a claim, though, because there's no way for the sync process to know whether or not a claim has already been submitted until reaching the chain tip, and submitting a duplicate one will leak information.)

crypto/src/dex/swap/payload.rs
```diff
@@ -0,0 +1,88 @@
...
+
+use super::{SwapCiphertext, SwapPlaintext};
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(try_from = "pb::SwapPayload", into = "pb::SwapPayload")]
+pub struct SwapPayload {
+    pub commitment: penumbra_tct::Commitment,
+    pub ephemeral_key: ka::Public,
+    pub encrypted_swap: SwapCiphertext,
+}
```
hdevalence, line 16: This is the analogue of the `EncryptedNote`. It's named this way because of the realization that, after all that refactoring and renaming, it might actually be best to call the `EncryptedNote` a `NotePayload` after all, so that we have "state payloads" which are an enum of "rolled up data", "note payloads", "swap payloads", etc.

crypto/src/dex/swap/plaintext.rs
```diff
@@ -49,6 +52,38 @@ impl SwapPlaintext {
         )
     }
 
+    pub fn output_notes(&self, batch_data: &BatchSwapOutputData) -> (Note, Note) {
+        let (output_1_blinding, output_2_blinding) = self.output_blinding_factors();
+
+        let (lambda_1_i, lambda_2_i) = batch_data.pro_rata_outputs((
+            // TODO: fix up amount conversions
+            self.delta_1_i.try_into().unwrap(),
+            self.delta_2_i.try_into().unwrap(),
```
hdevalence, line 61: Here and elsewhere there are a lot of issues where we do internal math on `u64` values, and bridging that with an `Amount` type is difficult. I think this speaks to some limitations of the existing `Amount` type we should address, in particular the lack of an analogue of a "widening multiply + divide" (which we can do for `u64` by means of `u128`).

We should fix those limitations and use `Amount` everywhere before trying to push forward on 128-bit `Amount`s.

crypto/src/proofs/transparent.rs
```diff
@@ -557,97 +472,49 @@ impl TryFrom<transparent_proofs::SwapClaimProof> for SwapClaimProof {
...
+                asset_id: self.swap_plaintext.trading_pair.asset_2(),
+            };
+        let transparent_balance_commitment = transparent_balance.commit(Fr::zero());
+
+        // XXX sign error here
+        // XXX we want to avoid having to twiddle signs for synthetic blinding factor in binding sig
+        ensure!(
+            balance_commitment == transparent_balance_commitment + fee_commitment,
+            "balance commitment mismatch"
+        );
```
hdevalence, line 517: Followup: review signs for fee commitment in Swap actions

I don't think either this code or the original code was correctly handling signs for the fee commitment. I suspect there are bugs lurking that we only miss because we set 0 fees by default.

We should do a followup pass and:
- [ ] review the signs of fees as they're used in `Swap`s
- [ ] write a state machine integration test that creates and executes complete, fully-valid `Transaction`s (using the test account's note state) that sets distinct nonzero fees for both `Swap` and `SwapClaim` transactions.

pcli/src/command/tx.rs
```diff
@@ -215,73 +217,53 @@ impl TxCmd {
...
-                let mut client = app.specific_client().await?;
-                let output_data: BatchSwapOutputData = client
-                    .batch_swap_output_data(BatchSwapOutputDataRequest {
-                        height: swap_nft_record.height_created,
-                        trading_pair: Some(swap_plaintext.trading_pair.into()),
-                    })
-                    .await?
-                    .into_inner()
-                    .try_into()
-                    .context("cannot parse batch swap output data")?;
```
hdevalence, line 272: Note: all of this goes away, we no longer have to leak any queries to a fullnode to be able to produce the `SwapClaim`

proto/proto/penumbra/core/dex/v1alpha1/dex.proto
```diff
@@ -27,43 +27,46 @@ message SwapClaim {
...
+  // The nullifier for the Swap commitment to be consumed.
   crypto.v1alpha1.Nullifier nullifier = 1;
   // The fee allows `SwapClaim` without an additional `Spend`.
   crypto.v1alpha1.Fee fee = 2;
   // Note output for asset 1.
-  crypto.v1alpha1.EncryptedNote output_1 = 3;
+  crypto.v1alpha1.StateCommitment output_1_commitment = 3;
   // Note output for asset 2.
-  crypto.v1alpha1.EncryptedNote output_2 = 4;
+  crypto.v1alpha1.StateCommitment output_2_commitment = 4;
```
hdevalence, line 43: The `EncryptedNote`s are replaced with bare `StateCommitment`s, because their contents are deterministically derived from the `SwapPlaintext`+`BatchSwapOutputData`.

proto/proto/penumbra/core/dex/v1alpha1/dex.proto
```diff
@@ -27,43 +27,46 @@ message SwapClaim {
...
-  crypto.v1alpha1.EncryptedNote swap_nft = 5;
-  // Encrypted version of the original `Swap`, symmetrically encrypted w/ viewing key.
-  bytes swap_ciphertext = 6;
+  // The swap commitment and encryption of the swap data.
+  SwapPayload payload = 5;
+}
+
+message SwapPayload {
+  crypto.v1alpha1.StateCommitment commitment = 1;
+  // TODO: replace with ovk-based encryption
```
hdevalence, line 67: Followup: should we be encrypting `SwapPlaintext`s with the IVK or OVK?

I'm not sure the distinction is particularly important, but we might be able to save a little bit of space + compute on the client side by using the (symmetric) OVK rather than the (asymmetric) IVK.

proto/proto/penumbra/core/transaction/v1alpha1/transaction.proto
```diff
@@ -81,6 +81,9 @@ message Action {
 message TransactionPerspective {
   repeated PayloadKeyWithCommitment payload_keys = 1;
   repeated NullifierWithNote spend_nullifiers = 2;
+  // The openings of note commitments referred to in the transaction
+  // but not included in the transaction.
+  repeated crypto.v1alpha1.Note advice_notes = 3;
```
hdevalence, line 86: Adding view support to the new `SwapClaim` where only the commitments, not encrypted notes, are present required adding a new piece of data to the `TransactionPerspective`, to be able to supply the openings of rolled-up state commitments.

hdevalence: We can also use this for, e.g., a future `RolledUpOutput` action that doesn't include the encrypted note in the compact block.

proto/proto/penumbra/view/v1alpha1/view.proto
```diff
@@ -46,6 +47,8 @@ service ViewProtocolService {
 
   // Query for a note by its note commitment, optionally waiting until the note is detected.
   rpc NoteByCommitment(NoteByCommitmentRequest) returns (NoteByCommitmentResponse);
+  // Query for a swap by its swap commitment, optionally waiting until the swap is detected.
+  rpc SwapByCommitment(SwapByCommitmentRequest) returns (SwapByCommitmentResponse);
```
hdevalence, line 51: Since swaps are different from notes now (we're not wrapping a swap commitment in a note, as with the former swap nft), we also need to be able to query the view service for swaps.  However, unlike notes, we probably don't need a flexible query, since we care about this for the purpose of looking up data on a swap we submitted and want to claim.

proto/proto/penumbra/view/v1alpha1/view.proto
```diff
@@ -268,3 +285,13 @@ message SpendableNoteRecord {
...
+
+message SwapRecord {
+  core.crypto.v1alpha1.StateCommitment swap_commitment = 1;
+  core.dex.v1alpha1.SwapPlaintext swap = 2;
+  uint64 position = 3;
+  core.crypto.v1alpha1.Nullifier nullifier = 4;
+  core.dex.v1alpha1.BatchSwapOutputData output_data = 5;
+  optional uint64 height_claimed = 6;
+  core.chain.v1alpha1.NoteSource source = 7;
+}
```
hdevalence, line 297: This is the analogue of the `SpendableNoteRecord`, with information on the `Swap` that's helpful for a client (e.g., it needs the position and output data to produce the claim).

transaction/src/action/swap_claim.rs
```diff
@@ -19,42 +21,26 @@ impl IsAction for SwapClaim {
...
-            (payload_key_1, payload_key_2)
-        {
-            // * Decrypt notes
-            let decrypted_note_1 =
-                Note::decrypt_with_payload_key(&self.body.output_1.encrypted_note, payload_key_1);
-            let decrypted_note_2 =
-                Note::decrypt_with_payload_key(&self.body.output_2.encrypted_note, payload_key_2);
+        // Get the advice notes for each output from the swap claim
+        let output_1 = txp.advice_notes.get(&self.body.output_1_commitment);
+        let output_2 = txp.advice_notes.get(&self.body.output_2_commitment);
```
hdevalence, line 26: To view a `SwapClaim`'s output notes, we need to use the new advice mechanism to be able to get the openings of the rolled-up output note commitments.

view/migrations/20220908223928_notes.sql
```diff
@@ -34,11 +34,17 @@ CREATE TABLE spendable_notes (
     height_spent            BIGINT 
 );
 
+CREATE INDEX spendable_notes_by_nullifier_idx ON spendable_notes (
+    nullifier
+);
+
+CREATE INDEX spendable_notes_by_source_idx ON spendable_notes (
+    source
+);
```
hdevalence, line 43: This is otherwise unrelated, but it seemed better to have separate indexes for nullifier / source, both of which are random data, than the more meaningful / structured data in the other index.

view/migrations/20221210013919_swaps.sql
```diff
@@ -0,0 +1,13 @@
...
+    swap_commitment         BLOB PRIMARY KEY NOT NULL,
+    swap                    BLOB NOT NULL,
+    position                BIGINT NOT NULL,
+    nullifier               BLOB NOT NULL,
+    output_data             BLOB NOT NULL,
+    height_claimed          BIGINT,
+    source                  BLOB NOT NULL
+);
+
+CREATE INDEX swaps_nullifier_idx ON swaps (nullifier);
```
hdevalence, line 13: Treating swaps distinctly from notes means having to record them in the view service separately; this uses the `spendable_notes` table as an example.

view/migrations/20221210013919_swaps.sql
```diff
@@ -0,0 +1,13 @@
+-- This table records the mapping from swap commitments to swap plaintexts.
+-- For now we just store the swap plaintexts as a blob.
+CREATE TABLE swaps (
+    swap_commitment         BLOB PRIMARY KEY NOT NULL,
+    swap                    BLOB NOT NULL,
+    position                BIGINT NOT NULL,
+    nullifier               BLOB NOT NULL,
+    output_data             BLOB NOT NULL,
+    height_claimed          BIGINT,
```
hdevalence, line 9: Followup: update the view server's `swaps.height_claimed` when a swap is claimed.

view/src/service.rs
```diff
@@ -256,19 +256,22 @@ impl ViewProtocolService for ViewService {
...
-
-                if spendable_note_record.is_err() {
-                    spend_nullifiers.insert(nullifier, None);
-                } else if let Ok(spendable_note_record) = spendable_note_record {
-                    spend_nullifiers.insert(nullifier, Some(spendable_note_record.note));
+                // An error here indicates we don't know the nullifier, so we omit it from the Perspective.
+                if let Ok(spendable_note_record) =
+                    self.storage.note_by_nullifier(nullifier, false).await
+                {
+                    spend_nullifiers.insert(nullifier, spendable_note_record.note);
```
hdevalence, line 263: This simplifies the `Perspective` data modeling: either the nullifier is known, in which case it's included, or it's not, in which case it isn't, but we don't record a sentinel `None` value we'd later discard.

view/src/service.rs
```diff
@@ -256,19 +256,22 @@ impl ViewProtocolService for ViewService {
...
                 }
             }
         }
 
+        // TODO: query for advice notes
+        let advice_notes = Default::default();
```
hdevalence, line 269: Followup: support advice notes in view server's `Perspective` implementation.

This is required to make viewing swap claims work properly. Currently we don't add in advice notes, and so all `SwapClaimView`s are `Opaque`, even if the user would actually be able to see them.

view/src/storage.rs
```diff
@@ -235,6 +235,37 @@ impl Storage {
...
+            .fetch_optional(&pool)
+            .await?
+            {
+                return Ok(record);
+            }
+
+            if !await_detection {
+                return Err(anyhow!("swap commitment {} not found", swap_commitment));
+            } else {
+                return Err(anyhow!("swap commitment await_detection not implemented"));
```
hdevalence, line 264: Followup: implement awaiting detection of swap commitments in the view server.

This should work (and be implemented exactly like) detection for note commitments.  We don't currently need this in `pcli`, though, because we always await transaction confirmation anyways, and at that point the swap commitment we care about for claiming will already be there. But in the future this might be useful for someone else who's writing a more complex interaction (e.g., a trading bot).

view/src/storage.rs
```diff
@@ -666,6 +697,97 @@ impl Storage {
...
+            asset_id,
+            blinding_factor,
+        )
+        .execute(&mut tx)
+        .await?;
+
+        tx.commit().await?;
+
+        Ok(())
+    }
```
hdevalence, line 732: Followup: commonize give_advice implementation with record_block

I made this a distinct method because it seemed useful to be able to expose externally in the future for rollup use cases (where there's some system, like Narsil, that's externally feeding advice to the view server).

But since we're duplicating the same insertion query in multiple places (here, and when recording a block, though there it's in the context of some bigger database transaction), it might be better to have a private `async fn record_note_inner(note: Note, dbtx: &mut _)` (not sure what type `dbtx` should have, but it should be the return of `self.pool.begin().await?`) that adds the note recording into a bigger db transaction.

Then `give_advice` can make a standalone dbtx and call `record_note_inner`, and `record_block` can use `record_note_inner` on the bigger transaction it constructs.

view/src/storage.rs
```diff
@@ -666,6 +697,97 @@ impl Storage {
...
+
+    /// Return advice about note contents for use in scanning.
+    ///
+    /// Given a list of note commitments, this method checks whether any of them
+    /// correspond to notes that have been recorded in the database but not yet
+    /// observed during scanning.
+    pub async fn scan_advice(
+        &self,
+        note_commitments: Vec<note::Commitment>,
+    ) -> anyhow::Result<BTreeMap<note::Commitment, Note>> {
```
hdevalence, line 742: This method is how the view server gets advice to use while scanning a compact block: it builds a list of all the rolled-up note commitments, then has the `Storage` cross-reference with known notes.

view/src/storage.rs
```diff
@@ -666,6 +697,97 @@ impl Storage {
...
+                        notes.blinding_factor
+                FROM notes
+                LEFT OUTER JOIN spendable_notes ON notes.note_commitment = spendable_notes.note_commitment
+                WHERE (spendable_notes.note_commitment IS NULL) AND (notes.note_commitment IN ({}))",
+                note_commitments
+                    .iter()
+                    .map(|cm| format!("x'{}'", hex::encode(cm.0.to_bytes())))
+                    .collect::<Vec<_>>()
+                    .join(", ")
+            )
```
hdevalence, line 761: Followup: document this SQL query and its rationale.

We want to only give advice about notes that are known (hence in `notes`) but not already recorded as spendable, in which case we shouldn't see them again in scanning anyways, and if we do, something bad is probably happening.  I didn't think it through carefully, but I was worried that since the `spendable_notes` table records our own scan results, we might open up the possibility of leaking information about our scan results to an adversary who sends us crafted `CompactBlock`, e.g., sending a compact block with a `StatePayload::RolledUp(target_commitment)` and then somehow observing that we'd detected it using our scan results as advice.

view/src/storage.rs
```diff
@@ -666,6 +697,97 @@ impl Storage {
...
+                    .expect("32 bytes"),
+            )?);
+            let blinding_factor = Fq::from_bytes(
+                row.get::<&[u8], _>("blinding_factor")
+                    .try_into()
+                    .expect("32 bytes"),
+            )?;
+
+            let note =
+                Note::from_parts(address, Value { amount, asset_id }, blinding_factor).unwrap();
```
hdevalence, line 783: Followup: refactor into a wrapper type with `sqlx::FromRow`

This is copied out of the `FromRow` impl for `SpendableNoteRecord`; we can't implement `FromRow` on `Note` directly, because of orphan rules, but we could probably clean this up with a newtype wrapper.

view/src/storage.rs
```diff
@@ -756,6 +870,9 @@ impl Storage {
             let position = (u64::from(note_record.position)) as i64;
             let source = note_record.source.to_bytes().to_vec();
 
+            // We might have already seen the notes in the form of advice,
+            // so we use ON CONFLICT DO NOTHING to skip re-inserting them
+            // in that case.
```
hdevalence, line 875: Since we now use the advice mechanism to pre-record notes, it's no longer the case that note insertions happen exactly once: we'll insert the note in the `notes` table when we get it as advice, and then attempt to insert again when processing a `SpendableNoteRecord` when we see it in scanning.  We just have to ignore the second insertion; we know the data can't be different because the note commitment is a hash of the entire note contents, so if the note commitment matches, so does the rest of the row.

view/src/storage.rs
```diff
@@ -816,6 +955,7 @@ impl Storage {
             )
             .fetch_optional(&mut dbtx)
             .await?;
+            // TODO: mark spent swaps as spent
```
hdevalence, line 958: Followup: mark claimed `Swap`s in the view server

Currently, although we have a `height_claimed` field in the `SwapRecord`, we never set it, and never mark the swap as claimed. We should use similar logic as for notes to update `SwapRecord`s with the claim height.

view/src/sync.rs
```diff
@@ -40,45 +44,67 @@ pub async fn scan_block(
...
+            StatePayload::RolledUp(commitment) => unknown_commitments.push(commitment.clone()),
+        }
+    }
+    // Having started trial decryption in the background, ask the Storage for scanning advice:
+    let mut note_advice = storage.scan_advice(unknown_commitments).await?;
+    for decryption in note_decryptions {
+        if let Some(note) = decryption.await.unwrap() {
+            note_advice.insert(note.commit(), note);
+        }
+    }
```
hdevalence, line 94: This is where the word "advice" comes from: the storage gives "scan advice" in the form of "known notes" that are then augmented with the results of trial decryption.

view/src/sync.rs
```diff
@@ -87,47 +113,80 @@ pub async fn scan_block(
...
-                        Some(record)
-                    } else {
-                        // Don't remember this commitment; it wasn't ours
-                        note_commitment_tree
-                            .insert(tct::Witness::Forget, note_commitment)
-                            .expect("inserting a commitment must succeed");
-
-                        None
-                    }
+        // TODO: this will desync on rolled up note commitments, rewrite using for loop technology
```
hdevalence, line 116:
```suggestion
```
already fixed

wallet/src/plan.rs
```diff
@@ -92,82 +83,6 @@ where
...
-        .payment_address(source_address.unwrap_or(0).into());
-
-    let mut planner = Planner::new(rng);
-    planner.fee(swap_fee);
-    planner.swap(input_value, into_denom, swap_claim_fee, claim_address)?;
-    planner
-        .plan(view, fvk, source_address.map(Into::into))
-        .await
-        .context("can't build send transaction")
-}
```
hdevalence, line 169: Part of #1396

wallet/src/plan.rs
```diff
@@ -306,6 +223,8 @@ where
     }
 
     Ok(plans)
+    */
+    unimplemented!("This function has not been implemented since we changed the swap mechanism to auto-claim in scanning.")
```
hdevalence, line 227: Followup: restore implementation of manually claiming unclaimed swaps

This is necessary if the command fails partway through for some reason. We may have to extend the view protocol to return a list of all unclaimed swaps to do so.  We should also restore the hook to claim unclaimed swaps in the `tx sweep` implementation, which is currently commented out.

wallet/src/plan/planner.rs
```diff
@@ -106,29 +106,13 @@ impl<R: RngCore + CryptoRng> Planner<R> {
...
-        .into();
-
+    pub fn swap_claim(&mut self, plan: SwapClaimPlan) -> &mut Self {
         // Nothing needs to be spent, since the fee is pre-paid and the
         // swap NFT will be automatically consumed when the SwapClaim action
         // is processed by the validators.
-        self.action(swap_claim);
+        // TODO: need to set the intended fee so the tx actually balances,
+        // otherwise the planner will create an output
+        self.action(plan.into());
```
hdevalence, line 115: Followup: ensure the `Planner` correctly makes use of the `SwapClaim` fee

Currently, I don't think the `Planner` is aware that the `SwapClaim` has a contribution to the transaction fee. Probably it should set the claim transaction's fee to the prepaid fee, if the fee is unset.

hdevalence:
I annotated some relevant points and identified follow-up issues, but otherwise I'm going to squash-merge this now so we can keep rolling forward.

PR #1739, opened 2022-12-12
penumbra: make `tonic` an optional dependency

### LINKED ISSUE #1717: Make `tonic` an optional dependency of `penumbra-proto`

The Zpoken team mentioned that the use of `tonic` in `penumbra-proto` is a problem, because Tonic depends on Tokio, and Tokio doesn't compile in wasm, but because `penumbra-proto` is a common dependency of many of our crates, this blocks using them from `wasm` contexts.

Make the tonic-specific parts of `penumbra-proto` behind an `rpc` feature flag (off by default), and make the crates that use RPC definitions enable the `rpc` flag.

We could make the `rpc` feature on by default, but making it off by default makes it more likely we'll notice if something breaks when not using that feature flag.

erwanor:
Fixes #1717  - uses [`tonic_build::client_mod_attribute`](https://docs.rs/tonic-build/latest/tonic_build/struct.Builder.html#method.client_mod_attribute), [`tonic_build::server_mod_attribute`](https://docs.rs/tonic-build/latest/tonic_build/struct.Builder.html#method.client_mod_attribute), to encapsulate tonic generated client/servers behind an `rpc` feature flag.

### DIFF
--- a/proto/Cargo.toml
+++ b/proto/Cargo.toml
@@ -10,7 +10,7 @@ decaf377-fmd = { path ="../decaf377-fmd" }
 decaf377-rdsa = { version = "0.5", git = "https://github.com/penumbra-zone/decaf377-rdsa" }
 bytes = { version = "1", features = ["serde"] }
 prost = "0.11"
-tonic = "0.8.1"
+tonic = { version = "0.8.1", optional = true }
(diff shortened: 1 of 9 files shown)


hdevalence (approved):
LGTM!

PR #1742, opened 2022-12-12
component(ibc): remove deprecated `prev_connection` logic

### LINKED ISSUE #1695: Restore the ability for an IBC connection attempt to use a previous connection
The recent releases of `ibc-rs` have deprecated usage of the [MsgConnectionOpenTry.previous_connection](https://docs.rs/ibc/latest/src/ibc/core/ics03_connection/msgs/conn_open_try.rs.html#50) identifier that we used to enable those messages to piggy-back off previously established connections (after proper validation).

We should investigate if it has been replaced with another way for a new connection to ask for reopening an older connection, and if so make sure that we port the current commented out code to use it.

- [x] Figure out what has replaced this field
- [x] If it has not been completely deprecated, find a way to restore the feature.

erwanor:
Fixes #1695

### DIFF
--- a/component/src/ibc/component/connection/execution.rs
+++ b/component/src/ibc/component/connection/execution.rs
@@ -53,12 +53,6 @@ pub mod connection_open_try {
             let new_connection_id =
                 ConnectionId::new(self.get_connection_counter().await.unwrap().0);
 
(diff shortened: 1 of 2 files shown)


erwanor:
Merging without review since this is just removing commented code.

Issue #1744, opened 2022-12-12
Slashing penalty not adequately specified in ChainParams


conorsch:
**Describe the bug**
Recent changes to ChainParams appear to have broken `pd testnet join` on preview.

**To Reproduce**
1. `pd testnet unsafe-reset-all`
2. `pd testnet join testnet-preview.penumbra.zone`
3. Observe error:

```
pd testnet join testnet-preview.penumbra.zone
2022-12-12T17:27:07.454822Z  INFO fetching genesis
Error: slashing_penalty_downtime_bps must be set in ChainParameters
```

**Expected behavior**
Can join testnet-preview fine and interact with preview network.

**Screenshots**
N/A

**Additional context**
Looks related to https://github.com/penumbra-zone/penumbra/pull/1721 ; some follow-up may be necessary.

conorsch:
Looks like preview was misconfigured to run the prior testnet tag; hat-tip to @erwanor for catching it. Pushed up a fix-it commit here: https://github.com/penumbra-zone/penumbra/commit/17a3267f63e1765d40cd5d3d071292b8bd4f7fbe monitoring deploy and will test again shortly.

conorsch:
That fixed it! I'm now able to join testnet-preview, because it's actually running the "main"-tagged container, as intended.

PR #1763, opened 2022-12-13
ci: fix cluster-based preview deployment

conorsch:
During the release of 038-kaylke, we declined to cut over to the cluster deployment pipelines, due to an obscure misconfiguration. The bug was subtle: the local container used for building the config files prior to deployment was properly referred to by its "main" tag, but without a `--pull=always` flag, stale containers were being used from the local cache. This caused old config-generation code to be used to create the genesis, which lacked updates for new slashing penalty logic.

The above is enough to unbreak `pd/tm` node functionality, but there's an additional tweak required to honor requests from client-side `pcli` over 8080/TCP: the `pd` service now binds to `0.0.0.0`, rather than localhost, ensuring that traffic to the NodePort for "fullnode" reaches pd as intended.

Makes a few other small changes, some pointed out by @aubrika during pairing on deploy:

  * remove AppVersion from Helm annotations, to avoid confusion
  * updates the bash parameter substitution to be more explicit about assignment
  * runs "shellcheck" and applies quoting where it was missed.
  * sprinkles in a few more test.sh examples for debugging

With these changes, there are now two k8s deployments active:

  * testnet-preview.penumbra.zone # preview
  * fullnode.testnet.penumbra.zone  # testnet 038-kalyke

Let's continue to bang on the preview deployment.

### DIFF
--- a/deployments/ci.sh
+++ b/deployments/ci.sh
@@ -16,14 +16,15 @@ set -euo pipefail
 # for the helm chart. N.B. these env vars are also configured
 # in GitHub Actions, so the values below may be out of date.
 WORKDIR="${WORKDIR:=$(pwd)/helm/pdcli}"
-IMAGE="${IMAGE:=ghcr.io/penumbra-zone/penumbra}"
-PENUMBRA_VERSION="${PENUMBRA_VERSION:=037-megaclite.1}"
-PENUMBRA_UID_GID="${PENUMBRA_UID_GID:=1000\:1000}"
-TENDERMINT_VERSION="${TENDERMINT_VERSION:=v0.34.23}"
-NVALS="${NVALS:=2}"
-NFULLNODES="${NFULLNODES:=1}"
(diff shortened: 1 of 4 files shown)


deployments/ci.sh
```diff
@@ -87,7 +88,10 @@ else
 fi
 echo "Generating new testnet files..."
 container_cli="$(get_container_cli)"
+# Silence shellcheck warning on 'preserve_chain_opt' being an empty string.
+# shellcheck disable=SC2086
 "$container_cli" run --user 0:0 \
+    --pull always \
```
conorsch, line 94: This one-line change addresses the out-of-sync chain data we were seeing on preview, even after resolution of #1744. Now, the remote container registry is always checked. This is especially necessary for using the "main" tag, which is frequently clobbered. (The weekly testnet tags, e.g. `038-kalyke`, are static.)

PR #1769, opened 2022-12-16
decouple IBCAction from specific proto format

avahowell:
previously, the IBCAction type was an explicit protobuf `oneof` that enumerated all of the specific IBC action types (CreateClient, UpdateClient, etc), binding to their specific proto version and forcing us to vendor IBC protobufs. This PR changes the IBCAction type to contain a `raw_action`, which is an `Any`. This way, we can rely solely on the decoding logic in the `ibc-proto` crate, instead of *both* the crate and the vendored protos. Also, this makes updates to the IBC protos less painful, since we only need to update our node-side decoding logic and not all of our clients.


hdevalence:
Looks like the `wkt-types` isn't working because it's deferring some codegen to build time, I guess it has to be bundled into the `tools/proto-compiler`?

avahowell:
I think we need to check in the generated types for `Any` etc in the proto compiler

avahowell:
Updated this by adding our own serde implementation instead of using `wkt-types`, rebased on to latest main, IBC test passing

hdevalence:
This subsumes #1768 right?

hdevalence (approved):

avahowell:
Yes

PR #1772, opened 2022-12-16
crypto: refactor gadgets

hdevalence:
(PR description edited by @redshiftzero)

This PR adds types for each R1CS variable we use in the ZK proofs. Instead of `FqVar` and `ElementVar`s, we now have specific types for each instance variable, i.e. `BalanceCommitmentVar`, `NoteCommitmentVar`, `NoteVar`, `ValueVar`, and so on. 

The API on those types mirrors their native equivalents, e.g. there is a `Note::commit` and a `NoteVar::commit`. 

In terms of organization, the R1CS types are moved next to their native types instead of being hidden away in the proofs module. In the future it may make sense to put them under a r1cs feature as we have done in the decaf377 and poseidon377 crates.


crypto/src/address/r1cs.rs
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    #[allow(dead_code)]
+    pub fn transmission_key(&self) -> ElementVar {
+        self.transmission_key.clone()
+    }
+
+    pub fn transmission_key_s(&self) -> FqVar {
+        self.transmission_key_s.clone()
+    }
```
hdevalence, line 29: Why do we want both of these? Since the `ElementVar` handles cached encodings internally now, could we just use `transmission_key()` everywhere?

crypto/src/address/r1cs.rs
```diff
@@ -0,0 +1,76 @@
...
+use ark_ff::PrimeField;
+use ark_r1cs_std::prelude::*;
+use ark_relations::r1cs::SynthesisError;
+use decaf377::{
+    r1cs::{ElementVar, FqVar},
+    Element, Fq,
+};
+
+pub struct AddressVar {
+    pub diversified_generator: ElementVar,
```
hdevalence, line 11: When working in the circuit, do we want to be working with the diversified generator `g_d` or the diversifier `d`? Do we need to do the `d -> g_d` map in the circuit?

crypto/src/asset/amount.rs
```diff
@@ -30,6 +34,34 @@ impl Amount {
...
+impl AllocVar<Amount, Fq> for AmountVar {
+    fn new_variable<T: std::borrow::Borrow<Amount>>(
+        cs: impl Into<ark_relations::r1cs::Namespace<Fq>>,
+        f: impl FnOnce() -> Result<T, SynthesisError>,
+        mode: ark_r1cs_std::prelude::AllocationMode,
+    ) -> Result<Self, SynthesisError> {
+        let ns = cs.into();
+        let cs = ns.cs();
+        let amount1 = f()?;
+        let amount: Amount = *amount1.borrow();
```
hdevalence, line 51: nit: since we're cloning the value, could this be
```suggestion
        let amount: Amount = *f()?.borrow();
```

PR #1774, opened 2022-12-16
ci: try using nextest in release mode

hdevalence:
Motivations:

- Property tests mean we spend a lot of time running code as well as compiling it, so optimizing it makes tests shorter overall.
- Cargo check isn't doing a lot of lifting here, since all the work it does is duplicated as part of a build, and it will error early if there's a problem.  Because of caching, it shouldn't be spending much time rebuilding dependencies anyways.
- Using nextest https://nexte.st/index.html should make test runs faster via parallelism.

### DIFF
--- a/.github/workflows/buf-pull-request.yml
+++ b/.github/workflows/buf-pull-request.yml
@@ -1,4 +1,4 @@
-name: Lint protobuf
+name: Protobuf
 on: pull_request
 jobs:
   lint:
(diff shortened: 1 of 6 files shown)


hdevalence:
It seems like this gives a reasonable speedup: 
Screen Shot 2022-12-16 at 4 23 31 PM
Screen Shot 2022-12-16 at 4 23 53 PM

PR #1789, opened 2023-01-05
narsil: stub out proto structures

hdevalence:


hdevalence:
Still TODO:

- Write up Narsil design + tracking issue
- Add proto messages for DKG setup phase

hdevalence:
That said, I think it'd be useful to merge this, so we get rendered docs

PR #1803, opened 2023-01-05
Align ActionView proto types with ActionView domain types

### LINKED ISSUE #1787: Refine `TransactionView` protobuf modeling using `oneof`s to match our Rust enums
Part of #1775 

From the blog:

> This means that rather than requiring every ecosystem tool to implement its own logic to decrypt and understand the effects of shielded transactions, tools can consume TransactionViews, without having to support Penumbra-specific cryptography. And, like the TransactionPlan, the TransactionView is also specified as a Protobuf, so it can be serialized and parsed by tooling in any language.

@hdevalence: Unfortunately this isn't really true at the moment, because the nice enums that save the user from having to interpret the meaning of the FooView are only in our Rust types, and not in the protos. In particular one anti-pattern that we've done is using optional proto fields as ad-hoc enums, like "if this field is missing then we treat it as this enum variant, otherwise that one". but those would be much better as proto `oneof`s, because, for instance, the typescript code generated from the protos won't have that logic in it.

This ticket is for repairing this misalignment of domain & proto types, to allow tooling to utilize the proto type without caring about the Rust types.

aubrika:
Closes #1787

### DIFF
--- a/proto/proto/penumbra/core/dex/v1alpha1/dex.proto
+++ b/proto/proto/penumbra/core/dex/v1alpha1/dex.proto
@@ -111,14 +111,34 @@ message SwapClaimPlan {
 }
 
 message SwapView {
-  dex.v1alpha1.Swap swap = 1;
-  dex.v1alpha1.SwapPlaintext swap_plaintext = 3;
+  message Visible {
+    dex.v1alpha1.Swap swap = 1;
+    dex.v1alpha1.SwapPlaintext swap_plaintext = 3;
+  }
+
+  message Opaque {
+    dex.v1alpha1.Swap swap = 1;
+  }
+
+  oneof swap_view {
+    Visible visible = 1;
+    Opaque opaque = 2;
+  }
 }
 
 message SwapClaimView {
-  dex.v1alpha1.SwapClaim swap_claim = 1;
-  crypto.v1alpha1.Note output_1 = 2;
-  crypto.v1alpha1.Note output_2 = 3;
+  message Visible {
+    dex.v1alpha1.SwapClaim swap_claim = 1;
+    crypto.v1alpha1.Note output_1 = 2;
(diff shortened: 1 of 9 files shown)


proto/proto/penumbra/core/dex/v1alpha1/dex.proto
```diff
@@ -111,16 +111,35 @@ message SwapClaimPlan {
...
+}
+
+message VisibleSwapView {
   dex.v1alpha1.Swap swap = 1;
   dex.v1alpha1.SwapPlaintext swap_plaintext = 3;
 }
 
+message OpaqueSwapView {
+  dex.v1alpha1.Swap swap = 1;
+}
```
hdevalence, line 127: This could be folded into the `SwapView` as a nested message, like

```
message SwapView {
  message Opaque { ... }
  message Visible { ... }
  oneof swap_view {
    Visible visible = 1;
    Opaque opaque = 2;
  }
}
```

hdevalence: there's an example of this here: https://buf.build/penumbra-zone/penumbra/docs/main:penumbra.custody.v1alpha1#penumbra.custody.v1alpha1.PreAuthorization

aubrika: that's what I was looking for, thank you!

aubrika: I'm still having trouble with the type Tetris here - I'm unsure how to refactor the from & try_from code to take the new proto implementation into account. @hdevalence given the above implementation of the proto you provided as an example, can you check out the conversions in OutputView and see if there is a better way to do those?

aubrika: Sorry, updated!

PR #1819, opened 2023-01-09
feat(crypto): switch `Amount` to 128 bits internally

erwanor:
Partial implementation of https://github.com/penumbra-zone/penumbra/issues/1378, this PR:

- make Amount use a u128 internally
- provide a getter Amount::value returning a u64
- adds the necessary conversion traits
- implements encoding a u128 into two u64s, and vice-versa

Previous #1571 - reopened with a clean remote

### DIFF
--- a/component/src/shielded_pool/note_manager.rs
+++ b/component/src/shielded_pool/note_manager.rs
@@ -62,7 +62,7 @@ pub trait NoteManager: StateWrite {
         let encrypted_note = note.encrypt();
 
         // Now record the note and update the total supply:
-        self.update_token_supply(&value.asset_id, i64::from(value.amount))
+        self.update_token_supply(&value.asset_id, value.amount.value() as i64)
             .await?;
         self.add_state_payload(StatePayload::Note {
             note: EncryptedNote {
--- a/crypto/src/asset/amount.rs
+++ b/crypto/src/asset/amount.rs
@@ -48,7 +48,7 @@ impl TryFrom<std::string::String> for Amount {
     type Error = anyhow::Error;
 
     fn try_from(s: std::string::String) -> Result<Self, Self::Error> {
-        let inner = s.parse::<u64>()?;
+        let inner = s.parse::<u128>()?;
         Ok(Amount { inner })
     }
 }
--- a/crypto/src/asset/amount.rs
+++ b/crypto/src/asset/amount.rs
@@ -53,9 +53,71 @@ impl TryFrom<pb::Amount> for Amount {
     fn try_from(amount: pb::Amount) -> Result<Self, Self::Error> {
         let lo = amount.lo as u128;
         let hi = amount.hi as u128;
+        // We want to encode the hi/lo bytes as follow:
+        //            hi: u64                          lo: u64
+        // ┌───┬───┬───┬───┬───┬───┬───┬───┐ ┌───┬───┬───┬───┬───┬───┬───┬───┐
+        // │   │   │   │   │   │   │   │   │ │   │   │   │   │   │   │   │   │
+        // └───┴───┴───┴───┴───┴───┴───┴───┘ └───┴───┴───┴───┴───┴───┴───┴───┘
+        //   15  14  13  12  11  10  9   8     7   6   5   4   3   2   1   0
+        //
+        // To achieve this, we shift hi 8 bytes to the right, and then add the
+        // lower order bytes (lo).
 
-        let inner: u128 = u128::from_be(hi) + lo;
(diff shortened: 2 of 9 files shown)


aubrika (approved):
LGTM! I'll remember the amount.rs TryFrom impl in case we ever do need to store u128s in SQLite as two u64 columns, because we'll need to do that same kind of math both directions on read and write. 😎

component/src/shielded_pool/note_manager.rs
```diff
@@ -66,7 +66,7 @@ pub trait NoteManager: StateWrite {
         let encrypted_note = note.encrypt(&esk);
 
         // Now record the note and update the total supply:
-        self.update_token_supply(&value.asset_id, i64::from(value.amount))
+        self.update_token_supply(&value.asset_id, value.amount.value() as i64)
```
hdevalence, line 69: Followup: not in this PR, but we have some cleanup to do around the token supply tracking -- it should use `Amount`s, but also, we should check that whenever we convert tokens from one type to another, we correctly update the supply. This will be tricky for cases like swaps, unbonding, etc., where the chain might not be able to perfectly track that information, and we need to do some design thinking about what the implications of that are...

crypto/src/asset/amount.rs
```diff
@@ -117,85 +211,110 @@ impl From<Amount> for Fq {
...
 }
 
 impl From<[u8; 8]> for Amount {
     fn from(bytes: [u8; 8]) -> Amount {
         Amount {
-            inner: u64::from_le_bytes(bytes),
+            inner: u64::from_le_bytes(bytes) as u128,
         }
     }
 }
```
hdevalence, line 307: similarly here -- do we want to be deserializing amounts in this way, rather than as protos?

crypto/src/balance/imbalance.rs
```diff
@@ -37,20 +37,20 @@ impl Add for Imbalance<NonZeroU64> {
             (Imbalance::Required(r), Imbalance::Required(s)) => {
                 if let Some(t) = r.get().checked_add(s.get()) {
                     Some(Imbalance::Required(
-                        NonZeroU64::new(t).expect("checked addition of nonzero u64 never is zero"),
+                        NonZeroU128::new(t).expect("checked addition of nonzero u64 never is zero"),
```
hdevalence, line 40: should be "nonzero u128", missed in renaming?

component/src/shielded_pool/note_manager.rs
aubrika: Is there already a related issue for this? Sounds like it could be several tickets

hdevalence: No, we should create one as a followup, and the first subtask should be to start with writing down every place where we create or destroy tokens, so we can think through those issues.

crypto/src/asset/amount.rs
```diff
@@ -117,85 +211,110 @@ impl From<Amount> for Fq {
...
+                "could not deserialize Amount: input vec is not 16 bytes"
+            ));
+        }
+        let mut array = [0u8; 16];
+        array.copy_from_slice(&bytes);
+        Ok(Amount {
+            inner: u128::from_le_bytes(array),
+        })
+    }
+}
```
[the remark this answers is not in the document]
hdevalence: Ahhh, the proto format is variable-length, but we need a fixed-length encoding for use in ciphertexts.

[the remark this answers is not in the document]
hdevalence: Maybe, in order to avoid confusion, rather than having a `From/TryFrom` pair here, we could have ordinary methods on the `Amount` type like

```
Amount::to_fixed_len_bytes(&self) -> [u8; 16]
Amount::from_fixed_len_bytes(&[u8]) -> Result<Self, _>
```
so that it's clear what the distinction is from ordinary serialization, and we can write a comment on the methods.

component/src/shielded_pool/note_manager.rs
aubrika: #1823

crypto/src/balance/imbalance.rs
erwanor: nice catch, yup

crypto/src/asset/amount.rs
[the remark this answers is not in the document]
erwanor: +1, the `Vec<u8>` conversions weren't actually used anywhere, so I removed them and use `to_le_bytes`/`from_le_bytes` as passthrough to `u128` instead to take care of the `Note` usecase

hdevalence (approved):

PR #1821, opened 2023-01-09
pd: change testnet join to use peers' listen addresses

hdevalence:
`pd testnet join` connects to a node of a running network and pulls information from it to generate configs for a node that runs as part of that network.  It needs to do this, because Tendermint's P2P layer is authenticated, so at some point a node needs to obtain the fingerprint of at least one other peer. However, if we only add the node we're scraping as a peer, then we might not be able to join the network if that peer has too many connections open.

To prevent this, we added logic to `pd testnet join` that scrapes the node's `net_info` RPC endpoint to obtain a list of bootstrap peers.  However, in the first version of that code, we used the `remote_ip`, which might not be the externally routable address.  This commit changes the `testnet join` logic to scrape the self-reported `listen_addr` instead, and then filters out addresses that probably aren't valid, such as loopback addresses, private addresses, the unspecified address, etc.

The code is a little messy, because it aims to minimally change the existing logic; it could be cleaned up if the `testnet join` logic were cleaned up and moved out of `main.rs`.

Longer-term, an even better solution would be to configure the "bootstrap" node in testnet deployments as a seed node, but this is a bigger change, and this fix is useful anyways.

### DIFF
--- a/pd/src/main.rs
+++ b/pd/src/main.rs
@@ -341,7 +341,11 @@ async fn main() -> anyhow::Result<()> {
                 .unwrap_or_default();
 
             let mut peers = Vec::new();
-            peers.push((node_id, node));
+            if node.contains(':') {
+                peers.push((node_id, node));
+            } else {
+                peers.push((node_id, format!("{}:26656", node)));
+            }
             tracing::info!(?peers);
 
             for raw_peer in net_info_peers {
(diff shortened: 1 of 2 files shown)


hdevalence:
We don't really have a way to write tests for this because the code is all in `main.rs`, but that's a bigger cleanup as noted in the pr, but, locally, i did

- using `pd testnet join testnet-preview.penumbra.zone` to check that i could stand up a node
- then, once it was running (`pd start` and `tendermint start`), i did `pd testnet --testnet-dir /tmp/pd testnet join testnet-preview.penumbra.zone` to run it again and check that my (non-externally routable) node was filtered out of configs, even though it is online as so appeared in `net_info`

conorsch:
After testing locally, and comparing results with a `pd` built from `main` branch, I can confirm that the unwanted local addresses are no longer added to the local TM config. That's great. Then, when starting up a node locally based on that config, I'm able to peer and sync with blocks, as expected. Querying `localhost:26657/net_info` shows no local IPs at all, which is excellent.

conorsch (approved):

PR #1833, opened 2023-01-10
ci: reuse existing PL validator definition

### LINKED ISSUE #1832: k8s: default validator config should use PL definition
The new k8s deployments use ad-hoc validator identities for the deployment. We've long had a blessed validator config that we reuse on every testnet; let's use that for k8s, too. It's not strictly necessary, but for continuity's sake and legibility on other projects like the block explorer, it will be good to see the familiar human-readable fields for the genesis validator.

In terms of implementation, we'll need to lower the number of validators in the k8s deployment from 2 -> 1, but that's worthwhile. We can customize the validator generation logic to contain metadata later if when we want to grow the deployment.

conorsch:
The config generation logic for k8s rightly supports any number of validators, but our `pd testnet generate` logic assumes there's only one validator. In order to retain the human-readable public validator identity, we'll dial down NVALS to 1 for now. We can circle back later to build out the `pd` generate logic to support multiple identities with autoincrementing values.

Closes #1832.


hdevalence:
Having multiple validators seems somewhat important; per https://github.com/penumbra-zone/penumbra/issues/1832#issuecomment-1377686148 I think it might be more important than having human-readable validator metadata. Is there another way we could solve that problem?

conorsch:
This is very much a nice-to-have, and doesn't need to be part of the new infra from the get-go. I'll close this for now, but leave the ticket open, and we can revisit at a more opportune time.

PR #1835, opened 2023-01-10
crypto: implement rseed mechanism

### LINKED ISSUE #1688: Use rseed mechanism
Placeholder issue to use the rseed mechanism as described as https://zips.z.cash/zip-0212, essentially:

We choose a random rseed.
We derive using a PRF the esk from this rseed.
We derive using a PRF the note commitment randomness from this rseed.
The note plaintext includes the rseed instead of the note blinding factor. 

During note decryption the recipient will check epk = [esk] g_d and discard the note if this check fails. 

This will obviate the ephemeral pubkey integrity check (epk = [esk] g_d) check in our proofs

redshiftzero:
Closes #1688

This modifies our notes to use the mechanism described in [ZIP212](https://zips.z.cash/zip-0212):
* When generating notes we choose a 32-byte uniformly random value called the rseed.
* We derive using a PRF the esk from this rseed.
* We derive using a PRF the note commitment randomness from this rseed.
* The note plaintext includes the rseed instead of the note blinding factor.
* During note decryption the recipient will check epk = [esk] g_d and discard the note if this check fails.
* We no longer need to do the ephemeral pubkey integrity check (epk = [esk] g_d) check in our proofs as we do it at note decryption time.

### DIFF
--- a/chain/src/genesis/allocation.rs
+++ b/chain/src/genesis/allocation.rs
@@ -1,6 +1,4 @@
-use ark_ff::Zero;
-use decaf377::Fq;
-use penumbra_crypto::{asset, Address, Note, Value};
+use penumbra_crypto::{asset, Address, Note, Rseed, Value};
 use penumbra_proto::{core::chain::v1alpha1 as pb, Protobuf};
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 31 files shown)


redshiftzero:
Going to merge this so I can work on #1772 on top of the rseed changes, but please add any feedback and I will address in followups

PR #1840, opened 2023-01-12
k8s: set externalTrafficPolicy=Local

conorsch:
During load-testing, we observed that the k8s setup stopped peering once its peer count surpassed the number of external IPs assigned to the deployment. Log diving showed that peers were identified by the closest hop, i.e. NAT'd, addresses. To resolve, here we update the Service spec externalTrafficPolicy from "Cluster" (the default) to "Local". See docs at [0]. After deploying this change on a devnet, each remote peer successfully peered with all nodes in the deployment. The remote peers do not peer with each other, but further config changes will be required to support that functionality, based on our generated tm configs.
    
[0] https://kubernetes.io/docs/tasks/access-application-cluster/create-external-load-balancer/#preserving-the-client-source-ip

### DIFF
--- a/deployments/ci.sh
+++ b/deployments/ci.sh
@@ -21,7 +21,7 @@ PENUMBRA_VERSION="${PENUMBRA_VERSION:-main}"
 PENUMBRA_UID_GID="${PENUMBRA_UID_GID:-1000\:1000}"
 TENDERMINT_VERSION="${TENDERMINT_VERSION:-v0.34.23}"
 NVALS="${NVALS:-2}"
-NFULLNODES="${NFULLNODES:-1}"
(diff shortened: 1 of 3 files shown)


conorsch:
Point of process: because this change touches the `Service` definitions for the cluster, we must manually destroy the existing Services prior to merge. This is because the CI deployment workflow by default does not update Services, which allows the externalIPs (and therefore DNS A records) to persistent between deployments.

conorsch:
Talked this over with @agouin synchronously today, got a thumbs-up. I'm going to destroy the current preview services manually, then merge to re-deploy preview, and run the load-tests again.

Issue #1836, opened 2023-01-10
Dex Implementation Tracking Issue


zbuc:
1. [ ] [Composition of AMMs (for routing)](https://github.com/penumbra-zone/penumbra/issues/1776)
2. [x] [Implement `Path` type for routing](https://github.com/penumbra-zone/penumbra/issues/1777)
3. [x] [Design indices for liquidity positions](https://github.com/penumbra-zone/penumbra/issues/1778)
4. [ ] [Implement routing](https://github.com/penumbra-zone/penumbra/issues/1722)
5. [ ] [Implement execution](https://github.com/penumbra-zone/penumbra/issues/1724)
6. [ ] [Implement arbitrage](https://github.com/penumbra-zone/penumbra/issues/1779)
7. [ ] [Implement liquidity management APIs](https://github.com/penumbra-zone/penumbra/issues/1725)
8. [ ] protocol.penumbra.zone documentation

Future work
- [x] #1942 
- [x] #1950
     
Reference: [Dex planning document](https://hackmd.io/rsgSwz3BQKuL0GrSF0LWXA?view)

hdevalence:
Replacing by #1724

erwanor:
Deleting this issue because it contains no information that isn't tracked in the actual tracking issue and clutters search results, and bugs the automatic sub-issue <> tracking issue tracking of #1724

PR #1844, opened 2023-01-17
Move existing dex code to stubdex component, add stubbed dex component

zbuc:
Preparation work for #1836


component/src/dex/mod.rs
```diff
@@ -1,9 +1,9 @@
+//! The dex component contains implementations of the real Penumbra CPMM with token supplies based on liquidity provider interactions.
```
erwanor, line 1: nit: I think it's CFMM (constant-function) or CSMM (constant-sum)

erwanor (approved):

PR #1845, opened 2023-01-17
Proposal NFTs

### LINKED ISSUE #1347: Length limit on withdrawal reasons
They should be short.

### LINKED ISSUE #1793: Replace governance auth sigs with proposal NFTs
This should follow the design of unbonding tokens, roughly: there should be a unique denom for each proposal. The `ProposalSubmit` action consumes the proposal deposit amount of penumbra and produces a single unit of the denomination `proposal_N_voting` where `N` is the number of the proposal being made (if a proposal is made for the incorrect number, the tx fails). Then, a `ProposalWithdraw` action marks a proposal as withdrawn in the public chain state, consuming a single unit of `proposal_N_voting` and producing a single unit of `proposal_N_withdrawn`. Finally, a `ProposalDepositClaim` action (only valid when a proposal has not been vetoed, and only after voting concludes) consumes a single unit of `proposal_N_voting` or `proposal_N_withdrawn` and produces a single unit of one of `proposal_N_passed`, `proposal_N_failed`, or `proposal_N_vetoed`, as well as the proposal deposit amount for that proposal (we need to store historical proposal deposit amounts in case this changes by governance), provided the proposal was not vetoed. Stateful checks ensure that the input and output match the actual state transition available.

plaidfinch:
Resolves #1793 and resolves #1347.

Before merging:

- [x] Re-enable governance smoke test


chain/src/note_source.rs
```diff
@@ -9,7 +9,6 @@ pub enum NoteSource {
     Unknown,
     Genesis,
     FundingStreamReward { epoch_index: u64 },
-    ProposalDepositRefund { proposal_id: u64 },
```
plaidfinch, line 12: Because the chain no longer mints refunds, this case can be removed from the `NoteSource` enum.

component/src/governance/check.rs
```diff
@@ -61,8 +60,16 @@ pub mod stateless {
...
-        // All the checks are stateful.
+    pub fn proposal_withdraw(proposal_withdraw: &ProposalWithdraw) -> Result<()> {
+        // Enforce a maximum length on proposal withdrawal reasons; 80 characters seems reasonable.
+        const PROPOSAL_WITHDRAWAL_REASON_LIMIT: usize = 80;
+
+        if proposal_withdraw.reason.len() > PROPOSAL_WITHDRAWAL_REASON_LIMIT {
+            return Err(anyhow::anyhow!(
+                "proposal withdrawal reason must fit within {PROPOSAL_WITHDRAWAL_REASON_LIMIT} characters"
+            ));
+        }
```
plaidfinch, line 71: This resolves #1347, incidentally :)

component/src/governance/check.rs
```diff
@@ -112,6 +128,16 @@ pub mod stateful {
...
 
+        // Check that the proposal ID is the correct next proposal ID
+        let next_proposal_id = state.next_proposal_id().await?;
+        if proposal.id != next_proposal_id {
+            anyhow::bail!(
+                "submitted proposal ID {} does not match expected proposal ID {}",
+                proposal.id,
+                next_proposal_id,
+            );
+        }
```
plaidfinch, line 139: Previously the chain decided the proposal ID for the next proposal, because it was in charge: now, because we are self-minting the governance NFTs, we have to specify up-front the proposal we want to propose, and get rejected if it's not the next one in the succession of IDs.

component/src/governance/check.rs
```diff
@@ -291,4 +261,81 @@ pub mod stateful {
 
         Ok(())
     }
+
+    pub async fn proposal_withdraw(
+        state: &State,
+        proposal_withdraw: &ProposalWithdraw,
+    ) -> Result<()> {
+        // Any voteable proposal can be withdrawn
+        proposal_voteable(state, proposal_withdraw.proposal).await?;
```
plaidfinch, line 270: Any voteable proposal can be withdrawn, because if the transaction balances, that means you are spending the corresponding proposal NFT, which you could only have if you submitted the proposal. No further checks here are necessary because of this.

component/src/governance/check.rs
```diff
@@ -291,4 +261,81 @@ pub mod stateful {
...
+        proposal_deposit_claim: &ProposalDepositClaim,
+    ) -> Result<()> {
+        // Any finished proposal can have its deposit claimed
+        proposal_claimable(state, proposal_deposit_claim.proposal).await?;
+        // Check that the deposit amount matches the proposal being claimed
+        proposal_claim_valid_deposit(
+            state,
+            proposal_deposit_claim.proposal,
+            proposal_deposit_claim.deposit_amount,
+        )
```
plaidfinch, line 285: These two checks make sure that (1) the proposal is in the "finished" state (not "voting", "withdrawn" (which means "withdrawn but not yet finished voting"), or "claimed") precisely, and (2) the deposit amount matches the amount put in when the proposal was submitted. Very important second check!

component/src/governance/check.rs
```diff
@@ -291,4 +261,81 @@ pub mod stateful {
...
+                }
+                Withdrawn { .. } => {
+                    anyhow::bail!(
+                        "proposal {} has been withdrawn but voting has not concluded",
+                        proposal_id
+                    )
+                }
+                Finished { .. } => {
+                    // This is when you can claim a proposal
+                }
```
plaidfinch, line 305: The only success case is when the proposal is precisely in the finished-but-not-claimed state. It transitions out of this when you claim it.

component/src/governance/component.rs
```diff
@@ -34,8 +34,5 @@ impl Component for Governance {
         execute::enact_pending_parameter_changes(state)
             .await
             .expect("failed to enact parameter changes");
-        execute::apply_proposal_refunds(state)
-            .await
-            .expect("failed to apply proposal refunds");
```
plaidfinch, line 39: We no longer need to apply pending proposal refunds, because they are now self-minted.

component/src/governance/execute.rs
```diff
@@ -126,6 +111,40 @@ pub async fn validator_vote(
...
+    if let Some(proposal::State::Finished { outcome }) = state.proposal_state(*proposal).await? {
+        // This should be prevented by earlier checks, but replicating here JUST IN CASE!
+        if *resupplied_outcome != outcome.as_ref().map(|_| ()) {
+            anyhow::bail!(
+                "proposal {} has outcome {:?}, but deposit claim has outcome {:?}",
+                proposal,
+                outcome,
+                resupplied_outcome
+            );
+        }
```
plaidfinch, line 137: Should we recheck this here? Should we also recheck the proposal deposit amount matches?

component/src/governance/view.rs
```diff
@@ -166,6 +118,13 @@ pub trait StateWriteExt: StateWrite {
     /// Store a new proposal with a new proposal id.
     async fn new_proposal(&mut self, proposal: &Proposal) -> Result<u64> {
         let proposal_id = self.next_proposal_id().await?;
+        if proposal_id != proposal.id {
+            return Err(anyhow::anyhow!(
+                "proposal id {} does not match next proposal id {}",
+                proposal.id,
+                proposal_id
+            ));
+        }
```
plaidfinch, line 127: Proposals now self-contain their ID, so we need to check that this matches here before modifying the state.

crypto/src/asset/registry.rs
```diff
@@ -293,5 +293,14 @@ pub static REGISTRY: Lazy<Registry> = Lazy::new(|| {
...
         )
+        .add_asset(
+            // Note: this regex must be in sync with ProposalNft::try_from
+            "^proposal_(?P<data>(?P<proposal_id>[0-9]+)_(?P<proposal_state>voting|withdrawn|passed|failed|vetoed))$",
+            &[ /* no display units - nft, unit 1 */ ],
+            (|data: &str| {
+                assert!(!data.is_empty());
+                denom::Inner::new(format!("proposal_{}", data), vec![])
+            }) as for<'r> fn(&'r str) -> _,
+        )
```
plaidfinch, line 304: Adding five new families of assets, the regex: `proposal_([0-9]+)_(voting|withdrawn|passed|failed|vetoed)` representing the lifecycle of a proposal from the point of view of its proposer.

crypto/src/governance/proposal_nft.rs
```diff
@@ -0,0 +1,213 @@
...
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "voting" => Ok(State::Voting),
+            "withdrawn" => Ok(State::Withdrawn),
+            "vetoed" => Ok(State::Vetoed),
+            "failed" => Ok(State::Failed),
+            "passed" => Ok(State::Passed),
+            _ => Err(anyhow::anyhow!("invalid proposal token state")),
+        }
+    }
```
plaidfinch, line 51: These are very importantly inverses!

pcli/src/command/tx/proposal.rs
```diff
@@ -54,6 +54,22 @@ pub enum ProposalCmd {
...
+        /// The transaction fee (paid in upenumbra).
+        #[clap(long, default_value = "0")]
+        fee: u64,
+        /// The proposal id to claim the deposit for.
+        #[clap(long = "on")]
+        proposal_id: u64,
+        /// Optional. Only spend funds originally received by the given address index.
+        #[clap(long)]
+        source: Option<u64>,
+    },
```
plaidfinch, line 72: A new `pcli` command to claim a proposal deposit. This does the lookups behind the scenes to find out what outcome the proposal had and what the deposit is, so you don't have to look them up and feed them into the command (this can't be integrated into the planner as it currently is, because these queries require the specific query service).

pcli/tests/network_integration.rs
```diff
@@ -286,61 +286,60 @@ fn swap() {
...
+        .args([
+            "--data-path",
+            tmpdir.path().to_str().unwrap(),
+            "query",
+            "governance",
+            "list-proposals",
+        ])
+        .timeout(std::time::Duration::from_secs(TIMEOUT_COMMAND_SECONDS));
+    proposals_cmd.assert().success();
+}
```
plaidfinch, line 342: Governance smoke test is re-enabled now.

transaction/src/plan/auth.rs
```diff
@@ -14,23 +14,14 @@ impl TransactionPlan {
...
             let auth_sig = rsk.sign(&mut rng, effect_hash.as_ref());
             spend_auths.push(auth_sig);
         }
-        for withdraw_proposal_plan in self.proposal_withdraws() {
-            let rsk = sk
-                .spend_auth_key()
-                .randomize(&withdraw_proposal_plan.randomizer);
-            let auth_sig = rsk.sign(&mut rng, effect_hash.as_ref());
-            withdraw_proposal_auths.push(auth_sig);
-        }
```
plaidfinch, line 29: The authenticating data no longer needs to deal with the proposal withdrawals, because we don't use signatures here.

plaidfinch (commented):
This is an overview self-review to note some of the most salient parts of this PR.

Issue #1778, opened 2022-12-23
Design/Implement Indices for Liquidity Positions


zbuc:
* Non-consensus index by price
    - Want to be able to query for cheapest position for some directional trading pair (check [DEX engine design doc](https://hackmd.io/jtoiMn0pQnSTZlOi2ig5Wg) for notes)
    - Want a big-endian key encoding so we can do range queries that return positions in order
    - Let’s maybe not bin these from the start, but plan to eventually, since the routing and execution phases have different semantics
        - Rabbit holes: How do we adjust bins by adding/removing positions? Are those operations numerically stable? Do we have error propagation, or do we have to rebuild the bins every time positions change?
* Consensus indices to track all positions
    - Unclear what these should look like
    - Probably want a stub `PositionManager` similar to `NoteManager` to encapsulate state-related functions and enforce consensus logic and indexing etc. (#1725)
* Implementation of stub action handlers for `Position` actions (#1725)

PR #1849, opened 2023-01-18
Implement big-endian fixed point encoding

zbuc:
Preliminary work for the keys to be used in #1778

### DIFF
--- a/crypto/src/dex/lp/trading_function.rs
+++ b/crypto/src/dex/lp/trading_function.rs
@@ -1,7 +1,10 @@
 use penumbra_proto::{core::dex::v1alpha1 as pb, Protobuf};
 use serde::{Deserialize, Serialize};
 
-use crate::{dex::TradingPair, Amount};
+use crate::{
+    dex::{fixed_encoding::FixedEncoding, TradingPair},
+    Amount,
+};
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(try_from = "pb::TradingFunction", into = "pb::TradingFunction")]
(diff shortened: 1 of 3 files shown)


crypto/src/dex/lp/trading_function.rs
```diff
@@ -52,11 +55,42 @@ impl Protobuf<pb::TradingFunction> for TradingFunction {}
...
+    /// Returns the effective price of the trading function.
+    /// Note: the float math is a placehodler
+    pub fn effective_price(&self) -> f64 {
+        self.gamma() * self.p.inner as f64 / self.q.inner as f64
+    }
+
+    /// Returns the fee of the trading function, expressed as a percentage (`gamma`).
+    /// Note: the float math is a placehodler
+    pub fn gamma(&self) -> f64 {
+        (10_000.0 + self.fee as f64) / 10_000.0
```
hdevalence, line 90: I think this should be `10000 - fee`, because `gamma` is multiplied into the amounts, so it should be in range `[0, 1]`, e.g., a 100bps = 1% fee should give `gamma = 0.99`

zbuc: I had it like that originally but it seemed weird that increasing the fee would decrease the effective price.

hdevalence: It should definitely be `1 - f`, if increasing the fee decreases the effective price we've done a different mistake somewhere.  The percentage fee `gamma` is premultiplied by the input amounts `Delta_i`, to account for the LP position taking some fraction of the input as a fee: `Delta_i = (1 - f + f)*Delta = (1-f)*Delta + f*Delta = gamma*Delta + f*Delta`, where the left term is what's traded and the right term is what's taken as the fee.

erwanor: @zbuc I only skimmed the PR but note that the larger the `fee` the smaller the `gamma`:

e.g. `1% fee -> gamma = 0.99` or `30 bps fee -> gamma = 0.997` etc. 

So it's normal that with a larger gamma (i.e. smaller fee) the effective price would improve

zbuc: I guess I'm totally confused on what `effective_price` is expressing.

zbuc: It's the price that you can get in terms of asset `q` for asset `p` according to the trading function's parameters? So with a greater fee you get a worse price? That makes sense. I'll add a comment.

PR #1850, opened 2023-01-18
ci: make smoke tests fail fast

### LINKED ISSUE #1780: Faster smoke tests by adjusting run order
Quoth @hdevalence in Discord:

> wondering if we could get faster smoke tests by changing the order of steps in the workflow: currently we start the network, run it for a duration, then stop it, then restart it, then run integration tests. but alternatively, we could start the network, wait a short timeout (say 10 seconds) for it to come up, then run integration tests, then continue running the network for a duration

For reference, the smoke test run typically takes around 25m: https://github.com/penumbra-zone/penumbra/actions/workflows/smoke.yml Let's see if we can get that down to below 20m consistently. (We're currently waiting 5m in the smoke test.) 

> then continue running the network for a duration

If we background the testnet process, it's difficult to check on status. We must check the error code for the backgrounded process in order to conclude that the additional runtime didn't lead to something like a chain halt (#1706). Might wrap the smoke test logic in a script to make error checking easier.

conorsch:
The goal here is to get faster feedback from the smoke test suite. We now run the integration tests almost immediately after creating the network, allowing for a tighter feedback loop if errors are encountered, then continue running the network for a while longer, in an attempt to catch more subtle problems.

The wall-time here is mostly unchanged; the smoke test suite still takes >20m to complete. It will now exit much faster if errors are found, though, which is a win.

Closes #1780.

### DIFF
--- a/.github/workflows/smoke.yml
+++ b/.github/workflows/smoke.yml
@@ -31,19 +31,7 @@ jobs:
           chmod -R 777 ~/.penumbra/testnet_data
           docker-compose build
 
-      - name: Run testnet for smoke test duration.
-        run: timeout --preserve-status $TESTNET_RUNTIME docker-compose up --exit-code-from pd-node0
+      - name: Run the smoke test suite
+        run: scripts/smoke-test.sh
         env:
           TESTNET_RUNTIME: 5m
-
-      - name: Now start testnet in the background so we can run integration tests.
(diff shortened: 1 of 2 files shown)


conorsch:
Works well; will force-push a few more times to gather data on wall-time.

Issue #1777, opened 2022-12-23
Implement `Path` type for routing


zbuc:
* Maintains a list of asset IDs and an end-to-end trading function
* For `Path` of length 1, list of asset IDs: `[asset1, asset2]` and trading function is the original trading function
* We’ll want a `Path.extend(&self, trading_function: TradingFunction) -> Path` method that puts the next asset ID into the list and composes the trading functions together
* Stub struct
```rust
struct Path {
    asset_ids: Vec<asset::Id>,
    trading_function: TradingFunction,
}

impl Path {
    pub fn start(&self) -> asset::Id {
        // return first element
    }
    
    pub fn end(&self) -> asset::Id {
        // return last element
    }
}

impl PartialOrd for Path {
    // comparison should return `None` if start and end asset IDs don't match
    // otherwise compare by price
}
```

conorsch:
Closed via https://github.com/penumbra-zone/penumbra/pull/1854 !

Issue #1776, opened 2022-12-23
Implement AMM Composition


zbuc:
Composition of AMMs (for routing): if we have one AMM trading between asset 1 and 2, and one trading between 2 and 3, we should be able to compose the two to trade between 1 and 3

- [x] Add type information to trading struct (done in #1764)
- [x] Implement formulas in [DEX engine design doc](https://hackmd.io/jtoiMn0pQnSTZlOi2ig5Wg). 
- [x] #1960
- [x] #1933
- [ ] #1951

conorsch:
We merged #1854 this week; @erwanor can you provide a status update on this ticket?

conorsch:
^ as above, @erwanor. Unlikely to make testnet 43, but we'll see how much work is still outstanding.

erwanor:
I'm going to move composition onto `BareTradingFunction` and push the coefficient scaling code. Last, I will write some tests.

PR #1854, opened 2023-01-19
dex: implement `TradingFunction` composition

erwanor:
Partial implementation of #1777 and #1776


crypto/src/dex/execution.rs
```diff
@@ -0,0 +1,42 @@
+use crate::asset;
+use crate::dex::lp::TradingFunction;
+use crate::dex::trading_pair::DirectedTradingPair;
+use anyhow::Result;
+
+pub struct Path {
+    pair: DirectedTradingPair,
+    route: Vec<asset::Id>,
+    phi: TradingFunction,
```
zbuc, line 9: nit: Not sure if we want to use the greek characters from the design docs in code; we avoided that in the implementation of swaps in favor of more descriptive property names.

crypto/src/dex/execution.rs
```diff
@@ -0,0 +1,42 @@
+use crate::asset;
+use crate::dex::lp::TradingFunction;
+use crate::dex::trading_pair::DirectedTradingPair;
+use anyhow::Result;
+
+pub struct Path {
```
zbuc, line 6: We'll need a matching message defined in protobuf + the protobuf trait impls

Issue #1826, opened 2023-01-10
Move `pd testnet` code out of `main.rs` and clean it up


hdevalence:
**Is your feature request related to a problem? Please describe.**

The current `pd testnet` code is large and unwieldy, and it's difficult to maintain because it's all in the body of `pd`'s `main.rs`. See comments in #1821, for instance.

**Describe the solution you'd like**

Extract the code out of `main.rs` and do minimal cleanup (breaking stuff into functions, for instance)

conorsch:
Closed via https://github.com/penumbra-zone/penumbra/pull/1897

PR #1855, opened 2023-01-19
pd: use type-aware tendermint-config

### LINKED ISSUE #1827: Replace `tm_config_template.toml` with the `tendermint-config` crate

Currently, we have a hardcoded TOML file in `testnets/tm_config_template.toml` that we do string interpolation into to generate configs.  This is only really workable because we edit only one or two fields.

Use the [`tendermint-config` crate](https://docs.rs/tendermint-config/latest/tendermint_config/) to construct and manipulate a config object, then save it.

The `TendermintConfig` struct doesn't have a `Default` impl, but we can load the existing template (with format parameters removed) using `TendermintConfig::parse_toml` and then modify it.

We could alternatively use a general-purpose templating system. But this seems better, because the data we work with is nicely typed.

**Additional context**
Doing this allows us to add additional flags to the `pd testnet` commands that insert data into the tendermint configs (e.g., the listen address)

conorsch:
We now use the "tendermint-config" crate to build our TM testnet config with types. We preserve the hardcoded TOML template as a baseline config, then substitute values dynamically at runtime. There are some superficial changes:

  * final tm config has no comments
  * peer strings are formatted with "tcp://" prefix

but those are well worth the additional assurances we get from using a typed config object.

Closes #1827. Towards #1826.

### DIFF
--- a/pd/src/testnet.rs
+++ b/pd/src/testnet.rs
@@ -16,11 +16,14 @@ use std::{
     fmt,
     fs::{self, File},
     io::{Read, Write},
+    net::{IpAddr, SocketAddr},
     path::PathBuf,
     str::FromStr,
 };
-use tendermint::{node::Id, Genesis, PrivateKey};
-use tendermint_config::{NodeKey, PrivValidatorKey};
+use tendermint::{node::Id, Genesis, Moniker, PrivateKey};
+use tendermint_config::{
+    net::Address as TendermintAddress, NodeKey, PrivValidatorKey, TendermintConfig,
+};
 
 /// Methods and types used for generating testnet configurations.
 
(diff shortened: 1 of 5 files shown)


pd/src/testnet.rs
```diff
@@ -42,28 +45,101 @@ pub fn parse_validators(input: impl Read) -> Result<Vec<TestnetValidator>> {
...
+        .json::<serde_json::Value>()
+        .await?
+        .get("result")
+        .and_then(|v| v.get("peers"))
+        .and_then(|v| v.as_array())
+        .cloned()
+        .unwrap_or_default();
+
+    let mut peers = Vec::new();
+    for raw_peer in net_info_peers {
```
conorsch, line 104: Open to feedback on this for-loop. Should this logic be separated out into a function? That'd allow us to `.collect` on the results more tersely. Maybe as a TryFrom trait that allows creating a `TendermintAddress` from a `serde_json::Value`?

pd/src/testnet.rs
```diff
@@ -42,28 +45,101 @@ pub fn parse_validators(input: impl Read) -> Result<Vec<TestnetValidator>> {
...
+/// values via a typed interface, and rerender as TOML.
+pub fn generate_tm_config(
+    node_name: &str,
+    peers: Vec<TendermintAddress>,
+) -> anyhow::Result<String> {
+    tracing::debug!("List of TM peers: {:?}", peers);
+    let moniker: Moniker = Moniker::from_str(node_name)?;
+    let mut tm_config =
+        TendermintConfig::parse_toml(include_str!("../../testnets/tm_config_template.toml"))
+            .context("Failed to parse the TOML config template for Tendermint")?;
```
hdevalence, line 58: Not worth changing here, but this could just as well be an `.expect`, since the `tm_config_template.toml` is compile-time data and so being unable to parse it is an invariant violation.

pd/src/testnet.rs
hdevalence: I think it's probably OK, if we were going to clean it up I think a potentially better route would be to use the `tendermint-rpc` crate, which will give us [parsed data](https://docs.rs/tendermint-rpc/latest/tendermint_rpc/endpoint/net_info/index.html).

hdevalence (approved):
LGTM!

PR #1856, opened 2023-01-19
test: ensure consistent behavior in debug and release mode

redshiftzero:
I just noticed while testing https://github.com/penumbra-zone/decaf377/pull/42 against latest penumbra that one test (`spend_proof_verification_diversified_address_integrity_failure`) had inconsistent behavior in debug and release mode: in release mode the test passed (hence CI was green), but in debug mode the test panicked, due to a [debug assertion in ark-groth16](https://github.com/arkworks-rs/groth16/blob/master/src/prover.rs#L253) that checks the constraint system is satisfied. This PR modifies the test to check the constraint system is unsatisfied, as it cannot be if the diversified address is incorrect at proving time (the case for the test in question).

### DIFF
--- a/crypto/src/proofs/groth16.rs
+++ b/crypto/src/proofs/groth16.rs
@@ -272,6 +272,7 @@ mod tests {
 
     proptest! {
             #![proptest_config(ProptestConfig::with_cases(2))]
+            #[should_panic]
         #[test]
         /// Check that the `SpendProof` verification fails when the diversified address is wrong.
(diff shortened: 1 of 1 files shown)


hdevalence:
What's the pro/cons of checking that the constraint system is satisfied, versus doing proving and then verification?

redshiftzero:
If we do proving and verification we get the following behavior:
* In debug mode, we get a panic due to hitting the [debug assertion](https://github.com/arkworks-rs/groth16/blob/master/src/prover.rs#L253) due to the constraint system being unsatisfied. This assertion is from an upstream dependency so we can't change the debug assert to a regular assert.
* In release mode, even though the constraint system is unsatisfied, no panic occurs at proving time. Instead, we create a proof that will not verify.

I just realized the best of both worlds is to prove and then expect on the verification, that way either way we get a panic: at proving time in debug mode, at release time in verification mode. Updated the PR to do this.

redshiftzero:
(I'm going to merge this as the protobuf / lint job failure is unrelated)

PR #1858, opened 2023-01-19
Expose a new externally-driven interface to the `Planner`

plaidfinch:
This is how you use it: given a `planner: Planner<R>`, loop through all its requests, fulfilling them and continuing onwards until it indicates that it is finished:

```rust
let mut step = planner.step_by_step(&chain_params, &fmd_params, fvk, source)?;
loop {
    match step {
        Step::Finished(mut plan) => return Ok(mem::take(&mut plan)),
        Step::Request { request, respond } => {
            let response: Vec<SpendableNoteRecord> =
                unimplemented!("get response for request"); // <-- USER OF API: FILL THIS IN
            step = respond.with(response)?;
        }
    }
}
```

### DIFF
--- a/wallet/src/plan.rs
+++ b/wallet/src/plan.rs
@@ -18,7 +18,7 @@ use rand_core::{CryptoRng, RngCore};
 use tracing::instrument;
 
 mod planner;
-pub use planner::Planner;
+pub use planner::{step, Planner};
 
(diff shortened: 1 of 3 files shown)


plaidfinch:
Closing in favor of #1861.

Issue #1847, opened 2023-01-17
Add seed node to testnet deployment


conorsch:
**Is your feature request related to a problem? Please describe.**
We're still seeing peering trouble on the public testnet, capping out around ~50 peers. Let's ship a `seed` node as part of the deployment, which should help with peering with large (~100) sets of peers.

**Describe the solution you'd like**
There's a tm config option called `p2p.seed_mode`: https://github.com/penumbra-zone/penumbra/blob/40f9de041c2c5cddaedebc05cd3696271f9c8cc7/testnets/tm_config_template.toml#L260-L264

Let's set that to `true` on the first fullnode in the deployment, which already has the DNS A record for e.g. "testnet.penumbra.zone" pointing to it. Doing so should ensure that new peers receive a list of current addrs, but don't hold onto a persistent connection on the backend. 

**Describe alternatives you've considered**
We tried raising the `max_num_inbound_peers` to 50 across a 4-node deployment, hoping to raise the peer conn limit to ~200 (50 * 4). That doesn't solve the larger problem of wanting to make sure that new nodes even joining late in the testnet cycle can still peer. 

**Additional context**
See relevant [discussion in Discord](https://discord.com/channels/824484045370818580/915710988312002610/1064967068727910441). See also [Tendermint docs describing seed behavior](https://docs.tendermint.com/v0.34/tendermint-core/using-tendermint.html#seed).

conorsch:
Unfortunately the naive approach in #1859 doesn't work as we'd hoped; testing with 10 client nodes joining the preview deployment, they aren't peering well:

![preview-seed-not-working](https://user-images.githubusercontent.com/657862/213594483-4f896e37-3a41-4028-aeb0-fb79f99b259d.png)

I'd expect the peer count for each node to be ~3 (4 instances - 1 seed), but it's not. I'll do a bit more testing, but right now my inclination is to revert the commit for next week's testnet, and continue evaluating the seed/peer behavior separately.

conorsch:
Oops, my testing methodology was flawed: the client nodes were using outdated versions of `pd`. I deployed the seed_node=true change to a devnet environment, and peers running the latest `pd` from main worked just fine:

![npeers-with-seed](https://user-images.githubusercontent.com/657862/213801222-cdc3cd3d-396e-4759-94b3-bf55b07c6ee8.png)

Had to raise our resource quotas on GKE yet again in order to run the devnet, but now I can test these types of config changes before merging to preview.

[...]

PR #1859, opened 2023-01-19
feat: use seed node in k8s deployment

conorsch:
The first fullnode in the deployment will now have `p2p.seed_mode=true` set in its Terndermint config. The first fullnode already has its external IPs targeted by DNS A records like "testnet.penumbra.zone", so folks joining the testnet will still talk to the same service on the backend. The seed_mode=true change is intended to allow clients to connect and find peers, but then they'll be automatically disconnected immediately, ensuring that the max connections of 50 isn't exhausted by the first 50 peers joining.

Refs #1847.


conorsch:
I haven't tested this change yet; on merge it'll go to preview, and I'll test interactively there, and record results on #1847.

PR #1861, opened 2023-01-19
Allow the planner to be used across an FFI boundary, more simply

plaidfinch:
You just have to call `Planner::note_requests` and then use the results of those queries to call `Planner::plan_with_spendable_notes`.


hdevalence:
Looks like we'll want to disable the governance-related parts of the smoke test pending #1845

PR #1865, opened 2023-01-20
feat: support external addr for testnet join

### LINKED ISSUE #1848: Add testnet join flag for setting external_address
Right now, public testnet participants peer directly with the PL-run infra when running `testnet join`. By default, the nodes run by testnet partipicants won't peer with each other, because our tendermint config template sets `external_address` to any-address:

https://github.com/penumbra-zone/penumbra/blob/40f9de041c2c5cddaedebc05cd3696271f9c8cc7/testnets/tm_config_template.toml#L207-L215

Our `pd testnet join` logic carefully excises special addresses like "0.0.0.0" (#1821), so only the PL-run instances will have their addresses set in a new node's `p2p.seeds` section of the tm config.

Allow public testnet participants to opt-in to setting an `external_address` when joining, e.g.:

```
pd testnet join --external-address 1.2.3.4
```

Doing so will enable folks to declare their nodes as external accessible, and cause their external addresses to land in tm configs of subsequently joining nodes.

We've considered leveraging the `remote_ip` attribute, but decided against it (#1815). It makes more sense to permit testnet participants to opt in to serving as an externally accessible peer.

**Additional context**
While it's not a hard requirement, let's perform some housekeeping on the testnet code before tackling this change: #1826, #1827. Also, make sure to update the documentation with an explanation of how the `--external-address` flag functions.

conorsch:
Our auto-generated client config inherits "0.0.0.0" for `external_address` from the `laddr` setting. That works well enough to connect to the testnet, but doesn't support inter-peer gossip. Let's provide a flag to opt-in to p2p gossip. This setting isn't appropriate for all environments; the node admin must ensure that the relevant firewall configs are correct. Still, for those who run externally accessible nodes, this will permit them to declare themselves as viable peers for others.

Closes #1848.

### DIFF
--- a/docs/guide/src/pd/join-testnet.md
+++ b/docs/guide/src/pd/join-testnet.md
@@ -36,9 +36,19 @@ cargo run --bin pd --release -- testnet join
 
 This command fetches the genesis file for the current testnet, and writes
 configs to a testnet data directory (by default, `~/.penumbra/testnet_data`).
-
 If any data exists in the testnet data directory, this command will fail.  See
-the section below on resetting node state.
+the section above on resetting node state.
+
+If you want your node to accept inbound connections from other peers,
+you can set an external IP address during the join operation:
+
+```shell
+cargo run --bin pd --release -- testnet join --external-address 1.2.3.4
+```
+
+where `1.2.3.4` is the public IP address of the node you're running.
+Other peers will try to connect to your node over port 26656/TCP,
+so make sure your firewall allows that traffic in.
 
 ### Running `pd` and `tendermint`
 
--- a/docs/guide/src/pd/join-testnet.md
+++ b/docs/guide/src/pd/join-testnet.md
@@ -31,24 +31,21 @@ This will delete the entire testnet data directory.
 Next, generate a set of configs for the current testnet:
 
 ```shell
-cargo run --bin pd --release -- testnet join
+cargo run --bin pd --release -- testnet join --external-address IP_ADDRESS --moniker MY_NODE_NAME
 ```
+
+where `IP_ADDRESS` (like `1.2.3.4`) is the public IP address of the node you're running,
+and `MY_NODE_NAME` is a moniker identifying your node. Other peers will try to connect
+to your node over port 26656/TCP.
+
+If your node is behind a firewall or not publicly routable for some other reason,
+skip the `--external-address` flag, so that other peers won't try to connect to it.
+You can also skip the `--moniker` flag to use a randomized moniker instead of selecting one.
 
 This command fetches the genesis file for the current testnet, and writes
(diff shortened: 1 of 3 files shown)


conorsch:
This is now ready for review. The logic is pretty minimal: it doesn't perform any validation on the string (e.g. it'd be nice to filter out RFC1918 addresses from this field), just inserts it into the config. Since it's opt-in, I'm OK with that for a first pass. Will make sure to add some language to the release notes, pointing to the new docs, and encouraging folks to try it out.

docs/guide/src/pd/join-testnet.md
```diff
@@ -36,9 +36,19 @@ cargo run --bin pd --release -- testnet join
...
+If you want your node to accept inbound connections from other peers,
+you can set an external IP address during the join operation:
+
+```shell
+cargo run --bin pd --release -- testnet join --external-address 1.2.3.4
+```
+
+where `1.2.3.4` is the public IP address of the node you're running.
+Other peers will try to connect to your node over port 26656/TCP,
+so make sure your firewall allows that traffic in.
```
hdevalence, line 51: Could we restructure this, so that we start with the external address by default, and then tell people they can skip it if they are _not_ routable? What do you think of something like the following:

Next, generate a set of configs for the current testnet:

```shell
cargo run --bin pd --release -- testnet join --external-address IP_ADDRESS --moniker MY_NODE_NAME
```

where `IP_ADDRESS` (like `1.2.3.4`) is the public IP address of the node you're running and `MY_NODE_NAME` is a moniker identifying your node. Other peers will try to connect to your node over port 26656/TCP.

If your node is not publicly routable or behind a firewall, skip the `--external-address` flag, so that other peers won't try to connect to it. You can also skip the `--moniker` flag to use a randomized moniker instead of selecting one.

hdevalence:
> This is now ready for review. The logic is pretty minimal: it doesn't perform any validation on the string (e.g. it'd be nice to filter out RFC1918 addresses from this field), just inserts it into the config. Since it's opt-in, I'm OK with that for a first pass. Will make sure to add some language to the release notes, pointing to the new docs, and encouraging folks to try it out.

Since the config will get parsed by Tendermint, I'm not sure we need to do a lot of heavyweight validation -- someone can already edit their config.toml however we want, and I would think of this as just being a convenient way to insert values into it.

docs/guide/src/pd/join-testnet.md
conorsch: Totally, I like the readability with that approach. Editing...

hdevalence: sounds good, just flagging that i did a slight ninja-edit to the comment above (rewording behind a firewall vs publicly routable)

conorsch: Ninja edit now included!

PR #1881, opened 2023-01-24
Implement delegator voting with receipts

plaidfinch:


component/src/action_handler/actions/delegator_vote.rs
```diff
@@ -0,0 +1,26 @@
...
+    async fn check_stateless(&self, context: Arc<Transaction>) -> Result<()> {
+        check::stateless::delegator_vote(self, context)
+    }
+
+    async fn check_stateful<S: StateRead>(&self, state: Arc<S>) -> Result<()> {
+        check::stateful::delegator_vote(&state, self).await
+    }
+
+    async fn execute<S: StateWrite>(&self, state: S) -> Result<()> {
+        execute::delegator_vote(state, self).await
```
hdevalence, line 24: Is there a particular benefit to having these be individual bare functions, rather than just part of the `ActionHandler` impl?

hdevalence: e.g., would we ever want to call `execute::delegator_vote` outside of the context of `DelegatorVote::execute`?

component/src/governance/check.rs
```diff
@@ -87,6 +92,41 @@ pub mod stateless {
...
+        let effect_hash = context.transaction_body().effect_hash();
+        let anchor = context.anchor;
+
+        // 1. Check spend auth signature using provided spend auth key.
+        rk.verify(effect_hash.as_ref(), auth_sig)
+            .context("delegator vote auth signature failed to verify")?;
+
+        // 2. Check that the proof verifies.
+
+        // Verify the proof with this position as the start position:
```
hdevalence, line 122: What is "this position" here, as distinct from "start position"? Does this comment mean something different from "Verify the proof with the provided anchor and start position"?

component/src/governance/check.rs
```diff
@@ -205,6 +251,39 @@ pub mod stateful {
...
+                    unbonded_amount,
+                    nullifier,
+                    rk: _, // We already used this to check the auth sig in stateless verification
+                },
+            auth_sig: _, // We already checked this in stateless verification
+            proof: _,    // We already checked this in stateless verification
+        }: &DelegatorVote,
+    ) -> Result<()> {
+        proposal_voteable(&state, *proposal).await?;
+        // TODO: once epochs are non-constant, this has to be re-done to be a stateful lookup:
```
hdevalence, line 272: I don't think this refactor needs to be done before merging, but I think it could usefully be done on its own prior to switching to non-constant epochs, by (1) recording a `start_position` in the state for each proposal and (2) querying it here.

Then even before epochs are made non-constant, this part of the code won't depend on epoch duration math, it just has to do a lookup.

component/src/governance/check.rs
```diff
@@ -226,6 +305,91 @@ pub mod stateful {
...
+                    "nullifier {} was already spent at block height {} before proposal started at block height {}",
+                    nullifier,
+                    spend_info.spend_height,
+                    start_height
+                );
+            }
+        }
+
+        Ok(())
+    }
```
hdevalence, line 345: Would it make sense to commonize this with the shielded pool's `StateReadExt::check_nullifier_unspent`, if not exactly the same method (passing an `Option` to signal whether a start height is relevant seems ugly), at least in the same trait?

PR #1903, opened 2023-01-27
build(chain): drop tonic dep to support wasm

### LINKED ISSUE #1857: Make `tonic` an optional dependency for `penumbra-chain`

Currently, `penumbra-chain` can't be used in wasm, because it depends on `tonic`.

Determine why `penumbra-chain` depends on `tonic`, and make it optional.

conorsch:
Removes the `tonic` dependency from the `chain` crate, in order to support compiling it to wasm. The only use of `tonic` in `chain` was for a few error return types. Here we swap those specific errors for more general anyhow versions, and re-map them back to tonic::Status::unknown in the calling crate, `pd`. The use of `map_err` is suboptimal, but in the interest of unblocking external partners, seems a reasonable choice.

Closes #1857.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3409,7 +3409,6 @@ dependencies = [
  "serde",
  "sha2 0.9.9",
  "tendermint",
- "tonic",
  "tracing",
 ]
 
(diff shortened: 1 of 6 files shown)


plaidfinch:
This seems like the right tack to me. The other alternative would be to use conditional compilation to change the error types depending on whether the tonic dependency is enabled, and that seems like a hairy mess of potential sadness.

PR #1910, opened 2023-01-28
Uncouple serde minimal

### LINKED ISSUE #1898: First pass on serde json reorg
- [x] Determine which places in the code use the Serde formats in a load-bearing way, and, for each of these use cases, remove the dependency on the current Serde format
  - [x] Patch the `custody` crate's `Config` struct to serialize keys and addresses as bech32 strings.
  - [x] Patch the vestigial `Wallet` struct to serialize the spend key as a bech32 string.
  - [x] Patch `pcli validator` to use a new, TOML-based config file format for validator configuration.

Refs #1896

plaidfinch:
Resolves #1898.

This is missing updated documentation to reflect that validator definitions are toml instead of json now.

### DIFF
--- a/pcli/src/legacy.rs
+++ b/pcli/src/legacy.rs
@@ -44,8 +44,8 @@ pub struct ClientState {
 
 /// A legacy wallet file.
 #[derive(Debug, Clone, Serialize, Deserialize)]
-#[serde(from = "serde_helpers::WalletHelper")] // Deserialize from legacy format
-#[serde(into = "penumbra_custody::soft_kms::Config")] // Serialize into non-legacy format
+#[serde(from = "serde_helpers::WalletHelper")]
+#[serde(into = "serde_helpers::WalletHelper")]
 pub struct LegacyWallet {
     pub spend_key: SpendKey,
 }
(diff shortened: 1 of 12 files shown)


pcli/src/legacy.rs
```diff
@@ -44,8 +44,8 @@ pub struct ClientState {
 
 /// A legacy wallet file.
 #[derive(Debug, Clone, Serialize, Deserialize)]
-#[serde(from = "serde_helpers::WalletHelper")]
-#[serde(into = "serde_helpers::WalletHelper")]
+#[serde(from = "serde_helpers::WalletHelper")] // Deserialize from legacy format
+#[serde(into = "penumbra_custody::soft_kms::Config")] // Serialize into non-legacy format
```
hdevalence, line 48: I think we don't want to do this just yet, because we haven't figured out how we want to structure `pcli` configs going forward (where there might be multiple custody providers); if we stick with the existing structure rather than using the new `Config`, we can keep it unchanged until we know what we want the new format to be.

hdevalence: I think we actually don't want to be touching this at all, this isn't the part of the `pcli` code that's currently used to store the key material.

Issue #1896, opened 2023-01-26
Align `serde_json` output with protobuf JSON


hdevalence:
**Is your feature request related to a problem? Please describe.**

Currently, we maintain two serialization formats: the "canonical" data structures, defined as Protobufs, as well as a set of Serde-derived formats which are "shaped" by a long and complex set of custom type annotations attached to the `prost`-generated types.

The original motivation for this was that it let us make the Serde formats, particularly JSON, have customized serialization behavior to make them more human-readable. For instance, addresses are encoded using bech32 prefixes, rather than just raw bytes.

However, this decision is now causing us significant problems, as it undercuts the cross-language support which is the point of using Protobufs. The only way to read or write the JSON format is to use our Rust code, and implementations derived from the supposedly canonical Protobuf files can't interoperate with them.  In particular, this makes interop with web code difficult, because the JSON produced by the Buf typescript package (which receives data from a remote server) can't be passed to Rust code, which expects JSON in a different format.

**Describe the solution you'd like**

We should:

- [x] #1898
- [ ] Replace all of the custom Serde annotations in `tools/proto-compiler` with an appropriate invocation of `pbjson_build`.

Later:
- [ ] Add `Serialize + Deserialize` bounds to the `penumbra_proto::Protobuf` trait, for good measure.
- [ ] Remove the `penumbra_proto` crate's `serializers` module.

**Describe alternatives you've considered**

We could continue to maintain two different serialization formats, but this will cause greater and greater pain as time goes on. Aligning on the Protobuf JSON format is somewhat suboptimal in each case -- it'd be nice to have bech32 this, or hex-encoded that -- but in the bigger picture, it means there's just one data format we need to worry about. We have the opportunity to skip recreating our own Amino, let's take it.

**Additional context**

There are use cases where we'd like to have "human-readable" (or at least "developer-readable") output, and we should try to come up with ways to handle them.  The dependency-removal steps above are a useful minimal test case.  We discussed trying to create a second, "prettier" serialization format using some additional traits and generic impls, but concluded it wouldn't work.

conorsch:
> Determine which places in the code use the Serde formats in a load-bearing way, and, for each of these use cases, remove the dependency on the current Serde format:

@plaidfinch is working on this today and tomorrow, with a goal of merging at least that much for #1877. :crossed_fingers: The rest of the work we'll carry forward on future sprints.

PR #1911, opened 2023-01-28
Use `pbjson_build` to derive Serde implementations

hdevalence:
Builds on #1898 towards #1896


hdevalence:
I tested this against the current testnet, running `cargo run --release --bin pcli -- query validator definition penumbravalid1ymlgthzee7nzn7nhrz38u4p5a580uzdupu6dzckr56v63v4tguysnpz7fm`
and got
```json
{
  "identityKey": {
    "ik": "Jv6F3FnPpin6dxiiflQ07Q7+CbwPNNFiw6aZqLKrRwk="
  },
  "consensusKey": "0Qgt3eWN2QbW8zG0l5WInnxrbebbj774s1ozfddiCxU=",
  "name": "alterkahn",
  "description": "Lion 🦁 Stake validator back in town 👨🏻‍💻",
  "enabled": true,
  "fundingStreams": [
    {
      "address": "penumbrav2t1yamv3ht64np8hn0e2e6zsnt094kkepm9p97c0w34udsr3j2mmhdwwrajcrcsekehplh2zcpunvkkf2dmm8p6unl22fcffnw6qk3a278s3gkcacxyjyf2s36ykrlsjwjjm9d8t8",
      "rateBps": 550
    }
  ],
  "governanceKey": {
    "gk": "Jv6F3FnPpin6dxiiflQ07Q7+CbwPNNFiw6aZqLKrRwk="
  }
}
```
It's weird that the address is in there as a Bech32 string. It looks like in the `FundingStreams` proto it's defined as `string address`, what's up with that?

hdevalence:
The round-trip test failure in the tct crate seems related to https://github.com/bincode-org/bincode/issues/276 ; one long term possibility would be to patch `pbjson-build` as suggested towards the end of that discussion, but we can disable the test for now, I don't think it's important to support bincode.

PR #1914, opened 2023-01-30
pclientd refactor to support custody mode

aubrika:


view/src/bin/pviewd.rs
```diff
@@ -32,22 +35,38 @@ struct Opt {
...
+    Custody {
+        /// Bind the view service to this host.
+        #[clap(long, default_value = "127.0.0.1")]
+        host: String,
+        /// Bind the view gRPC server to this port.
+        #[clap(long, default_value = "8081")]
+        view_port: u16,
+        // The seed phrase providing spend capability
+        seed_phrase: String,
+    },
```
hdevalence, line 69: Rather than being able to start it in two different modes, I think we should have a config file generated by `init` that either has custody configuration or not, and then on start, run the custody service based on whether the config value is present.

hdevalence: It might be useful to start with a PR that adds a TOML-able `Config` structure; then we just have to change `init` to write an appropriate config and `start` to read it.

view/src/bin/pviewd.rs
```diff
@@ -32,15 +49,31 @@ struct Opt {
...
+    View {
+        /// Bind the view service to this host.
+        #[clap(long, default_value = "127.0.0.1")]
+        host: String,
+        /// Bind the view gRPC server to this port.
+        #[clap(long, default_value = "8081")]
+        view_port: u16,
+    },
+    /// Initialize the custody service with a seed phrase
+    Custody {
```
hdevalence, line 76: Rather than having subcommands on `start`, we should have the `view`/`custody` subcommands be on `init`, and have `start` run in whatever mode is specified in the config file.

aubrika: is there a use case where users will want to use a custody file containing a spend key to run the client in view mode? because if so, I figured it made sense to put the switch on `start` instead of `init`, but do we want to maybe add an explicit "start mode" field to the config file instead of just relying on the presence or absence of a spend key?

view/src/bin/pviewd.rs
```diff
@@ -1,15 +1,32 @@
...
+use std::{env, fs};
 use tonic::transport::Server;
+#[serde_as]
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub struct ClientConfig {
+    /// Optional KMS config for custody mode
+    pub kms_config: Option<soft_kms::Config>,
+    /// FVK for both view and custody modes
+    pub fvk: FullViewingKey,
+}
```
hdevalence, line 29: This should be expanded to move most of the current options out of the command-line flags and into the config. We can change the top-level `Opt` structure to take a `--home` argument (similar to `pd` and `tendermint` do) that sets the home directory with all of the configs and data, and load the other options (like the node URL, the host address, etc) from the config structure.

view/src/bin/pviewd.rs
```diff
@@ -77,28 +115,119 @@ async fn main() -> Result<()> {
...
+                        AuthPolicy::PreAuthorization(PreAuthorizationPolicy::Ed25519 {
+                            required_signatures: 1,
+                            allowed_signers: vec![pvk],
+                        }),
+                    ];
+                    Some(soft_kms::Config {
+                        spend_key: spend_key.clone(),
+                        auth_policy,
+                    })
+                }
```
hdevalence, line 148: This could probably be factored out into a standalone function that would build a default `soft_kms::Config` from a seed phrase

view/src/bin/pviewd.rs
hdevalence: No, I think they could do that by removing the custody config (or commenting it out)

view/src/bin/pviewd.rs
hdevalence: We should do this before merging, so we're not rearranging the binary's interface.

hdevalence:
We'll also want to rename `pviewd` to `pclientd`.

PR #1923, opened 2023-02-01
crypto: integrate Groth16 output and spend proofs

redshiftzero:
This PR replaces the transparent `OutputProof` and `SpendProof` with their ZK versions for testnet use. 

TODOs before merge:
- [ ] Debug smoke test failures here - after discussion we are going to debug these issues on `main` and resolve in smaller followups - see #2041
- [x] Debug why pcli is very slow - fixed (see comment [here](https://github.com/penumbra-zone/penumbra/pull/1923#issuecomment-1442679889))
- [x] Same for SpendProof
- [x] Delete transparent output proofs
- [x] Delete transparent spend proofs - Dropped this commit as the spend proof is used by the new delegator vote proof internally

Followup issues I will file after this is merged:
- [x] Reduce size of proving and verification keys: I disabled compression of the proving and verification keys which can be added back to reduce the size of the keys by ~2x - filed in https://github.com/penumbra-zone/penumbra/issues/2038
- [x] Cleanup r1cs module in the TCT crate: impl AllocVar, R1CSVar for the appropriate variables - filed in https://github.com/penumbra-zone/penumbra/issues/2040
- [x] Redo benchmarking / add benchmarks somewhere in this repo - filed in https://github.com/penumbra-zone/penumbra/issues/2039


hdevalence:
A fix for the CPU issue is in https://github.com/penumbra-zone/penumbra/pull/1923/commits/b6ade69cbed3ecb1cc0c25a6403473ad71143725 ; without disabling the `r1cs` tracing target, we'll spend an enormous amount of CPU and memory tracing execution of the constraint system.

Rebased on `main`.

PR #1926, opened 2023-02-02
feat(pcli): use https url by default for pd

conorsch:
Updated the option handling for pcli so that the `--node` arg is now a URL, rather than a hostname. Defaults to
`https://grpc.testnet.penumbra.zone`, which routes connections to pd's grpc service on the testnet deployment. Because it's a URL, HTTP is still supported, e.g. `http://testnet.penumbra.zone:8080`.

Requires disabling support for HTTP1 in the pd service. Was not able to identify any breakage from this change; external services like Buf Studio were still able to speak protobuf fine to the patched pd instance.

Includes a wrapper script that was helpful to me during debugging, so I'm tossing it in the repo.

### DIFF
--- a/pcli/src/opt.rs
+++ b/pcli/src/opt.rs
@@ -2,7 +2,7 @@ use crate::{
     box_grpc_svc::{self, BoxGrpcService},
     legacy, App, Command,
 };
-use anyhow::{Context, Result};
+use anyhow::Result;
 use camino::Utf8PathBuf;
 use clap::Parser;
 use directories::ProjectDirs;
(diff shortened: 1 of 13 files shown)


conorsch:
Draft because there's still some cleanup required of the option names, e.g. `PENUMBRA_NODE_HOSTNAME` -> `PENUMBRA_NODE_URL`.

pcli/src/opt.rs
```diff
@@ -123,9 +115,9 @@ impl Opt {
             let path = self.data_path.join(crate::VIEW_FILE_NAME);
             tracing::info!(%path, "using local view service");
 
-            let svc =
-                ViewService::load_or_initialize(path, fvk, self.node.to_string(), self.pd_port)
-                    .await?;
+            tracing::debug!("Connecting to URL: {}", &self.node);
+            let svc = ViewService::load_or_initialize(path, fvk, self.node.clone()).await?;
+            tracing::debug!("ALL DONE");
```
hdevalence, line 120: non-blocking nit: here, rather than BEGIN/END annotations, it'd probably be better to use a tracing span, so that all the events that happen in the context of `load_or_initialize` are annotated with the contextual information. this could either be done by constructing a span manually and using the `Instrument` trait to wrap the future before it's awaited, or by using the `#[instrument]` macro on `ViewService::load_or_initialize` (probably using `skip` on the `fvk` parameter).

view/Cargo.toml
```diff
@@ -54,6 +54,7 @@ async-trait = "0.1"
 tendermint = "0.26.0"
 tendermint-rpc = { version = "0.26.0", features = ["http-client"] }
 sha2 = "0.10.1"
+url = "*"
```
hdevalence, line 57: We probably want to specify a version here, right?

hdevalence (approved):
LGTM but we should specify a version of the `url` crate, to avoid surprise breakage later.

view/Cargo.toml
conorsch: Also done, pinning at `"2"`, so we'll continue to get minor and patch versions when updating.

PR #1919, opened 2023-01-31
fix fedo/rhel/centos

Tocard:
currently building on fedora, and this fit as well

PR #1936, opened 2023-02-03
docs: install git-lfs from package managers

### LINKED ISSUE #1928: docs: install git-lfs from package managers
In #1885 we updated the docs to include git-lfs, now a build-time dependency when compiling from source. Rather than link out to the tarball page, though, let's add it to the `apt` install command, along with the system dependencies for Linux. Relatedly, while we're in there:

* snip out the protobuf tooling on https://guide.penumbra.zone/main/pcli/install.html ; it's already mentioned in https://guide.penumbra.zone/main/dev/protobuf.html
* pull in https://github.com/penumbra-zone/penumbra/pull/1919 and test on multiple linux distros

Since #1885 is slated for release in Testnet 43 (#1902), these changes should be made by EOW.

conorsch:
Updates install instructions to avoid tarball-based git-lfs instructions. Also touches up the Fedora commands, ensuring they use `dnf` as `yum` is deprecated. Closes #1928. Supersedes and therefore closes #1919.

### DIFF
--- a/docs/guide/src/pcli/install.md
+++ b/docs/guide/src/pcli/install.md
@@ -26,10 +26,16 @@ run:
 sudo apt-get install build-essential pkg-config libssl-dev clang
 ```
 
-For a minimal Fedora image, you can run:
+For a minimal Fedora/Centos/Rhel image, you can run:
 
+This one to keep/update your system, this step is optional but recommended (this is important to keep system up to date)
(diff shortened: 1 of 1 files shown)


Tocard:
@conorsch you are wrong about something, dnf was supposed to replace yum, but that has been rollbacks.

PR #1956, opened 2023-02-06
staking: optimize validator uptime tracking

hdevalence:
This uses the changes to StateRead introduced in #1940 to concurrently fetch all data required for performing validator uptime tracking.  Note that we're not actually processing the _updates_ concurrently, because that would require mediating access to our shared mutable data.  Instead, we just concurrently fetch all of the data we'll need to process those updates, and then do the updates sequentially.

### DIFF
--- a/component/src/stake/component.rs
+++ b/component/src/stake/component.rs
@@ -158,8 +158,7 @@ pub(crate) trait StakingImpl: StateWriteExt {
                         self.get_block_height().await?,
                         self.signed_blocks_window_len().await? as usize,
                     ),
-                )
-                .await;
+                );
 
                 // Inform tendermint that the validator is now active.
                 let power = self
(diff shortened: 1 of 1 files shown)


component/src/stake/component.rs
```diff
@@ -635,16 +634,44 @@ pub(crate) trait StakingImpl: StateWriteExt {
...
         // Since we don't have a lookup from "addresses" to identity keys,
         // iterate over our app's validators, and match them up with the vote data.
-        for v in self.validator_list().await?.iter() {
-            let info = self
-                .validator_info(&v.identity_key)
-                .await?
-                .ok_or_else(|| anyhow::anyhow!("validator missing info"))?;
+        // We can fetch all the data required for processing each validator concurrently:
+        let mut js = JoinSet::new();
+        for v in self.validator_identity_list().await? {
```
hdevalence, line 639: The real optimization here is only iterating over the list of validator _identity_ keys, rather than full validator data; the concurrent fetching is just additional gravy.

hdevalence:
Not the most precise benchmark, but you can see the effect this has on the relative time cost of BeginBlock/EndBlock before
Screen Shot 2023-02-06 at 4 37 04 PM
after
Screen Shot 2023-02-06 at 4 37 45 PM

Issue #1884, opened 2023-01-24
pd crash loop if tendermint exits between init_chain and first block


conorsch:
**Describe the bug**
The `pd` application will get stuck in a failed state if tendermint sends an `init_chain` event to `pd`, but then exits before the first block is handled and submitted to the app. On subsequent service start, tendermint will again send an `init_chain` event, because it hasn't sent a block before, causing `pd`  to reject the second `init_chain` msg, crashing.

**To Reproduce**
1. `pd testnet unsafe-reset-all`
2. `pd testnet join`
3. start node by your favorite method
4. kill tendermint process (e.g. via `ctrl+c` if run interactively) quickly (~<2s) after starting
5. restart tendermint process, observe error msg in `pd` logs:

```
thread 'tokio-runtime-worker' panicked at 'init_chain must succeed: database already initialized', /home/penumbra/penumbra/pd/src/consensus/worker.rs:45:26
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: consensus worker terminated or panicked', /home/penumbra/.cargo/registry/src/github.com-1ecc6299db9ec823/tower-abci-0.2.0/src/server.rs:164:70
```

**Expected behavior**
pd should resume sipping events from tendermint and not crash. 

**Additional context**
We believe that the problem of tendermint not sending blocks correlates with inability to find peers. The longer peers cannot be found, the more likely this crash loop is to occur. The current best workaround is to `pd testnet unsafe-reset-all` and try joining again.

To solve the problem, we can avoid calling commit in the app code until the first block has been processed. Quoth @hdevalence:

> one possibility would be that we should not commit state changes after `InitChain`, and wait until we get the first `BeginBlock/DeliverTx/EndBlock/Commit` stanza for it to be committed

Currently we use a u64 as version, where the max value indicates an uninitialized state:

https://github.com/penumbra-zone/penumbra/blob/3fdb59a21eb2a8890c14d523308c79409ec0aeb8/storage/src/storage.rs#L118-L124

If we change that logic, we should be mindful of off-by-one errors in other parts of the code that reference of the version.

conorsch:
> To solve the problem, we can avoid calling commit in the app code until the first block has been processed.

Further discussion with Tendermint experts confirms this is the right approach. If Tendermint is halted early, and submits init_chain, the app should respond kindly.

[...]

PR #1959, opened 2023-02-07
Reconcile pd/tm handling of initchain requests

conorsch:
Tried using `[0u8; 32]` and `[255u8; 32]`, and Tendermint rejected both. Including a tiny repro script to aid in reproducing the issue and therefore evaluating potential fixes.

Towards #1884. Posting early for visibility.

### DIFF
--- a/scripts/test-1884.sh
+++ b/scripts/test-1884.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+# The purpose of this script is to replicate the failure conditions in GH1884.
+# It attempts to restart tendermint fast enough after first start that
+# the InitChain request gets resent on subsequent service start.
+set -euo pipefail
+set -x
+
+
+sudo systemctl disable --now penumbra tendermint
+
+# build new binaries
+cargo build --release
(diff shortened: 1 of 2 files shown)


scripts/test-1884.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+# The purpose of this script is to replicate the failure conditions in GH1884.
+# It attempts to restart tendermint fast enough after first start that
+# the InitChain request gets resent on subsequent service start.
+set -euo pipefail
+set -x
+
+
+sudo systemctl disable --now penumbra tendermint
```
hdevalence, line 9: Wait, do we have native systemd unit files now?

conorsch: Ha, no, I just use it myself because it's the most convenient option.

conorsch: Filed #1974 to improve the docs a bit. Closing this PR because I don't expect it to be merged; more research required on approach. Will continue discussion in #1884 in subsequent sprints.

PR #1963, opened 2023-02-07
jmt: use `jmt@main`

erwanor:


hdevalence:
Could we do a release instead of using a git dep? Why or why not?

erwanor:
@hdevalence yes good idea

PR #1970, opened 2023-02-08
perf: concurrent proving

### LINKED ISSUE #717: pcli: parallelize proof generation
The zk proving time is ~10s per action (based on benchmarking on the zk `OutputProof` in #665), but each proof is independent. So, once either #716 or #715 are implemented, we should parallelize proving in `pcli`.

redshiftzero:
Needed for #1923, will close #717

### DIFF
--- a/transaction/src/plan/build.rs
+++ b/transaction/src/plan/build.rs
@@ -163,4 +163,182 @@ impl TransactionPlan {
             binding_sig,
         })
     }
+
+    /// Build the transaction this plan describes while proving concurrently.
+    /// This can be used in environments that support tokio tasks.
+    pub async fn build_concurrent<R: CryptoRng + RngCore>(
+        self,
+        rng: &mut R,
+        fvk: &FullViewingKey,
+        auth_data: AuthorizationData,
+        witness_data: WitnessData,
+    ) -> Result<Transaction> {
+        // Do some basic input sanity-checking.
+        let spend_count = self.spend_plans().count();
+        if auth_data.spend_auths.len() != spend_count {
+            return Err(anyhow::anyhow!(
+                "expected {} spend auths but got {}",
+                spend_count,
+                auth_data.spend_auths.len()
+            ));
+        }
+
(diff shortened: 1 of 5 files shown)


transaction/src/plan/action/output.rs
```diff
@@ -61,6 +61,19 @@ impl OutputPlan {
...
+        &self,
+        ovk: &OutgoingViewingKey,
+        memo_key: &PayloadKey,
+    ) -> Output {
+        Output {
+            body: self.output_body(ovk, memo_key),
+            proof: self.output_proof(),
+        }
+    }
+
```
hdevalence, line 76: what's the reason to have a duplicate method here, rather than calling the existing method in an async context?

transaction/src/plan/build.rs
```diff
@@ -163,4 +163,186 @@ impl TransactionPlan {
...
+        // If the memo_key is None, then there is no memo, and we populate the memo key
+        // field with a dummy key.
+        for output_plan in self.output_plans().cloned() {
+            // Outputs subtract from the transaction's value balance.
+            synthetic_blinding_factor += output_plan.value_blinding;
+            let ovk = fvk.outgoing().clone();
+            let memo_key = memo_key.as_ref().unwrap_or(&dummy_payload_key).clone();
+            in_progress_output_actions.push(tokio::spawn(async move {
+                output_plan.output_concurrent(&ovk, &memo_key).await
+            }));
```
hdevalence, line 236: in re: https://github.com/penumbra-zone/penumbra/pull/1970/files#r1102024216 , couldn't this just be
```suggestion
            in_progress_output_actions.push(tokio::spawn(async move {
                output_plan.output(&ovk, &memo_key)
            }));
```

redshiftzero: ah! totally, it should be

redshiftzero:
Note that CI is failing here due to #1977 so this PR should be rebased to get a full CI run prior to merge

PR #1971, opened 2023-02-09
ci: reorganize deploy files into subdir

### LINKED ISSUE #621: repo cleanup: move deployment-related files into a `deploy/` subdirectory

conorsch:
Strictly moving things around in this commit, shoving docker/compose/scripts/grafana-config into "deployments/". Requires corresponding path changes to the CI workflows, and some docs touch-ups.
    
Closes #621.


conorsch:
Smoke test passes. Also made sure to kick off an ad-hoc container build run, to make sure the CI workflows are using the proper new filepaths. Success: https://github.com/penumbra-zone/penumbra/actions/runs/4129685083

conorsch:
One minor gripe encountered when testing the compose files in their new location; the network names are now different, so on first up, it failed with:

```
❯ docker compose -f docker-compose.yml -f docker-compose.prod.yml  -f docker-compose.override.yml up --abort-on-container-exit --remove-orphans
[+] Running 0/0
 ⠿ Network compose_localnet  Error                                                                                                                                                                                         0.0s
failed to create network compose_localnet: Error response from daemon: Pool overlaps with other one on this address space
```

We can prevent that from happening by explicitly naming the net, rather than allowing compose to infer a network name based on the parent directory (whence `penumbra_localnet` vs `compose_localnet`). Then even in the new location a previously existing network will be honored and used.

PR #1973, opened 2023-02-09
dex: bound LP positions to at most 112 bits

### LINKED ISSUE #1960: Restrict LP position and reserve amounts to 112 bits
It's much more convenient to do multi-word arithmetic on unsaturated coefficients, because it's possible to avoid intermediate carries. If we restrict LP position and reserve amounts to be at most 112 bits, we can represent them as two 56-bit limbs, with 8 bits of headroom for carries in intermediate computations.

Rather than implementing a whole new type for this, I think it'd be cleanest to start by just implementing a check function for LP positions as they're created.

erwanor:
Closes #1960

### DIFF
--- a/component/src/dex/position_manager.rs
+++ b/component/src/dex/position_manager.rs
@@ -1,5 +1,6 @@
 use anyhow::{anyhow, Result};
 use async_trait::async_trait;
+use penumbra_crypto::dex::lp::position::MAX_RESERVE_AMOUNT;
 use penumbra_crypto::dex::{
     lp::{
         position::{self, Position},
@@ -9,7 +10,6 @@ use penumbra_crypto::dex::{
 };
 use penumbra_proto::{DomainType, StateReadProto, StateWriteProto};
 use penumbra_storage::{StateRead, StateWrite};
-use penumbra_crypto::dex::lp::position::MAX_RESERVE_AMOUNT;
 
 use super::state_key;
 
--- a/component/src/dex/position_manager.rs
+++ b/component/src/dex/position_manager.rs
@@ -1,4 +1,4 @@
-use anyhow::Result;
+use anyhow::{anyhow, Result};
 use async_trait::async_trait;
 use penumbra_crypto::dex::{
     lp::{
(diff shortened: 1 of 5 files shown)


component/src/dex/position_manager.rs
```diff
@@ -41,6 +42,17 @@ pub trait PositionManager: StateWrite + PositionRead {
         position: Position,
         initial_reserves: Reserves,
     ) -> Result<LpNft> {
+        // We limit the sizes of reserve amounts to at most 112 bits. This is to give us extra
+        // headroom to perform intermediary calculations during composition.
+        // TODO: remove the extra casting once `Amount` gets full 128 bits support.
+        if initial_reserves.r1.value() as u128 > MAX_RESERVE_AMOUNT
+            || initial_reserves.r2.value() as u128 > MAX_RESERVE_AMOUNT
```
hdevalence, line 49: We'll want to check the position coefficients as well as the reserves.

In terms of structuring the checks, ideally we'd check this before starting execution, and then again as a sanity check here.  So I think it'd make sense to factor the check out into methods on `Position` and `Reserves`, maybe a `fn check_bounds(&self) -> Result<()>` on each, so we have the check defined only in one place.

transaction/src/action/position.rs
```diff
@@ -42,6 +42,22 @@ impl IsAction for PositionOpen {
...
+            Err(anyhow::anyhow!(format!(
+                "position reserves too big (limit: {MAX_RESERVE_AMOUNT})"
+            )))
+        } else {
+            Ok(Self {
+                position,
+                initial_reserves,
+            })
+        }
+    }
```
hdevalence, line 59: Possibly counterintuitively, but I think it might be better to not have this constructor: either we decide that we want to make the excess-size values totally unrepresentable, or we should check them in the place where the check is required. Having a check in this constructor might give a false sense of security, because it's not the only way that a `PositionOpen` can be constructed.

component/src/dex/position_manager.rs
erwanor: thanks, this results in a very legible `PositionManager::position_open`

erwanor:
Postponing this one until 45, because we're thinking about revamping completely how we do fixed point arithmetic.

Issue #1843, opened 2023-01-17
Restore metrics dashboard on k8s deployment


conorsch:
**Is your feature request related to a problem? Please describe.**
As part of the k8s migration (#1818), we lost automatic display of grafana metrics for the testnet. Let's add that feature into the cluster deploy logic.

**Describe the solution you'd like**
Port the pre-existing grafana setup so it's available on k8s. Subtasks include:

* [x] Ensure dashboards are working via docker-compose, as a baseline
* [x] Build a container image for grafana with the Penumbra dashboards included
* [x] Add prom/graf container spec to existing helm deployment logic
   *  [x] grafana dashboards should be bundled in container
   * [x] grafana.ini config file should be bundled in container
   * [x] grafana datasources config should be a helm template (for dynamic values of prom endpoint)
   * [x] prometheus scrape config should be helm template (for dynamic values of tm/pd endpoints)
   * [x] therefore, no custom container required for prom
* [x] Add ingress and DNS records per deployment (e.g. `grafana.testnet.penumbra.zone`)

**Additional context**
Once we've restored the dashboards, we can revisit the metrics tracked, and document a workflow for iterating on the configs over time., as part of the development cycle.

conorsch:
After discussion with @hdevalence, I'm prioritizing this work near-term above the ongoing HTTPS work. Will try to get a first pass on changes into the upcoming testnet, and continue expanding as outlined above.

conorsch:
Progress update. As of #1980, dashboards are working via `compose`: 

![metrics-compose-1](https://user-images.githubusercontent.com/657862/218210751-5ef1b0fe-2083-4a98-99c2-b5fef726ff45.png)

That instance is connected to "preview", hence few peers tracked. In #1981 there's a partial deployment spec, but still waiting on an HTTPS cert:

![gke-cert-provisioning](https://user-images.githubusercontent.com/657862/218212348-c6a2a131-9d56-48aa-92c8-155e4ae8527f.png)

I'll need that request to be fulfilled before I can test it interactively. I'll double check the logs, because I'm surprised it's taken this long already. The "devnet" deployment is an ephemeral environment that I can push to directly from CLI, and edit on the fly, without disrupting the longer-running preview and testnet deployments. 

So, we're quite close to barebones metrics making it into testnet 44, but the fallback will be the dashboards landing on preview early next week.

conorsch:
Ah, found it. Events show failed lookup on new backend svc:

```
[...]

PR #1981, opened 2023-02-10
metrics: display grafana dashboards on k8s

conorsch:
Exposes Grafana dashboards, unique per testnet deployment (i.e. "testnet-preview" separate from "testnet"). Includes no persistence to honor data retention for metrics, so data will be destroyed on every deploy. That's ideal for "testnet" but it will make preview quite short-lived. We'll continue to iterate on the config.

Refs #1843.


conorsch:
Merging, will monitor deploy to preview, and update ingress settings as needed.

PR #1983, opened 2023-02-11
Nestable state deltas

hdevalence:
One things we've been discussing recently is that the current approach of separating routing and execution might not be the right thing to do, because it means that we can’t cleanly account for what happens as we move around the liquidity graph — for instance, if we have cycles, intuitively, there shouldn’t be any problem, because once we go “around”, we end up in a different market state with different available liquidity, but this requires a way to track state changes without having to commit to them. The approach of just trying to compose together trading functions doesn’t really capture that, and also caused us to run into other problems.

Instead, it could be cleaner if we did a different approach, where we directly simulated potential executions, then picked the one we like, i.e., as we explore a path, we actually execute trades along that path, and then see how much output we get from that trade, and then later, when we find a path we like, we can pick its execution trace and apply it.  Our current storage system doesn't allow this, however, because our `StateTransaction` isn't nestable. 

This PR introduces a new `StateDelta` type, representing an arbitrarily-deeply nested stack of delta updates to an underlying state. This API allows exploring a tree of possible execution paths concurrently, before finally selecting one and applying it to the underlying state.

On creation, `StateDelta::new` takes ownership of a `StateRead + StateWrite` instance, acquiring a "write lock" over the underlying state (since `&mut S` is `StateWrite` if `S: StateWrite`, it's possible to pass a unique reference).

The resulting `StateDelta` instance is a "leaf" state, and can be used for reads and writes, following the some execution path.

When two potential execution paths diverge, `delta.fork()` can be used to fork the state update.  The new forked `StateDelta` will include all previous state writes made to the original (and its ancestors).  Any writes made to the original `StateDelta` after `fork()` is called will not be seen by the forked state.

Finally, after some execution path has been selected, calling `delta.apply()` on one of the possible state updates will commit the changes to the underlying state instance, and invalidate all other delta updates in the same family.  It is a programming error to use the other delta updates after `apply()` has been called, but ideally this should not be a problem in practice: the API is intended to explore a tree of possible execution paths; once one has been selected, the others should be discarded.

## TODO

- [x] Remove the GATs currently introduced in this PR and change the associated stream types to have a `'static` lifetime, and update the internals of the existing `State`/`StateTransaction` code so that their streams can meet that requirement (this probably means wrapping internals in `Arc<RwLock>`).
- [x] Connect the stub prefix stream to the `StateDelta`'s `nonconsensus_prefix` method, then copy-paste its internals to provide stream impls for the other stream types (which are almost but not exactly the same; I think copy-pasting code will be clearer and more maintainable than trying to make the behavior generic)


hdevalence:
The new `StateDelta` code is now fully functional, and side-by-side with the existing `State`/`StateTransaction` code. It passes all the same tests as those structures do.

I'm unsure whether the best course is to (1) merge it as-is, then deduplicate usage later, or (2) to carry through with
- removing `State`/`StateTransaction`;
- exposing `Snapshot`;
- replacing `State` by `StateDelta<Snapshot>` and `StateTransaction` by `StateDelta<StateDelta<Snapshot>>`
- possibly renaming `StateDelta` to `StateTransaction` (?)

in this PR.

hdevalence:
I did the replacement in https://github.com/penumbra-zone/penumbra/pull/1983/commits/47a13e89068f9673d0dbbe272facffbe2e67efdc ; it ended up touching a fair amount of the component code because the trait signatures changed.

PR #1989, opened 2023-02-14
make `tokio` an optional dependency of `penumbra-transaction`

redshiftzero:
Followup to #1970 
See [this Discord discussion](https://discord.com/channels/824484045370818580/979785365437165658/1075021011809992784)

### DIFF
--- a/transaction/Cargo.toml
+++ b/transaction/Cargo.toml
@@ -45,9 +45,13 @@ pbjson-types = "0.5"
 num-bigint = "0.4"
 serde_json = "1"
 tracing = "0.1"
-tokio = { version = "1.21.1", features = ["full"]}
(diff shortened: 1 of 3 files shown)


conorsch:
This change alone isn't enough to get wasm compilation working:

> `error: the wasm32-unknown-unknown target is not supported by default, you may need to enable the "js" feature. For more information see: https://docs.rs/getrandom/#webassembly-support

Is that worth doing alongside these changes, @redshiftzero?

redshiftzero:
Thanks for pointing this out - I was going to add it (despite downstream wasm users being able to add the `js` getrandom feature) so that we could add a CI check, but I ran into a bunch more compilation errors. Since they've likely been resolved in the wasm-bindgen rust code, I'm going to hold off on other changes until it's clear how to proceed re: #1990.

PR #1993, opened 2023-02-14
docs: prefer symlinks to 'cargo run' in guide

### LINKED ISSUE #1976: docs: remove excessive "cargo run" invocations
In https://guide.penumbra.zone, nearly every CLI action takes the form of `cargo run --release --bin <binary> -- <args`. It'd be much more readable to use the short form instead: `<binary> <args>`. 

Let's recommend that folks:

1. run `cargo build --release`
2. run `sudo ln -s "$PWD/target/release/{pd,pcli,pviewd}" /usr/local/bin/`
3. optionally do the same for tm: `sudo cp ~/Downloads/tendermint_0.34.23_darwin_arm64/tendermint /usr/local/bin/`

We've discussed precompiling binaries (#1961), which is a great idea, but we can get immediate wins from a few docs-only changes. 

**Additional context**
Recent discussion in Discord, a testnet participant stated:

>  If we can just move builded binaries to usr/local/bin and instead of this long command use just pd / pcli. It looks more convenient for me. If there any reason for using long, just let me know.

Indeed, that's precisely what I'm doing locally, so I'm in favor of simplifying the workflow for all. During a recent pairing session with @aubrika, we felt it'd be much easier for new folks to run e.g. `pcli view balance` rather than `cd ~/repos/penumbra && cargo run --release --bin pcli -- view balance`.

conorsch:
We're trying to make the guide more accessible. In practice, symlinking the compiled binaries onto a PATH location allows for less typing at the command line, and makes the documentaiton significantly easier to read.

Closes #1976.

### DIFF
--- a/docs/guide/src/dev/devnet-quickstart.md
+++ b/docs/guide/src/dev/devnet-quickstart.md
@@ -10,7 +10,7 @@ To start, you'll need to [install Tendermint `v0.34`](../pd/build.md#installing-
 To generate a clean set of configs, run
 
 ```shell
-cargo run --release --bin pd -- testnet generate
+pd testnet generate
 ```
 
 This will write configs to `~/.penumbra/testnet_data/`.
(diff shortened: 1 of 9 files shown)


hdevalence:
I don't think we should do this. The cargo output directory isn't a stable location for binaries, and we shouldn't tell people to configure their systems in a way that assumes it is.

The reason we tell people to run commands with cargo is because we don't support any mechanism other than from-source builds, and so we use a tool that handles in-source builds (e.g., by automatically recompiling when necessary).

If we don't want to require people to build from source with cargo, we should provide installable binaries, rather than inventing our own ad-hoc installer.

PR #1994, opened 2023-02-15
Refactor to support account groups

### LINKED ISSUE #1972: Introduce "Accounts" and "Account Groups" in place of current address index system

It's useful to be able to maintain multiple logical accounts that record distinct groups of funds for accounting purposes. On transparent chains, this is done by creating distinct on-chain accounts with different addresses, each with a distinct signing key hierarchically derived from a single root secret.

This solution doesn't work well on Penumbra, however, because clients have to scan the chain to learn which transactions are relevant to that user, and naively creating `N` spending keys would mean multiplying the scanning work by a factor of `N`.  To handle this, the diversified address mechanism allows each viewing key to simultaneously scan 2^128 addresses, tagged by a _diversifier_. The diversifier is an encryption of an _address index_ with the viewing key's diversifier key.

We'd like to use this to provide a multiple logical accounts, but our current scheme for address indices doesn't fit this use case very well.

Currently, we use the lower `u64`-sized part of the address index space for numbered addresses, and treat 128-bit address indices as ephemeral addresses. Ephemeral addresses are useful for cases where an address is revealed publicly, such as in an IBC transfer.  However, because all of the ephemeral addresses are in one group, it's not possible to treat numbered addresses as accounts and use ephemeral addresses.

As discussed [in Discord](https://discord.com/channels/824484045370818580/864913084028158012/1072975470200619099), we have four points:

1. Should we use bits of the address index space to hierarchically organize a single spending key's addresses into logical accounts?
2. How many bits of address space should we use?
3. Should these be called "accounts", "sub-accounts", or something else?
4. Should we have a default address for each logical account?

Our decision was:

1. Yes, this is a useful feature for us to support
2. We should split the address index space as a `32-bit account id || 96-bit randomizer`. The division should be made so that the account id can be indexed with a standard sized machine word, so the options are 16, 32, and 64 bits. The birthday bound says that the choice of randomly selecting a colliding randomizer happens with probability `2^{-k/2}`, giving collision probabilities `2^-56`, `2^-48`, `2^-32`. This suggests a 32-bit account id, as `2^-48` is more comfortable than `2^-32` as a collision probability but `2^32` accounts seems more useful than `2^16` accounts.
3. We should call them "accounts", so that they're the primary abstraction. If we need to refer to the collection of all of the addresses, we could use "account group".
4. Yes, we can divide each account's address set into the "default address" (with `randomizer = 0`) and "ephemeral addresses" (with nonzero randomizers).

To make this change, we'll need to:

- [ ] Change the `AddressIndex` structure from its current enum to be defined as
```rust
struct AddressIndex {
    pub account: u32,
    pub randomizer: [u8; 12],
}

impl From<u32> for AddressIndex {
    fn from(account: u32) -> Self {
        Self { account, randomizer: [0; 12] }
    }
}

impl AddressIndex {
    fn is_ephemeral(&self) -> bool {
        self.randomizer != [0; 12]
    }
}
```
- [ ] Rename the (internal) `AccountId` structure to `AccountGroupId` (this is an internal structure, so the name isn't as important to settle on for now, we just want to avoid confusing with the new meaning of "account").
- [ ] Update the view service protos and documentation to reflect the new meaning of "account". In particular, we currently have a way to filter notes in queries by the address index. Instead, we'll want to filter notes by _account index_, matching all of the notes whose address index matches the given account, regardless of randomizer (so that we filter for funds sent to _any_ address of that account).
  - At some point in the future, we may want to change the SQL tables to store the account and randomizer components separately, in order to be able to do that filtering in the database, but since all our data is small at the moment, we don't need to do this now, we can skip it for now and just filter on the returned records using the existing codepaths.
- [ ] Change `pcli`'s interface to handle accounts, by:
  - [ ] Changing `pcli view address` and its docs so that the number indexes an account, not an address
  - [ ] Changing the `--source` flag on the `tx` subcommands to an `--account` flag, and changing it from `Option<u32>` to `u32` with default 0.  (This means that transactions won't commingle funds between accounts, using the existing "source" mechanics in the view service queries).
  - [ ] Change `pcli view balance` to remove the `--by-address` flag (this behavior should be the default), and to group all funds based on the `account` field of the address index.
  - [ ] Check if there are other references to address indices that should be replaced by accounts.

One convenient thing about this change is that previously, the address index encoding format for numeric addresses was `LE64(address index) || 0...0`, and now it will be `LE32(account index) || randomizer`, so the previous default address remains the new default address (for the default account, account 0).  This means that as long as their address index fits in 32 bits, all existing user addresses will remain unchanged.

aubrika:
Closes #1972

### DIFF
--- a/crypto/src/keys/ivk.rs
+++ b/crypto/src/keys/ivk.rs
@@ -11,7 +11,7 @@ use decaf377::{
 
 use super::{AddressIndex, Diversifier, DiversifierKey};
 use crate::{
-    address, fmd, ka,
+    fmd, ka,
     keys::{AuthorizationKeyVar, NullifierKeyVar, IVK_DOMAIN_SEP},
     prf, Address, Fr,
 };
@@ -138,7 +138,7 @@ mod test {
         let rng = rand::rngs::OsRng;
         let spend_key = SpendKey::from_seed_phrase(SeedPhrase::generate(rng), 0);
         let ivk = spend_key.full_viewing_key().incoming();
-        let own_address = ivk.payment_address(AddressIndex::from(0u64)).0;
+        let own_address = ivk.payment_address(AddressIndex::from(0u32)).0;
         assert!(ivk.views_address(&own_address));
     }
 
@@ -151,7 +151,7 @@ mod test {
         let other_address = SpendKey::from_seed_phrase(SeedPhrase::generate(rng), 0)
             .full_viewing_key()
             .incoming()
-            .payment_address(AddressIndex::from(0u64))
+            .payment_address(AddressIndex::from(0u32))
             .0;
 
         assert!(!ivk.views_address(&other_address));
--- a/crypto/src/keys/ivk.rs
+++ b/crypto/src/keys/ivk.rs
@@ -11,7 +11,7 @@ use decaf377::{
 
 use super::{AddressIndex, Diversifier, DiversifierKey};
 use crate::{
-    fmd, ka,
+    address, fmd, ka,
     keys::{AuthorizationKeyVar, NullifierKeyVar, IVK_DOMAIN_SEP},
     prf, Address, Fr,
 };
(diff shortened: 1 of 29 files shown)


crypto/src/keys/ivk.rs
```diff
@@ -54,7 +54,8 @@ impl IncomingViewingKey {
         while u128::from_le_bytes(random_index) <= 2u128.pow(64) {
             rng.fill_bytes(&mut random_index);
         }
-        let index = AddressIndex::Random(random_index);
+        let index = AddressIndex::try_from(random_index.as_slice())
+            .expect("can form address index from slice");
```
hdevalence, line 58: We don't want to do this, now that we changed the semantics of how ephemeral addresses work — the ephemeral_address function should take an account index as an input, rather than generating a random address for a random account.

wallet/src/plan/planner.rs
```diff
@@ -333,7 +333,7 @@ impl<R: RngCore + CryptoRng> Planner<R> {
         // For any remaining provided balance, make a single change note for each
         let self_address = fvk
             .incoming()
-            .payment_address(source.unwrap_or(AddressIndex::Numeric(0)))
+            .payment_address(source.unwrap_or(AddressIndex::default()))
```
hdevalence, line 336: Here the source looks like it was formerly an AddressIndex being used as a logical account, probably the method should be changed so that source is a u32 account index, rather than being an option? There was a note about that change on the original issue and I think this might be running up against that change.

Issue #1209, opened 2022-07-22
Add tombstoning to smoke test


hdevalence:
**Is your feature request related to a problem? Please describe.**

We don't currently have an integration test that checks that a validator who double-signs a block actually gets tombstoned, and since this is kind of annoying to manually run, we've never actually executed that code path. Attempting to fix #1207 revealed that the necessary indexes weren't being populated, so tombstoning wouldn't have worked, even if it had happened. This is bad, and exposes a gap in our testing.

**Describe the solution you'd like**

We should add a tombstoning test to the smoke test.  This would involve (1) creating a new validator (2) inducing it to double-sign (3) checking that it is tombstoned.  One way to induce double-signing is to run two instances of tendermint+pd that have a common `priv_validator_key.json`.

conorsch:
Given recent progress toward #1664, I assume we'd want to run the tombstoning check in rust code against the state machine, rather than via integration tests. Do you agree, @hdevalence?

conorsch:
Scratch that: we discussed at sprint-planning today, and having this as a true integration test is a Good Thing:tm:. Reducing priority for now, though.

conorsch:
I'll take a first look at this for 44; might have to reorganize the compose files a bit to make coerce the double-sign action. Not sure it'll ship in 44, but will document progress as I go.

conorsch:
Took a first pass. I'd like to wire up a new CI workflow much like the smoke test, but that uses two validators, one with a duplicated identity. The small changes suggested in #1992 will make this a bit more straightforward, so I'm going to tackle that first.

hdevalence:
We won't be adding any new checks to the smoke test. This should be done as a staking state machine test.

PR #2003, opened 2023-02-17
feat(pcli): flag for reading tendermint validator keyfile

### LINKED ISSUE #1992: pcli: add flag for reading in tendermint priv key
When configuring a validator, our [docs currently say](https://guide.penumbra.zone/main/pd/join-testnet.html#validator-definitions-penumbra):

> By default template-definition will use a random consensus key that you won’t have access to. Make sure to set the consensus_key correctly otherwise your instance of tendermint won’t be using the key expected in the validator definition. You can get the correct value for consensus_key from your tendermint configs:

There's a bit of interface tension because `pcli` cannot assume that it has access to the same context as `pd`; it may, but we don't know that for sure, thus the instruction to run a shell command manually. We can do better. Let's add an optional flag to pcli to point to the `priv_validator_key.json` file on disk. If that flag is set and the file exists, read the key material from it for use in the template. Otherwise, proceed with filling in a random value, as we do now.

The docs can then be update to recommend using that flag, and qualifying when it's appropriate. See [related discussion in Discord](https://discord.com/channels/824484045370818580/915710988312002610/1075118245964685382).

conorsch:
Rather than recommending a two-step copy/paste workflow for creating validator identities, we now provide a CLI flag that supports reading in the `priv_validator_key.json` file for Tendermint, and uses its pubkey file when rendering the TOML template. Updated docs accordingly.

Closes #1992.

### DIFF
--- a/pcli/src/command/validator.rs
+++ b/pcli/src/command/validator.rs
@@ -14,6 +14,7 @@ use penumbra_proto::{core::stake::v1alpha1::Validator as ProtoValidator, DomainT
 use penumbra_transaction::action::{ValidatorVote, ValidatorVoteBody, Vote};
 use penumbra_wallet::plan;
 use rand_core::OsRng;
+use serde_json::Value;
 
 use crate::App;
 
@@ -47,7 +48,7 @@ pub enum ValidatorCmd {
 pub enum DefinitionCmd {
     /// Create a ValidatorDefinition transaction to create or update a validator.
     Upload {
-        /// The JSON file containing the ValidatorDefinition to upload.
+        /// The TOML file containing the ValidatorDefinition to upload.
         #[clap(long)]
         file: String,
         /// The transaction fee (paid in upenumbra).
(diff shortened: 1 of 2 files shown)


pcli/src/command/validator.rs
```diff
@@ -186,29 +194,39 @@ impl ValidatorCmd {
...
+                            &f,
+                        )?)
+                        .context(format!(
+                            "Could not parse file as Tendermint validator config: {}",
+                            f
+                        ))?;
+                        serde_json::value::from_value::<tendermint::PublicKey>(
+                            tm_key_config["pub_key"].clone(),
+                        )
+                        .context(format!("Tendermint JSON file malformed: {}", f))?
```
hdevalence, line 223: Not sure if this is worth documenting, but a nice property of this implementation is that it will ignore the private key component of the file, so it's still possible to use even if the JSON file only has the public key

hdevalence (approved):

pcli/src/command/validator.rs
conorsch: That's rad. I tested mostly with a completely wrong file, e.g. a non-json ascii text file, and confirmed the error message was displayed. That's enough to get someone to check the contents manually.

conorsch:
The flag is solid, I'm comfortable shipping as-is, but I'm going to take a moment to look at the smoke tests and see if I can trim down the logic there ahead of #1209.

conorsch:
More work required on the smoke test refactor; best handled as part of #1209. Merging here for upcoming testnet.

PR #2008, opened 2023-02-17
Add 128x128-bit fixed-point arithmetic.

### LINKED ISSUE #2014: Implement checked fixed-point arithmetic
Building off the work in #2008 - we want to be able to perform fixed-point arithmetic that is checked by default. One approach to do this, is to implement the `std::ops` binary operators such that their associated type [`Output`](https://doc.rust-lang.org/std/ops/trait.Add.html#associatedtype.Output) wraps the fixed-point struct into an `Option` (`None` when overflow).

The idea is to end-up with an api that lets you combine a sequence of operations which result in an `Option` that the user has either `unwrap`, or `expect` and motivate why its specific application would work.  One of the advantage of this approach is that the failure mode for an out-of-bounds error becomes easier to model (fail-stop). 

- [x] implement basic arithmetic operations (`Add`, `Sub`, `Mul`, `Div`)
- [ ] an efficient way to divide 
- [x] faillible conversion to `Amount` (fails for non-integral fixed-points 
- [x] bytes encoding/decoding
- [ ] property testing

hdevalence:
Closes #2014.

### DIFF
--- a/crypto/src/fixpoint.rs
+++ b/crypto/src/fixpoint.rs
@@ -1,55 +1,97 @@
-use std::{
-    fmt::Display,
-    ops::{Add, Div, Mul, Sub},
-};
+use std::fmt::{Debug, Display};
 
 mod div;
+mod from;
+mod ops;
+
+#[cfg(test)]
+mod tests;
 
 use ethnum::U256;
 
-use crate::Amount;
-
-#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord)]
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
 pub struct U128x128(U256);
 
-impl From<u128> for U128x128 {
-    fn from(value: u128) -> Self {
-        Self(U256([0, value]))
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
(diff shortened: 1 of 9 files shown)


crypto/src/fixpoint.rs
```diff
@@ -0,0 +1,199 @@
...
+        let (integral, fractional) = self.0.into_words();
+        if fractional >= one_half {
+            let Some(integral) = integral.checked_add(1) else {
+                return None;
+            };
+            Some(Self(U256::from_words(integral, 0u128)))
+        } else {
+            Some(self.truncate())
+        }
+    }
```
hdevalence, line 45: I think this would be better as two functions:
```
pub fn round_up(&self) -> Self { ... }
pub fn round_down(&self) -> Self { ... }
```
so that the caller is always choosing whether to round up or down, since that will probably depend on the use case. For instance, we'll generally want to round one direction or the other when working with the DEX, so that we're always burning rounding error rather than minting dust, and which direction is the right one might depend on the call location.

hdevalence: Fixed in https://github.com/penumbra-zone/penumbra/pull/2008/commits/a10d983a26e473fd8e003aa01cdfbae4fac466f8

PR #2018, opened 2023-02-22
account groups: add from_str impl

conorsch:
Toward addressing recent Galileo breakage, as of Testnet 45. Implements FromStr to aid in opt-parsing for galileo; see related PR at: https://github.com/penumbra-zone/galileo/pull/27

Refs https://github.com/penumbra-zone/galileo/issues/26

### DIFF
--- a/crypto/src/keys/diversifier.rs
+++ b/crypto/src/keys/diversifier.rs
@@ -1,5 +1,6 @@
 use std::convert::{TryFrom, TryInto};
 use std::fmt::Debug;
+use std::str::FromStr;
 
 use aes::cipher::{generic_array::GenericArray, BlockDecrypt, BlockEncrypt, KeyInit};
 use aes::Aes128;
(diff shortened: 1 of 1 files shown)


hdevalence (approved):

crypto/src/keys/diversifier.rs
```diff
@@ -177,6 +178,18 @@ impl From<u32> for AddressIndex {
     }
 }
 
+impl FromStr for AddressIndex {
+    type Err = anyhow::Error;
+
+    fn from_str(s: &str) -> anyhow::Result<Self> {
+        let i: u32 = s.parse()?;
```
aubrika, line 185: Is this meant not to support string representations of ephemeral addresses? If so, should we add a comment here to that effect? Otherwise LGTM 😎

aubrika (commented):
LGTM other than nit re: clarifying comment on `from_str` impl

crypto/src/keys/diversifier.rs
conorsch: That was not my intention, simply an oversight. It's not clear to me what changes would be necessary for ephemeral address, so I'll tack on a comment for now!

PR #2022, opened 2023-02-23
update services

Pa1amar:


Pa1amar:
for example root home dir is `/root/` not `/home/root/`

hdevalence:
Thanks!

Issue #1990, opened 2023-02-14
Merge penumbra-web-assembly into monorepo


redshiftzero:
## Overview
We want to merge the external create https://github.com/zpoken/penumbra-web-assembly into this monorepo. 

## Next steps

- [x] Get upstream wasm-pack build logic from Zpoken, PR into this repo: https://github.com/penumbra-zone/penumbra/pull/2163
  - [x] Adapt our CI workflows to add `wasm32-unknown-unknown` target for the `penumbra-web-assembly` crate: https://github.com/penumbra-zone/penumbra/pull/2163/commits/7cc20770067265a66f5d1310918d060a89fe8b7d
  - [x] Identify problematic areas in our codebase, where our deps cause build to fail
  - [x] Plan refactors for async (particularly tokio) code, add feature-gating where possible to unblock wasm32 builds
  - [x] Implement code changes requested by Zpoken; taken from [related discussion in Discord](https://discord.com/channels/824484045370818580/979785365437165658/1075184618178875462): 
    - [x] remove  penumbra-chain -> penumbra-storage -> tonic dependency  - #1857
    - [x] remove  penumbra-chain -> penumbra-proto -> tonic dependency 
    - [x] make jmt optional for penumbra-chain and penumbra-crypto - https://github.com/penumbra-zone/penumbra/pull/2025
    - [x] remove  penumbra-transaction -> tokio dependency -  #1989
  - [x] Get all CI workflows passing on PR
- [x] Clean up Rust code:
  - [x] Delete the `wasm-compat` crate (assuming it's actually just for build testing?)
  - [x] Rename the `penumbra-web-assembly` crate to `penumbra-wasm` in a `/wasm` directory, to match the other crates (e.g., `penumbra-storage` is in `/storage`)
  - [x] Run `cargo fmt` on `penumbra-wasm`
  - [x] Add `cargo fmt` to the `penumbra-wasm` CI (assuming we keep separate CI)
- [ ] Merge PR into penumbra/main with CI guarding against WASM-breaking regressions

Later:
    - [ ] Optional: have the possibility to use in penumbra-web-assembly  struct from penumbra-view (SpendableNoteRecord, SwapRecord) 

## CI changes
We should have a CI check to prevent breaking downstream users that are compiling our crates to wasm

To do so:
1. enable the `wasm32-unknown-unknown` target using rustup 
[...]

PR #2025, opened 2023-02-24
build: remove unused jmt dep for chain & crypto

conorsch:
Towards #1990. Here we mark the `jmt` dependency as optional in `chain` and `crypto`. This change alone isn't enough for a working wasm compilation; see in-line comments about the `js` feature for `getrandom`, which was also discussed in #1989. 

If I enable the js feature (currently commented out) in `crypto`, then I can build an artifact via:

```
cargo build --release --target wasm32-unknown-unknown -p penumbra-crypto
```

However, I cannot build the `chain` crate. This is still progress, though, so I propose we merge and we continue tracking work in #1990.


crypto/Cargo.toml
```diff
@@ -34,6 +35,10 @@ bytes = "1"
 derivative = "2.2"
 hex = "0.4"
 hmac = "0.12.0"
+# Not enabling js feature for getrandom, which is a transitive dep of rand_core,
+# because the docs recommend against doing it in libraries: https://docs.rs/getrandom/latest/getrandom/#webassembly-support
+# Downstream client projects can modify their Cargo.toml files to enable it.
+# getrandom = { version = "0.2", features = ["js"] }
```
hdevalence, line 41: 👍

chain/Cargo.toml
```diff
@@ -13,7 +13,8 @@ penumbra-tct = { path = "../tct" }
 
 # Penumbra dependencies
 decaf377 = { git = "https://github.com/penumbra-zone/decaf377" }
-jmt = "0.3"
+# We make JMT optional to support WASM for web extension.
+jmt = { version = "0.3", optional = true }
```
hdevalence, line 17: Hold on, this makes the crate optional, but there's no other changes inside the library. Does that mean we weren't using it anyways, and could just remove it, or is something going to break when it's disabled?

conorsch: Hey, nice! You're right: we can fully remove `jmt` in both chain and crypto; still builds fine. Updated, CI is running now. Separately I'll scan for unused deps elsewhere as a housekeeping exercise.

PR #2029, opened 2023-02-24
Add --home argument to specify home directory for pclientd

### LINKED ISSUE #2019: Fix `pclientd` to use a `--home` directory, like `pd`
**Describe the bug**

When adding config support to `pclientd`, we didn't update the command line arguments, so it scatters state all over the file system. Instead, it should have a `--home` argument that specifies a directory, and only writes state into that directory.

This would replace the existing `sqlite_path` argument, which should now be computed relative to the home directory (e.g. `PCLIENTD_HOME/pclientd-db.sqlite`.  The config file should also be located in the same directory, e.g., as `PCLIENTD_HOME/config.toml`.

aubrika:
Closes #2019

### DIFF
--- a/view/src/bin/pclientd.rs
+++ b/view/src/bin/pclientd.rs
@@ -1,6 +1,6 @@
 #![allow(clippy::clone_on_copy)]
 use anyhow::{Context, Result};
-use camino::Utf8PathBuf;
+use camino::{Utf8Path, Utf8PathBuf};
 use clap::Parser;
 use penumbra_crypto::keys::{SeedPhrase, SpendKey};
 use penumbra_crypto::FullViewingKey;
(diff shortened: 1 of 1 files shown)


view/src/bin/pclientd.rs
```diff
@@ -163,33 +173,43 @@ async fn main() -> Result<()> {
...
+                "Starting view service at node {:?} and port {:?}",
+                &opt.node, &opt.pd_port
+            );
 
             let service = ViewService::new(storage, opt.node, opt.pd_port).await?;
+            let config_file_path = &mut opt.home.clone();
+            config_file_path.push("config.toml");
 
-            let config_contents = match fs::read_to_string("./client_data/pclientd/config.toml") {
+            let config_contents = match fs::read_to_string(config_file_path) {
```
hdevalence, line 212: Do we maybe want to read and validate the config data before starting the view service?

hdevalence (approved):

Issue #1647, opened 2022-11-21
Add an additional RocksDB column family to look up keys by key hash


hdevalence:
**Is your feature request related to a problem? Please describe.**

To make ICS23 nonexistence proofs, we need to be able to create existence proofs for the left and right splits from the nonexistent key.  This isn't currently possible, because we'll look up those leaf nodes by key hash, but the existence proof requires a full key.

**Describe the solution you'd like**

Add an additional RocksDB column family that allows looking up keys by key hashes. The (RocksDB) keys would be (JMT) key hashes and the (RocksDB) values would be (JMT) keys.  Like the existing JMT keys column family, this column family should be inserted into when a key is added and removed when a key is deleted.

redshiftzero:
Status update? Is this done @plaidfinch?

conorsch:
We think not; but there's WIP in https://github.com/penumbra-zone/penumbra/pull/1681

conorsch:
Update: we still need functionality to support non-existence proofs. The work in #1681 should be revisited; however, this work won't make 042 testnet (#1877) given other priorities.

conorsch:
This will ship as part of #2032, which is slated for 51, so we're on track. Won't land until the associated IBC changes are in, also for 51.

PR #2032, opened 2023-02-24
storage: integration of `jmt@0.4.0` & reverse index for jmt key hashes

### LINKED ISSUE #2030: Upgrade to `jmt` `0.4`

If we upgrade to `jmt` `0.4`, we'll pick up a ~10-15x speedup on JMT queries for every state read in our application, because of https://github.com/penumbra-zone/jmt/pull/77 . That PR also changed the `StateRead` trait, however, so we'll have to change how we manage RocksDB column families slightly.

erwanor:
Closes #2030  and #1647


conorsch:
Heads up, will need a rebase after https://github.com/penumbra-zone/penumbra/pull/2025 lands, hopefully within the hour.

erwanor:
Sure! thanks for the heads up

conorsch:
#2025 landed, as did https://github.com/penumbra-zone/penumbra/pull/2034. I believe that's the last time today that _I_ will ask you to rebase... can't speak for others, though. :wink:

storage/src/snapshot.rs
```diff
@@ -215,8 +219,9 @@ impl StateRead for Snapshot {
                         let k = std::str::from_utf8(key_preimage.as_ref())
                             .expect("saved jmt keys are utf-8 strings")
                             .to_string();
+
                         let v = self2
-                            .get_jmt(k.as_bytes().into())?
+                            .get_jmt(KeyHash::with::<Sha256>(k.clone()))?
```
hdevalence, line 224: wait, why doesn't `.into()` work to construct key hashes any more?

storage/src/snapshot.rs
```diff
@@ -309,6 +314,16 @@ impl StateRead for Snapshot {
 /// A reader interface for rocksdb. NOTE: it is up to the caller to ensure consistency between the
 /// rocksdb::DB handle and any write batches that may be applied through the writer interface.
 impl TreeReader for Inner {
+    /// Gets a value by identifier, returning the newest value whose version is *less than or
+    /// equal to* the specified version.  Returns None if the value does not exist.
+    fn get_value_option(
```
erwanor, line 319: note: this probably should be a default implementation

storage/src/snapshot.rs
erwanor: we removed it with `0.4.0`, I think to make space for the generic hasher, see:

https://github.com/penumbra-zone/jmt/commit/2475dfb7b73bf16de518c2de34f10e00cd4fa909#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759L160

storage/src/snapshot.rs
hdevalence: I don't think it can be a default implementation, the point of the `jmt` change was to have implementors store segmented data...

storage/src/snapshot.rs
hdevalence: Ah, hmm, that seems unfortunate. I think it'd be better if there was a `From` impl, but not really relevant for this PR.

storage/src/snapshot.rs
erwanor: perhaps something like:

```rust
pub struct KeyHash<H: SimpleHasher>(pub [u8; 32], PhantomData<H>);
```

?

storage/src/snapshot.rs
erwanor: perhaps something like:

```rust
pub struct KeyHash<H: SimpleHasher>(pub [u8; 32], PhantomData<H>);
```

?

conorsch:
@erwanor I'm able to reproduce the WASM compat CI failure locally:

```
cd wasm/
cargo build --release --target wasm32-unknown-unknown
```

Whereas on `main`, that same command is still passing for me locally. So it's definitely a regression in this PR, that much I'm convinced off. The build error is:

```
error[E0433]: failed to resolve: use of undeclared crate or module `imp`
  --> /home/ubuntu/.cargo/registry/src/github.com-1ecc6299db9ec823/wait-timeout-0.2.0/src/lib.rs:66:9
   |
66 |         imp::wait_timeout(self, dur)
   |         ^^^ use of undeclared crate or module `imp`

For more information about this error, try `rustc --explain E0433`.
error: could not compile `wait-timeout` due to previous error
warning: build failed, waiting for other jobs to finish...
```

Looking at the docs/repo for `wait-timeout`, it appears that dependency does not and will not support wasm: https://github.com/alexcrichton/wait-timeout/issues/18 So why do we need it?

```
❯ cargo tree --invert wait-timeout 
wait-timeout v0.2.0
└── rusty-fork v0.3.0
    └── proptest v1.1.0
        └── jmt v0.4.0
            └── penumbra-chain v0.1.0 (/home/conor/src/penumbra/chain)
                ├── penumbra-transaction v0.1.0 (/home/conor/src/penumbra/transaction)
                │   └── penumbra-wasm v0.48.0 (/home/conor/src/penumbra/wasm)
                └── penumbra-wasm v0.48.0 (/home/conor/src/penumbra/wasm)
```

Looks like jmt requires proptest, which pulls in the the incompatible dep. But we surely we don't need proptest in the full dependencies, right? Indeed, if I hack the jmt crate Cargo.toml locally:

```diff
diff --git a/Cargo.toml b/Cargo.toml
index bcebf19..2dbe6fc 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,8 +24,6 @@ mirai-annotations = "1.10.1"
 num-derive = "0.3.3"
 num-traits = "0.2.14"
 once_cell = "1.7.2"
-proptest = { version = "1.0.0" }
-proptest-derive = { version = "0.3.0"}
 serde = { version = "1.0.124", features = ["derive"] }
 thiserror = "1.0.24"
 prometheus = "0.13"
@@ -36,3 +34,5 @@ tracing = "0.1"
 
 [dev-dependencies]
 rand = { version = "0.8.3" }
+proptest = { version = "1.0.0" }
+proptest-derive = { version = "0.3.0"}
```

Then I'm able to build the wasm target on this PR again, referencing my local dir for the jmt dep. So I'm sorry to say it looks like to satisfy that check, we'll need to adjust the deps in the jmt crate and re-release.

hdevalence:
Adjusting the deps in the jmt crate seems like the right move to me. If they're not used in the public API, it could be done in a patch release.

hdevalence:
Looks like we'll have to update our `ibc` dependency:
```
~/c/p/penumbra (jmt4|✔) $ cargo tree -i ics23@0.9.0
ics23 v0.9.0
├── ibc v0.29.0
```

erwanor:
yes, this is on hold until routing execution is completed.

erwanor:
@conorsch  This is blocked on #2391, because we cannot update our ibc dependency with the recent changes to `ibc-rs`, but our goal is still to include this for 51.

hdevalence:
This should be unblocked as of https://github.com/penumbra-zone/ibc-types/pull/13 ; I made an `0.2.x` branch on that repo that the Penumbra monorepo can use instead of the current `0.1.x` branch.

erwanor:
Unfortunately, this PR coalesce a few different changes:

- #2391 (+ support for `tendermint-rs@0.31.1`)
- #1647 
- https://github.com/penumbra-zone/ibc-types/issues/3
- and #2030 which it initially set out to do

that have to be tackled together. As it stands, it likely won't make it into 52 as it needs a round of debugging (related to deserialization failures that happen when fetching jmt `Node`s from RocksDB), and a round of quality assurance. With that said, pushing this across the finish line will be my primary focus once the DEX engine, and associated work is completed.

hdevalence:
We won't be able to merge these changes, since they've been rebased too many times, they don't apply cleanly any more, and we don't have a way to know that they actually work (in fact, merging them causes the storage tests to fail). There are also a lot of significant design changes buried within these changes -- e.g., why do we need a dirty bit on the JMT backing data now?

Instead, we'll need to start over, and plan a phased sequence of changes that only does one thing at a time, changing just the Tendermint version, or just the JMT version, etc, rather than combining them all into one PR.

erwanor:
The failing test is triggered by using the wrong serialization methods, though for some reason doing `Node::serialize -> Node::deserialize` seems to always fail. I haven't gotten to the bottom of this yet. The dirty bit is an interim solution that ought to be replaced by using an `Option`, it was needed because  `get_value_option` will do a cf lookup for the value with the greatest version up to the specified height, so you need a way to signal that the value is deleted after a certain version.

PR #2033, opened 2023-02-24
use MemoPlaintext structured data instead of String for memos

### LINKED ISSUE #1880: Change memo plaintexts to include a structured sender address.

For payment transactions, it's important that the receiver of a payment can view information about who sent them funds. To address this use case, we have a memo field, similar to Zcash.  Our memo field is 512 bytes of UTF-8 text.  While this is enough space to encode a good amount of information about the sender and the purpose of the payment, that information is unstructured.

Having unstructured data at the protocol level is good, because it leaves flexibility for evolving use-cases in the future. But it also imposes costs, because clients have to agree on how to encode data in the memo.  Currently, the entirety of the memo field is unstructured, but we could do better by defining a sender address field.  This would give client software a standard way to display the sender of a payment.

Instead of 512 bytes of freeform UTF-8 text, memo plaintexts should be encoded as `80-byte address || 432-byte text`, i.e., exactly the same encoding as now, but with a fixed 80-byte prefix.

This should be represented in the client code by a `Memo` domain/proto type defined as
```rust
struct Memo {
    pub sender: Address,
    pub text: String,
}
```
This change should propagate through the `TransactionView` code.  Along the way, we should tweak the memo handling in the `TransactionView` so that the memo ciphertext is included even when the memo plaintext is not visible, in support of #1788.

When constructing a transaction, the `Planner` should fill in the `sender` field with the default address if it is unset. In `pcli`, we should set the address based on the source-index parameter if it is set.

avahowell:
closes #1880

### DIFF
--- a/crypto/src/address.rs
+++ b/crypto/src/address.rs
@@ -98,22 +98,27 @@ impl Address {
 
     /// A randomized dummy address.
     pub fn dummy<R: CryptoRng + Rng>(rng: &mut R) -> Self {
-        let mut diversifier_bytes = [0u8; 16];
-        rng.fill_bytes(&mut diversifier_bytes);
+        loop {
+            let mut diversifier_bytes = [0u8; 16];
+            rng.fill_bytes(&mut diversifier_bytes);
 
-        let mut pk_d_bytes = [0u8; 32];
-        rng.fill_bytes(&mut pk_d_bytes);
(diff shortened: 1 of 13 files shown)


avahowell:
Fixed the failing smoke test and rebased on latest main

PR #2034, opened 2023-02-24
build: remove unused cargo dependencies

conorsch:
Ran `cargo +nightly udeps --all-targets` and slimmed down the Cargo.toml files based on that output. Followed the recommendations rather blindly, but still builds fine.

Refs #1990, which tries to eliminate a few specific dependencies. Motivation for scanning for unused deps in #2025.


hdevalence (approved):

conorsch:
Tests are passing! :shipit:

PR #2049, opened 2023-02-27
pcli: add output describing proving latency

hdevalence:
we don't have to keep this forever, but it seems nice as we enable zk proofs


hdevalence:
example output
```
building transaction...
finished proving in 3.379 seconds [9 actions, 9 proofs, 4370 bytes]
broadcasting transaction...
```

PR #2058, opened 2023-02-28
Delete `measure` because it's unused

plaidfinch:
We don't  use this anymore, and it costs build time. Let's delete it.


hdevalence:
This looks good, but I think it needs edits to the Cargo.toml workspace definition.

hdevalence:
On second thought I think this could still be useful — let's keep it for now.

PR #2061, opened 2023-02-28
proof-params: use prepared verification keys

hdevalence:
Not sure if this clobbers any other changes, I did it just to see if there would be a performance impact and then put it up.


redshiftzero (approved):
LGTM!

PR #2091, opened 2023-03-04
view: add BroadcastTransaction with confirmation detection to ViewService

hdevalence:
This pulls code out of `pcli` and puts it in the view service, so that users of
the view service can submit transactions and await detection without having to
write custom code.


hdevalence:
Of note while testing this: the `pcli view tx` command is broken.

PR #2118, opened 2023-03-07
Further implementation of buy orders in pcli

zbuc:


pcli/src/command/tx.rs
```diff
@@ -785,45 +783,100 @@ impl TxCmd {
...
+                        // and r2 will be set to the amount of the asset being sold that would be needed to buy the desired amount of the asset being bought
+                        // (divided by 1_000_000 to correct the scaling)
+                        r2: (buy_order.price.amount * buy_order.desired.amount)
+                            / 10u128.pow(6).into(),
+                    }
+                } else {
+                    Reserves {
+                        // r1 will be set to the amount of the asset being sold that would be needed to buy the desired amount of the asset being bought
+                        // (divided by 1_000_000 to correct the scaling)
+                        r1: (buy_order.price.amount * buy_order.desired.amount)
```
erwanor, line 804: question: in this case, we're provisioning `asset_1` and expecting the order to fill with some `asset_2`, so shouldn't the trading function coefficients (`p` and `q`) be swapped as well? I guess they would be swapped in the other branch (when `asset_1 = desired_asset`)

pcli/src/command/tx.rs
```diff
@@ -785,45 +783,100 @@ impl TxCmd {
...
+                }
+                let trading_function = TradingFunction::new(trading_pair, *spread, p, q);
+
+                // This represents an "ask" in the order book, where bids are placed in the asset type without initial reserves.
+                //
+                // The [`Position`] constructor expects the ordering of [`Reserves`] to match the ordering of the assets in the [`TradingPair`].
+                //
+                // When opening a liquidity position, the initial reserves will only be set for one asset.
+                let reserves = if trading_pair.asset_1() == sell_order.selling.asset_id {
+                    Reserves {
```
erwanor, line 854: same question here, shouldn't we swap `p` and `q`

pcli/src/command/tx.rs
zbuc: Good catch

Issue #2056, opened 2023-02-27
rework how proof parameters are loaded


redshiftzero:
Currently we have a `penumbra-proof-params` crate that requires git LFS at build time to build the parameters. This makes it hard for third parties to use penumbra as a dependency, see  https://github.com/penumbra-zone/galileo/issues/29. Also otherwise it creates friction for our build process. 

Since the VKs are small (~1KB each) we can skip git LFS for them. That size halves if we implement #2038.

Suggested next steps from Discord by @hdevalence :
- [x] keep VKs out of git lfs and have them just be ordinary files
- [x] unconditionally include VKs in the params crate
- [x] conditionally include PKs in the params crate with a “proving-key” feature
- [x] add a “auto-download” feature to the params crate that fetches proving keys if missing

For the last step, we can use the Git LFS Batch API to convert the pointer files to URLs. See documentation [here](https://github.com/git-lfs/git-lfs/blob/main/docs/api/batch.md#requests).

Added based on feedback:
- [x] Add a WASM-friendly way to load the parameters (no longer needed)

hdevalence:
I think we should prioritize the last point, so that we can allow dependencies outside of the monorepo to use the proof parameter crates (cf https://github.com/penumbra-zone/galileo/issues/29)

PR #2121, opened 2023-03-08
wip: rework `penumbra-proof-params`

redshiftzero:
Towards #2056 - two items remaining there


redshiftzero:
From discussion with downstream users, it sounds like the ability to load the parameters is resolved and was due to the build system used (wasm-pack IIRC) rather than a fundamental limit in the wasm binary itself, so I'm going to back out the custom wasm family logic here

PR #2134, opened 2023-03-10
pcli: extend functionality of `pcli view address` to include account refactor

### LINKED ISSUE #2007: Add `pcli address view`
Following #1972, it would be nice to have a way to view addresses via pcli and be able to determine whether those addresses are part of / visible to a given account.

aubrika:
Closes #2007

### DIFF
--- a/pcli/src/command/view/address.rs
+++ b/pcli/src/command/view/address.rs
@@ -2,12 +2,13 @@ use anyhow::Result;
 use base64::Engine;
 use rand_core::OsRng;
 
-use penumbra_crypto::FullViewingKey;
+use penumbra_crypto::{Address, FullViewingKey};
 
 #[derive(Debug, clap::Parser)]
 pub struct AddressCmd {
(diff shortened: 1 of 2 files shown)


pcli/src/command/view/address.rs
```diff
@@ -28,13 +31,21 @@ impl AddressCmd {
...
+        let address_index = fvk.incoming().address_index(&address);
+
+        match address_index {
+            Some(address_index) => {
+                if self.base64 {
+                    println!(
+                        "Account {0}: {1}",
+                        address_index.account,
+                        base64::engine::general_purpose::STANDARD.encode(address.to_vec())
+                    );
```
hdevalence, line 43: I think this breaks the base64 output format by making it text-based instead.

aubrika: It doesn't seem to based on testing this implementation alongside the legacy implementation - both base64 address strings match.

PR #2146, opened 2023-03-10
client: add witness_and_build rpc

### LINKED ISSUE #1883: TransactionBuild RPC
Following #1782, we need another RPC on the view service to receive a TransactionPlan and build the Transaction

aubrika:
Closes #1883


aubrika (commented):
Makes sense, looks good, thanks for wrapping it up :)

Issue #1915, opened 2023-01-30
Replace transaction expiry height with a time-based mechanism


hdevalence:
**Is your feature request related to a problem? Please describe.**

The current `TransactionBody` has an `expiry_height` field, which I originally added because it was present in Zcash (which, in turn, inherited it from Bitcoin).  In that context, using a height makes sense because it's the only notion of temporality in the system. With Tendermint, however, validators come to consensus on a latest time (and we use that notion of time elsewhere, e.g., in IBC).  From an end-user perspective, specifying transaction validity periods in terms of time is much more useful than specifying them in terms of block heights. In addition, we'd like to have the chain logic be relatively independent of the block height -- if the chain produces blocks faster than normal, the rest of the logic should be as unaffected as possible.

**Describe the solution you'd like**

- [ ] Remove the current `expiry_height` field from the transaction body;
- [ ] Add new fields, `valid_before` and `valid_after`. These messages should have the following semantics, where `latest_time()` is the most recent block header timestamp:
  - If `valid_before` is set, the transaction is valid iff `latest_time() < valid_before`;
  - If `valid_after` is set, the transaction is valid iff `latest_time() > valid_after`;
 - [ ] Add state machine unit tests that
   1. Initialize a genesis state
   2. Set a test timestamp (see the IBC state machine test for an example of how to do this)
   3. Create transactions with all four combinations of `valid_before` and `valid_after` present and not present, and check that validation rules behave as expected.

mikayla-maki:
Note on Checkbox 3.ii: This seems to refer to the test in `component/src/app/ibc/component/client.rs` file

hdevalence:
> This seems like it removes the ability for someone to specify "I want this transaction to execute if and only if it is included in this specific block". That seems useful. Is it less useful than I think it is?

To the extent that the desired behavior can't be specified by an equivalent time interval, I don't think we should support it. Ideally, I think we want the block interval to be an implementation detail, and keep the primary mechanism time-based.

I think there are two use cases:

- I want to bound when the transaction can execute — in which case, a time bound is better than a block height bound, because it names the actually desired range
- I want to bound when this transaction can execute relative to something else — in which case I think the solution is to combine transactions coinjoin-style (eg I want my swap to execute in the same block as yours, so we'll aggregate them into one transaction).

So I think we should get rid of height-based mechanics entirely, so that we're prepared for a setting where block intervals float and vary over time.

conorsch:
[...]

Issue #2070, opened 2023-03-01
Tracking Issue: Dynamic Epochs


avahowell:
**Is your feature request related to a problem? Please describe.**

Currently, the epoch system in Penumbra's state machine is static: we have an `is_epoch_end` method on `chain.Epoch` which tracks the epoch cycle using the configured static epoch `duration` (in blocks) and index. It's up to components (such as the staking component) to design their own state transition logic around this simple helper function, which assumes that epochs have a static length in blocks. There is no global (i.e., at the Component level) notion of the current epoch, or current position in the epoch. In practice, this leads to quite of bit of complexity and challenges, such as for handling mid-epoch slashing events. Currently all of the mechanisms in Penumbra assume static-length epochs and handle this complexity on a case-by-case basis.

**Describe the solution you'd like**

Implement *dynamic* epochs that can be cleanly interrupted by various components of the state machine. For this change to preserve the current mechanism design of Penumbra, we need to also rework the mechanisms (such as staking rate) to operate based on clock time rather than index in the epoch. This second piece also allows us to safely use the Tendermint `skip-empty-blocks` flag, improving performance.

Using wall clock time for mechanisms has its own drawbacks. There are potential concerns around manipulation of timestamps (the source of truth for time-based epochs) by malicious (1/3 + 1) quorums of validators (see https://github.com/cosmos/cosmos-sdk/issues/5411 and https://github.com/tendermint/tendermint/issues/2653 for deeper discussion of this issue). We can avoid the bulk of concern (h/t @zmanian) by adding a synchrony assumption: Penumbra nodes should reject blocks whose timestamp is >~1hr different than their own. This aligns with the `Node Based Rejection` approach described [here](https://github.com/tendermint/tendermint/issues/2653).

hdevalence:
I think it makes sense to conceptually separate two distinct but related changes here:

1. Changing from static to dynamic epochs;
2. Making Penumbra's economic mechanisms, like staking rates, time-based.

On (1), the original reason to have epochs in the first place was so that we could ensure that the validator set only changes at defined epoch boundaries, rather than in any arbitrary block.  This is necessary so that when we have the validator set do threshold cryptography, we can ensure that the set of participants in the MPC isn't constantly changing, and that when it does change, there's a place to do epoch-specific setups, like a DKG.

[...]

PR #2148, opened 2023-03-13
wip: Add valid_before and valid_after fields to transaction

mikayla-maki:
This PR aspires to implement https://github.com/penumbra-zone/penumbra/issues/1915

I have removed expiry height, and added 2 new fields for valid_before / valid_after.

TODO:
- [x] Figure out how to add a state machine test for this behavior
- [x] Remove expiry_height
- [x] Do due diligence on time representations
- [ ] Get a pass from the penumbra dev team on:
  - [ ] Wording in documentation / errors
  - [ ] Proto numbering
- [ ] Have the PR pass review :D

### DIFF
--- a/component/src/action_handler/transaction/stateful.rs
+++ b/component/src/action_handler/transaction/stateful.rs
@@ -41,22 +41,13 @@ pub(super) async fn timestamps_are_valid<S: StateRead>(
     let current_time: Timestamp = state.get_block_timestamp().await?.into();
 
     let after = transaction.transaction_body().valid_after;
-
-    let before = transaction.transaction_body().valid_before;
-
-    if current_time.check_expiry(&after) == ibc::timestamp::Expiry::Expired {
+    if after.check_expiry(&current_time) == ibc::timestamp::Expiry::Expired {
         anyhow::bail!("Too late!");
     }
 
-    if before.check_expiry(&current_time) == ibc::timestamp::Expiry::Expired {
+    let before = transaction.transaction_body().valid_before;
+    if current_time.check_expiry(&before) == ibc::timestamp::Expiry::Expired {
         anyhow::bail!("Too early!");
-    }
-
-    if transaction.transaction_body().expiry_height != 0 {
-        let current_block = state.get_block_height().await?;
-        if current_block > transaction.transaction_body().expiry_height {
-            anyhow::bail!("Too late!");
-        }
     }
 
     Ok(())
--- a/component/src/action_handler/transaction/stateful.rs
+++ b/component/src/action_handler/transaction/stateful.rs
@@ -1,4 +1,5 @@
 use anyhow::Result;
+use ibc::timestamp::Timestamp;
 use penumbra_chain::StateReadExt as _;
 use penumbra_storage::StateRead;
 use penumbra_transaction::Transaction;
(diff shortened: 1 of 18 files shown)


component/src/action_handler/transaction/stateful.rs
```diff
@@ -32,3 +33,31 @@ pub(super) async fn fmd_parameters_valid<S: StateRead>(
         height,
     )
 }
+
+pub(super) async fn timestamps_are_valid<S: StateRead>(
```
mikayla-maki, line 37: I wasn't sure exactly where to add the checks for these fields, and what the difference between a panic and an `Err()` is in this context, this is my best guess. I also couldn't find any check for the expiry height anywhere, so I added it here.

proto/proto/penumbra/core/transaction/v1alpha1/transaction.proto
```diff
@@ -41,6 +41,24 @@ message TransactionBody {
...
   optional bytes encrypted_memo = 6;
+  // A unix timestamp, in nanoseconds.
+  // If this value is set, the transaction is valid iff the most recent block header
+  // timestamp is < this value.
+  //
+  // If the value is zero, this value is considered unset.
+  // [TODO: Double check that this actually does behave like google's protobufs]
+  // The behavior of this time stamp should match google's protobuf specification:
+  // [specification]: https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.Timestamp
+  uint64 valid_before = 7;
```
mikayla-maki, line 52: This is supposed to match exactly the Timestamp type in IBC, which has this note on it. I don't have the context to verify this yet, hence the TODO.

proto/proto/penumbra/view/v1alpha1/view.proto
```diff
@@ -106,6 +106,10 @@ message TransactionPlannerRequest {
     core.crypto.v1alpha1.Fee fee = 2;
     // The memo for the requested TransactionPlan
     string memo = 3;
+    // The valid_before timestamp for the requested TransactionPlan, formatted as an RFC 3339 timestamp
+    optional string valid_before = 4;
```
mikayla-maki, line 110: From what I remember of the penumbra architecture, this is  the internal RPC call that converts from UI to internal penumbra data types. As this is more 'user facing', I thought a string would be nicer here.

view/src/service.rs
```diff
@@ -354,6 +354,19 @@ impl ViewProtocolService for ViewService {
             )
             .expiry_height(prq.expiry_height);
 
+        if let Some(timestamp) = prq.valid_after {
+            let time = tendermint::Time::parse_from_rfc3339(timestamp.as_str()).map_err(|e| {
+                tonic::Status::invalid_argument(format!("Could not parse valid_after: {e:#}"))
+            })?;
+            planner.valid_after(time);
```
mikayla-maki, line 361: Note that the view service does not depend on IBC, only tendermint, so this code has to use the tendermint `Time` type

transaction/src/plan.rs
```diff
@@ -40,6 +41,8 @@ pub struct TransactionPlan {
     pub fee: Fee,
     pub clue_plans: Vec<CluePlan>,
     pub memo_plan: Option<MemoPlan>,
+    pub valid_before: Timestamp,
```
mikayla-maki, line 44: Note that the transaction crate does not depend on tendermint, only IBC, and so has to use the IBC `Timestamp` type.

hdevalence:
This looks great!

The only thing I'm not sure about is modeling the timestamps using the IBC timestamp types -- I'm not sure we want to link our transaction domain types to the IBC types that way.  Using a Tendermint timestamp type might also be problematic, because IIRC there are a bunch of weird rules it implements to be exactly compatible with some accidental behavior of a Go library for timestamp formatting, and we might not want to pull in those consensus rules by accident.  That said, at the moment we don't have a timestamp type of our own, so we don't have a ready-to-go alternative.

We're currently in the process of adding time-based dynamic epochs (#2070), which will have to define our own timestamp type. So I think the cleanest thing to do would be:

- We will finish and merge work on #2070 that includes a timestamp type;
- We will rebase your branch on top of that work, and swap out the timestamp types;
- Then, pick up code review again.

Sorry for the mis-planning and the delay -- I know we'd listed it as a good first issue, and it turned out to be more intertwined than expected -- but I think if we wait a bit, we can rebase it on top of that work and then resume the process!

mikayla-maki:
Sounds good to me! Guess we'll leave this hanging for a while, excited to see what y'all come up with :D

Also, glad to hear that the code I wrote is otherwise OK, this is a lot more of the system than I'm used to interacting with :)

hdevalence:
We didn't land this :( 

Sorry to have let you down on it, but I think it's best that we close this out for now.

PR #2161, opened 2023-03-15
proof-params: add hardcoded hashes for proving keys

hdevalence:
Not sure if this is the best approach; AFAICT there's not a way to run the parameter generator without overwriting the existing parameters, so I manually set the debug hashes to the values for the existing parameters.


hdevalence:
Hoping this is useful in debugging https://discord.com/channels/824484045370818580/915710988312002610/1085418581753339924

redshiftzero (approved):
LGTM!

PR #2163, opened 2023-03-15
added penumbra-web-assembly crate

zpoken:


conorsch:
FYI, when I tried to push to the PR's branch on the upstream fork, it failed with:

```
error: Authentication error: Authentication required: You must have push access to verify locks                                                   
error: failed to push some refs to 'github.com:zpoken/penumbra'
```

That's a side-effect of our use of git-lfs, and can be sidestepped by running:

```
git config lfs.https://github.com/zpoken/penumbra.git/info/lfs.locksverify false
```

custody/Cargo.toml
```diff
@@ -11,7 +11,7 @@ penumbra-proto = { path = "../proto" , features = ["rpc"] }
 penumbra-crypto = { path = "../crypto" }
 penumbra-transaction = { path = "../transaction" }
 
-tokio = { version = "1.21.1", features = ["full"]}
+tokio = { version = "1.21.1", features = ["rt", "sync"]}
```
hdevalence, line 14: I think it'd actually be better to stick with `features = ["full"]` here and elsewhere, because always pulling in the full feature set limits the number of possible feature-sets that Tokio will be compiled with, which reduces the number of times the crate has to be recompiled (once for each combination of feature flags).

On the other hand, pruning features out of the Tokio crate doesn't really help with the goal of getting wasm builds working, since the wasm builds can't depend on Tokio at all.

hdevalence: On further inspection -- I'm confused why this is even an issue, it looks like the `wasm-compat` crate doesn't depend on Tokio at all:
```
~/c/p/p/wasm-compat (wasm_integration|✔) $ cargo tree -i tokio
error: package ID specification `tokio` did not match any packages

	Did you mean `toml`?
```
This is when resolving dependencies for the `wasm-compat` crate specifically, rather than as part of the workspace.  (Since workspaces resolve their dependencies and features all at once, it's not possible to have different feature sets, so we need to build the `wasm-compat` crate separately from the rest of the workspace anyways).

hdevalence:
I was able to get a variant of this branch working: https://github.com/penumbra-zone/penumbra/pull/2206

conorsch:
Closing in favor of #2206!

Issue #465, opened 2022-03-02
Tracking issue: MVP IBC Relayer infrastructure


hdevalence:
## Overview
In order to relay IBC packets into Penumbra, we need a way to include the packets in transactions (#464), but we also need software that relayers can run to actually create those transactions.

What does this entail? TBD: let's figure this out (cc: @jackzampolin)

## Next steps
- [x] Implement Penumbra support in [relayer](https://github.com/penumbra-zone/relayer)  
  - [x] #2123
  - [x] Resolve buf build logic https://github.com/penumbra-zone/penumbra/issues/1867
  - [x] Resolve ICS23 verification failures https://github.com/penumbra-zone/penumbra/issues/2252
  - [x]  Land PR on upstream main; work w/ Strangelove as necessary on review - https://github.com/cosmos/relayer/pull/1144
- [ ] Perform local testing to exercise relaying functionality
  - [x] Confirm testnet <-> preview config works
  - [ ] Confirm testnet <-> bablyon-testnet works https://github.com/penumbra-zone/penumbra/issues/2284
- [x] Run a relayer as part of testnet deployment
  - [x] Build container image of Penumbra-compatible relayer code - https://github.com/penumbra-zone/penumbra/pull/2298
  - [x] Deploy relayer as part of "testnet-preview"
  - [x] Implement preview -> testnet packet sync

zbuc:
Acceptance criteria: Relayer can execute a full chain of IBC actions

redshiftzero:
Next step here is to use the go buf build process

conorsch:
@avahowell Could you provide an update here about where things stand on the Relayer work?

conorsch:
Update from sync today: still untangling gnarly protobuf build logic. Latest WIP PR is here: https://github.com/penumbra-zone/relayer/pull/2 @avahowell will continue to consult with SL folks and drive progress forward. See relevant discussion in https://discord.com/channels/824484045370818580/930154881040404480/1067611197970595932 Work will continue past shipping #1877.

conorsch:
Based on discussion yesterday, we consider this blocked until #1867 is resolved.

conorsch:
Moved to future column; we'll return to this in a few sprints.

conorsch:
@avahowell For clarity's sake, could you update the OP on this ticket to make it into a epic, with cross-links to other (potentially blocking) issues related to IBC?

conorsch:
[...]

PR #2166, opened 2023-03-15
wip: relayer config generation

conorsch:
Towards #465. Records initial config generation logic for using `rly` (built from the WIP PR at https://github.com/cosmos/relayer/pull/1124). 

The plan is to post transactions between testnet & preview, using the relayer, which must be configured with chain info for both networks, as well as wallet identities. No path configuration logic exists yet; will work with @avahowell to get that straightened out soon.

### DIFF
--- a/deployments/relayer/README.md
+++ b/deployments/relayer/README.md
@@ -1,0 +1,21 @@
+# relayer config
+
+Stores config generation scripts for use with the [relayer](https://github.com/cosmos/relayer),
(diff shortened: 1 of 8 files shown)


hdevalence:
@conorsch I'm going to optimistically merge #2249 into this and this into `main`; although the configs don't work, they don't work for Penumbra-related reasons also, so it seems better to keep them as part of the main repo state and edit them there along with any needed changes to the Penumbra state machine.

PR #2168, opened 2023-03-15
add authorize function to transaction build to allow concurrent proving and authorization

### LINKED ISSUE #1670: Change `TransactionPlan::build` to allow proving to be concurrent with authorization

After switching from mock "transparent" proofs to real zk-SNARKs, building transactions will require proving, which takes user-meaningful amounts of time: aspirationally, our goal would be to get this to be around 1 second. However, any use case with a human-in-the-loop for authorization will already have some authorization latency (while the human reviews the transaction). It would be ideal if we could hide proving latency with authorization latency by running proving concurrently with authorization, rather than having proving latency occur *after* authorization latency, where the user will notice.

Currently, the flow for transaction construction looks like this:
```
┌───────────────┐
│TransactionPlan│
└───────────────┘
        │           ┏━━━━━━━━━━━━━━━┓
        ├──────────▶┃ View Service  ┃
        │           ┗━━━━━━━━━━━━━━━┛
        │                   │           ┏━━━━━━━━━━━━━━━┓
        ├───────────────────┼──────────▶┃Custody Service┃
        │                   │           ┗━━━━━━━━━━━━━━━┛
        │                   │                   │
        ▼                   ▼                   ▼
┌───────────────┐   ┌───────────────┐   ┌───────────────┐
│TransactionPlan│   │  WitnessData  │   │   AuthData    │
└───────────────┘   └───────────────┘   └───────────────┘
        │                   │                   │
        └───────────────────┼───────────────────┘
                            ▼
                    ┌───────────────┐
                    │  Transaction  │
                    └───────────────┘
```
The `TransactionPlan` is sent to the view service to get the `WitnessData` needed for proving, and to the custody service to get the `AuthData` needed to authorize the transaction. Then the `TransactionPlan`, `WitnessData`, and `AuthData` are combined using [`TransactionPlan::build`](https://rustdoc.penumbra.zone/main/penumbra_transaction/plan/struct.TransactionPlan.html#method.build):
```rust
impl TransactionPlan {
    pub fn build<R: CryptoRng + RngCore>(
        self,
        rng: &mut R,
        fvk: &FullViewingKey,
        auth_data: AuthorizationData,
        witness_data: WitnessData,
    ) -> Result<Transaction> { /* ... */ }
}
```

Unfortunately, this means that we have to wait for the auth request to complete **before** we can start proving, even though the auth data isn't used during proving.  Instead, it'd be better to have a flow like this:
```
┌───────────────┐
│TransactionPlan│
└───────────────┘
        │           ┏━━━━━━━━━━━━━━━┓
        ├──────────▶┃ View Service  ┃
        │           ┗━━━━━━━━━━━━━━━┛
        │                   │           ┏━━━━━━━━━━━━━━━┓
        ├───────────────────┼──────────▶┃Custody Service┃
        │                   │           ┗━━━━━━━━━━━━━━━┛
        │                   │                   │
        ▼                   ▼                   ▼
┌───────────────┐   ┌───────────────┐   ┌───────────────┐
│TransactionPlan│   │  WitnessData  │   │   AuthData    │
└───────────────┘   └───────────────┘   └───────────────┘
        │                   │                   │
        └─────────┬─────────┘                   │
                  ▼                             │
         ┌─────────────────┐                    │
         │UnauthTransaction│──────────┬─────────┘
         └─────────────────┘          │
                                      ▼
                              ┌───────────────┐
                              │  Transaction  │
                              └───────────────┘
```
that would allow proving and authorization to happen concurrently.

To do this, we should change the `TransactionPlan::build` API to be a two-step process:
```rust
impl TransactionPlan {
    pub fn build<R: CryptoRng + RngCore>(
        self,
        rng: &mut R,
        fvk: &FullViewingKey,
        witness_data: WitnessData,
    ) -> Result<UnauthTransaction> {
        // Current contents of TransactionPlan::build(), but fills in [0;64] in place of missing auth sigs
    }
}

/// A partially-constructed transaction awaiting authorization data.
struct UnauthTransaction {
    inner: Transaction
}

impl UnathTransaction {
    pub fn authorize(self, auth_data: AuthorizationData) -> Result<Transaction> {
        // Overwrites the placeholder auth sigs with the real ones from `auth_data`
    }
}
```

This leaves the existing API flow almost unchanged for users that don't care about the sequencing: instead of writing
```rust
let tx = plan.build(rng, &fvk, auth_data, witness_data)?;
```
they write
```rust
let tx = plan.build(rng, &fvk, witness_data)?
             .authorize(auth_data)?;
```
and so they could begin proving in advance of authorization.

aubrika:
Closes #1670

### DIFF
--- a/component/src/action_handler/actions/proposal/submit.rs
+++ b/component/src/action_handler/actions/proposal/submit.rs
@@ -300,19 +300,22 @@ const DAO_TRANSACTION_RNG_SEED: &[u8; 32] = b"Penumbra DAO's tx build rng seed";
 
 async fn build_dao_transaction(transaction_plan: TransactionPlan) -> Result<Transaction> {
     let effect_hash = transaction_plan.effect_hash(&DAO_FULL_VIEWING_KEY);
-    transaction_plan.build(
-        &mut ChaCha20Rng::from_seed(*DAO_TRANSACTION_RNG_SEED),
-        &DAO_FULL_VIEWING_KEY,
-        AuthorizationData {
-            effect_hash,
-            spend_auths: Default::default(),
-            delegator_vote_auths: Default::default(),
-        },
-        WitnessData {
-            anchor: penumbra_tct::Tree::new().root(),
-            state_commitment_proofs: Default::default(),
-        },
-    )
+    transaction_plan
+        .build(
+            &DAO_FULL_VIEWING_KEY,
+            WitnessData {
+                anchor: penumbra_tct::Tree::new().root(),
+                state_commitment_proofs: Default::default(),
+            },
+        )?
+        .authorize(
+            &mut ChaCha20Rng::from_seed(*DAO_TRANSACTION_RNG_SEED),
+            &AuthorizationData {
+                effect_hash,
+                spend_auths: Default::default(),
+                delegator_vote_auths: Default::default(),
+            },
+        )
 }
 
 #[cfg(test)]
--- a/crypto/src/symmetric.rs
+++ b/crypto/src/symmetric.rs
@@ -11,9 +11,7 @@ use crate::{
     keys::{IncomingViewingKey, OutgoingViewingKey},
     note,
 };
-use penumbra_proto::core::transaction::v1alpha1::{
-    self as pb, NullifierWithNote, PayloadKeyWithCommitment,
-};
+use penumbra_proto::core::transaction::v1alpha1::{self as pb};
 pub const PAYLOAD_KEY_LEN_BYTES: usize = 32;
 pub const OVK_WRAPPED_LEN_BYTES: usize = 48;
 pub const MEMOKEY_WRAPPED_LEN_BYTES: usize = 48;
--- a/transaction/src/effect_hash.rs
+++ b/transaction/src/effect_hash.rs
@@ -794,7 +794,6 @@ mod tests {
     };
     use penumbra_tct as tct;
     use rand_core::OsRng;
-    use tokio::runtime::Runtime;
 
     use crate::{
         plan::{CluePlan, MemoPlan, OutputPlan, SpendPlan, SwapPlan, TransactionPlan},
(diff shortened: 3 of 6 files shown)


hdevalence:
I haven't checked carefully, but I think the reason CI is failing is that the binding signature is wrong...

Recall that penumbra transactions have three relevant hashes:
- the tx id, the hash of the entire transaction encoding
- the auth hash, the hash of the authorizing data (= the hash of the transaction body, which includes the proofs and auth sigs but excludes the binding signature)
- the effect hash, the hash of the effecting data (which excludes all the zk proofs, auth sigs, and binding signatures)

From inside out, the way to understand this is:
- the effecting data describes the effects the transaction will have on the chain state. it’s hashed into the effect hash
- the auth sigs sign the effect hash: they authorize spending funds in the context of the exact effects the transaction will have on the chain state.
- the auth sigs and the zk proofs authorize the effects of the transaction, so they (together with the effecting data) are hashed together into the auth hash
- the binding sig signs the auth hash, in order to bind the proofs to the specific transaction they were originally part of
- and the nested structure is because the signature can’t sign over itself, so you need to have the “inner” signatures sign over something that excludes them

Ok so, with all of that context in mind, what’s the problem? The binding sig depends on the auth hash, which hashes over the spend auth signatures, so it can’t be computed before they’re inserted.  However, in the current state of the branch, we've replaced all the auth sigs with `[0; 64]` placeholder values, but the binding sig is still being computed in `build()`, so it’s signing over the placeholder values.  This means that when they’re filled in later, the binding signature will be wrong, since it signed a different transaction.

The fix is that the computation of the binding signature has to be delayed until after the auth sigs are inserted. The computation of the binding signature is done just above the end, 165-171, and it takes as input the `synthetic_blinding_factor`, so we’ll want to add a `synthetic_blinding_factor` field to the `UnauthTransaction` struct, that we can save for use computing the binding signature, and do the exact same computation as now — the `// Finally` stanza — but stick it after inserting the auth sigs, in `UnauthTransaction::authorize`, so that it will sign over the correct auth hash. The not-yet-computed binding signature could be filled in with a [0; 64] placeholder value, or alternatively, the `UnauthTransaction` could record the `body` and `anchor` fields separately rather than as a full `Transaction` and not have to have a dummy value.

This is a bit of a subtle point to trip over, but wasn't mentioned in the issue description, and I think the reason is that the issue has been sitting around for a while, and when it was made, the binding signature was made over the effect hash, not the auth hash.  Then we fixed this, so that the binding signature would bind the proofs, but this queued change wasn't top-of-mind then, whoops.

PR #2172, opened 2023-03-16
pcli: make `pcli v tx` display external transactions

### LINKED ISSUE #2171: `pcli v tx` could fetch unknown transactions and attempt to view them

When running `pcli v tx` with a third-party transaction, the command fails:
```
Error: transaction 810d6db339044c512888327bae6e9ed85488c402083a4f5c454325184f0c7c71 not found in view service
```
It might be nice if we fetched it from the chain as a fallback, and then rendered it using the default transaction perspective, in order to illustrate to users what information is disclosed to third parties (by showing them what they can see about *other* people's transactions).

hdevalence:
Closes #2171

### DIFF
--- a/pcli/src/command/view/tx.rs
+++ b/pcli/src/command/view/tx.rs
@@ -1,14 +1,17 @@
 use anyhow::{Context, Result};
 use comfy_table::{presets, Table};
 use penumbra_crypto::{
-    asset::Cache, dex::swap::SwapPlaintext, keys::IncomingViewingKey, Address, FullViewingKey,
-    Note, Value,
+    asset::Cache, dex::swap::SwapPlaintext, keys::IncomingViewingKey, Address, Note, Value,
 };
+use penumbra_proto::{client::v1alpha1::GetTxRequest, DomainType};
 use penumbra_transaction::{
     action::{Swap, SwapClaim},
     view::action_view::{OutputView, SpendView, SwapClaimView, SwapView},
+    Transaction,
 };
 use penumbra_view::ViewClient;
+
+use crate::App;
 
 /// Queries the chain for a transaction by hash.
 #[derive(Debug, clap::Args)]
@@ -176,19 +179,35 @@ impl TxCmd {
     pub fn offline(&self) -> bool {
         false
     }
-    pub async fn exec<V: ViewClient>(&self, fvk: &FullViewingKey, view: &mut V) -> Result<()> {
+    pub async fn exec(&self, app: &mut App) -> Result<()> {
+        let fvk = app.fvk.clone();
         let hash = self
             .hash
             // We have to convert to uppercase because `tendermint::Hash` only accepts uppercase :(
             .to_uppercase()
             .parse()
             .context("invalid transaction hash")?;
-        // Retrieve Transaction
-        let tx = view.transaction_by_hash(hash).await?.ok_or_else(|| {
-            anyhow::anyhow!("transaction {} not found in view service", self.hash,)
-        })?;
+
+        // Retrieve Transaction from the view service first, or else the fullnode
+        let tx = if let Some(tx) = app.view().transaction_by_hash(hash).await? {
+            tx
+        } else {
+            println!("Transaction not found in view service, fetching from fullnode...");
+            // Fall back to fetching from fullnode
+            let mut client = app.tendermint_proxy_client().await?;
+            let rsp = client
+                .get_tx(GetTxRequest {
+                    hash: hex::decode(self.hash.clone())?,
+                    prove: false,
+                })
+                .await?;
+
+            let rsp = rsp.into_inner();
+            let tx = Transaction::decode(rsp.tx.as_slice())?;
+            tx
+        };
         // Retrieve full TxP
-        let txp = view.transaction_perspective(hash).await?;
+        let txp = app.view().transaction_perspective(hash).await?;
         // Generate TxV using TxP
         let txv = tx.view_from_perspective(&txp);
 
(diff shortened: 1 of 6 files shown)


view/src/service.rs
```diff
@@ -513,14 +513,12 @@ impl ViewProtocolService for ViewService {
...
-            .ok_or_else(|| {
-                tonic::Status::failed_precondition(format!(
-                    "No transaction found with this hash {}",
-                    hex::encode(&request.tx_hash)
-                ))
             })?;
 
+        let Some(tx) = maybe_tx else {
+            return Ok(tonic::Response::new(pb::TransactionPerspectiveResponse::default()));
+        };
```
hdevalence, line 520: Not totally happy with this -- what if the view service didn't have the transaction, but it was still viewable in principle?  (i.e., what if we didn't have the current behavior where pcli always downloads every matching transaction before continuing its sync process?)

hdevalence:
Example output:
```
     Running `target/release/pcli v tx 9aa0d147f3678309b987e1bac70c69f890768097d5d1f2426379713d0af0f65f`
Scanning blocks from last sync height 19739 to latest height 19768
[0s] ██████████████████████████████████████████████████      29/29      291/s ETA: 0s
Transaction not found in view service, fetching from fullnode...
 Action Type  Description
 Spend        [?] spent [?]
 Spend        [?] spent [?]
 Spend        [?] spent [?]
 Output       [?] to [?]
 Output       [?] to [?]
 Output       [?] to [?]
 Output       [?] to [?]
 Output       [?] to [?]
 Output       [?] to [?]

 Transaction Fee                0upenumbra
 Transaction Expiration Height  0
 ```

File: view/src/service.rs (hunk 1 above)
Comment (aubrika) [reply view/src/service.rs:520]: Maybe we want something like a, uh, phantom response? Or pending? That way we can indicate this state of affairs to the user at least?

File: pcli/src/command/view/tx.rs [hunk 2]
```diff
@@ -176,19 +179,35 @@ impl TxCmd {
...
+                .get_tx(GetTxRequest {
+                    hash: hex::decode(self.hash.clone())?,
+                    prove: false,
+                })
+                .await?;
+
+            let rsp = rsp.into_inner();
+            let tx = Transaction::decode(rsp.tx.as_slice())?;
+            tx
+        };
```
hdevalence, line 208: Question: should this code be here, or should it be part of the view server implementation?

i.e., should we make the view server's transaction lookup query a fullnode if the transaction isn't found locally?

I think that that might be a bit cleaner than having to have this fallback code in (every) user of the view service.

transaction/src/view/transaction_perspective.rs
```diff
@@ -6,6 +6,7 @@ use std::collections::BTreeMap;
 
 /// This represents the data to understand an individual transaction without
 /// disclosing viewing keys.
+#[derive(Debug, Clone, Default)]
```
hdevalence, line 9: I was surprised that the `Default` impl was missing, I thought we'd added it a while ago on realization that the "default" TxP is the "exterior view" of the implementation.

pcli/src/command/view/tx.rs
aubrika: I think I agree with this. Another option is making a flag to enable/disable this behavior?

view/src/client.rs
```diff
@@ -705,7 +705,14 @@ where
...
             );
-            rsp.await?.into_inner().txp.unwrap().try_into()
+            // If no txp was supplied, use the default (empty) perspective.
+            Ok(rsp
+                .await?
+                .into_inner()
+                .txp
+                .map(|txp| txp.try_into())
+                .transpose()?
+                .unwrap_or_default())
```
hdevalence, line 715: I don't love this code, it feels really non-intuitive to read and understand what's going on. Probably it should be rewritten to not use so many combinators, or rather, it should be rewritten so that the option/result combinators are doing the non-semantically meaningful "data juggling", and have explicit control flow for the semantically meaningful "if no txp supplied, return default" behavior

pcli/src/command/view/tx.rs
hdevalence: Would that flag be in the request, or in the `pclientd` config? It seems better to have in in each individual request, like adding a
```
boolean fetch_if_missing = FIELDNUM;
```
to the request object

aubrika (approved):
LGTM

view/src/service.rs
hdevalence: I think the bigger issue is: currently we make sync speed slower in order to have the property that, if the view service reports it has synced to height `h`, it's guaranteed to have _all_ relevant data up to height `h`.

So, for instance, as soon as we report that we've synced a block, we're guaranteed to have the full transaction info for anything we might have detected.

However, basically every trick that we could do to make sync faster / lighter than it is now would require breaking that assumption.

For instance, we probably don't want to stop scanning once we detect something (because we're waiting to fetch full details in the background), but that means that if we notify someone about a detection, they might not be able to immediately download the relevant transaction.  Is that a problem?

Or, we could do "DAGSync", where as soon as we detect one transaction, we query for transactions spending its output notes, and walk the DAG of our own transaction graph.  This means we'll end up with a "partial" view of sync state up to the chain tip, which we progressively backfill in the background.  What should the query behavior be in that case?

As long as we have the current behavior, we don't really have to worry about it, so the current code in this specific place is currently fine -- but it points to this bigger issue of how we want to handle partially-synced state.

aubrika: we could add an additional function mode? something like "full sync" vs. "fast sync" where fast sync will return some phantom/pending tx info as you describe that returns a partial view of the sync state. Could we make that as an additive change? Because the existing behavior seems worth keeping as an option even if it's more time-consuming

PR #2179, opened 2023-03-16
`pqli q dex liquidity-positions` implementation

zbuc:
This adds a new `pcli q dex liquidity-positions` command with an optional filter (`--only-open`) that will query a node for known liquidity positions and print them in a basic table view.


hdevalence (approved):
This looks good for now. Going forward we'll probably want to rethink it, because we may end up having a lot of liquidity positions, too many to return all of. But we can wait to add those RPCs until we have a clearly identified caller with needs etc.

PR #2183, opened 2023-03-16
#2157: Change semantics of BatchSwapOutputData

### LINKED ISSUE #2157: Update semantics of `BatchSwapOutputData` to support partial fills
Currently, `BatchSwapOutputData` presumes that the entirety of the batch swap will succeed or fail. However as we progressed on implementation of execution, it became clear that we actually need to represent two directional trades for every batch, and so `BatchSwapOutputData` will need to be updated.

We have a trade (1=>2) which we execute in a batch and a trade (2=>1) which we execute in a batch. Our current data structures assume that both of those trades execute at a common clearing price, but in our execution implementation we are going to be doing them as two separate trades.

Concretely, this means:

- [x] Remove the `success` boolean property from `BatchSwapOutputData`
- [x] Update the fields of `BatchSwapOutputData` to contain `lambda_1_1`, `lambda_1_2`, `lambda_2_1`, `lambda_2_2` where:
```
lambda_1_1 <- amount of asset 1 output from (1=>2) trade
lambda_2_1 <- amount of asset 2 output from (1=>2) trade
lambda_1_2 <- amount of asset 1 output from (2=>1) trade
lambda_2_2 <- amount of asset 2 output from (2=>1) trade
```
- [x] Update the `pro_rata_outputs` method to sum the outputs from both directions of the trade (i.e. the pro-rata share of `lambda_1_1` plus the pro-rata share of `lambda_1_2` and the pro-rata share of `lambda_2_1` plus the pro-rata share of `lambda_2_2`)

zbuc:
Closes #2157

### DIFF
--- a/component/src/stubdex/component.rs
+++ b/component/src/stubdex/component.rs
@@ -81,13 +81,13 @@ impl Component for StubDex {
                         state.set_stub_cpmm_reserves(&trading_pair, amm.reserves);
                         (lambda_1, lambda_2, true)
                     }
-                    None => (0, 0, false),
+                    None => (0u64.into(), 0u64.into(), false),
                 };
 
             let (lambda_1_1, lambda_2_2, lambda_2_1, lambda_1_2) = if success {
-                (0, 0, lambda_2, lambda_1)
+                (0u64.into(), 0u64.into(), lambda_2, lambda_1)
             } else {
-                (delta_1, delta_2, 0, 0)
+                (delta_1, delta_2, 0u64.into(), 0u64.into())
             };
 
             let output_data = BatchSwapOutputData {
--- a/component/src/stubdex/component.rs
+++ b/component/src/stubdex/component.rs
@@ -65,6 +65,12 @@ impl Component for StubDex {
             let (delta_1, delta_2) = (swap_flows.0.mock_decrypt(), swap_flows.1.mock_decrypt());
 
             tracing::debug!(?delta_1, ?delta_2, ?trading_pair);
+            // Currently the stub CPMM supports only simple one-directional trades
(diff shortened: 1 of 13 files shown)


component/src/stubdex/component.rs
```diff
@@ -78,14 +84,21 @@ impl Component for StubDex {
                     None => (0, 0, false),
                 };
 
+            let (lambda_1_1, lambda_2_2, lambda_2_1, lambda_1_2) = if success {
+                (0, 0, lambda_2, lambda_1)
+            } else {
+                (delta_1, delta_2, 0, 0)
+            };
```
hdevalence, line 91: This looks right to me

crypto/src/dex/swap.rs
```diff
@@ -29,40 +29,39 @@ pub static DOMAIN_SEPARATOR: Lazy<Fq> =
...
                 .checked_div(self.delta_2 as u128)
                 .unwrap_or(0);
 
-            (lambda_1_i as u64, lambda_2_i as u64)
-        } else {
-            // The swap failed, so the pro rata share is a share of the input amount of
-            // the same token type. But this is exactly the delta_j_i.
-            (delta_1_i, delta_2_i)
-        }
+        (lambda_1_i as u64, lambda_2_i as u64)
```
hdevalence, line 64: This code was written before we had the U128x128, and I think it's probably not going to correctly handle trades with 128-bit inputs (actually, the BSOD can't even currently model those, since it uses `u64` rather than `Amount`s)...

Maybe we should take the opportunity to change the BSOD's data types from `u64` to `Amount`, and then use the `U128x128` to implement the clean versions of the formulas rather than doing manual casting and rounding.

PR #2185, opened 2023-03-17
storage: track jmt keys by their keyhash

erwanor:


erwanor:
I'm going to put this into #2032  directly instead.

PR #2187, opened 2023-03-17
Add fullnode GRPC proxy to `pclientd`

hdevalence:
One point that came up in the IBC meeting with Strangelove is that, for InterchainTest integration (and other use cases), it'd be convenient if `pclientd` could proxy GRPC requests through to a fullnode.  That way, client code only has to have one endpoint to configure and talk to.

This feels like something that should be relatively easy to do with `tower`/`tonic`, so I tried it out (with pointers from @hawkw).

Remaining work:

- [x] Figure out how to convert body types:
```
   = note: expected struct `http::Request<http_body::combinators::box_body::UnsyncBoxBody<tonic::codegen::Bytes, Status>>`
              found struct `http::Request<tonic::transport::Body>`
```
- [x] Figure out how to convert a `Channel` error into an HTTP response with an appropriate status code (the `Service` we plug into the `tonic::Server` has to have `Error = Infallible`, i.e., it has to produce an `Ok(HTTP response with error code)`
- [x] Test the proxy by telling `pcli` that `pclientd` is actually a fullnode, so it makes requests to it.


aubrika (approved):
Great refactor 👍🏻

PR #2189, opened 2023-03-18
pcli: mark `pcli q` as not requiring a chain sync

hdevalence:
It's just querying a fullnode, so there's no need to sync local state, and not waiting to sync local state makes it more useful in debugging (e.g., it's possible to query multiple chains simultaneously).

### DIFF
--- a/pcli/src/command/query/governance.rs
+++ b/pcli/src/command/query/governance.rs
@@ -13,7 +13,6 @@ use penumbra_transaction::{
     proposal::{self, Proposal},
     vote::Vote,
 };
-use penumbra_view::ViewClient;
 use serde::Serialize;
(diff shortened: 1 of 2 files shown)


pcli/src/command/query/governance.rs
```diff
@@ -64,7 +63,7 @@ impl GovernanceCmd {
                     let mut unfinished = client
                         .prefix_value(PrefixValueRequest {
                             prefix: all_unfinished_proposals().into(),
-                            chain_id: app.view().chain_params().await?.chain_id,
+                            ..Default::default()
```
hdevalence, line 66: cc @plaidfinch ; IMO we should keep a separation between `pcli q` and `pcli v` here

PR #2190, opened 2023-03-18
pclientd: fix config file format for FVK

### LINKED ISSUE #2186: `pclientd` config file format serializes the FVK incorrectly
**Describe the bug**

The `config.toml` for `pclientd` to control the test account is
```
[kms_config]
spend_key = 'penumbraspendkey1e9gf5g8jfraap4jqul7e80vv0zrnwpsm4ke0df38ejrfh430nu4s9gc22d'

[fvk]
inner = 'TGKRxEMHLn0WEbna8MjVSHpUk4tXFu8ahPGbgwtC9g8EJpeHx2CmO7narTOJgA5Na1i5nmfDj9dJEdnc9UXHAA=='
```
Instead, it should be
```
fvk = 'penumbrafullviewingkey1.....'
[kms_config]
spend_key = 'penumbraspendkey1e9gf5g8jfraap4jqul7e80vv0zrnwpsm4ke0df38ejrfh430nu4s9gc22d'
```
using the bech32 string format of the full viewing key.  The reason it's using Base64 is that we're falling back to the derived `Serde` format, which is aligned with the protobuf json format.  

To fix this, we should add annotations to the `PclientdConfig` struct that tell it to serialize the FVK as a string.

This will be a tiny change, but a breaking one; let's be sure to get it in before the next `pclientd` release.

hdevalence:
It now uses the string format, like so (shown is the test account):
```toml
fvk = 'penumbrafullviewingkey1f33fr3zrquh869s3h8d0pjx4fpa9fyut2utw7x5y7xdcxz6z7c8sgf5hslrkpf3mh8d26vufsq8y666chx0x0su06ay3rkwu74zuwqq9w8aza'

[kms_config]
spend_key = 'penumbraspendkey1e9gf5g8jfraap4jqul7e80vv0zrnwpsm4ke0df38ejrfh430nu4s9gc22d'
```
Closes #2186

### DIFF
--- a/pclientd/src/lib.rs
+++ b/pclientd/src/lib.rs
@@ -16,7 +16,7 @@ use penumbra_proto::{
 };
 use penumbra_view::{Storage, ViewService};
 use serde::{Deserialize, Serialize};
-use serde_with::serde_as;
+use serde_with::{serde_as, DisplayFromStr};
 
 use std::fs;
 use std::fs::File;
@@ -27,10 +27,11 @@ use tonic::transport::Server;
 #[serde_as]
 #[derive(Serialize, Deserialize, Clone, Debug)]
 pub struct PclientdConfig {
+    /// FVK for both view and custody modes
+    #[serde_as(as = "DisplayFromStr")]
+    pub fvk: FullViewingKey,
     /// Optional KMS config for custody mode
     pub kms_config: Option<soft_kms::Config>,
-    /// FVK for both view and custody modes
-    pub fvk: FullViewingKey,
 }
 
 impl PclientdConfig {


pclientd/src/lib.rs
```diff
@@ -27,10 +27,11 @@ use tonic::transport::Server;
 #[serde_as]
 #[derive(Serialize, Deserialize, Clone, Debug)]
 pub struct PclientdConfig {
-    /// Optional KMS config for custody mode
-    pub kms_config: Option<soft_kms::Config>,
     /// FVK for both view and custody modes
+    #[serde_as(as = "DisplayFromStr")]
     pub fvk: FullViewingKey,
+    /// Optional KMS config for custody mode
+    pub kms_config: Option<soft_kms::Config>,
```
hdevalence, line 34: This has to be reordered for toml reasons (ordinary keys can't follow tables).

PR #2192, opened 2023-03-18
testnets: add an ERC20-like 10e18 asset to Pasiphae allocations.

hdevalence:
We updated the `Amount` type to be 128-bits, but stalled out on eliminating all our previous uses of `u64`s, and on ensuring that we correctly handled 128-bit amount values in the client logic.

This commit adds a `pusd` denom to the testnets (a fake USDC) with a 10e18 display denomination to force the issue, adds it to the base allocations (including the test account), and updates the network integration test to send 20pusd (an amount of the base denom which is greater than u64::MAX).

To merge this PR and ship the Pasiphae allocations, we'll need to ensure that we can actually correctly handle 128-bit amounts.


hdevalence:
We might want to let this slip into the following testnet, but I think this will be the best way to ensure that we can actually handle 128-bit note values.  In the meantime, we can poke at this PR when we have time.

hdevalence:
Closing this out, I think we should aim to do this in the next sprint instead.

PR #2214, opened 2023-03-21
component(dex): implement `PositionRead::positions_by_price stream`

### LINKED ISSUE #2211: Add a `PositionRead` method to stream position IDs by price
We index positions for a given trading pair by price, maintaining a list of positions with nonzero reserves ordered by the effective price.

This index is written manually as part of `PositionManager`'s `Inner::index_position`, but we don't currently have a method on `PositionRead` to read it.  This API should probably look like:
```
fn positions_by_price(&self, pair: DirectedTradingPair) -> Pin<Box<dyn Stream<Item=position::Id> + Send + 'static>>;
```
Internally, this would call `StateRead::nonconsensus_prefix_raw` to get the stream and then map the results to a `position::Id`.

erwanor:
Closes #2211

### DIFF
--- a/component/src/dex/position_manager.rs
+++ b/component/src/dex/position_manager.rs
@@ -25,8 +25,7 @@ pub trait PositionRead: StateRead {
         self.nonconsensus_prefix_raw(&prefix)
             .map(|entry| match entry {
                 Ok((k, _)) => {
-                    let raw_id=
-                        <&[u8; 32]>::try_from(&k[103..135])?.to_owned();
+                    let raw_id = <&[u8; 32]>::try_from(&k[103..135])?.to_owned();
                     Ok(position::Id(raw_id))
                 }
                 Err(e) => Err(e),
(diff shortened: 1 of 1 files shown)


component/src/dex/position_manager.rs
```diff
@@ -11,9 +11,29 @@ use penumbra_proto::{DomainType, StateReadProto, StateWriteProto};
...
+    async fn positions_by_price(
+        &self,
+        pair: DirectedTradingPair,
+    ) -> Pin<Box<dyn Stream<Item = Result<position::Id>> + Send + 'static>> {
+        let prefix = state_key::internal::price_index::prefix(&pair);
+        self.nonconsensus_prefix_raw(&prefix)
+            .map(|entry| match entry {
+                Ok((k, _)) => {
+                    let raw_id=
+                        <&[u8; 32]>::try_from(&k[103..135])?.to_owned();
```
erwanor, line 29: I think it might be nice to encapsulate the decoding of keys into `state_key`, now that we use the kv store as a set, but deferring this to later.

PR #2215, opened 2023-03-22
switch from static to dynamic epochs

### LINKED ISSUE #2181: Move from static to dynamic epochs
Towards #2070 and following refactoring the SCT and compactblock handling out of the shielded pool, we want to change from static to dynamic epochs without moving to a time-based system. This "breaks" the economic mechanisms, but they are already broken, so we don't really lose anything. 

- [x] Determine how we want to change the Epoch data modeling. Currently, this holds an index and a duration. Should we change to just be an index, or an index, start_height, end_height?
- [x] Determine how we want the chain to record the current epoch, and the history of epoch transitions. Should we have a key-value mapping like "chain/epoch_by_height/{height:010}"  -> epoch_index, using 0-padding so that the lex order of keys we use in iteration is our numeric ordering? What RPC methods do we need to allow clients to query for current or historical epochs, now that they can't compute them locally?
- [x] Replace all uses of Epoch::from_height with these new APIs.
Update the client code in the MockClient and in the Rust view server to check if an epoch root is present in the compact block and use its presence or absence to determine whether to trigger an epoch transition in its copy of the SCT, now that they are dynamic.
  - [ ] Add a test to the `MockClient` that ensures it can maintain SCT synchronization across epoch boundaries
- [x] Ensure validator changes only happen at epoch boundaries, by changing the code so that slashing triggers an ad-hoc epoch boundary (how should the staking component signal an epoch end?) and only sending validator updates at epoch boundaries.

avahowell:
Closes #2181


avahowell:
Should be ready for review

chain/src/epoch.rs
```diff
@@ -1,62 +1,39 @@
...
-    type Output = Epoch;
-
-    fn add(self, other: u64) -> Self::Output {
-        Epoch {
-            index: self.index + other,
-            duration: self.duration,
-        }
+impl Epoch {
+    // NOTE: this should be changed when we move to time-based mechanisms
+    pub fn is_epoch_end(&self, current_height: u64, epoch_duration: u64) -> bool {
```
hdevalence, line 36: Even if we're planning to change it later, maybe this should be `is_scheduled_epoch_end`, to avoid confusion? Or we could move the logic into a private method on the `App`, so that we can't have any other part of the code predict when epochs will end (and thereby become surprised if they end early).

chain/src/view.rs
```diff
@@ -112,11 +113,18 @@ pub trait StateReadExt: StateRead {
...
 
-        // The current epoch
-        Ok(Epoch::from_height(height, epoch_duration))
+    /// Gets the pending epoch change, if any
+    async fn pending_epoch(&self) -> Result<Option<Epoch>> {
+        Ok(self
+            .get(&state_key::epoch_change_at_height(
+                self.get_block_height().await?,
+            ))
+            .await?)
```
hdevalence, line 127: What's the meaning of a pending epoch change?

chain/src/state_key.rs
```diff
@@ -27,3 +27,11 @@ pub fn halt_now() -> &'static str {
 pub fn chain_params_changed() -> &'static str {
     "chain_params_changed"
 }
+
+pub fn epoch_by_height(height: u64) -> String {
+    format!("chain/epoch_by_height/{}", height)
+}
```
hdevalence, line 33: I started wondering about how we could avoid storing a height => epoch lookup for every height, but I think that the current storage APIs are a bit limiting, in that they only give us prefix queries rather than real range operations (even though those operations could be implemented). So this seems reasonable to do in the meantime.

chain/src/view.rs
```diff
@@ -150,11 +164,32 @@ pub trait StateWriteExt: StateWrite {
...
+            index: current_epoch.index + 1,
+            start_height: change_height,
+        };
+
+        tracing::info!(%change_height, "scheduling epoch change at next block");
+
+        self.put(state_key::epoch_change_at_height(change_height), new_epoch);
+
+        Ok(())
+    }
```
hdevalence, line 182: Do we want to be scheduling epoch changes and committing the schedule to the chain state, rather than just having the `App` suddenly decide to roll over the epoch? What is going to be consuming the schedule?

component/src/dex/tests.rs
```diff
@@ -29,6 +29,13 @@ mod test {
         let height = 1;
 
         state_tx.put_block_height(height);
+        state_tx.put_epoch_by_height(
+            height,
+            penumbra_chain::Epoch {
+                index: 0,
+                start_height: 0,
+            },
+        );
```
hdevalence, line 38: Do we ever want to set the block height without also setting the epoch? If not, should we fold both of these methods into a `put_block_height_and_epoch` method?

proto/proto/penumbra/client/v1alpha1/client.proto
```diff
@@ -25,6 +25,7 @@ import "tendermint/types/validator.proto";
 service ObliviousQueryService {
   rpc CompactBlockRange(CompactBlockRangeRequest) returns (stream CompactBlockRangeResponse);
   rpc ChainParameters(ChainParametersRequest) returns (ChainParametersResponse);
+  rpc Epoch(EpochRequest) returns (EpochResponse);
```
hdevalence, line 28: I think we might want this to be
```
rpc EpochByHeight(EpochByHeightRequest) returns (EpochByHeightResponse);
```
in case we later want to query for the start/end or other epoch data by some otehr means (e.g. epoch data by index)

chain/src/view.rs
avahowell: This was meant to convey to begin_block that it should increment the epoch, but I changed this so that the epoch is set by the app for the next block every time in `end_block instead, so this is no longer needed

chain/src/view.rs
avahowell: [1fca310](https://github.com/penumbra-zone/penumbra/pull/2215/commits/1fca310e1abcd78daf123fe599950b9d99654ae6) changed this to using the `App` to roll over the epoch

component/src/dex/tests.rs
avahowell: In `init_chain`, we want to set the epoch for both the current block (0) and the first block (1), since we only set the epoch in `end_block`, and `end_block` isn't called after init chain on block 0

hdevalence:
This looks good, nice work!

PR #2222, opened 2023-03-23
Use URLs and SocketAddrs for CLI args

### LINKED ISSUE #2196: Align command-line interfaces around endpoint URIs (outbound) and socket addresses (inbound)

Currently, the command-line interfaces for pd, pclientd, and pcli have a disjointed mix of addresses and ports. This is a problem, both because all the tools have a slightly different interface, but also and more importantly because it means it's not easy to handle different protocols (http vs https).

We should go through the command-line arguments for
- [ ] pd
- [ ] pclientd
- [ ] pcli
and modify them so that
- all _inbound_ addresses (eg a service hosted by that program) are specified as SocketAddr, which includes both the address and port in one item
- all _outbound_ addresses are specified as a [Uri](https://docs.rs/tonic/latest/tonic/transport/struct.Uri.html), allowing specification of the protocol, address, and port all in one item

Then, we'll need to propagate those changes through the smoke test script until it passes CI.

### LINKED ISSUE #2240: pclientd docs are out of date
The [pclientd docs in the guide](https://guide.penumbra.zone/main/pcli/pclientd.html) appear to be out of date.

* `pclientd init FVK_STRING` will throw an error; use `pclient init --home /path/to/state FVK_STRING`
* `pclientd start` will throw an error; use `pclientd --home /path/to/state start`
* the `-s` parameter doesn't exist; it's been renamed to `--home`.

Additionally the `pcli -v 127.0.0.1:8081` command should probably be expanded to a working invocation, e.g. `pcli -v 127.0.0.1:8081 q validator list`.

conorsch:
Updates the CLI args for pcli, pd, and pclientd to use `net::SocketAddr` types for local bind addresses, and `url::Url` types for remote endpoint specification. This makes the CLI args more verbose, but also more explicit and more flexible. For instance, it's now possible to bind pd's metrics port to localhost, but the grpc pd service to a different interface.
    
Also adds env var options for the new CLI args, to standardize config from other wrapper tools and testing environments.

To see the new options at a glance:

```
cargo build --release
./target/release/pcli -h
./target/release/pclientd -h
./target/release/pclientd start -h
./target/release/pd start -h
./target/release/pd testnet join -h
```

Closes #2196. Closes #2240.

### DIFF
--- a/deployments/helm/templates/fn-deployments.yaml
+++ b/deployments/helm/templates/fn-deployments.yaml
@@ -163,8 +163,10 @@ spec:
             - start
             - --home
             - /home/pd
-            - --host
-            - 0.0.0.0
+            - --grpc-bind
+            - 0.0.0.0:8080
+            - --metrics-bind
+            - 0.0.0.0:9000
         - name: health-check
           image: "{{ $.Values.health.image }}:{{ $.Values.health.version }}"
           imagePullPolicy: IfNotPresent
--- a/pclientd/src/lib.rs
+++ b/pclientd/src/lib.rs
@@ -1,4 +1,5 @@
 use std::env;
+use std::net::SocketAddr;
 use std::path::Path;
 
 use anyhow::Result;
@@ -23,6 +24,7 @@ use std::fs::File;
 use std::io::{self, BufRead, Write};
 use std::str::FromStr;
 use tonic::transport::Server;
+use url::Url;
 
 mod proxy;
 pub use proxy::{ObliviousQueryProxy, SpecificQueryProxy, TendermintProxyProxy};
@@ -61,19 +63,16 @@ pub struct Opt {
     #[clap(subcommand)]
     pub cmd: Command,
     /// The path used to store pclientd state and config files.
-    #[clap(long)]
+    #[clap(long, env = "PENUMBRA_PCLIENTD_HOME")]
     pub home: Utf8PathBuf,
-    /// The address of the pd+tendermint node.
+    /// The URL of the gRPC endpoint for pd.
     #[clap(
         short,
         long,
-        default_value = "testnet.penumbra.zone",
-        env = "PENUMBRA_NODE_HOSTNAME"
+        default_value = "http://testnet.penumbra.zone:8080",
+        env = "PENUMBRA_NODE_PD_URL"
     )]
-    pub node: String,
-    /// The port to use to speak to pd's gRPC server.
-    #[clap(long, default_value = "8080", env = "PENUMBRA_PD_PORT")]
-    pub pd_port: u16,
+    pub node: Url,
 }
 
 #[derive(Debug, clap::Subcommand)]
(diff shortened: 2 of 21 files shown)


conorsch:
Needs a bit more work on the `testnet join` args handling; keeping as draft for now.

conorsch:
Fixed up the `testnet join` flow. Still remaining is interactive testing of the new args, after which I'll update the docs.

conorsch:
OK, marking as ready for review! The changes are fairly straightforward, with the exception of the external-address support in `pd testnet join`; 

```
# old version:
pd testnet join --external-address 1.2.3.4
# new version
pd test join --external-address tcp://1.2.3.4:26656
```

That's significantly less ergonomic, but it's consistent with the URL-versus-socket terminology used throughout, so I think it's worth it. We'll need callouts in the release notes so folks can update their scripts.

deployments/helm/templates/fn-deployments.yaml
```diff
@@ -163,8 +163,10 @@ spec:
             - start
             - --home
             - /home/pd
-            - --host
-            - 0.0.0.0
+            - --grpc-bind
+            - 0.0.0.0:8080
+            - --metrics-bind
+            - 0.0.0.0:9000
```
conorsch, line 169: Post-merge I'll check on the preview deploy to confirm these new values are working as expected.

pd/src/main.rs
```diff
@@ -114,18 +132,35 @@ enum TestnetCommand {
...
+            default_value = "0.0.0.0:26657"
+        )]
+        tendermint_rpc_bind_addr: SocketAddr,
+        /// When generating Tendermint config, use this socket to bind the Tendermint P2P service.
+        #[clap(
+            long,
+            env = "PENUMBRA_PD_TM_P2P_BIND_ADDR",
+            default_value = "0.0.0.0:26656"
+        )]
+        tendermint_p2p_bind_addr: SocketAddr,
```
hdevalence, line 163: This may not matter, but one thing I noticed looking at the `pd start` options is that there was a consistent terminology between `foo_bind` (specifying inbound addresses) and `foo_addr` (specifying outbound addresses), so that you can tell just by glancing at the name whether a parameter is an inbound or outbound address.  Should we align these with that, and make them `tendermint_foo_bind` rather than `tendermint_foo_bind_addr`?

pclientd/src/lib.rs
```diff
@@ -88,12 +87,13 @@ pub enum Command {
...
-        /// Bind the view gRPC server to this port.
-        #[clap(long, default_value = "8081")]
-        view_port: u16,
+        /// Bind the view service to this socket.
+        #[clap(
+            long,
+            env = "PENUMBRA_PCLIENTD_BIND_ADDR",
+            default_value = "127.0.0.1:8081"
+        )]
+        bind_addr: SocketAddr,
```
hdevalence, line 96: ditto https://github.com/penumbra-zone/penumbra/pull/2222/files#r1148237861

conorsch:
Updated to standardize the `bind/addr` naming. I also pushed another commit rolling back the awkward `tcp://1.2.3.4:26656` for the `--external-address` flag; it now takes a socketaddr, so `1.2.3.4:26656`. Either way is a bit of a compromise, so I made sure to keep that change in a separate commit. If anyone feels strongly, feel free to drop it.

PR #2226, opened 2023-03-24
wip: change unbonding mechanism so end_epoch is chosen by chain

### LINKED ISSUE #2224: Change the unbonding token mechanism so that the unbonding end epoch is chosen by the chain
Currently, these are statically scheduled by the user when they create the unbonding tokens, but we'll need to have the end epoch chosen by the chain. In other words, rather than unbonding tokens being parameterized by the start and end epoch indexes, they'll only be parameterized by the start epoch, and the chain will be responsible for determining whether finishing the unbonding is allowed in the current epoch.

avahowell:
closes #2224

### DIFF
--- a/proto/proto/penumbra/core/stake/v1alpha1/stake.proto
+++ b/proto/proto/penumbra/core/stake/v1alpha1/stake.proto
@@ -146,16 +146,14 @@ message Undelegate {
   crypto.v1alpha1.IdentityKey validator_identity = 1;
   // The index of the epoch in which this undelegation was performed.
   uint64 start_epoch_index = 2;
-  // The index of the epoch in which unbonding should complete.
-  uint64 end_epoch_index = 3;
   // The amount to undelegate, in units of unbonding tokens.
-  crypto.v1alpha1.Amount unbonded_amount = 4;
+  crypto.v1alpha1.Amount unbonded_amount = 3;
   // The amount of delegation tokens consumed by this action.
   //
   // This is implied by the validator's exchange rate in the specified epoch
   // (and should be checked in transaction validation!), but including it allows
   // stateless verification that the transaction is internally consistent.
-  crypto.v1alpha1.Amount delegation_amount = 5;
+  crypto.v1alpha1.Amount delegation_amount = 4;
 }
 
 // A transaction action finishing an undelegation, converting (slashable)
--- a/crypto/src/stake/unbonding_token.rs
+++ b/crypto/src/stake/unbonding_token.rs
@@ -13,28 +13,22 @@ use crate::asset;
 pub struct UnbondingToken {
     validator_identity: IdentityKey,
     start_epoch_index: u64,
-    end_epoch_index: u64,
     base_denom: asset::Denom,
 }
 
 impl UnbondingToken {
-    pub fn new(
-        validator_identity: IdentityKey,
-        start_epoch_index: u64,
-        end_epoch_index: u64,
-    ) -> Self {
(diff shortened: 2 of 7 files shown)


pcli/src/command/tx.rs
```diff
@@ -570,7 +553,6 @@ impl TxCmd {
                             .undelegate_claim(UndelegateClaimPlan {
                                 validator_identity,
                                 start_epoch_index,
-                                end_epoch_index,
                                 penalty,
```
avahowell, line 556: why do we have the client send the penalty at all? as opposed to just having the chain compute it

hdevalence: because the penalty is an input to proof verification, and we want to be able to do all of the cryptographic verification up front, before doing any state accesses

avahowell: should the end epoch index then still be a part of the undelegate claim? as of this change, the client calculates the penalty assuming that the current epoch is >= the (chain enforced) end epoch

hdevalence: i think it's fine to just include the penalty, since that's enough to verify the proofs, we can do some state access + computation to enforce that the supplied penalty is the correct one

component/src/action_handler/actions/undelegate.rs
```diff
@@ -36,23 +36,6 @@ impl ActionHandler for Undelegate {
             ));
         }
 
-        // Check whether the end epoch is correct for the given validator (it
-        // may already be unbonding).  This will only differ from the generic
-        // unbonding time if the validator is already unbonding, so we only
-        // require that the declared unbonding period is at least as long as the
-        // expected one, to allow clients the option of skipping a query to
-        // discover if the validator is already unbonding.
```
avahowell, line 44: i'm not as keyed into the unbonding token mechanisms, but i think this is the correct change? this PR just removed this validation, and replaced it with a check in `undelegate_claim` that verifies that at least `unbonding_epochs` have passed since the undelegate action was triggered

component/src/action_handler/actions/undelegate_claim.rs
```diff
@@ -42,7 +43,7 @@ impl ActionHandler for UndelegateClaim {
             .compounded_penalty_over_range(
                 &self.body.validator_identity,
                 self.body.start_epoch_index,
-                self.body.end_epoch_index,
+                end_epoch_index,
             )
```
hdevalence, line 47: @avahowell per your question in discord: this is where the penalty is computed -- all that matters here is what slashings occurred during the epoch, not how many there were.

proto/proto/penumbra/core/stake/v1alpha1/stake.proto
```diff
@@ -146,16 +146,14 @@ message Undelegate {
   crypto.v1alpha1.IdentityKey validator_identity = 1;
   // The index of the epoch in which this undelegation was performed.
   uint64 start_epoch_index = 2;
-  // The index of the epoch in which unbonding should complete.
-  uint64 end_epoch_index = 3;
   // The amount to undelegate, in units of unbonding tokens.
-  crypto.v1alpha1.Amount unbonded_amount = 4;
+  crypto.v1alpha1.Amount unbonded_amount = 3;
```
hdevalence, line 150: It's not worth undoing, but in general it's better to leave the other field numbers unchanged, and just leave a gap, because that way parsing can be backwards/forwards compatible at the data interchange level.

component/src/action_handler/actions/undelegate.rs
hdevalence: LGTM!

PR #2238, opened 2023-03-24
ibc: try changing event formatting

hdevalence:


hdevalence:
During relayer debugging, we noticed that the `update_client` event was encoding the updated header using a string-encoded debug format, instead of hex-encoded proto bytes as expected by the relayer.  This commit is branched off of `048-carme.1` so that it can be run on the existing testnet.

PR #2245, opened 2023-03-25
IBC improvements

hdevalence:
This PR makes a few changes, based on observations doing IBC debugging:

- It adds an `IbcAction` domain type.  This is still represented in protos as an `Any`, but it means that we can only parse once, in our Rust code.
- It adds an `IbcAction::create_span` method, similar to `Action::create_span`, that adds contextual information while processing the IBC action.
- It refactors the `ActionHandler` implementations for IBC actions, so that we're able to have multiple IBC packets in a single transaction.


hdevalence:
Merging this optimistically to deploy to `testnet-preview`.

PR #2250, opened 2023-03-27
pd: use tower-actor for Consensus, Mempool

hdevalence:
Now that we have a basic implementation of `tower-actor`, this PR replaces our existing ad-hoc `Service` impls with it.


pd/Cargo.toml
```diff
@@ -28,6 +28,7 @@ penumbra-wallet = { path = "../wallet" }
 decaf377 = {version = "0.3", features = ["parallel"]}
 tower-abci = "0.5.0"
 jmt = "0.3"
+tower-actor = { git = "https://github.com/penumbra-zone/tower-actor", branch = "main" }
```
hdevalence, line 31: TODO: this should use a published version of the crate, I just wanted to check that it could typecheck first.

Issue #2218, opened 2023-03-22
Docker builds are failing due to recursion error on pd


conorsch:
See [example failing build](https://github.com/penumbra-zone/penumbra/actions/runs/4494704940/jobs/7907512701):

```
#24 213.5    Compiling penumbra-transaction v0.1.0 (/usr/src/transaction)
#24 ...

#49 [linux/amd64->arm64 build-env 7/9] RUN if [ "arm64" = "arm64" ] && [ "amd64" != "arm64" ]; then       export CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER=aarch64-linux-gnu-gcc         CC_aarch64_unknown_linux_gnu=aarch64-linux-gnu-gcc         CXX_aarch64_unknown_linux_gnu=aarch64-linux-gnu-g++         PKG_CONFIG_SYSROOT_DIR=/usr/aarch64-linux-gnu;       cargo build --release --target aarch64-unknown-linux-gnu;     elif [ "arm64" = "amd64" ] && [ "amd64" != "amd64" ]; then       export CARGO_TARGET_x86_64_UNKNOWN_LINUX_GNU_LINKER=x86_64-linux-gnu-gcc         CC_x86_64_unknown_linux_gnu=x86_64-linux-gnu-gcc         CXX_x86_64_unknown_linux_gnu=x86_64-linux-gnu-g++         PKG_CONFIG_SYSROOT_DIR=/usr/x86_64-linux-gnu;       cargo build --release --target x86_64-unknown-linux-gnu;     else       cargo build --release --target $(uname -m)-unknown-linux-gnu;    fi;
#49 191.2    Compiling sqlx v0.6.2
#49 ...

[...]

PR #2257, opened 2023-03-27
fix: raise pd recursion limit 256 -> 512

conorsch:
Refs #2218.


conorsch:
I kicked off an ad-hoc run of the container workflow here: https://github.com/penumbra-zone/penumbra/actions/runs/4534103763 to evaluate whether the new limit is sufficient.

conorsch:
Didn't work. Trying again with the recursion limit added to main.rs. https://github.com/penumbra-zone/penumbra/actions/runs/4534350606 If that doesn't work, will try another power of 2.

PR #2265, opened 2023-03-27
#2223: Remove use of view service from `pcli q` commands

### LINKED ISSUE #2223: Update existing `pcli q dex` commands to avoid use of view service
Currently, the various `pcli q dex` commands rely on the view service, primarily to fetch the asset list and chain ID.

It should be possible to instead use the oblivious query service's asset list for fetching the asset cache.

Also, it's probably unnecessary to pass in the "expected" chain ID in the `pcli q dex` commands after this change is made, since there will be no view service interaction any more, and no synced chain data to worry about.

zbuc:
This allows the various `pcli q` commands to run entirely in offline mode.

I realized after writing this that I possibly could have used the existing `TendermintProxyService` rather than adding an `info` call proxy to `ObliviousQueryService`, but this can be cleaned up in a subsequent PR.

Closes #2223

### DIFF
--- a/chain/src/known_assets.rs
+++ b/chain/src/known_assets.rs
@@ -1,4 +1,4 @@
-use penumbra_crypto::asset::Asset;
+use penumbra_crypto::asset::{self, Asset};
 use penumbra_proto::{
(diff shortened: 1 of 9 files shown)


hdevalence (approved):
LGTM!

PR #2266, opened 2023-03-28
refactor: impl `AllocVar` for all TCT R1CS variables

### LINKED ISSUE #2040: crypto: clean up the TCT R1CS gadgets
This is low priority but the TCT R1CS gadgets are messy, it would be good to instead of using a `new()` method implement the `AllocVar` trait as we do for our other R1CS variables and also `impl R1CSVar` for each variable in the TCT crate (nice for debugging as it lets one access the inner value).

redshiftzero:
closes #2040


hdevalence (approved):

hdevalence:
@redshiftzero does this require circuit regeneration?

redshiftzero:
Nope, no circuit-breaking changes in this

PR #2268, opened 2023-03-28
pd: respond to queries with specified height

hdevalence:
This may be the cause of IBC issues.


conorsch:
Pushed a fix-it commit for the `buf mod update`. Let's merge optimistically after CI passes, so the height logic makes it into 49.

PR #2280, opened 2023-03-29
build: straighten out feature flags

conorsch:
Several of our crates failed to build under certain combinations of feature flags. Stepped through those crates and adjusted the feature logic where required.
    
Regarding runtime for this CI check, the wall time isn't bad, especially if we parallelize it with the other rust checks:
    
      command: cargo check-all-features --release
      warm cache: 1m39s
      cold cache: 12m4s
    
Notably, the disk usage is an order of magnitude higher:
    
      dir size of `target/`
      cargo build --release: ~3.4GB
      cargo check-all-features --release 25G(!)
    
In order to support that additional disk usage in CI caching layers, we'll likely have to pony up for more space on our CI runners.  Still, that's well worth the confidence that we have working builds across all feature combinations.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,6 @@ target/
 
 # Python3
 .venv
+
+# Ignoring lock file for the parameter-setup tool.
+tools/parameter-setup/Cargo.lock
--- a/chain/Cargo.toml
+++ b/chain/Cargo.toml
@@ -31,6 +31,6 @@ tracing = "0.1"
 num-rational = "0.4"
 
 [features]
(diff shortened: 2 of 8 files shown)


hdevalence:
I think the bigger problem with the space is that we pull all of that data to load every time. I am pretty hesitant about adding this to CI, and wonder if we could get most of the value by just occasionally running the command locally. If nobody notices any breakage, it's because nobody is actually using that combination of features, so having occasional gaps seems OK, at least up until someone actually runs into the problem at least once.

PR #2282, opened 2023-03-30
workspace: rename `penumbra-component` to `penumbra-app`

hdevalence:
Currently, the Penumbra code is poorly organized, as a result of us stumbling around trying to figure out the right way to organize our code.  As a result, we've ended up with pretty much everything in the `penumbra-component` crate. However, we now see a path to cut the gordian knot of internal dependencies, to be written up in a separate issue.

This commit does the first and most invasive step to unlock gradual refactoring over time: renaming the current `penumbra-component` crate to `penumbra-app`. This recasts the current crate as the top-level "application" crate, allowing us to gradually peel out functionality bit by bit.


conorsch:
Squashed in a rename on the `include_str` call in the integration tests, to get CI passing.

hdevalence:
Thanks, squashed in a `cargo fmt` on top of that path fix

Issue #2216, opened 2023-03-22
Host static site for Penumbra webapp client


conorsch:
Let's take the static site here: https://github.com/penumbra-zone/penumbra_dapp, recently transferred by the Zpoken team to the penumbra-zone org, and host it as a static website via Firebase, same as our other sites. We want automatic redeploys on changes to `main`. 

## Progress
- [x] Choose URL: `app.testnet.penumbra.zone`
- [x] Add Firebase config and update DNS
- [x] Document build steps for compiling the site (Zpoken is working on this)
- [x] Confirm build works locally, and static site can serve locally cleanly
- [x] Add CI workflow to publish site via GH actions from that repo
- [x] Perform interactive testing on deployed site

conorsch:
Marking as blocked due to 

> Document build steps for compiling the site (Zpoken is working on this)

I'll plan to pick this up next sprint.

conorsch:
The site repo has a Dockerfile, which currently fails on the `yarn build` step. I assume that's because the `penumbra-web-assembly` npm dependency must be updated with the latest code from the renamed `penumbra-wasm` in #2206. The Zpoken team maintains that module; let's make sure to add a note to the site README about that critical piece.

conorsch:
Refs https://github.com/penumbra-zone/wallet/issues/2

conorsch:
Progress in the upstream repo: as of https://github.com/penumbra-zone/penumbra_dapp/commit/eda5a42c3a9b2aff67e44932e3d138452f704a1e, I can build a container with the static site docroot inside of it. Loading the site just shows a splash page though:

![penumbra-dapp-splash](https://user-images.githubusercontent.com/657862/228060144-99797a4c-5f27-43b5-9b24-88a10b238415.png)

Will ask for more updates in chat.

conorsch:
The following steps worked for me:

1. Install extension from Chrome store
2. Create a new wallet inside the extension flow
3. Send some funds to that new wallet via `pcli`
4. Spin up localhost webserver for the `penumbra_dapp` docroot
5. Navigate to localhost webserver, connect to wallet
6. Voila!

![penumbra-wallet-dapp-demo](https://user-images.githubusercontent.com/657862/228358297-cd3734a4-c24d-4af4-a932-43517eaa64ab.png)

Will work on deploying that static site to our domain next.

conorsch:
We're unblocked on docs for building, so I'm going to work on rehosting the site.

[...]

PR #2293, opened 2023-04-03
docs: include wasm in rustdoc

### LINKED ISSUE #2275: Add `penumbra-wasm` to Rustdoc actions
Let's add the `penumbra-wasm` crate to the list of crates built with Rustdoc, so it shows up in the index.

conorsch:
Pulls the wasm crate into the parent workspace, so `cargo doc` and associated scripting can find it.

In the process, I promoted the release profile setting of `opt-level=s`, which optimizes for minimal binary size [0], from the wasm crate Cargo.toml to the workspace. I compared output sizes (but not runtime metrics) between the two profiles, and saw that pcli is smaller, but pd grew slightly. The `.wasm` artifacts were indeed slightly smaller with opt-level=s. In the future, we can further optimize with e.g. lto=true, but I'm not doing that here because the goal is publishing docs for wasm, and that's accomplsihed.

Closes #2275.

[0] https://doc.rust-lang.org/cargo/reference/profiles.html#opt-level
[1] https://rustwasm.github.io/book/reference/code-size.html#optimizing-builds-for-code-size


wasm/Cargo.toml
```diff
@@ -37,8 +37,3 @@ wasm-bindgen-test = "0.3.13"
 
 [profile.release]
 opt-level = "s"
-
-# We set a blank 'workspace' block disassociate this crate
-# with the workspace at the repo root. Doing so models
-# how an external developer would use it.
-[workspace]
```
hdevalence, line 44: without necessarily assuming this was good to start, why don't we need it now?

conorsch: In order for our docs-generation scripting to include the `penumbra-wasm` crate, we must pull it into the top-level workspace. Otherwise:

```
error: package ID specification `penumbra-wasm` did not match any packages
```

While we could rig up some special-casing where we build the `wasm` docs separately and then edit the index HTML to include it, it doesn't seem worthwhile. We'd do better to have the docs integrated with the other crates on rustdoc, and flesh out the tests and functionality in the wasm crate.

hdevalence: Right, that makes sense, what I meant was more, are we giving up anything important by moving it into the workspace?

conorsch: No, I don't think we lose much: it was always aspirational to have a way to test how the wasm bindings would be used, especially when they lived in a different repo (#1990). Now that we have the dapp site built using the wasm bindings published to npm (#2216), that's a better canary for the build pipeline, because it's actually being used in a deploy. The Zpoken folks are still maintaining the publish-to-npm workflow, but I don't see a need for us to try to host that.

hdevalence:
I don't think I follow what this PR does with the `opt-level` parameter -- it seems like it's still restricted to the `wasm` crate config, does that get lifted to workspace-wide config? If so, I think we should drop it, and only put the setting in the script that's actually building the wasm code.

PR #2297, opened 2023-04-04
Add stub narsil crate to workspace (#2260)

### LINKED ISSUE #2260: Add a stub `narsil` crate to the workspace
This should be structured as a library, with

- [x] A stub ledger implementation in the library that uses `penumbra_storage` as a backing store and can process ABCI messages using `tower_actor`
- [x] A `narsild` binary (for the server) that spins up the stub ledger implementation
- [x] A `narsilcli` binary to poke at it during development; this can be empty to start.

zbuc:
Closes #2260


hdevalence:
Pulling out the `penumbra-tower-trace` code is a good idea, looks like it might have been missed from the commits and the workspace though?

Issue #2210, opened 2023-03-21
CI check to run build on all features


redshiftzero:
It could be good now that we have feature flags to enable: https://crates.io/crates/cargo-all-features

For example the `penumbra-wallet` crate doesn't compile on `main` without the `parallel` feature (I first noticed when testing on 3a5aa9464ab5c1db85b3fd9b83204698029f6a50, but it was not introduced in that commit).

If this check is slow, could be fine to just do on `main` as a guard against releasing breaking changes to downstream users.

conorsch:
Interesting! Tried a quick `cargo check-all-features` on `main` and it failed. Looks like there might be some small cleanup we can do on the edge cases; otherwise, we can ignore certain conflicting feature flags in the package metadata. I'll take a closer look next sprint and document results here.

Issue #2184, opened 2023-03-17
Replace `tools/proto-compiler` with `buf gen` and `protoc-gen-prost`


hdevalence:
**Is your feature request related to a problem? Please describe.**

As described in #1867, we have two different variants of protobuf generation tooling, one using Buf and the other using prost/tonic.  That issue described a strategy for trying to hack together `prost` and the Buf cache.  An almost-certainly better alternative would be to use Buf to generate both kinds of code, and instruct it on how to generate Rust code using https://github.com/neoeinstein/protoc-gen-prost  (details on Buf integration here: https://github.com/neoeinstein/protoc-gen-prost/blob/main/protoc-gen-prost-serde/README.md#usage-with-buf )

**Describe the solution you'd like**

- Replace `tools/proto-compiler` with Buf generate
- We'll need a combination of
  - `protoc-gen-prost` (replicating our existing configs with the options it provides)
  - `protoc-gen-prost-serde` (replicating our invocation of `pbjson-build`)
  - `protoc-gen-tonic` (replicating our existing configs, in particular replicating the cfg-module options we use to feature-gate the tonic server code)
- The output location and file structure might be slightly different; I suspect it will be easier to update the `include!` paths where we arrange the generated code into the module tree rather than trying to exactly match what we currently do with `tools/proto-compiler`

conorsch:
Picking this up. Given the recent crate reorg, our plan is currently:

1. Move all `.proto` files to top-level `proto/` directory.
2. Use `buf`-compatible tooling for all stages of the codegen.
3. Continue to store output of the generated rust code inside `crates/proto/src/gen/`, updating include paths as necessary.

First attempt in branch `2184-let-there-be-buf` shows progress. Currently having trouble sorting out how to use `prost-gen-serde` correctly: without it, some of the deserialization impls are missing, which makes sense. But enabling prost-gen-serde causes conflicting duplicate impls of deserialize, which clearly isn't right either. Will keep wrestling with it.

conorsch:
Updated the feature branch to use generated protos for go and rust, entirely via buf/prost plugins. All hardcoded vendored protos have been removed. The major regression right now is that gRPC server reflection is broken. Haven't yet figured out how to cobble together the file descriptor info and expose that to tonic during service construction, but it seems to be possible. Will try on that a bit longer.

hdevalence:
[...]

PR #2300, opened 2023-04-05
proto: enable grpc reflection

### LINKED ISSUE #2299: Add support for GRPC server reflection

Third-party GRPC tooling makes use of server reflection, and doesn't work with `pd` or other Penumbra GRPC servers, which don't support reflection.

Hook in [`tonic-reflection` to support server reflection].

**Additional context**
Possible interaction with #2184, in that both involve adjusting the protobuf build process.

hdevalence:
To do this, we have to save the `proto_descriptor.bin` generated by Prost/Tonic.  It's currently 300KB and changes with every proto edit, so I attempted to configure it to use git lfs.

Then, in the `pclientd` and `pd` binaries, we add a grpc-web enabled version of the reflection service.

Closes #2299.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,2 @@
 *.bin filter=lfs diff=lfs merge=lfs -text
+proto/src/gen/proto_descriptor.bin filter=lfs diff=lfs merge=lfs -text
--- a/.github/workflows/buf-pull-request.yml
+++ b/.github/workflows/buf-pull-request.yml
@@ -69,6 +69,9 @@ jobs:
         run: |
           cd tools/proto-compiler
           cargo run
+          cd ../..
+          # Ignore changes to the unstable proto_descriptor output
+          git checkout -- proto/src/gen/proto_descriptor.bin
           s="$(git status --porcelain)"
           if [[ -n "$s" ]]; then
               echo "ERROR: protobuf files must be regenerated and committed:"
--- a/.github/workflows/smoke.yml
+++ b/.github/workflows/smoke.yml
@@ -14,6 +14,8 @@ jobs:
     environment: smoke-test
     steps:
       - uses: actions/checkout@v2
+        with:
+          lfs: true
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
--- a/proto/src/lib.rs
+++ b/proto/src/lib.rs
@@ -266,3 +266,6 @@ pub mod tendermint {
         include!("gen/tendermint.p2p.rs");
     }
 }
+
+#[cfg(feature = "rpc")]
+pub const FILE_DESCRIPTOR_SET: &[u8] = include_bytes!("gen/proto_descriptor.bin");
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3464,6 +3464,7 @@ dependencies = [
  "tokio-stream",
  "toml 0.5.11",
  "tonic",
+ "tonic-reflection",
  "tonic-web",
  "tower",
  "tracing",
@@ -3536,6 +3537,7 @@ dependencies = [
  "tokio-util 0.7.7",
  "toml 0.5.11",
  "tonic",
+ "tonic-reflection",
  "tonic-web",
  "tower",
  "tower-abci",
@@ -5956,6 +5958,20 @@ dependencies = [
  "tower-service",
  "tracing",
  "tracing-futures",
+]
+
+[[package]]
+name = "tonic-reflection"
+version = "0.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "67494bad4dda4c9bffae901dfe14e2b2c0f760adb4706dc10beeb81799f7f7b2"
+dependencies = [
+ "bytes",
+ "prost",
+ "prost-types",
+ "tokio",
+ "tokio-stream",
+ "tonic",
 ]
 
 [[package]]
--- a/pclientd/Cargo.toml
+++ b/pclientd/Cargo.toml
@@ -40,6 +40,7 @@ http-body = "0.4.5"
 tower = "0.4.0"
 tonic = "0.8.1"
 tonic-web = "0.4.0"
+tonic-reflection = "0.6"
 bytes = { version = "1", features = ["serde"] }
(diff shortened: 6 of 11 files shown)


hdevalence:
The smoke test is failing with
```
 2023-04-05T00:29:09.788739Z  INFO starting pd abci_bind=127.0.0.1:26658 grpc_bind=127.0.0.1:8080 metrics_bind=127.0.0.1:9000
2023-04-05T00:29:09.788850Z  INFO opening rocksdb path="/home/ubuntu/.penumbra/testnet_data/node0/pd/rocksdb"
Error: could not configure grpc reflection service

Caused by:
    error decoding FileDescriptorSet from buffer
```
Is it possible that the smoke test isn't using git-lfs correctly, or that I misconfigured the git-lfs in the PR?

hdevalence:
There are also problems with the protobuf generation CI check:
```
      Running `target/debug/penumbra-proto-compiler`
/home/ubuntu/actions-runner/_work/penumbra/penumbra/tools/proto-compiler
/home/ubuntu/actions-runner/_work/penumbra/penumbra/tools/proto-compiler/../../proto/src/gen
cargo:rerun-if-changed=../../proto/proto/penumbra/client/v1alpha1/client.proto
cargo:rerun-if-changed=../../proto/proto/penumbra/view/v1alpha1/view.proto
cargo:rerun-if-changed=../../proto/proto/penumbra/custody/v1alpha1/custody.proto
cargo:rerun-if-changed=../../proto/rust-vendored/cosmos/base/tendermint/v1beta1/query.proto
cargo:rerun-if-changed=../../proto/rust-vendored/tendermint/types/validator.proto
cargo:rerun-if-changed=../../proto/rust-vendored/tendermint/p2p/types.proto
cargo:rerun-if-changed=../../proto/proto/
cargo:rerun-if-changed=../../proto/rust-vendored/
ERROR: protobuf files must be regenerated and committed:
 M proto/src/gen/proto_descriptor.bin
Error: Process completed with exit code 1.
```
I don't see changes when re-running the proto compiler locally. Perhaps the output is not stable? If so, should we explicitly exclude it from the workflow?

hdevalence:
I added a printout of the fd set size and it's small, so it looks like there's an `lfs` problem:
```
 Error: could not configure grpc reflection service, fd set size 131

Caused by:
    error decoding FileDescriptorSet from buffer
```

conorsch:
> Is it possible that the smoke test isn't using git-lfs correctly, or that I misconfigured the git-lfs in the PR?

We intentionally disabled lfs support for the smoke test in 492458973ac7d053f31eb3bdcbc252a64051b509, to ensure that our auto-download logic for params was working as expected. The auto-download logic is feature-gated behind `penumbra-proof-params/download-proving-keys`, and currently specific to the proof-params crate, in its `build.rs` logic. So our options are:

1. re-enable lfs in the smoke test ci
2. port the auto-download-lfs logic to the proto crate

1 is the least effort; 2 is more complex, but presumably more durable long-term. However, given the special case of the `proto` crate, and the fact that the build logic is likely to change soon (#2184), maybe now isn't the best time for 2. 

Thoughts?

hdevalence:
Got it. I think (1) is a good option; I suspect it's unlikely that we'll break the auto-downloading on purpose, and we have other software, like `galileo`, that uses the auto-download logic and that will detect an accidental breakage.

.github/workflows/buf-pull-request.yml
```diff
@@ -69,6 +69,9 @@ jobs:
         run: |
           cd tools/proto-compiler
           cargo run
+          cd ../..
+          # Ignore changes to the unstable proto_descriptor output
+          git checkout -- proto/src/gen/proto_descriptor.bin
```
hdevalence, line 74: @conorsch does this look OK to you? I asked in the Tokio discord and was told that the descriptor output depends on the specific protoc and so might not be stable / deterministic.  My thinking is that since it's generated by the tool, in practice there's no situation where we'd need to catch a change that wouldn't be caught by changes to the Rust codegen.

conorsch: Yes, this makes sense! There's actually a seldom-encountered race where the buf lock file can change due to an out of band (i.e. not a dep we control) version bump. I'll file an issue to track that, because I suspect the upcoming rewrite of the proto-compiler tool might fix it, too.

.github/workflows/smoke.yml
```diff
@@ -14,6 +14,8 @@ jobs:
     environment: smoke-test
     steps:
       - uses: actions/checkout@v2
+        with:
+          lfs: true
```
conorsch, line 18: :+1:

.gitattributes
```diff
@@ -1 +1,2 @@
 *.bin filter=lfs diff=lfs merge=lfs -text
+proto/src/gen/proto_descriptor.bin filter=lfs diff=lfs merge=lfs -text
```
conorsch, line 2: Nit: no changes required to `.gitattributes`; the preexisting `*.bin` rule is sufficient for our needs:

```
❯ fd -e bin
proof-params/src/gen/output_pk.bin
proof-params/src/gen/spend_pk.bin
proof-params/src/gen/swap_pk.bin
proto/src/gen/proto_descriptor.bin
```

proto/src/lib.rs
```diff
@@ -266,3 +266,6 @@ pub mod tendermint {
         include!("gen/tendermint.p2p.rs");
     }
 }
+
+#[cfg(feature = "rpc")]
+pub const FILE_DESCRIPTOR_SET: &[u8] = include_bytes!("gen/proto_descriptor.bin");
```
conorsch, line 271: FYI, I ran `cargo check-all-features --release` (#2210) locally, and it failed on this branch. It's also failing on main, but it was passing earlier this week, so will dig around a bit more and see if we can unbreak it. Doesn't need to block here.

.github/workflows/buf-pull-request.yml
conorsch: > There's actually a seldom-encountered race where the buf lock file can change due to an out of band (i.e. not a dep we control) version bump. I'll file an issue to track that

Ahoy: #2321

PR #2301, opened 2023-04-05
add support for ICS20 withdrawals to pcli

avahowell:


app/src/action_handler/actions/proposal/submit.rs
```diff
@@ -78,7 +78,8 @@ impl ActionHandler for ProposalSubmit {
                 use penumbra_transaction::plan::ActionPlan::*;
                 for action in &transaction_plan.actions {
                     match action {
-                        Spend(_) | Output(_) | Swap(_) | SwapClaim(_) | DelegatorVote(_) => {
+                        Spend(_) | Output(_) | Swap(_) | SwapClaim(_) | DelegatorVote(_)
+                        | WithdrawalPlan(_) => {
```
hdevalence, line 82: wait, why do we want to ban the DAO from making withdrawal plans? it seems potentially useful to be able to send funds directly from a proposal to, e.g., a multisig contract on another ibc chain.

i don't think the withdrawal plan requires any proving, so it seems like it shouldn't be a dos risk

pcli/src/command/tx.rs
```diff
@@ -162,6 +165,24 @@ pub enum TxCmd {
     /// Currently, only zero-fee sweep transactions are implemented.
     #[clap(display_order = 990)]
     Sweep,
+
+    /// Perform an ICS-20 withdrawal.
+    #[clap(display_order = 1000)]
+    Ics20Withdrawal {
```
hdevalence, line 171: The command name will get inferred from the enum variant, so maybe we should call it just `Withdraw`, and set the display order to 350, so it's slotted in after swaps but before voting?

pcli/src/command/tx.rs
```diff
@@ -162,6 +165,24 @@ pub enum TxCmd {
...
+        destination_chain_id: String,
+        destination_chain_address: String,
+        denom: String, //TODO: should we pull this out of amount
+        amount: String,
```
hdevalence, line 175: Yeah, I think combining these into a single `value: String` and parsing them as in `tx send` would be a better UI

pcli/src/command/tx.rs
```diff
@@ -162,6 +165,24 @@ pub enum TxCmd {
...
+        source_channel: String,
+
+        #[clap(long, default_value = "0", display_order = 100)]
+        timeout_height: u64,
+        #[clap(long, default_value = "0", display_order = 150)]
+        timeout_timestamp: u64,
+
+        #[clap(long, default_value = "0", display_order = 200)]
+        return_address_index: u32,
```
hdevalence, line 184: This seems like it should be a `source` flag, like the other commands (e.g., `tx send`), and to match those, we should have it (1) always select funds only from the given source account and (2) always choose a return address that's a randomization of the given source account.

This way, we uphold the design that the pcli commands respect accounts as an (internal) accounting boundary.

pcli/src/command/tx.rs
```diff
@@ -162,6 +165,24 @@ pub enum TxCmd {
     /// Currently, only zero-fee sweep transactions are implemented.
     #[clap(display_order = 990)]
     Sweep,
+
+    /// Perform an ICS-20 withdrawal.
+    #[clap(display_order = 1000)]
+    Ics20Withdrawal {
+        destination_chain_id: String,
+        destination_chain_address: String,
```
hdevalence, line 173: Can we make this a `--to` field, like in `tx send`?  I'm also wondering if it would be better to lump the destination chain address together with the destination chain ID, like
```
--to cosmos1grgelyng2v6v3t8z87wu3sxgt9m5s03xvslewd@cosmoshub-4
```
That way, it's harder to accidentally send funds to the wrong chain, because you have to specify a fully-qualified address in one piece

pcli/src/command/tx.rs
```diff
@@ -162,6 +165,24 @@ pub enum TxCmd {
...
+        denom: String, //TODO: should we pull this out of amount
+        amount: String,
+        source_channel: String,
```
hdevalence, line 176: Why do we need to specify the source channel explicitly? Could we infer it from the command?

pcli/src/command/tx.rs
```diff
@@ -808,6 +830,63 @@ impl TxCmd {
...
+                destination_chain_id,
+                destination_chain_address,
+                amount,
+                denom,
+                timeout_height,
+                timeout_timestamp,
+                source_channel,
+                return_address_index,
+            } => {
+                let fee = Fee::from_staking_token_amount(1_000_000u64.into());
```
hdevalence, line 843: This should probably default to `0` rather than `1` -- since we're going to do a fee overhaul as part of distributions anyways, I think we don't need to worry about setting it for now.

pcli/src/command/tx.rs
```diff
@@ -808,6 +830,63 @@ impl TxCmd {
...
+                        timeout_height,
+                        timeout_timestamp,
+                        ephemeral_return_address,
+                        ChannelId::from_str(source_channel)?,
+                    )
+                    .fee(fee)
+                    .plan(
+                        app.view.as_mut().unwrap(),
+                        app.fvk.account_group_id(),
+                        AddressIndex::new(0),
```
hdevalence, line 885: I think this should match the `source` (current `return_address_index`), so that we're always withdrawing from the specified account and returning funds to it.

proto/proto/penumbra/core/ibc/v1alpha1/ibc.proto
```diff
@@ -78,6 +78,17 @@ message Ics20Withdrawal {
...
+message Ics20WithdrawalPlan {
+  string destination_chain_id = 1;
+  string destination_chain_address = 2;
+  crypto.v1alpha1.Denom denom = 3;
+  crypto.v1alpha1.Amount amount = 4;
+  string source_channel = 5;
+  uint64 timeout_height = 6;
+  uint64 timeout_timestamp = 7;
+  crypto.v1alpha1.Address return_address = 8;
+}
```
hdevalence, line 90: Does this differ from the `Ics20Withdrawal` action? If there's no blinding factors or other encrypted data etc., we can just use the action as a plan of itself, similar to the other transparent actions.

transaction/src/plan/action/ics20_withdrawal.rs
```diff
@@ -0,0 +1,98 @@
...
+    pub amount: Amount,
+    pub source_channel: ChannelId,
+    pub timeout_height: u64,
+    pub timeout_timestamp: u64,
+
+    pub return_address: Address,
+}
+
+impl Ics20WithdrawalPlan {
+    // NOTE: these are duplicated from action::Ics20Withdrawal. should they be deduplicated?
```
hdevalence, line 25: Commented above, but AFAICT the best deduplication would be to just use the action itself as the plan.

view/src/planner.rs
```diff
@@ -326,6 +327,32 @@ impl<R: RngCore + CryptoRng> Planner<R> {
...
+            destination_chain_address,
+            denom,
+            amount,
+            timeout_height,
+            timeout_timestamp,
+            source_channel,
+            return_address,
+        }));
+        self
+    }
```
hdevalence, line 354: Rather than duplicating all of the fields, I think this method should just take the ICS20 withdrawal action itself, letting the caller construct it themselves with named fields if they want, or pass an object if they already had one.

hdevalence:
Rebased; it'd be good if we could get the review comments addressed so we don't miss shipping this.

avahowell:
review comments addressed, with the exception of the unified `value` field, since it seems like the `Value` type can't parse unknown denoms into a full denom? the withdrawal type needs a full denom, not just an asset id

hdevalence:
Oh, hmm, good catch. Maybe in the future we can make a FromStr for ValueView or something. Anyways, not super important now.

PR #2306, opened 2023-04-05
narsil: rename package without penumbra prefix

conorsch:
Follow up to #2297. The docs deploy started to fail after merge of 2297, because the narsil crate was sometimes referred to as `narsil` and other times as `penumbra-narsil`. I've updated it to use the former throughout, similar to how we name the `pcli` package.


conorsch:
I confirmed this unbreaks the docs setup by running locally:

```
REGENERATE_INDEX=1 ./deployments/scripts/rust-docs
```

On merge, the docs deploy will be unbroken. Submitting as a PR for visibility, in case folks feel strongly about the naming convention.

hdevalence (approved):
LGTM

conorsch:
Docs build working again: https://github.com/penumbra-zone/penumbra/actions/runs/4623286774 which means we have narsil included in rustdoc now: https://rustdoc.penumbra.zone/main/narsil/index.html

PR #2327, opened 2023-04-08
pd,pclientd,narsild: auto-color log output

### LINKED ISSUE #2314: pd logs are always colored, even with no tty
**Describe the bug**
The logs emitted by `pd` are _always_ colored, which makes them hard to read in plaintext contexts, such as log aggregation, or grepping/searching through logs.

**To Reproduce**
Here's what the logs look like in a terminal—colored and pretty:

pd-logs-colored

Pipe that same command to `less`, however, and it's difficult to make sense of:

pd-logs-unescaped

**Expected behavior**
Logs should be colored if-and-only-if pd is run inside a tty, such as an interactive terminal window. In other contexts, when there's no tty, such as run under systemd or as an orchestrated container, the logs should be plaintext. 

**Additional context**
We'll need to tweak the settings on the tracing config to alter the behavior. Ideally we can do that by detecting whether there's a tty, but failing that, let's prefer plaintext logs for readability.

hdevalence:
Closes #2314

(I haven't tested that this does the desired behavior)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3505,6 +3505,7 @@ version = "0.1.0"
 dependencies = [
  "anyhow",
  "async-stream 0.2.1",
+ "atty",
  "clap 3.2.23",
  "console-subscriber",
  "futures",
@@ -4014,6 +4015,7 @@ dependencies = [
  "assert_cmd",
  "async-stream 0.2.1",
  "async-trait",
+ "atty",
  "base64 0.20.0",
  "bytes",
  "camino",
@@ -4066,6 +4068,7 @@ dependencies = [
  "ark-ff",
  "async-stream 0.2.1",
  "async-trait",
+ "atty",
  "base64 0.20.0",
  "bincode",
  "blake2b_simd 0.5.11",
--- a/narsil/Cargo.toml
+++ b/narsil/Cargo.toml
@@ -57,6 +57,7 @@ tower-service = "0.3.2"
 tracing = "0.1"
 tracing-subscriber = { version = "0.3", features = ["env-filter", "ansi"] }
 url = "2"
+atty = "0.2"
 
 [build-dependencies]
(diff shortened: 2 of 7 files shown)


conorsch:
pclientd-atty-memleak

The smoke test currently fails due to oomkiller; running locally, the pclientd smoke tests will use all available memory. Will poke at the formatting changes to remove the leak.

conorsch:
The minimal diff for isolating the memleak is:

```
--- a/pclientd/src/main.rs
+++ b/pclientd/src/main.rs
@@ -5,7 +5,8 @@ use pclientd::Opt;
 
 #[tokio::main]
 async fn main() -> Result<()> {
-    tracing_subscriber::fmt::init();
+    tracing_subscriber::fmt()
+        .init();
     let opt = Opt::parse();
 
     opt.exec().await
```

Resolved by porting over the envfilter builder construction from the other bins.

hdevalence:
> The minimal diff for isolating the memleak is:
> 
> ```
> --- a/pclientd/src/main.rs
> +++ b/pclientd/src/main.rs
> @@ -5,7 +5,8 @@ use pclientd::Opt;
>  
>  #[tokio::main]
>  async fn main() -> Result<()> {
> -    tracing_subscriber::fmt::init();
> +    tracing_subscriber::fmt()
> +        .init();
>      let opt = Opt::parse();
>  
>      opt.exec().await
> ```
> 
> Resolved by porting over the envfilter builder construction from the other bins.

Wait, what?

Ah, looking at the docs: https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/fn.init.html  maybe the problem is that this wasn't configuring any filtering?

conorsch:
That's right, tacking on the filters works just fine. No more leaks, and piping to `tee` shuts off the color, just like we wanted.

Issue #2330, opened 2023-04-10
relayer deployment failing due to unsupported value type


conorsch:
**Describe the bug**
The CI step to deploy the relayer has [started failing](https://github.com/penumbra-zone/penumbra/actions/runs/4653657699/jobs/8235041831), post merge of https://github.com/penumbra-zone/penumbra/pull/2315.

**To Reproduce**
Easily reproducible locally:

```
cd deployments/relayer
just
```

**Expected behavior**
Relayer config scripts can build a connection between the Penumbra chains.

**Additional context**
Log messages contain failure messages such as:

```
ts=2023-04-10T16:51:52.724670Z lvl=debug msg="Assembled connection handshake message" path_name=penumbra_path chain_id=penumbra-testnet-theb-ff3b9445 client_id=07-tendermint-143 msgError="unsupported value type"
ts=2023-04-10T16:51:52.724735Z lvl=debug msg="Will broadcast connection handshake message" path_name=penumbra_path chain_id=penumbra-testnet-theb-ff3b9445 client_id=07-tendermint-143 msgError="unsupported value type"
ts=2023-04-10T16:51:52.724735Z lvl=debug msg="Will broadcast connection handshake message" path_name=penumbra_path chain_id=penumbra-testnet-theb-ff3b9445 client_id=07-tendermint-143 msgError="unsupported value type"
ts=2023-04-10T16:51:52.743759Z lvl=debug msg="Broadcasting penumbra tx" provider_type=penumbra
ts=2023-04-10T16:51:52.761052Z lvl=debug msg="Received response from sending messages" provider_type=penumbra response= responseError="unsupported value type"
ts=2023-04-10T16:51:52.761090Z lvl=debug msg="Successfully broadcasted connection handshake message" path_name=penumbra_path chain_id=penumbra-testnet-theb-ff3b9445 client_id=07-tendermint-143 msgError="unsupported value type"
```

Note the ubiquitous "unsupported value type". Let's figure out what broke and see if we can get it working again.

conorsch:
Progress with the updated protobufs from PR into upstream. With the latest code, we still encounter failure, but the config building step shows a different error:

```
rly --debug transact channel penumbra_path
Error: identifier cannot be blank: invalid identifier
```
[...]

PR #2338, opened 2023-04-11
ci: pin tag for upstream relayer image

conorsch:
Trying out a new version of the relayer, including updated protobufs, from PR [0]. Built that container locally, based on the checked out PR, and pushed to ghcr within our GH org.

Refs #2330 

[0] https://github.com/cosmos/relayer/pull/1168


conorsch:
No longer necessary, since upstream PR is merged. Will continue to debug relayer in #2330.

Issue #2310, opened 2023-04-06
Represent denom metadata compatibly with Cosmos SDK


plaidfinch:
Right now we use a custom `REGISTRY` to keep track of denomination metadata. This only works in the closed-world assumption where we know about all the families of assets that could possibly exist, and does not work in the open-world context of IBC. Instead, we should transition to a uniform representation of denominations, ideally one compatible with the Cosmos SDK so that we can have easy interoperability with external asset registries.

The Cosmos SDK [represents denomination metadata like this](https://buf.build/cosmos/cosmos-sdk/docs/main:cosmos.bank.v1beta1#cosmos.bank.v1beta1.Metadata):

```proto
// Metadata represents a struct that describes a basic token.
message Metadata {
    string description = 1;
    // denom_units represents the list of DenomUnit's for a given coin
    repeated DenomUnit denom_units = 2;
    // base represents the base denom (should be the DenomUnit with exponent = 0).
    string base = 3;
    // display indicates the suggested denom that should be
    // displayed in clients.
    string display = 4;
    // name defines the name of the token (eg: Cosmos Atom)
    //
    // Since: cosmos-sdk 0.43
    string name = 5;
    // symbol is the token symbol usually shown on exchanges (eg: ATOM). This can
    // be the same as the display.
    //
    // Since: cosmos-sdk 0.43
    string symbol = 6;
    // URI to a document (on or off-chain) that contains additional information. Optional.
    //
    // Since: cosmos-sdk 0.46
    string uri = 7;
    // URIHash is a sha256 hash of a document pointed by URI. It's used to verify that
    // the document didn't change. Optional.
    //
    // Since: cosmos-sdk 0.46
    string uri_hash = 8;
}

// DenomUnit represents a struct that describes a given denomination unit of the basic token.
message DenomUnit {
    // denom represents the string name of the given denom unit (e.g uatom).
    string denom = 1;
    // exponent represents power of 10 exponent that one must
    // raise the base_denom to in order to equal the given DenomUnit's denom
    // 1 denom = 10^exponent base_denom
    // (e.g. with a base_denom of uatom, one can create a DenomUnit of 'atom' with
    // exponent = 6, thus: 1 atom = 10^6 uatom).
    uint32 exponent = 2;
    // aliases is a list of string aliases for the given denom
    repeated string aliases = 3;
}
```

This is entirely compatible with our representation of denoms, except that it does not represent the Penumbra asset ID, which is computed like:

```rust
[...]

PR #2340, opened 2023-04-11
Vendor Cosmos SDK `DenomMetadata` and serve for `AssetInfoRequest`

plaidfinch:
We don't currently consume `AssetInfoResponse` in the client (we fetch all the assets in one go), so this doesn't need to alter client implementation.

Currently this does not report any special information for penumbra native denominations; that is, it only reports the base denom and no information about suggested name or symbol, etc. Should that be in a follow-up PR or in this one?

### DIFF
--- a/pd/src/info/specific.rs
+++ b/pd/src/info/specific.rs
@@ -22,6 +22,7 @@ use penumbra_proto::{
         ProposalInfoResponse, ProposalRateDataRequest, ProposalRateDataResponse,
         StubCpmmReservesRequest, ValidatorStatusRequest,
     },
+    core::crypto::v1alpha1::{DenomMetadata, DenomUnit},
     StateReadProto as _,
 };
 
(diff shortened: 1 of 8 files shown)


pd/src/info/specific.rs
```diff
@@ -317,8 +318,26 @@ impl SpecificQueryService for Info {
...
+                        display: base_unit.denom.clone(),
+                        denom_units: vec![base_unit],
+                        penumbra_asset_id: Some(id.into()),
+                        // TODO: Keep track of this metadata and report it for Penumbra's own assets
+                        name: Default::default(),
+                        description: Default::default(),
+                        symbol: Default::default(),
+                        uri: Default::default(),
+                        uri_hash: Default::default(),
+                    }),
```
hdevalence, line 340: Why are we constructing the `DenomMetadata` manually, rather than just pulling it out of the state?

hdevalence:
We need to change the on-chain lookup to store a `DenomMetadata` rather than just a denomination: https://github.com/penumbra-zone/penumbra/issues/2310#issuecomment-1502053339

pd/src/info/specific.rs
```diff
@@ -317,8 +318,26 @@ impl SpecificQueryService for Info {
...
+                        base: base_unit.denom.clone(),
+                        display: base_unit.denom.clone(),
+                        denom_units: vec![base_unit],
+                        penumbra_asset_id: Some(id.into()),
+                        // TODO: Keep track of this metadata and report it for Penumbra's own assets
+                        name: Default::default(),
+                        description: Default::default(),
+                        symbol: Default::default(),
+                        uri: Default::default(),
+                        uri_hash: Default::default(),
```
hdevalence, line 339: I don't think we need to handle these fields at all at the moment, we could leave them out of our domain types for now (and in some cases, like `uri`, possibly forever).

hdevalence:
Yes, I think we should just do the minimal change in #2310 in one go.

hdevalence:
Yes, the minimal change is to make the shielded pool's asset lookup return `DenomMetadata`: https://github.com/penumbra-zone/penumbra/issues/2310#issuecomment-1499420018

To do that, the `DenomMetadata` has to be populated when the asset is registered. There's no way for the `REGISTRY` and the chain state to disagree, because they're both hardcoded, so we can just not hardcode a disagreement.

erwanor:
This is stalled, pending a technical solution.

PR #2342, opened 2023-04-11
Retroactively remove tombstoned validators & their delegators from tally

### LINKED ISSUE #2111: Prevent tombstoned validators or their delegators from voting on governance proposals
Tombstoned validators no longer have a stake in the protocol. I posit that their delegators, who are forcibly unbonded, also should be barred from voting, though this seems less important.

plaidfinch:
Resolves #2111.

This also refactors the emergency proposal pass logic into a common function on the state which can be called from the action handler and the governance component. We call it in both places, because now a validator tombstoning can result in a change to the proposal passing state for an emergency proposal, and we should check immediately when the epoch rolls over (which is now triggered by the slashing event).


hdevalence:
Per https://github.com/penumbra-zone/penumbra/issues/2111#issuecomment-1504006308, I don't think we should do this.

PR #2345, opened 2023-04-11
Resolves #2085, preventing unintentional chain restarts

### LINKED ISSUE #2085: Prevent un-intentional external restarts when chain tries to halt
Software like systemd and kubernetes restarts processes when they stop. This will likely mean that if the majority use this to manage their deployments, an emergency chain halt... won't halt the chain. We need to prevent this.

Suggested solution: if `pd` starts up and the last chain halt occurred in the most recent block, check for an environment variable `PD_RESTART_CHAIN_FROM=N` where `N` is the block height of the last halt. This variable can't be set once-and-for-all, and must be intentionally set by the person restarting the software, which means auto-restarts will fail until manual intervention. Thanks @conorsch for bringing up this issue.

plaidfinch:
This PR writes to a key `chain/halted/N` where `N` is the number of times the chain has previously halted, every time the chain halts. Then, upon initialization of the `App`, the key `chain/halted/TOTAL_HALT_COUNT` is checked to determine if the chain has halted before without the software having been updated. The constant `TOTAL_HALT_COUNT` defined separately in `app` and `narsil` (which both reuse the `chain` crate) must be manually incremented in the code every time the chain halts, in order for the new version of the software to pass over the block height of the chain halt (and all previous halts); otherwise, the chain will not start up again. If `TOTAL_HALT_COUNT` is incremented, synchronizing from genesis will not cause the chain to halt on the previous halt block height, because the check in the `App` for the halt key `chain/halted/TOTAL_HALT_COUNT` will fail, and the chain will continue to sync.

### DIFF
--- a/narsil/src/ledger/app.rs
+++ b/narsil/src/ledger/app.rs
@@ -77,20 +77,11 @@ impl App {
         let state = Arc::try_unwrap(std::mem::replace(&mut self.state, Arc::new(dummy_state)))
             .expect("we have exclusive ownership of the State at commit()");
 
-        // Check if someone has signaled that we should halt.
-        let should_halt = state.should_halt();
-
         // Commit the pending writes, clearing the state.
         let jmt_root = storage
             .commit(state)
             .await
             .expect("must be able to successfully commit to storage");
-
-        // If we should halt, we should end the process here.
-        if should_halt {
-            tracing::info!("committed block when a chain halt was signaled; exiting now");
-            std::process::exit(0);
-        }
 
         let app_hash: AppHash = jmt_root.into();
 


hdevalence:
This shouldn't touch the `narsil` code, which shouldn't be dependent on the `penumbra-chain` crate at all.

hdevalence:
Yes, the narsil crate should be stubbed out with its own state machine, not reusing the Penumbra one -- and we shouldn't make it aware of Penumbra-specific state machine changes.

app/src/app/mod.rs
```diff
@@ -349,7 +357,10 @@ impl App {
             .expect("we have exclusive ownership of the State at commit()");
 
         // Check if someone has signaled that we should halt.
-        let should_halt = state.should_halt();
+        let should_halt = state
+            .is_chain_halted(TOTAL_HALT_COUNT)
+            .await
+            .expect("must be able to read halt flag");
```
hdevalence, line 363: How can we have halted prior to genesis?

pd/src/main.rs
```diff
@@ -214,14 +214,21 @@ async fn main() -> anyhow::Result<()> {
...
+                    }
                 }));
             let mempool = tower::ServiceBuilder::new()
                 .layer(request_span::layer(|req: &MempoolRequest| {
                     req.create_span()
                 }))
                 .service(tower_actor::Actor::new(10, |queue: _| {
-                    pd::Mempool::new(storage.clone(), queue).run()
+                    let storage = storage.clone();
+                    async move { pd::Mempool::new(storage.clone(), queue).await?.run().await }
```
hdevalence, line 231: Can we avoid propagating this noisy sequence onto API users?

app/src/app/mod.rs
hdevalence: Got it, sorry about that!

app/src/app/mod.rs
```diff
@@ -32,13 +32,21 @@ pub struct App {
 }
 
 impl App {
-    pub fn new(snapshot: Snapshot) -> Self {
+    pub async fn new(snapshot: Snapshot) -> Result<Self> {
```
hdevalence, line 35: It feels wrong for this method to be fallible.  Every `Snapshot` should represent a valid chain state, so I think creating an `App`, which is a view of that valid chain state, should also be infallible.

Making it fallible and asynchronous taints all of its callers, who now have to deal with errors, regardless of whether or not that makes sense (e.g., we will have to `await` and `unwrap` every time we reset the mempool fork).

Why don't we just do the check in the one place it matters -- in `commit`?

pd/src/main.rs
hdevalence: https://github.com/penumbra-zone/penumbra/pull/2345/files/c06530db6d049b770a7d70104667e308e509e4b6#r1163369559

hdevalence:
> Should this PR do that stubbing-out?

We could just not include the `narsil` changes, right?

hdevalence:
OK, let's just delete the `let should_halt = state.should_halt();` in the Narsil crate, which is already slated for removal?

PR #2346, opened 2023-04-11
Signal the end of an epoch for `active => disabled|jailed|tombstoned` validator transitions

plaidfinch:


hdevalence (approved):

hdevalence:
Nice, this is really clean

Issue #1825, opened 2023-01-10
Epic: Cryptography tasks prior to mainnet


redshiftzero:
This is a tracking ticket for the core protocol cryptography tasks that need to be completed prior to mainnet. 

# Finalize core cryptography

The current status as of filing this issue is:
* We have finalized the crypto primitives (decaf377, poseidon377)
* We have finalized the key/address derivation
* We have integrated the pieces we need to support FMD in the future (FMD clues are in transactions, etc.)
* We lifted the memo up to the tx level and added another layer of key wrapping to enable each output in the tx to decrypt the memo

Remaining tasks prior to mainnet:
- [x] https://github.com/penumbra-zone/penumbra/issues/1688
- [x] https://github.com/penumbra-zone/penumbra/issues/1716
- [x] https://github.com/penumbra-zone/penumbra/issues/1747
- [x] https://github.com/penumbra-zone/penumbra/issues/1746
- [x] https://github.com/penumbra-zone/penumbra/issues/1962

# ZK Proofs

The current status of this work is that we've implemented:
* low level r1cs gadgets for decaf377 and poseidon377
* high level r1cs gadgets for all the functionality we need for the spend and output proofs: value commitment integrity, merkle path integrity, nullifier integrity, ephemeral public key integrity, checking generator points are not identity, randomized verification key integrity.
* ZK spend and output proofs are implemented (but not in use in testnets yet)

Tasks prior to mainnet that can be done immediately:
- [x] https://github.com/penumbra-zone/penumbra/issues/711
- [x] https://github.com/penumbra-zone/penumbra/issues/717 (will be closed by https://github.com/penumbra-zone/penumbra/pull/1970)
- [x] Replace transparent spend and output proofs with their ZK versions (M) (will be closed by #1923)
- [x] Benchmarking: How many constraints per high-level gadget? (results [here](https://github.com/penumbra-zone/penumbra/issues/1825#issuecomment-1377477140))
- [x] [Fix intermittent verification issues uncovered during #1923](https://github.com/penumbra-zone/penumbra/issues/2041)
- [x] https://github.com/penumbra-zone/penumbra/issues/2040
- [x] https://github.com/penumbra-zone/penumbra/issues/2056
- [x] https://github.com/penumbra-zone/penumbra/issues/1824
- [x] #2152 (closed by #2272)
- [x] Fixed point arithmetic: will be closed by #2389 
[...]

PR #2355, opened 2023-04-13
spec: add description of delegator vote zk-SNARK statements

redshiftzero:
Adding more docs towards #1825

### DIFF
--- a/docs/protocol/src/protocol/action_descriptions/delegator_vote.md
+++ b/docs/protocol/src/protocol/action_descriptions/delegator_vote.md
@@ -1,0 +1,98 @@
+# Delegator Vote Descriptions
+
+Each delegator vote contains an DelegatorVoteBody and a zk-SNARK delegator vote proof.
+
+## Delegator Vote zk-SNARK Statements
+
+The delegator vote proof demonstrates the properties enumerated below for the following private witnesses known by the prover:
+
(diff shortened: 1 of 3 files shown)


docs/protocol/src/protocol/action_descriptions/delegator_vote.md
```diff
@@ -0,0 +1,98 @@
...
+
+The zk-SNARK certifies that the witnessed Merkle authentication path is a valid Merkle path to the provided public anchor.
+
+### Diversified Base is not Identity
+
+The zk-SNARK certifies that the diversified basepoint $B_d$ associated with the address on the note is not identity.
+
+### Spend Authorization Key is not Identity
+
+The zk-SNARK certifies that the spend authorization key $ak$ is not identity.
```
hdevalence, line 97: why are these checks important? we don't really care that the diversified basepoint is not the identity, in itself, right? that's done in service of some other goal?

redshiftzero: These two checks are the decaf equivalent of ZCash's small-order checks, here from the spend statements:

Screenshot 2023-04-18 at 6 44 02 PM

For us I'm not aware of any specific attack that this mitigates, though the cost of these checks is small since we already have witnessed $B_d$ and $ak$ for checking the transmission key on the note is properly derived and that the spend authorization key has been correctly randomized.

PR #2361, opened 2023-04-13
Attempt to integrate `arti-hyper` into `pcli`

### LINKED ISSUE #2358: Experiment with Tor support in `pcli`

Now that we have support for HTTPS in `pcli`, as well as automatic HTTPS in `pd`, we can make requests from `pcli` over untrusted networks.

In a time-boxed way (say, not more than ~90mins for a first attempt), let's try adding Tor support to `pcli`, using [`arti-hyper`](https://docs.rs/arti-hyper/latest/arti_hyper/) and [`arti-client`](https://docs.rs/arti-client/latest/arti_client/).

To do this, we'll probably want to:

- [x] Add a top-level `--use-tor` option to `pcli`, alongside the RPC endpoint.
- [ ] Require that if `--use-tor` is present, the RPC endpoint is an HTTPS URL.
- [ ] Create a new `PCLI_HOME/arti-state` subdirectory for Arti data, and [configure Arti to only store data there](https://docs.rs/arti-client/latest/arti_client/config/struct.TorClientConfigBuilder.html#method.from_directories), ensuring that `pcli` doesn't read or write state outside of its home directory.
  - Arti, due to its Tor heritage, is apparently [fussy](https://docs.rs/fs-mistrust/0.6.3/fs_mistrust/index.html) about filesystem permissions. I personally don't believe this is a particularly useful threat model, since most systems are effectively single-user anyways. If we run into problems, let's take the path of least resistance for now (which might be [`dangerously_trust_everyone`](https://docs.rs/fs-mistrust/0.6.3/fs_mistrust/struct.MistrustBuilder.html#method.dangerously_trust_everyone)).
- [ ] Hook the `arti-hyper` connector up to the places in `pcli` where we connect to `pd`: https://github.com/penumbra-zone/penumbra/blob/main/pcli/src/network.rs#L82-L94
- [ ] Connect to a `pd` instance over Tor!

Note: depending on timing, it might be helpful to do this work branching directly off of the `050-thebe` tag, so that the client can be aimed at https://pd-tls.penumbra.zone, which is currently serving HTTPS gRPC for Testnet 50 in a way we know to be compatible with `pcli` (thus avoiding a dependency on #2341 if it's not finished yet).

plaidfinch:
Closes #2358

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -422,9 +422,9 @@ checksum = "8da52d66c7071e2e3fa2a1e5c6d088fec47b593032b254f5e980de8ea54454d6"
 
 [[package]]
 name = "arti-client"
-version = "0.8.3"
+version = "0.9.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "03e724eb75d44a875ab5f33fc17dbb06c20fc39e059e431e86dc9bcec5a9174b"
+checksum = "0bc0b55ecee92a250b55535f0957c3ce0953779f9c48ff05e9b5f8a430f1b7aa"
 dependencies = [
  "cfg-if 1.0.0",
  "derive_builder_fork_arti",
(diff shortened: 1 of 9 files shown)


hdevalence:
Closing as per https://github.com/penumbra-zone/penumbra/issues/2358#issuecomment-1554987758 (but preserving this branch as a reference)

PR #2362, opened 2023-04-13
Further routing tests

zbuc:
Seeing unexpected behavior with pricing (see cheap_price/expensive_price comparisons)

### DIFF
--- a/app/src/dex/router/path.rs
+++ b/app/src/dex/router/path.rs
@@ -72,12 +72,14 @@ impl<S: StateRead + 'static> Path<S> {
         self.state.deindex_position(&best_price_position);
 
         // Update and return the path.
-        // TODO: gross
-        let hop_price = if self.end() == &best_price_position.phi.pair.asset_1() {
-            best_price_position.phi.component.ask_price()
-        } else {
-            best_price_position.phi.component.bid_price()
-        };
+        // Orient the best priced position's canonical trading form to a [`DirectionalTradingPair`]
+        // ending at the path's current end and retrieve the bid price.
+        let hop_price = best_price_position
+            .phi
+            .orient_end(*self.end())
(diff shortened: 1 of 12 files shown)


zbuc:
I think the "find best priced position" logic is wrong and what's causing the weirdness in the test. The mispriced position is at 3 GN to 1 PUSD, i.e. it's a _worse_ price going from GN -> PUSD (the direction the test goes), yet that position is taken when it's available.

app/src/dex/router/path.rs
```diff
@@ -72,12 +72,14 @@ impl<S: StateRead + 'static> Path<S> {
...
-            best_price_position.phi.component.ask_price()
-        };
+        // Orient the best priced position's canonical trading form to a [`DirectionalTradingPair`]
+        // ending at the path's current end and retrieve the bid price.
+        let hop_price = best_price_position
+            .phi
+            .orient_end(*self.end())
+            .ok_or_else(|| anyhow::anyhow!("path end not in position"))?
+            .component
+            .bid_price();
```
hdevalence, line 82: This is backwards: we're extending from the current end to the `new_end`, not from the new end to the current `self.end()`.

hdevalence:
The problem was introduced in https://github.com/penumbra-zone/penumbra/commit/517c4dc4833f1247a66349a2edf2deb6c4b2c23f , and it's that we no longer correctly orient the trading function: we always need to be extending _to_ the `new_end`.  I think the "bid/ask" terminology is confusing, because it obscures the fact that the direction is _already chosen_ by the direction of the trading function, and all that's happening there is changing which way the rounding behavior "flows" (are we starting with the output amount and rounding the input amount, or vice versa?).

Like I mentioned in Discord, I think we should add `orient_` methods, but I don't think we should make extensive changes to the rest of the data modeling until we get the routing working.

PR #2366, opened 2023-04-14
Add comment to query in storage::scan_advice

### LINKED ISSUE #1751: Followup: document this SQL query and its rationale.
Followup: document this SQL query and its rationale.

We want to only give advice about notes that are known (hence in `notes`) but not already recorded as spendable, in which case we shouldn't see them again in scanning anyways, and if we do, something bad is probably happening.  I didn't think it through carefully, but I was worried that since the `spendable_notes` table records our own scan results, we might open up the possibility of leaking information about our scan results to an adversary who sends us crafted `CompactBlock`, e.g., sending a compact block with a `StatePayload::RolledUp(target_commitment)` and then somehow observing that we'd detected it using our scan results as advice.

_Originally posted by @hdevalence in https://github.com/penumbra-zone/penumbra/pull/1736#discussion_r1045175372_

aubrika:
Closes #1751

### DIFF
--- a/view/src/storage.rs
+++ b/view/src/storage.rs
@@ -942,6 +942,8 @@ impl Storage {
             return Ok(BTreeMap::new());
         }
 
+        // This query gives advice about notes which are known but which have not already been recorded as spendable,
+        // in part to avoid revealing information about which notes have been spent.
         let rows = sqlx::query(
             format!(
                 "SELECT notes.address,


view/src/storage.rs
```diff
@@ -942,6 +942,8 @@ impl Storage {
             return Ok(BTreeMap::new());
         }
 
+        // This query gives advice about notes which are known but which have not already been recorded as spendable,
+        // in part to avoid revealing information about which notes have been spent.
```
hdevalence, line 946: I don't think that's the reason -- we're already inside the trust boundary, the reason we want to exclude notes that have been marked as spendable is that we only want to give info on notes that have not been scanned yet: 
```
    /// Given a list of note commitments, this method checks whether any of them
    /// correspond to notes that have been recorded in the database but not yet
    /// observed during scanning.
```

Two points:

1. Do we remember why this was important?
2. I think this means the TxP code from earlier this week that uses `scan_advice` is incorrect and won't give the necessary information.

aubrika: 1. Not personally, I was trying to divine the reasoning from the comment in #1736 but didn't remember context
2. Oh hmm, what will we need to do to fix that?

hdevalence: Looks like this was the reason: https://github.com/penumbra-zone/penumbra/pull/1736#discussion_r1045175372

For (2), instead of using scan_advice for TxPs, we should just query notes by commitment.

hdevalence:
Now that https://github.com/penumbra-zone/penumbra/pull/2372 is landed, we should rebuild this. Since the code change is small, I suspect that it will be the least work to just redo it on the new code using this PR as a reference, rather than trying to rebase it.

PR #2370, opened 2023-04-14
Add record_note_inner to storage to commonize recording of notes

### LINKED ISSUE #1750: Followup: commonize give_advice implementation with record_block
Followup: commonize give_advice implementation with record_block

I made this a distinct method because it seemed useful to be able to expose externally in the future for rollup use cases (where there's some system, like Narsil, that's externally feeding advice to the view server).

But since we're duplicating the same insertion query in multiple places (here, and when recording a block, though there it's in the context of some bigger database transaction), it might be better to have a private `async fn record_note_inner(note: Note, dbtx: &mut _)` (not sure what type `dbtx` should have, but it should be the return of `self.pool.begin().await?`) that adds the note recording into a bigger db transaction.

Then `give_advice` can make a standalone dbtx and call `record_note_inner`, and `record_block` can use `record_note_inner` on the bigger transaction it constructs.

_Originally posted by @hdevalence in https://github.com/penumbra-zone/penumbra/pull/1736#discussion_r1045174885_

aubrika:
Closes #1750

### DIFF
--- a/view/sqlx-data.json
+++ b/view/sqlx-data.json
@@ -320,16 +320,6 @@
     },
     "query": "SELECT block_height, tx_hash, tx_bytes\n            FROM tx\n            WHERE tx_hash = ?"
   },
-  "741709fac6ac70295ed4ac0986ea81f230e817a57da49a0c9e2311c91594211f": {
-    "describe": {
-      "columns": [],
-      "nullable": [],
-      "parameters": {
-        "Right": 5
-      }
-    },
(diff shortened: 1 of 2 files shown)


hdevalence:
Let's hold this off until after #2372 lands.

hdevalence:
Now that #2372 is landed, we should rebuild this. Since the code change is small, I suspect that it will be the least work to just redo it on the new code using this PR as a reference, rather than trying to rebase it.

PR #2372, opened 2023-04-14
Attempt to convert `pcli`'s ORM to use `rusqlite` instead of `sqlx`

plaidfinch:
- [x] Rip out sqlx completely by commenting out anything that uses it
- [x] Replace the connection pool with an `Arc<Mutex<rusqlite::Connection>>` (this is temporary)
- [x] Rewrite all the queries using rusqlite
- [x] Test `pcli`, `pclientd`, and `galileo` (things might be really slow because of the global mutex, especially galileo)
- [x] Reintroduce connection pooling by replacing the `Arc<Mutex<...>>` with an `r2d2::Pool<r2d2_sqlite::SqliteConnectionManager>`, so we can (hopefully) read while writing, etc.
- [x] Test everything again
- [x] ~~Might need to re-enable table-level locking using the "shared cache" like we did previously even though rusqlite (and the sqlite docs) warn against it? Consider this carefully...~~

### DIFF
--- a/view/src/swap_record.rs
+++ b/view/src/swap_record.rs
@@ -79,7 +79,7 @@ impl TryFrom<&Row<'_>> for SwapRecord {
             nullifier: row.get::<_, Vec<u8>>("nullifier")?[..].try_into()?,
             output_data: BatchSwapOutputData::decode(&row.get::<_, Vec<u8>>("output_data")?[..])?,
             source: row.get::<_, Vec<u8>>("source")?[..].try_into()?,
-            swap: row.get::<_, Vec<u8>>("swap")?[..].try_into()?,
+            swap: SwapPlaintext::decode(&row.get::<_, Vec<u8>>("swap")?[..])?,
         })
     }
 }


view/src/storage/schema.sql
```diff
@@ -0,0 +1,119 @@
+-- The hash of this schema file
```
hdevalence, line 1: Does this need to live inline with the Rust code? Could we keep it in `view/schema/view.sql` (rather than "migrations")?

view/src/storage.rs
```diff
@@ -343,37 +399,40 @@ impl Storage {
...
-            )
-            .fetch_optional(&pool)
-            .await?
-            {
-                let spent = record.height_spent.is_some();
+        // Check if we already have the nullifier in the set of spent notes
+        if let Some(height_spent) = spawn_blocking(move || {
+            pool.get()?
+                .prepare_cached("SELECT height_spent FROM spendable_notes WHERE nullifier = ?1")?
+                .query_and_then([nullifier_bytes], |row| {
```
hdevalence, line 406: Maybe this would be a good place for some temporary `tracing::warn!` statements to what the query is returning?

view/src/storage.rs
```diff
@@ -343,37 +399,40 @@ impl Storage {
 
         let nullifier_bytes = nullifier.0.to_bytes().to_vec();
```
hdevalence, line 400: Compare this ....

view/src/storage.rs
```diff
@@ -1060,258 +1084,244 @@ impl Storage {
...
-                "UPDATE swaps SET height_claimed = ? WHERE nullifier = ? RETURNING swap_commitment",
-                height_spent,
-                nullifier,
-            )
-            .fetch_optional(&mut dbtx)
-            .await?;
+            // Update any rows of the table with matching nullifiers to have height_spent
+            for nullifier in &filtered_block.spent_nullifiers {
+                let height_spent = filtered_block.height as i64;
+                let nullifier = nullifier.to_bytes().to_vec();
```
hdevalence, line 1181: ... to this ...

view/src/storage.rs
hdevalence: https://github.com/penumbra-zone/penumbra/pull/2372/files#r1170603076

hdevalence:
```
 Apr 19 02:23:05.889  INFO view worker error e=near "RETURNING": syntax error

Caused by:
    Error code 1: SQL error or missing database
```

hdevalence:
> ```
>  Apr 19 02:23:05.889  INFO view worker error e=near "RETURNING": syntax error
> 
> Caused by:
>     Error code 1: SQL error or missing database
> ```

For posterity: this was caused by not using the `bundled` feature for bringing our own sqlite, causing us to use whatever sqlite was available on the host system, and the `RETURNING` feature was added only two years ago, so the build worker had an outdated version.

view/src/storage.rs
hdevalence: This turns out to be fine by accident, since both of these give the same array serialization.

Issue #2373, opened 2023-04-15
Update `ibc` crate version


hdevalence:
**Is your feature request related to a problem? Please describe.**

We need to update the `ibc` crate version to be able to pull in `ics23` `0.10.0`.  There isn't yet a version with `ics23` `0.10.0`, but we're already several versions behind so it'd be good to start upgrading. 

**Describe the solution you'd like**

Upgrade to `ibc` `0.37.0`.

erwanor:
we will also need it to complete the jmt integration, unfortunately. I'm planning to get to it once routing execution is completed, or shortly after anyway.

redshiftzero:
No longer using the ibc crate

PR #2374, opened 2023-04-15
[WIP] Upgrade to `ibc` `0.37`

hdevalence:
Work towards #2373.

This does the easy and mechanical refactoring.  However, there are a number of annoying changes:

- There are now two different proto messages that correspond to the same domain type. This makes it impossible to do `MsgUpdateClient::decode`, because it's now also possible to decode a `MsgSubmitMisbehavior` as a `MsgUpdateClient`. I don't know why it is so hard to just have one Rust domain type for each proto message. Instead we have to write code like
```rust
                let msg = <MsgUpdateClient as ibc_proto::protobuf::Protobuf<
                    ibc_proto::ibc::core::client::v1::MsgUpdateClient,
                >>::decode(raw_action_bytes)?;
```
- The `MsgUpdateClient` no longer has a `header`, because it was merged with `MsgSubmitMisbehavior`. So we have to parse it back out of the "domain type" (or fork the `ibc` crate so it actually has a faithful representation of the IBC messages).
- The methods on the `TendermintClientState` are all gone or renamed.


hdevalence:
The methods on the `TendermintClientState` are all gone because they've been folded into internal details using the `ibc` crate's `ValidationContext` interface -- which we can't use, because it's stateful and incompatible with our async state system. Since the component methods we used to build our implementation have been removed, we'll have to fork and vendor them.  Considering the issues with the `ibc-rs` direction, though, this may be the better option in the long term.

PR #2383, opened 2023-04-18
Update validator transitions in protocol spec to match implementation

### LINKED ISSUE #2363: Update protocol docs for validator state transitions
The [protocol specs on validator transitions](https://protocol.penumbra.zone/main/concepts/validators.html) currently don't represent the "tombstoned" state and also don't correctly describe the mechanics of the "jailed" state. This should be fixed to match the current implementation.

zbuc:
Closes #2363


docs/protocol/src/concepts/validators.md
```diff
@@ -3,44 +3,52 @@
...
 
-* **Inactive**, where the validator is not part of the consensus set, the stake in the validator's delegation pool is not bonded;
+* **Inactive**, where the validator is not part of the consensus set, and the stake in the validator's delegation pool is not bonded;
 * **Active**, where the validator is part of the consensus set, and the stake in the validator's delegation pool is bonded;
-* **Unbonding**, where the validator is not part of the consensus set, but the stake in the validator's delegation pool is still bonded;
-* **Slashed**, where the validator is not part of the consensus set, and the stake in the validator's delegation pool is not bonded.
+* **Jailed**, where the validator is not part of the consensus set, and the stake in the validator's delegation pool is not bonded;
+* **Tombstoned**, where the validator is not part of the consensus set, and the stake in the validator's delegation pool is not bonded;
+* **Disabled**, where the validator is not part of the consensus set, and the stake in the validator's delegation pool is not bonded.
```
hdevalence, line 34: IMO the ", and the stake is/is not bonded" isn't useful here, now that we separate out bondingness and the state. I think this would be clearer if we just had each bullet point have a brief description of the state, like "Inactive: a validator whose delegation pool is too small to participate in consensus", and so on.

docs/protocol/src/concepts/validators.md
```diff
@@ -3,44 +3,52 @@
...
+First, the validator could be jailed and slashed for inactivity.  This can happen in any block, not just at an epoch transition.  Jailed validators are immediately removed from the consensus set. The validator's rates are updated to price in the slashing penalty, and are then held constant. Validators jailed for inactivity are not permanently prohibited from participation in consensus, and their operators can re-activate them by re-uploading the validator definition. Stake cannot be delegated to a slashed validator. Stake already contributed to a slashed validator's delegation pool will enter an unbonding period to hold the validator accountable for any byzantine behavior during the unbonding period. Re-delegations may occur after the validator enters the "Inactive" state.
```
hdevalence, line 40: > This can happen in any block, not just at an epoch transition.

This isn't accurate any more: validator changes always happen at epoch transitions, but those transitions are dynamically triggered:

> This can happen in any block, triggering an unscheduled epoch transition.

hdevalence: and ditto in the tombstoning section

docs/protocol/src/concepts/validators.md
```diff
@@ -3,44 +3,52 @@
...
+
+Third, the validator could be manually disabled by the operator. The validator is then in the disabled state.  It does not participate in consensus, and the stake in its delegation pool does not earn rewards (the validator's rates are held constant).  The stake in its delegation pool will enter an unbonding period at the time the validator becomes disabled. The only valid state a disabled validator may enter into is "inactive", if the operator re-activates it by updating the validator definition.
```
hdevalence, line 44: cc @avahowell @plaidfinch currently, disabling a validator triggers an epoch boundary, correct? is that desirable? should we have disabled validators only exit at the next scheduled epoch boundary?

when we do flow encryption, the first block or two after a new epoch won't be able to do trading, while we wait for the dkg to occur. could a validator disable and re-enable to trigger as many epoch boundaries as they want?

docs/protocol/src/concepts/validators.md
```diff
@@ -3,44 +3,52 @@
...
-3. they can become inactive, if neither (1) nor (2) occurs before the unbonding period passes.
+2. they can be tombstoned, if evidence of misbehavior arises during the unbonding period;
+3. they can be disabled, if the operator chooses.
 
-If (2) occurs, the same state transitions as in regular slashing occur: all pending undelegations are cancelled, etc.
-If (3) occurs, all pending undelegations are immediately removed from quarantine, short-circuiting the unbonding period that began when the undelegation was performed.  If (1) occurs, the validator stops unbonding, but this has no effect on pending undelegations, since they were quarantined with an unbonding period that started when the undelegation was performed (i.e., as if they were undelegations from an active validator).
+If (2) occurs, the same state transitions as in regular tombstoning occur: all pending undelegations are cancelled, etc.
+If (3) occurs, the unbonding period continues and the validator enters the disabled state.
+If (1) occurs, the validator stops unbonding, and all delegations are no longer subject to unbonding restrictions.
```
hdevalence, line 54: should this be ", and all delegations become bonded again"? since stopping unbonding => bonded ?

docs/protocol/src/concepts/validators.md
hdevalence: actually maybe we should update the part of the spect that describes epochs to say that they're time based and dynamic?

zbuc: I don't think the docs ever specify the epoch in terms of block height _or_ time, but maybe it should be more clear? Closest I can find is in the `batching_flows.md`:

```markdown
To allow batching across block boundaries,
Penumbra organizes blocks into epochs, and applies changes to the validator set
only at epoch boundaries.  Decryption keys live for the duration of the epoch,
allowing value flows to be batched over any time interval from 1 block up to the
length of an epoch. We propose epoch boundaries on the order of 1-3 days.
```

PR #2389, opened 2023-04-19
crypto: R1CS U128x128 implementation

redshiftzero:
Needed for #2151 for the pro rata output calculation

### DIFF
--- a/crypto/src/fixpoint.rs
+++ b/crypto/src/fixpoint.rs
@@ -8,6 +8,7 @@ mod ops;
 mod tests;
 
 use ark_ff::{BigInteger, Field, PrimeField};
+use ark_r1cs_std::fields::fp::FpVar;
 use ark_r1cs_std::prelude::*;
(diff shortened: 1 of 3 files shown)


crypto/src/fixpoint.rs
```diff
@@ -181,3 +172,645 @@ impl U128x128 {
...
+    pub fn checked_add(
+        self,
+        rhs: &Self,
+        cs: ConstraintSystemRef<Fq>,
+    ) -> Result<U128x128Var, SynthesisError> {
+        // x = [x0, x1, x2, x3]
+        // x = x0 + x1 * 2^64 + x2 * 2^128 + x3 * 2^192
+        // y = [y0, y1, y2, y3]
+        // y = y0 + y1 * 2^64 + y2 * 2^128 + y3 * 2^192
+        let x0 = convert_uint64_to_fqvar(&self.limbs[0]);
```
redshiftzero, line 241: For later it could make sense to lazily/optionally store the limbs as field elements on the `U128x128Var` struct in addition to `UInt64`s

PR #2397, opened 2023-04-21
ibc: replace ibc-rs with ibc-types

### LINKED ISSUE #2380: Update the IBC component in `penumbra_app` to use `ibc-types` `0.1.0`.
This should be a drop-in change. Let's do it in two steps:

- [ ] Replace the `ibc` dependency with a git dependency on `ibc-types`, until we get green CI (fixing any issues that pop up along the way)
- [ ] Publish `ibc-types` `0.1.0` once we know it works as a drop-in replacement.

avahowell:
closes #2380


avahowell:
We can do a follow-up PR once we publish the first ibc-types version

hdevalence:
Oops, did I not publish the main branch of ibc-types to crates.io?

erwanor:
@hdevalence You did, but I think we need the org to own https://crates.io/crates/ibc-types to publish new versions!

PR #2402, opened 2023-04-24
Implement `SwapExecution` as collection of `TradeTrace`s

zbuc:


proto/proto/penumbra/core/dex/v1alpha1/dex.proto
```diff
@@ -327,19 +327,23 @@ message Path {
...
+  crypto.v1alpha1.Value input = 1;
+  // The output asset and amount.
+  crypto.v1alpha1.Value output = 2;
 }
 
 // Contains the entire execution of a particular swap.
 message SwapExecution {
-  repeated Trade trades = 1;
+  repeated TradeTrace traces = 1;
 }
```
hdevalence, line 347: I don't think we need input and output values, since the output of one step should be the input of the next.  So I think this could just be
```proto
message SwapExecution {
  message Trace {
    repeated crypto.v1alpha1.Value value = 1;
  }
  repeated Trace traces = 1;
}
```
so that we're grouping the `Trace` with the `SwapExecution` in other codegen, and in our domain type we can simplify to
```rust
pub struct SwapExecution {
    pub traces: Vec<Vec<Value>>,
}
```
rather than having to define extra domain types.

Issue #1961, opened 2023-02-07
Build and publish binaries for testnet releases


conorsch:
**Is your feature request related to a problem? Please describe.**
Folks wanting to participate the public testnets must recompile the project from source every week. 

**Describe the solution you'd like**
On a testnet release, we should compile the binaries and post them as artifacts on the release. It'd be nice to create the release object in CI, too, rather than asking a human to do it manually. 

**Describe alternatives you've considered**
Alternatives are: status quo, requiring builds from source; and offering container images, as described in #1927. Shipping container images doesn't help much in the case of `pcli`, though, so if folks need to compile `pcli`, they might as well compile `pd` and `pviewd`.

**Additional context**

Initial testing by @hdevalence showed that binaries using the `stable-x86_64-unknown-linux-gnu` target transferred fine between Debian 11 and Fedora 37, as long as the OS dependencies were installed. Let's give that a shot and see if they're useful for folks.

conorsch:
Let's try with https://github.com/axodotdev/cargo-dist#way-too-quick-start and see how that goes.

conorsch:
Took a first look at using `cargo-dist`. The tool has some rough edges still—it underwent a [full rewrite between 0.0.2 and 0.0.3](https://blog.axo.dev/2023/02/cargo-dist-rewrite), just a few weeks ago, for example—but still gives us a ton of functionality cheaply. Tinkering with it surfaced a few workflow snags that we should address on our own, prior to adoption.

### Crate versions
All our crate versions—which are not yet published to crates.io—are currently set naively to `0.1.0`, and we do not update those versions when releasing new testnets. When generating a release, `cargo-dist` will create artifacts named like so:

```
target/distrib/
├── pcli-v0.1.0-x86_64-unknown-linux-gnu
├── pd-v0.1.0-x86_64-unknown-linux-gnu
└── penumbra-view-v0.1.0-x86_64-unknown-linux-gnu
```

To resolve, we should re-version these crates to `0.0.n` where `n` is the version of the testnet. Then, releasing a new testnet would involve bumping all these versions in lockstep. (In the future, we can use tooling like [cargo-release](https://github.com/crate-ci/cargo-release) to keep that process scriptable.) 

### Tag/version format
[...]

PR #2404, opened 2023-04-24
docs: update release steps

conorsch:
Documents the new release workflow, based on cargo-release, as described in #1961. We used these steps for the first time in testnet 51 #2360.


conorsch:
Updated docs to include slight tweaks based on discussion in https://github.com/penumbra-zone/penumbra/pull/2405#issuecomment-1522027080. Considering these docs adjustments "good enough" for our purposes with next testnet, but let's make sure to update again after 52 if anything's confusing. 

Completes and therefore closes #1961.

PR #2405, opened 2023-04-25
view,custody: make account_group_id be actually optional

hdevalence:
It's marked as an optional field in the protos, but it wasn't treated as optional by the rust view server.  This is annoying, since it makes it very hard to use the view RPCs without already having the account group ID in hand. We should let it be optional, since in every current implementation of the view RPC, the account group ID is already implied by the endpoint (all current implementations only support one account group).

### DIFF
--- a/custody/src/request.rs
+++ b/custody/src/request.rs
@@ -10,7 +10,7 @@ pub struct AuthorizeRequest {
     /// The transaction plan to authorize.
     pub plan: TransactionPlan,
     /// Identifies the FVK (and hence the spend authorization key) to use for signing.
-    pub account_group_id: AccountGroupId,
+    pub account_group_id: Option<AccountGroupId>,
     /// Optionally, pre-authorization data, if required by the custodian.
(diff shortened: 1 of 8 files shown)


hdevalence:
@conorsch I did this directly off of the `0.51` tag, since it might be useful for the strangelove team, WDYT of using it as an opportunity to test the minor-version deployment flow? Maybe we could cherry-pick it into an `0.51.x` branch and then try doing an `0.51.1` release.

hdevalence:
I made a `v0.51.x` branch we could use.

I think my preference would be to make changes against `main` and then later cherry-pick them into the release branch. This seems like much less work, and ensures we never have outstanding changes that aren't included in the `main` branch. I'm not sure we need to PR the cherry-picks; we could just push them up and see that CI-on-push succeeds, then tag the commit on the release branch. This way we keep the process relatively lightweight. Ideally we should not spend a lot of time backporting fixes at this stage.

PR #2414, opened 2023-04-25
Consolidate transaction info request RPCs

### LINKED ISSUE #2311: Unify transaction query view service RPCs

Currently we have somewhat awkward and inconsistent RPCs in the view service for getting transactions / transaction views / transaction perspectives:

- [`TransactionByHash`](https://buf.build/penumbra-zone/penumbra/docs/main:penumbra.view.v1alpha1#penumbra.view.v1alpha1.ViewProtocolService.TransactionByHash)
- [`TransactionPerspective`](https://buf.build/penumbra-zone/penumbra/docs/main:penumbra.view.v1alpha1#penumbra.view.v1alpha1.ViewProtocolService.TransactionPerspective)
- Nothing that gives a pre-built `TransactionView`
- [`Transactions`](https://buf.build/penumbra-zone/penumbra/docs/main:penumbra.view.v1alpha1#penumbra.view.v1alpha1.ViewProtocolService.Transactions) which streams a collection of `TransactionsResponse` (an entirely different object than a `TransactionByHashResponse`!)
- [`TransactionHashes`](https://buf.build/penumbra-zone/penumbra/docs/main:penumbra.view.v1alpha1#penumbra.view.v1alpha1.ViewProtocolService.TransactionHashes) which is like `Transactions` but only gives the hashes (it's not clear that we really save much bandwidth here)

Unify and commonize these RPCs, handing out all relevant data and allowing the client to decide what they want to use. 

First, define a common
```proto
message TransactionInfo {
    // The height the transaction was included in a block, if known.
    uint64 height = 1;
    // The hash of the transaction.
    transaction.Id id = 2;
    // The transaction data itself.
    Transaction transaction = 3;
    // The transaction perspective, as seen by this view server.
    TransactionPerspective perspective = 4;
    // A precomputed transaction view of `transaction` from `perspective`, included for convenience of clients that don't have support for viewing transactions on their own.
    TransactionView view = 5;
}
```

Then, replace all of these RPC methods with just two:

- `TransactionByHash`, which returns a `TransactionInfo`;
- `Transactions`, which streams `TransactionInfo`s.

In other words, we change the view server to always return all possible information about the requested transactions.

aubrika:
Closes #2311


transaction/src/view/transaction_info.rs
```diff
@@ -0,0 +1,15 @@
...
+    pub height: u64,
+    // The hash of the transaction.
+    pub id: Id,
+    // The transaction data itself.
+    pub transaction: Transaction,
+    // The transaction perspective, as seen by this view server.
+    pub perspective: TransactionPerspective,
+    // A precomputed transaction view of `transaction` from `perspective`, included for convenience of clients that don't have support for viewing transactions on their own.
+    pub view: TransactionView,
+}
```
hdevalence, line 15: This structure is RPC-specific, so it should live in the view crate rather than the transaction crate.  And we won't be able to use it without defining a bunch of boilerplate conversions (it's not registered as a domain type, so we'd need a bunch of tryfrom/into).

Instead I think we could drop it entirely and just work with the proto types.

view/src/client.rs
```diff
@@ -146,32 +145,18 @@ pub trait ViewClient {
...
+    ) -> Pin<Box<dyn Future<Output = Result<TransactionInfo>> + Send + 'static>>;
 
     /// Queries for transactions in a range of block heights
-    fn transactions(
+    fn transaction_info(
         &mut self,
         start_height: Option<u64>,
         end_height: Option<u64>,
-    ) -> Pin<Box<dyn Future<Output = Result<Vec<(u64, Transaction)>>> + Send + 'static>>;
+    ) -> Pin<Box<dyn Future<Output = Result<Vec<TransactionInfo>>> + Send + 'static>>;
```
hdevalence, line 159: This could be nicer if it took a https://doc.rust-lang.org/std/ops/struct.Range.html

hdevalence: never mind, it should really be a https://doc.rust-lang.org/std/ops/trait.RangeBounds.html but we can't use that because it would make the trait not be object safe

view/src/client.rs
```diff
@@ -648,82 +633,47 @@ where
...
+            let perspective = rsp.perspective.unwrap_or_default();
+            let view = rsp.view.unwrap_or_default();
+
+            let tx_info = TransactionInfo {
+                height: rsp.height.unwrap().into(),
+                id: rsp.id.unwrap().try_into()?,
+                transaction: transaction.try_into()?,
+                perspective: perspective.try_into()?,
+                view: view.try_into()?,
+            };
```
hdevalence, line 662: We can't `unwrap` here, we need to `ok_or_else` and error

PR #2418, opened 2023-04-26
vendor verify_membership and verify_non_membership from ibc-types

avahowell:


hdevalence:
We decided not to do this, and to work on splitting out the `ibc-types` subcrates separately from the current `ibc-types` crate.

PR #2424, opened 2023-04-27
router(dex): document safety assumptions in routing execution

erwanor:
The initial goal of this PR was to prove (via a dust fill test) that during routing execution, it was possible to craft positions such that a saturating input would be zero. It turns out that this isn't possible, so this PR instead lays out why this isn't possible, and document various `SAFETY` assumptions that the DEX engine makes.


app/src/dex/router/fill_route.rs
```diff
@@ -188,11 +210,20 @@ pub trait FillRoute: StateWrite + Sized {
...
+                // There are two reasons why no saturating can be zero:
+                // + delta_1_star = ceil(delta_1), so unless delta_1 == 0, we
+                // have delta_1_star > 0.
+                // + delta_1 = lambda_2 * accumulated_effective_price
+                // And lambda_2 != 0, since lambda_2 = r2 and r2 != 0 because
+                // otherwise it would not have been indexed for that pair's
+                // direction.
+                // + accumulated_effective_price != 0, because for each position
+                //   we have p, q != 0.
+                unreachable!("current_value is nonzero")
```
erwanor, line 226: @hdevalence you were right! it should be impossible to hit that branch, which is a relief. In the process of proving why i realized that a much more load bearing aspect of routing is in the end-to-end effective price calculation, so this is the next item on my TODO of double-checking assumptions. They are already marked as TODOs in the code.

PR #2448, opened 2023-05-02
dex: add epoch height to `BatchSwapOutputData`

### LINKED ISSUE #2351: Include epoch and intra-epoch block height in `BatchSwapOutputData`
Currently, the `BatchSwapOutputData` uses the global block height of inclusion, and the associated proofs used math on the epoch to calculate the global block height (since the commitment only has the intra-epoch block height available).

Now that we are using non-fixed block length epochs, the swap claim proof can no longer assume the epochs are fixed-length. An easy fix would be to include the epoch height and an intra-epoch block height in the `BatchSwapOutputData` in place of (or alongside) the global block height.

redshiftzero:
Closes #2351

Note also that I don't think we need the `epoch_duration` on the `SwapClaim`, but I don't want to make a breaking change to the protos.

### DIFF
--- a/app/src/action_handler/actions/swap_claim.rs
+++ b/app/src/action_handler/actions/swap_claim.rs
@@ -20,7 +20,6 @@ impl ActionHandler for SwapClaim {
                 context.anchor,
                 self.body.nullifier,
                 self.body.output_data,
-                self.epoch_duration,
                 self.body.output_1_commitment,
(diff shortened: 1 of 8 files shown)


zbuc:
The `epoch_duration` on SwapClaim is definitely not needed any more after this

zbuc (approved):

PR #2453, opened 2023-05-03
Fix bad re-streams

zbuc:
There were some instances where we were attempting to re-stream streams across RPC boundaries in a way that only returned the last item. This replaces all instances of that pattern with a simpler implementation using stream combinators.


zbuc:
@plaidfinch i haven't directly tested it but yes it should

PR #2456, opened 2023-05-03
ci: cancel pending preview deploys

conorsch:
Sometimes we merge several PRs in succession, and currently that causes preview deploys to trigger in parallel, which leads to a race on container build and cluster deploy. Let's instead always prefer the most recent merge, cancelling prior builds if they're still ongoing.

### DIFF
--- a/.github/workflows/deploy-preview.yml
+++ b/.github/workflows/deploy-preview.yml
@@ -16,6 +16,12 @@ on:
     branches:
       - main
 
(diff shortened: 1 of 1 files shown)


conorsch:
Smoke test is failing, unrelated to this PR, same as https://github.com/penumbra-zone/penumbra/pull/2460#issuecomment-1533720567

PR #2460, opened 2023-05-03
ci: update rust cache action v1 -> v2

conorsch:
We've gone over the 10GB per-repo limit for workflow caches on GH. GH will automatically evict caches not used for a week or longer, but we've had a bunch of active PRs with plenty of rebases, and went over, ~22GB/10GB. Manually cleaned up old caches in the GitHub UI [1] to unbreak workflow runs. This change brings us up to speed with the latest release for the associated GH action helper, which has some updated caching settings in the changelog. May need to follow up with refinements to ensure that we don't hit the cache limit again.

N.B. GH does not allow us to purchase a cache increase per repo, which is unfortunate.

[0] https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy
[1] https://github.com/penumbra-zone/penumbra/actions/caches

### DIFF
--- a/.github/workflows/notes.yml
+++ b/.github/workflows/notes.yml
@@ -24,7 +24,7 @@ jobs:
           toolchain: nightly
           override: true
       - name: Load Rust caching
-        uses: Swatinem/rust-cache@v1
+        uses: Swatinem/rust-cache@v2
       - name: Load get-version action to grab version component of deployment path
         uses: battila7/get-version-action@v2
         id: get_version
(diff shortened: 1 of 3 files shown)


conorsch:
Smoke test [is failing](https://github.com/penumbra-zone/penumbra/actions/runs/4876051650/jobs/8699049039?pr=2460) on swap integration test:

```
error: test failed, to rerun pass `-p pcli --test network_integration`
failures:

failures:
    swap

test result: FAILED. 6 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 164.19s
```

Rerunning...

redshiftzero:
Heads up the smoke test CI failure is unrelated to the changes in this PR and is occurring on main so I think it's safe to ignore here

conorsch:
Missed that it was on main already, tks!

PR #2463, opened 2023-05-03
test(integration): reverse swap direction

redshiftzero:
This is predicated on getting at least 0.01gn which seems reasonable?


redshiftzero:
Closed in favor of #2469

PR #2464, opened 2023-05-03
ibc: improve logging

avahowell:
differentiate between client type and client state lookups


crates/ibc/src/component/client.rs
```diff
@@ -265,14 +265,14 @@ pub trait StateReadExt: StateRead {
     async fn get_client_type(&self, client_id: &ClientId) -> Result<ClientType> {
         self.get_proto(&state_key::client_type(client_id))
             .await?
-            .ok_or_else(|| anyhow::anyhow!("client not found"))
+            .context("could not find client type")
```
conorsch, line 268: What about "could not find client type for {client_id}"?

avahowell: good idea, done

PR #2465, opened 2023-05-03
dev: script for relayer local devnet

conorsch:
Adds a repeatable script for debugging and testing the relayer, which attempts to build a path between the preview deployment and a local devnet, ignoring testnet entirely. This allows us to test the very latest versions of the Penumbra code, and rules out any variation caused by proto definition changes or other code drift.

Renames the `run-relayer` to `build-path"`to describe accurately what it does. After path is built, then one can run `rly start` but we're not quite ready for that yet.

Also updates the `configure-relayer` script to target devnet & preview, rather than preview & testnet. N.B. the directionality is important here: we want to be able to inspect the local devnet logs and see the relayer's events in pd's logs.

Maybe worth editing the scripts to allow for this customization ad-hoc, not bothering right now.


conorsch:
We've been using this to spin up testing environments while chasing #2330, submitting to main so we don't have to keep rebasing it.

PR #2469, opened 2023-05-04
test(integration): Fix dex smoketests

zbuc:
The existing dex smoketests assumed the genesis liquidity pools would exist and were failing as a result.

They've been updated to exercise liquidity position opening to enable the swap to run.

We should probably add a full LP lifecycle test.


hdevalence:
Rather than use `1cube`, which is an NFT more subject to rounding errors, I think we should use `gm` or `gn`.

PR #2472, opened 2023-05-04
ibc: connection_open_try: fix client id handling

avahowell:
we were using the wrong client ID here; this wasn't noticed because our testing coincidentally used chains with the same client count, resulting in the counterparty & base chain client ids always being the same.


conorsch:
Refs #2330, specifically https://github.com/penumbra-zone/penumbra/issues/2330#issuecomment-1533875743. 

Looks great. Will merge as soon as CI passes, so we can test-drive on preview.

PR #2494, opened 2023-05-05
dex_utils: map individual payoffs to `Position`s

### LINKED ISSUE #2477: Map the approximated payoffs into `Position`s

erwanor:
Closes #2477

### DIFF
--- a/crypto/src/fixpoint.rs
+++ b/crypto/src/fixpoint.rs
@@ -12,6 +12,9 @@ use ethnum::U256;
 #[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
 pub struct U128x128(U256);
 
+// binary repr of 2^128
+const BASE: u64 = 0x47f0000000000000;
+
 #[derive(thiserror::Error, Debug)]
 pub enum Error {
     #[error("attempted to convert non-integral value {value:?} to an integer")]
--- a/crypto/src/fixpoint.rs
+++ b/crypto/src/fixpoint.rs
@@ -12,9 +12,6 @@ use ethnum::U256;
 #[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
 pub struct U128x128(U256);
 
-// binary repr of 2^128
-const BASE: u64 = 0x47f0000000000000;
-
 #[derive(thiserror::Error, Debug)]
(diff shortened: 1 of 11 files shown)


crypto/src/fixpoint.rs
```diff
@@ -36,17 +39,47 @@ impl Debug for U128x128 {
...
+    let fractional_as_u128 = (fractional * f64::from_bits(BASE)).trunc() as u128;
+
+    // Combine the integral and fractional parts into a U128x128.
+    let combined = U256::from_words(integral, fractional_as_u128);
+    U128x128(combined)
+}
+
+/// Converts an `U128x128` into an `f64`, this is a hack to make the `Display`
+/// impl work. This should NOT be used in mission critical contexts.
+pub fn to_f64_unsafe(value: &U128x128) -> f64 {
```
hdevalence, line 75: There's a conflict since I promoted this to a first-class `From` impl.  I think it's fine to have a conversion; floats aren't exact so shouldn't be used in consensus-critical contexts anyways, but not having conversions at all makes things difficult to use.

crypto/src/fixpoint.rs
```diff
@@ -36,17 +39,47 @@ impl Debug for U128x128 {
...
+    // Convert the fractional part to an unsigned 128-bit integer.
+    // 1. Multiply the fractional part by 2^128 to move the significant bits to the left.
+    // 2. Round down the result to the nearest integer.
+    // 3. Convert the rounded result to an unsigned 128-bit integer.
+    let fractional_as_u128 = (fractional * f64::from_bits(BASE)).trunc() as u128;
+
+    // Combine the integral and fractional parts into a U128x128.
+    let combined = U256::from_words(integral, fractional_as_u128);
+    U128x128(combined)
+}
```
hdevalence, line 71: I think this should probably just be a `TryFrom` impl, and we should document that it's only an approximate conversion. It definitely shouldn't panic, though, I don't think that's a good way to discourage bad use, it just makes everything more crashy.

erwanor: Fair enough, this is fixed in the latest batch of changes!

erwanor:
I'm going to work out pretty printing positions and interactivity in a follow-up PR, merging early.

Issue #1887, opened 2023-01-25
Add support for HTTPS calls via pcli


conorsch:
The `pcli` binary should be able to talk to HTTPS endpoints. Right now, it assumes all connections are HTTP, but we should update the option handling to infer the proper scheme based on a [`url::Url`](https://docs.rs/url/latest/url/struct.Url.html) or similar and make the relevant requests. 

We already have a long-running HTTPS endpoints for the testnet deployments: `https://{,g}rpc.testnet{,-preview}.penumbra.zone ` Traffic on 443/TCP to those FQDNs will talk to pd or tm as necessary. Let's use those to build the new option handling, and default client requests to an HTTPS endpoint in future releases.

conorsch:
Took a first look at this, documenting some progress below. The test command is `pcli view sync`.

### Naive approach
After updating the relevant args to default to an HTTPS URL for `https://grpc.testnet.penumbra.zone`:
```
Error: status: Unknown, message: "transport error", details: [], metadata: MetadataMap { headers: {} }

Caused by:
    0: transport error
    1: http2 error: connection error detected: frame with invalid size
    2: connection error detected: frame with invalid size

```
The error messages are less descriptive than I'd like, but they make sense: we haven't even clicked on TLS support in tonic yet.

### Features: tls
Updating URLs to be HTTPS, and enabling the ["tls" feature for tonic](https://docs.rs/tonic/latest/tonic/#feature-flags):


```
Error: transport error

Caused by:
    0: error trying to connect: Connecting to HTTPS without TLS enabled
    1: Connecting to HTTPS without TLS enabled
```

Progress. Docs indicate that we may need to pass in a [ClientTlsConfig](https://docs.rs/tonic/latest/tonic/transport/struct.ClientTlsConfig.html) struct; or, we can try more features

### Features: tls & tls-roots
Also enabling feature `tls-roots` for tonic:

```
Error: status: Internal, message: "grpc-status header missing, mapped from HTTP status code 400", details: [], metadata: MetadataMap { headers: {"content-length": "0", "date": "Tue, 31 Jan 2023 23:14:33 GMT", "via": "1.1 google", "alt-svc": "h3=\":443\"; ma=2592000,h3-29=\":443\"; ma=2592000"} }
```

OK, that's much better: we're now establishing a valid connection over HTTPS and receiving a response. The headers match what we'd see via curl, leaving grpc aside:

```
curl -I https://grpc.testnet.penumbra.zone
[...]

PR #2497, opened 2023-05-05
Fix AddressByIndex handler for cases of AddressIndexes with empty randomizers

### LINKED ISSUE #2490: Rust view server should allow default randomizers in `AddressByIndex`
**Describe the bug**

The `AddressByIndex` request handling currently errors when an empty `randomizer` slice is passed (which is what will happen if the caller doesn't include it).

Instead of forcing callers to pass a 12-byte array of zeros, we should internally use a `[0u8; 12]` whenever the slice is empty.

This came up when the Strangelove team tried to pull `pclientd` into InterchainTest.

aubrika:
Closes #2490

### DIFF
--- a/crypto/src/keys/diversifier.rs
+++ b/crypto/src/keys/diversifier.rs
@@ -245,13 +245,18 @@ impl TryFrom<pb::AddressIndex> for AddressIndex {
     type Error = anyhow::Error;
 
     fn try_from(d: pb::AddressIndex) -> Result<AddressIndex, Self::Error> {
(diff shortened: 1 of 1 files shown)


hdevalence:
I tested this locally using Buf studio and confirmed that it worked.  I also discovered that removing HTTP/1 support, which [broke the web extension for any non-https endpoints](https://github.com/penumbra-zone/penumbra/issues/1887#issuecomment-1503909705), broke our CORS headers, making it impossible to query a local view service.

Issue #2499, opened 2023-05-05
DEX panic during execution


hdevalence:
**Describe the bug**

Logs here: https://gist.github.com/hdevalence/a4aa06b19594219ae1ff58c5f05c494b

**To Reproduce**

```
cargo run --release --bin pcli -- -n http://localhost:8080 view reset
cargo run --release --bin pcli -- -n http://localhost:8080 view balance
cargo run --release --bin pcli -- -n http://localhost:8080 tx lp order sell 100gm@1gn
cargo run --release --bin pcli -- -n http://localhost:8080 tx lp order sell 100gm@1gn --spread 10
cargo run --release --bin pcli -- -n http://localhost:8080 tx lp order sell 100gm@1gn --spread 25
cargo run --release --bin pcli -- -n http://localhost:8080 tx lp order sell 100gm@1gn --spread 50
cargo run --release --bin pcli -- -n http://localhost:8080 tx lp order sell 80gn@1gm
cargo run --release --bin pcli -- -n http://localhost:8080 tx lp order sell 80gn@1gm --spread 10
cargo run --release --bin pcli -- -n http://localhost:8080 tx lp order sell 80gn@1gm --spread 25
cargo run --release --bin pcli -- -n http://localhost:8080 tx lp order sell 80gn@1gm --spread 50
cargo run --release --bin pcli -- -n http://localhost:8080 q dex positions --help
cargo run --release --bin pcli -- -n http://localhost:8080 q dex help
cargo run --release --bin pcli -- -n http://localhost:8080 tx lp order sell 10penumbra@2gm
cargo run --release --bin pcli -- -n http://localhost:8080 tx lp order sell 27penumbra@2.5gn
cargo run --release --bin pcli -- -n http://localhost:8080 tx lp order sell 30penumbra@3gm --spread 100
cargo run --release --bin pcli -- -n http://localhost:8080 tx lp order sell 1000gm@1.5test_usd --spread 25
cargo run --release --bin pcli -- -n http://localhost:8080 tx lp order sell 1000gn@1.5test_usd --spread 25
cargo run --release --bin pcli -- -n http://localhost:8080 q dex all-positions
cargo run --release --bin pcli -- -n http://localhost:8080 q dex positions test_usd:gm
cargo run --release --bin pcli -- -n http://localhost:8080 q dex positions test_usd:gn
cargo run --release --bin pcli -- -n http://localhost:8080 q dex positions gm:gn
[...]

PR #2500, opened 2023-05-05
dex: fill in position-by-price RPCs

hdevalence:
This allows querying what positions are available on the chain by price and pair.


hdevalence:
I spotted #2499 while testing this.

Issue #2505, opened 2023-05-06
Broken CORS headers for grpc-web


hdevalence:
**Describe the bug**

Our services no longer correctly set CORS headers, making it impossible to query them from studio.buf.build or [other web contexts](https://github.com/penumbra-zone/penumbra/issues/1887#issuecomment-1503909705). Restoring `accept_http1(true)` fixes the problem.

xref #2497

PR #2506, opened 2023-05-06
grpc: re-enable http/1

hdevalence:
This restores the ability to use web clients, like https://studio.buf.build, or the browser extension, against local instances of these services.  I also tested (by temporarily running this branch on `pd-tls.penumbra.zone` with auto-https enabled) to check that it doesn't interfere with https support, either for browser clients or for `pcli`.


hdevalence:
Closes #2505 , xref https://github.com/penumbra-zone/penumbra/issues/1887#issuecomment-1503909705

PR #2513, opened 2023-05-09
dex: rework fill_route to ensure we always make progress

### LINKED ISSUE #2507: Redesign routing execution
#### How does routing execution currently work?

We build off the assumption that there is a way to route a positive amount of capacity over the provided route that has an end-to-end effective that's better or similar to the specified `spill_price`. 

First, we pick the first position on the first hop and simulate execution over the entire route. If routing flow is bottlenecked at any position, we identify that position as a limiting constraint, track it, and continue routing capacity through the remaining hops. At the end of the simulation, we know whether we can execute the entirety of the first order, or if we must deal with a more constraining position first ("lift the constraint").

If no limiting hops were found, that's great we execute that amount and try again until spill price is hit. If some constraints were found, we pick the most limiting one, compute its saturating input i.e. $\Delta^*_1$ such that when we execute over the route again we exhaust that constraint. The reason we are able to do this without fearing getting bottlenecked by another hop along the route is because we explicitly picked the most limiting constraint i.e. no other position along the route will bottleneck execution. Once the constraint is lifted, we can proceed again from the start, trying to route as much flow as possible until the position is filled or the `spill_price` is hit. 

#### Why this does not work

However, the current design of routing execution does not take into account the possibility that lifting the most limiting constraint may result in an output that exceeds the capacity of the next constraint on the route. This is a capital error (manifested in #2484) and a serious flaw in the current design of routing execution.

##### Interim proposed solution:

In addition to selecting the saturating input ($\Delta^*_1$) that lifts the most limiting constraint, we should simulate execution and backtrack until we find some amount of routable flow across the route.  This solution adds overhead to an already quadratic algorithm, but will let us work on the DEX concurrently as we work to improve it. 

### Dust swaps and dust positions

Dust swaps and dust positions are a class of undesirable behavior that happen when provisioning an action (position, or swap) with infinitesimal amounts of reserves. While these limitations of the DEX engine are not directly related to routing execution, they are still important to consider and should be given priority in any future improvements to the system. The DEX engine works with limited precision: a `U128x128` fixed-point type which handles pricing and quantity transformations, and an integral `Amount` type which is 128 bits wide and stores denoms scaled numbers such as the reserves $R_1, R_2$, the objective valuations of each assets $p, q$, and the input/output amounts $\langle (\Delta_1, \Delta_2), (\Lambda_1, \Lambda_2) \rangle$

When we execute against a position, we feed a tuple $(\Delta_1, \Delta_2)$ and obtain an output $(\Lambda_1, \Lambda_2)$ from execution. Internally, execution is handled by multiplying the input $\Delta_1$ by the exchange rate $\frac p {q \gamma}$ between assets 1 and 2. However, when $\Delta_1 \lt \epsilon$ is very small, computing the output $\Lambda_2 = \lfloor {\Delta_1 \cdot \frac p q \cdot \gamma} \rfloor = 0$. (as manifested in #2499) as well as in the early design discussions about implementing routing execution. This is a dust swap i.e. a swap that submits inputs and receives no outputs, and the reason why the current dex design does not support NFT transactions of `cube`. Our thought process was that this was an acceptable tradeoff for our first iteration of the DEX engine, and that swap inputs were to be denom scaled anyway (i.e. swapping 1gm -> $\Delta_{gm} = 1,000,000$). However, in the process of executing a route, it is possible that a swap will be partially filled up to some dust amount, triggering this scenario, or that attempting to lift a very limiting constraint becomes tantamount to routing a dust flow that recreates this scenario, as it is the case in #2499. 

##### Interim proposed solution:

I propose that burning dust might actually be the correct behavior for the current system we have, which is not to say that a better system is not possible or what we should be aiming for.

hdevalence:
Closes #2507  , along the lines of the comments there.


app/src/dex/router/fill_route2.rs
```diff
@@ -0,0 +1,197 @@
...
+                .last()
+                .cloned()
+                .ok_or_else(|| anyhow::anyhow!("called fill_route with empty route"))?,
+        };
+
+        // TODO: this does not allow traversing the same macro-edge in opposite directions
+        // (This impl will produce WRONG RESULTS if you try to do that!)
+        // later: implement fast path for non-opposite directions, slow path (check for duplicate position ids)
+
+        let mut frontier = Vec::new();
```
hdevalence, line 68: I think we should change `frontier` to be a `Vec<(Position, position::Id)>`, saving the position IDs as well as the positions.  Then, before adding or replacing an existing position on the frontier, we should check whether its ID is already present in the frontier.

hdevalence: Actually, this could just be a standalone BTreeSet, seems more efficient

app/src/dex/router/tests.rs
```diff
@@ -568,7 +568,7 @@ async fn fill_route_constraint_stacked() -> anyhow::Result<()> {
     let spill_price = U128x128::from(Amount::from(1_000_000_000u64) * pusd.unit_amount());
 
     let (unfilled, output) =
-        FillRoute::fill_route(&mut state_tx, delta_1, &route, Some(spill_price))
+        FillRoute2::fill_route(&mut state_tx, delta_1, &route, Some(spill_price))
```
erwanor, line 571: Oh, fyi the `fill_route_*` tests are skipped at the moment!

app/src/dex/router/fill_route2.rs
```diff
@@ -0,0 +1,370 @@
...
+                        Amount::zero(),
+                        "unfilled amount for unconstrained frontier should be zero"
+                    );
+
+                    position.reserves = new_reserves;
+                    current_value = output;
+                    trace[i + 1].amount += output.amount;
+                }
+
+                tracing::debug!("filled input amount completely, breaking loop");
```
erwanor, line 180: I think we need to update `input.amount` and `output.amount` here.

```rust
            // Update the input and output amounts tracked outside of the loop:
            input.amount = input.amount - current_input.amount;
            output.amount = output.amount + current_output.amount;
 ```

like done further below

File: app/src/dex/router/fill_route2.rs [hunk 4]
```diff
@@ -0,0 +1,370 @@
...
+            // Work backwards along the path from the constraining position.
+            let mut current_value = exactly_consumed_reserves;
+            for i in (0..=constraining_index).rev() {
+                trace[i + 1].amount += current_value.amount;
+
+                let (new_reserves, prev_input) = frontier[i]
+                    .phi
+                    .fill_output(&frontier[i].reserves, current_value)
+                    .expect("asset ids should match")
+                    .expect("working backwards from most-constraining position should not exceed reserves");
```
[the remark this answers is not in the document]
hdevalence: But the last constraint _is_ the most constraining one, because we know all the positions are distinct, and we apply each position's constraints, if any, as we pass through it.

If there were an earlier constraining position, we would have reduced the amount of output we got for the next hop, so the fact that we see a later position as constraining _that_ output means that the later position is more constraining than the earlier one.

app/src/dex/router/tests.rs
hdevalence: Got it, I guess we should unskip them?

app/src/dex/router/fill_route2.rs
```diff
@@ -0,0 +1,370 @@
...
+                };
+
+                let next_position = self
+                    .position_by_id(&next_position_id)
+                    .await?
+                    .expect("indexed position should exist");
+
+                frontier_position_ids.insert(next_position_id);
+                frontier[constraining_index] = next_position;
+                // TODO: should we remove the old position's ID from the ID set?
```
erwanor, line 368: I think, but I'm not certain, that we are missing a `break 'next_position`. AFAICT, it would make sense to have it here when we have a bona fide `next_position` to roll with

app/src/dex/router/fill_route2.rs
```diff
@@ -0,0 +1,412 @@
...
+
+                // Having updated the frontier, it's time to break out of the loop.
+                break 'next_position;
+            }
+        }
+
+        // We need to save these positions, because we mutated their state, even
+        // if we didn't fully consume their reserves.
+        for position in frontier {
```
erwanor, line 381: I think this should be in the filling loop, otherwise we might execute against stale orders on later fils?

PR #2520, opened 2023-05-09
Refactor Denom to DenomMetadata for compatibility with Cosmos SDK

### LINKED ISSUE #2310: Represent denom metadata compatibly with Cosmos SDK

aubrika:
Closes #2310


crypto/src/asset.rs
```diff
@@ -21,7 +23,7 @@ pub use registry::{Registry, REGISTRY};
 #[serde(try_from = "pb::Asset", into = "pb::Asset")]
 pub struct Asset {
     pub id: Id,
-    pub denom: Denom,
+    pub denom: DenomMetadata,
 }
```
hdevalence, line 27: What's the point of the `Asset` proto message, now that we have `DenomMetadata` ?

aubrika: Hmm, yes, seems redundant. Should I add removing `Asset` and replacing with `DenomMetadata` to this refactor?

crypto/src/asset.rs
```diff
@@ -137,7 +144,7 @@ mod tests {
 
         assert_eq!(
             base_denom.best_unit_for(0u64.into()).to_string(),
-            "penumbra"
+            "upenumbra"
```
hdevalence, line 147: This should be set back to `penumbra`, the tweak to `best_unit_for` seems to have been dropped, which is also causing the smoke tests to fail

crypto/src/asset/denom.rs
```diff
@@ -1,360 +1,25 @@
...
-    /// Returns the "best" unit for the given amount (expressed in units of the
-    /// base denomination).
-    ///
-    /// This is defined as the largest unit smaller than the given value (so it
-    /// has no leading zeros when formatted).
-    pub fn best_unit_for(&self, amount: asset::Amount) -> Unit {
-        // Special case: use the default unit for 0
-        if amount == 0u64.into() {
-            return self.default_unit();
-        }
```
hdevalence, line 162: Here's the special case that seems to be missing in the new code

crypto/src/asset.rs
aubrika: If it was recent, it may not have been part of my latest rebase, or got lost in a merge. Will investigate and revert this change

crypto/src/asset/denom_metadata.rs
```diff
@@ -0,0 +1,414 @@
...
+    }
+
+    /// Returns the "best" unit for the given amount (expressed in units of the
+    /// base denomination).
+    ///
+    /// This is defined as the largest unit smaller than the given value (so it
+    /// has no leading zeros when formatted).
+    pub fn best_unit_for(&self, amount: asset::Amount) -> Unit {
+        for (unit_index, unit) in self.inner.units.iter().enumerate() {
+            let unit_amount = asset::Amount::from(10u128.pow(unit.exponent as u32));
```
hdevalence, line 214: Here's where `best_unit_for` moved to, without the newly added special case

crypto/src/asset/denom.rs
aubrika: Oh, that makes sense! All changes to `denom.rs` will not be taken into account on this branch on account of the renaming of that file/struct & creation of a new one

crypto/src/asset.rs
hdevalence: I don't know. This PR is already terrifyingly large, maybe we should try to land it and then do the `Asset` removal in another step?

aubrika: I was hoping you would suggest that :) sounds great

hdevalence:
This looks like a move in the right direction -- there are probably further things we'll want to do, but we should merge this now to avoid generating conflicts (cc @plaidfinch )

PR #2529, opened 2023-05-10
Disentangle compact block from shielded pool and dex

### LINKED ISSUE #2440: Disentangle SCT and CompactBlock responsibilities

As described in #2288, we want to have a "diamond" dependency graph, with a few core crates at the bottom of the diamond, with a "wide" group of components depending on each other, and then a few crates at the top that depend on all of the components and stitch them back together.

I tried doing this for the `dex` component in #2438, which failed because of a cyclic dependency I didn't notice before starting. The problem is that the `CompactBlock` is currently forced to be both at the top and at the bottom of the diamond:

- it must go at the bottom, because it's used to track state commitments inserted into the SCT, and many components need to insert state commitments into the SCT;
- it must go at the top, because those state commitments are tracked in the `CompactBlock` using `StatePayload`s, which are an enumeration of all possible kinds of state payload (note payloads, swap payloads, etc).  This is convenient for clients, but means the ordering of the interleaved payloads is critical.

This isn't just a problem of code quality but also an extensibility issue -- we want to be forward-compatible with many other types of state commitments, but we can't do that with this kind of circularity.

To fix this, we should disentangle the responsibilities of "SCT management" (which we'll push down to the bottom of the diamond) and the "state payload managment" (which we'll pull up to the top):

- the `penumbra-sct` crate should just manage the SCT, exposing a read-only interface and a higher-level `NoteManager`-style trait for mutation that allows:
  - adding an opaque state commitment to the tree, returning the position;
  - sealing a block/epoch (as currently implemented in the `app` crate)
- "middle" components (like the shielded pool) should be changed to use only the SCT, not the compact block, and use the ephemeral object store to record a `Vec<(Position, Payload)>` of data they wish to save for later (we don't need any fancy `im` stuff here, we can just clone it wastefully, like we already do with the existing `CompactBlock`, it doesn't even show up in profiles).
- the top-level `CompactBlock` management code should have a higher-level `NoteManager`-style trait that queries all relevant components for their payloads, extracts their lists of positioned payloads, "lifts" the per-component payload type to the common one, sorts by position so that the payloads are in the right order, and then assembles the `CompactBlock`.
  - I think this code will need to live in the `app` crate for now, at least until the `dex` crate can be extracted, because if it's in the `penumbra-compact-block` crate, it won't be able to pull from the dex component in the `app` crate.  But we can't extract the `dex` crate until this code exists.  So we should write it as a (relatively) self-contained module of the `app` crate that we can plan to extract together with the `dex` crate extraction (V2).

plaidfinch:
Resolves #2440.

### DIFF
--- a/crates/shielded-pool/src/component/note_manager.rs
+++ b/crates/shielded-pool/src/component/note_manager.rs
@@ -105,6 +105,16 @@ pub trait NoteManager: StateWrite {
         self.object_put(state_key::pending_notes(), payloads);
     }
 
+    async fn pending_note_payloads(&self) -> im::Vector<(tct::Position, NotePayload, NoteSource)> {
+        self.object_get(state_key::pending_notes())
+            .unwrap_or_default()
+    }
+
+    async fn pending_rolled_up_payloads(&self) -> im::Vector<(tct::Position, Commitment)> {
+        self.object_get(state_key::pending_notes())
+            .unwrap_or_default()
+    }
+
     #[instrument(skip(self, source))]
     async fn spend_nullifier(&mut self, nullifier: Nullifier, source: NoteSource) {
         tracing::debug!("marking as spent");
--- a/app/src/app/mod.rs
+++ b/app/src/app/mod.rs
@@ -15,7 +15,7 @@ use penumbra_crypto::NotePayload;
 use penumbra_ibc::component::IBCComponent;
 use penumbra_proto::DomainType;
 use penumbra_sct::component::SctManager;
-use penumbra_shielded_pool::component::ShieldedPool;
+use penumbra_shielded_pool::component::{NoteManager, ShieldedPool};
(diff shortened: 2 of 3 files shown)


crates/shielded-pool/src/component/note_manager.rs
```diff
@@ -16,6 +12,42 @@ use crate::{event, state_key};
...
 use super::SupplyWrite;
 
+#[derive(Clone)]
+pub enum StatePayload {
+    RolledUp(note::Commitment),
+    Note {
+        source: NoteSource,
+        note: Box<NotePayload>,
+    },
+}
```
hdevalence, line 22: This `StatePayload` is only a subset of the possible state payload variants, meaning that we'll need to have another `StatePayload` somewhere else with more variants, right?

I'm wondering if it would be simpler to not have this enum at all, and instead have separate handling for note payloads and bare commitments -- what's the situation where we want to be handling "either bare commitment OR note payload" that's not "either bare commitment OR note payload OR swap payload"?

Issue #2350, opened 2023-04-12
check and document all constraints added after out of circuit computation


redshiftzero:
In our R1CS gadgets the most likely places where constraints could be missing are anywhere we are manually doing out-of-circuit computation and then adding constraints based on the result. Before audit I think we should (1) Double check all these locations and (2) Ensure that documentation is added inline in the form of comments (similar to [here](https://github.com/penumbra-zone/penumbra/blob/208e94116d16df376b001b08f47d5ce38b95250a/crypto/src/asset/amount.rs#L86)).

These are the places we need to check: 

## decaf377

- [x] Square roots: https://github.com/penumbra-zone/decaf377/blob/main/src/r1cs/fqvar_ext.rs#L35 - addressed by https://github.com/penumbra-zone/decaf377/pull/52

## penumbra-crypto

- [x] Amount division: https://github.com/penumbra-zone/penumbra/blob/main/crypto/src/asset/amount.rs#L55
- [x] Applying a `Penalty`: https://github.com/penumbra-zone/penumbra/blob/main/crypto/src/stake/penalty.rs#L102
- [x] Constrain a < b for `TradingPair` 

## TCT

- [x] Computing `WhichWay` is done out of circuit and then witnessed in circuit: https://github.com/penumbra-zone/penumbra/blob/main/tct/src/r1cs.rs#L146 (note that doing this in circuit I think would solve #2135 but doing the computation involves [bit ops](https://github.com/penumbra-zone/penumbra/blob/208e94116d16df376b001b08f47d5ce38b95250a/tct/src/internal/path.rs#L87) which is why I did it this in this way)  - resolved by #2533
- [x] Adding variables for the epoch, block, and commitment index from position: https://github.com/penumbra-zone/penumbra/blob/main/tct/src/r1cs.rs#L32 - I think we need to add additional constraints here based on the position var - resolved by #2533

PR #2533, opened 2023-05-10
crypto: rework position and merkle path verification

### LINKED ISSUE #2135: bug: spend proofs do not verify for 0 valued spends
The idea with dummy spends was that spending notes of 0 value should always verify. This is not happening, as found by @zbuc. 

Reproducer here: https://github.com/penumbra-zone/penumbra/compare/dummy-spends-fail-merkle?expand=1

Two issues:
1. The proofs don't verify when the anchor of the dummy spend used at proving time is that from the dummy spend commitment proof. 
2. This bug was found because a note has 0 value but was not intended to be a dummy spend, meaning that indicating dummy spends only by their 0 value may not be the right path.

redshiftzero:
Closes #2135
This also resolves the last two checkboxes in #2350


redshiftzero:
Performance changes compared to `main`:
* Spend
    * Proving time: 2.19s -> 2.27s
    * Number of constraints: 33484 -> 33740
* Delegator vote
    * Proving time: 2.20s -> 2.24s
    * Number of constraints: 35497 -> 36425

PR #2536, opened 2023-05-11
Swap p/q for pcli order sell

zbuc:
This results in sell orders with the expected `p`/`q`

### DIFF
--- a/crypto/src/dex/lp.rs
+++ b/crypto/src/dex/lp.rs
@@ -1,10 +1,12 @@
 mod nft;
+mod order;
 mod reserves;
 mod trading_function;
 
 pub mod position;
 
 pub use nft::LpNft;
+pub use order::{BuyOrder, SellOrder};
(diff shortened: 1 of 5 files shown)


hdevalence:
Why are these the expected ones? Out of the 4 combinations of (buy, sell) x (direction choice), which two do we expect to be equivalent and which two do we expect to be distinct?

zbuc:
> Why are these the expected ones? Out of the 4 combinations of (buy, sell) x (direction choice), which two do we expect to be equivalent and which two do we expect to be distinct?

Since `p/q` expresses the price relationship between `asset_1` and `asset_2` shouldn't it be the same for all orders at the same price, i.e.

`buy 1btc@2700usd` should have the same `p/q` as `sell 1btc@2700usd`
and the same `p/q` as `buy 0.000036btc@1usd` and `sell 0.000036btc@1usd`

PR #2538, opened 2023-05-11
dex: prelude to on-chain arb implementation

hdevalence:
This PR has changes along the way to starting an arb implementation, split out since they're all mostly finished.

### DIFF
--- a/app/src/dex/router/route_and_fill.rs
+++ b/app/src/dex/router/route_and_fill.rs
@@ -41,7 +41,7 @@ pub trait HandleBatchSwaps: StateWrite + Sized {
         let traces: im::Vector<Vec<Value>> = im::Vector::new();
         Arc::get_mut(self)
             .expect("one mutable reference to state")
-            .object_put("swap_execution", traces);
+            .object_put("trade_traces", traces);
 
         // Depending on the contents of the batch swap inputs, we might need to path search in either direction.
         let (lambda_2, unfilled_1) = if delta_1.value() > 0 {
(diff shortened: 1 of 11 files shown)


app/src/dex/router/route_and_fill.rs
```diff
@@ -93,42 +86,45 @@ pub trait RouteAndFill: StateWrite + Sized {
             lambda_2_2: unfilled_2,
         };
 
+        // TODO: how does this work when there are trades in both directions?
+        // Won't that mix up the traces? Should the SwapExecution be indexed by
+        // the _DirectedTradingPair_?
```
hdevalence, line 91: @zbuc I noticed this and was a bit confused, don't we want to be recording two distinct executions for each pair per block? otherwise the traces would get mixed up between them, right? Or?

app/src/dex/router/route_and_fill.rs
```diff
@@ -164,6 +159,7 @@ trait RouteAndFillInner: StateWrite + Sized {
 
                 tracing::debug!(?path, delta_1 = ?delta_1.amount, "found path, starting to fill up to spill price");
 
+                // TODO: in what circumstances should we use fill_route_exact?
```
hdevalence, line 162: Ideally we'd have one `route_and_fill` method that we could use for both swaps and arbitrage.  I'm not sure, however, how we want to handle price limits in the two cases, relative to the changes in #2535

app/src/dex/router/route_and_fill.rs
zbuc: I think you're right. That does seem wrong. It seems like we need to execute a batch swap per trading direction.

PR #2539, opened 2023-05-11
Implement arbitrage

hdevalence:
Builds on #2538


hdevalence:
The current commit is triggering an `.expect` in the filling code, which should not happen: https://gist.github.com/hdevalence/4a5e6c7e8cf8fb981528fa455c4a8e43

Since that code wasn't changed by this PR, we should figure out what problem is uncovered by this PR.

PR #2555, opened 2023-05-12
crypto: fixes to TradingPair allocation and division of Amounts

### LINKED ISSUE #2350: check and document all constraints added after out of circuit computation

redshiftzero:
Closes #2350

This PR:
- Reworks the allocation of `TradingPairVar` (represents a `TradingPair` in R1CS), to note that we are intentionally not checking that the asset IDs are ordered as expected in the canonical representation of the out of circuit `TradingPair`. This is because we are only using the `TradingPair`'s two asset IDs in the swap commitment integrity check, which will fail if the asset IDs are switched.
- Fixes two issues with the R1CS implementation of Euclidean division of `AmountVar`s, where we now ensure that the remainder cannot equal the divisor, and that the divisor is non-zero. 
- Adds tests for all circuits to ensure that the parameters and circuits are in sync, i.e. that a circuit-breaking change has not unintentionally been made.


crypto/src/dex/trading_pair.rs
```diff
@@ -121,6 +123,40 @@ impl TryFrom<[u8; 64]> for TradingPair {
...
+        let ns = cs.into();
+        let cs = ns.cs();
+        let trading_pair = f()?.borrow().clone();
+        let asset_1 = AssetIdVar::new_variable(cs.clone(), || Ok(trading_pair.asset_1()), mode)?;
+        let asset_2 = AssetIdVar::new_variable(cs, || Ok(trading_pair.asset_2()), mode)?;
+        // A valid `TradingPair` has a canonical ordering.
+        // Constrain: asset_1 <= asset_2.
+        asset_1
+            .asset_id
+            .enforce_cmp(&asset_2.asset_id, core::cmp::Ordering::Less, true)?;
```
redshiftzero, line 147: I think this doesn't work because `enforce_cmp` [expects](https://docs.rs/ark-r1cs-std/latest/ark_r1cs_std/fields/fp/enum.FpVar.html#method.enforce_cmp) the variables to be `<= (p - 1)/2` in size. The other places where we use the `enforce_cmp` method should be checked too.

PR #2556, opened 2023-05-12
docs: add grpc docs to dev guide

conorsch:
Documents basic workflow for using Buf Studio to interact with the gRPC endpoints for pd and pclientd.

### DIFF
--- a/docs/guide/src/dev/rpc.md
+++ b/docs/guide/src/dev/rpc.md
@@ -1,0 +1,49 @@
+# Using Buf Studio
+
+The [Buf Studio](https://studio.buf.build) webapp provides a GUI
+for interacting with gRPC endpoints. Developers can use it to explore Penumbra's
+use of gRPC, for example, viewing transactions and transaction plans.
+
+## Using the public testnet
+
+<!---
+The use of the buf-studio agent is only necessary because the public gRPC endpoint
+is HTTP-only; as soon as we have HTTPS support on the public testnet,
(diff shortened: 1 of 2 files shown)


docs/guide/src/dev/rpc.md
```diff
@@ -0,0 +1,46 @@
...
+use of gRPC, for example, viewing transactions and transaction plans.
+
+## Using the public testnet
+
+<!---
+The use of the buf-studio agent is only necessary because the public gRPC endpoint
+is HTTP-only; as soon as we have HTTPS support on the public testnet,
+we can revise these docs to drop use of the agent, and connect directly to the HTTPS
+endpoint. See for details: https://github.com/penumbra-zone/penumbra/issues/2341
+--->
```
conorsch, line 14: These HTML comments _will_ render in the final document, visible when viewing the page source. That's fine!

docs/guide/src/dev/rpc.md
```diff
@@ -0,0 +1,46 @@
...
+
+To get started quickly, you can use the publicly available gRPC endpoint
+from the testnet deployment run by Penumbra Labs, `http://testnet-preview.penumbra.zone:8080`.
```
conorsch, line 17: Intentionally using `preview` in the URL because the http1 support from https://github.com/penumbra-zone/penumbra/pull/2506 hasn't landed on `testnet` yet. That'll change next week, at which point we can update this. Or, we can just leave as-is for now and expand these docs as we go.

PR #2559, opened 2023-05-13
crypto(fixpoint): add overflow failure test

erwanor:
This test should fail and overflow loudly, but instead it fails silently producing a nonsensical value.

### DIFF
--- a/crypto/src/fixpoint.rs
+++ b/crypto/src/fixpoint.rs
@@ -94,8 +94,11 @@ impl U128x128 {
 
     /// Performs checked multiplication, returning `Some` if no overflow occurred.
     pub fn checked_mul(self, rhs: &Self) -> Option<Self> {
-        let [x0, x1] = self.0 .0;
-        let [y0, y1] = rhs.0 .0;
+        // It's important to use `into_words` because the `U256` type has an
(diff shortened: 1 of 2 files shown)


hdevalence:
The bug is in this code I wrote in `U128x128::checked_mul`:
```
        x1y1.checked_shl(128)
            .and_then(|acc| acc.checked_add(x0y1))
            .and_then(|acc| acc.checked_add(x1y0))
            .and_then(|acc| acc.checked_add(x0y0 >> 128))
            .map(U128x128)
```
The first call was intended to return None when there are nonzero bits in the high part of the product. But that's not what [checked_shl](https://doc.rust-lang.org/std/primitive.u128.html#method.checked_shl) does: the check is on the shift amount, not the shifted bits. Oops.

To remediate this, we should:

- [ ] Use U256::into_words to split the high and low parts of x1y1, and check that the high part is zero, erroring out if so
- [ ] Add a `fixpoint::Error` enum with an `Error::Overflow` variant using thiserror, and swap out the use of Options in the U128x128 API for Results
- [ ] Make a plan of how to review use of the U128x128 arithmetic in the DEX code to check that it either doesn't overflow or errors out semi-gracefully

hdevalence:
(The last point can be a followup issue)

PR #2561, opened 2023-05-14
fix: temporarily limit dex values to 60 bits

### LINKED ISSUE #2560: Temporarily avoid large values in DEX calculations

#2559 identified a bug in the `U128x128` fixed-point arithmetic implementation that hid integer overflows -- which were supposed to cause the computation to fail.  Because of this, at this instant we don't have confidence that the DEX implementation correctly prevents or handles overflows.  While the test suite passes after correcting the bug, if there are any problems, we won't have noticed them during existing experimentation.  Gaining this confidence will require going through the DEX implementation and analyzing bounds on values used during the computation.

However, we'd still like to ship the first working version of the DEX on the original timeline (on Monday), so that people can start experimenting with the economic mechanisms.  For that reason, it would be useful to find a way to avoid the possibility of integer overflows in the near term, while we take time to polish the code and gain assurance that they're not possible.  To do this, we can temporarily disallow large values in DEX calculations, keeping them small enough to be reasonably sure that overflows can't happen.

- [ ] Add a new, temporary consensus rule to the `PositionOpen` action handler that checks that `p`, `q`, `r1`, `r2` are all less than 60 bits in size (i.e., are less than `1u128 << 60`).  Then multiplying them can only result in a 120-bit value, giving 8 bits of headroom.  Positions exceeding this bound should be rejected with an error message indicating they will be supported in a future testnet. Adding this as a new consensus rule emphasizes that it's temporary and special-purpose.

This doesn't pose a problem for most native assets, which use 1e6 display denoms, but does pose a problem for `test_eth` and `test_usd`, which use 1e18 to simulate large-denom assets.  To handle this, we should:

- [ ] Temporarily redenominate `test_usd` to have a 1e6 display exponent instead of 1e18
- [ ] Comment out `test_eth` from the asset registry, and comment out anywhere we currently use it
- [ ] Regenerate allocations with the redenominated `test_usd` and without `test_eth`.

Once we ship the testnet, we can work through any precision issues and remove the temporary restrictions, then ship a later testnet release with support for 1e18-denominated assets in the DEX.

erwanor:
Closes #2560, the diff is large because we're regenerating testnet allocations but each individual commit should consist of a relevant subtask.


hdevalence:
Could we leave the existing check as-is, and have a dedicated code path for the temporary restrictions? Then when it's time to remove, we can just delete the temporary check in isolation — in the course of fixing the underlying issue we might determine, for instance, that we needed some restriction on the bitsize, and it'd be cleaner if we could write it separately from the checks we intend to remove.

testnets/new-testnet.sh
```diff
@@ -2,18 +2,23 @@
 
 set -euo pipefail
 
-PREVIOUS_TESTNET_DIRECTORY=$(find . -mindepth 1 -type d | tail -n1)
+echo "#####################################################"
+PREVIOUS_TESTNET_DIRECTORY=$(find . -mindepth 1 -type d | sort | tail -n1)
```
erwanor, line 6: The change is here, I've added a `sort`.

Issue #2544, opened 2023-05-11
SwapExecutions need to be directed


hdevalence:
**Describe the bug**

Currently we make one execution trace for both directions of a swap; instead, there should be two: https://github.com/penumbra-zone/penumbra/pull/2538#discussion_r1190755104

hdevalence:
Along the way, we could also change the traces to be at the "frontier" level rather than at the route level, by recording each `FrontierTx`'s trace as a separate trace entry.

PR #2570, opened 2023-05-15
dex: make SwapExecutions directed

hdevalence:
I'm not sure we should land this for testnet 52; it was more complicated to make these changes and it feels like the code is doing the wrong thing.  It's not really clear to me why the execution traces should be stored as ambient global state in the object store, rather than just being returned by the methods that do execution. But that refactoring isn't something we should stuff in before the deadline.


hdevalence:
Closes #2544

PR #2572, opened 2023-05-15
dex: add additional RPC methods

hdevalence:
These don't have an implementation yet; the first commit just defines a spec for some methods that seem useful.

### DIFF
--- a/pd/src/info/specific.rs
+++ b/pd/src/info/specific.rs
@@ -1,5 +1,6 @@
 use std::pin::Pin;
 
+use async_stream::try_stream;
 use futures::StreamExt;
 use futures::TryStreamExt;
 use penumbra_app::dex::PositionRead;
@@ -8,6 +9,7 @@ use penumbra_app::governance::StateReadExt as _;
 use penumbra_chain::component::AppHashRead;
 use penumbra_chain::component::StateReadExt as _;
 use penumbra_crypto::asset::{self, Asset};
+use penumbra_crypto::dex::execution::SwapExecution;
 use penumbra_crypto::dex::lp::position;
 use penumbra_crypto::dex::lp::position::Position;
 use penumbra_crypto::dex::DirectedTradingPair;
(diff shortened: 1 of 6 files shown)


proto/proto/penumbra/client/v1alpha1/client.proto
```diff
@@ -111,11 +111,19 @@ service SpecificQueryService {
...
   rpc BatchSwapOutputData(BatchSwapOutputDataRequest) returns (BatchSwapOutputDataResponse);
   rpc SwapExecution(SwapExecutionRequest) returns (SwapExecutionResponse);
+  rpc ArbExecution(ArbExecutionRequest) returns (ArbExecutionResponse);
+
+  rpc SwapExecutions(SwapExecutionsRequest) returns (stream SwapExecutionsResponse);
+  rpc ArbExecutions(ArbExecutionsResponse) returns (stream ArbExecutionsResponse);
+
   rpc LiquidityPositions(LiquidityPositionsRequest) returns (stream LiquidityPositionsResponse);
   rpc LiquidityPositionById(LiquidityPositionByIdRequest) returns (LiquidityPositionByIdResponse);
+  rpc LiquidityPositionsById(LiquidityPositionsByIdRequest) returns (stream LiquidityPositionsByIdResponse);
```
hdevalence, line 124: The idea is that this would be useful for displaying the current state of a user's set of positions -- they can send a request with all of the position IDs they care about and get those specific positions back.

pd/src/info/specific.rs
```diff
@@ -429,6 +495,80 @@ impl SpecificQueryService for Info {
...
+                        ))));
+                    }
+
+                    let (key, swap_execution) = i.unwrap();
+                    let parts = key.split('/').collect::<Vec<_>>();
+                    let height = parts[2].parse().expect("height is not a number");
+                    let asset_1: asset::Id =
+                        parts[3].parse().expect("asset id formatted improperly");
+                    let asset_2: asset::Id =
+                        parts[4].parse().expect("asset id formatted improperly");
```
hdevalence, line 539: After #2606 we could pull these out of the `SwapExecution` directly rather than parsing keys

pd/src/info/specific.rs
```diff
@@ -429,6 +495,80 @@ impl SpecificQueryService for Info {
...
+            .await
+            .map_err(|e| tonic::Status::unknown(format!("chain_id not OK: {e}")))?;
+        let request_inner = request.into_inner();
+        let start_height = request_inner.start_height;
+        let end_height = request_inner.end_height;
+        let trading_pair = request_inner.trading_pair;
+
+        // Convert to domain type ahead of time if necessary
+        let trading_pair: Option<TradingPair> = if let Some(trading_pair) = trading_pair {
+            Some(trading_pair.try_into().expect("invalid trading pair"))
```
hdevalence, line 515: I think we don't want to `expect` here, since then someone can crash the node by submitting a request with an invalid trading pair.

hdevalence:
Merged optimistically to allow some crate splitting over the weekend.

PR #2581, opened 2023-05-16
spec: restructure protocol specification

hdevalence:
This commit starts (but does not finish) rearranging the existing protocol specification by functional section.  In the future, this should align more closely with crate boundaries.

I didn't move files around, so they're not fully organized yet -- but we can align them with the structure in the table of contents as we edit them.


hdevalence:
Resulting TOC:

Screen Shot 2023-05-15 at 11 26 40 PM
Screen Shot 2023-05-15 at 11 26 55 PM

PR #2582, opened 2023-05-16
Fix `anyhow::Error` type bounds to use `From` instead of `Into`

kevinji:
As documented in [this Stack Overflow post](https://stackoverflow.com/questions/72626885/using-tryinto-generics-with-anyhowerror), the correct type bound to use is `From` instead of `Into` because `?` desugars to `.from()` instead of `.into()`. This also removes the need for the explicit `Send + Sync + 'static` bounds.


hdevalence:
Nice! Thanks for the PR!

PR #2593, opened 2023-05-17
view: add support for querying owned position IDs

### LINKED ISSUE #2531: Add support for storing and querying owned positions to the view service
The implementation should support queries in the form of `select {position ID} where state is {desired state} and trading pair is {trading pair} from {set of owned positions}`.

aubrika:
Closes #2531


aubrika:
Closed in favor of #2637

PR #2602, opened 2023-05-18
fix(pd): look up listenaddr from tm rpc

conorsch:
When joining a testnet, `pd` will inspect the RPC endpoint for the bootstrap node and copy its known peers into the generated Tendermint config. To date, since #1847, that logic has also interpolated an incorrect address for the bootstrap node (by default currently `testnet.penumbra.zone:26657`): the RPC address shouldn't be used, the P2P address should be. This is particularly important when as we move toward updating our configs to default to HTTPS: we can no longer assume that the host portion of a service URL will be the same between the RPC and P2P endpoints. Instead, let's use the RPC to inspect what the correct ListenAddress is.

## Testing and review

First, verify the misbehavior on main branch:

```
git checkout main
cargo run --release --bin pd -- testnet unsafe-reset-all
cargo run --release --bin pd -- testnet join https://rpc.testnet.penumbra.zone
grep ^seed ~/.penumbra/testnet_data/node0/tendermint/config/config.toml
```

Observe that the first host in there is `rpc.testnet.penumbra.zone:26656`, which won't work: 26656 isn't even open on that IP. Next, try it on this feature branch:

```
git checkout smarter-seed-addr-lookup
cargo run --release --bin pd -- testnet unsafe-reset-all
cargo run --release --bin pd -- testnet join https://rpc.testnet.penumbra.zone
grep ^seed ~/.penumbra/testnet_data/node0/tendermint/config/config.toml
```

Observe that the first host in there is no longer a DNS name, but an IP address. Critically, the IP address is not the same as that for the RPC service:

```
❯ host rpc.testnet.penumbra.zone
rpc.testnet.penumbra.zone has address 34.111.241.130
```

Instead, the first node in the TM config should be `35.226.255.25:26656`, which maps to the exposed p2p port for the tendermint service on fn-0 in the testnet deployment.

### DIFF
--- a/deployments/scripts/rust-docs
+++ b/deployments/scripts/rust-docs
@@ -24,6 +24,7 @@ export RUSTDOCFLAGS="--enable-index-page -Zunstable-options --cfg docsrs"
 # shellcheck disable=SC2086
 cargo +nightly doc --no-deps \
   -p tendermint \
+  -p tendermint-config \
   -p tower-abci \
   -p jmt@0.3.0 \
   -p ark-ff \
--- a/pd/src/testnet.rs
+++ b/pd/src/testnet.rs
@@ -56,7 +56,6 @@ pub fn generate_tm_config(
         tm_config.p2p.laddr =
             parse_tm_address(None, &Url::parse(format!("tcp://{}", p2p).as_str())?)?;
     }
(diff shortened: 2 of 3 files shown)


zbuc (approved):
LGTM!

PR #2619, opened 2023-05-22
build: ditch proto-compiler, use buf

conorsch:
Overhauls the protobuf codegen by standardizing on `buf`, along with associated tooling like `protoc-gen-prost` and `proto-gen-tonic`. The output dirs for the supported langs are now:
    
* Rust: `./crates/proto/src/gen/`
* Go: `./proto/go/gen/`
    
We've removed the proto-compiler tool from the repo. Also, we no longer vendor upstream protos by committing directly to our repo; instead, we declare dependencies in the buf yaml files and offload the resolution to `buf generate`.
    
One major regression is that gRPC server reflection no longer works. These changes remove the `proto_descriptor.bin` build.rs logic, with no replacement in the new buf logic. As of right now, preview supports reflection:
    
        ❯ grpcurl grpc.testnet-preview.penumbra.zone:443 list
        cosmos.base.tendermint.v1beta1.Service
        grpc.reflection.v1alpha.ServerReflection
        penumbra.client.v1alpha1.ObliviousQueryService
        penumbra.client.v1alpha1.SpecificQueryService
        penumbra.client.v1alpha1.TendermintProxyService
        penumbra.custody.v1alpha1.CustodyProtocolService
        penumbra.narsil.ledger.v1alpha1.LedgerService
        penumbra.view.v1alpha1.ViewAuthService
        penumbra.view.v1alpha1.ViewProtocolService
    
but this code does not:
    
        ❯ grpcurl -plaintext localhost:8080 list
        Failed to list services: server does not support the reflection API
    
We'll circle back on reflection support.
    
Notably the tonic rpc defs are now in their own files, a side-effect of using the `prost-gen-tonic` plugin for buf. The original proto files, without the `.tonic.proto` suffix, now automatically `include!` their tonic neighbors, so the interface hasn't changed.
    
Refs #2184.


hdevalence:
@conorsch Question in re: https://github.com/penumbra-zone/penumbra/issues/2184#issuecomment-1560274681 : is there any other regression _other_ than the reflection data not being updated? If not, I think it could be good to merge this now, so that we don't generate conflicts with other changes to `.proto` files, and loop back on the reflection issue.

conorsch:
So far, so good. I'll get this cleaned up and RFR today, then circle back on the reflection!

conorsch:
The CI check for breaking changes in buf [is failing](https://github.com/penumbra-zone/penumbra/actions/runs/5071291239/jobs/9107492623?pr=2619), reporting a single change to the TM types, due to the switch from vendoring to dynamic dependency resolution. Running locally, I see:

```
❯ buf breaking --against 'https://github.com/penumbra-zone/penumbra.git#branch=main,subdir=crates/proto' proto/
tendermint/types/types.proto:1:1:Previously present enum "BlockIDFlag" was deleted from file.
```

Indeed, on `main` the vendored proto had a slightly different case to its type:

```
❯ rg -i 'enum BlockIDFlag'
crates/proto/rust-vendored/tendermint/types/types.proto
13:enum BlockIDFlag {

crates/proto/src/gen/tendermint.types.rs
228:pub enum BlockIdFlag {
```

Whereas on this branch, there's only the one casing:

```
❯ rg -i 'enum BlockIDFlag'
crates/proto/src/gen/tendermint.types.rs
228:pub enum BlockIdFlag {
```

Our code compiles fine, so I think we can disregard this as a blocker.

conorsch:
Regarding testing to ensure we don't have regressions (other than disabling gRPC server reflection support), I compared the output from `pcli q tx <id>`, built from this branch and from a precompiled binary for `v0.52.2`. No differences.

.github/workflows/buf-pull-request.yml
```diff
@@ -33,7 +33,7 @@ jobs:
       - name: Check for breaking changes
         uses: bufbuild/buf-breaking-action@v1
         with:
-          input: 'crates/proto'
+          input: 'proto'
           # The 'main' branch of the GitHub repository that defines the module.
           against: 'https://github.com/${GITHUB_REPOSITORY}.git#branch=main,subdir=crates/proto'
```
conorsch, line 38: N.B. the `against` line should be updated post-merge, but is accurate as of right now, showing us potentially breaking changes in the CI output of this PR.

crates/bin/pclientd/src/lib.rs
```diff
@@ -247,14 +247,6 @@ impl Opt {
...
                     .add_service(tonic_web::enable(specific_query_proxy))
                     .add_service(tonic_web::enable(tendermint_proxy_proxy))
-                    .add_service(tonic_web::enable(
-                        tonic_reflection::server::Builder::configure()
-                            .register_encoded_file_descriptor_set(
-                                penumbra_proto::FILE_DESCRIPTOR_SET,
-                            )
-                            .build()
-                            .with_context(|| "could not configure grpc reflection service")?,
-                    ))
```
hdevalence, line 257: Let's make a follow-up issue to restore grpc reflection, and copy this code snippet into the issue description, so that we don't lose track of it once it's deleted.

crates/proto/src/gen/penumbra.core.transparent_proofs.v1alpha1.rs
```diff
@@ -1,25 +1,121 @@
...
-    #[prost(message, optional, tag = "21")]
+    #[prost(message, optional, tag="21")]
     pub lambda_2_i: ::core::option::Option<super::super::crypto::v1alpha1::Amount>,
 }
+/// Encoded file descriptor set for the `penumbra.core.transparent_proofs.v1alpha1` package
+pub const FILE_DESCRIPTOR_SET: &[u8] = &[
+    0x0a, 0xe1, 0x0b, 0x0a, 0x42, 0x70, 0x65, 0x6e, 0x75, 0x6d, 0x62, 0x72, 0x61, 0x2f, 0x63, 0x6f,
+    0x72, 0x65, 0x2f, 0x74, 0x72, 0x61, 0x6e, 0x73, 0x70, 0x61, 0x72, 0x65, 0x6e, 0x74, 0x5f, 0x70,
+    0x72, 0x6f, 0x6f, 0x66, 0x73, 0x2f, 0x76, 0x31, 0x61, 0x6c, 0x70, 0x68, 0x61, 0x31, 0x2f, 0x74,
+    0x72, 0x61, 0x6e, 0x73, 0x70, 0x61, 0x72, 0x65, 0x6e, 0x74, 0x5f, 0x70, 0x72, 0x6f, 0x6f, 0x66,
```
hdevalence, line 28: 👀 Looks like the new setup is putting the file descriptor set in the output directly?

hdevalence (approved):

PR #2620, opened 2023-05-23
compactblock: store in non-consensus store

hdevalence:
We don't ever use the compact block inside the chain state.  Strictly speaking, it's not necessary to store it inside the verifiable part of the chain state, and keeping it outside allows it to be accessed more efficiently when serving clients (it can just be read directly from RocksDB, in a linear order...)

We don't currently verify the CB data with a light client anyways, but even if we wanted to, we probably wouldn't want to verify every single CB's merkle inclusion proof separately. One idea we discussed informally some time ago is to write the hash of the current CB into the state (to allow individual verification) and also to write a hash chain accumulator that hashes in each successive CB, so a client can stream many CBs all at once, hash them together in sequence, and then verify the accumulator hash of all-downloaded-CBs using a Merkle proof.

This commit does none of these options, leaving them for later work, and just makes sure that the CB data is stored in ordinary RocksDB in linear order.

### DIFF
--- a/crates/core/component/compact-block/src/component/view.rs
+++ b/crates/core/component/compact-block/src/component/view.rs
@@ -1,16 +1,19 @@
 use anyhow::Result;
 use async_trait::async_trait;
-use penumbra_proto::{StateReadProto, StateWriteProto};
+use penumbra_proto::DomainType;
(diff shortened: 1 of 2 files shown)


hdevalence:
Nope, it'd be good to file them though.

PR #2622, opened 2023-05-23
ci: debug output during failed wasm checks

conorsch:
We want to see the output from `cargo tree` inspecting the dep graph of `mio` on any failure of the wasm workflow, since that's one of the first actions a dev would take in debugging.


conorsch:
This failed CI run shows that the `cargo tree` command ran at the end, despite the intentional failure earlier in the workflow: https://github.com/penumbra-zone/penumbra/actions/runs/5060018873/jobs/9082358374?pr=2622

PR #2628, opened 2023-05-24
storage: integration work for `jmt@0.6`

### LINKED ISSUE #2391: Use updated ICS23 proof spec supporting noninclusion proofs

Now that we've been able to merge our extensions to ICS23 upstream, we should ensure that we ship a version of Penumbra that uses the correct ICS23 proof spec.  This is important to do as part of the next testnet release, because otherwise, we'll have `testnet` and `testnet-preview` using different proof specs, and the relayer doesn't know how to support that (and shouldn't need to, because there should just be one correct proof spec).

Getting this done has a somewhat deep dependency chain, so we should focus to make sure we get all of the steps done in time:

- [x] Ensure that the published `jmt` crate uses `ics23 = "0.10"` and has the correct proof spec
- [ ] Use the updated `jmt` crate, partially done in #2032 ...
  - but blocked on #2374, which is impossible ...
  - so we need the first part of #2379 instead:
    - [x] Finish and merge  https://github.com/penumbra-zone/penumbra/pull/2392
    - [x] #2380
    - [x] https://github.com/penumbra-zone/ibc-types/issues/2
    - [x] #2381
    - [x] https://github.com/penumbra-zone/ibc-types/issues/3

erwanor:
Closes #2391 #2030


crates/core/component/chain/src/component/app_hash.rs
```diff
@@ -106,36 +110,43 @@ impl AppHashRead for Snapshot {
...
-                proofs: vec![jmt_commitment_proof, root_proof],
+                proofs: vec![membership_proof, root_proof],
             },
         ))
     }
 
+    /// Returns the value and a proof of inclusion up to the current app hash,
+    /// using Tendermint's proof format. Returns `None` if the key is not present.
+    ///
+    /// Review: it looks like the Tendermint proof format does not support
```
erwanor, line 123: Reviewer scrutiny appreciated here, I'm not sure this is true (re: tendermint merkle proof format) - or that we want to reply to ABCI queries with nonexistence proofs regardless. @avahowell

PR #2638, opened 2023-05-25
build: re-enable prost-serde buf plugin

conorsch:
Adds an invocation of the `protoc-gen-prost-serde` plugin to the buf codegen config. This plugin was mistakenly omitted from the recent proto-compiler -> buf-generate rewrite, but many of the emitted serde files were still present. This change regenerates the files.

Unfortunately, the tendermint types now lack deserialization impls. Ideally we'd be able to add `buf.build/tendermint/tendermint` as a dependency of our buf module, at `proto/penumbra/buf.yaml`, but doing so results in an error:

    ibc/core/commitment/v1/commitment.proto:7:8:gogoproto/gogo.proto
    exists in multiple locations: gogoproto/gogo.proto gogoproto/gogo.proto

That's due to the fact that both the ibc and tendermint BSR modules depend on gogoproto. Unsure how to resolve this conflict without resorting to vendoring.


conorsch:
Recent proto changes introduced conflicts. Will continue to iterate on this logic locally.

PR #2646, opened 2023-05-26
Clean up SwapExecution handling

### LINKED ISSUE #2606: Clean up `SwapExecution` handling

The current implementation of `SwapExecution` has a few issues:

- it uses the object store to store data in a global variable, even though the data is scoped to the current execution;
- it aggregates trades along each sub-route, making it impossible to see fine-grained price information;
- we make one `SwapExecution` for both directions of a trade, even though there are actually two executions.

I made an attempt to clean up part of this in #2570, but decided not to merge the code in advance of the initial DEX release.

To fix these issues, we should clean up the `SwapExecution` handling:

- [x] Change the `SwapExecution` proto and domain types to have `input: Value`, `output: Value` fields recording the total input and output values.
- [x] Change the `FillRoute` and `RouteAndFill` traits to return a `Result<SwapExecution>`, passing the traces up the call stack rather than storing them in a global variable. Note that this will change the semantics of the return value: rather than returning the _unfilled_ input, and the output as an `(Amount, Amount)`, they'll return the filled input and output in the `SwapExecution`; the caller will need to compute how much was unfilled.
- [x] Store one `SwapExecution` for each direction of the trade, along the lines of the code in #2570 (but simpler, since there's no object store involved)
- [x] Update the `FillRoute` code to record "precise" traces, along the lines of the code in #2570.
- [x] Remove any remaining references to the `trade_traces` state key that was used to access the global variable.

**Additional context**

This unblocks #2580, by making the traces precise (allowing them to be used for introspection of market depth) and by providing a convenient object to return with the results of the execution.

Supersedes #2544

zbuc:
closes #2606

### DIFF
--- a/crates/core/component/dex/src/component/router/fill_route.rs
+++ b/crates/core/component/dex/src/component/router/fill_route.rs
@@ -604,7 +604,7 @@ impl<S: StateRead + StateWrite> Frontier<S> {
         tracing::debug!("filling backward along frontier");
         let mut current_value = output;
         for i in (0..=start_index).rev() {
-            tx.trace.push(Some(current_value.amount));
+            tx.trace.insert(0, Some(current_value.amount));
 
             let (new_reserves, prev_input) = self.positions[i]
                 .phi
@@ -627,6 +627,6 @@ impl<S: StateRead + StateWrite> Frontier<S> {
             current_value = prev_input;
         }
 
-        tx.trace[0] = Some(current_value.amount);
+        tx.trace.insert(0, Some(current_value.amount));
     }
 }
--- a/crates/core/component/dex/src/component/router/fill_route.rs
+++ b/crates/core/component/dex/src/component/router/fill_route.rs
@@ -48,8 +48,10 @@ pub trait FillRoute: StateWrite + Sized {
         input: Value,
         hops: &[asset::Id],
         spill_price: Option<U128x128>,
-    ) -> Result<Vec<Vec<Value>>> {
-        fill_route_inner(self, input, hops, spill_price, true).await
+    ) -> Result<SwapExecution> {
+        Ok(SwapExecution::new(
+            fill_route_inner(self, input, hops, spill_price, true).await?,
+        ))
     }
 
     /// Like `fill_route`, but with exact spill price checks at the cost of
--- a/crates/core/component/dex/src/component/router/fill_route.rs
+++ b/crates/core/component/dex/src/component/router/fill_route.rs
@@ -49,9 +49,7 @@ pub trait FillRoute: StateWrite + Sized {
         hops: &[asset::Id],
         spill_price: Option<U128x128>,
     ) -> Result<SwapExecution> {
-        Ok(SwapExecution::new(
-            fill_route_inner(self, input, hops, spill_price, true).await?,
-        ))
+        fill_route_inner(self, input, hops, spill_price, true).await
     }
 
     /// Like `fill_route`, but with exact spill price checks at the cost of
@@ -85,7 +83,7 @@ async fn fill_route_inner<S: StateWrite + Sized>(
     hops: &[asset::Id],
     spill_price: Option<U128x128>,
     ensure_progress: bool,
-) -> Result<Vec<Vec<Value>>> {
+) -> Result<SwapExecution> {
     // Build a transaction for this execution, so if we error out at any
     // point we don't leave the state in an inconsistent state.  This is
     // particularly important for this method, because we lift position data
@@ -220,7 +218,7 @@ async fn fill_route_inner<S: StateWrite + Sized>(
     this.apply();
 
     // cleanup / finalization
-    Ok(trace)
+    Ok(SwapExecution::new(trace))
 }
 
 /// Breaksdown a route into a collection of `DirectedTradingPair`, this is mostly useful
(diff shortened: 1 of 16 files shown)


crates/core/component/dex/src/component/router/fill_route.rs
```diff
@@ -48,8 +48,10 @@ pub trait FillRoute: StateWrite + Sized {
         input: Value,
         hops: &[asset::Id],
         spill_price: Option<U128x128>,
-    ) -> Result<(Value, Value)> {
-        fill_route_inner(self, input, hops, spill_price, true).await
+    ) -> Result<SwapExecution> {
+        Ok(SwapExecution::new(
+            fill_route_inner(self, input, hops, spill_price, true).await?,
```
hdevalence, line 53: Shouldn't `fill_route_inner` already have computed all of the fields of the `SwapExecution`? In that case, could we just have `fill_route_inner` return a `SwapExecution` directly?

crates/core/component/dex/src/component/router/fill_route.rs
```diff
@@ -81,7 +85,7 @@ async fn fill_route_inner<S: StateWrite + Sized>(
     hops: &[asset::Id],
     spill_price: Option<U128x128>,
     ensure_progress: bool,
-) -> Result<(Value, Value)> {
+) -> Result<Vec<Vec<Value>>> {
```
hdevalence, line 88: I think this should return a `SwapExecution`, since it already computes `input` and `output`.

crates/core/component/dex/src/component/router/fill_route.rs
```diff
@@ -351,15 +350,15 @@ impl<S: StateRead + StateWrite> Frontier<S> {
...
+        let trace: Vec<Vec<Value>> = vec![std::iter::once(Value {
             amount: 0u64.into(),
             asset_id: pairs.first().unwrap().start,
         })
         .chain(pairs.iter().map(|pair| Value {
             amount: 0u64.into(),
             asset_id: pair.end,
         }))
-        .collect();
+        .collect()];
```
hdevalence, line 361: This will insert an all-zeros execution at the beginning, instead, it should be `let trace = Vec::new()` to represent an empty set of traces.

crates/core/component/dex/src/component/router/fill_route.rs
```diff
@@ -380,17 +379,28 @@ impl<S: StateRead + StateWrite> Frontier<S> {
     /// Apply the [`FrontierTx`] to the frontier, returning the input and output
     /// amounts it added to the trace.
     fn apply(&mut self, changes: FrontierTx) -> (Amount, Amount) {
-        self.trace[0].amount +=
-            changes.trace[0].expect("all trace amounts must be set when applying changes");
+        let mut trace: Vec<Value> = vec![];
+
+        trace[0] = Value {
+            amount: changes.trace[0].expect("all trace amounts must be set when applying changes"),
+            asset_id: self.trace[0][0].asset_id,
```
hdevalence, line 386: Rather than pulling the asset ID out of `self.trace[0]`, which may not be present, they can be pulled out of `self.pairs` (`.start` for `trace[0]` and `.end` for all other `trace[i]`).

crates/core/component/dex/src/component/router/route_and_fill.rs
```diff
@@ -55,10 +52,20 @@ pub trait HandleBatchSwaps: StateWrite + Sized {
...
+                traces: vec![],
+                input: Value {
+                    asset_id: trading_pair.asset_1(),
+                    amount: Amount::zero(),
+                },
+                output: Value {
+                    asset_id: trading_pair.asset_2(),
+                    amount: Amount::zero(),
+                },
+            }
```
hdevalence, line 65: Since we skipped execution, shouldn't we not write a `SwapExecution` into the state at all, rather than making an empty one? Otherwise a caller has to differentiate between "swap execution that is present and happened" and "swap execution that is present but did not happen" and "swap execution that is not present and did not happen".

crates/core/component/dex/src/component/router/route_and_fill.rs
```diff
@@ -70,7 +77,17 @@ pub trait HandleBatchSwaps: StateWrite + Sized {
...
+                traces: vec![],
+                input: Value {
+                    asset_id: trading_pair.asset_2(),
+                    amount: Amount::zero(),
+                },
+                output: Value {
+                    asset_id: trading_pair.asset_1(),
+                    amount: Amount::zero(),
+                },
+            }
```
hdevalence, line 90: ditto above

crates/core/component/dex/src/component/router/route_and_fill.rs
```diff
@@ -79,34 +96,25 @@ pub trait HandleBatchSwaps: StateWrite + Sized {
...
-                SwapExecution {
-                    traces: trade_traces.into_iter().collect(),
-                },
-            );
-
-        // Clean up the swap execution object store now that it's been persisted.
-        Arc::get_mut(self)
-            .expect("expected state to have no other refs")
-            .object_delete("trade_traces");
+            .set_output_data(output_data, swap_execution_1_for_2, swap_execution_2_for_1);
```
hdevalence, line 117: Probably we want to decouple setting the BSOD from the `SwapExecution`s, so that we can write the `SwapExecution` only in the branch where we actually execute it.

PR #2655, opened 2023-06-01
view: add RateData field to TransactionPlannerRequest.Delegate

aubrika:


crates/proto/proto/penumbra/view/v1alpha1/view.proto
```diff
@@ -128,6 +129,7 @@ message TransactionPlannerRequest {
   message Delegate {
     core.crypto.v1alpha1.Amount amount = 1;
     core.crypto.v1alpha1.IdentityKey identity_key = 2;
+    core.stake.v1alpha1.RateData rate_data = 3;
```
hdevalence, line 132: Since the `RateData` contains an `IdentityKey`, we should omit the standalone `IdentityKey`:
```suggestion
    core.stake.v1alpha1.RateData rate_data = 3;
```

hdevalence (approved):
LGTM!

PR #2686, opened 2023-06-07
dex(router): track state before checking for termination

erwanor:
This PR makes sure that all the execution state is tracked _prior_ to checking termination conditions, so that interrupting a route and fill does not lead to inconsistencies between the reported state and the actual positions.

### DIFF
--- a/crates/core/component/dex/src/swap_execution.rs
+++ b/crates/core/component/dex/src/swap_execution.rs
@@ -13,21 +13,21 @@ pub struct SwapExecution {
 }
 
 impl SwapExecution {
+    /// Returns the price of the latest execution trace.
     pub fn max_price(&self) -> Result<Option<U128x128>> {
-        let Some(aggregate_input )= self.aggregate_input() else {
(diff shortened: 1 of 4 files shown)


crates/core/component/dex/src/swap_execution.rs
```diff
@@ -13,21 +13,21 @@ pub struct SwapExecution {
...
-        };
-
-        let price = U128x128::ratio(aggregate_input, aggregate_output)?;
-
+        let price = U128x128::ratio(input.amount, output.amount)?;
         Ok(Some(price))
     }
 
-    fn aggregate_input(&self) -> Option<Amount> {
+    pub fn aggregate_input(&self) -> Option<Amount> {
```
hdevalence, line 30: This is just the input field already recorded in the swap execution though right? So we could delete these two methods entirely?

erwanor: yup good catch , i was testing something, ripping them out entirely. fyi, the last few commits aren't showing up yet (github incident).

PR #2698, opened 2023-06-08
deps: temporarily remove ibc-types2

hdevalence:
Adding it to the repo caused breakage, because the existing `ibc-types` dep pins dependencies, and those pins conflict with `ibc-types2`.  The right fix is to remove the pins, but that is a slightly bigger fix than this temporary one.


conorsch:
Tested a local build of galileo against this branch, and it works, so this resolves https://github.com/penumbra-zone/galileo/issues/49

PR #2713, opened 2023-06-14
Add display_confirm boolean to AddressByIndexRequest and EphemeralAddressRequest

### LINKED ISSUE #2711: Add method to view/custody service RPC that computes and optionally displays an address

We'd like for a custodian to be able to display an address to the user, so that the display happens along a trusted codepath.  As an example use case, we want a frontend site to be able to ask the extension to display a user's address in an extension popup, rather than on the untrusted frontend.

We already have almost exactly the desired RPC method:
```proto
  // Query for an address given an address index
  rpc AddressByIndex(AddressByIndexRequest) returns (AddressByIndexResponse);
  // Query for an ephemeral address
  rpc EphemeralAddress(EphemeralAddressRequest) returns (EphemeralAddressResponse);
```
with
```proto
message AddressByIndexRequest {
  core.crypto.v1alpha1.AddressIndex address_index = 1;
}
message EphemeralAddressRequest {
  core.crypto.v1alpha1.AddressIndex address_index = 1;
}
```

To add the desired functionality, we could add optional fields to these requests:
```proto
message AddressByIndexRequest {
  core.crypto.v1alpha1.AddressIndex address_index = 1;
  bool display_confirm = 2;
}
message EphemeralAddressRequest {
  core.crypto.v1alpha1.AddressIndex address_index = 1;
  bool display_confirm = 2;
}
```

The Rust view server, which has no UI, can ignore these fields, but the extension could check whether the field is set and display a popup.

aubrika:
Closes #2711


crates/view/src/client.rs
```diff
@@ -763,6 +763,7 @@ where
         async move {
             let address = self2.address_by_index(tonic::Request::new(pb::AddressByIndexRequest {
                 address_index: Some(address_index.into()),
+                display_confirm: false,
```
hdevalence, line 766: Not particularly important here, but I think it's slightly cleaner to do
```
..Default::default()
```
instead; the result is the same but the syntax signals that we don't care about the `display_confirm` value rather than that we explicitly want it to be false for some reason.

hdevalence (approved):

hdevalence (approved):
LGTM; @aubrika can you regenerate the protos?

aubrika:
> LGTM; @aubrika can you regenerate the protos?

Running the script indicated by @conorsch generated changes to `view.pb.go` but still none to `proto_descriptor.bin`.

PR #2723, opened 2023-06-19
update `create_span` to allow Prepare/ProcessProposal

noot:
we're using the `tower-trace` crate in astria, and using it with tendermint v0.37, so we needed this to not have panics.


hdevalence:
@conorsch looks like something may have changed with the protobuf workflow, it's now reporting a dirty state because the `.bin` file was changed, whereas previously we excluded changes only to that file.

hdevalence (approved):

PR #2728, opened 2023-06-21
Implement staking/slashing tests

### LINKED ISSUE #2630: Delegations and undelegations should use exchange rate from current epoch start
Currently, we compute the next epoch's `BaseRateData` in the preceding epoch's `end_epoch`, and (un)delegations use the exchange rate of the upcoming epoch to compute how many delegation tokens a certain amount of staking tokens should be converted into upon delegation, or how many unbonding tokens (1:1 for staking tokens) a certain amount of delegation tokens should be converted into upon undelegation. This works because we are currently able to predict the exchange rate for the next epoch ahead of time, because the change in `BaseRateData` is a constant multiplier each epoch.

However, with time-based epochs, we should like to have a chain-controlled effective staking token issuance (via changes to `BaseRateData` and thereby changes to the exchange rate of active validators, plus payouts to funding streams) that is determined by a _time-based rate of issuance_. That is, shorter epochs should issue proportionately fewer staking tokens. If the change to `BaseRateData` is fixed one epoch ahead of time, as it is presently (and as it must be when delegations and undelegations use the exchange rates for the upcoming epoch boundary—they must bind this information, so it must be known), then an unexpected early epoch end will result in more issuance per unit time for that epoch than desired.

In the design meeting, we outlined several possible solutions to this problem:

1. Make a new family of denominations dual to unbonding tokens called "bonding tokens" which are also 1:1 for the staking token. This allows late binding of the exchange rate until after the rate is known in the future. An explicit `DelegateClaim` action would be needed to perform the conversion after the rate is known.
2. Track the difference between the previous epoch's actual duration in clock time and the desired epoch length in clock time. If the previous epoch ran short, subtract a proportional amount from the next epoch's issuance; if it ran long, add a proportional amount to the next epoch's issuance. This "fudge factor" over time would average to the correct issuance.
3. Have both delegations and undelegations use the `RateData` from the current epoch rather than the next epoch. This information is available at the time the action is performed, so it can be early-bound, and the exactly correct `BaseRateData` for an epoch can be computed at its own `end_epoch`, based on its actual length.

We have decided to go with (3), because of shortcomings in (1) and (2). In (1), making a new family of denominations complicates wallet logic, and requires active user participation to claim delegation tokens. If the user does not do this, temporal correlation between `DelegateClaim` and `Undelegate` (when the user goes to undelegate) lead to privacy degradation by probabilistically linking the delegation and its corresponding undelegation. Flow encryption would mitigate this to some degree. In (2), it is possible that there exist economic attacks which could be performed by malicious active validators against the staking rate. For example, two active validators could collude to alternate enabling and disabling their validators, which would allow them to trigger epoch transitions arbitrarily. This could disrupt normal issuance of staking rewards indefinitely, creating a variety of undesirable price manipulation opportunities. By contrast, (3) requires no extra mechanism to track another intermediate state for delegations, and calculates rates exactly, because it knows all information needed when it is needed.

To do this, the `check_stateful` for `Delegate` and `Undelegate` should use the _current_ validator rate data, not the _next_ validator rate data, as is presently done. Client support will also need to be updated, because we'll want to fetch the current, not next, rates to properly prepare these actions.

In order for this to work, it's necessary to ensure that the slashing risk period for undelegation claims covers slashing that occurs in the same epoch as the undelegation. This appears to already be the case, but more thought should be put to it: https://github.com/penumbra-zone/penumbra/blob/e75ea5f6b5fbcaea0a114c02b86e0fe7a3342d07/crates/core/component/stake/src/action_handler/undelegate_claim.rs#L30-L57

When this is done, the tracking of next validator rates can and should be entirely removed.

This is intentionally separated from the work necessary to implement the distributions component and time-based rates, which should be tracked separately and is blocked on completion of this ticket.

### LINKED ISSUE #2754: Implement integration tests for staking and slashing mechanics
Currently no integration tests exist for the staking and slashing mechanics, resulting in a low degree of assurance when we are making changes and requiring more manual testing.

While implementing https://github.com/penumbra-zone/penumbra/issues/2630 this shortcoming was noted, since the slashing mechanics are somewhat complicated it would be great to get some tests in place for the expected behavior.

The tests should use the existing `MockClient`-style integration testing.

zbuc:
WIP Closes #2754

### DIFF
--- a/crates/core/app/src/mock_client.rs
+++ b/crates/core/app/src/mock_client.rs
@@ -47,8 +47,8 @@ impl MockClient {
             anyhow::ensure!(
                 root == expected_root,
                 format!(
-                    "client sct root should match chain state: {:?} != {:?}",
-                    root, expected_root
+                    "client sct root should match chain state: {:?} != {:?} (height: {})",
+                    root, expected_root, height
                 )
             );
         }
--- a/crates/core/app/src/mock_client.rs
+++ b/crates/core/app/src/mock_client.rs
@@ -133,6 +133,10 @@ impl MockClient {
         self.notes.get(commitment).cloned()
     }
 
+    pub fn notes(&self) -> &BTreeMap<note::Commitment, Note> {
+        &self.notes
+    }
+
     pub fn swap_by_commitment(&self, commitment: &note::Commitment) -> Option<SwapPlaintext> {
         self.swaps.get(commitment).cloned()
     }
--- a/crates/core/app/src/tests/mod.rs
+++ b/crates/core/app/src/tests/mod.rs
@@ -1,1 +1,2 @@
+mod staking_and_slashing;
 mod swap_and_swap_claim;
--- a/crates/core/app/src/tests/staking_and_slashing.rs
+++ b/crates/core/app/src/tests/staking_and_slashing.rs
@@ -43,7 +43,10 @@ async fn unbonding_slashing_penalties() -> anyhow::Result<()> {
     let mut height = 1;
 
     // 0. Simulate BeginBlock
+    // TODO: Would be nice to call `App::begin_block` but creating the `abci::request::BeginBlock` is tricky
+    // App::begin_block(&mut state_tx).await;
     let mut state_tx = state.try_begin_transaction().unwrap();
+    state_tx.put_block_height(height);
     state_tx.put_epoch_by_height(
         height,
         penumbra_chain::Epoch {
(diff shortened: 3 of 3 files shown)


erwanor:
@zbuc should we close this and re-try it when we have figured up our cometbft mocking story?

zbuc:
> @zbuc should we close this and re-try it when we have figured up our cometbft mocking story?

Yeah, it would be much neater to implement then

PR #2732, opened 2023-06-22
proto: allow specifying a base denom in `AssetId` protos

hdevalence:
This allows RPC clients to use a base denom string wherever they would encode an `AssetId`.


hdevalence:
I branched this off of `0.54.1` to allow backports, but it looks like that generated a merge conflict with the proto descriptors. Rebasing onto `main`.

PR #2733, opened 2023-06-23
pcli: import seed phrase from stdin

### LINKED ISSUE #2693: `pcli keys import phrase` should not pass the seed phrase on the command line
`pcli keys import phrase` takes the seed phrase as a command line argument rather than reading from stdin, so the user's seed phrase will end up in shell history.

`pcli` should read from stdin.

conorsch:
Changes the behavior for `pcli keys import phrase` to prompt for a masked password input interactively. When run via a pipe, reads from stdin instead, which preserves the ability to import seed phrases from other tools.

Closes #2693.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3933,6 +3933,7 @@ dependencies = [
  "ark-ff",
  "assert_cmd",
  "async-stream 0.2.1",
+ "atty",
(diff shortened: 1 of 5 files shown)


conorsch:
Oops, missed 55 window for this. It'll be in 56.

Issue #2736, opened 2023-06-23
Newer versions of protoc don't regenerate proto descriptor


conorsch:
As part of the proto codegen process, we use `protoc` via `tools/proto-compiler`. Our dev docs currently say:

>  After installing, run protoc --version and confirm you’re running at least 3.21.8 (or newer). Don’t install protoc from package managers such as apt, as those versions are often outdated, and will not work with Penumbra.

Current version of `protoc` is [v23.3](https://github.com/protocolbuffers/protobuf/releases/tag/v23.3). If that sounds astonishingly newer than v3.21, you are correct. Looks like the project jumped from v3 to v20 last year:

```
❯ curl -s https://api.github.com/repos/protocolbuffers/protobuf/releases | jq '.[] | .tag_name + " " + .published_at' -r
v23.3 2023-06-14T19:22:45Z
v23.2 2023-05-26T16:43:43Z
v23.1 2023-05-17T17:28:25Z
v22.5 2023-05-10T03:56:57Z
v23.0 2023-05-08T18:31:51Z
v23.0-rc3 2023-05-05T17:20:25Z
v22.4 2023-05-04T20:35:56Z
v23.0-rc2 2023-04-28T18:59:14Z
v23.0-rc1 2023-04-26T23:16:56Z
v22.3 2023-04-12T23:47:13Z
v22.2 2023-03-10T18:35:01Z
v22.1 2023-03-07T22:31:27Z
v22.0 2023-02-16T18:17:29Z
v22.0-rc3 2023-02-10T20:17:58Z
v22.0-rc2 2023-02-02T21:18:01Z
v22.0-rc1 2023-01-31T18:24:35Z
v21.12 2022-12-14T14:45:50Z
v21.11 2022-12-08T06:34:59Z
v21.10 2022-11-30T19:15:02Z
v21.9 2022-10-26T20:28:55Z
v21.8 2022-10-18T18:18:56Z
v21.7 2022-09-29T22:22:53Z
v3.20.3 2022-09-29T21:22:59Z
v3.19.6 2022-09-29T20:46:59Z
v3.16.3 2022-09-29T22:17:28Z
v21.6 2022-09-14T21:37:03Z
v3.20.2 2022-09-14T18:05:45Z
v3.19.5 2022-09-14T21:36:20Z
v3.18.3 2022-09-14T21:02:42Z
v21.5 2022-08-09T19:53:44Z
```

[...]

PR #2739, opened 2023-06-26
build: require protoc version 23.x

conorsch:
Our dev docs were badly out of date in terms of recommended version for `protoc` and related tooling. Installed the latest and greatest, updated the CI workflows, and clarified the docs. We now mandate a specific major version of the protoc tooling, so we can ensure that
    
1) protos have been regenerated in a CI check
2) binary artifacts like the descriptor file don't flap between
    dev workstations
    
There's a bit of friction introduced here, in that folks will need to update the tooling locally, but we've been ignoring that need to date and paying the cost in confusion tokens, so I view the new workful as a reasonable tradeoff.
    
Also regenerates all code emitted by the proto build process so we're up to spec with the latest.
    
Refs #2736.


aubrika (approved):
LGTM! Maybe we can announce the need for the tooling upgrade in a couple relevant dev channels so nobody is caught unaware?

Issue #2644, opened 2023-05-26
Restructure DEX computations to not be wide


erwanor:
Following #2643, and part of #2589, most DEX operations combine `U128x128` into ratios, we should tentatively take advantage of this fact to find a workaround widening operations.

erwanor:
Done in #2740

Issue #2589, opened 2023-05-16
Tracking issue: analyze and improve our fixed-point arithmetic usage 


erwanor:
`U128x128` is a specialized fixed-point arithmetic type used throughout the DEX, capable of maintaining 128 bits for the integral portion and another 128 bits for the decimal portion, which is handy to deal with denominations that have large unit amounts (for instance, 1 Ether is 1e18 wei).

In #2559, we discovered a silent overflow bug, and a hard limit on computing the constant invariants of trading functions ($pR_1 + qR_2 = k$ will overflow if $p, q, R_1, R_1 \geq 2^{64}$). As a result, we want to increase our assurance efforts by refactoring the DEX engine to cut the surface area where fixed point arithmetic is used, starting with `fill_route`, `route_and_fill`), perform bound analysis at each of the call sites, and restructure DEX computations to not be "wider" at each step, taking advantage of the fact that a lot of compute takes the form of dividing large integers together.

Finally, we should develop a strategy to ensure that we have a robust analysis of whether the DEX can encounter liveness issues when stretching the limits of our arithmetic system, and develop a mitigation strategy if necessary. 

- [x] #2565 
- [x] Refactor away `fill_against` in favor of `fill_route`
- [x] Inspect the DEX call hierarchy and remove vestigial/unused methods
- [x] #2643
- [x] #2644
- [x] #2677

erwanor:
We're going to change strategies with this story. Our first approach was to work backwards from a worst case (ETH-style denom units of $10^18$) to determine individual bounds on trading function coefficients, LP reserves and Swap amounts so that $\varphi$ calculations (transforming input into output) would be infaillible.

Unfortunately, we found that such bounds would be too limiting, especially in the case of cross denom interactions (e.g. an ETH denominated NFT). So instead, we are shifting gear towards a dynamic bound checking approach which focuses on 1/ having a clear path to recover from overflow failures, this is particularly important in the frontier execution logic 2/ isolate failures from interfering with other benign positions, this is more difficult because batched execution makes it impossible to attribute failure to a specific interaction.

We will still implement #2644, in particular, making sure that we first compute the trading function effective price hoping that in most cases the denom scaling factors cancels out. This means that the overall sequence of dependencies that this ticket tracks is a follow:

1. Make DEX engine computations handle overflow gracefully
2. #2644 
3. Switch numeric bounds to $2^{80}$ (experimentally), unblocking #2677 
4. and finally #2643

PR #2740, opened 2023-06-26
dex: faillible dex execution

erwanor:
Part of #2589, this PR:
- [x] makes internal dex operations faillible i.e. catching overflow failures
- [x] specifies and implement gracefully recovery from those errors
- [x] #2644


erwanor:
Merging this optimistically, along with #2805

PR #2741, opened 2023-06-26
Replace REGISTRY::parse_unit calls with Cache::with_known_assets().get_unit when dealing with known assets

aubrika:
Some additional work to the end of replacing the asset registry - in the course of replacing `REGISTRY::parse_unit` some decisions needed to be made around using specific test assets know to the asset cache, so there were some changes needed to tests and some in particular could use some expert review.


aubrika:
Hey @erwanor , in the course of this work I broke one of your tests, and I'm not sure I fixed it right. :)

Can you take a look at the test assets being utilized in `router/tests.rs` for `path_search_commutative()` in this branch and tell me if we should instead change the unit exponents of any of them, instead of changing the path and spill prices? 🙏🏻 

My concern is that, while it is passing with my changes, I lack the domain expertise here to know if I've done something foolish by moving the goalposts along with the test assets in play.

erwanor (approved):
@aubrika thank you, that's much cleaner and using the correct denom strings for the `path_search_commutative` test! the target values are correct as well!

aubrika:
> @aubrika thank you, that's much cleaner and using the correct denom strings for the `path_search_commutative` test! the target values are correct as well!

Hmm, I think the target values are correct now, but I don't quite understand why they changed by a factor of 100? We will need to update the comment to the test to account for this, I think. Can you tell which asset swap (test_btc for btcumbra or penumbra for test_pen) would have caused that change in value? 🤔

erwanor:
This is because we're starting of `test_btc` with a denom unit amount of $10^6$ to swap into `penumbra` with a denom unit amount of $10^8$. Before you fixed the test, we were using the incorrect btc denom which defaulted to $10^8$! Now that we use the correct denom strings, the path price is multiplied by the cross-denom difference $10^2$

PR #2753, opened 2023-06-28
dex: implement limit orders

### LINKED ISSUE #2722: dex: enable user to specify "fill-once" liquidity positions

The current DEX engine only supports two-sided liquidity provision, which means that there is no way to specify a limit order selecting for a specific asset on a pair. Users interested in those types of order must remain online in order to close their position when it gets filled against.

Instead, a user should be able to open a limit order, that is a liquidity position which gets filled exactly-once. This could be done via a `fill_once` or `close_on_fill` boolean that is included in a `Position`. We do not need particular logic to handle partial fills because the arbitrage dynamics in the DEX engine cause positions at the tip to be either untouched or completely filled. After their position is executed and closed, the user will only need to withdraw their liquidity.

- [x] Handling `PositionClose`: now that the chain can close positions, we must change the handling of `PositionClose` to not assume that the client is necessarily caught up on the state of a position. 
- [x] Add a bool `close_on_fill` to `Position`
- [x] Modify `put_position` to inspect whether a given position is being completely filled, and if so update its state to `Closed`
- [ ] Add some unit-tests
- [ ] Add support for limit-orders in `pcli`

Originally posted in discord by @hdevalence

erwanor:
Closes #2722


crates/core/component/dex/src/lp/position.rs
```diff
@@ -29,6 +29,11 @@ pub struct Position {
     /// duplicate position [`Id`]s, so it can track position ownership with a
     /// sequence of stateful NFTs based on the [`Id`].
     pub nonce: [u8; 32],
+    /// Set to `true` if a position is a limit-order, meaning that it will be closed
+    /// after being filled against.
+    pub close_on_fill: bool,
+    /// Set to the `asset::Id` of the asset provided in a limit-order, `None` otherwise.
+    pub limit_order_asset: Option<asset::Id>,
```
erwanor, line 36: @hdevalence I've added this field to our `Position` struct. It's relevant when the LP is a limit-order. It is initialized to `None`, and will be tagged with the `asset::Id` of the inventory that the user provides when opening the position. That means that when we handle a position write, we know which asset to inspect to check their reserves. The drawback of course is that `Position` and `pb::Position` don't match perfectly. Curious to read your thoughts on this, when you have time.

hdevalence: Wait, why do we need an extra field, rather than just using the close_on_fill bool, and closing the position when either asset is set to zero?

hdevalence: In #2744 we have put_position become async so it can read the current state, that change seems relevant here even if the liquidity indexing part isn't implemented

erwanor: Ah yes, if `put_position` can read the current state then we can pull the position from the index, check the state of its reserves, determine that the asset provided is `1` or `2`, and inspect the reserves in `put_position` to figure out if that corresponds to a total fill.

erwanor:
The remaining items for this ticket are: adding tests and client support in `pcli`. I will follow-up on them later, for now I am merging this to avoid conflicts with other changes.

PR #2763, opened 2023-06-29
dex(router): define ordering on `Path<S>`

### LINKED ISSUE #2758: dex(router): `path_search` must produce unique results
In order to execute trades against the entire liquidity graph, the DEX engine performs a path search to detect the lowest cost path from a given source $S$ to a target $T$. Our path search implementation follows a modified version of Bellman-Ford that incorporates the SPFA optimization, trimming the search space by marking paths that are strictly worse than the current known best path as "inactive". 

Because candidate paths are evaluated concurrently, in an order that should presumed random and non-deterministic, it's critical that the update rule for tracking the best two paths be commutative (i.e. always settles to the same value regardless of the order of evaluation).

However, the current update rule:

```rust
    pub fn update(&mut self, new_path: Path<S>) {
        if new_path.price < self.path.price {
            tracing::debug!(new_price = %new_path.price, old_price = %self.path.price, "new path is better than best path, updating cache");
            self.spill = Some(std::mem::replace(&mut self.path, new_path));
            self.active = true;
        } else {
            // The new path is worse than the best path, but it might be better than the spill path.
            self.update_spill(new_path);
        }
    }
```
falls short of that requirement, because it assumes that the candidate paths' cost are **distinct**. However, this is not always the case, and in a liquidity graphs that contains multiple stable-kind assets, this is frequently not the case.

The consequence of two penumbra state-machines disagreeing over what is the best route to take is an eventual app hash mismatch. To see this, consider two routes: $S \rightarrow A \rightarrow T$ and $S \rightarrow B \rightarrow T$ with equal cost $c$. Executing over each route will result in a diverging view of the liquidity graph, in the first case the set of update position is: $\langle P_s, P_a, P_t \rangle$ in the second case, it is $\langle P_s, P_b, P_t \rangle$. Even though the output of the trade is identical, the views of the liquidity graph have diverged. This will trigger an app hash mismatch when a user attempts to swap $A$ for $T$ (many other triggers exist).

The update rule should be amended to:

```rust
    pub fn update(&mut self, new_path: Path<S>) {
        if new_path.price < self.path.price {
            tracing::debug!(new_price = %new_path.price, old_price = %self.path.price, "new path is better than best path, updating cache");
            self.spill = Some(std::mem::replace(&mut self.path, new_path));
            self.active = true;
        } else if new_path.price == self.path.price {
            // tie breaking rule deciding to branch to update or not.
        } else {
            // The new path is worse than the best path, but it might be better than the spill path.
            self.update_spill(new_path);
        }
    }
```

erwanor:
Closes #2758

### DIFF
--- a/crates/core/component/dex/src/component/router/path.rs
+++ b/crates/core/component/dex/src/component/router/path.rs
@@ -115,8 +115,6 @@ impl<S: StateRead + 'static> PartialOrd for Path<S> {
 impl<S: StateRead + 'static> Eq for Path<S> {}
 
 impl<S: StateRead + 'static> Ord for Path<S> {
-    /// First, we compare the price, then the path length, and
-    /// finally the nodes themselves.
     fn cmp(&self, other: &Self) -> Ordering {
         self.price.cmp(&other.price).then_with(|| {
             self.nodes
--- a/crates/core/component/dex/src/component/router/path.rs
+++ b/crates/core/component/dex/src/component/router/path.rs
@@ -115,14 +115,14 @@ impl<S: StateRead + 'static> PartialOrd for Path<S> {
 impl<S: StateRead + 'static> Eq for Path<S> {}
 
 impl<S: StateRead + 'static> Ord for Path<S> {
+    /// First, we compare the price, then the path length, and
+    /// finally the nodes themselves.
     fn cmp(&self, other: &Self) -> Ordering {
-        match self.price.cmp(&other.price) {
-            Ordering::Equal => match self.nodes.len().cmp(&other.nodes.len()) {
-                Ordering::Greater => Ordering::Greater,
-                Ordering::Equal => self.nodes.cmp(&other.nodes),
-                Ordering::Less => Ordering::Less,
-            },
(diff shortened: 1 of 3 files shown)


crates/core/component/dex/src/component/router/path.rs
```diff
@@ -98,3 +99,30 @@ impl<S: StateRead + 'static> Path<S> {
...
+    fn cmp(&self, other: &Self) -> Ordering {
+        match self.price.cmp(&other.price) {
+            Ordering::Equal => match self.nodes.len().cmp(&other.nodes.len()) {
+                Ordering::Greater => Ordering::Greater,
+                Ordering::Equal => self.nodes.cmp(&other.nodes),
+                Ordering::Less => Ordering::Less,
+            },
+            less_or_more => less_or_more,
+        }
+    }
```
hdevalence, line 127: Could this use `.then_with` to express the sequence of comparisons more succinctly? https://doc.rust-lang.org/stable/std/cmp/enum.Ordering.html#method.then_with

Issue #2765, opened 2023-06-29
Proposed `penumbra-crypto` code reorganization


redshiftzero:
This is a more detailed proposal for the final item in #2288. As described there, `penumbra-crypto` has become a catch-all crate for a bunch of diverse functionality. The scope of this ticket is to:

1. Rehome the various modules of `penumbra-crypto` to existing crates that make sense, or create new crates if needed, 
2. Remove code in `penumbra-crypto` we're not using, i.e. stuff that was originally targeted for Penumbra mainnet but will end up going into PenumbraV2, possibly with significant changes. 

## Concrete steps

Note that the following paths refer to the various modules in `penumbra-crypto` in `crates/core/crypto`.

- [x] Move `SwapFlow` (in `flow.rs`) to the `penumbra-dex` component. Remove `MockFlowCiphertext` and instead just use `Amount`s. This removes `flow.rs`. - #2766
- [x] Remove `eddy.rs` for now - we'll bring it back in the appropriate place if needed for PenumbraV2. - #2766
- [x] Move code in `governance*` to the `penumbra-governance` crate. Currently this crate doesn't exist, so in the meantime we could move it alongside where the governance component lives. Then this governance code would all be moved together to the `penumbra-governance` crate, as described/tracked in #2288. - https://github.com/penumbra-zone/penumbra/pull/2776
- [x] Move code in `stake*` to the `penumbra-stake` crate. - https://github.com/penumbra-zone/penumbra/pull/2770
- [x] Move `memo.rs` to the `penumbra-transaction` crate, since as of #1222 the memo is at the transaction level. - https://github.com/penumbra-zone/penumbra/pull/2773
- [x] Move `effect_hash.rs` and `transaction.rs` to the `penumbra-component` crate (we need to implement `EffectingHash` for component-specific actions in the component crates). - an attempt in #2779 that is failing CI, next attempt in #2798
- [x] Create new crate `penumbra-num` for `Amount` and `U128x128` types. - #2781
- [x] Move `asset*` to a new `penumbra-asset` crate, along with `Value` and `Balance` - #2781
- [x] Move `note_payload.rs`, `rseed.rs`, to `penumbra-shielded-pool` - #2795
- [x] For the proofs, we'll move each of the proofs to the appropriate component crate, as was already done for the `SwapProof` and `SwapClaimProof`. Specifically:
[...]

PR #2766, opened 2023-06-30
refactor: remove unused code in `penumbra-crypto`

redshiftzero:
First two items in #2765, as well as renames `tct::Commitment` to `tct::StateCommitment` to reflect the corresponding proto


redshiftzero:
Note: The protobuf Lint job will fail here in CI as there is a breaking change to the protobufs (removal of `MockFlowCiphertext`)

PR #2770, opened 2023-06-30
refactor: move penumbra-crypto staking-related types to staking crate 

redshiftzero:
Towards #2765

Unfortunately to do this without introducing a cyclic dependency (between `penumbra-stake` and `penumbra-chain`), the `Penalty` struct usage had to be removed from `penumbra-chain`. Please comment if that seems like a bad call!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4954,6 +4954,7 @@ dependencies = [
  "bech32",
  "bitvec",
  "decaf377",
+ "decaf377-rdsa",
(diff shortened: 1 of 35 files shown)


hdevalence:
I think moving the `Penalty` types out is fine. Longer term, I think we should evolve the chain parameters to be a collection of per-component parameters that are joined together in a top-level parameters structure at the `app` level.

Issue #2785, opened 2023-07-03
Stateless TxP/TxV generation for wasm code


hdevalence:
**Is your feature request related to a problem? Please describe.**

The browser extension needs to generate TxPs and TxVs in wasm'd Rust code, but that code doesn't have direct access to the state.  This is the cause of some visibility bugs in the current implementation. Transactions, even those created by the user, are being marked as "opaque", indicating that the extension is not be properly populating the TxP/TxVs.

Upon examining the Typescript and penumbra-wasm code, we have identified potential root causes related to how the extension fetches transactions and the data it passes to the Rust view server. Specifically, the problem seems to lie with the `ViewServer` object not being initialized with every single note and nullifier ever seen, and hence it fails to find the necessary notes for TxP population.

The method in question can be found at: [transaction_info](https://github.com/penumbra-zone/wallet/blob/main/src/utils/WasmViewConnector.ts#L350) in WasmViewConnector.ts, which is defined in [view_server.rs](https://github.com/penumbra-zone/penumbra/blob/main/crates/wasm/src/view_server.rs#L385) and calls into [transaction_info_inner](https://github.com/penumbra-zone/penumbra/blob/main/crates/wasm/src/view_server.rs#L416), which in turn ports the logic for the Rust view server.

To fix this, we need to restructure the TxP/TxV generation so that it is stateless -- or rather, is an explicit state machine, where explicit context structures feed necessary data in at each step.

Valentine1898:
I don't seem to have been able to properly associate this issue to PR https://github.com/penumbra-zone/penumbra/pull/2787

PR #2787, opened 2023-07-05
Fixed penumbra-zone/penumbra#2785 stateless tx p/tx v generation for wasm code

zpoken:
Part of #2785

### DIFF
--- a/crates/wasm/Cargo.toml
+++ b/crates/wasm/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "penumbra-wasm"
-version = "0.55.0"
+version = "0.55.6-indexeddb"
 authors = ["Valentine <valentine@zpoken.io>"]
 edition = "2021"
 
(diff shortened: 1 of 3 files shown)


crates/wasm/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "penumbra-wasm"
-version = "0.55.0"
+version = "0.55.6-indexeddb"
```
conorsch, line 3: Version should remain `0.55.0`, because we set the version to the same value across all crates, and increment in tandem at release time.

crates/wasm/src/view_server.rs
```diff
@@ -16,7 +19,7 @@ use tct::storage::{StoreCommitment, StoreHash, StoredPosition, Updates};
 use tct::{Forgotten, Tree};
 use wasm_bindgen::prelude::wasm_bindgen;
 use wasm_bindgen::JsValue;
-use web_sys::console as web_console;
+use web_sys::{console as web_console};
```
conorsch, line 22: L22 fails `cargo fmt`, will patch locally and push up to make CI happy.

conorsch (commented):
A few cosmetic changes. Haven't tested functionality, but if this works for you, happy to land it.

PR #2793, opened 2023-07-06
ibc: use ibc-types 0.3.0 instead of main

avahowell:


conorsch:
Failing WASM check: @avahowell suspects we just need to click on the `std` feature for the ibc-types dep.

conorsch:
Heads up, we just merged https://github.com/penumbra-zone/ibc-types/pull/38 which we'll definitely want in for Testnet 55. If subsequent testing with https://github.com/penumbra-zone/penumbra/pull/2803 looks good, we'll either want to bump ibc-types to 0.3.1 to include the latest, or keep the dep on main for now. What we _don't_ want is to pin 0.3.0 of ibc-types for 55, which would ship a known-bad ibc implementation.

PR #2798, opened 2023-07-07
refactor: move remaining code in `penumbra-crypto` to `penumbra-chain`

### LINKED ISSUE #2765: Proposed `penumbra-crypto` code reorganization

redshiftzero:
Closes #2765


redshiftzero:
I split out #2800 into another PR since this PR is failing the wasm build and I need to debug further to understand why.

conorsch:
We aim to land this PR prior to shipping Testnet 56 (https://github.com/penumbra-zone/penumbra/issues/2761).

redshiftzero:
For posterity, the issue with the wasm build was:

On main,
* `penumbra-component` depends on `penumbra-storage`
* `penumbra-storage` depends on `tokio`

In this PR previously:
* `penumbra-component` is added to the dependencies for `penumbra-transaction`, among other crates that need to be wasm-friendly -> the problem
* so the approach of moving `effect_hash.rs` to `penumbra-component` just doesn't work

Two ways address this would be to:
1. either move `effect_hash.rs` to `penumbra-chain` (already a dependency of `penumbra-wasm` ) OR
2. make a minimal crate e.g. `penumbra-effect-hash` that can be used by penumbra-transaction and the various penumbra component crates

I've done approach 1 here to resolve since the amount of code in question is fairly small.

PR #2799, opened 2023-07-07
Add ability to query for balances by account and asset type

### LINKED ISSUE #2768: Allow filtering view service balance queries by asset

Currently, the only way to query balances is the `BalanceByAddress` RPC, which has one argument, the address. It would be convenient if it could filter the returned balances by asset ID.  In particular, now that the `AssetId` proto structure allows [alternate encodings](https://buf.build/penumbra-zone/penumbra/docs/35a7026ca3bf4b1090ac1a49833114bc:penumbra.core.crypto.v1alpha1#penumbra.core.crypto.v1alpha1.AssetId), this would allow an RPC user to query balances using a base denom string.

We may also want to re-examine the RPC. It's `BalanceByAddress`, but addresses aren't a unit of accounting, accounts are.  Also, a user might want to request the total balance across all accounts.

Change the RPC to `rpc Balances(BalancesRequest) returns (stream BalancesResponse)`.

Change `BalanceByAddressRequest` to
```proto
message BalancesRequest {
  // If present, filter balances to only include the account specified by the `AddressIndex`.
  core.crypto.v1alpha1.AddressIndex account_filter = 1;
  // If present, filter balances to only include the specified asset ID.
  core.crypto.v1alpha1.AssetId asset_id_filter = 2;
}
```

Change `BalanceByAddressResponse` to
```proto
message BalancesResponse {
  penumbra.core.crypto.v1alpha1.AddressIndex account = 1; 
  penumbra.core.crypto.v1alpha1.Value balance = 2;
}
```

Update the Rust view server implementation to reflect these changes. For now, I think it's fine for the Rust view server to return `tonic::Status::unimplemented` if `account_filter` is unset, to minimize changes to the existing code; if anyone ever complains, we'll know that someone wants to request multiple accounts simultaneously.

**Additional context**
This unblocks work on InterchainTest, by making it easier to write balance tests without having to build asset ID lookup tables.

aubrika:
Closes #2768


conorsch:
@aubrika Thanks for tackling this on short notice. Still very interested in shipping this feature, but it's OK if it slips off 55: we can work with SL to use custom container versions, so whenever this lands, we'll have an associated hash; plus, all of the `view.*` protos are currently inaccessible to us in go until we resolve https://github.com/strangelove-ventures/interchaintest/issues/658

aubrika:
Broken CI is just due to proto changes, merging in as-is in advance of potential future feature requests

Issue #2748, opened 2023-06-27
docs: web wallet, how to upgrade between testnets


conorsch:
We should figure out what the recommended user actions are for web wallet when a new testnet is released. It's not even clear to me yet. In the past, I've simply uninstalled-and-reinstalled the extension to get the latest version, which is clunky. Questions I have:

1. Is it possible to upgrade the version of the extension from within the Chrome Web Store? I've never seen an "upgrade" button anywhere, but I'm a Firefox user by day, so perhaps I'm missing something obvious.
2. When a new testnet is released, should we assume a new version of the extension is also released to bump the chain id? If so, we need a solution to 1.
3. If the answer to 2 is no (and IMO it should be), then we need to document how to edit the chain settings within the extension. There's a UI flow for this, but it might be tricky for less technical users.

All these improvements should land in https://guide.penumbra.zone/main/extension.html

conorsch:
> Is it possible to upgrade the version of the extension from within the Chrome Web Store? 

According to Zpoken, one can expect the web extension to update _automatically_ within Chrome. I haven't observed this myself yet, as I was expecting a specific button to perform the upgrade manually.

> If the answer to 2 is no (and IMO it should be), then we need to document how to edit the chain settings within the extension. 

The proper procedure here is to "reset" the extension, then it will resync on the new testnet. This must be done manually, and therefore should be documented.

conorsch:
> According to Zpoken, one can expect the web extension to update automatically within Chrome.

A new version of the extension, v0.2.3, was just released: https://github.com/penumbra-zone/wallet/releases/tag/0.2.3 and it's already in the Web Store: https://chrome.google.com/webstore/detail/penumbra-wallet/lkpmkhpnhknhmibgnmmhdhgdilepfghe/ My local browser has v0.2.2 installed, so I'm leaving it running and will monitor for an automatic update.

conorsch:
> My local browser has v0.2.2 installed, so I'm leaving it running and will monitor for an automatic update.

After an hour, no change, but then restarting the browser instantly updated it to 0.2.3. So I've observed automatic upgrades of the web extension, as described.

PR #2810, opened 2023-07-10
docs: web extension testnet upgrades

conorsch:
Just ran through the procedure after release of v0.56.0, and jotted down some notes for folks to use.

Closes https://github.com/penumbra-zone/penumbra/issues/2748


aubrika (approved):
LGTM! :)

PR #2818, opened 2023-07-12
chore: update chain id for 057

conorsch:


conorsch:
Smoke test is failing, presumably due to changes in https://github.com/penumbra-zone/penumbra/pull/2817/files. I think that's just a missing newline. Will fix on this PR until CI is passing.

PR #2820, opened 2023-07-12
fix ics20 withdrawal: include withdrawals in effect hash and tx builder

avahowell:


conorsch:
Merging, will test between preview & local devnet shortly.

PR #2822, opened 2023-07-12
fix: update rocksdb from v0.19 to v0.21

cronokirby:
Depending on the C++ toolchain used by the underlying system, builds of rocksdb might fail because of an issue documented in:

https://github.com/facebook/rocksdb/pull/11118

(In short, the code relied on headers importing other headers, rather than what the standard guarantees headers to export; thus code may not compiler on all toolchains).

Version 0.21 of the crate fixes this issue, by relying on a later version of the underlying rocksdb library.

For the rust crate, this issue showed up in:

https://github.com/rust-rocksdb/rust-rocksdb/issues/772

On Linux/Manjaro I ran into this issue as well, but I expect that other people following along with the docs might run into it as well, so probably worth updating, unless there's some issue with doing so.


redshiftzero (approved):
When CI passes this LGTM!

PR #2821, opened 2023-07-12
spec: fix links to value commitments section

redshiftzero:

PR #2824, opened 2023-07-12
docs: fix broken links in protocol spec

conorsch:
Used `cargo install mdbook-linkcheck` to identify broken internal links. There are still a few failing checks in the protocol spec, but they're all about threshold encryption articles not being listed in SUMMARY.md. Fine to circle back on that.


redshiftzero (approved):
thanks for fixing these! I merged #2821 so this can be rebased

PR #2828, opened 2023-07-13
pd: add EventIndexLayer to enable tendermint event indexing

avahowell:


hdevalence:
Under-advertised in this PR: the commit adds configurable event filtering based on regex matching, so that in the future, if we hook it up to a config mechanism, users can configure regex-based filters for which events they want indexed.

PR #2829, opened 2023-07-13
ibc: use ibc-types typed events, add SendPacket event

avahowell:


crates/core/component/ibc/src/component/client.rs
```diff
@@ -32,44 +28,6 @@ use super::state_key;
 
 #[async_trait]
 pub(crate) trait Ics2ClientExt: StateWrite {
-    // execute a UpdateClient IBC action. this assumes that the UpdateClient has already been
-    // validated, including header verification.
-    async fn execute_update_client(&mut self, msg_update_client: &MsgUpdateClient) -> Result<()> {
```
avahowell, line 37: this is dead code; we inlined it into the msg handler for update client a while ago

PR #2831, opened 2023-07-13
ci: use posix regex

erwanor:
grep PCREs isn't supported in OS X. I'm replacing this one by a POSIX friendly one


erwanor:
@conorsch i think the cargo warning error is spurious, feel free to merge this (or not!) if it works on linux

PR #2833, opened 2023-07-13
Use commitment in NullfierStateDerivation proof

### LINKED ISSUE #2764: modify nullifier derivation circuit to take a state commitment
In #2526 we added a new ZK proof that certifies that a nullifier was derived correctly given a witnessed note. However, @hdevalence pointed out that we only need to demonstrate that a nullifier was derived correctly given a state commitment, i.e. to link the revealed nullifiers in a spend with a specific state commitment. 

This means that instead of having `Note` as a public input in the `NullifierDerivationCircuit`, we can just have a `StateCommitment` as a public input. Since we no longer have to derive the commitment in-circuit, we should update the benchmarks [here](https://guide.penumbra.zone/main/dev/parameter_setup.html#benchmarks) to reflect the lower constraint cost and proving time.

cronokirby:
Fixes #2764.

Previously, this proof would take in a public note, and compute the commitment in circuit. This PR changes the proof to take in the commitment directly, thus saving on constraints.


cronokirby:
For merging, we may not want to be modifying proof-related stuff quite at the moment, but this is a minor change.

crates/core/component/shielded-pool/src/nullifier_derivation.rs
```diff
@@ -60,12 +61,12 @@ impl ConstraintSynthesizer<Fq> for NullifierDerivationCircuit {
 
         // Public inputs
         let claimed_nullifier_var = NullifierVar::new_input(cs.clone(), || Ok(self.nullifier))?;
-        let note_var = note::NoteVar::new_input(cs.clone(), || Ok(self.note.clone()))?;
+        let note_commitment_var =
```
redshiftzero, line 64: One other change we should make as part of this is below in `NullifierDerivationProof::verify` we'll want to pass in the note commitment instead of the `Note` as an argument, and then when we construct the public inputs the second element in the public input vector ([here](https://github.com/penumbra-zone/penumbra/pull/2833/files#diff-f4e7e8f2b1ba9fb55a1dbdc472f800bb724bc65e810b56db8a9f4db2a13c1d4fL150)) should be the note commitment. Otherwise this LGTM!

cronokirby: Fixes in the latest commits 8^)

redshiftzero (approved):
LGTM!

redshiftzero:
Looks like the CI failure here was unrelated to the diff in this PR and was resolved with #2838, so merging

PR #2835, opened 2023-07-13
fix(pcli): help text for ibc withdrawals

conorsch:


crates/bin/pcli/src/command/tx.rs
```diff
@@ -828,8 +845,10 @@ impl TxCmd {
                     timeout_timestamp = current_time_u64 + 172800u64;
                 }
 
-                let denom = asset::REGISTRY.parse_denom(denom).unwrap();
-                let amount = Amount::try_from(amount.clone()).unwrap();
+                let denom = asset::REGISTRY.parse_denom(denom).ok_or_else(|| {
+                    anyhow::anyhow!(format!("unable to parse denomination '{}'", denom))
+                })?;
```
conorsch, line 850: Withdrawing `penumbra` throws an error, but sending `upenumbra` does not. That's confusing. We should look for opportunities to reconcile the "base denom" vs "display denom" nomenclature throughout user-facing views.

hdevalence: Ideally there shouldn't be any reconciliation necessary, since the command line tooling should work with any known unit.

conorsch: If we address the TODO about parsing the denom and amount from a concatenated string, as we do with `pcli tx send`, this problem should go away.

Issue #2834, opened 2023-07-13
ibc: relayer can't find nodekey during packet proof query


conorsch:
**Describe the bug**
The [relayer](https://github.com/cosmos/relayer) currently fails to process withdrawals between Penumbra chains with the following error:

```
2023-07-13T21:58:00.807818Z     error   Error assembling packet message {"path_name": "penumbra_path", "chain_id": "penumbra-devnet-local-986947bb", "client_id": "07-tendermint-0", "msg": {"type": "timeout_packet", "src_port": "transfer", "src_channel": "channel-0", "dst_port": "transfer", "dst_channel": "channel-2", "sequence": 1, "timeout_height": "0-28824", "timeout_timestamp": 1689458273, "data": "CkFwYXNzZXQxOTg0ZmN0ZW53OG0yZnBsOGE5d3pndXpwN2ozNGQ3dnJhdnJ5dWhmdDgwOG55dDlmZGdncXhtYW5xbRIBMRqSAXBlbnVtYnJhdjJ0MXg0OGhkY2pxYTVqbGtneDJ3bmpzMHNnNjducWh3dHgwOGxreHVmNmo5YTh1cTk2MHlwbGNudnd1ZjJubm55NTM3OTJxZDg0bTIzY2o3azl0MjNlaDdlNmt3NTJsam4zcjNnZXZqbGttaGE5em54OXVxcXo3cWc0ZHQ4NmM5eTJuazA2bmh5IpIBcGVudW1icmF2MnQxa25lcWp4eDhndDJ0d2M4ODVnZm43bjJ0azY4bTczMDQzMmh4bHpuNDdocnptejI5YTR0aGNjejR5cWZrZnZqdm1qa3d4cHk0N2NscWhmNDVkOHpxOWo5ZWZ4ajk0YzJqeWhnem01OXNsZjhha3Bnd3cwNnh6ZWthNncyazA5ZnFhcWtncW4=", "ack": ""}, "error": "error querying packet proof: error querying tendermint proof for packet receipt: rpc error: code = Unknown desc = Missing node at NodeKey { version: 2338, nibble_path: 49c5 }."}
```

**To Reproduce**
1. Fire up a local devnet/preview connection, following docs in https://github.com/penumbra-zone/penumbra/tree/ed1cb0734aadb36059eb147c57b30694ff578002/deployments/relayer#running-a-local-devnet
2. Run the relayer via `rly start`
[...]

PR #2837, opened 2023-07-14
fix packet commitment paths: use `sequences` correctly

avahowell:
Our packet commitment paths were wrong, because they were modeled on the IBC spec, which contains state commitment paths different than what is currently deployed in `ibc-go`:

https://github.com/cosmos/ibc/blob/854f2e13ccab539dd99d9889dd9367339875f73e/spec/core/ics-004-channel-and-packet-semantics/README.md?plain=1#L231

vs ibc-go:

https://github.com/cosmos/ibc-go/blob/56a519c188ae841818b9ecd11244240941cc5d90/modules/core/24-host/keys.go#L212

this PR brings the penumbra commitment paths inline with `ibc-go`.

Should address #2834, but let's avoid closing that issue until we confirm with testing.

### DIFF
--- a/crates/core/component/ibc/src/component/state_key.rs
+++ b/crates/core/component/ibc/src/component/state_key.rs
@@ -75,14 +75,14 @@ pub fn seq_send(channel_id: &ChannelId, port_id: &PortId) -> String {
 
 pub fn packet_receipt(packet: &Packet) -> String {
     format!(
-        "receipts/ports/{}/channels/{}/receipts/{}",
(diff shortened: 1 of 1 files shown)


avahowell:
In the longer term, think we should probably move these commitment paths to `ibc-types`. cc @hdevalence

conorsch:
As usual, will merge this onto preview, and re-run the devnet/preview script locally to evaluate behavior.

Issue #2808, opened 2023-07-10
Rework EffectHash


redshiftzero:
This is a tracking ticket for modifying the `EffectHash`:

- [x] Refactor TransactionBody such that its component pieces can be incrementally hashed (e.g. by a hardware wallet): #2784
- [x] Combine those fields into the full transaction effect hash e.g. `H(fixed size hashes of per tx data packets || num actions || fixed size hashes per action)`
- [x] Use the protos to encode the effect hash for individual actions: #2000, ensuring for each action that where needed the current data modeling separates the effecting data (e.g. SpendBody vs Spend), and that only that field is included in the effect hash - https://github.com/penumbra-zone/penumbra/pull/2931

conorsch:
Raising priority on this ticket, moving to top of Next column. We still need more discussion on design. The motivation for this work is to reduce the likelihood that a developer will overlook these interfaces and thereby introduce bugs; see for example https://github.com/penumbra-zone/penumbra/pull/2820/files.

cronokirby:
Some ideas after more discussion with @redshiftzero:

- For handling padding, we could use a similar api to the [GroupHasher](https://github.com/penumbra-zone/penumbra/blob/main/crates/crypto/proof-setup/src/group.rs#L130) utility we've just introduced for the setup ceremony code.
- For handling ordering of items we need to include in the effect hash (like the recent PR in https://github.com/penumbra-zone/penumbra/pull/2820/), one solution would be to hash all of the items, and then sort the hashes, before including those in the hash state. This would provide a canonical total ordering for any item that can be hashed, and thus avoid the need to worry about having a consistent ordering elsewhere in the code. Not sure if that's totally solving ordering, but seems to move the issue away from the effect hash side of things.
- For handling code reuse with proto serialization, the tricky part is when a structure mixes both "atomic" fields (string, ints, etc.) and composite substructures that need to be hashed. The issue is that we can't then just hash the serialization of the struct, because the sub-structures need to be hashed first. One way around this, that requires touching many protos, would be to make it so that each structures is always either a collection of sub-structures, or a collection of atomic fields. This would mean that the only two codepaths would be hashing all sub-structures before combining those hashes, or serializing the entire structure as a protobuf and hashing that.

PR #2839, opened 2023-07-14
refactor: add `TransactionParameters` and `DetectionData` fields in `TransactionBody`

### LINKED ISSUE #2784: Refactor TransactionBody to include a TransactionParameters

Conceptually, a Penumbra transaction is just a bundle of actions that have effects on the chain state. In practice, though, transactions have a bunch of other transaction-wide fields, like the encrypted memo, the chain ID, expiry heights, and so on.

Currently, these fields are all put into the transaction body without any organization. Instead, it would be cleaner to bundle them together. Other than the encrypted memo, all of the existing fields in the transaction body specify parameters about when the transaction should be accepted to a chain — for instance, the chain id the transaction is for, the expiry height after which it is no longer valid, etc.

This suggests a way to conceptually clarify the definition of the transaction data, as a few conceptually distinct items:

- a list of actions that have effects on the chain state;
- an encrypted memo describing information about the purpose of the transaction;
- detection information helping other users detect it;
- a fee that pays for its processing;
- a set of transaction parameters controlling when the transaction should be accepted.

Refactor the transaction data modeling to introduce a TransactionParamaters proto and domain type.

We may also want to introduce a DetectionData proto message containing the FMD clues, so that it can be hashed as a single unit (see below) and for forward extensibility at the protocol level.

**Additional Context**

This will also allow us to refactor the effect hash computation so that a signer can consider each of the parts of the transaction in turn, without having to load an entire transaction plan all at once. The effect hash could be computed by, for instance, individually hashing the memo packet, the parameters packet, and each of the actions, and then combining these hashes.

redshiftzero:
Closes #2784

The scope of this is refactoring the `TransactionBody` to contain two new fields:
* the `TransactionParameters` (new struct) that contains the expiry height and chain ID (both previously fields directly on the `TransactionBody`). 
* the optional `DetectionData` (new struct) that contains the Fuzzy Message Detection (FMD) clues for that transaction (previously a field directly on the `TransactionBody`).

In this PR I have _not_ modified the EffectHash implementation (see #2808). The idea is (per the comment at the end of #2784):
> This will also allow us to refactor the effect hash computation so that a signer can consider each of the parts of the transaction in turn, without having to load an entire transaction plan all at once. The effect hash could be computed by, for instance, individually hashing the memo packet, the parameters packet, and each of the actions, and then combining these hashes.

Note: CI job `Protobuf / Lint protobuf (pull_request)` will fail during the check for breaking changes to the protobufs, but breaking changes to the protobufs are unavoidable here.


proto/penumbra/penumbra/core/transaction/v1alpha1/transaction.proto
```diff
@@ -31,20 +31,32 @@ message EffectHash {
...
   repeated Action actions = 1;
+  // Parameters determining if a transaction should be accepted by this chain.
+  TransactionParameters transaction_parameters = 2;
+  // The transaction fee.
+  crypto.v1alpha1.Fee fee = 3;
+  // Detection data for use with Fuzzy Message Detection
+  DetectionData detection_data = 4;
+  // An optional encrypted memo. It will only be populated if there are
+  // outputs in the actions of this transaction. 528 bytes.
+  optional bytes encrypted_memo = 5;
```
hdevalence, line 42: Can we make this a sub-message (maybe `MemoData`?) rather than an `optional bytes`?  That way, we'll have better extensibility (if we ever want to add fields, we can add it to the `MemoData`) and it will be easier to do packetwise effect hashing (we can just hash the `MemoData`).

hdevalence (changes requested):
This looks good, but I think we should pull out the encrypted memo into another proto message so that the `TransactionBody` only has sub-messages as fields.  (We don't have to do anything on the domain types, I think just changing the protos is sufficient for forward compatibility).

crates/core/transaction/src/transaction.rs
```diff
@@ -445,6 +469,66 @@ impl From<TransactionBody> for Vec<u8> {
...
+}
+
+impl From<DetectionData> for pbt::DetectionData {
+    fn from(msg: DetectionData) -> Self {
+        let mut fmd_clues = Vec::new();
+        for fmd_clue in msg.fmd_clues {
+            fmd_clues.push(fmd_clue.into());
+        }
+
+        pbt::DetectionData { fmd_clues }
```
cronokirby, line 528: Maybe we could have this match the `TryFrom` impl above, using an iterator? Not sure what convention is preferred.

```suggestion
        let fmd_clues = proto
            .fmd_clues
            .into_iter()
            .map(|x| x.try_into())
            .collect();
            
        pbt::DetectionData { fmd_clues }
```

cronokirby (commented):
I agree with @hdevalence about the protobuf comments.

Had a very minor nit in terms of code simplification as well.

Otherwise, looks good to me! :)

redshiftzero:
This has had review comments addressed - lmk if there is any other feedback else I'll plan to merge in the next few hours

cronokirby (approved):
Looks good to me!

PR #2843, opened 2023-07-18
Index available liquidity as part of put_position

### LINKED ISSUE #2378: Maintain indices of available liquidity for open positions

zbuc:
This is missing tests.

Closes #2378

### DIFF
--- a/crates/core/component/dex/src/component/position_manager.rs
+++ b/crates/core/component/dex/src/component/position_manager.rs
@@ -1,3 +1,4 @@
+use std::future;
 use std::{pin::Pin, sync::Arc};
 
 use anyhow::Result;
@@ -12,7 +13,7 @@ use penumbra_proto::DomainType;
 use penumbra_proto::{StateReadProto, StateWriteProto};
 use penumbra_storage::{EscapedByteSlice, StateRead, StateWrite};
 
-use crate::lp::Reserves;
+use crate::lp::position::State;
 use crate::{
     lp::position::{self, Position},
     state_key, DirectedTradingPair,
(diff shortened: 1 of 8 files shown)


crates/core/component/dex/src/component/position_manager.rs
```diff
@@ -252,5 +288,128 @@ pub(super) trait Inner: StateWrite {
...
+                        Value {
+                            asset_id: trading_pair.end,
+                            amount: position.reserves_for(trading_pair.end).unwrap_or_default(),
+                        },
+                        &position.reserves,
+                    )?
+                    .2
+                    .amount
+            }
+            _ => Amount::zero(),
```
erwanor, line 363: We might want to re-organize the method around a match statement over the previous position and the state of the current one. The state machine for this method looks like something like this:

```rust
match (current_position.state, prev_position) {
    (State::Opened, None) => we just have to add to the index, no deletion necessary
    (State::Opened, Some(prev)) => we add to the index, deleting previous entry
    (State::Closed, None) => we return early nothing to be done,
    (State::Closed, Some(prev)) => compute the previous contribution and erase it from the current index
}
```



I think we want to re-organize this method around a match statement on the current position's state, returning early if the current position is `Closed` for example.

crates/core/component/dex/src/component/position_manager.rs
```diff
@@ -187,16 +197,42 @@ pub trait PositionManager: StateWrite + PositionRead {
...
+            .take(DYNAMIC_ASSET_LIMIT);
+        try_stream! {
+            // First stream the fixed candidates, so those can be processed while the dynamic candidates are fetched.
+            for candidate in fixed_candidates.iter() {
+                yield candidate.clone();
+            }
+
+            // Yield the liquidity-based candidates. Note that this _may_ include some assets already included in the fixed set.
+            while let Some(candidate) = dynamic_candidates
+                .next().await {
```
erwanor, line 216: Thinking more about it, I think you're right that dupes are going to be more the rule than the exception. Henry suggested using a filter on the stream iterator. That's probably a good idea

PR #2848, opened 2023-07-19
Fixes transaction hash parsing bug

### LINKED ISSUE #2844: Transaction Hash parsing error in transaction_info_by_hash
_Originally posted by @aubrika in https://github.com/penumbra-zone/penumbra/issues/2547#issuecomment-1639130720_

Repro steps:

Start pclientd: `cargo run --release --bin pclientd -- --home ./ start`

Using buf studio, no transaction info is returned.

Screen Shot 2023-07-18 at 09 36 46

A `pclientd` debugging branch with added console output shows the wrong hash on the storage level:

Screen Shot 2023-07-18 at 09 38 08

But, the transaction is viewable in `pcli v tx`:

Screen Shot 2023-07-18 at 09 40 16

Likewise, the transaction hash is available in DBBrowser in local storage:

Screen Shot 2023-07-18 at 09 41 19

From this research, it appears we are doing the conversion wrong somewhere - pcli is calling the same storage function in a way that returns data, while the ViewClient protocol is not digesting the hash correctly from the web frontend. 

Still unknown where exactly the hash is being read or parsed incorrectly, debugging continues.

aubrika:
Closes #2844

### DIFF
--- a/proto/penumbra/penumbra/core/transaction/v1alpha1/transaction.proto
+++ b/proto/penumbra/penumbra/core/transaction/v1alpha1/transaction.proto
@@ -20,7 +20,7 @@ message Transaction {
 
 // A transaction ID, the Sha256 hash of a transaction.
 message Id {
-  bytes hash = 1;
+  string hash = 1;
 }
 
 message EffectHash {
--- a/crates/bin/pcli/src/command/view/tx.rs
+++ b/crates/bin/pcli/src/command/view/tx.rs
@@ -225,7 +225,8 @@ impl TxCmd {
     }
     pub async fn exec(&self, app: &mut App) -> Result<()> {
         let fvk = app.fvk.clone();
-        let hash = self
+
+        let hash: penumbra_transaction::Id = self
             .hash
             // We have to convert to uppercase because `tendermint::Hash` only accepts uppercase :(
             .to_uppercase()
--- a/crates/core/transaction/src/id.rs
+++ b/crates/core/transaction/src/id.rs
@@ -36,6 +36,7 @@ impl FromStr for Id {
         }
         let mut id = [0u8; 32];
         id.copy_from_slice(&bytes);
+
         Ok(Id(id))
     }
 }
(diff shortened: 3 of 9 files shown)


proto/penumbra/penumbra/core/transaction/v1alpha1/transaction.proto
```diff
@@ -20,7 +20,7 @@ message Transaction {
 
 // A transaction ID, the Sha256 hash of a transaction.
 message Id {
-  bytes hash = 1;
+  string hash = 1;
```
hdevalence, line 23: The transaction hash should be `bytes`, so that we transmit it as binary data. Otherwise, the proto doesn't specify the data format properly, because it doesn't specify how the string encodes the hash (is it base64? hex? something else?)

We should leave the encoding as `bytes`, and fix any translation errors where they occur, rather than propagating the transation errors into our own data modeling.

aubrika: Heard, and accounted for in a later debugging branch (#2852) which leaves the proto hash field as-is.

`pcli` is currently operating under the assumption that the bytes field in this request proto will be populated from a string, and is relying on .parse()/from_str to accordingly and successfully interpret those bytes.

In #2852 as a debugging sanity check I was able to successfully read the `hash` field in the proto (using the web gRPC UI) as copied from pcli output, but doing so breaks `pcli v tx` insofar as it starts failing to return transaction info from the client.

I'm out of time today, but for tomorrow:

1. Implement your suggestion (https://github.com/penumbra-zone/penumbra/issues/2844#issuecomment-1641860008) of adding an alt_hex field to the proto & ensure that the client code continues to work with pcli while also supporting retrieving transaction info from hex-encoded strings.
2. Once both sides of this (pcli and pclientd-via-webgrpcui) are working reliably, annotate the various field encodings appropriately so that we know which data format should be specified/what people should be passing to each optional field.

aubrika:
superseded by #2852

PR #2851, opened 2023-07-19
dex(router): internally split execution

### LINKED ISSUE #2842: dex: split execution of large batches
## Context

When an overflow happen during DEX execution, the engine closes the specific position that was tentatively filled and restarts the process of routing and filling the batch. This approach is predicated on attributing the overflow onto a position (or trading function) rather than the specified swap input. But since trades are aggregated into batches, the effective upper bound on a batch is  the minimum of `u128::MAX`, or all the the circulating supply for that token in the shielded pool. While processing such a large batch at once is unlikely, we still have to deal with the fact that approaching this number increases the likelihood of overflow occurring during internal dex calculations.

## Solution

We should detect and split large batches into smaller chunks that are sequentially executed by the DEX engine (to avoid introducing any meaningful notion of economic ordering).

erwanor:
Closes #2842

### DIFF
--- a/crates/core/component/dex/src/component/router/route_and_fill.rs
+++ b/crates/core/component/dex/src/component/router/route_and_fill.rs
@@ -13,6 +13,7 @@ use crate::{
         router::{FillRoute, PathSearch, RoutingParams},
         PositionManager, StateWriteExt,
     },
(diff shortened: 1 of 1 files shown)


erwanor:
This is going into assurance.

PR #2853, opened 2023-07-20
miscellaneous ICS20 fixes

avahowell:
- fix inversion in is_source logic
- fix inversion in Unreceived{Packets,Acks} query logic
- fix RecvPacket proof verification path
- fix encoding of packet commitments in RecvPacket proof verification
- fix timestamp encoding in `pcli tx withdraw`

After these fixes, I was able to execute a test ICS20 withdrawal from testnet-preview to a local devnet.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -151,9 +151,9 @@ dependencies = [
 
 [[package]]
 name = "allocator-api2"
-version = "0.2.15"
+version = "0.2.16"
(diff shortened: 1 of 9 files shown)


conorsch:
Nice, looking forward to testing this interactively!

N.B. the `tendermint` crates were bumped `0.32.0` -> `0.32.2` in this PR, which broke the docs build. Will fix that shortly.

PR #2861, opened 2023-07-21
fix query overflow: don't subtract one when height is zero

avahowell:


crates/bin/pd/src/info.rs
```diff
@@ -99,7 +99,8 @@ impl Info {
             //
             // Try to do this behavior by querying at height = latest-1.
             0 => {
-                let height = self.storage.latest_snapshot().version() - 1;
+                let version = self.storage.latest_snapshot().version();
+                let height = if version == 0 { version } else { version - 1 }
```
erwanor, line 103: alternatively, `saturating_sub` works too

avahowell: ooh good idea

PR #2863, opened 2023-07-22
ICS20 withdrawal: remove unnecessary chain id, parse amount from command

### LINKED ISSUE #2849: Change `pcli` ICS20 withdrawal command to take a value as a string
If we address the TODO about parsing the denom and amount from a concatenated string, as we do with `pcli tx send`, this problem should go away.

_Originally posted by @conorsch in https://github.com/penumbra-zone/penumbra/pull/2835#discussion_r1263934343_

avahowell:
Closes #2849 

This PR changes the `ICS20Withdrawal` type to remove the chain ID, which I don't think is required (the chain ID is implied by the `--source-channel`), and parses the value as a string from the command, similar to `tx send`. This cleans up the UX for withdrawals:

`pcli transaction withdraw --to penumbrav2t1xsfq50zc8kctc6k8e0lr9uy7qeku3c3racv3f6keu0w9qk2cmfyfkm5dwqk50ffuftjjauz7zu436h6gde3p5yq8w5pkg9nken48qzpvfcvclpadmxkk9u9enfys7j9p8ml5y5 --source-channel channel-0 100penumbra`

instead of

`pcli transaction withdraw --to penumbrav2t1xsfq50zc8kctc6k8e0lr9uy7qeku3c3racv3f6keu0w9qk2cmfyfkm5dwqk50ffuftjjauz7zu436h6gde3p5yq8w5pkg9nken48qzpvfcvclpadmxkk9u9enfys7j9p8ml5y5@penumbra-testnet-ganymede-e6a65bff --timeout-timestamp 16898230646320000000 --source-channel channel-0 upenumbra 100000000`

this also includes a security fix: we should include the source channel and source port and denom in the ICS20Withdrawal's effectingdata.


crates/core/component/ibc/src/ics20_withdrawal.rs
```diff
@@ -84,17 +80,18 @@ impl EffectingData for Ics20Withdrawal {
...
-        let destination_chain_id_hash =
-            blake2b_simd::Params::default().hash(self.destination_chain_id.as_bytes());
         let destination_chain_address_hash =
             blake2b_simd::Params::default().hash(self.destination_chain_address.as_bytes());
+        let return_address = blake2b_simd::Params::default().hash(&self.return_address.to_vec());
 
-        state.update(destination_chain_id_hash.as_bytes());
         state.update(&self.value().amount.to_le_bytes());
         state.update(&self.value().asset_id.to_bytes());
+        state.update(&self.denom.encode_to_vec());
```
avahowell, line 89: i don't think we actually need this, since asset_id is computed by denom and covered by the hash, but I included it here for completeness

PR #2865, opened 2023-07-24
ibc: elide source port in Ics20Withdrawal

avahowell:
Since withdrawals always use the transfer port, there is no reason to include `source_port` in the Ics20Withdrawal action


conorsch:
@avahowell Do you mind rebasing this and regenerating the proto descriptor file? Looks like it'll pass CI then.

avahowell:
rebased and updated proto descriptor

conorsch:
The sole failing CI check is flagging breaking changes to the protos. We're in good contact with SL about potential implications, so proceeding with merge for inclusion in Testnet 58.

PR #2866, opened 2023-07-24
spec: fix out of date note commitment docs, add closing parentheses

redshiftzero:
Once we started adding the clue key as part of the note commitment, we forgot to update the documentation. This PR fixes that, as well as adds missing closing parentheses for the state commitments.


cronokirby (approved):
Nice catches!

Issue #2867, opened 2023-07-24
pd doesn't behave well under heavy sync load


conorsch:
Today on Testnet 56 we observed a large spike in client traffic to the `pd` endpoint at `https://grpc.testnet.penumbra.zone`. 

![active-client-conns-2](https://github.com/penumbra-zone/penumbra/assets/657862/7f92d54d-d95e-4e2e-8ca6-9d55bf5c5b44)

As for provenance of the traffic, let's assume it's organic interest, in the form of many people downloading the web extension and synchronizing blocks for the first time. After about an hour or two, memory consumption—in the `pd` container specifically—balloons to the point that OOMkiller kicks in and terminates the pod. An example of resource consumption shortly before kill:

```
❯ kubectl top pod penumbra-testnet-fn-0-54d9f68d76-kl9gm --containers
POD                                      NAME   CPU(cores)   MEMORY(bytes)   
penumbra-testnet-fn-0-54d9f68d76-kl9gm   pd     3055m        11249Mi         
penumbra-testnet-fn-0-54d9f68d76-kl9gm   tm     47m          162Mi 
```

According to the logs, `pd` is serving a lot of two types of requests, CompactBlockRange and ValidatorInfoStream:

```
❯ wc -l logs-penumbra-testnet-fn-0-54d9f68d76-kl9gm.txt
8329 logs-penumbra-testnet-fn-0-54d9f68d76-kl9gm.txt
❯ rg validator_info logs-penumbra-testnet-fn-0-54d9f68d76-kl9gm.txt -c
4639
❯ rg -i compact_block_range logs-penumbra-testnet-fn-0-54d9f68d76-kl9gm.txt -c
2107
```

Intriguingly both those types are `Box`ed return values in our RPCs. Also intriguing is this comment:

https://github.com/penumbra-zone/penumbra/blob/bfda3a85ba92b1d4112ebe402c0fb2a8e6271c60/crates/bin/pd/src/info/oblivious.rs#L332-L336

We need to understand why pd consumes large amounts of memory when handling these types of concurrent requests. For now, I'm assuming the traffic is well-formed, honest clients trying to synchronize.

hdevalence:
Maybe https://github.com/koute/bytehound would be useful here? I think a good goal would be that `pd` can maintain many active connections without using excess memory.

hdevalence:
[...]

PR #2868, opened 2023-07-25
fix(pd): drop unused snapshots

conorsch:
Trying to resolve a memory leak in pd. These manual drops are a first pass, and appear to reduce memory consumption, but there's still a leak, according to bytehound. We'll continue to investigate.

Includes a missed `to_proto` that's a nice to have, but likely doesn't constitute a fix for our problem.

Refs https://github.com/penumbra-zone/penumbra/issues/2867.

Used this script to generate loadtests against a local pd instance, for memory profiling efforts. Steps I used to test:

    pd testnet unsafe-reset-all
    pd testnet join https://rpc.testnet-preview.penumbra.zone
    # start pd & tendermint, wait for genesis to complete
    ./deployments/scripts/pcli-client-test -n 1000 -p 100

Refs #2867.

### DIFF
--- a/crates/bin/pd/src/info/oblivious.rs
+++ b/crates/bin/pd/src/info/oblivious.rs
@@ -207,6 +207,9 @@ impl ObliviousQueryService for Info {
             .await
             .map_err(|e| tonic::Status::unavailable(format!("error getting block height: {e}")))?;
 
+        // Perform housekeeping, so long-lived connections don't cause pd to leak memory.
+        std::mem::drop(snapshot);
+
         // Treat end_height = 0 as end_height = current_height so that if the
         // end_height is unspecified in the proto, it will be treated as a
         // request to sync up to the current height.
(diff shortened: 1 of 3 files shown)


crates/bin/pd/src/info/oblivious.rs
```diff
@@ -263,6 +263,8 @@ impl ObliviousQueryService for Info {
                         .await??
                         .expect("compact block for in-range height must be present");
                     tx.send(Ok(block.to_proto())).await?;
+                    // Don't let the block hang around while we serve other clients.
+                    std::mem::drop(block);
```
hdevalence, line 267: This isn't necessary, because the compact block doesn't have any shared references in it, and the explicit drop distracts from the case that is important (dropping the snapshot)

conorsch:
Cleaned up the commit history. After CI runs, I'm going to merge this, then cherry-pick the pd changes onto a release branch and tag as `v0.56.1`. As such, refs #2761. We still plan to release Testnet 57 tomorrow (#2827).

PR #2878, opened 2023-07-27
measure: add connection opener tool

hdevalence:


conorsch:
Thanks for writing this! I'll tinker with it today. On the face looks like it supersedes slapdash script in #2868. Might add some logging messages based on information I want to see in realtime. Refs #2867.

PR #2881, opened 2023-07-28
pd: use a range query for client synchronization

erwanor:
This PR:
- implements a `stream_compact_block` method to `compact_block::StateReadExt` 
- remove the `block_fetcher` tokio tasks in `ObliviousQueryService::compact_block_range`
- use a single range query to catch-up on existing `CompactBlock`s

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4674,6 +4674,7 @@ dependencies = [
  "blake2b_simd 0.5.11",
  "bytes",
  "decaf377-rdsa",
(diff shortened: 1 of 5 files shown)


erwanor:
Merging optimistically since the range query functionality is relatively well tested. The main unknown is whether the anticipated performance improvement will materialize or not. If it doesn't this PR will be easy to revert. This closes the loop on a series of `pd` improvements as a response to #2867.

PR #2883, opened 2023-07-28
ci: use buildjet cache for less flakiness

joroshiba:
Astria recently started using buildjet runners and ran into caching flakiness with the reliance on github cache, as alluded to [here](https://buildjet.com/for-github-actions/blog/launch-buildjet-cache#the-buildjet-cache-action). Created a fork of the rust cache that is built on top of `buildjet/cache` instead of `action/cache`, it's otherwise identical.

Figured I'd give you all the option to adopt, since I was inspired to use buildjet to speed up our builds by you.


hdevalence:
Awesome, thanks for sending this over!

PR #2885, opened 2023-08-01
storage: implement `StateRead::nonverifiable_range`

erwanor:
This PR adds support for range queries over the `nonverifiable` storage. It does so by:
- extending the `StateRead` trait with a new method:
```rust
    /// Retrieve all values for keys in a range from the non-verifiable key-value store, as raw bytes.
    /// This method does not support inclusive ranges, and will return an error if passed one.
    ///
    /// Users should generally prefer to use wrapper methods in an extension trait.
    fn nonverifiable_range_raw(
        &self,
        prefix: Option<&[u8]>,
        range: impl RangeBounds<Vec<u8>>,
    ) -> Result<Self::NonconsensusRangeRawStream>;
```
- use a rocksdb range iterator to create an underlying stream of keys and values
- implements a `StateDelta`-level stream that interleaves caching layers and underlying storage.

Example API:

The first intended use of range queries is to fetch a range of `compact_blocks` from an arbitrary height:

```rust
let start_height = compact_block::height(20); // "000000000000000020"
// Return a stream of compact blocks starting at height 20 to the latest one.
let mut compact_block_stream = state.nonverifiable_range_query(Some(compact_block::prefix()), start_height..);
```

Alternative APIs considered:

```rust
        fn nonverifiable_range_raw<T, U>(
        &self,
        prefix: Option<&[u8]>,
        range: T,
    ) -> anyhow::Result<Self::NonconsensusRangeRawStream>
    where
        T: std::ops::RangeBounds<U>,
        U: Into<Vec<u8>>;
```

and

```rust
    fn nonverifiable_range_raw<T, U>(
        &self,
        prefix: Option<&[u8]>,
        range: T,
    ) -> anyhow::Result<Self::NonconsensusRangeRawStream>
    where
        T: std::ops::RangeBounds<U>,
        U: AsRef<[u8]>;
```

either are fine choices, but my preference goes to the simpler signature.

### DIFF
--- a/crates/storage/src/delta.rs
+++ b/crates/storage/src/delta.rs
@@ -9,7 +9,7 @@ use crate::{
         CacheFuture, StateDeltaNonconsensusPrefixRawStream, StateDeltaNonconsensusRangeRawStream,
         StateDeltaPrefixKeysStream, StateDeltaPrefixRawStream,
     },
-    Cache, EscapedByteSlice, StateRead, StateWrite,
+    utils, Cache, EscapedByteSlice, StateRead, StateWrite,
 };
 
 /// An arbitrarily-deeply nested stack of delta updates to an underlying state.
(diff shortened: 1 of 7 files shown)


hdevalence:
Is there any reason not to merge this for the current testnet?

erwanor:
Not really, I left it open to invite reviews but merging now

PR #2892, opened 2023-08-02
ibc: use ibc-types for ibc component state paths

### LINKED ISSUE #2859: ibc: replace duplicated state_keys with IBC paths from ibc-types
Currently, the IBC component duplicates IBC paths, defining in its own state the canonical paths for IBC (for example, for packet commitments). Instead, it should use the paths defined in the `ibc-types` crate.

avahowell:
Closes #2859


conorsch:
Hurray for net negative diffs! Thanks, @avahowell !

Issue #2891, opened 2023-08-02
RPC changes for handling SwapClaims in the extension


conorsch:
Zpoken is currently working on adding support for Swap and SwapClaims to the wallet/dapp. They've requested two changes to our RPC interfaces:

- [x] In `TransactionPlannerRequest` replace `string memo = 3` with `MemoPlaintext memo`. This is necessary to allow dapp to hide or show the sender's address.  Alternatively, it can also be done with a combination of the fields `string memo` and `bool hide_sender`
- [ ] There is currently no way to build a `TransactionPlan` for a `SwapClaim` transaction through the planner interface. It would be cool to add `repeated TransactionPlannerRequest.SwapClaim`

Reported by the [Zpoken team in Discord](https://discord.com/channels/824484045370818580/979785365437165658/1135955908908822659).

conorsch:
We should absolutely make the changes requested, ideally by 2023-08-07 (#2888). But these changes may be necessary but not sufficient for Swap support; see for example https://buf.build/penumbra-zone/penumbra/docs/986b465b7fc4428d95d12d94f313c25f:penumbra.core.dex.v1alpha1#penumbra.core.dex.v1alpha1.SwapClaimPlan which requires an input of https://buf.build/penumbra-zone/penumbra/docs/986b465b7fc4428d95d12d94f313c25f:penumbra.core.dex.v1alpha1#penumbra.core.dex.v1alpha1.BatchSwapOutputData. How is the dapp expected to obtain the inputs for `SwapClaimPlan`?

conorsch:
> There is currently no way to build a TransactionPlan for a SwapClaim transaction through the planner interface. It would be cool to add repeated TransactionPlannerRequest.SwapClaim

I took a look around, and this is best understood as a documentation problem: it _is_ currently possible to build a TransactionPlan for a SwapClaim transaction, by leveraging the ActionPlan enum, which can be a `dex.v1alpaha1.SwapPlan`. I'll update the docstrings to make that clearer.

conorsch:
Ah, I see the problem: `TransactionPlan` supports swapclaims, but `TransactionPlannerRequest` does not. Will try to add that.

PR #2893, opened 2023-08-02
proto: use MemoPlaintext in TxPlannerRequest

conorsch:
We overlooked updating the memo field to its type. Fixed now. While I was in there, added some doc strings to make the interfaces abit more intuitive. Still need to follow up with deeper changes to support SwapClaims via the Planner interface; will consult with @aubrika on how best to do that.

Refs #2891.


conorsch:
These are breaking proto changes. To my eye, looks like the interchaintest work is largely unaffected, given its use of defaults, but the Zpoken folks will need to update. Given that they requested these changes, that's just fine.

PR #2899, opened 2023-08-03
client: refactor balances query

### LINKED ISSUE #2894: New Balances RPC request returns empty results
**Describe the bug**
The new (as of https://github.com/penumbra-zone/penumbra/pull/2799) RPC method [`penumbra.view.v1alpha1.BalancesRequest`](https://buf.build/penumbra-zone/penumbra/docs/main:penumbra.view.v1alpha1#penumbra.view.v1alpha1.BalancesRequest) returns empty responses, regardless of request params. 

**To Reproduce**

1. Configure and run `pclientd` locally. 
2. Run [grpcui](https://github.com/fullstorydev/grpcui) locally: `grpcui -plaintext localhost:8081`
3. Navigate to grpcui web ui, and make a request for `penumbra.view.v1alpha1.BalancesRequest`. 
4. Observe empty results.

**Expected behavior**
I expected to see balance info, meaning denoms and amounts, returned from the query. Instead, I saw nothing at all, which was disappointing. 

**Additional context**
You must run pclientd locally so there's an FVK available to parse response. If you try to use e.g. https://grpcui.testnet.penumbra.zone, the method will return empty, which is expected. (Maybe a custom response error code would be nice for that case.)

aubrika:
Closes #2894

### DIFF
--- a/crates/view/src/storage.rs
+++ b/crates/view/src/storage.rs
@@ -236,43 +236,60 @@ impl Storage {
         let pool = self.pool.clone();
 
         spawn_blocking(move || {
-            // If set, only return notes with the specified address index.
-            let address_clause = address_index
-            .map(|d| format!("x'{}'", hex::encode(d.to_bytes())))
-            .unwrap_or_else(|| "address_index".to_string());
+            let query = format!(
+                "SELECT notes.asset_id, notes.amount, spendable_notes.address_index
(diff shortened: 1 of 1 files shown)


conorsch:
Thanks, @aubrika! I will test this on a local pclientd instance prior to our IBC sync today.

PR #2903, opened 2023-08-06
Add log verification / searching for setup ceremony

### LINKED ISSUE #2870: Setup Ceremony: Allow searching logs for next tip to build on
Part of #712.

The idea is that the contribution logs will include not just valid contributions, but also potentially invalid contributions.

A contribution can be invalid in one of three ways:
1. The CRS elements of that contribution are inconsistent
2. The proof that these elements build on their claimed parent elements is invalid.
3. The elements don't build on the most recent valid contribution as their parent.

We need logic to take a transaction logic, and then use these conditions to find the most recent valid contribution, thus determining where we should build our own contribution.

We also want this logic to be resumable, making syncing easier, by allowing us to sync up a large portion of the log in advance, and then syncing smaller portions later without having to re-verify the head of the log.

cronokirby:
Fixes #2870.

This also adds an abstraction over the two phases, which was need to consolidate the common behavior they share for log verification.


redshiftzero (approved):
Looks great!

PR #2909, opened 2023-08-09
Reverse staking rate calculations

### LINKED ISSUE #1481: staking: reverse causality of rate calculations

Currently, the staking rate calculations are derived from the code that was originally used just to build the exchange rate mechanism and test that it works. As a result, the flow of causality is roughly: `fixed base rate` → `validator exchange rate` → (implicitly) change to size of validators' delegation pool.

However, this isn't what we actually want: we want to take the amount of new issuance for a delegation pool as the *given* input, then compute the validator exchange rate that prices in that change to the delegation pool.  This way, we can split apart two responsibilities: (1) the overall tokenomics for staking, which determines the new issuance for each delegation pool [i.e., some new code] and (2) the mechanism that actually does the accounting to track changes to the pool size [i.e., the validator rate mechanism]

For instance, this allows us to implement #1450, because we can (a) compute the new staking issuance for each delegation pool and then (b) add in the fees collected by that validator before (c) computing a validator exchange rate that prices in a+b.

It will also allow us to implement variable staking rewards, or variable splits of newly issued tokens between stakers or other streams, since those just change the "top" of the pipe.

- Create a new piece of code that determines the new staking issuance for each validator's delegation pool at the end of the epoch (to start, this can use a fixed inflation rate, like we do currently);
- Refactor the validator exchange rate computations introduced in #1480 to take an amount of new issuance as the input, and the validator exchange rate as the output.

### LINKED ISSUE #2802: Remove "next rates" APIs
In https://github.com/penumbra-zone/penumbra/issues/2630 we added and began using the current epoch's staking rates in the node and client, however the APIs for fetching the next epoch rates remained in place. This is an issue because they can no longer be predicted accurately; the API should be removed entirely.

plaidfinch:
Resolves #1481, resolves #2802. This happens in two steps:
- [x] Remove all remaining references to next-rate-data
- [x] In the staking module, determine the correct modification of the base rate data to effect a particular amount of inflation specified in penumbra by the distributions module
- [x] In the distributions module, determine the correct amount of inflation in penumbra based on the chain parameter specified base rate and send it to the staking module (this is temporary until we transition to dynamic distributions in a follow-up PR)

The formula to adjust the `previous_base_rate_data` to reflect a particular desired `issuance` is:

```rust
let total_active_stake = 
    todo!("sum over only active validators of `unbonded_stake` of their delegation amount");
const BPS_SQUARED: u64 = 1_0000_0000; // reward rate is measured in basis points squared
let upcoming_base_reward_rate = issuance * BPS_SQUARED / total_active_stake;
let upcoming_base_rate_data = previous_base_rate_data.next(upcoming_base_reward_rate);
```

Rewards are issued by inflating the stake bonded only to *active* validators, so to invert the staking calculation, we need to calculate a reward rate (in basis points of basis points) such that the upcoming base rate data when applied to the active validator set will inflate the value of their stake by the desired issuance. This works out to be that we should multiply the base rate by `(issuance + total_active_stake) / total_active_stake`, or in other words calculate the `next` base rate using a reward rate of `issuance * BPS_SQUARED / total_active_stake`.

Efficiently summing over all active validators requires an index tracking which validators are active. This could be done in this PR, or stubbed out with a filter over all validators and fixed up in #2921 to use the index of active validators defined there.

It is likely that the calculation of the new base rate data will result in a not-perfectly-precise creation of inflation (it will undershoot due to precision loss in division in several places). This can't be fixed in any given epoch, but to ensure that the total amount issued over time ends up being correct, we can carry forward to the next epoch a remainder to be fed forward into that epoch's issuance. This remainder can be calculated by summing the `unbonded_amount` across all active validators' _upcoming_ `RateData` (i.e. the new ones generated from the new `BaseRateData`) of each validator's entire delegation pool, and then subtracting that amount from the actual desired issuance. It is *not correct* to calculate a hypothetical `unbonded_amount` based only on the `BaseRateData`, because there are additional precision losses in the conversion to `RateData` for each validator. The sum should always be positive, because division rounds down. The remainder should be stored in the state — it should _not_ be fed forward into the next epoch's staking issuance _directly_, but rather should be treated as an _additional input to the distributions module_ (because this gives the distributions module freedom to re-allocate that issuance elsewhere in the upcoming epoch, which it might want to do).

Finally, the temporary determination of the correct issuance of staking rewards in the distributions module should be given by the product of the per-epoch staking reward rate set as a chain parameter, and the total value of all delegation tokens to any validator. This is, as noted above, temporary, and will be replaced with a per-epoch allocation that is scaled to epoch duration, and dynamically allocated based on other factors considered by the distributions module.


conorsch:
We're worried about drift making rebasing difficult. Let's aim to get this one in during 61.

hdevalence:
This won't be mergeable any more, following #3077, and the work will have to be redone on top of the current `main`. We should try to avoid leaving large PRs like this open for so long to prevent this in the future.

erwanor:
Picking this up in a cherry-picked branch.

PR #2910, opened 2023-08-09
fix: make `U128x128::round_up` fallible

redshiftzero:
Followup from one of the audits. 

Inside the DEX logic for the infallible methods that call `U128x128::round_up`, what should happen? (cc @erwanor)


erwanor (changes requested):
I need to check a couple assumptions, marking this as needing review to make sure it doesn't get merged accidentally before I can focus on this.

erwanor (approved):

PR #2916, opened 2023-08-11
feat(pd): support real addrs in testnet generate

conorsch:
Removed the default RFC1918 addresses from the `pd testnet generate` logic. Added two new options:

  * --peer-address-template
  * --external-addresses

The goal is to allow a node operator to pass in public IPv4 address information at the time of config generation, so the emitted config files are valid without further munging, particularly with respect to the `external_address` Tendermint advertises for its P2P service.

This change unblocks a leaner deployment pipeline, where we actually use the results of the `pd testnet generate` output, as intended.

Also: fix(pd) support urls in testnet join

We were inadvertently dropping paths on URLs by prepending `/` to `Url.join` args, which caused the join operation to be performed on the base domain, ignoring any path. We didn't notice this because all our services had been on their own subdomains.


conorsch:
I've done some manual testing of the deploy logic, against an ad-hoc devnet, to ensure that these changes will work fine with existing deployments. These changes unblock simplifying the CI logic, but those changes will come in separately, in order to keep things rolling smoothly, and to aid in review.

zbuc (approved):
LGTM

Issue #2877, opened 2023-07-27
Setup Ceremony: Implement benchmarks for contributions, log verification, etc.


cronokirby:
Part of #712.

One we implement these aspects, having benchmarks will be very useful to inform our decisions about the orchestration aspects of the ceremony.

cronokirby:
To note that one reason motivating this split is that Phase 1 is the most expensive, so having benchmarks for that sooner is already representative, and can be done before the linking code for phase 1 is ready, so no need to block on that.

PR #2922, opened 2023-08-11
zk ceremony phase 1 benchmarking

redshiftzero:
Towards #2877


cronokirby (changes requested):
I think we also want to benchmark the linking proof in verification as well, using the `is_linked_to` method. This probably won't add much to the cost of verification, but is part of it.

cronokirby (approved):
LGTM!

PR #2931, opened 2023-08-14
crypto: rework action-level `EffectingData` implementations

### LINKED ISSUE #2000: Change `EffectingData` implementations to hash the proto encoding of the body
Right now our `EffectingData` implementations are ad-hoc, and it would be easy to forget to bind an important field. Instead, we should hash the proto-encoding of the entire action body, for every action. This ensures that we will not ever fail to bind something necessary in an effecting hash.

redshiftzero:
Towards one of the checkboxes in #2808
Closes #2000

The scope of this PR is to in each `EffectingData` implementation hash a single proto that represents the effecting data of that transaction action.

### DIFF
--- a/crates/core/component/dao/src/action/dao_deposit.rs
+++ b/crates/core/component/dao/src/action/dao_deposit.rs
@@ -1,29 +1,14 @@
 use anyhow::{Context, Error};
-use prost::Message;
 use serde::{Deserialize, Serialize};
 use std::convert::{TryFrom, TryInto};
 
 use penumbra_asset::{Balance, Value};
-use penumbra_chain::{EffectHash, EffectingData};
 use penumbra_proto::{core::governance::v1alpha1 as pb, DomainType, TypeUrl};
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
 #[serde(try_from = "pb::DaoDeposit", into = "pb::DaoDeposit")]
 pub struct DaoDeposit {
     pub value: Value,
-}
-
-impl EffectingData for DaoDeposit {
-    fn effect_hash(&self) -> EffectHash {
-        let effecting_data: pb::DaoDeposit = self.clone().into();
-
-        let mut state = blake2b_simd::Params::default()
-            .personal(b"PAH:daodeposit")
-            .to_state();
-        state.update(&effecting_data.encode_to_vec());
-
-        EffectHash(*state.finalize().as_array())
-    }
 }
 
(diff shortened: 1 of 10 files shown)


crates/core/component/dao/src/action/dao_deposit.rs
```diff
@@ -1,25 +1,28 @@
...
 pub struct DaoDeposit {
     pub value: Value,
 }
 
 impl EffectingData for DaoDeposit {
     fn effect_hash(&self) -> EffectHash {
+        let effecting_data: pb::DaoDeposit = self.clone().into();
+
         let mut state = blake2b_simd::Params::default()
             .personal(b"PAH:daodeposit")
```
cronokirby, line 21: One thought I had here as I was reading this is that we might be able to remove the personalization strings. However, I don't think we can, because we might have a theoretical scenario where it's possible for two different protos to encode to the same bytes, and we'd want their hashes to be different if they're actually different things, to avoid shenanigans.

I don't think protobufs encode the "name" of the struct anywhere either, which would also avoid this need.

crates/core/component/dao/src/action/dao_deposit.rs
```diff
@@ -1,25 +1,28 @@
...
             .to_state();
+        state.update(&effecting_data.encode_to_vec());
 
-        state.update(&self.value.amount.to_le_bytes());
-        state.update(&self.value.asset_id.to_bytes());
-
-        EffectHash(state.finalize().as_array().clone())
+        EffectHash(*state.finalize().as_array())
```
cronokirby, line 25: Maybe we should add a helper for this flow of creating the state with personalization, updating, and then producing an effect hash

crates/core/component/dao/src/action/dao_deposit.rs
hdevalence: What if, instead of using a personalization string, we hashed the `TYPE_URL` provided by the [`TypeUrl` trait implemented for any domain type](https://rustdoc.penumbra.zone/main/penumbra_proto/trait.TypeUrl.html) and then the message encoding?

redshiftzero: this is a good idea! I've made that change in cdbfac7391fc15770f08e0b347962959c71a5be9, where we first hash the length of the variable-length `TYPE_URL`, then the `TYPE_URL`, and finally the corresponding message

cronokirby (approved):
LGTM

PR #2934, opened 2023-08-15
ibc: fix channel in recv_transfer_packet_inner

avahowell:
The channel to be used for value balance should be the source channel on the chain that originally sent the funds, i.e. channel B in recv_packet, not channel A.


conorsch:
Refs #2935, I will test post-merge to confirm the problem is resolved.

Issue #2284, opened 2023-03-30
Test IBC relaying between Penumbra and any Cosmos testnet


conorsch:
Part of #465. Let's attempt to relay between the Penumbra Pasiphae 49 (#2255) and the [Babylon](https://docs.babylonchain.io/docs/introduction/babylon-overview) testnet. Relevant docs:

* https://docs.babylonchain.io/docs/testnet/getting-funds
* https://github.com/babylonchain/networks/tree/main/bbn-test1

We'll need to implement a relayer config for the Babylon testnet, similar to the approach in #2166. Then we'll need to confirm that `rly transact link <path>` works without errors.

conorsch:
Some notes documenting the setup. At a high level, I created a keypair, documented its seed phrase in the usual place, tapped the Babylon testnet faucet for some test tokens, and confirmed the balance is reflected in my local wallet. 

```
# create keypair
$ babylond keys add test
# view pubkey
$ babylond keys list
- address: bbn18vpy897s8m4wm8wfmcw6llh4zmzdvxvd865tw9
  name: test
  pubkey: '{"@type":"/cosmos.crypto.secp256k1.PubKey","key":"A6yEPaXIyBK+TH3fT9Ez7jZMBypkVEDUE2lsBtzqIa6Z"}'
  type: local
# retrieve address
$ babylond keys list --output json | jq '.[0].address' -r
bbn18vpy897s8m4wm8wfmcw6llh4zmzdvxvd865tw9
# query balance
$ babylond query bank balances bbn18vpy897s8m4wm8wfmcw6llh4zmzdvxvd865tw9 --node http://rpc.testnet.babylonchain.io:26657
balances:
- amount: "100"
  denom: ubbn
pagination:
  next_key: null
  total: "0"
```

The `--node` argument is required because I'm not running a local node; the value defaults to `http://127.0.0.1:26657`. There's no corresponding env var, so it must be specified on the CLI for now.

conorsch:
The [relayer](https://github.com/cosmos/relayer) code we've been working with doesn't support Babylon. Instead, Babylon has its own version of the relayer: https://github.com/babylonchain/babylon-relayer Will try to get Penumbra support working in the latter, rather than Babylon support in the former.

hdevalence:
Ah, hmm, that might be a good signal to pause, then — I think we might be overrunning our headlights if we try to combine features from two different forks of the Go relayer.

conorsch:
[...]

PR #2938, opened 2023-08-15
docs: ibc pcli actions in guide

### LINKED ISSUE #2907: docs: ibc withdrawal workflow in the guide
We've landed a lot of IBC functionality lately, but we don't yet have example commands for moving funds between two test chains. Let's get that information added to the guide, alongside the other `pcli` actions. 

For these commands to work, we'll need to run the relayer between preview & testnet, and explain the need to open a separate terminal to sync with the preview chain. Fundamental actions should be covered:

* observe funds in primary chain
* create new wallet on preview chain
* check no funds in new wallet
* send funds via ibc to counterparty chain
* confirm receipt

That's tricky to document clearly, but we'll do our best.

conorsch:
Documents the rough-and-ready steps we've been using to exercise the developing IBC implementation. We only support preview/testnet for now, but we hope soon to add support for other, non-Penumbra testnet chains (#2284).

Closes #2907.

### DIFF
--- a/docs/guide/src/pcli/transaction.md
+++ b/docs/guide/src/pcli/transaction.md
@@ -152,7 +152,7 @@ This will handle generating the swap transaction and you'd soon have the market-
 in `gm` tokens returned to you, or the original investment of 1 `penumbra` tokens returned if there wasn't
 enough liquidity available to perform the swap.
 
-## Replicating a UniswapV2 (`x*y=k`) pool 
+## Replicating a UniswapV2 (`x*y=k`) pool
 
 Penumbra's constant-price pool is a versatile market primitive, allowing users extensive control over their trading strategies. It's not solely for active DEX quoters; with our AMM replication tool, users can emulate any passive AMM of their choice. The testnet comes with a built-in UniswapV2 replicator that is utilized as such:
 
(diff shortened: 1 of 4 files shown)


docs/guide/src/pcli/transaction.md
```diff
@@ -170,3 +170,65 @@ You will be prompted a disclaimer which you should read carefully, and accept or
...
+
+<!--
+N.B. These steps were written based on a local relayer setup, transferring packets
+between a local devnet and preview. The plan is to run a public relayer between
+testnet & preview again, starting with Testnet 59 Enceladus. We cannot run the relayer
+until 59, due to pcli incompatibilities between the branches.
+-->
+
+Penumbra aims to implement full IBC support for cross-chain asset transfers. For now, however,
+only withdrawals between Penumbra testnet chains are supported. To exercise this logic,
```
avahowell, line 184: Technically withdrawals to any testnet are supported, but we're just not running any relayers other than testnet <> preview

conorsch: Fair point! Overall, I'm struggling to balance the tone between what's technically true and what's comprehensible. Not sure I'm the best judge of either. :upside_down_face: I can rephrase to emphasize that the present restriction—which will change rapidly over time—is due to the operation of relayers.

conorsch: Addressed. Thanks for the review!

PR #2940, opened 2023-08-16
feat(proto)!: remove optional fields in protos

### LINKED ISSUE #2933: Remove optional fields throughout protos
The `protoc-gen-gocosmos` tool, used to generate golang code from protobuf definitions for many Cosmos projects, does not support the use of `optional` fields in proto3 syntax. We encountered this on https://github.com/strangelove-ventures/interchaintest/issues/644#issuecomment-1677750362. To unblock, let's drop use of optional fields, updating code where necessary.

conorsch:
The cosmos fork of gogoproto, used to generate golang code from BSR, doesn't support "optional" fields in proto3 syntax. Rather than sort that out, we're removing use of optional to unblock integration work on interchaintest.
    
In order to preserve the representation of optionality, we use submessages in the proto types, so we can include or omit the named field. A notable exception is for encrypted memos: since the encrypted memo payload must be a specific length, we can check for whether it's empty, and map an empty encrypted memo to None.
    
H/t to @plaidfinch for describing the submessage pattern to me, and implementing it on the governance changes.
    
Closes #2933.


proto/penumbra/penumbra/core/governance/v1alpha1/governance.proto
```diff
@@ -212,7 +218,12 @@ message Proposal {
   // It optionally contains a reference to a commit which contains code to upgrade the chain.
   message Signaling {
     // The commit to be voted upon, if any is relevant.
-    optional string commit = 1;
+    Commit commit = 1;
+  }
+
+  // The sha1 hash of a git commit, used to indicate a specific version of the software.
+  message Commit {
+      string commit = 1;
```
hdevalence, line 226: Alternatively, this could just be `string commit = 1` rather than a submessage, with the empty string meaning `None`.

proto/penumbra/penumbra/core/stake/v1alpha1/stake.proto
```diff
@@ -90,7 +90,14 @@ message BondingState {
...
   BondingStateEnum state = 1;
-  optional uint64 unbonding_epoch = 2;
+
+  message UnbondingEpoch {
+    // The epoch in which stake becomes unbonded.
+    uint64 epoch = 1;
+  }
+
+  // Present if needed to specify unbonding epoch.
+  UnbondingEpoch unbonding_epoch = 2;
```
hdevalence, line 100: We could also just have `uint64 unbonding_epoch = 2`, since `0` is not a valid unbonding epoch.

proto/penumbra/penumbra/view/v1alpha1/view.proto
```diff
@@ -128,8 +128,8 @@ message TransactionPlannerRequest {
     // The memo for the requested TransactionPlan.
     // The memo must be unspecified unless `outputs` is nonempty.
     core.transaction.v1alpha1.MemoPlaintext memo = 3;
-    // Identifies the account group to query.
-    optional core.crypto.v1alpha1.AccountGroupId account_group_id = 14;
+    // Identifies the account group to query. Defaults to 0.
+    core.crypto.v1alpha1.AccountGroupId account_group_id = 14;
```
hdevalence, line 132: 0 isn't a valid account group ID, right? So this comment should read
```
// If present, identifies the account group to query.
```
?

hdevalence: Or we could just drop these entirely, since we don't use them at all and have no concrete plans to do multi-account view servers.

proto/penumbra/penumbra/view/v1alpha1/view.proto
```diff
@@ -418,15 +418,23 @@ message TransactionInfoByHashRequest {
...
+  // Submessage to represent optionality for specifying heights.
+  message BlockHeight {
+      uint64 height = 1;
+  }
   // If present, return only transactions after this height.
-  optional uint64 start_height = 1;
+  BlockHeight start_height = 1;
   // If present, return only transactions before this height.
-  optional uint64 end_height = 2;
+  BlockHeight end_height = 2;
```
hdevalence, line 428: Rather than defining a submessage, these should just be `uint64 start_height` and `uint64 end_height`, since 0 is not a valid block height.

proto/penumbra/penumbra/view/v1alpha1/view.proto
```diff
@@ -418,15 +418,23 @@ message TransactionInfoByHashRequest {
...
 }
 
 message TransactionInfo {
+    // Submessage to represent optionality for block height.
+    message BlockHeight {
+        uint64 height = 1;
+    }
     // The height the transaction was included in a block, if known.
-    optional uint64 height = 1;
+    BlockHeight height = 1;
```
hdevalence, line 437: ditto above, this should just be a `uint64`.

proto/penumbra/penumbra/view/v1alpha1/view.proto
```diff
@@ -467,28 +481,35 @@ message SpendableNoteRecord {
   // The height at which the note was created.
   uint64 height_created = 5;
   // Records whether the note was spent (and if so, at what height).
-  optional uint64 height_spent = 6;
+  // Present if the note was spent, otherwise absent.
+  BlockHeight height_spent = 6;
```
hdevalence, line 485: ditto above, this should just be a `uint64`.

proto/penumbra/penumbra/view/v1alpha1/view.proto
```diff
@@ -467,28 +481,35 @@ message SpendableNoteRecord {
...
+  }
   core.crypto.v1alpha1.StateCommitment swap_commitment = 1;
   core.dex.v1alpha1.SwapPlaintext swap = 2;
   uint64 position = 3;
   core.crypto.v1alpha1.Nullifier nullifier = 4;
   core.dex.v1alpha1.BatchSwapOutputData output_data = 5;
-  optional uint64 height_claimed = 6;
+
+  // If present, height at which Swap was claimed.
+  BlockHeight height_claimed = 6;
```
hdevalence, line 504: ditto above, this should just be a `uint64`.

hdevalence:
This PR changes a lot of scalar fields to have submessages, but in almost every case it would be simpler and easier (as well as less work, for ourselves making this change as well as all downstream users) to just use ordinary values.

conorsch:
> This PR changes a lot of scalar fields to have submessages, but in almost every case it would be simpler and easier (as well as less work, for ourselves making this change as well as all downstream users) to just use ordinary values.

Great feedback, thanks for thorough review. I've opened a separate PR, https://github.com/penumbra-zone/penumbra/pull/2951, with a fresh implementation as you describe.

conorsch:
Superseded by #2951.

PR #2942, opened 2023-08-16
ibc: decrement token supply tracking on ics20 withdrawal

avahowell:


conorsch:
Merging optimistically, so this change makes it into the next testnet.

Issue #2597, opened 2023-05-17
Add swap claims to `TransactionPlannerRequest`


hdevalence:
**Is your feature request related to a problem? Please describe.**

The `TransactionPlannerRequest` allows planning transactions with swaps, but doesn't allow planning swap claims.  Swap claims are designed so that they can be submitted autonomously with no user interaction, but we might not want to assume that the view server is the part of the software stack that does so. In any case, it seems useful to allow planning of swap claims, too.

**Describe the solution you'd like**

- [ ] Add swap claims to the planner request
- [ ] Test that it works by adding a network integration test to `pclientd` that makes a swap and then claims the outputs, using only GRPC methods and not any of the Rust libraries.

conorsch:
Looks like I filed a duplicate issue in #2891. I didn't take the time to add an integration test, as specified here.

zbuc:
It looks like the actual view service implementation of swaps and swap claims in transaction planner request handling wasn't implemented yet either.

PR #2947, opened 2023-08-17
Support swap/swap claim in TransactionPlannerRequest

zbuc:
Part 1 of #2597, missing integration tests


conorsch:
Merging this so that the breaking proto changes in #2940 can be rebased on top of it.

PR #2954, opened 2023-08-22
ibc: add grpc queries for connections, channels, and clients 

avahowell:


avahowell:
All of the previously implemented ibc queries are implemented now, this just needs to be hooked up to the grpc api

avahowell:
Hooked up the grpc api

hdevalence:
Looks like this is just blocked on fixing build flags for wasm?

conorsch:
Discussed at sprint planning today: we need to land these necessary-but-not-sufficient queries to unblock a Hermes relaying spike. However, further queries are required. For Testnet 60 https://github.com/penumbra-zone/penumbra/issues/2984, we won't have full support merged, but we'll press on with support.

crates/core/component/ibc/src/component/grpc_query.rs
```diff
@@ -0,0 +1,637 @@
...
+
+        Ok(tonic::Response::new(res))
+    }
+    /// ClientConnections queries the connection paths associated with a client
+    /// state.
+    async fn client_connections(
+        &self,
+        _request: tonic::Request<QueryClientConnectionsRequest>,
+    ) -> std::result::Result<tonic::Response<QueryClientConnectionsResponse>, tonic::Status> {
+        todo!()
```
hdevalence, line 150: This will make the node panic if we get one of these requests, can we search/replace these `todo!()` impls with a snippet that returns an `Ok` with [`Status::unimplemented`](https://docs.rs/tonic/latest/tonic/struct.Status.html#method.unimplemented) ?

hdevalence (changes requested):
Let's avoid the `todo!()`s but otherwise this LGTM!

crates/bin/pd/src/main.rs
```diff
@@ -341,6 +346,9 @@ async fn main() -> anyhow::Result<()> {
                 .add_service(tonic_web::enable(SpecificQueryServiceServer::new(
                     info.clone(),
                 )))
+                .add_service(tonic_web::enable(ClientQueryServer::new(ibc.clone())))
+                .add_service(tonic_web::enable(ChannelQueryServer::new(ibc.clone())))
+                .add_service(tonic_web::enable(ConnectionQueryServer::new(ibc.clone())))
```
hdevalence, line 351: Non-blocking: going forward, we want to split up the RPCs and move them into the component crates. When we do that, we'll have LOTS of different services to stitch together, and we should figure out how to cleanly add them all. Maybe `tower::Layer`s would be helpful there.

conorsch:
We want this in for 60 (#2984). @avahowell is aiming for merge by EOD.

PR #2957, opened 2023-08-25
governance: create upgrade messages

erwanor:
This PR adds an `UpgradePlan` proposal type that specifies a height for an upgrade. If an upgrade is ratified, then the chain will trigger an epoch change at that height and gracefully shutdown.


erwanor:
I am going to add the actual mechanism for triggering a chain halt / epoch change in a follow-up PR. The rough idea is that we will write a `height: u64` to a `chain/next_upgrade_height` key in nonverifiable storage, compute a predicate during the commit phase (`should_chain_upgrade = current_height.sat_add(1) == upgrade_height`), and exit if that's the case; this is much like how emergency halts work.

This PR does not add any stateless/stateful check for the `UpgradePlan` handling.

PR #2959, opened 2023-08-26
chain: implement upgrade halts

erwanor:
This PR implement chain halts when the upgrade height specified in an `UpgradePlan` has been reached. It does this in three steps:
- when an `UpgradePlan` is enacted, we write the scheduled upgrade height in nonverifiable storage at key `chain/next_upgrade_height`
- when the target height is reached:
    - trigger an early epoch change so that the upgraded chain starts at a clean epoch boundary
    - after committing, shutdown gracefully


erwanor:
There has been some minor additions since review, mainly:
- a small fix to the logic fetching spendable note records for voting
- re-using the signal mechanism to force a binary upgrade when after an upgrade halt is complete.

Issue #2946, opened 2023-08-17
BIP44 support


redshiftzero:
In #419 (corresponding PR #443) we modified the `SpendSeed` generation to be derived deterministically from a `SeedPhrase`, instead of randomly generated. When we did this, we followed [BIP39](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki) which is a standard for generating a seed from a mnemonic phrase. However, for hardware wallet support, we need BIP44 support, which is the default derivation mode [used by Ledger Live](https://developers.ledger.com/docs/coin/address-derivation/). BIP44 is a standard that defines an organizational hierarchy allowing the user to remember a single seed phrase for multiple cryptocurrencies. To support this we need to:

# Add experimental BIP44 support

- [x] Add BIP44 support in `penumbra-keys`, using the [BIP44 path](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki) (`m / purpose' / coin_type' / account' / change / address_index`).
- [x] Add it in `pcli` with an optional flag e.g. `--bip44-derivation` to allow for testing
- [x] Add a Penumbra `coin_type` to the list of registered coins: https://github.com/satoshilabs/slips/blob/master/slip-0044.md - in https://github.com/satoshilabs/slips/pull/1592

# Make BIP44 derivation default

- [x] Update documentation in protocol spec
- [x] Determine if changes are needed on the web extension side (one change needed: https://github.com/penumbra-zone/penumbra/blob/f482b485ad265511e2040fa2d9090c5bc04273c7/crates/wasm/src/keys.rs#L16C17-L16C17)
- [x] Make above change to web extension
- [x] Make this BIP44 derivation the default in `pcli` and `pclientd`, and reserve the old BIP39 derivation in `pcli` with an optional flag `--legacy-raw-bip39-derivation`
- [x] Replace test account (and its allocations) with one that uses BIP44 derivation using the seed phrase in `bip44_test_ledger` which has addresses as below:

```rust
pub const ADDRESS_0_STR: &str = "penumbrav2t147mfall0zr6am5r45qkwht7xqqrdsp50czde7empv7yq2nk3z8yyfh9k9520ddgswkmzar22vhz9dwtuem7uxw0qytfpv7lk3q9dp8ccaw2fn5c838rfackazmgf3ahhvhypxd";
[...]

PR #2964, opened 2023-09-01
crypto: BIP44 derivation

redshiftzero:
Towards #2946. 

This adds experimental BIP44 derivation wherein we:
1. derive a 64-byte binary seed from the seed phrase as described in BIP39 using PBKDF2 but using only `mnemonic` as the salt
2. define a BIP44 path `m/44'/6532'/0'` which matches the Penumbra `coin_type` and the account number (`0`) used in the test vector generator (the apostrophes indicate hardened derivation)
3. derive the spend key bytes from BIP32 child key derivation from the seed and the above BIP44 path

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -368,17 +368,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "ark-secp256k1"
-version = "0.4.0"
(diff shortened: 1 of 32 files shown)


cronokirby (approved):
A few minor nits, looks good to me otherwise!

PR #2973, opened 2023-09-06
Improvements and fixes to wasm transaction planner and wasm view server

zpoken:
- added wasm planner which allows more flexible creation of TransactionPlan on TS side 
- added logic of storing and reading advice when scanning blocks
- added function for generating ephemeral address 
- planner reads data from indexedDB directly from rust

### DIFF
--- a/crates/wasm/Cargo.toml
+++ b/crates/wasm/Cargo.toml
@@ -29,15 +29,19 @@ penumbra-tct           = { path = "../crypto/tct" }
 penumbra-transaction   = { path = "../core/transaction", default-features = false }
 
 anyhow                   = "1.0.75"
+ark-ff                   = { version = "0.4.2", features = ["std"] }
 base64                   = "0.21.2"
 console_error_panic_hook = { version = "0.1.7", optional = true }
+decaf377                 = { version = "0.5", features = ["r1cs"] }
 hex                      = "0.4.3"
 indexed_db_futures       = "0.3.0"
 rand_core                = { version = "0.6.4", features = ["getrandom"] }
 serde                    = { version = "1.0.186", features = ["derive"] }
 serde-wasm-bindgen       = "0.5.0"
+thiserror                = "1.0"
 wasm-bindgen             = "0.2.87"
 wasm-bindgen-futures     = "0.4.37"
+wasm-bindgen-test        = "0.3.0"
 web-sys                  = { version = "0.3.64", features = ["console"] }
 
 [dev-dependencies]
--- a/crates/wasm/src/error.rs
+++ b/crates/wasm/src/error.rs
@@ -1,0 +1,69 @@
+use base64::DecodeError;
+use hex::FromHexError;
+use penumbra_tct::error::{InsertBlockError, InsertEpochError, InsertError};
+use serde_wasm_bindgen::Error;
+use std::convert::Infallible;
+use thiserror::Error;
+use wasm_bindgen::{JsError, JsValue};
+use web_sys::DomException;
+
+pub type WasmResult<T> = Result<T, WasmError>;
+
+#[derive(Error, Debug)]
+pub enum WasmError {
+    #[error("{0}")]
+    Anyhow(#[from] anyhow::Error),
+
+    #[error("{0}")]
+    DecodeError(#[from] DecodeError),
+
+    #[error("{0}")]
+    Dom(#[from] DomError),
+
+    #[error("{0}")]
+    FromHexError(#[from] FromHexError),
+
+    #[error("{0}")]
+    Infallible(#[from] Infallible),
+
+    #[error("{0}")]
+    InsertBlockError(#[from] InsertBlockError),
+
+    #[error("{0}")]
+    InsertEpochError(#[from] InsertEpochError),
+
+    #[error("{0}")]
+    InsertError(#[from] InsertError),
+
+    #[error("{0}")]
+    Wasm(#[from] serde_wasm_bindgen::Error),
+}
+
+impl From<WasmError> for serde_wasm_bindgen::Error {
+    fn from(wasm_err: WasmError) -> Self {
+        Error::new(wasm_err.to_string())
+    }
+}
+
+impl From<WasmError> for JsValue {
+    fn from(error: WasmError) -> Self {
+        JsError::from(error).into()
+    }
+}
+
+#[derive(Debug)]
+pub struct DomError(DomException);
+
+impl std::fmt::Display for DomError {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "DOM Exception: {:?}", self.0)
+    }
+}
+
+impl std::error::Error for DomError {}
+
+impl From<DomException> for WasmError {
+    fn from(dom_exception: DomException) -> Self {
+        WasmError::Dom(DomError(dom_exception))
+    }
+}
--- a/crates/wasm/src/keys.rs
+++ b/crates/wasm/src/keys.rs
@@ -1,0 +1,106 @@
+use crate::error::WasmResult;
+use penumbra_keys::keys::{SeedPhrase, SpendKey};
+use penumbra_keys::{Address, FullViewingKey};
+use penumbra_proto::{core::crypto::v1alpha1 as pb, serializers::bech32str, DomainType};
+use rand_core::OsRng;
+use std::str::FromStr;
+use wasm_bindgen::prelude::*;
+
+/// generate a spend key from a seed phrase
+/// Arguments:
+///     seed_phrase: `string`
+/// Returns: `bech32 string`
+#[wasm_bindgen]
+pub fn generate_spend_key(seed_phrase: &str) -> WasmResult<JsValue> {
+    let seed = SeedPhrase::from_str(seed_phrase)?;
+    let spend_key = SpendKey::from_seed_phrase_bip39(seed, 0);
+
+    let proto = spend_key.to_proto();
+
+    let spend_key_str = bech32str::encode(
+        &proto.inner,
+        bech32str::spend_key::BECH32_PREFIX,
+        bech32str::Bech32m,
+    );
+
+    Ok(JsValue::from_str(&spend_key_str))
+}
+
+/// get full viewing key from spend key
+/// Arguments:
+///     spend_key_str: `bech32 string`
+/// Returns: `bech32 string`
+#[wasm_bindgen]
+pub fn get_full_viewing_key(spend_key: &str) -> WasmResult<JsValue> {
+    let spend_key = SpendKey::from_str(spend_key)?;
+
+    let fvk: &FullViewingKey = spend_key.full_viewing_key();
+
+    let proto = fvk.to_proto();
+
+    let fvk_bech32 = bech32str::encode(
+        &proto.inner,
+        bech32str::full_viewing_key::BECH32_PREFIX,
+        bech32str::Bech32m,
+    );
+    Ok(JsValue::from_str(&fvk_bech32))
+}
+
+/// get address by index using FVK
+/// Arguments:
+///     full_viewing_key: `bech32 string`
+///     index: `u32`
+/// Returns: `pb::Address`
+#[wasm_bindgen]
+pub fn get_address_by_index(full_viewing_key: &str, index: u32) -> WasmResult<JsValue> {
+    let fvk = FullViewingKey::from_str(full_viewing_key)?;
+    let (address, _dtk) = fvk.incoming().payment_address(index.into());
+    let proto = address.to_proto();
+    let result = serde_wasm_bindgen::to_value(&proto)?;
+    Ok(result)
+}
+
+/// get ephemeral (randomizer) address using FVK
+/// The derivation tree is like "spend key / address index / ephemeral address" so we must also pass index as an argument
+/// Arguments:
+///     full_viewing_key: `bech32 string`
+///     index: `u32`
+/// Returns: `pb::Address`
+#[wasm_bindgen]
+pub fn get_ephemeral_address(full_viewing_key: &str, index: u32) -> WasmResult<JsValue> {
+    let fvk = FullViewingKey::from_str(full_viewing_key)?;
+    let (address, _dtk) = fvk.ephemeral_address(OsRng, index.into());
+    let proto = address.to_proto();
+    let result = serde_wasm_bindgen::to_value(&proto)?;
+    Ok(result)
+}
+
+/// Check if the address is FVK controlled
+/// Arguments:
+///     full_viewing_key: `bech32 String`
+///     address: `bech32 String`
+/// Returns: `Option<pb::AddressIndex>`
+#[wasm_bindgen]
+pub fn is_controlled_address(full_viewing_key: &str, address: &str) -> WasmResult<JsValue> {
+    let fvk = FullViewingKey::from_str(full_viewing_key)?;
+    let index: Option<pb::AddressIndex> = fvk
+        .address_index(&Address::from_str(address)?)
+        .map(Into::into);
+    let result = serde_wasm_bindgen::to_value(&index)?;
+    Ok(result)
+}
+
+/// Get canonical short form address by index
+/// This feature is probably redundant and will be removed from wasm in the future
+/// Arguments:
+///     full_viewing_key: `bech32 string`
+///     index: `u32`
+/// Returns: `String`
+#[wasm_bindgen]
+pub fn get_short_address_by_index(full_viewing_key: &str, index: u32) -> WasmResult<JsValue> {
+    let fvk = FullViewingKey::from_str(full_viewing_key)?;
+
+    let (address, _dtk) = fvk.incoming().payment_address(index.into());
+    let short_address = address.display_short_form();
+    Ok(JsValue::from_str(&short_address))
+}
--- a/crates/wasm/src/lib.rs
+++ b/crates/wasm/src/lib.rs
@@ -1,131 +1,15 @@
-#![deny(clippy::unwrap_used)]
 #![allow(dead_code)]
 extern crate core;
 
+mod error;
+mod keys;
 mod note_record;
 mod planner;
+mod storage;
 mod swap_record;
 mod tx;
 mod utils;
 mod view_server;
-use penumbra_proto::{core::crypto::v1alpha1 as pb, serializers::bech32str, DomainType};
+mod wasm_planner;
 
-use penumbra_keys::{Address, FullViewingKey};
-use std::convert::TryFrom;
-use std::str::FromStr;
-
-use penumbra_keys::keys::{SeedPhrase, SpendKey};
-use wasm_bindgen::prelude::*;
-
-use penumbra_transaction::Transaction;
-
-pub use tx::send_plan;
 pub use view_server::ViewServer;
-
-#[wasm_bindgen]
-pub fn generate_spend_key(seed_phrase: &str) -> JsValue {
-    utils::set_panic_hook();
-    let seed =
-        SeedPhrase::from_str(seed_phrase).expect("the provided string is a valid seed phrase");
-    let spend_key = SpendKey::from_seed_phrase_bip39(seed, 0);
-
-    let proto = spend_key.to_proto();
-    let spend_key_str = &bech32str::encode(
-        &proto.inner,
-        bech32str::spend_key::BECH32_PREFIX,
-        bech32str::Bech32m,
-    );
-
-    serde_wasm_bindgen::to_value(&spend_key_str).expect("able to serialize spend key")
-}
-
-#[wasm_bindgen]
-pub fn get_full_viewing_key(spend_key_str: &str) -> JsValue {
-    utils::set_panic_hook();
-    let spend_key =
-        SpendKey::from_str(spend_key_str).expect("the provided string is a valid spend key");
-
-    let fvk: &FullViewingKey = spend_key.full_viewing_key();
-
-    let proto = pb::FullViewingKey::from(fvk.to_proto());
-
-    let fvk_str = &bech32str::encode(
-        &proto.inner,
-        bech32str::full_viewing_key::BECH32_PREFIX,
-        bech32str::Bech32m,
-    );
-    serde_wasm_bindgen::to_value(&fvk_str).expect("able to serialize full viewing key")
-}
-
-#[wasm_bindgen]
-pub fn get_address_by_index(full_viewing_key: &str, index: u32) -> JsValue {
-    utils::set_panic_hook();
-    let fvk = FullViewingKey::from_str(full_viewing_key.as_ref())
-        .expect("the provided string is a valid FullViewingKey");
-
-    let (address, _dtk) = fvk.incoming().payment_address(index.into());
-
-    let proto = address.to_proto();
-    let address_str = &bech32str::encode(
-        &proto.inner,
-        bech32str::address::BECH32_PREFIX,
-        bech32str::Bech32m,
-    );
-
-    serde_wasm_bindgen::to_value(&address_str).expect("able to serialize address")
-}
-
-#[wasm_bindgen]
-pub fn base64_to_bech32(prefix: &str, base64_str: &str) -> JsValue {
-    utils::set_panic_hook();
-
-    let bech32 = &bech32str::encode(
-        &base64::Engine::decode(&base64::engine::general_purpose::STANDARD, base64_str)
-            .expect("the provided string is a valid base64 string"),
-        prefix,
-        bech32str::Bech32m,
-    );
-    serde_wasm_bindgen::to_value(bech32).expect("able to serialize bech32 string")
-}
-#[wasm_bindgen]
-pub fn is_controlled_address(full_viewing_key: &str, address: &str) -> JsValue {
-    utils::set_panic_hook();
-    let fvk = FullViewingKey::from_str(full_viewing_key.as_ref())
-        .expect("the provided string is a valid FullViewingKey");
-
-    let index = fvk.address_index(&Address::from_str(address.as_ref()).expect("valid address"));
-
-    serde_wasm_bindgen::to_value(&index).expect("able to serialize address index")
-}
-
-#[wasm_bindgen]
-pub fn get_short_address_by_index(full_viewing_key: &str, index: u32) -> JsValue {
-    utils::set_panic_hook();
-    let fvk = FullViewingKey::from_str(full_viewing_key.as_ref())
-        .expect("The provided string is not a valid FullViewingKey");
-
-    let (address, _dtk) = fvk.incoming().payment_address(index.into());
-    let short_address = address.display_short_form();
-    serde_wasm_bindgen::to_value(&short_address).expect("able to serialize address")
-}
-
-#[wasm_bindgen]
-pub fn decode_transaction(tx_bytes: &str) -> JsValue {
-    utils::set_panic_hook();
-    let tx_vec: Vec<u8> =
-        base64::Engine::decode(&base64::engine::general_purpose::STANDARD, tx_bytes)
-            .expect("the provided tx string is a valid base64 string");
-    let transaction: Transaction =
-        Transaction::try_from(tx_vec).expect("the provided tx string is a valid transaction");
-    serde_wasm_bindgen::to_value(&transaction).expect("able to serialize transaction")
-}
-
-#[wasm_bindgen]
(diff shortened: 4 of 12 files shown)


crates/wasm/Cargo.toml
```diff
@@ -39,6 +39,10 @@ serde-wasm-bindgen       = "0.5.0"
 wasm-bindgen             = "0.2.87"
 wasm-bindgen-futures     = "0.4.37"
 web-sys                  = { version = "0.3.64", features = ["console"] }
+tracing = "0.1"
```
grod220, line 42: Unused?

crates/wasm/src/lib.rs
```diff
@@ -74,6 +77,20 @@ pub fn get_address_by_index(full_viewing_key: &str, index: u32) -> JsValue {
...
 }
 
+#[wasm_bindgen]
+pub fn get_ephemeral_address(full_viewing_key: &str, index: u32) -> JsValue {
+    utils::set_panic_hook();
+    let fvk = FullViewingKey::from_str(full_viewing_key.as_ref())
+        .context("The provided string is not a valid FullViewingKey")
+        .unwrap();
+
+    let (address, _dtk) = fvk.ephemeral_address(OsRng, index.into());
```
grod220, line 87: How does `OsRng` work in the web context? Does it just work?

crates/wasm/src/lib.rs
```diff
@@ -74,6 +77,20 @@ pub fn get_address_by_index(full_viewing_key: &str, index: u32) -> JsValue {
     return serde_wasm_bindgen::to_value(&address_str).unwrap();
 }
 
+#[wasm_bindgen]
+pub fn get_ephemeral_address(full_viewing_key: &str, index: u32) -> JsValue {
```
grod220, line 81: `JsValue` is quite painful trying to understand what comes back from this function. We also need to start adding docstrings above every imported function so it helps the frontend more. Something like:

```rust
/// Generates a kind of "burner" address in which the user can use to maintain their anonymity without sharing their main address.
/// Returns: `u128`
#[wasm_bindgen]
pub fn get_ephemeral_address(full_viewing_key: &str, index: u32) -> JsValue {
```

and this would work for parameters too:

```rust
/// Scans block for notes, swaps...
/// Arguments: 
///     compact_block: `v1alpha1::CompactBlock`
/// Returns: `ScanBlockResult`
#[wasm_bindgen]
    pub fn scan_block_without_updates(&mut self, compact_block: JsValue) -> JsValue {
```

crates/wasm/src/planner.rs
```diff
@@ -69,44 +82,41 @@ impl<R: RngCore + CryptoRng> Planner<R> {
...
+                        address_index: None,
+                        ..Default::default()
+                    },
+                )
+                .collect(),
+        )
+    }
 
     /// Set the expiry height for the transaction plan.
+    #[instrument(skip(self))]
```
grod220, line 119: What is this and what is it used for?

crates/wasm/src/planner.rs
```diff
@@ -341,22 +392,55 @@ impl<R: RngCore + CryptoRng> Planner<R> {
...
+    //     // Plan the transaction using the gathered information
+    //
+    //     self.plan_with_spendable_and_votable_notes(
+    //         &chain_params,
+    //         &fmd_params,
+    //         spendable_notes,
+    //         voting_notes,
+    //         self_address,
+    //     )
+    // }
```
grod220, line 418: Delete?

crates/wasm/src/view_server.rs
```diff
@@ -625,3 +533,42 @@ pub fn load_tree(stored_tree: StoredTree) -> Tree {
...
     return tree;
 }
+
+pub async fn store_advice(note: Note) {
+    let db_req: OpenDbRequest = IdbDatabase::open_u32("penumbra", 12).ok().unwrap();
+
+    let db: IdbDatabase = db_req.into_future().await.ok().unwrap();
+
+    let tx = db.transaction_on_one("notes").ok().unwrap();
+    let store = tx.object_store("notes").ok().unwrap();
```
grod220, line 543: I noticed we use a lot of unwrap()'s. Also these patterns:

```rust
.ok()?.await.ok()?;
.ok().unwrap()
```

In general, we need to move to the standard `Result` pattern and not use unwraps unless they are right as we coerce them to a `JsValue` and encode them for javascript.

Also, doing `.ok()` on everything has downsides:
- We lose error information, making debugging harder
- `?` is usually used on Result types. Using it on an Option is confusing.
- This will cause a panic if a None is received.
- We are treating Options as Results.

crates/wasm/src/wasm_planner.rs
```diff
@@ -0,0 +1,225 @@
...
+        self.planner.swap_claim(swap_claim_plan);
+        Ok(())
+    }
+
+    pub fn swap(
+        &mut self,
+        input_value: JsValue,
+        into_denom: JsValue,
+        swap_claim_fee: JsValue,
+        claim_address: JsValue,
```
grod220, line 95: Can't these be?:

```rust
input_value: u128,
into_denom: &str,
swap_claim_fee: u128,
claim_address: &str,
```

Whenever possible, we should not do JsValue if it is allowed. You may have to test this.

crates/wasm/src/wasm_planner.rs
```diff
@@ -0,0 +1,225 @@
...
+        let claim_address_proto: Address = serde_wasm_bindgen::from_value(claim_address)?;
+
+        let _ = self.planner.swap(
+            input_value_proto.try_into().unwrap(),
+            into_denom_proto.try_into().unwrap(),
+            swap_claim_fee_proto.try_into().unwrap(),
+            claim_address_proto.try_into().unwrap(),
+        );
+
+        Ok(())
```
grod220, line 109: What's the purpose of this function?

crates/wasm/src/wasm_planner.rs
```diff
@@ -0,0 +1,225 @@
...
+            .unwrap();
+
+        let plan_proto: TransactionPlan = plan.to_proto();
+
+        Ok(serde_wasm_bindgen::to_value(&plan_proto)?)
+    }
+}
+
+pub async fn get_chain_parameters() -> Option<ChainParameters> {
+    let db_req: OpenDbRequest = IdbDatabase::open_u32("penumbra", 12).ok()?;
```
grod220, line 149: I think we should have a `IndexDb` struct that has all of these helper functions in one place in another file.

grod220 (commented):
Please also run `rustfmt` and `clippy` on the wasm crate 🙏

crates/wasm/src/lib.rs
hdevalence: Yeah, I believe the crate has compile-time features that hook the platform APIs.

conorsch:
Discussed during sprint planning meeting today. We're currently planning to release in one business day, on Monday, #2984. Is this PR QOL, meaning it's OK to let slip for Monday? @grod220 will ping Zpoken folks to get a sense of urgency. It's technically still possible to dogpile on this and get it in for Monday if necessary.

crates/wasm/src/view_server.rs
Valentine1898: In most cases we will have to use a rather massive construction with `map_err`
` .map_err(|err| JsError::new(&err.to_string()))?`
On this [issue](https://github.com/rustwasm/wasm-bindgen/issues/1742) , there is no more ergonomic solution yet

crates/wasm/src/wasm_planner.rs
Valentine1898: This is part of the functionality for flexible TransactionPlan creation
An example of use can be found [here](https://github.com/penumbra-zone/wallet/blob/dev/src/services/ViewProtocolService.ts#L279C4-L279C4)
With this function we add Swap actions to planner

grod220:
Take a look at this example. At the top level, all `#[wasm_bindgen]` functions need to return  `Result<JsValue, Error>` if consuming functions that return results or just `JsValue` if not using any functions that that return results (should be very rare). 

However, every other function in this crate should be using the `Result` type (in my example the custom `WasmResult`). This example uses `thiserror` crate to handle the mapping of all possible errors in the code. This means, error messages are retained, you can use the succinct `?` without mapping errors, and the code is responsible for result handling like normal Rust code.

```rust

use std::str::FromStr;

use indexed_db_futures::IdbDatabase;
use indexed_db_futures::prelude::OpenDbRequest;
use serde_wasm_bindgen::Error;
use thiserror::Error;
use wasm_bindgen::JsValue;
use wasm_bindgen::prelude::wasm_bindgen;
use web_sys::DomException;

use penumbra_keys::keys::{SeedPhrase, SpendKey};
use penumbra_proto::DomainType;
use penumbra_proto::serializers::bech32str;

// --------------- //
// DomException doesn't play nicely with thiserror's automated tooling and needs to be wrapped

#[derive(Debug)]
pub struct DomError(DomException);

impl std::fmt::Display for DomError {
    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
        write!(f, "DOM Exception: {:?}", self.0)
    }
}

impl std::error::Error for DomError {}

impl From<DomException> for WasmError {
    fn from(dom_exception: DomException) -> Self {
        WasmError::Dom(DomError(dom_exception))
    }
}
// --------------- //

pub type WasmResult<T> = Result<T, WasmError>;


#[derive(Error, Debug)]
pub enum WasmError {
    #[error("{0}")]
    Anyhow(#[from] anyhow::Error),

    #[error("{0}")]
    Dom(#[from] DomError),
}

impl From<WasmError> for serde_wasm_bindgen::Error {
    fn from(wasm_err: WasmError) -> Self {
        Error::new(wasm_err.to_string())
    }
}

#[wasm_bindgen]
pub async fn some_wasm_fn() -> Result<JsValue, Error> {
    let result = get_spend_key().await?;
    Ok(JsValue::from_str(&result))
}

pub async fn get_spend_key() -> WasmResult<String> {
    let db_req: OpenDbRequest = IdbDatabase::open_u32("penumbra", 12)?;
    // do some stuff
    let spend_key = spend_key_new("xyz")?;
    Ok(spend_key)
}


pub fn spend_key_new(seed_phrase: &str) -> WasmResult<String> {
    let seed = SeedPhrase::from_str(seed_phrase)?;
    let spend_key = SpendKey::from_seed_phrase(seed, 0);

    let proto = spend_key.to_proto();
    Ok(bech32str::encode(
        &proto.inner,
        bech32str::spend_key::BECH32_PREFIX,
        bech32str::Bech32m,
    ))
}

```

Valentine1898:
@grod220 
I've completed a major refactoring of wasm 
- added documentation for all wasm_bindgen functions 
- unwrap() and ok() constructs are no longer used 
- all indexedDB interactions have been moved to a separate file, and some of the code is being reused

Valentine1898:
@zbuc 
In this PR, I have already gotten rid of using unwrap() in wasm crate 
I suggest leaving my variant as it uses succinct `?` and allows for proper error handling in JS

crates/wasm/src/tx.rs
```diff
@@ -103,34 +82,174 @@ pub fn build_tx(
...
+pub async fn transaction_info(full_viewing_key: &str, tx: JsValue) -> Result<JsValue, Error> {
+    let fvk = string_to_fvk(full_viewing_key)?;
+
+    let transaction = serde_wasm_bindgen::from_value(tx)?;
+    let response = transaction_info_inner(fvk, transaction).await?;
+
+    serde_wasm_bindgen::to_value(&response)
+}
+
+/// deprecated
```
grod220, line 100: If deprecated, can we delete? This applies to all the functions labelled with `/// deprecated`

crates/wasm/src/tx.rs
```diff
@@ -103,34 +82,174 @@ pub fn build_tx(
...
 }
 
+/// Get transaction view, transaction perspective
+/// Arguments:
+///     full_viewing_key: `bech32 String`
+///     tx: `pbt::Transaction`
+/// Returns: `TxInfoResponse`
+#[wasm_bindgen]
+pub async fn transaction_info(full_viewing_key: &str, tx: JsValue) -> Result<JsValue, Error> {
+    let fvk = string_to_fvk(full_viewing_key)?;
```
grod220, line 92: I'm not sure the `string_to_fvk` helper function makes sense if it's only one line. Think it can be done in-line.

crates/wasm/src/wasm_planner.rs
```diff
@@ -0,0 +1,221 @@
...
+        self.planner
+            .expiry_height(serde_wasm_bindgen::from_value(expiry_height)?);
+        Ok(())
+    }
+
+    /// Add memo to plan
+    /// Arguments:
+    ///     memo: `MemoPlaintext`
+    pub fn memo(&mut self, memo: JsValue) -> Result<(), Error> {
+        self.memo_inner(memo)?;
```
grod220, line 49: Same feedback as the others. I'm not sure the point of having `inner` helper functions here given the body of the helper function is so small.

crates/wasm/src/wasm_planner.rs
grod220: Same feedback as before 👆

crates/wasm/src/wasm_planner.rs
```diff
@@ -0,0 +1,221 @@
...
+
+    pub async fn swap_claim_inner(&mut self, swap_commitment: JsValue) -> WasmResult<()> {
+        let swap_commitment_proto: StateCommitment =
+            serde_wasm_bindgen::from_value(swap_commitment)?;
+
+        let swap_record: SwapRecord = self
+            .storage
+            .get_swap_by_commitment(swap_commitment_proto)
+            .await?
+            .expect("Swap record not found")
```
grod220, line 139: If you are doing .expect() on these, why not just have `get_swap_by_commitment` return a Result instead of an Option?

crates/wasm/src/storage.rs
```diff
@@ -0,0 +1,179 @@
...
+    pub async fn get_chain_parameters(&self) -> WasmResult<Option<ChainParameters>> {
+        let tx = self.db.transaction_on_one("chain_parameters")?;
+        let store = tx.object_store("chain_parameters")?;
+
+        store
+            .get_owned("chain_parameters")?
+            .await?
+            .map(serde_wasm_bindgen::from_value)
+            .transpose()
+            .map_err(WasmError::from)
```
grod220, line 147: With `thiserror`, you shouldn't need to be doing `.map_err(WasmError::from)` anymore. Just put the error in the new Error struct.

crates/wasm/src/storage.rs
```diff
@@ -0,0 +1,179 @@
...
+
+    pub async fn get_chain_parameters(&self) -> WasmResult<Option<ChainParameters>> {
+        let tx = self.db.transaction_on_one("chain_parameters")?;
+        let store = tx.object_store("chain_parameters")?;
+
+        store
+            .get_owned("chain_parameters")?
+            .await?
+            .map(serde_wasm_bindgen::from_value)
+            .transpose()
```
grod220, line 146: I would not convert this to an Option as every consumer is treating it as a Result.

grod220: Same feedback for the rest of these as well

grod220 (commented):
Substantially better and more succinct with Error handling now fixed 🎉 

Some additional comments. Let me know if you wanted to pair on this to make the remaining fixes or if you wanted to continue on your own.

grod220:
Also, it looks like there are some merge conflicts with main. Please rebase/merge to fix 🙏

crates/wasm/Cargo.toml
```diff
@@ -39,6 +39,13 @@ serde-wasm-bindgen       = "0.5.0"
 wasm-bindgen             = "0.2.87"
 wasm-bindgen-futures     = "0.4.37"
 web-sys                  = { version = "0.3.64", features = ["console"] }
+ark-ff = { version = "0.4", features = ["std"] }
+decaf377 = { version = "0.5", features = ["r1cs"] }
+thiserror = "1.0"
+wasm-bindgen-test = "0.3.0"
```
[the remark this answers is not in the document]
Valentine1898: Align and alphabetical order  :white_check_mark: 
As for the full version, I use the version that uses in other penumbra crates

Valentine1898:
> Substantially better and more succinct with Error handling now fixed 🎉
> 
> Some additional comments. Let me know if you wanted to pair on this to make the remaining fixes or if you wanted to continue on your own.

I think I can make the fixes myself

crates/wasm/src/storage.rs
```diff
@@ -0,0 +1,73 @@
...
+use penumbra_proto::view::v1alpha1::NotesRequest;
+
+enum Store {
+    SpendableNotesStore,
+}
+
+impl Store {
+    fn as_str(&self) -> &'static str {
+        match *self {
+            // *self has type Direction
```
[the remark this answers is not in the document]
Valentine1898: I removed it even earlier

crates/wasm/src/wasm_planner.rs
Valentine1898: This is only necessary in order to use WasmResult<()> in the helper function

crates/wasm/src/wasm_planner.rs
Valentine1898: The point is that on the TS side we will be using mostly [proto](https://buf.build/penumbra-zone/penumbra) types to interact with wasm. 
And in this case, having an object of the [Address](https://buf.build/penumbra-zone/penumbra/docs/main:penumbra.core.crypto.v1alpha1#penumbra.core.crypto.v1alpha1.Address) type on the TS side, we will need to 
1. convert Address -> string on the TS side 
2. on the wasm side get an Address from string object
This leads to additional writing of specific conversion code, while the use of JsValue will be unified everywhere else

crates/wasm/src/storage.rs
Valentine1898: I proceeded from the fact that the situation when we successfully read data from indexedDb by a given key, but there was no record by this key in indexedDB, should not lead to an error.

crates/wasm/src/tx.rs
```diff
@@ -103,34 +82,174 @@ pub fn build_tx(
...
+        match action {
+            Action::Spend(spend) => {
+                let nullifier = spend.body.nullifier;
+                // An error here indicates we don't know the nullifier, so we omit it from the Perspective.
+                if let Some(spendable_note_record) =
+                    storage.get_note_by_nullifier(&nullifier).await?
+                {
+                    txp.spend_nullifiers
+                        .insert(nullifier, spendable_note_record.note.clone());
+                }
```
Valentine1898, line 129: For example, Option is used here when indexedDB reading

crates/wasm/Cargo.toml
```diff
@@ -39,6 +39,13 @@ serde-wasm-bindgen       = "0.5.0"
...
+
+
+
```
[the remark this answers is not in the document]
Valentine1898: ✅

crates/wasm/publish/publish-wasm.ts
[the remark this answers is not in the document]
Valentine1898: ✅

crates/wasm/src/error.rs
```diff
@@ -0,0 +1,63 @@
...
+}
+
+impl std::error::Error for DomError {}
+
+impl From<DomException> for WasmError {
+    fn from(dom_exception: DomException) -> Self {
+        WasmError::Dom(DomError(dom_exception))
+    }
+}
+// --------------- //
```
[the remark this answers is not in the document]
Valentine1898: ✅

crates/wasm/src/error.rs
```diff
@@ -0,0 +1,63 @@
...
+
+pub type WasmResult<T> = Result<T, WasmError>;
+
+#[derive(Error, Debug)]
+pub enum WasmError {
```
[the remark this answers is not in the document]
Valentine1898: ✅

crates/wasm/src/keys.rs
```diff
@@ -0,0 +1,145 @@
...
+    let result = spend_key_from_seed(seed_phrase)?;
+
+    Ok(JsValue::from_str(&result))
+}
+
+/// get full viewing key from spend key
+/// Arguments:
+///     spend_key_str: `bech32 string`
+/// Returns: `bech32 string`
+#[wasm_bindgen(js_name = getFullViewingKey)]
```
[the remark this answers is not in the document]
Valentine1898: I agree, I'll remove the `js_name`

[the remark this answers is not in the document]
Valentine1898: ✅

crates/wasm/src/keys.rs
```diff
@@ -0,0 +1,145 @@
...
+pub fn get_full_viewing_key(spend_key: &str) -> Result<JsValue, Error> {
+    let result = get_full_viewing_key_from_spend(spend_key)?;
+    Ok(JsValue::from_str(&result))
+}
+
+/// get address by index using FVK
+/// Arguments:
+///     full_viewing_key: `bech32 string`
+///     index: `u32`
+/// Returns: `pb::Address`
```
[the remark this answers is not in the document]
Valentine1898: Yes
image

crates/wasm/src/keys.rs
```diff
@@ -0,0 +1,145 @@
...
+///     full_viewing_key: `bech32 string`
+///     index: `u32`
+/// Returns: `String`
+#[wasm_bindgen(js_name = getShortAddressByIndex)]
+pub fn get_short_address_by_index(full_viewing_key: &str, index: u32) -> Result<JsValue, Error> {
+    let result = short_address_by_index(full_viewing_key, index)?;
+    Ok(JsValue::from_str(&result))
+}
+
+pub fn spend_key_from_seed(seed_phrase: &str) -> WasmResult<String> {
```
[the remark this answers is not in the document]
Valentine1898: Removed `pub` wherever it was unnecessary

crates/wasm/src/planner.rs
```diff
@@ -366,67 +390,88 @@ impl<R: RngCore + CryptoRng> Planner<R> {
...
-        let self_address = fvk.incoming().payment_address(source).0;
 
         for value in self.balance.provided().collect::<Vec<_>>() {
             self.output(value, self_address);
         }
 
+        // All actions have now been added, so check to make sure that you don't build and submit an
+        // empty transaction
+        if self.plan.actions.is_empty() {
+            anyhow::bail!("planned transaction would be empty, so should not be submitted");
```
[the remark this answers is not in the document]
Valentine1898: Actually this code is copied from [here](https://github.com/penumbra-zone/penumbra/blob/main/crates/view/src/planner.rs), we just can't reuse the code from view crate because it is not compatible with wasm

crates/wasm/src/tx.rs
Valentine1898: That means it still uses [penumbra-zone/wallet](https://github.com/penumbra-zone/wallet) , so we can't remove it yet.

Valentine1898:
I think I've fixed all the comments except those related to these two issues 
1. Should functions reading from indexedDb return `WasmResult<Option<SwapRecord>> `
2. most helper functions are meaningless, they are only needed so that `#[wasm_bindgen]` functions return `Result<JsValue, Error>` and we can use succinct `?`

crates/wasm/src/keys.rs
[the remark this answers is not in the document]
grod220: Amazing 🎉

crates/wasm/src/storage.rs
grod220: It feels easier to make this a Result as we are nearly always expecting data to be there. It looks like there is only one Option case that could easily work the same as a Result. If you feel strongly otherwise, I can live with it though.

crates/wasm/src/wasm_planner.rs
grod220: In that case we should do this:

```rust
// In wasm_planner.rs
pub fn memo(&mut self, memo: JsValue) -> WasmResult<()> {
    let memo_proto: MemoPlaintext = serde_wasm_bindgen::from_value(memo)?;
    let _ = self.planner.memo(memo_proto.try_into()?);
    Ok(())
}

// in error.rs
impl From<WasmError> for JsValue {
    fn from(error: WasmError) -> Self {
        JsError::from(error).into()
    }
}
```

And then we can use WasmResult everywhere, even the `#[wasm_bindgen]` top level functions. It will remove the need of using the `inner` pattern too. Helper functions are not bad, but we should use them when necessary and not to please the `wasm_bindgen` gods.

crates/wasm/Cargo.toml
```diff
@@ -29,15 +29,19 @@ penumbra-tct           = { path = "../crypto/tct" }
 penumbra-transaction   = { path = "../core/transaction", default-features = false }
 
 anyhow                   = "1.0.75"
+ark-ff                   = { version = "0.4", features = ["std"] }
```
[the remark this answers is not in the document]
grod220: More information on this: https://users.rust-lang.org/t/psa-please-specify-precise-dependency-versions-in-cargo-toml/71277

crates/wasm/src/storage.rs
Valentine1898: Another point is that when initially reading from indexedDB, we get `Option<JsValue> `
```rust
 let read_result: Option<JsValue> = store
            .get_owned("chain_parameters")?
            .await?;
``` 
This means we'd need to convert `Option -> Result`,  and it doesn't seem to be very convenient
For some functions like `get_chain_parameters()` and `get_fmd_parameters()`, we do expect that either the record exists or something is wrong. 
But in other cases, there is a high probability that we will successfully read from indexedDb, but the data will not be there. And in the future, if we want more interaction with indexedDb, there will be more logic similar to 
```rust
  let option = storage.get_note_by_nullifier().await?;
    match option {
        None => {
            // forget nullifier
        }
        Some(note_record) => {
            // Mark record_note as spent
        }
    }
```

crates/wasm/Cargo.toml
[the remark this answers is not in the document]
Valentine1898: I agree, fixed it

crates/wasm/src/tx.rs
Valentine1898: ✅

crates/wasm/src/wasm_planner.rs
Valentine1898: I removed some of the helper functions that seemed redundant. ✅

crates/wasm/src/keys.rs
grod220, line 1: Think this file has some helper functions that are small enough to be moved into the main function. It's similar feedback as the other. Will have to move `WasmResult` to the the top level function.

grod220 (approved):
Wow! 🎊 

Though this PR took a lot of iterations, think the wasm crate is in a _much_ better place with these new patterns. Going to approve, but think there's one file where the helpers are small enough to move into the main function. After you do that, feel free to merge 👍

crates/wasm/src/keys.rs
Valentine1898: ✅

Valentine1898:
@conorsch can merge this PR?

conorsch:
There's just one more conflict, on the cargo lock file, that's blocking merge. I'll fix that now, and get it in!

conorsch:
Turns out rebasing this PR on latest main was not straightforward. After an hour of unsuccessful interactive rebasing, I ended up performing a mixed reset on top of main, selectively adding only the wasm-related changes, and discarding the rest. That approach looks to have succeeded in preserving the new content of the PR while catching up the base with latest main.

In case I made a mistake, I made sure to push a "backup" branch to the penumbra remote: https://github.com/penumbra-zone/penumbra/tree/wasm_transaction_planner-backup That branch captures the state of this PR prior to my rebasing.

There's only one commit on this PR now, including the entirety of changes over the course of review.  Tests and linters pass locally, so I think we're in good shape.

@grod220 and @Valentine1898: given your familiarity with the changes here, can you take one more look at the diff and confirm it conforms to expectations? If so, we're good to merge!

conorsch (approved):

Valentine1898:
Oh, I didn't think merge would be a problem. 
I reviewed the code again, it's all right.

grod220 (approved):
@Valentine1898, please merge 🙏

Valentine1898:
I can't merge, I don't have permission to do that

grod220:
@Valentine1898 oh! Didn't realize. I feel like an approved PR by penumbra team member should enable external folks to merge 🤔 . I'll do it, all good.

PR #2974, opened 2023-09-06
debug(rpc): witness and build logging

conorsch:
Adding some debug logging statements to the witness and build logic, to aid in tracing a failure encountered during interchaintest integration. The WitnessAndBuild RPC never returns: pclientd balloons in memory consumption until oomkilled. Let's isolate where in the witness-and-build logic it's getting stuck.


conorsch:
This was in pursuit of #2977. The real fix was #2980, so closing this PR without merging.

PR #2978, opened 2023-09-07
crypto: use `DetectionData` proto encoding for effect hash

redshiftzero:
Followup from #2931 and [recent discussion in discord](https://discord.com/channels/824484045370818580/864913084028158012/1149384015032365117)

### DIFF
--- a/crates/core/transaction/src/effect_hash.rs
+++ b/crates/core/transaction/src/effect_hash.rs
@@ -486,15 +486,8 @@ impl EffectingData for PositionRewardClaim {
 
 impl EffectingData for DetectionData {
     fn effect_hash(&self) -> EffectHash {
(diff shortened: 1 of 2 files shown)


cronokirby:
With this change everything is done with the protos now, right?

redshiftzero:
> With this change everything is done with the protos now, right?

For the individual transaction fields and actions, that's correct. The only place in the effect hash computation where we _don't_ use the proto is for combining those individual effect hashes together, i.e. in  `TransactionBody::effect_hash`, which makes sense since we want to use the effect hashes of each component instead of the full proto-encoded data.

conorsch:
Merging for 60. This is a breaking change, so let's get it in for Monday's release.

PR #3002, opened 2023-09-11
view: expose source control in TransactionPlanner RPC

hdevalence:
The transaction planner allows controlling the source of spent funds, but we didn't expose this to the `TransactionPlanner` RPC. This commit adds a new field to the protos that threads this parameter through.


hdevalence:
@conorsch would it be possible to merge this and deploy it as `0.60.1`? It doesn't affect consensus at all, it's purely client-side.

PR #3006, opened 2023-09-12
proto: stub proto definitions for setup coordinator

hdevalence:
Results of pairing with @redshiftzero and @cronokirby .


conorsch:
Rebased to resolve conflicts from merging #3016.

conorsch:
Saw a new buf lint on this PR:

```
Error: Service name "CeremonyCoordinator" should be suffixed with "Service".
Error: buf found 1 lint failures.
```

We've honored that rule elsewhere, so make since to me. I'll tack it on.

PR #3016, opened 2023-09-13
build(proto): update buf dep to cosmos/ibc v8

conorsch:
At the request of collaborators from Strangelove Ventures, bumping the cosmos/ibc Buf Build dependency v7 -> v8. It's good to stay on top of this version stream, and the golang integrations of Penumbra are a good excuse to motivate timely upgrades.

Notably we're not bumping cosmos/sdk, as the comment is still accurate: to upgrade, we must update our tendermint protos with knowledge of vote extensions. Not doing that now.


conorsch:
Merging to unblock SL work.

PR #3023, opened 2023-09-14
[Bug fix] Forgetting delegation tokens

grod220:
Two bugs:
1. Spent commitments are being forgotten if they _are_ delegation tokens. Should flip the logic on that conditional.
2. Spent swap commitments are never delegation tokens.

Found while pairing 🍐 with @plaidfinch

### DIFF
--- a/crates/view/src/storage.rs
+++ b/crates/view/src/storage.rs
@@ -1352,20 +1352,14 @@ impl Storage {
                     // which must be saved to allow voting on proposals that might or might not be
                     // open presently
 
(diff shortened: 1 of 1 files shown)


grod220:
@hdevalence fyi, we were pairing yesterday and found some oddities while trying to write a smoke test for this:
- In main (before this change), after undelegating, voting appeared to work (not sure how if bug was forgetting)
- Tallys did not show the vote count
- Spent denom not showing unbonding token when we used tracing

@plaidfinch will sync with Aubrey to see get additional eyes on the `spent_denom` sql query.

PR #3036, opened 2023-09-15
ibc: allow legacy proof specs without prehash_key_before_comparison

avahowell:


conorsch:
This change was included in https://github.com/penumbra-zone/penumbra/pull/3043 and so has already landed, therefore closing.

hdevalence:
We should however make a follow-up issue to track reverting this rule; ideally we don't allow it during mainnet since it breaks timeouts.

Issue #2937, opened 2023-08-15
ibc: add ability to query all channels


conorsch:
**Is your feature request related to a problem? Please describe.**
We need a way to enumerate all available ibc transfer channels, via `pcli query ibc channels`.

When a Penumbra user wants to make an IBC withdrawal, they need to confirm that a given channel exists, and also that it's the appropriate channel for the desired counterparty chain. In our early steps toward full cross-chain support for Penumbra, we plan to document IBC relaying between Penumbra testnet and Penumbra preview (#2907). Given that preview is frequently destroyed and recreated, on every merge into main, this means that the valid, working channel on Penumbra testnet will always been `n - 1`, where `n` is the number of merges into main since the last testnet release.

Currently, we only have the ability to query for specific channel ids:

```
❯ pcli query ibc channel transfer channel-0
{
  "state": "Open",
  "ordering": "Unordered",
  "remote": {
    "port_id": "transfer",
    "channel_id": "channel-0"
  },
  "connection_hops": [
    "connection-0"
  ],
  "version": "ics20-1"
}
```

It's the most recent channel that will be valid for our use case, however. Right now, a user must tediously increment the query by one, until an error occurs, and then they can reasonable conclude that the last successful query is the desired channel. Hardly ideal UX. 

**Describe the solution you'd like**
Let's implement at least a way to find the most recent channel. We could do `pcli query ibc channels` (note the "s") to dump _all_ channels. This is the shortest path, although eventually that list could get quite long. When it becomes a problem, perhaps `--latest=1` would be a right-size fit to review the newest opened channels.

**Describe alternatives you've considered**

**Additional context**
I believe it's reasonable to provide a default value of "transfer" for the channel "port" value, given that it's hardcoded elsewhere in our implementation.

conorsch:
Foundation is in https://github.com/penumbra-zone/penumbra/pull/2954 ; after that's in, we just need to wire up a `pcli` action.

conorsch:
Started on this today. The changes in #2954 are a great start, but several of the relevant queries are still marked "unimplemented." Will wire them up as part of these changes. The bare minimum for completeness should be support for the following:

* `pcli q ibc client 0`
* `pcli q ibc clients`
* `pcli q ibc connection 0`
* `pcli q ibc connections`
* `pcli q ibc channel 0`
* `pcli q ibc channels`

[...]

PR #3037, opened 2023-09-15
docs: relayer setup and usage

conorsch:
These docs changes are based on feedback from a pairing session with @aubrika today.

### DIFF
--- a/deployments/relayer/build-path
+++ b/deployments/relayer/build-path
@@ -27,6 +27,6 @@ visible_log "These commands should all yield results:"
 
 cat <<EOF
   * pcli -n http://localhost:8080 q ibc client 07-tendermint-0
-  * pcli -n http://localhost:8080 q ibc channel transfer channel-0
+  * pcli -n http://localhost:8080 q ibc channel --channel-id 0
   * pcli -n http://localhost:8080 q ibc connection connection-0
 EOF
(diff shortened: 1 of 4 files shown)


deployments/relayer/build-path
```diff
@@ -27,6 +27,6 @@ visible_log "These commands should all yield results:"
 
 cat <<EOF
   * pcli -n http://localhost:8080 q ibc client 07-tendermint-0
-  * pcli -n http://localhost:8080 q ibc channel transfer channel-0
+  * pcli -n http://localhost:8080 q ibc channel --channel-id 0
```
hdevalence, line 30: why does this command have different syntax to all of the other `q ibc` commands?

docs/guide/src/pcli/transaction.md
```diff
@@ -198,7 +198,7 @@ while read -r i ; do
         echo "Most recently active channel is: $(( i - 1))" ;
         break ;
     fi ;
-done <<< "$(seq 0 10)"
+done <<< "$(seq 0 100)"
```
hdevalence, line 201: side note, but why don't we just do the real query command? wouldn't that be a comparable effort to documenting and updating the bash script?

deployments/relayer/build-path
conorsch: We haven't been exercising these paths much! That's changing now, as IBC comes into focus.

docs/guide/src/pcli/transaction.md
conorsch: Totally agree: enumerating all the matches to a query is way more useful cli functionality that spot-checking individual query targets. That's tracked in #2937.

PR #3047, opened 2023-09-18
chore: update chain id for 061-dione

conorsch:
Also pulls in the latest faucet allocations from discord, via galileo.


conorsch:
I removed the new discord allocations; OK to merge once CI passes.

Issue #3056, opened 2023-09-18
Rethink peering logic for testnet join


conorsch:
**Describe the bug**
A new node joining a testnet via `pd testnet join` will either get very few peers, or a whole bunch. We should stabilize this behavior. 

**To Reproduce**
```
❯ for i in $(seq 10) ; do \
curl -s https://rpc.testnet.penumbra.zone/net_info | jq -r .result.n_peers ; \
sleep 1 ; done 
84
2
84
2
84
2
2
84
84
3
```

**Expected behavior**
I expect to receive a large number of peers, stably, on repeated attempts to join the network.

**Screenshots**

**Additional context**
The reason this is happening is because:

1. The `testnet join` logic inspects peer info from the bootstrap node address and adds those addresses to the `seeds` line of the Tendermint config;
2. The "bootstrap node address" is load-balanced across two (2) full nodes, each of which can have different numbers of peers;
3.  Our limits for maximum number of peers (currently 75 in/out) are being hit across all the nodes we run.

We've also been using `seed_mode=true` on fn-0, which will automatically disconnect peering attempts after exchanging info about other peers. If we instead run more nodes, we should be able to trust Tendermint/CometBFT's address book functionality to track valid peers over the life of the network, rather than hardcoding a few known peers as "seeds" at join time.

1613557499:
I[2023-09-19|13:52:09.591] Saving AddrBook to file                      module=p2p book=/root/.penumbra/testnet_data/node0/tendermint/config/addrbook.json size=5
I[2023-09-19|13:52:09.594] Ensure peers                                 module=pex numOutPeers=0 numInPeers=0 numDialing=0 numToDial=10
I[2023-09-19|13:52:39.594] Ensure peers                                 module=pex numOutPeers=0 numInPeers=0 numDialing=0 numToDial=10
I[2023-09-19|13:53:09.593] Ensure peers                                 module=pex numOutPeers=0 numInPeers=0 numDialing=0 numToDial=10


Look abnormal

conorsch:
Talked this over with @zbuc. Changes to make:

- [ ] Add more nodes to our default CI deployment
- [ ] Raise the default limits for max in/out peers (we're currently using ~half of what e.g. osmosisd configures)
- [ ] Adjust `join` logic to take a random subset of the found nodes
- [ ] Stop using seed_mode=true on fn-0

conorsch:
> Add more nodes to our default CI deployment

[...]

PR #3063, opened 2023-09-19
feat(pd): raises default peering limits

conorsch:
Makes a few changes to `pd`, designed to improve peering behavior on testnets:

  * raises the default values for inbound/outbound peers
  * limits the number of peers discovered during join to 5, randomized

We originally decided to add node discovery logic to aid in peering. There's a downside, however, to reusing all peers of the bootstrap node: the CI deployment nodes get swamped, and the network remains centralized on the nodes that Penumbra Labs run. Instead, let's just take a handful, at random, and trust the p2p gossip to discover more over time.

Refs #3056.


zbuc (approved):
LGTM!

Issue #2288, opened 2023-03-31
Tracking issue: Modularize the Penumbra application code into coherent crates


hdevalence:
**Is your feature request related to a problem? Please describe.**

Since there isn't already a Rust application framework for building Tendermint chains, we've had to build one at the same time as we've built out our application.  This has been a messy and iterative process, and along the way, we've passed from having separate crates for separate components of the application (which turned out not to be entirely separate) to having effectively all of our application logic in a single crate, formerly called `penumbra_component` and now called `penumbra_app` (after #2282). 

However, while looking at the current structure of our code, we realized there may be a path to re-modularizing it, not in a single stop-the-world refactor, but gradually peeling out chunks of functionality.

**Describe the solution you'd like**

The following plan is a sketch; the first steps are concrete and definitely a good idea, while the later steps are more speculative. As we do this, let's think step by step, and notice any unexpected frictions.

- [x] Change the `Component` trait so that the `AppState` passed to `init_chain` is an associated type.
  
  This makes the definition of the `Component` trait independent of the rest of the Penumbra code. Later, it will allow us to split up the genesis state into per-component chunks, with each different `Component` operating on its own definition of its genesis data.  This avoids cross-component dependencies (e.g., component X has to know about component Y because the genesis data passed to component X has types from component Y).
  
  It might be worth delaying this until we can clean up the IBC code, which has a number of `Component` implementations that may not be necessary.

- [x] Change the `ActionHandler` trait to add a `CheckStatelessContext: Clone + Send + Sync + 'static` associated type.

  This makes the definition of the `ActionHandler` trait independent of the rest of the Penumbra code.  Most `ActionHandler`s don't make use of the context, and could have `CheckStatelessContext = ()`. Others that do probably just use the anchor, and could have that specified as the verification context, rather than a full `Arc<Transaction>`.  This avoids cross-action dependencies (e.g., every action has to know about every other action, because it has to know about the `Transaction` type that contains all of them including itself).
  
  Longer-term, this also presents some interesting opportunities.  For instance, we could use the [`tower-batch`](https://zfnd.org/composable-futures-based-batch-verification/) strategy to create a context struct that had channels communicating with `tower::Service`s performing batch verification of signatures and proofs.

- [x] Pull the revised `Component` and `ActionHandler` traits out into a new `penumbra_component` crate.

[...]

PR #3067, opened 2023-09-20
refactor: move governance code to `penumbra-governance`

### LINKED ISSUE #2950: Pull governance out of the `app` crate
Currently the governance mechanism is still in the `app` crate, rather than in its own component crate. This should be fixed.

redshiftzero:
Closes #2950
This addresses the remaining governance items in #2288

Note that there is one non-refactoring change here which is in [this commit](6a8634e7ffc10526ac3e749475804ea5bcc3801f), which includes `UndelegateClaim` in the list of actions that have proofs and thus should not be possible in a `DaoSpend` proposal due to DoS risk. Since claiming undelgate outputs is not possible, `Delegate` and `Undelegate` actions are also now disallowed.


hdevalence:
Merging in order to allow proto refactoring.

PR #3073, opened 2023-09-21
ibc client upgrades

### LINKED ISSUE #635: ibc: implement upgradeclient to allow counterparty chains to upgrade
There are two cases where we need to consider handling upgrades for IBC clients:

- The host chain (penumbra) performs an upgrade.
- A counterparty, a client of which exists on penumbra, performs an upgrade.

First case is tracked in https://github.com/penumbra-zone/penumbra/issues/2985.

This issue refers to the second case. At a high level, in order to handle this case, we must:

- Implement a state handler for `MsgUpgradeClient`, similarly to `MsgUpdateClient`, that allows us to verify that a counterparty included an UpgradeClient in their state which specifies a new ConsensusState and ClientState for that client.
- Once we have verified that the counterparty has committed this data, overwrite the existing client with the newly committed (on the counterparty) ConsensusState and ClientState.

We should replicate existing ecosystem implementations of this functionality, to maintain compatibility. These can be found here:

- [ibc-rs](https://github.com/cosmos/ibc-rs/blob/95704fcede4146fd9039580a8a8ca2218180e400/crates/ibc/src/core/ics02_client/handler/upgrade_client.rs#L16)

ibc-go:
- https://github.com/cosmos/ibc-go/blob/56a519c188ae841818b9ecd11244240941cc5d90/modules/core/02-client/types/msgs.go#L168
- https://github.com/cosmos/ibc-go/blob/56a519c188ae841818b9ecd11244240941cc5d90/modules/core/02-client/keeper/client.go#L111

At a high level,

- Add a new `UpgradeClient` type to the existing IBC action [enum](https://github.com/penumbra-zone/penumbra/blob/a97a0855f2c830c723a678eed7311ff4901b974d/crates/core/component/ibc/src/ibc_action.rs#L21)
- Add new branches to every switch on that enum
- Inside the execution handlers, implement the logic that constructs and verifies the merkle inclusion proofs necessary (see `proof_verification.rs`)
- Once proofs are verified, use the StateWrite trait in client to overwrite the client's consensus and client states

avahowell:
Closes #635

### DIFF
--- a/crates/core/component/ibc/src/component/action_handler/ibc_action.rs
+++ b/crates/core/component/ibc/src/component/action_handler/ibc_action.rs
@@ -14,6 +14,7 @@ impl ActionHandler for IbcAction {
         match self {
             IbcAction::CreateClient(msg) => msg.check_stateless().await?,
             IbcAction::UpdateClient(msg) => msg.check_stateless().await?,
+            IbcAction::UpgradeClient(msg) => msg.check_stateless().await?,
             IbcAction::SubmitMisbehavior(msg) => msg.check_stateless().await?,
             IbcAction::ConnectionOpenInit(msg) => msg.check_stateless().await?,
             IbcAction::ConnectionOpenTry(msg) => msg.check_stateless().await?,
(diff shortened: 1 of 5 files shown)


conorsch:
@avahowell What's the status of this PR? Mergeable for 61 (#3046), or still more work required?

hdevalence:
I think we should delay it until the next cycle.

avahowell:
Yes, this PR only contains the proof verification. we still need to add the message itself and its state machine handling

conorsch:
Merging! This'll ship in 62.

PR #3077, opened 2023-09-22
proto: align proto packages to rust crates

hdevalence:
After reorganizing the Rust crates into a coherent heirarchy, we never did the same for the .proto files. This PR does that.

Currently, this just reorganizes the proto files, updates the `proto-compiler` tool, and regenerates the source code.  What's remaining is to update all of the import paths in all of the domain type implementations that reference the proto structures. I'd suggest we swarm that part tomorrow morning, using live share + search&replace, since this PR will conflict with everything else (and break all of the protos used by other tooling like the web interfaces).

I **don't** think we should try to update all of the `TYPE_URL`s at this point; those are used for generic `Any` support, which we don't usually make use of, and landing the changes ASAP is more important than restoring that functionality (except where we know we need it because eg the smoke test fails).


hdevalence:
A follow-on step once this PR lands (which we should do as soon as the Rust code compiles again) is to split up the RPC definitions in the `client` proto package into per-component `QueryService`s. We might also be able to land that as part of the next release cycle.

hdevalence:
It was easier than I expected to patch up the import paths; we'll still need to go through and fix up the `TYPE_URL` constants.

hdevalence:
It looks like `prost` `0.12` has a `Name` trait: https://docs.rs/prost/latest/prost/trait.Name.html so we could skip updating all the `TYPE_URL`s, switch to `prost` `0.12`, and delete all the impls instead of updating them.

conorsch:
Refs #2288

Issue #3078, opened 2023-09-22
Update to `prost` `0.12`, `tonic` `0.10`.


hdevalence:
This is useful for two reasons:

> It looks like `prost` `0.12` has a `Name` trait: https://docs.rs/prost/latest/prost/trait.Name.html so we could skip updating all the `TYPE_URL`s, switch to `prost` `0.12`, and delete all the impls instead of updating them.

_Originally posted by @hdevalence in https://github.com/penumbra-zone/penumbra/issues/3077#issuecomment-1730866734_

and also, because Tonic `0.10.0` has an `add_routes` feature that will let us composably build up RPC services defined in each component: https://docs.rs/tonic/latest/tonic/transport/server/struct.Server.html#method.add_routes

This will require pulling all our dependencies over.

hdevalence:
https://github.com/cosmos/ibc-proto-rs/issues/145

conorsch:
Resolved via https://github.com/penumbra-zone/penumbra/pull/3192. Notably we've not yet implemented: 

> and also, because Tonic 0.10.0 has an add_routes feature that will let us composably build up RPC services defined in each component: https://docs.rs/tonic/latest/tonic/transport/server/struct.Server.html#method.add_routes

but we're now unblocked from doing so.

PR #3079, opened 2023-09-22
ibc: split up query rpc implementations

hdevalence:


crates/core/component/ibc/src/component/rpc.rs
```diff
@@ -0,0 +1,21 @@
...
+
+impl IbcQuery {
+    pub fn new(storage: penumbra_storage::Storage) -> Self {
+        Self(storage)
+    }
+}
+
+pub fn routes(_storage: Storage) -> Routes {
+    unimplemented!("functionality we need is only in tonic 0.10")
+}
```
hdevalence, line 21: The hope here was to be able to expose just an `rpc::routes()` method that would return a partial bundle of `Service`s. This would avoid the code in `pd/main.rs` from having to recite them all, but that API was only added in Tonic 0.10, so it needs #3078 for that approach to work.

PR #3086, opened 2023-09-22
WIP: Split ChainParameters into component-specific parameters

### LINKED ISSUE #3084: Split chain parameters into per-component parameters

As part of #2288, we split up the Penumbra application into a coherent collection of "components" (from its previous state as a big-ball-of-mud). This refactoring split the crates into two rough "categories":

- "base" crates that do parts of the application logic, and can depend on each other as needed (e.g., shielded pool, stake, dex)
- "superstructure" crates that pull in all the base crates and assemble them into a unified whole (e.g., transaction, which has all the actions, app, which has all the components, etc)

The problem we have is that our current chain parameters structure mixes categories. We have one monolithic `ChainParameters`, which must be accessible to every component, but which also has data relevant to every component.  This creates dependency cycles whenever we try to add structured data to it, as we recently discovered with #2970.

Split the chain parameters up into per-component parameters, e.g., a `DexParameters` in the dex component, `StakeParameters` in the stake crate, and so on, then unify these into an `AppParameters` in the app crate:
```rust
struct AppParameters {
    pub DexParameters,
    pub StakeParameters,
    ...
}
```
Each component's parameters are recorded in that component, so components can access their own parameters as well as all the parameters of components they depend on.  The top-level `AppParameters` wouldn't be stored in a specific location; it would need to be reassembled from the component parameters, but this seems fine.

**Additional context**
A final part of #2288.

zbuc:
Closes #3084


crates/core/component/shielded-pool/src/component/shielded_pool.rs
```diff
@@ -20,17 +21,16 @@ pub struct ShieldedPool {}
 
 #[async_trait]
 impl Component for ShieldedPool {
-    type AppState = genesis::AppState;
+    type AppState = GenesisContent;
 
     #[instrument(name = "shielded_pool", skip(state, app_state))]
-    async fn init_chain<S: StateWrite>(mut state: S, app_state: &genesis::AppState) {
+    async fn init_chain<S: StateWrite>(mut state: S, app_state: &GenesisContent) {
         match app_state {
```
erwanor, line 28: I think something like this could work:

```rust
async fn init_chain<S: StateWrite>(mut state: S, app_state: Option<&GenesisContent>) {
    match app_state {
        None => { /* no-op */ },
        Some(app_state) => {
            /* bootstrapping */
        }
    }
    /* logic that is common to both checkpointed and full config genesis */
}
```

In most cases, this can be refactored even further since the checkpointing case is a no-op. Either way, the most important thing is that the common logic be executed, in particular in the staking component as we build the cometbft view of the validator set.

erwanor: This means we would only be able to perform checkpointing verification logic and health checks at the application-level `init_chain`, but that seems fine.

erwanor: The application state argument to the component's `init_chain` would be the component specific application state I imagine

crates/core/app/src/app/mod.rs
```diff
@@ -87,7 +86,15 @@ impl App {
             .expect("state Arc should not be referenced elsewhere");
         match app_state {
             genesis::AppState::Content(app_state) => {
-                state_tx.put_chain_params(app_state.chain_params.clone());
+                // Put all the initial component parameters in the JMT:
+                // TODO: should the components themselves handle this in their
+                // `init_chain` methods?
```
hdevalence, line 91: Yeah, I don't think we need to do it in this PR, but I think ideally we shouldn't have the `App` do any initialization at all, and only have it call into specific components' `init_chain` methods.  In order to do that, we'd need to have a clear assignment of responsibility of functionality per component, which we don't currently have (but should).

For instance, why is the epoch being written directly by the `App`, if the epoch handling is intertwined with the SCT? Why are the FMD parameters set by the `App` rather than by the shielded pool? etc.

crates/core/app/src/app/mod.rs
```diff
@@ -113,16 +120,26 @@ impl App {
...
                     },
                 );
+
+                Distributions::init_chain(&mut state_tx, Some(&())).await;
+                ShieldedPool::init_chain(&mut state_tx, Some(&app_state.shielded_pool_content))
+                    .await;
+                Staking::init_chain(&mut state_tx, Some(&app_state.stake_content)).await;
+                IBCComponent::init_chain(&mut state_tx, Some(&())).await;
+                Dex::init_chain(&mut state_tx, Some(&())).await;
+                Governance::init_chain(&mut state_tx, Some(&())).await;
```
hdevalence, line 130: If the component parameters were set by the component itself, we'd need each of these impls to have the associated type be a container that had the component parameters (left open for other genesis data in the future), right?

That would involve making a proto message for each component's genesis data, which would be some amount of boilerplate, and we could definitely leave it out of this PR, but wanted to raise it for discussion or as a next step.

hdevalence: Oh, wait, it looks like there are already `genesis::Content` types in each component?

crates/core/app/src/params/change.rs
```diff
@@ -1,68 +1,92 @@
...
 // The checks below validate that a parameter change is valid, since some parameter settings or
 // combinations are nonsensical and should be rejected outright, regardless of governance.
 
 #[deny(unused)] // We want to be really careful here to not examine fields!
-impl ChainParameters {
-    pub fn check_valid_update(&self, new: &ChainParameters) -> Result<()> {
+impl AppParameters {
+    pub fn check_valid_update(&self, new: &AppParameters) -> Result<()> {
         new.check_valid()?;
+        // TODO: move the checks below into their respective components
```
hdevalence, line 19: 👍 commenting to flag something we should track as a follow-up. Probably we could do it while updating the parameter change logic?

crates/core/app/src/params/change.rs
```diff
@@ -71,26 +95,40 @@ impl ChainParameters {
...
+                    proposal_voting_blocks,
+                    proposal_deposit_amount,
+                    proposal_valid_quorum,
+                    proposal_pass_threshold,
+                    proposal_slash_threshold,
+                },
+            dao_params:
+                DaoParameters {
+                    dao_spend_proposals_enabled: _,
+                }, // IMPORTANT: Don't use `..` here! We want to ensure every single field is verified!
```
hdevalence, line 131: Now that we have per-component parameters, it feels like we can have a much better solution to this problem, since we can do all of the parameter validation "locally" in each component, whereas before, our only option was to do all of that logic "at a distance" here in the governance component (cc @plaidfinch)

crates/core/component/dao/src/component/state_key.rs
```diff
@@ -1,10 +1,14 @@
 use penumbra_asset::asset;
 
+pub fn dao_params() -> &'static str {
+    "dao_params"
```
hdevalence, line 4: nit: we probably want this to be `dao/params` so that it's under the same prefix as the other keys in the component?

crates/core/component/dao/src/component/view.rs
```diff
@@ -38,6 +47,17 @@ impl<T> StateReadExt for T where T: StateRead + ?Sized {}
 
 #[async_trait]
 pub trait StateWriteExt: StateWrite {
+    /// Writes the provided DAO parameters to the JMT.
+    fn put_dao_params(&mut self, params: DaoParameters) {
+        // TODO: this needs to be handled on a per-component basis or possibly removed from the compact block
+        // entirely, currently disabled, see https://github.com/penumbra-zone/penumbra/issues/3107
+        // Note to the shielded pool to include the chain parameters in the next compact block:
+        // self.object_put(state_key::chain_params_changed(), ());
```
hdevalence, line 55: Can we flag this in that issue, so we don't forget? Or maybe add a checklist item to that issue to do `rg 3107` and ensure that all the references to the issue were resolved as part of the fix?

hdevalence: The latter seems like less work and more resilient to forgetting to tag something on the issue.

crates/core/component/ibc/src/component/state_key.rs
```diff
@@ -4,6 +4,10 @@ use penumbra_asset::asset;
 
 use std::string::String;
 
+pub fn ibc_params() -> &'static str {
+    "ibc_params"
```
hdevalence, line 8: ditto nit on prefix

crates/core/component/stake/src/state_key.rs
```diff
@@ -3,6 +3,10 @@ use tendermint::PublicKey;
 
 use crate::IdentityKey;
 
+pub fn stake_params() -> &'static str {
+    "stake_params"
```
hdevalence, line 7: nit on prefix

hdevalence (approved):
Looks amazing!!!!

Thanks for pushing this through, it's a huge refactor.

crates/core/component/dao/src/component/view.rs
zbuc: Yeah that was my intent with leaving the link to the issue everywhere, I'll make sure to add that to #3107

PR #3088, opened 2023-09-22
crypto: add helper function for decrypting transaction memo

redshiftzero:
This PR adds a small helper method for decrypting memos given a `Transaction` cc @aubrika


aubrika (approved):
LGTM!

PR #3091, opened 2023-09-23
proto: split RPCs into per-component QueryServices

### LINKED ISSUE #3083: Split existing GRPC services into per-component `QueryService`s.
The final part of the modularization in #2288 is to split up the GRPC service definitions into per-component `QueryService`s.

One example of this so far is the IBC query support, implemented in the IBC component. See also #3079 for restructuring of that code.

Ideally we'd do this as part of Testnet 61, so that we do all the breaking protobuf changes in one go.

_Originally posted by @hdevalence in https://github.com/penumbra-zone/penumbra/issues/2288#issuecomment-1731636563_

hdevalence:
Closes #3083.

Still outstanding work to do:

- [x] Split up the server-side implementations currently in `pd`.
- [ ] Resolve the governance issues.

On the latter point, I discovered that large parts of the governance code don't have any RPCs defined, and instead rely on having knowledge of the Rust code to layer together raw state accesses.  This is practically impossible for any third party client to do, so it's not an acceptable solution.  The shim code that enabled this was removed as part of the refactoring, so we'll either need to define the relevant RPCs, or leave that functionality disabled until we do.


hdevalence:
I ran into another difficulty while making these changes. We have an RPC to provide raw K/V queries to the underlying storage, and it seemed like this should conceptually be attached to the `penumbra-storage` crate. Because `penumbra-storage` (intentionally) doesn't depend on `penumbra-proto` (so that it can be more independent of the Penumbra crates and reusable by other projects), doing this meant duplicating parts of our proto compilation logic to have a parallel production of proto codegen in the `penumbra-storage` crate.

However, the RPC isn't actually really generic to the storage system, for the following reason: it returns K/V entries with ICS23 proofs up to the _apphash_, and the Penumbra apphash isn't the JMT merkle root. We add an extra layer of hashing to have a second "stub" proof to satisfy requirements of counterparty chains. So I put the RPCs in the `app` package's `QueryService` for now.

conorsch:
Refs #2288

Issue #3080, opened 2023-09-22
Passing indexedDB constans to wasm


Valentine1898:
Problem: wasm crate uses hard-coded values for indexedDB

Solution: pass IndexedDB constants values to the constructor 

Related:  #https://github.com/penumbra-zone/web/issues/39

PR #3097, opened 2023-09-25
3080 passing indexeddb constans to wasm

Valentine1898:
#3080


crates/wasm/src/storage.rs
```diff
@@ -9,23 +9,45 @@ use penumbra_proto::view::v1alpha1::{NotesRequest, SwapRecord};
...
+    chain_parameters: String,
+    fmd_parameters: String,
+    last_block_synced: String,
+    notes: String,
+    spendable_notes: String,
+    swaps: String,
+    tree_commitments: String,
+    tree_hashes: String,
+    tree_last_forgotten: String,
+    tree_last_position: String,
```
grod220, line 29: Don't think we need to list more than this crate needs access to

crates/wasm/src/storage.rs
```diff
@@ -9,23 +9,45 @@ use penumbra_proto::view::v1alpha1::{NotesRequest, SwapRecord};
...
+    name: String,
+    version: u32,
+
+    assets: String,
+    chain_parameters: String,
+    fmd_parameters: String,
+    last_block_synced: String,
+    notes: String,
+    spendable_notes: String,
+    swaps: String,
```
grod220, line 25: Let's nest this within another struct called `Tables`

crates/wasm/src/wasm_planner.rs
```diff
@@ -27,10 +28,10 @@ pub struct WasmPlanner {
 #[wasm_bindgen]
 impl WasmPlanner {
     #[wasm_bindgen(constructor)]
-    pub async fn new() -> Result<WasmPlanner, Error> {
+    pub async fn new(idb_constants: IndexedDbConstants) -> Result<WasmPlanner, Error> {
         let planner = WasmPlanner {
             planner: Planner::new(OsRng),
-            storage: IndexedDBStorage::new().await?,
+            storage: IndexedDBStorage::new(idb_constants).await?,
```
grod220, line 34: It looks like the constructor now requires opening up an IndexedDb connection. Can you add a code comment about this fact?

Or, do we need to opening up that connection when we instantiate the class? Could we not do it when we make the request? Just a thought.

crates/wasm/src/storage.rs
```diff
@@ -9,23 +9,45 @@ use penumbra_proto::view::v1alpha1::{NotesRequest, SwapRecord};
...
 
 pub struct IndexedDBStorage {
     db: IdbDatabase,
+    constants: IndexedDbConstants,
 }
 
 impl IndexedDBStorage {
-    pub async fn new() -> WasmResult<Self> {
-        let db_req: OpenDbRequest = IdbDatabase::open_u32("penumbra", 12)?;
+    pub async fn new(constants: IndexedDbConstants) -> WasmResult<Self> {
```
grod220, line 38: Does wasm-pack serialize `IndexedDbConstants` properly?

Valentine1898: I need to add set and get functions to properly export the struct. But actually, it seems more convenient to just use `JsValue`

crates/wasm/src/wasm_planner.rs
Valentine1898: We can choose not to open IndexedDb connection in constructor, but then we will have to open IndexedDb connection on every request
Added documentation

grod220: Yeah, that's fair

grod220 (approved):

PR #3098, opened 2023-09-25
Add amount_by_sender query to storage to return transaction amounts and denominations grouped by sender address

### LINKED ISSUE #3068: Storage Method for Recipient Tallying
As part of our ongoing work on summoning ceremony coordination, we ultimately need a way to easily figure out how much of a given asset has been sent to us by specific addresses, as declared in the recipient part of the transaction memo.

One simple way to add this feature would be to add extra functionality to the storage struct in the view crate. One way of dividing up the feature would be to add a method which streams out pairs of `(Address, Amount)` by joining the transaction table and the note table, giving us the total amount sent to the owner of a viewing key by various addresses.

This would potentially be useful for other situations as well, and would be enough for using in the coordinator.

aubrika:
Closes #3068

### DIFF
--- a/crates/view/src/storage.rs
+++ b/crates/view/src/storage.rs
@@ -1225,6 +1225,8 @@ impl Storage {
         let scanned_nullifiers_tx = self.scanned_nullifiers_tx.clone();
         let scanned_swaps_tx = self.scanned_swaps_tx.clone();
 
+        let fvk = self.full_viewing_key().await?;
+
         // Cloning the SCT is cheap because it's a copy-on-write structure, so we move an owned copy
         // into the spawned thread. This means that if for any reason the thread panics or throws an
         // error, the changes to the SCT will be discarded, just like any changes to the database,
@@ -1378,12 +1380,13 @@ impl Storage {
                 let tx_hash_owned = sha2::Sha256::digest(&tx_bytes);
                 let tx_hash = tx_hash_owned.as_slice();
                 let tx_block_height = filtered_block.height as i64;
+                let return_address = transaction.decrypt_memo(&fvk).map_or(None, |x| Some(x.sender.to_vec()));
 
                 tracing::debug!(tx_hash = ?hex::encode(tx_hash), "recording extended transaction");
 
                 dbtx.execute(
-                    "INSERT INTO tx (tx_hash, tx_bytes, block_height) VALUES (?1, ?2, ?3)",
-                    (&tx_hash, &tx_bytes, tx_block_height),
+                    "INSERT INTO tx (tx_hash, tx_bytes, block_height, return_address) VALUES (?1, ?2, ?3, ?4)",
+                    (&tx_hash, &tx_bytes, tx_block_height, return_address),
                 )?;
 
                 // Associate all of the spent nullifiers with the transaction by hash.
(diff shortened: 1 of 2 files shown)


crates/view/src/storage.rs
```diff
@@ -1493,4 +1500,54 @@ impl Storage {
...
+
+        let pool = self.pool.clone();
+
+        let query = "SELECT return_address, denom, amount
+                    FROM tx
+                    JOIN spendable_notes ON spendable_notes.source = tx.tx_hash
+                    JOIN notes ON spendable_notes.note_commitment = notes.note_commitment
+                    JOIN assets ON assets.asset_id = notes.asset_id
+                    WHERE block_height BETWEEN ?1 AND ?2
+                    GROUP BY return_address";
```
cronokirby, line 1520: What does GROUP BY do for Blobs?

cronokirby (commented):
I think this is what we want, although I'm not sure if the GROUP BY in the query does what we want it to do. I think the code should work without it.

crates/view/src/storage.rs
aubrika: Without grouping by the address, you won't filter out asset amounts for the other addresses, because the JOIN does not otherwise take this into account/is not unique to the address. I initially thought the same but testing revealed the need to specify a specific grouping to return the appropriate values per address.

cronokirby: Have we tested the case where you send the exact same amount twice? I think the GROUP BY might clobber that

cronokirby: Like my understanding of how GROUP BY works without aggregation functions is that it's like doing SELECT DISTINCT, which would have this undesired behavior when you see a repeat of (return_addr, denom, amount)

crates/view/src/storage.rs
```diff
@@ -1493,4 +1500,54 @@ impl Storage {
         })
         .await?
     }
+
+    pub async fn amount_by_sender(
+        &self,
+        start_height: Option<u64>,
+        end_height: Option<u64>,
+    ) -> anyhow::Result<BTreeMap<(Address, String), u128>> {
```
hdevalence, line 1508: Why do we want this to do the summarization internally? Wouldn't it be more flexible to have something like
```
notes_by_sender(&self, return_address: &Address) -> Result<Vec<SpendableNoteRecord>>
```
and then let the caller figure out how to interpret the results?

cronokirby: Well, we want the signature to be such that the output contains all the addresses, rather than us having to figure out what addresses we need to pass into the function. This would be a return type of like `Iterator<(Address, SpendableNoteRecord)>` with the address appearing multiple times.

While this might be useful in more situations, right now the single consumer of this API just wants the total amount received. I think we can just punt the issue of providing a more generally useful version of the API once there's more consumers of that API.

cronokirby: Okay actually I think implementing Henry's suggestion makes sense now :p

hdevalence: The key point is that for the coordinator we're making this API for, we only ever want to query for single, specific addresses, and we don't want to have to scan through a potentially unbounded amount of data to run the query.

cronokirby (approved):
Looks good to me!

PR #3100, opened 2023-09-26
ibc: compute IBC heights from revision number derived by chain id

avahowell:
Previously, we hardcoded `0` when constructing IBCHeights (which consist of a revision number, and revision height) to verify that counterparties committed correct penumbra consensus states. This PR instead computes the revision number from the chain ID.

There is an underlying issue here, which is that [ibc-go enforces a relationship between the chain ID and the revision number](https://github.com/cosmos/ibc-go/blob/56a519c188ae841818b9ecd11244240941cc5d90/modules/light-clients/07-tendermint/client_state.go#L140). The revision number is always parsed from the chain ID, matching with a regex that matches chain ids in the form `.*[^\n-]-{1}[1-9][0-9]`. If the matching fails, the revision number is assumed to be zero. This is why, despite our chain ID not fitting this regex, our handshakes worked.

In the future, we can either

* keep non-ibcgo-standard chain IDs, and commit to app_version (revision number) always being zero

or

* amend chain IDs to fit this format, eg `penumbra-testnet-dione-1`

### DIFF
--- a/crates/bin/pd/src/info.rs
+++ b/crates/bin/pd/src/info.rs
@@ -26,10 +26,7 @@ use ibc_types::core::channel::{ChannelId, PortId};
 use ibc_types::core::client::ClientId;
 use ibc_types::core::connection::ConnectionId;
 use ibc_types::core::connection::IdentifiedConnectionEnd;
-use penumbra_chain::{
-    component::{AppHashRead, StateReadExt},
-    APP_VERSION,
-};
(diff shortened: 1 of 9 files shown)


erwanor (approved):
I think we can drop the revision number suffix, and default to zero. My understanding is that its purpose is to disambiguate block numbers for chains that do height-reseting upgrades.

Issue #2263, opened 2023-03-27
Migrate to CometBFT 0.37


conorsch:
The Tendermint project has been forked as CometBFT. We should plan to migrate our use Tendermint accordingly. At first, let's aim 

## Required changes
- [x] Spike on compatibility by swapping cometbft for tendermint binary, joining the testnet-preview deployment. For now, let's keep the version the on the `0.34.x` series.
- [x] Update testnet generate logic to use new config dir paths.
- [x] Update container references to use cometbft, rather than tendermint, container images
- [x] (Later) Update cometbft to latest release series, e.g. `0.37.x`. Check whether LTS channel exists.

## Relevant docs
* [CometBFT announcement post](https://medium.com/the-interchain-foundation/cosmos-meet-cometbft-d89f5dce60dd)
* [CometBFT: Upgrading from Tendermint Core](https://docs.cometbft.com/v0.34/guides/upgrading-from-tm)

hdevalence:
There's currently no LTS channel for CometBFT releases.  We can't easily switch to 0.37.x because it changes the ABCI version, so we'll need to update `tendermint-rs` and `tower-abci` and also change our `Consensus` and `Mempool` implementations to match the new ABCI semantics.

conorsch:
@erwanor has some work in flight that's related to this https://github.com/penumbra-zone/tower-abci/pull/25. To be clear, we're not blocked here: we can still proceed with testing compat on the _current version_. Before we reach for upgrading versions, let's make sure to check compat in e.g. tower-abci.

hdevalence:
Updated issue title to reference a goal of switching to 0.37; the ABCI support should have landed in `tower-abci` by now.

conorsch:
As of https://github.com/penumbra-zone/penumbra/pull/3033, we are running cometbft `v0.34.27` in the preview environment. So far, so good. Updated the checklist. Final remaining task is to increment minor versions, which we'll need to do carefully, making changes to `pd` and/or `tower-abci` where necessary.

erwanor:
The only changes that we will need to do are to `pd`, specifically:

- [ ] use the `v037` module of `tower-abci` and `tower-trace`
- [ ] handle `v037` new messages like `PrepareProposal`, `ProcessProposal`

[...]

PR #3115, opened 2023-09-28
feat(pd)!: migrate to cometbft v0.37

conorsch:
Bumps the rust imports for tendermint-rs and tower-abci to use the v037, rather than v034, modules. The substantive change is the addition of two new ABCI event types:

* PrepareProposal
* ProcessProposal
    
The consensus logic has been updated according to spec [0].
    
Refs #2263.
    
[0] https://github.com/cometbft/cometbft/blob/v0.37.2/spec/abci/abci%2B%2B_comet_expected_behavior.md#adapting-existing-applications-that-use-abci


crates/narsil/narsil/src/ledger/consensus.rs
```diff
@@ -81,6 +81,10 @@ impl Consensus {
                         .await
                         .expect("commit must succeed"),
                 ),
+                // The proposal types were added in CometBFT v0.37.x.
+                // How should they be handled in narsil?
```
erwanor, line 85: The same way as we do in the main application. This can be done later

erwanor (approved):
Nice work!

PR #3118, opened 2023-09-28
ibc: fix connection paths to use ibc-types

avahowell:
IBC on main will be broken until this is merged; we still had some of the ibc standard paths defined in the `penumbra-ibc` crate, which recently got changed. This broke compatibility, but these keys shouldn't even be in this crate. This PR removes the hardcoded `connection` keys and replaces them with the ones exposed by `ibc-types`, and also fixes the `ClientConnections` mapping logic to be consistent with the spec and the ecosystem.

This is blocking on release of a new version of ibc-types which includes a domain type we previously missed, `ClientPaths`

### DIFF
--- a/crates/bin/pcli/Cargo.toml
+++ b/crates/bin/pcli/Cargo.toml
@@ -48,7 +48,7 @@ tendermint = { version = "0.33.0", features = ["rust-crypto"] }
 jmt = "0.7"
 
 # External dependencies
-ibc-types = { version = "0.6.0", default-features = false, features = ["std", "with_serde"]}
+ibc-types = { git = "https://github.com/penumbra-zone/ibc-types.git", branch = "client-paths-type", features = ["std", "with_serde"] }
 
 ibc-proto = { version = "0.33.0" }
 ark-ff = { version = "0.4", default-features = false }
--- a/crates/bin/pcli/Cargo.toml
+++ b/crates/bin/pcli/Cargo.toml
@@ -48,7 +48,7 @@ tendermint = { version = "0.33.0", features = ["rust-crypto"] }
 jmt = "0.7"
 
 # External dependencies
-ibc-types = { git = "https://github.com/penumbra-zone/ibc-types.git", branch = "client-paths-type", features = ["std", "with_serde"] }
+ibc-types = { version = "0.6.3", features = ["std", "with_serde"] }
 
 ibc-proto = { version = "0.33.0" }
 ark-ff = { version = "0.4", default-features = false }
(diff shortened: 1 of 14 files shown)


crates/core/component/ibc/src/component/state_key.rs
```diff
@@ -8,44 +8,17 @@ pub fn ibc_params() -> &'static str {
...
-// TODO (ava): move these to ibc-types eventually
+// these are internal helpers that are used by penumbra-ibc, but not part of the IBC spec (that is,
+// counterparties don't expect to verify proofs about them)
 pub fn client_processed_heights(client_id: &ClientId, height: &Height) -> String {
-    format!("ibc/clients/{client_id}/processedHeights/{height}")
+    format!("clients/{client_id}/processedHeights/{height}")
 }
 pub fn client_processed_times(client_id: &ClientId, height: &Height) -> String {
-    format!("ibc/clients/{client_id}/processedTimes/{height}")
+    format!("clients/{client_id}/processedTimes/{height}")
```
hdevalence, line 17: if these aren't to be used by counterparties, and are component-internal, should we be prefixing them like the other component keys?

crates/bin/pcli/Cargo.toml
```diff
@@ -48,7 +48,7 @@ tendermint = { version = "0.33.0", features = ["rust-crypto"] }
 jmt = "0.7"
 
 # External dependencies
-ibc-types = { version = "0.6.0", default-features = false, features = ["std", "with_serde"]}
+ibc-types = { version = "0.6.3", features = ["std", "with_serde"] }
```
hdevalence, line 51: side point, but, i think we should change the `ibc-types` crate to make the feature be called `serde` rather than `with_serde`, since that's a more conventional name

hdevalence (approved):
LGTM modulo path nits!

crates/bin/pcli/Cargo.toml
erwanor: @hdevalence Cargo gets confused when using colliding feature names that overlap with an optional dependency, even if you specify `dep:serde`, I never was able to find a way to bundle the `serde = [dep:serde, dep:serde_derive]` etc. maybe it's a skill issue.

hdevalence: yeah, i remember it being annoying to figure out. it seems potentially worth it longer-term though.

erwanor: Cargo would not complain about it but the end result is that, somehow, `serde_derive` would not be available to the consumer crate (internal to the ibc-types graph)

Issue #3129, opened 2023-10-01
Add substore support to `penumbra-storage`


hdevalence:
The IBC protocol requires that IBC data is recorded at specific merkle tree paths, so that all counterparties can have a standardized way to verify each others' messages. To provide domain separation, IBC implementations are allowed to specify a prefix that separates their IBC data from the rest of the state hashed upwards to their light-client-verifiable state root. 

The way this is accomplished, however, poses complications.  Rather than being just a prefix on the _keys_, the prefix is a complete sub-tree. IBC data is expected to be in a separate subtree ("substore"), whose root hash is recorded as a leaf of a higher-level state tree. Inclusion proofs are lists of multiple individual proofs, chaining verification up to the global root. This is actually required by `ibc-go`, which will reject proof specifications without at least two proofs.

While this is natural for the IAVL tree used by the SDK, it's less natural for us, using the JMT, which provides a flat keyspace. To work around this, so far we've created a bogus, depth-1, single-node "merkle tree" that re-hashes the JMT root. But not only is this ugly, it creates complications:

* Our app hash is no longer the JMT root, but some derived data. So we have to have code to manage this and layer Penumbra-specific notions of app hashing onto the JMT root. This also means we can't just have the k/v RPC methods in the storage crate (cf #3091) because they return ICS23 proofs. This means that none of those RPC implementations are directly reusable by Astria.

* We don't have domain separation between IBC data and other Penumbra data. If some future IBC protocol added new data at conflicting paths, we'd have an unresolvable problem. While we could add a custom domain separation to all of _our_ data, this is quite ugly.

One idea to resolve this problem is to add transparent substore support to `penumbra-storage`. Roughly speaking, this would be a way to "mount" JMT subtrees at specific, fixed prefixes of the tree:

- instead of `Storage::load` doing implicit init, we'd have explicit init
- initialization would specify a list of "substore prefixes". these would only be configurable on init and never afterwards, and they'd be stored in rocksdb itself, so `load()` can be used without reciting substore configuration.
- the `StateRead`/`StateWrite` interfaces are unaffected, and the storage continues to present one namespace, just as before. nothing changes about any caching layers.
- in the base layer of the stack, reads are compared against the substore prefix list. if they match, the prefix is removed and the suffix is routed to the substore instead of to the default store.
- writes are also intercepted, if new keys match the prefix, the suffix gets routed to the substore.
- substore writes happen first, so that the updated substore roots can be inserted as writes to the main store.
[...]

PR #3131, opened 2023-10-02
storage: implement substore support

hdevalence:
Towards #3129, this PR implements support for prefixed substores.

A substore is an independent section of storage parametrized over a prefix, and a set of column families. It embeds its own independent merkle tree, flat KV store (nonverifiable storage), and key preimage index.

A substore has three logical components:
- a stateless configuration that specifies its scope and columns: `penumbra_storage::store::SubstoreConfig`.
- a read-only view to a specific version of its state: `penumbra_storage::store::SubstoreSnapshot`.
- a write-handle to its storage instance: `penumbra_storage::store::SubstoreStorage`.

The default store (aka. "main store") has an empty prefix and hosts both non-prefixed values and substore root hashes stored at their respective prefix. For example, consider this simplified layout:

```
/// Main store layout:
/// ┌────────────────┬──────────────────┐
/// │                │    merkle tree   │
/// │ empty prefix   │                  │
/// │                │       Root R0    │
/// └────────────────┴─────────┬────────┘
///                ┌───────────┼───────────┐
///                ▼           ▼           ▼
///   ┌─────┬───────────┐    ┌───┐  ┌─────┬───────────┐
///   │ibc/ │    R1     │    │   │  │misc/│     R2    │
///   └─────┴─────┬─────┘    └┬─┬┘  └─────┴─────┬─────┘
///               │           │ │               │
///          ┌────┼───┐       │ │            ┌──┴──┐
///          │    │   │       ▼ ▼            │     │
///          ▼    ▼   ▼       ▩ ▩            ▼     │
///        ┌───┐  ▩   ▩                      ▩     ▼
///        └┬─┬┘                                 ┌───┐
///         │ │                                  └┬─┬┘
///         │ │                                   │ │
///         ▼ ▼                                   ▼ ▼
///         ▩ ▩                                   ▩ ▩
```

In addition to non-prefixed keys, the main store hosts two substores: one with prefix `ibc/` and the other with prefix `misc/`. These prefixes are both defining a namespace and are actual keys that map to the root hash of their respective merkle trees.

Rather than requiring consumers to track storage namespaces, we perform routing transparently. When querying a key from the JMT, we determine whether the key belongs to a substore, and if so, we fetch the latest known version of that store. This means that state access interfaces do not change.

Versioning is done on a per-substore basis, each batch of writes yielding an increase in version numbers.

Writes are performed atomically, using `rocksdb::WriteBatch`. Substore writes are performed first, so that we can then collect their root hashes and store them at their respective prefix key in the main store.


hdevalence:
(If someone wants to try running with this idea for a bit, feel free, just be sure to push up WIP changes as you go).

crates/storage/src/store/substore.rs
```diff
@@ -0,0 +1,233 @@
...
+        Self {
+            // TODO: harmonize cf format throughout.
+            cf_jmt: "jmt".to_string(),
+            cf_jmt_keys: "jmt_keys".to_string(),
+            cf_jmt_values: "jmt_values".to_string(),
+            cf_jmt_keys_by_keyhash: "jmt_keys_by_keyhash".to_string(),
+            cf_nonverifiable: "nonverifiable".to_string(),
+            prefix: "".to_string(),
+        }
+    }
```
hdevalence, line 67: Why do we need this? Why can't we just call `new()` with an empty string? We're breaking the disk format anyways to change the tree shape, we don't need to preserve existing column families -- everything that does access via column family should be running through a substore, or it will be broken for other reasons.

crates/storage/src/store/substore.rs
```diff
@@ -0,0 +1,233 @@
...
+    // writing to every substore atomic.
+    pub fn _commit(&self, _changeset: ()) -> Result<()> {
+        todo!("commit changeset to rocksdb")
+    }
+}
+
+pub struct SubstoreSnapshot {
+    pub config: Arc<SubstoreConfig>,
+    pub snapshot: Snapshot,
+}
```
hdevalence, line 129: We could add a note that this is intended to be constructed on-demand.

crates/storage/src/storage.rs
```diff
@@ -122,13 +125,19 @@ impl Storage {
...
                     })))
                 })
             })?
             .await?
     }
 
+    // TODO: hack, will consolidate later in the pr.
+    pub async fn load(path: PathBuf) -> Result<Self> {
+        Storage::init(path, vec![]).await
+    }
```
hdevalence, line 139: We'll want this to read the prefixes out of a config column family, right?

crates/storage/src/snapshot.rs
```diff
@@ -43,18 +45,30 @@ impl std::fmt::Debug for Snapshot {
...
+    pub(crate) fn new(
+        db: Arc<rocksdb::DB>,
+        version: jmt::Version,
+        mut substore_configs: Vec<Arc<store::substore::SubstoreConfig>>,
+    ) -> Self {
+        if substore_configs.is_empty() {
+            let transparent_config = store::substore::SubstoreConfig::transparent_store();
+            substore_configs.push(Arc::new(transparent_config));
+        }
+
```
hdevalence, line 66: I could be missing something but this doesn't seem right to me. Why are we creating a new config object here?

crates/storage/src/store/multistore.rs
```diff
@@ -0,0 +1,69 @@
+use std::sync::Arc;
+
+use super::substore::SubstoreConfig;
+
+/// A collection of substore, each with a unique prefix.
+#[derive(Debug, Clone)]
+pub struct Multistore {
```
hdevalence, line 7: Should this be called `MultistoreConfig`, since it just has the configs and isn't a store in itself?

crates/storage/src/store/multistore.rs
```diff
@@ -0,0 +1,69 @@
...
+    transparent_store: Arc<SubstoreConfig>,
```
hdevalence, line 8: I'm not sure `transparent_store` is the best name here. Maybe `root_store` would be better, since it's at the root of the tree?

crates/storage/src/store/multistore.rs
```diff
@@ -0,0 +1,69 @@
...
+    substores: Vec<Arc<SubstoreConfig>>,
+}
+
+impl Multistore {
+    pub fn new(mut substores: Vec<Arc<SubstoreConfig>>) -> Self {
+        assert!(
+            !substores.is_empty(),
+            "multistore must have at least one substore"
+        );
```
hdevalence, line 17: Why do we want to require this?

For that matter, why do we have a constructor at all -- since the `MultistoreConfig` is set on storage configuration, we should never be creating more than one, so it can be created once just with struct initialization in the part of the code that loads the configuration.

crates/storage/src/store/multistore.rs
```diff
@@ -0,0 +1,69 @@
...
+        }
+    }
+
+    pub fn _get(&self, prefix: &str) -> Option<Arc<SubstoreConfig>> {
+        if prefix.is_empty() {
+            Some(self.transparent_store.clone())
+        } else {
+            self.substores.iter().find(|s| s.prefix == prefix).cloned()
+        }
+    }
```
hdevalence, line 31: What do we do with this? When do we want to return `None` here, isn't the point of the root store that it handles the case where no prefix matches?

hdevalence: (I guess this also applies to the `find_substore` method below, though in that case the question is whether we want the caller to have to deal with filling in the root store, or whether the routing method should do that).

crates/storage/src/store/multistore.rs
```diff
@@ -0,0 +1,69 @@
...
+                    .expect("key has the prefix of the matched substore"),
+                config,
+            ),
+            None => (key, self.transparent_store.clone()),
+        }
+    }
+
+    /// Route the key to the correct substore, or the transparent store if no prefix matches.
+    /// Returns the truncated key, and the target snapshot.
+    pub fn route_key<'a>(&self, key: &'a [u8]) -> (&'a [u8], Arc<SubstoreConfig>) {
```
hdevalence, line 59: maybe this should be `route_key_bytes` for symmetry with the above? i think a bit of duplication is cleaner here per your comment above.

crates/storage/src/store/substore.rs
```diff
@@ -0,0 +1,233 @@
...
+
+impl SubstoreSnapshot {
+    fn rocksdb_snapshot(&self) -> &RocksDbSnapshot {
+        &self.snapshot.0.snapshot
+    }
+}
```
hdevalence, line 135: The existence of this method suggests that the data modeling for the `SubstoreSnapshot` is wrong, and that it would be better to have it just hold on to the `RocksDbSnapshot`. That way we don't have this ugly pass-the-parent-container-to-the-sub-object pattern of cloning the snapshot inside itself.

crates/storage/src/storage.rs
```diff
@@ -36,15 +36,20 @@ struct Inner {
...
     jh_dispatcher: Option<tokio::task::JoinHandle<()>>,
     db: Arc<DB>,
 }
 
 impl Storage {
-    pub async fn load(path: PathBuf) -> Result<Self> {
+    /// Initializes a new storage instance at the given path, along with a substore
+    /// defined by each of the provided prefixes.
+    /// TODO: potentially restore the `impl IntoIterator` if it makes sense later in the pr.
+    pub async fn init(path: PathBuf, store_prefixes: Vec<String>) -> Result<Self> {
```
hdevalence, line 50: nit: should this be `substore_prefixes`, since they're defining the substores?

crates/storage/src/snapshot.rs
```diff
@@ -293,36 +328,41 @@ impl StateRead for Snapshot {
 
     fn nonverifiable_prefix_raw(&self, prefix: &[u8]) -> Self::NonconsensusPrefixRawStream {
         let span = Span::current();
-        let self2 = self.clone();
+        let snapshot = self.clone();
 
+        let (prefix, config) = self.0.multistore.route_key(prefix);
```
hdevalence, line 333: For the prefix queries, I think we need to do another check: we need to ensure that the iteration doesn't cross a "mount point". I think we can do this by banning queries supplying prefixes that are themselves prefixes of mount points, returning an error instead.  (I don't think this is a meaningful restriction, maybe it means we can't iterate over every single key but I think that's OK)

crates/storage/src/snapshot.rs
```diff
@@ -112,8 +126,8 @@ impl Snapshot {
     /// Reads from the JMT will fail if the root is missing; this method
     /// special-cases the empty tree case so that reads on an empty tree just
     /// return None.
-    fn get_jmt(&self, key: jmt::KeyHash) -> Result<Option<Vec<u8>>> {
-        let tree = Sha256Jmt::new(&*self.0);
+    fn get_jmt(&self, key: jmt::KeyHash, substore: SubstoreSnapshot) -> Result<Option<Vec<u8>>> {
+        let tree = Sha256Jmt::new(&substore);
```
hdevalence, line 130: Why is this a method on `Snapshot` taking the `SubstoreSnapshot` that will actually be used, instead of moving the method out to the `SubstoreSnapshot`? Wouldn't it be cleaner to move all of the tree accesses out to the `SubstoreSnapshot`?

crates/storage/src/snapshot.rs
```diff
@@ -69,73 +77,103 @@ impl Snapshot {
...
-        let span = Span::current();
-        let snapshot = self.clone();
-
+        let span = tracing::Span::current();
+
+        let (_, substore_config) = self.0.multistore.route_key_bytes(&key);
+        let substore = store::substore::SubstoreSnapshot {
+            config: substore_config,
+            rocksdb_snapshot: self.0.snapshot.clone(),
+            version: self.version(),
```
erwanor, line 86: note: we are going to keep separate versioning per substore/tree, the strategy should be something along the lines of keeping track of (prefix -> version) after writes, possibly persisting them on disk.

crates/storage/src/snapshot.rs
```diff
@@ -69,73 +62,81 @@ impl Snapshot {
...
     /// Returns the root hash of this `State`.
     ///
     /// If the `State` is empty, the all-zeros hash will be returned as a placeholder value.
     ///
     /// This method may only be used on a clean [`State`] fork, and will error
     /// if [`is_dirty`] returns `true`.
-    pub async fn root_hash(&self) -> Result<crate::RootHash> {
-        let span = Span::current();
-        let snapshot = self.clone();
+    pub async fn prefix_root_hash(&self, prefix: &str) -> Result<crate::RootHash> {
```
hdevalence, line 97: Can we update this comment? We don't even have a `State` any more ...

crates/storage/src/snapshot.rs
```diff
@@ -69,73 +62,81 @@ impl Snapshot {
...
+        let rocksdb_snapshot = self.0.snapshot.clone();
+        let db = self.0.db.clone();
+
+        let config = self
+            .0
+            .multistore_cache
+            .config
+            .find_substore(prefix.as_bytes());
+
+        // If a substore is not found, we default to the main store.
```
hdevalence, line 108: I was initially confused by this comment, I think it would be clearer if it read

> If a substore is not found, find_substore will default to the main store

to clarify what exactly is doing the defaulting (it's the part of the code that returns `config`)

crates/storage/src/snapshot.rs
```diff
@@ -238,104 +280,150 @@ impl StateRead for Snapshot {
...
             })
             .expect("should be able to spawn_blocking");
 
-        tokio_stream::wrappers::ReceiverStream::new(rx)
+        tokio_stream::wrappers::ReceiverStream::new(rx_prefix_query)
     }
 
+    /// Returns a stream of all key-value pairs with the given prefix, and range
+    /// from nonverifiable storage.
+    /// TODO(erwan): For now this method only supports range queries over the main store.
```
hdevalence, line 398: It's further afield, but one possible future use case could be to put all of the DEX indexes in a substore, so that DEX queries would operate only on DEX data. But I don't think this is currently the lowest hanging fruit for performance there.

crates/storage/src/storage.rs
```diff
@@ -36,50 +38,149 @@ struct Inner {
     tx_dispatcher: watch::Sender<Snapshot>,
     tx_state: Arc<watch::Sender<Snapshot>>,
     snapshots: RwLock<SnapshotCache>,
+    multistore_config: MultistoreConfig,
     #[allow(dead_code)]
     /// A handle to the dispatcher task.
+    /// This is used by `Storage::release` to wait for the task to terminate.
```
hdevalence, line 44: if it's used, why do we need `allow(dead_code)`?

hdevalence (commented):
Thanks @erwanor for carrying this forward. There are a lot of changes here but they look good to me. There are a lot of new tests. It seems like the smoke test indicates that we can drop this in without problems.

Is there any reason we should not merge this now, and then attempt to remove the penumbra-specific apphash in favor of an IBC substore?

crates/storage/src/snapshot.rs
erwanor: Good point, that would be neat. Prefix queries work with substores, but the more general range queries don't yet. There isn't a hard blocker, but it requires massaging the API. It's tracked in [#3327](https://github.com/penumbra-zone/penumbra/issues/3327).

crates/storage/src/storage.rs
erwanor: It's gated on `#[cfg(test)]`. This is used to implement the async drop that we use to await shutting down the dispatcher task and releasing the RocksDB lock.

erwanor:
Thanks for the review, really appreciate it. I'll wait for CI to pass before merging. Next steps like adjusting the ICS23 proof api are scoped in [this comment](https://github.com/penumbra-zone/penumbra/issues/3129#issuecomment-1810013873).

Issue #3107, opened 2023-09-26
Re-enable governance param change proposals


zbuc:
In the refactoring efforts for #3084/#3085 it was simpler to disable the ability to submit param change proposals for now, since splitting apart the configurable parameters per-component introduces more complexity.

These should be re-enabled after the refactoring work is complete. This will require redesigning the proposal storage and proposal RPCs in the governance component.

- [ ] grep the codebase for references to `3107` and ensure all functionality is re-added
- [ ] split the param change validation per-component [crates/core/app/src/params/change.rs](https://github.com/penumbra-zone/penumbra/pull/3086/files/26314039d08bb61d1fbe98447c053817abdc4b1b#diff-1a67e737348dc179b290adb5e1a05057a0d10deec233c67dc23ecd171e6167ef)

conorsch:
Optimistic about completion for Testnet 63.

conorsch:
We have a CI regression since merge of #3212: https://github.com/penumbra-zone/penumbra/actions/runs/6566057359/job/17835924660 it's flagging allegedly unused protos.

PR #3133, opened 2023-10-02
Wrapup of multidimensional gas fees

### LINKED ISSUE #2970: Groundwork for Multidimensional Gas Fees

Processing transactions consumes resources. To account for this, transactions should pay fees. Currently, we've designed a way for Penumbra transactions to include a fee payment, but we haven't designed the system that controls what those fees should be.

Penumbra transactions can consume different types of resources. For instance, proof verifications are parallelizable and purely CPU-bound, while IBC packets may require sequential, stateful accesses. Most importantly, new output notes create work not just for every _node_, but every _user_, who must scan the new notes to determine whether they are relevant (or offload that scanning work to a server, using fuzzy message detection).

These resources have different capacity limits, and those capacity limits show themselves at different time scales. For instance, Zcash experienced a DoS attack where an adversary posted many transactions that included many outputs. While these transactions didn't exceed the available block space, they did overwhelm other clients' effective scanning capacity, cause all other users to have degraded UX.  (See [this discussion](https://twitter.com/hdevalence/status/1597115071033118722) for details).

In Penumbra's context, we can express the problem as follows: while _block space_ describes synchronization bandwidth between full nodes, _compact block space_ describes synchronization bandwith among client devices. On long time scales, compact block space is much more constrained than block space, because client devices are assumed to have much more restrictive bandwidth constraints than full nodes, but on short time scales, there's no particular constraint -- a client trying to synchronize the last week's worth of activity doesn't particularly care about whether it was evenly spread or concentrated in a few bursts.

So the resources we need to account for differ not just in their capacity but also in their time horizons. How can we handle this? One solution is provided by the [multidimensional resource pricing](https://baincapitalcrypto.com/multi-dimensional-on-chain-resource-pricing/) paper, which shows why multidimensional gas fees are better than a scalar gas quantity, and how to automatically price different resources according to actual utilization.

We should implement this for Penumbra, in a phased approach that lets us frontload foundational structure and backload whatever implementation complexity we can get away with.  I think we can decompose the work into three phases:

_Resource Selection_
 
First, what resources do we need to price? Based on discussions with the team, it seems like the following is a reasonable starting point:

1. **Block Space**, a proxy for synchronization bandwidth between full nodes.
2. **Compact Block Space**, a proxy for synchronization bandwidth for clients.
3. **Verification**, a proxy for parallelizable, CPU-bound work.
4. **Execution**, a proxy for non-parallelizable stateful execution.

_Gas Schedules_

In the `transaction` crate:

- [x] Define a `Gas` structure to represent a multidimensional gas vector.

This should look like
```rust
struct Gas {
    pub block_space: u64,
    pub compact_block_space: u64,
    pub verification: u64,
    pub execution: u64,
}
```
It should have an `Add` impl to allow adding gas costs.

- [x] Define a `GasCost` trait that assigns a gas cost to each `Action`.

Making this a standalone trait means we can put all the impls in one place, so we have one file declaring the entire gas cost schedule.  We should make the gas costs statically determined, because that's easier for clients to reason about, and we should avoid allowing transactions to do arbitrary work on-chain anyways, so static gas costs are an additional forcing function.

The `GasCost` trait should be implemented for each `Action` and for the `Transaction` overall.  The `Transaction` implementation should sum up all of the constituent actions' gas costs and add its own contribution. This way, we can have each `Action`'s implementation have `block_space: 0` and handle the block space in the `Transaction` impl by using the encoded size of the transaction. (Yes, this means wastefully re-encoding. We can deal with that if and when it shows up in a profile).

For the gas schedule itself, we can just make up some numbers and fine tune them later. Because gas needs to be multiplied by prices, only the relative costs of different actions matter, the rest gets factored into the prices. As a rough starting point, what about:

- Use the encoded size (in bytes) of whatever data the action adds to the compact block as the compact block space;
- Use `1000` as the `verification` cost for anything with a snark and `0` otherwise;
- Use `10` for `execution` cost.

Later, we can fine-tune these.

- [x] Define a `GasPrices` proto and domain type 

This should also be a four-dimensional vector, but instead it expresses the price of each unit of gas in terms of the staking token:
```rust
struct GasPrices {
    pub block_space_price: u64,
    pub compact_block_space_price: u64,
    pub verification_price: u64,
    pub execution_price: u64,
}

impl GasPrices {
    pub fn price(&self, gas: &Gas) -> Amount {
        // dot product
    }
}
```

- [x] Add hardcoded gas prices as a chain parameter

To start, we can hardcode the gas prices as a chain parameter, and have the default parameter choice be all-zero. This lets us experiment with nonzero fees on devnets (by editing the generated parameters in the `genesis.json`) prior to committing to the change and potentially breaking client software.

- [x] Add gas checks to the transaction-level fee logic.

The chain logic should enforce that the transaction-level fee is greater than or equal to the _base fee_, which is computed as the transaction's gas use, priced by the current gas prices. (Initially, gas prices are static, so the static prices are always current).  For now, we can continue to do nothing with the fee, effectively burning it, though we may want to revisit this later once there's a working distributions component (at which point we can distribute everything above the base fee to the proposer, giving an in-protocol first-price auction for block space).

- [x] Add an RPC for fetching the current gas prices.

This just amounts to adding the GRPC method to the `SpecificQuery` RPC; the initial implementation can return the hardcoded chain parameters. Although the implementation is trivial, adding it as an RPC sets the expectation that client software should query for gas prices (so that when we make them dynamic later, no changes are required).

- [x] Change the `TransactionPlanner` to account for fees.

This means the `TransactionPlanner` will need to be configured with a `GasPrices` at the beginning. If none is supplied, it should internally set an all-zero gas prices (to retain compatibility with current API use).

We want to retain the existing two-phase planner, which is convenient for use in the wasm context where the planner code may not have state access.  In the current planner API, the caller specifies a list of constraints on a planner object, and the planner object returns information about what it needs to balance the transaction (i.e., what notes it needs to be able to spend).

This potentially poses a problem: if the planner estimates gas based on the constraints, but those constraints are satisfied by adding additional actions, the resulting transaction may underpay fees. One can imagine pathological cases where this cascades over multiple rounds, etc.

As a workaround that would be much simpler and probably work in 99% of cases is to just "overbudget" the estimated gas based on the provided constraints, and then adjust it into change outputs according to the actual gas use.  For instance, we could multiply the estimated gas by 2x or 3x, request funds to balance the transaction with overpaid gas, and then refund any excess into the transaction's existing change output in the second phase of planning.  (In particular, we expect that outputs are much more expensive than spends, since they consume >5x the compact block space, so satisfying funding constraints should be much less expensive than adding them).

**Followup work**

At this point, we've laid the groundwork for multidimensional gas prices, but ensured that the prices are all set to zero (except potentially in a local devnet with customized parameters). This means that we don't need to change anything about the client code for it to keep working.

In a later phase, we should follow on this groundwork:

- [ ] #2971
- [ ] #2972

zbuc:
* Add hardcoded gas prices as a fee component parameter
* Add gas checks to the transaction-level fee logic.
* Allow fetching `fee_params` in the `AppParameters`.
* Include gas prices in the compact block. 
* Change the TransactionPlanner to account for fees.

Closes #2970


hdevalence:
> Allow fetching `fee_params` including current gas prices in the `AppParameters`.

Two design questions here:

1. I think we should have an RPC that fetches the current gas prices specifically, defined in the fee component.
2. I'm not sure we should have the gas prices be part of the `AppParameters`.

The reasoning behind both of these is that if we do automatic gas pricing based on resource usage, we'll be adjusting the gas prices dynamically, and the chain may be frequently editing the prices. In that case, it would be a bit incongruous for the `AppParameters` to have some data that's chosen by humans (either in source code or governance) and changing infrequently, together with other data that's chosen automatically.  And, in that case, suppose that we have parameters that control the auto-updater for the prices. If the gas prices themselves are part of the `AppParameters`, where do we put the higher-level parameters that control them? In the same structure? To the side? Also, if the gas prices change reasonably often, someone might be querying them frequently, in which case it'd be easier and more efficient to just get the data they need.

So, even though we're punting on actually implementing an auto-pricing mechanism, I think we should plan with it in mind, and I think it would be cleaner if we thought of the gas prices as being data, rather than parameters.

zbuc:
Sure, I implemented them as parameter based on this comment from the original issue:

> Add hardcoded gas prices as a chain parameter
> To start, we can hardcode the gas prices as a chain parameter, and have the default parameter choice be all-zero. This lets us experiment with nonzero fees on devnets (by editing the generated parameters in the genesis.json) prior to committing to the change and potentially breaking client software.

It sounds like we want to change the design and remove them from being parameters?

zbuc:
FWIW I don't have a strong opinion on whether they should be parameters or not -- my gut says it probably fine and even if they are managed automatically in the future that isn't necessarily a reason to exclude them from being parameters; we already have other parameters that aren't controllable by governance.

hdevalence:
Sounds good. Let's leave them as parameters for now and iterate on it later.

crates/bin/pcli/src/command/tx.rs
```diff
@@ -289,32 +263,50 @@ impl TxCmd {
...
-                )
-                .await?;
+                let params = app
+                    .view
+                    .as_mut()
+                    .context("view service must be initialized")?
+                    .app_params()
+                    .await?;
+
+                let gas_prices = params.fee_params.gas_prices;
```
hdevalence, line 273: Can we lift this stanza out of the match statement so that we don't have to repeat it in every command?

Also, is it possible for the planner to know about the gas prices without having to have an extra round trip communication with the view service? I think the answer is probably no, right, because the current logic over-budgets gas in order to cover for spends that may be required to fill in the rest of the transaction?

crates/bin/pcli/src/command/tx.rs
```diff
@@ -356,27 +348,39 @@ impl TxCmd {
...
-                planner.fee(swap_fee);
-                planner.swap(input, into.id(), swap_claim_fee.clone(), claim_address)?;
+                planner.set_gas_prices(gas_prices.clone());
+                // The swap claim requires a pre-paid fee, however gas costs might change in the meantime.
+                // This shouldn't be an issue, since the planner will account for the difference and add additional
+                // spends alongside the swap claim transaction as necessary.
+                //
+                // Since the swap claim fee needs to be passed in to the planner to build the swap (it is
+                // part of the `SwapPlaintext`), we can't use the planner to estimate the fee and call the
+                // helper method directly.
```
hdevalence, line 380: I think a gas adjustment mechanism like the Cosmos SDK uses might also help here. On SDK chains it's common to estimate the gas prices client side, then multiply the estimate by a fudge factor (I've seen 1.3, commonly) to account for the possibility that the estimate is wrong. In our case, we have perfect estimation, but possibly evolving fees.

We could multiply the estimated fee by a gas adjustment, to insulate against the possibility that the prices change.

crates/bin/pcli/src/command/tx.rs
```diff
@@ -289,32 +263,50 @@ impl TxCmd {
...
-                    AddressIndex::new(*from),
-                    Some(memo_plaintext),
-                )
-                .await?;
+                let params = app
+                    .view
+                    .as_mut()
+                    .context("view service must be initialized")?
+                    .app_params()
+                    .await?;
```
hdevalence, line 271: This is pulling the gas prices from the `AppParams` stored in the view service. How does the view service have the current `AppParams`?

I believe the `CompactBlock` currently contains a way for the full node to signal changes in parameters by including a new `AppParams` in the compact block. But this design was a compromise in the case where parameter changes were rare and infrequent, and would be much more problematic if the chain were to frequently adjust the gas prices.

In that case, we'd end up filling the compact blocks with the complete history of changes to gas prices, increasing bandwidth costs for all clients, and supplying them with lots of information they don't care about -- the intermediate history of gas prices is useless, only the gas prices at the moment they submit the application are relevant.

So, I wonder if we should be fetching the latest gas prices from the full node, rather than the view service, and not baking in an assumption that the view server has synchronized the current gas prices.

Alternatively, we could decide that the bandwidth cost of gas price changes isn't a big deal, and decide that the view service should synchronize gas prices.

hdevalence:
Branching off of https://github.com/penumbra-zone/penumbra/pull/3133/files#r1344827368, here's a further rabbithole to go down: how do gas prices affect transaction linkability?

Historically, one of the reasons that Zcash was resistant to having a fee market was that the fee is necessarily revealed in the clear, and so it's a vector to leak information about the client. For instance, if different clients have different logic for fee selection, it might be possible to infer which client was used to create a specific transaction, or to link two transactions by the same user (if that user overrode the fee amount in a unique way shared only among those transactions).

We're exposed to this because of the ability to set a custom fee amount. However, we do need to price resources, to avoid the DoS problems later experienced with Zcash (and with the perspective that a fee market _will_ exist, regardless of whether we'd like it to or not), and so we should take a harm reduction approach.

One way that we could reduce the anonymity risk of setting a custom fee amount would be to use the fact that we have static gas amounts, and to have a few preset values for "low" "medium" "high" that would have different multiples of the base price (e.g., low = 1.05x, medium = 1.3x, high = 2x), rather than an arbitrary setting.

However, while we can quantize the multiple into a few settings, this still allows an observer who sees the transaction to infer what prices were used by that user's client, and hence learn information about when they started planning the transaction. Consider, for instance, the "randomized TWAP" use case, where a user wants to perform a swap over some time period without revealing their complete flow, by splitting their trade into sub-trades with randomized sub-amounts. The `TransactionPlan` mechanism allows all sub-trades to be planned and authorized by the user at once, storing a collection of pre-authorized transaction plans that can be proved just before submission so that the time of authorization isn't revealed by the transaction contents. But auto-adjusting gas prices potentially messes this up, since the plan has to include what fee it will pay, and computing the fee requires using a specific choice of gas prices.

Perhaps this suggests that we should try to change prices less often than every block, to have longer chunks of time in which gas prices don't change. On the other hand, that messes up the resource pricing goal.  Another solution could be to apply some kind of quantization _after_ computnig the fees, so that if gas prices are slightly different the fee would still be the same, although this also seems potentially problematic to define in advance (how to we know how much to quantize without knowing what the prices will be?)

hdevalence:
This looks good, I think there might be some init logic happening, since the smoke test fails to query app parameters:
```
Error: status: Unavailable, message: "error getting fee parameters: Missing FeeParameters", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "date": "Tue, 03 Oct 2023 21:24:12 GMT", "content-length": "0", "access-control-allow-credentials": "true", "vary": "origin", "vary": "access-control-request-method", "vary": "access-control-request-headers", "access-control-expose-headers": "grpc-status,grpc-message,grpc-status-details-bin"} }
2023-10-03T21:24:12.650217Z  INFO abci:BeginBlock{height=111}: pd::consensus: beginning block time=Time(2023-10-03 21:24:12.100867336)
2023-10-03T21:24:12.737014Z  INFO abci:Commit: pd::consensus: committed block app_hash=AppHash("c1443b696ab9e40fa4be17aa96fbed67a98245377b3895865149edfc80ecdc23")
Error: pclientd exited early: ExitStatus(unix_wait_status(256))
FOct 03 21:24:12.933  INFO pclientd: starting pclientd opt.home="/tmp/.tmpaNGjYD" config.bind_addr=127.0.0.1:8081 config.grpc_url=http://127.0.0.1:8080/
Error: status: Unavailable, message: "error getting fee parameters: Missing FeeParameters", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "date": "Tue, 03 Oct 2023 21:24:12 GMT", "content-length": "0", "access-control-allow-credentials": "true", "vary": "origin", "vary": "access-control-request-method", "vary": "access-control-request-headers", "access-control-expose-headers": "grpc-status,grpc-message,grpc-status-details-bin"} }
2023-10-03T21:24:13.198481Z  INFO abci:BeginBlock{height=112}: pd::consensus: beginning block time=Time(2023-10-03 21:24:12.648688611)
```

crates/bin/pcli/src/command/tx.rs
zbuc: > I believe the CompactBlock currently contains a way for the full node to signal changes in parameters by including a new AppParams in the compact block.

It _did_ but this was broken (see #3107). I think you raise a valid issue in that updating the `AppParams` in the compact block _requires storing the entire `AppParams` structure in the compact block_. If only the gas prices are changing, and they are changing on a block-by-block basis, this is rather expensive.

Another potential alternative (that might be desirable when implementing #3107) is to have the param change updates in the compact block be at the component-level. That would mean we don't need to store a complete `AppParams` for a gas prices change. Or we poll the gas prices directly from the full node. 🤷

hdevalence: What if:

1. We pull the gas prices out of the `AppParameters`, so that we can have a separation between "governance-controlled parameters that change infrequently and that clients can be updated on" and "auto-controlled prices that change frequently"
2. We add the gas prices into the `CompactBlock`, and decide that the chain will stream them to the client when they change.

This way we don't have to do a bigger architectural change about per-component parameter change updates etc. I think that the bandwidth costs might be pretty minimal, we could do some back of the envelope estimates.

hdevalence:
Unfortunately, adding noise only to the RTWAP transactions doesn't really help, because then they're distinguishable from other transactions that would use quantized fee amounts.  We could make all transactions have noisy fee levels, but this is also tricky, because that opens up a large surface area for implementation fingerprinting (what if different clients add noise in different ways that turn out to be statistically distinguishable?).

One possibility that might be workable is to stick with preset tiers for gas adjustments (e.g., "low" = 1.05x base, "medium" = 1.3x base, "high" = 2x base), but instead of using fixed multipliers, define specific distributions for clients to draw from to choose a multiplier. So setting "medium" would cause your client to sample a random value near 1.3, using the same distribution as other people's clients.

The hope would then be that slight variation in the gas prices used to compute the fee amount would hide in the noise of the randomized adjustment. This wouldn't be the case if prices changed very rapidly, but if they change only a little bit, it might work.

For a first pass, though, I think we should stick to fixed tiers and figure out randomization later.

zbuc:
@hdevalence can you expand on this a bit further:

> However, while we can quantize the multiple into a few settings, this still allows an observer who sees the transaction to infer what prices were used by that user's client, and hence learn information about when they started planning the transaction. Consider, for instance, the "randomized TWAP" use case, where a user wants to perform a swap over some time period without revealing their complete flow, by splitting their trade into sub-trades with randomized sub-amounts. The TransactionPlan mechanism allows all sub-trades to be planned and authorized by the user at once, storing a collection of pre-authorized transaction plans that can be proved just before submission so that the time of authorization isn't revealed by the transaction contents. But auto-adjusting gas prices potentially messes this up, since the plan has to include what fee it will pay, and computing the fee requires using a specific choice of gas prices.

Even in the case that the various transactions constituting the TWAP are linkable via the gas price, the assets and transaction contributions are still shielded. Is the concern revealing alpha re: the TWAP, or something else?

hdevalence:
> Even in the case that the various transactions constituting the TWAP are linkable via the gas price, the assets and transaction contributions are still shielded. Is the concern revealing alpha re: the TWAP, or something else?

Yeah, exactly, the concern is that the individual transactions comprising the TWAP may be linkable to each other.

More generally, one of the goals of making transaction authorization bind only to the effecting data (which does not include the ZKPs or the proof anchor) was so that the final transaction would not reveal _when_ it was authorized. This prevents distinguishing transactions made with a custody setup that has more latency. For instance, if we had the signatures sign over the proofs, the proofs would have to be made first, and then an observer could tell that a transaction took a long time to sign by noticing that it was submitted with a much older anchor than when it was included in a block.

The fact that the fee is part of the effecting data (necessarily! it does matter that users authorize the fees they pay, and it affects spends etc), together with the fact that fees now depend on gas prices, means we somewhat reintroduce this property.  So it's somewhat desirable if the gas prices don't change super super rapidly.  On the other hand, having accurate representation of the costs of resource use is also important, so there will be a balancing act.

PR #3143, opened 2023-10-03
feat(pcli): support new ibc query logic

### LINKED ISSUE #2937: ibc: add ability to query all channels

conorsch:
Previously pcli was only capable of looking up info on IBC clients, connections, and channels if the user provided a unique identifier for the resource in question. To facilitate IBC in practice, we want to enable querying for lists of these values, so the end user (or developer) can inspect the current options in a legible format.

Closes #2937.

### DIFF
--- a/crates/bin/pcli/src/command/query/ibc_query.rs
+++ b/crates/bin/pcli/src/command/query/ibc_query.rs
@@ -1,7 +1,12 @@
 use anyhow::{Context, Result};
 use colored_json::ToColoredJson;
+use ibc_proto::ibc::core::channel::v1::query_client::QueryClient as ChannelQueryClient;
+use ibc_proto::ibc::core::channel::v1::QueryChannelsRequest;
+use ibc_proto::ibc::core::client::v1::query_client::QueryClient as ClientQueryClient;
+use ibc_proto::ibc::core::client::v1::{QueryClientStateRequest, QueryClientStatesRequest};
+use ibc_proto::ibc::core::connection::v1::query_client::QueryClient as ConnectionQueryClient;
+use ibc_proto::ibc::core::connection::v1::{QueryConnectionRequest, QueryConnectionsRequest};
 use ibc_types::core::channel::ChannelEnd;
-use ibc_types::core::connection::ConnectionEnd;
 use ibc_types::lightclients::tendermint::client_state::ClientState as TendermintClientState;
 
 use penumbra_proto::core::app::v1alpha1::{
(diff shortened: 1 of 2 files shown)


conorsch:
Requesting feedback from @avahowell. It might be useful to compare results between the current state of this PR and after applying this diff:

```
diff --git a/crates/bin/pcli/src/command/query/ibc_query.rs b/crates/bin/pcli/src/command/query/ibc_query.rs
index e821b93bf..6de19a816 100644
--- a/crates/bin/pcli/src/command/query/ibc_query.rs
+++ b/crates/bin/pcli/src/command/query/ibc_query.rs
@@ -83,12 +83,7 @@ impl IbcCmd {
                     .await?
                     .into_inner()
                     .client_states;
-                let clients_ids: Vec<String> =
-                    client_states.into_iter().map(|c| c.client_id).collect();
-                // TODO: Resolve formatting for these representations. The key_value lookup
-                // for IbcCmd::Client recasts as a TendermintClientState; the ibc ClientState
-                // is less informative. For now, just emitting the names.
-                let clients_json = serde_json::to_string_pretty(&clients_ids)?;
+                let clients_json = serde_json::to_string_pretty(&client_states)?;
                 println!("{}", clients_json.to_colored_json_auto()?);
             }
             IbcCmd::Connection { connection_id } => {

```

conorsch:
Wasn't able to drag the `channel` (singular) logic over to the new format, so I left that as the old k/v lookup.

Issue #2997, opened 2023-09-11
Add Bech32 encoding for `AccountGroupId`


hdevalence:
**Is your feature request related to a problem? Please describe.**

We'd like to domain-separate the IndexedDB instance used by the web extension's view server so that it's scoped to the network and to the viewing key. Currently this is done with an address, but the correct domain separator is an `AccountGroupId`. However, we can't use an `AccountGroupId` at the moment, because we don't have a Bech32 encoding defined for it.

**Describe the solution you'd like**

Add Bech32 encoding methods for an `AccountGroupId`, with a `penumbraaccountgroupid` HRP.  This allows Rust code to go from an FVK to an `AccountGroupId` (using existing methods) and then to Bech32-encode the resulting ID.

**Additional context**

Maybe this needs to be made accessible from the wasm crate as well?

PR #3145, opened 2023-10-04
account_group_id -> wallet

grod220:
Closes #https://github.com/penumbra-zone/penumbra/issues/2997


hdevalence:
One thing I don't think we settled on: up to this point, the "account group ID" has been a purely internal name, only used in internal code and datastructures. Adding a Bech32 prefix for it feels like making it in some sense user-facing (at least to a subset of users).

Did we want to align on terminology first? We'd talked about having a conceptual hierarchy of "wallet" / "account" / "address" for spend capability / logical account / payment capability that we currently call "account group" / "account" / "address". Should we commit to that?

grod220:
@hdevalence I think that terminology makes sense! There also seemed to be consensus on the team meeting with it too. Changed all instances of account group id to wallet.

hdevalence:
I hate to ask to redo everything, but I'm wondering if we could call it `WalletId` rather than `Wallet`. It seems kind of important to have the name clarify that it's just an identifier, it doesn't actually do anything itself -- if we call it "wallet", I think someone reading the code or API docs could easily get confused thinking that it was the thing that manages funds, rather than _just_ being an identifier.

grod220:
I wasn't sure if this struct would be something later down the line we'd extend. I had a hunch about Id though. Wasn't terribly hard to add.

grod220:
@hdevalence thanks for the re-review 🙏

PR #3146, opened 2023-10-04
#3091 separate block scanning and receiving updates for wasm

Valentine1898:
Changes wasm needed to save all updates to indexed-db transactionally 
Related: [#34](https://github.com/penumbra-zone/web/issues/34)

These changes allow to use different block saving modes:
1. Saving the updates after processing each block
Example:
```javascript
await wasmViewServer.scanBlock();
const scanResult = wasmViewServer.get_updates(last_position, last_forgotten)
// process scan result
``` 
2. Saving updates every N blocks, or when we encounter new notes, swaps
Example:
```javascript
const found_new_data = await wasmViewServer.scanBlock();
if ( block.height % interval === 0n || found_new_data) {
    const scanResult = wasmViewServer.get_updates(last_position, last_forgotten)
    // process scan result
}
``` 
That way you don't have to check nullifier twice (on the wasm and ts side). And this is important, because double processing of nullifier could greatly affect the synchronization speed  (Related: [sync.rs#L23](https://github.com/penumbra-zone/penumbra/blob/main/crates/view/src/sync.rs#L231) )


crates/wasm/src/view_server.rs
```diff
@@ -101,20 +101,19 @@ impl ViewServer {
...
-    /// Arguments:
+    /// Returns true if the block contains new notes, swaps or false if the block is empty for us
     ///     compact_block: `v1alpha1::CompactBlock`
+    /// Scan results are saved in-memory rather than returned
+    /// Use `get_updates()` to get the scan results
     /// Returns: `ScanBlockResult`
     #[wasm_bindgen]
-    pub async fn scan_block(&mut self, compact_block: JsValue) -> Result<JsValue, Error> {
+    pub async fn scan_block(&mut self, compact_block: JsValue) -> Result<bool, Error> {
         let result = self.scan_block_inner(compact_block).await?;
```
grod220, line 111: Didn't we get rid of this need for `_inner` functions? Would be ideal if we could fix that here too.

crates/wasm/src/view_server.rs
```diff
@@ -101,20 +101,19 @@ impl ViewServer {
...
 
     /// Scans block for notes, swaps
-    /// This method does not return SCT updates (nct_updates).
-    /// Although we can make it do if we want to save SCT updates after each blob is processed
-    /// Arguments:
+    /// Returns true if the block contains new notes, swaps or false if the block is empty for us
     ///     compact_block: `v1alpha1::CompactBlock`
+    /// Scan results are saved in-memory rather than returned
+    /// Use `get_updates()` to get the scan results
     /// Returns: `ScanBlockResult`
```
grod220, line 108: `Returns: 'ScanBlockResult'` needs updating

crates/wasm/src/view_server.rs
```diff
@@ -338,13 +331,7 @@ impl ViewServer {
 
         self.latest_height = block.height;
 
-        let result = ScanBlockResult {
-            height: self.latest_height,
-            nct_updates: Default::default(),
-            new_notes,
-            new_swaps,
```
grod220, line 345: When I later to `get_updates()` does it return every note stored on this struct despite the range I ask for? Does that mean when I do `get_updates()` many times, I keep getting that same list, but bigger? We may need to think through this more.

Valentine1898: Hmm, we can just clear notes and swaps during the `get_updates()` call

crates/wasm/src/view_server.rs
Valentine1898: Some large _inner functions remain, I'll fix them

crates/wasm/src/view_server.rs
grod220: Would clearing the notes affect this line?

```rust
          StatePayload::RolledUp(commitment) => {
                    if let std::collections::btree_map::Entry::Occupied(mut e) =
                        self.notes.entry(commitment)
                    {
```

I'm not sure what `RolledUp` is

grod220: Also, if we move to this method (get_updates() clears state), we probably can start storing:
- last_position
- last_forgotten

On the struct itself. And when an get_updates() is called, we update those struct fields with the updates:
- set_position
- set_forgotten

grod220: And likely we should change `get_updates` do something like `flush_updates`.

Valentine1898:
```javascript
      StatePayload::RolledUp(commitment) => {
                    if let std::collections::btree_map::Entry::Occupied(mut e) =
                        self.notes.entry(commitment)
                    {
``` 
Now it seems to me that it's actually a mistake.
The `RolledUp` handler is associated with `swap claim`. When a swap occurs we save the advice so that in the future, when a `swap claim` occurs, we can use it to generate a `SpendableNote`
I'll test it

Valentine1898:
```rust
 StatePayload::RolledUp(commitment) => {
                    if let std::collections::btree_map::Entry::Occupied(mut e) =
                        self.notes.entry(commitment)
                    {
``` 
This condition was indeed wrong. It is almost impossible to reproduce the bug related to it, because swap claim is automatically executed immediately after swap.

Valentine1898: I removed it, and tested correctly scanning the block with the swap claim.

Valentine1898: And after that, completed the redesign of `get_updates() `into `flush_updates()`

crates/wasm/src/view_server.rs
```diff
@@ -91,112 +92,30 @@ impl ViewServer {
...
         let block_proto: penumbra_proto::core::component::compact_block::v1alpha1::CompactBlock =
             serde_wasm_bindgen::from_value(compact_block)?;
 
         let block: CompactBlock = block_proto.try_into()?;
 
-        // Newly detected spendable notes.
-        let mut new_notes = Vec::new();
-        // Newly detected claimable swaps.
-        let mut new_swaps: Vec<SwapRecord> = Vec::new();
+        let mut found_new_data: bool = false;
```
grod220, line 118: Do you see a usecase for this in the web code? I don't think this will be used.

crates/wasm/src/view_server.rs
```diff
@@ -91,112 +92,30 @@ impl ViewServer {
...
-
-    /// get SCT state updates
-    /// This method is necessary because we save SCT updates to indexedDB once every 1000 blocks
-    /// rather than after each block
-    /// Arguments:
-    ///     last_position: `Option<StoredPosition>`
-    ///     last_forgotten: `Option<Forgotten>`
-    /// Returns: `ScanBlockResult`
+    /// Scan results are saved in-memory rather than returned
+    /// Use `get_updates()` to get the scan results
```
grod220, line 109: `flush_updates()`

grod220 (commented):
Have you manually tested this with the new web repo? Without tests in the wasm crate, it's hard to tell if this breaks anything.

crates/wasm/src/view_server.rs
Valentine1898: Yes, of course.  
I changed the web repo code to test the new wasm logic. 
Most of the tests is to compare the state of `indexed_db` and` pcli view db` for different cases. 

Regarding `found_new_data`, this is an important part of block scanning. If we don't handle `found_new_data` values, we run the risk of missing nullifier and not marking some notes as spent.

I did it this way 
```typescript
 const found_new_data: boolean = await this.viewServer.scanBlock(res.compactBlock);

      await this.markNotesSpent(res.compactBlock.nullifiers, res.compactBlock.height);

      if (isDevEnv()) {
        await this.assertRootValid(res.compactBlock.height);
      }

      if (found_new_data || shouldStoreProgress(res.compactBlock, lastBlockHeight)) {
        await this.saveSyncProgress(res.compactBlock.height);
      }
``` 

And moved the saving of new notes to `saveSyncProgress()`
```typescript
 async saveSyncProgress(height: bigint) {
    const updates = await this.viewServer.updatesSinceCheckpoint();
    await this.storeNewNotes(updates.new_notes);
    await this.indexedDb.updateStateCommitmentTree(updates.nct_updates, height);
  }
``` 
Save `notes` and `nct_updates` also need to be saved in the same transaction, but for my tests this wasn't critical, so I left it out.

grod220: That means our db writes have that one exception (nullifiers) and everything else gets saved in a single transaction every 1000 blocks.

Could we just collect nullifiers as we find them and put them in memory on the struct? And then when we call `flush_updates()` we get those back/clears as well? That will allow us to handle all db writes only then.

grod220: Also, do those `nullifiers` belong to anyone else? Is it guaranteed to belong to the user? If it is specifically for the user, I see that it makes sense to have `found_new_data`. If it contains public info, then I'd say it doesn't make sense for `found_new_data` to be triggered by nullifiers being present.

Valentine1898: There's a little miscommunication 
`nullifiers` are not processed in wasm and do not affect the `found_new_data` value in any way
`found_new_data = true` only when we found our notes or swaps when scanning the block 
`Nullifiers` should be checked on TS side, comparing `nullifiers` that are in block and `nullifiers` that we have in indexed_db. 
In 99.99% of cases this requires only reading from `indexed_db`, because we will not find nullifiers and we will not need to change data in `indexed_db` (mark `note` as spent).
Marking notes as spent doesn't affect SCT  state, so I don't see anything wrong with having it in a separate transaction

grod220: Ah yes, that's right. I misunderstood. Ok, I think all looks good then.

grod220 (approved):

Issue #3116, opened 2023-09-28
Exercise ICS20 timeouts end-to-end against a Cosmos-SDK chain.


hdevalence:
**Is your feature request related to a problem? Please describe.**

In advance of getting automatic testing as part of ongoing InterchainTest integration, we should manually exercise the end-to-end flow of timeouts against a Cosmos-SDK chain.

This will require a counterparty chain with ICS23 non-inclusion proof support (#3045) and using our Hermes fork with the config option added in https://github.com/penumbra-zone/hermes/pull/3  to create a client with the correct proof spec.

Currently, this is blocked by:

- [x] #3082

conorsch:
To test this, we can make a packet with near-immediate timeouts. This test should be manual, so we can observe behavior and document results.

conorsch:
Achieved! @avahowell managed to get a local testing setup together to validate the behavior. She's going to write up those testing steps in this ticket, then close out.

avahowell:
I was able to successfully perform a timeout of a packet from a local osmosis testnet (built using the osmosis `make localnet-*` tooling) to a penumbra testnet.

The steps were as follows,

1. perform an ibc handshake using hermes 
2. perform a transfer using `rly`, with a timeout height of +5 blocks
3. wait 5 blocks, then run `hermes start`
4. observe that hermes constructs a timeout, using the nonexistence proof from the penumbra chain, commits it to the osmosis chain,
5. try another transfer and verify that it still succeeds after timeout

A small change to our query api was required. When that's merged, I'll close this issue

https://github.com/penumbra-zone/penumbra/pull/3163

conorsch:
Closing as completed.

conorsch:
Let's try this again, but this time using public testnets, not local nets. This is recently made possible by the dep changes in Osmosis testnet.

aubrika:
This has been demonstrated, per @avahowell

PR #3163, opened 2023-10-06
abci query: support nonexistence proofs

avahowell:
Previously, we would error in the ABCI query interface if a key was not found. This PR changes that behavior to instead return an empty vec for the requested value, along with a nonexistence proof to the root hash showing the value doesn't exist in the state.

This is required for relayers to build timeout packets. I verified that this change works locally using hermes.


avahowell:
Closes #3116

PR #3170, opened 2023-10-10
app: register metrics of components

hdevalence:


conorsch:
Good catch! Will inspecting the metrics on the current testnet 61, I also noticed a small regression in that we're no longer capturing some concurrent connection stats:

metrics-regression-in-pd

I don't expect these changes will resolve that, but will proceed with release as planned today. I'll follow up separately to track down the root cause on the regression in the screenshot above.

hdevalence:
I think those should have moved to the `penumbra_compact_block` crate, but I don't see a `register_metrics` method there. Maybe I forgot to add that while splitting up RPCs.

PR #3173, opened 2023-10-10
docs: add ibc channel info for testnet 62 iapetus

conorsch:
WIP: started this PR after release of 62, but wasn't able to figure out the right steps. In particular, the `--timeout-height` default arg will not work.

### DIFF
--- a/crates/bin/pcli/src/command/tx.rs
+++ b/crates/bin/pcli/src/command/tx.rs
@@ -173,8 +173,11 @@ pub enum TxCmd {
         #[clap(long)]
         channel: u64,
 
-        /// Block height on the current chain, after which the withdrawal will be considered
-        /// invalid if not already relayed.
+        /// Block height on the counterparty chain, after which the withdrawal will be considered
+        /// invalid if not already relayed. Must be specified as a tuple of revision number and block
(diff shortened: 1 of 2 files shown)


conorsch:
The error I'm seeing on the timeout-height is: `error: Invalid value "0-0" for '--timeout-height <TIMEOUT_HEIGHT>': attempted to parse an invalid zero height`.

If I override the default value with a custom one, such as "5-100000":


```
Error: status: Internal, message: "could not broadcast transaction: Error submitting transaction: code 1, log: timeout height 5-100000 is less than the latest height 5-3153474", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc"} }
```

What structure should this arg take? What is a sane default value?

conorsch:
@hdevalence clarified in chat that the timeout-height applies to the _counterparty_ chain, and additional is a tuple of two values: chain revision and block height. I can't think of a sane default value for this, so I've updated the docs to be explicit about declaring it. Relevant docs:  https://rustdoc.penumbra.zone/main/penumbra_ibc/struct.Ics20Withdrawal.html

avahowell:
LGTM

Issue #3158, opened 2023-10-06
Summonerd: Implement transitions between phase1 and phase2


cronokirby:
This needs to:
- Read the phase1 table, to get the latest contribution
- Use the circuit information to transition, producing a phase2 root, along with some auxiliary information for reconstruction.

PR #3183, opened 2023-10-11
summoner: provide phase 1 root during ceremony initialization

### LINKED ISSUE #3156: Summonerd: Initializing phase 1 root
We want to be able to at least setup the start of phase1 by using a "blank" set of parameters, which are still valid, just deterministically generated.

This command can later be modified to read some kind of file with a desired starting output.

redshiftzero:
Closes #3156

Note that I have intentionally not done the transition from the provided phase 1 to phase 2, which will be done in #3158 which I'll pick up after this one.

### DIFF
--- a/tools/summonerd/src/main.rs
+++ b/tools/summonerd/src/main.rs
@@ -12,6 +12,11 @@ use coordinator::Coordinator;
 use metrics_tracing_context::MetricsLayer;
 use penumbra_keys::FullViewingKey;
 use penumbra_proto::tools::summoning::v1alpha1::ceremony_coordinator_service_server::CeremonyCoordinatorServiceServer;
+use penumbra_proto::tools::summoning::v1alpha1::CeremonyCrs;
+use penumbra_proto::Message;
+use std::fs::File;
+use std::io::BufReader;
+use std::io::Read;
 use std::net::SocketAddr;
 use storage::Storage;
 use tonic::transport::Server;
@@ -19,7 +24,7 @@ use tracing_subscriber::{prelude::*, EnvFilter};
 use url::Url;
 
 use crate::{penumbra_knower::PenumbraKnower, server::CoordinatorService};
-use penumbra_proof_setup::all::Phase1CeremonyCRS;
+use penumbra_proof_setup::all::{Phase1CeremonyCRS, Phase1RawCeremonyCRS};
 
 /// 100 MIB
 const MAX_MESSAGE_SIZE: usize = 100 * 1024 * 1024;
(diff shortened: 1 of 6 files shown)


cronokirby (changes requested):
My thinking was that instead initialization should be a separate command, rather than something done automatically.

The reason behind that is that then we could make this command take in a non-dummy root from an external source to populate things instead.

redshiftzero:
The commands we have now are:

```
Running `target/debug/summonerd --help`
summonerd v0.62.0-11-g6d156c5
Penumbra summoning ceremony coordinator

USAGE:
    summonerd [OPTIONS] <SUBCOMMAND>

OPTIONS:
    -h, --help                             Print help information
        --tokio-console <tokio-console>    Enable Tokio Console support [default: false]
    -V, --version                          Print version information

SUBCOMMANDS:
    generate-phase1    Generate a phase 1 root (for testing purposes)
    help               Print this message or the help of the given subcommand(s)
    init               Initialize the coordinator
    start              Start the coordinator
```

I tested this via:

1. Generate a test phase 1 root and store in a file (if we use an externally provided phase 1 root later we can delete this): 
```
cargo run --bin summonerd -- generate-phase1 --output testphase1.txt
```

2. Initialize the database using the provided phase 1 root:
```
cargo run --bin summonerd -- init --storage-dir /tmp/summonerd --phase1-root testphase1.txt
```

After this step the database exists and we can validate that the phase 1 root has been inserted in the database:

```
% sqlite3 /tmp/summonerd/ceremony.db
SQLite version 3.39.5 2022-10-14 20:58:05
Enter ".help" for usage hints.
sqlite> sqlite> select * from phase1_contributions;
0|1|������n���N9��H�\�7�Ȃ��A��,��T���%b��g
t��H�|
```

3. Now we start the summonerd server as before:

```
cargo run --bin summonerd -- start --storage-dir /tmp/summonerd --fvk YOURFVKHERE --node https://grpc.testnet-preview.penumbra.zone
```

tools/summonerd/src/main.rs
```diff
@@ -87,6 +105,40 @@ impl Opt {
...
+                let mut phase_1_bytes = Vec::new();
+                let mut buffer = [0; 4096 * 10]; // 40 KB chunks
+
+                loop {
+                    let bytes_read = reader.read(&mut buffer)?;
+                    if bytes_read == 0 {
+                        break;
+                    }
+                    phase_1_bytes.extend_from_slice(&buffer[..bytes_read]);
+                }
```
cronokirby, line 124:
```suggestion
                reader.read_to_end(&mut phase_1_bytes)?;
```

I believe this should work

cronokirby: (We're already doing buffered reads via BufReader)

cronokirby (approved):
LGTM, I'll just apply my suggestion, check CI, and then merge

PR #3185, opened 2023-10-12
Make Summonerd a bit more responsive & Fix ranking bug

cronokirby:
No issue addressed here, but this is a quick fix to the coordinator loop to be able to send updates to clients whenever the queue changes.

I also took the opportunity to fix a bug I noticed, which was that by sorting addresses by the amount bid, we were then choosing the lowest bid, and not the highest.

The idea with the new queue is to create a merged event stream of new participants, and the completion of contributions, and then react to which of those is happening first in the loop accordingly.


redshiftzero (approved):
nice! LGTM

PR #3186, opened 2023-10-12
View service: wallet id rpc

grod220:
We need a way to retrieve the wallet id from the view service. Adds new:
- proto definition
- pcli command


crates/view/src/service.rs
```diff
@@ -1549,4 +1551,19 @@ impl ViewProtocolService for ViewService {
...
+
+    async fn wallet_id(
+        &self,
+        _: Request<WalletIdRequest>,
+    ) -> Result<Response<WalletIdResponse>, Status> {
+        let fvk = self
+            .storage
+            .full_viewing_key()
+            .await
+            .map_err(|_| Status::failed_precondition("Error retrieving full viewing key"))?;
```
erwanor, line 1563: nit: I would either propagate the storage module error, or mention that the FVK was not found to hint at an actionable item for the user.

grod220: We can't propagate the error unfortunately:
```
the trait `From<anyhow::Error>` is not implemented for `tonic::Status`
```
I suppose why it's being mapped like this a bunch of times in this file. I'm personally a big fan of `thiserror` to prevent all the manually mapping. Could be a good candidate to fix later.

>  or mention that the FVK was not found to hint at an actionable item for the user.

Is that not what the failed pre-condition text is doing? Or am I missing something?

erwanor: You can use `Error::to_string` on the anyhow struct, at your discretion but the gRPC status code might not be surfaced to the user depending on how the API is consumed. On external APIs, I like error messages that tell me what to do

PR #3192, opened 2023-10-16
penumbra: integration work for new tendermint bindings and `prost@0.12`

erwanor:
This PR moves the penumbra codebase to using `tendermint-rs@0.34`, `ibc-proto-rs@0.37`, `ics23@0.11`, `jmt@0.9`, `prost@0.12` and `tonic@0.10`


erwanor:
We need to adjust our `MsgUpdateClient` handling to use the `ClientState` abstraction, merging handling of header update, batch of headers, and misbehavior, see: https://github.com/cosmos/ibc-go/blob/830982adfb4448573f8898bbd07a73d06bff640a/docs/docs/03-light-clients/01-developer-guide/04-updates-and-misbehaviour.md#handling-clientmessages-updates-and-misbehaviour

erwanor:
I took the path of least resistance to handle `client_message` continuing to not support misbehavior submission via `MsgUpdateClient`. I think that's something we will want to do later, however. I will create a ticket to track this.

conorsch:
@erwanor There's a failing CI check that alleges some files need to be committed: https://github.com/penumbra-zone/penumbra/actions/runs/6538666650/job/17756717056?pr=3192

```
Finished! Consider running a 'cargo check --release' to confirm builds still work.
Updated 1 path from the index
Updated 1 path from the index
ERROR: protobuf files must be regenerated and committed.
Run this command locally: ./deployments/scripts/protobuf-codegen
These are the files that reported differences:
 M crates/proto/src/gen/penumbra.view.v1alpha1.rs
 M crates/proto/src/gen/penumbra.view.v1alpha1.serde.rs
Error: Process completed with exit code 1.
```

Try as I might, I cannot reproduce that error locally, not even if with shifts in tooling. I suspect it's a caching issue in the CI environment, but haven't been able to resolve. Given the importance of this PR, I suggest we press on and merge it. If the same codegen flapping occurs on other PRs after rebasing, then I'll jump on it and get a fix.

erwanor:
Thanks for your help debugging this, merging optimistically

conorsch:
> If the same codegen flapping occurs on other PRs after rebasing, then I'll jump on it and get a fix.

I am indeed seeing the same failure on other rebased PRs. Will look into it tomorrow.

Issue #3179, opened 2023-10-11
[wasm] Wasm code related to transaction building is deprecated and requires refactoring 


Valentine1898:
[ViewProtocolService](url) on the extension side must support such requests related to transaction building as follows 
1. rpc
[Witness](https://buf.build/penumbra-zone/penumbra/docs/main:penumbra.view.v1alpha1#penumbra.view.v1alpha1.ViewProtocolService.Witness)
2. rpc
[WitnessAndBuild](https://buf.build/penumbra-zone/penumbra/docs/main:penumbra.view.v1alpha1#penumbra.view.v1alpha1.ViewProtocolService.WitnessAndBuild)
3. rpc
[AuthorizeAndBuild](https://buf.build/penumbra-zone/penumbra/docs/main:penumbra.view.v1alpha1#penumbra.view.v1alpha1.ViewProtocolService.AuthorizeAndBuild)
4. rpc
[Authorize](https://buf.build/penumbra-zone/penumbra/docs/main:penumbra.custody.v1alpha1#penumbra.custody.v1alpha1.CustodyProtocolService.Authorize)

At the same time as the current wasm code allows you to do Authorize, Witness, Build in [one function ](https://github.com/penumbra-zone/penumbra/blob/027faa0640a348d1c394244501062b071fa2a156/crates/wasm/src/tx.rs#L63)

It is necessary to make separate wasm functions for Authorize, Witness, Build that can be called from TS for proper implementation of the rpc requests given above

grod220:
can we close this issue?

Valentine1898:
Yeah, I'm closing

PR #3197, opened 2023-10-16
[wasm] Wasm code related to transaction building is deprecated and requires refactoring #3179

Valentine1898:
#3179


grod220:
The dapp code needs to do something like:

```typescript
    sendTx: async () => {
      const { plan } = await viewClient.transactionPlanner(reqFromForm());
      const { data } = await custodyClient.authorize({ plan });
      const { transaction } = await viewClient.authorizeAndBuild({
        transactionPlan: plan,
        authorizationData: data,
      });
      const { id } = await viewClient.broadcastTransaction({ transaction, awaitDetection: true });
      return uint8ArrayToHex(id.hash);
    },
```

And the extension needs to be able to handle each of those messages and use the wasm functions to help fulfill these requests. I'm currently working on this now, but can you ensure that these changes are compatible with the grpc messages?

Valentine1898:
Yes, I can confirm that wasm functions will allow to do this, processing of `authorizeAndBuild` request will be like this 
```typescript
    const plan: TransactionPlan;
    const authData: AuthorizationData;
            
    const witnessData: WitnessData = witness(plan.toJson(), loadStoredTree());
    const transaction: Transaction = build(loadFullViewingKey(), plan.toJson(), witnessData.toJson(), authData.toJson());
```

Valentine1898:
It seems that some checks have failed, but that's not related to my changes. This PR is ready for review

grod220 (approved):

Issue #3194, opened 2023-10-16
ibc: wasm planner should support gas fees


conorsch:
During a recent review of web code (https://github.com/penumbra-zone/web/issues/97) we noticed that the wasm planner interface lacks knowledge of the gas pricing estimates, which already exist in the view service crate. Let's update the wasm code so it can leverage that knowledge, which will reduce code duplication, and reduce surprises as the fee implementation stabilizes.

Issue #3178, opened 2023-10-11
[wasm] Add ICS20 withdrawals support for transaction plan building 


Valentine1898:
[Transaction planner](https://buf.build/penumbra-zone/penumbra/docs/main:penumbra.view.v1alpha1#penumbra.view.v1alpha1.ViewProtocolService.TransactionPlanner) interface now supports 
```typescript
repeated penumbra.core.component.ibc.v1alpha1.IbcAction ibc_actions = 60;
repeated penumbra.core.component.ibc.v1alpha1.Ics20Withdrawal ics20_withdrawals = 61;
``` 
We need to add support for these actions (`ibc_actions`, `ics20_withdrawals`) to [wasm_planner](https://github.com/penumbra-zone/penumbra/blob/main/crates/wasm/src/wasm_planner.rs
) 
There may be some complications when importing crates related to IBC, previously we avoided them because they were not compatible with `wasm32-unknown-unknown `target

Valentine1898:
I don't have permission, so I can't assign this and some other wasm issues to me

PR #3198, opened 2023-10-17
wasm_planner support gas fees & add ICS20 withdrawal

Valentine1898:
#3194 
#3178


crates/wasm/src/wasm_planner.rs
```diff
@@ -172,6 +191,12 @@ impl WasmPlanner {
     pub async fn plan(&mut self, refund_address: JsValue) -> WasmResult<JsValue> {
         utils::set_panic_hook();
 
+        // Calculate the gas that needs to be paid for the transaction based on the configured gas prices.
+        // Note that _paying the fee might incur an additional `Spend` action_, thus increasing the fee,
+        // so we slightly overpay here and then capture the excess as change later during `plan_with_spendable_and_votable_notes`.
```
grod220, line 196: Is overpaying what is being done in plci as well?

crates/wasm/src/planner.rs
```diff
@@ -135,6 +147,22 @@ impl<R: RngCore + CryptoRng> Planner<R> {
         self
     }
 
+    /// Calculate gas cost-based fees and add to the transaction plan.
+    ///
+    /// This function should be called once.
+    pub fn add_gas_fees(&mut self) -> &mut Self {
```
grod220, line 153: Instead of having this be a separate call that needs to be done at the end, why not put it inside `set_gas_prices` and it gets done as someone adds the gas price?

crates/wasm/src/wasm_planner.rs
Valentine1898: Yeah, it's just copied from pcli

crates/wasm/src/planner.rs
Valentine1898: `add_gas_fees` depends on the actions that have been added to the plan. We must ensure that new actions cannot be added after `add_gas_fees() `is executed, so it is better to do it once at the end. And yes, this solution is also inherited from pcli

PR #3199, opened 2023-10-17
penumbra: regen proto with `protoc@24.4`

erwanor:
For some reason, I was able to work around the caching issue locally by upgrading to `protoc@24.4` and `buf@1.27`


conorsch:
Nice. I'll to the same locally to confirm behavior (I tried 24.4 but not a buf upgrade), then update the docs to match, so we can link the rest of the team.

PR #3207, opened 2023-10-18
Tx planner updates [wasm]

grod220:
- Moving chain+fmd params to be passed in versus grabbing global indexdb tables
- adding error console hook to all wasm-bindgen functions. Without this, we cannot debug from javascript.


Valentine1898 (approved):
Some changes have already been made here at #3198 
After merging this PR I will make changes to #3198 so we don't have any conflicts

PR #3208, opened 2023-10-18
Improve summonerd performance through parallelism

### LINKED ISSUE #3202: Summonerd: Use parallelism to improve contribution speed
The reason this would be nice is that it allows us to slot more contributions within a given window. At the very least, it helps the coordinator process contributions faster, given more cores to grind away with.

Implementation wise, we can likely make good progress just with enabling parallel features in arkworks, and using rayon for to parallelize operations across each circuit we're making a setup for.

cronokirby:
Closes #3202, although I'm not quite sure we're exploiting everything we can.

In particular, the way you enable parallelism in arkworks through spooky action at a distance always makes me unsure if I've done things correctly. Would appreciate a review of that specifically.

(Also, I refactored some of the code so this PR ends up being a net decrease in loc 8^))

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5312,6 +5312,7 @@ dependencies = [
  "penumbra-stake",
  "rand_chacha 0.3.1",
  "rand_core 0.6.4",
+ "rayon",
 ]
 
 [[package]]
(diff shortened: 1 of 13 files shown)


redshiftzero (approved):
LGTM!

PR #3225, opened 2023-10-25
Simplify state management in summonerd

### LINKED ISSUE #3221: Summonerd shared state refactor
One way to simplify state management which would help make a progress web view easier is to centralize the management of the queue into a single piece of locked state, rather than spreading it out via channels. This would allow the server, the coordinator, and the web view to all coordinate via accessing this state, which would greatly simplify the current code.

More concretely, we would have an `Arc<RWLock<ParticipantQueue>>` or something akin to that, which would then be accessed as needed by these different components.

cronokirby:
Closes #3221

### DIFF
--- a/tools/summonerd/src/coordinator.rs
+++ b/tools/summonerd/src/coordinator.rs
@@ -1,74 +1,56 @@
-use std::{cmp, collections::HashMap, time::Duration};
+use std::time::Duration;
 
-use anyhow::{anyhow, Result};
-use futures::FutureExt;
-use penumbra_keys::Address;
-use penumbra_num::Amount;
-use tokio::sync::mpsc::{self};
-use tokio_stream::{wrappers::ReceiverStream, StreamExt};
+use anyhow::Result;
 
-use crate::{participant::Participant, phase::Phase, storage::Storage};
+use crate::{participant::Participant, phase::Phase, queue::ParticipantQueue, storage::Storage};
 
-struct ContributionHandler {
+const QUEUE_SLEEP_TIME_SECS: u64 = 1;
+
+pub struct Coordinator {
     storage: Storage,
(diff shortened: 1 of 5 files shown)


cronokirby:
https://github.com/penumbra-zone/penumbra/actions/runs/6635035943 the full summoning smoke test passed successfully as well, but a careful review is still warranted imo

tools/summonerd/src/coordinator.rs
```diff
@@ -1,90 +1,69 @@
...
         .await
         {
             Ok(Ok(_)) => Ok(()),
             Err(_) => {
                 tracing::info!("STRIKE (timeout)");
-                self.storage.strike(&contributor).await?;
+                self.storage.strike(&address).await?;
                 Ok(())
             }
             Ok(Err(e)) => Err(e),
```
hdevalence, line 56: Why do we want to bubble up an error here if `contribute_inner` fails? If a participant does something bad (e.g., submitting something that fails validation), won't this still crash the server?

tools/summonerd/src/participant.rs
```diff
@@ -11,14 +11,14 @@ use penumbra_proto::{
...
-use tokio::sync::mpsc;
+use tokio::sync::{mpsc, Mutex};
 use tonic::{Status, Streaming};
 
 use crate::phase::Phase;
 
 pub struct Participant {
     address: Address,
-    rx: Streaming<pb::ParticipateRequest>,
+    rx: Mutex<Streaming<pb::ParticipateRequest>>,
```
hdevalence, line 21: I don't think we should be using a `Mutex` here, why are we sharing access to the participant's messages?

hdevalence:
The comments above are largely stylistic; I'm not sure that the new code structure is optimal, but I think it may be good enough for the purposes of our ceremony, so we could merge this and move forward.

Issue #2312, opened 2023-04-06
Refactor `pcli` transaction viewing to be an extension trait on `TransactionView`


hdevalence:
**Is your feature request related to a problem? Please describe.**

`pcli` has existing code to render transactions, which uses `comfy_table` to render data in a terminal.  This code was written before we even had transaction views, so it assumes that the full viewing key is present.  But the design we're aiming at is that the `TransactionView` is supposed to be a self-contained, pre-computed view of the transaction.

**Describe the solution you'd like**

To reflect this, we could refactor the transaction viewing code in `pcli` into a method on `TransactionView` in an extension trait:

```rust
// in pcli/src/transaction_view_ext.rs
pub trait TransactionViewExt {
    /// Render this transaction view on stdout.
    fn render_terminal(&self);
}

impl TransactionViewExt for TransactionView {
    fn render_terminal(&self) {
        // use comfy_table to render transaction contents
    }
}
```

This would be similar, in spirit, to the `TransactionView => HTML` pipeline we're planning for the web extension.  We could adapt the existing code, but it's not necessary to stick to it, we can do whatever format we want.

jackmcasey:
I recently came across a post on Bluesky from @aubrika , one of your team members, mentioning that you're hiring Rust engineers. I wanted to take a look at Penumbra and get familiar with what it is and how it works. This seemed like as good a jumping off point as any.

Will be pushing the code to a GitHub repo later today, but wanted to post this to get some feedback. It's more or less "playing with the fonts and margins", but I've added a visualization for encrypted ActionViews that is a bit more human friendly. Might be a bit loud though.

![image](https://github.com/penumbra-zone/penumbra/assets/23410235/ffeaa66c-97fe-466a-a1f9-1f9e8482a788)

Anyway, let me know what you think!

hdevalence:
That looks really cool! I think the colors are a bit loud, but maybe in black and white it would convey the "this data is encrypted" part really well...

cronokirby:
Another cool idea for the aesthetics would be to use a random string of braille characters, which kind of has this look to it:
`⠨⡳⠡⡳⡸⢠⠌⢊⠋⣢⣶⣑⣕⠃⢻⠜⢮⡬⢌⠎⣴⣓⣿⡮⣈⣽⠓⣹⡷⡜⡤⢘`

jackmcasey:
I agree. It also dodges the case where the foreground_color and background_color happen to be the same color.

[...]

Issue #2547, opened 2023-05-11
`SwapClaim` transactions are not viewable by the view server


hdevalence:
**Describe the bug**

`SwapClaim` transactions are not viewable by the view server:
```
~/c/p/penumbra (bsod-rename-unfilled|✔) $ cargo run --release --bin pcli -- -n http://localhost:8080 tx swap 1000test_usd --into penumbra
    Finished release [optimized] target(s) in 0.35s
     Running `target/release/pcli -n 'http://localhost:8080' tx swap 1000test_usd --into penumbra`
Scanning blocks from last sync height 48 to latest height 222
[0s] ██████████████████████████████████████████████████     174/174     4628/s ETA: 0s
building transaction...
finished proving in 0.980 seconds [2 actions, 1 proofs, 1169 bytes]
broadcasting transaction and awaiting confirmation...
transaction confirmed and detected: fcd213335856605d149ba685953bcff86dbac7c2d616a7f988b0b1d6c489d6e4
Swap submitted and batch confirmed!
You will receive outputs of 0test_usd and 100penumbra. Claiming now...
building transaction...
finished proving in 0.001 seconds [1 actions, 0 proofs, 3179 bytes]
broadcasting transaction and awaiting confirmation...
transaction confirmed and detected: e80cfeef60ba0d1e6be7aefcb3b121eb9435508f1c3b8575769620d0e2c5f419
~/c/p/penumbra (bsod-rename-unfilled|✔) $ cargo run --release --bin pcli -- -n http://localhost:8080 view tx e80cfeef60ba0d1e6be7aefcb3b121eb9435508f1c3b8575769620d0e2c5f419
    Finished release [optimized] target(s) in 0.34s
     Running `target/release/pcli -n 'http://localhost:8080' view tx e80cfeef60ba0d1e6be7aefcb3b121eb9435508f1c3b8575769620d0e2c5f419`
Scanning blocks from last sync height 223 to latest height 227
[0s] ██████████████████████████████████████████████████       4/4       236/s ETA: 0s
Transaction not found in view service, fetching from fullnode...
 Action Type  Description
 Swap Claim   Opaque swap claim for trading pair: passet14h46dmcyy6fl5vyz7xx933n8l7jy202ahrxg92kad6yfladsvcyqczcwda <=> passet1984fctenw8m2fpl8a9wzguzp7j34d7vravryuhft808nyt9fdggqxmanqm with fee 0penumbra

 Transaction Fee                0penumbra
[...]

PR #3240, opened 2023-10-30
Refactor pcli transaction viewing to be an extension trait on TransactionView

wtf-jik:
#2312 

-    Introduces the TransactionViewExt extension trait.
-    Implements the render_terminal method in the extension trait.
-    Adds helper functions for improved formatting.
-    Proposes relocation of certain helper functions to more appropriate modules.


crates/bin/pcli/src/transaction_view_ext.rs
```diff
@@ -0,0 +1,349 @@
...
+        ValueView::UnknownDenom { amount, asset_id } => {
+            format!("{} {}", amount, format_opaque_bytes(&asset_id.to_bytes()))
+        },
+    }
+}
+
+
+fn format_fee_view(fee: &Fee) -> String {
+    format!("{} {}", fee.amount(), fee.asset_id())
+}
```
jackmcasey, line 103: Does a FeeView exist?

crates/bin/pcli/src/transaction_view_ext.rs
```diff
@@ -0,0 +1,349 @@
...
+
+// when handling ValueViews inside of a Visible variant of an ActionView, handling both cases might be needlessly verbose
+// potentially this makes sense as a method on the ValueView enum
+// propose moving this to core/asset/src/value.rs
+fn value_view_amount(value_view: &ValueView) -> Amount {
+    match value_view {
+        ValueView::KnownDenom { amount, .. } | ValueView::UnknownDenom { amount, .. } => {
+            *amount
+        }
+        _ => panic!("Unexpected ValueView variant!"),
```
jackmcasey, line 114: Unsure if this should be explicit. Don't really expect new enum variants.

crates/bin/pcli/src/transaction_view_ext.rs
```diff
@@ -0,0 +1,349 @@
...
+                                    swap_plaintext.delta_1_i,
+                                    swap_plaintext.trading_pair.asset_2(),
+                                ),
+                                // The pathological case (both assets have output values).
+                                _ => (
+                                    swap_plaintext.trading_pair.asset_1(),
+                                    swap_plaintext.delta_1_i,
+                                    swap_plaintext.trading_pair.asset_1(),
+                                )
+                            };
```
jackmcasey, line 229: Similarly to issues displaying Fee - does TradingPairView exist?

crates/bin/pcli/src/transaction_view_ext.rs
```diff
@@ -0,0 +1,349 @@
...
+                    ["Ics20 Withdrawal", &action]
+                }
+                penumbra_transaction::ActionView::PositionOpen(position_open) => {
+                    let position = &position_open.position;
+                    /* TODO: leaving this around since we may want it to render prices
+                    let _unit_pair = DirectedUnitPair {
+                        start: unit_1.clone(),
+                        end: unit_2.clone(),
+                    };
+                    */
```
jackmcasey, line 302: Brought this along from the prior implementation.

crates/bin/pcli/src/transaction_view_ext.rs
```diff
@@ -0,0 +1,349 @@
...
+                    '6' => "⠋", // "\u{2586}",
+                    '7' => "⠛", // "\u{2587}",
+                    '8' => "⠓", // "\u{2588}",
+                    '9' => "⠊", // "\u{2589}",
+                    'A' => "⠊", // "\u{259A}",
+                    'B' => "⠋", // "\u{259B}",
+                    'C' => "⠌", // "\u{259C}",
+                    'D' => "⠍", // "\u{259D}",
+                    'E' => "⠎", // "\u{259E}",
+                    'F' => "⠏", // "\u{259F}",
```
jackmcasey, line 59: The incredibly difficult decision of blocks versus braille. IMHO the blocks fit better.

crates/bin/pcli/src/transaction_view_ext.rs
```diff
@@ -0,0 +1,349 @@
...
+pub trait TransactionViewExt {
+    /// Render this transaction view on stdout.
+    fn render_terminal(&self);
+}
+
+impl TransactionViewExt for TransactionView {
+    fn render_terminal(&self) {
+        println!("Tx Hash"); // Not available here?
+        println!("Tx Sig"); // Probably not needed
+        println!("Anchor"); // Probably not needed
```
jackmcasey, line 127: Metadata placeholders that may not actually be needed

crates/bin/pcli/src/transaction_view_ext.rs
```diff
@@ -0,0 +1,349 @@
...
+                                ),
+                            };
+                            
+                            action = format!(
+                                "Claimed {} with fee {:?}",
+                                claimed_value,
+                                format_fee_view(&swap_claim.body.fee),
+                            );
+                            ["Swap Claim", &action]
+                        }
```
jackmcasey, line 276: Was unable to test the SwapView::Visible match since the view service could never find transactions containing the swaps I made. When fetching from a fullnode, I don't expect to get a SwapView::Visible, but do expect the view service to find the transaction I just executed. Not sure why this is happening.

jackmcasey (commented):
Added some notes, some questions.

crates/bin/pcli/src/transaction_view_ext.rs
hdevalence: No — but it probably should, I actually noticed this just a day or two ago while scaffolding some code for the web extension: https://github.com/penumbra-zone/web/pull/159/files#diff-5686a31bfb26a32d7a5f0f112c3b6cb9f0cfc7c57218264d1f1f87ae184b5f28R28

(cc @grod220) 

However, since that would require making breaking proto changes, we should wait to align it with a testnet release and I think we are probably too close to the wire in this cycle. So for now this could just render the number and not worry too much about the formatting.

jackmcasey: Gotcha, will do. Am I correct in saying that all fees are denominated in penumbra?

hdevalence: Yes, but the fees are a bit of a WIP, so it's probably better to save code and just render the raw number, and then we could improve the rendering when we fix the data modeling.

crates/bin/pcli/src/transaction_view_ext.rs
hdevalence: Nope -- so doing something simple is good, and we can collect a list of missing `View`s to add. Nice catch!

crates/bin/pcli/src/transaction_view_ext.rs
hdevalence: I think the tx hash is available, but in a slightly baroque way: clone the `TransactionView`, then "demote" the `TransactionView` into a `Transaction`, then hash that. Not sure it's important to render here, though, since so far this is only used in a command that takes the hash as input.

crates/bin/pcli/src/transaction_view_ext.rs
hdevalence: It could be related to a long-running bug that we have not got around to fixing yet: https://github.com/penumbra-zone/penumbra/issues/2547

So probably not important to go to great lengths to exercise in this PR.

crates/bin/pcli/src/transaction_view_ext.rs
jackmcasey: cargo fmt --all decided this should be implicit

crates/bin/pcli/src/transaction_view_ext.rs
jackmcasey: Agree

jackmcasey:
Rebased onto v0.63.0 as main was not building for me.

hdevalence:
Thanks! Don't worry too much about the other tooling errors, we can fix them for you.

hdevalence:
Hey @jackmcasey  --  I'm really sorry this fell through the cracks, we got sidetracked with the ceremony launch in November. It looks really great, though, and I'd love to get it merged.

I went through and rebased all the changes onto current `main`, but I wasn't able to push them to your fork, I think because of an LFS-related issue (googling the error led to threads about LFS configuration). I pushed up the rebased branch as https://github.com/penumbra-zone/penumbra/pull/3546/ if you want to replace the contents of this branch with it (or we can merge it through there, whichever is easier).

jackmcasey:
Appreciate that! Will do

jackmcasey:
Sorry for the delay. Rebased onto main since the community pool got renamed. Should be good? I hope!

PR #3244, opened 2023-10-30
ci: enable debug logging for validators

conorsch:
On Testnet 62 Iapetus, we experienced a chain halt at block height
279655. We recently removed debug logging from the testnet nodes, which caused problems for figuring out what went wrong. Full debug output generates a lot of text, which raises hosting costs. Here, we selectively re-enable debug just on the validator nodes, where it's most useful.


conorsch:
Test deploy of this change failed against an ad-hoc devnet, need to finagle the yaml.

Issue #3128, opened 2023-10-01
Reverse dependency relationship between `penumbra-ibc` and `penumbra-shielded-pool`


hdevalence:
Following the design work in #3127, we should implement it, reversing the dependency relationship between the IBC crate and the shielded pool. The ICS20 app handler should move into the shielded pool crate, where it can do Penumbra-specific stuff, and we can have the `penumbra-ibc` crate be independent of Penumbra-specific details.

conorsch:
There's some design work to do on this yet, but it's a great starting point for cross-org collab. Assigning @avahowell, given how she's been driving much of the IBC implementation work to date.

Issue #3127, opened 2023-09-30
Design a mechanism to allow dynamic `AppHandler`s in the IBC component


hdevalence:
Currently, `penumbra-ibc` depends on `penumbra-shielded-pool`, because the `penumbra-ibc` crate contains the implementation of the ICS20 token transfer handler, which edits the shielded pool state. This isn't the design we want, though, because we want the `penumbra-ibc` crate to be a "foundational" crate like `penumbra-storage` that's also usable by other projects. #3125 brings this problem from the backburner to the front burner.

Packet handling in `penumbra-ibc` is performed by the `AppHandler` trait, which exposes a generic interface for custom behavior. But currently, this is hardcoded to use an implementation that does Penumbra-specific ICS20 handling. Instead, we need a way to allow a _dependency_ to implement the `AppHandler`, and then somehow cause that app handler to be used by the `penumbra-ibc` code.

This may also mean the `AppHandler` trait needs to be object-safe, and that may cause issues.

Two ideas on how we could do this:

1. We could pass some kind of configuration object with the `AppHandler` implementation in somewhere. This may be difficult, because the existing `Component` interface used by `penumbra-ibc` may not have a way to pass in extra parameters.
2. We could use global state, and have a registration mechanism.  Dependencies would register app handlers at startup, and those handlers would be used during execution.

The downside of (2) is that it wouldn't be possible to have two different applications both using `penumbra-ibc` to do different things inside of the same process. This is kind of aesthetically displeasing, but I'm not sure it's a real problem at the moment, since it seems unlikely someone would, e.g., import all of Penumbra and all of Astria into one program. Otherwise, it seems potentially simpler.

In a previous iteration of informal design discussions with @avahowell about this issue, I think we'd felt that (2) was probably the right option, but we should compare both approaches.

We'll also need to design/refine the external interface, now that the `AppHandler` will be public.

hdevalence:
@noot To provide context for this design work, it would be helpful to know about how Astria works, in particular, how Astria represents accounts and balances, and what Astria's ICS20 handler would need to do in order to record the results of token transfers.

hdevalence:
> 1. We could pass some kind of configuration object with the `AppHandler` implementation in somewhere. This may be difficult, because the existing `Component` interface used by `penumbra-ibc` may not have a way to pass in extra parameters.

[...]

PR #3245, opened 2023-10-30
IBC refactor to remove dependence on `penumbra-shielded-pool`

noot:
At a high level, this wraps the `IbcAction` type with an `IbcActionWithHandler`, which contains an additional generic parameter that implements `AppHandler`. Also updated the `MsgHandler` trait methods to accept additional `AppHandler` generic parameters, so then the IBC messages which implement `MsgHandler` can pass in these parameters when calling `check_stateless`/`try_execute`.

I went with this instead of using global state with an `AppHandler` with a registration mechanism as the `AppHandler` contains generics, and trait objects with generics don't work :/ 

Also moved `Ics20Withdrawal` and the `component::transfer::Ics20Transfer` type into the shielded pool crate.

closes https://github.com/penumbra-zone/penumbra/issues/3127 and also closes https://github.com/penumbra-zone/penumbra/issues/3128.


hdevalence:
Huh, this seems like a really neat approach. Nice work. I'm not sure I see a downside to doing it this way, the `.with_handler` pattern seems really clean. @avahowell does this seem good?

avahowell:
I like this approach. This PR looks good to me, thanks for taking this on!

noot:
awesome, thanks @hdevalence and @avahowell ! i'll finish up the refactor and then open this up for review :)

erwanor (approved):
This is really great work, thanks for driving this.

PR #3248, opened 2023-10-30
Pcli ceremony: add contribution hash / receipt

cronokirby:
This can make it unto the upcoming release, and allows us to include these hashes into the database later.


hdevalence (commented):
LGTM!

PR #3251, opened 2023-10-30
Proof-Setup: Use faster verification method for phase1

cronokirby:
There was some vestigial debugging aid code doing many pairings instead of a batched pairing checks; this was very slow. Removing this code, we get a verification time of 4s, rather than 2+ minutes.


redshiftzero (approved):
LGTM!

PR #3255, opened 2023-10-31
Summonerd: add transcript hashes and timestamps

### LINKED ISSUE #3230: Summonerd: Add hashes and timestamps to scheme
The hashes would allow a person to easily verify inclusion, by saving a hash they would get from pcli, and checking that it appears in the sqlite file. Using timestamps allows for some cool information about contribution times in the web view.

cronokirby:
Closes #3230.


redshiftzero (approved):
nice thanks! i'll update the webapp to use these

PR #3257, opened 2023-10-31
updates for summonerd webapp

### LINKED ISSUE #3220: ceremony progress web view
Now that we have the pieces for the ceremony, it would be nice to have a web view that displays the progress of the ceremony, including the:
* Number of contributions thus far
* Current length of contribution queue (i.e. the number of people currently connected waiting to contribution)
* Last winning bid

See example code here: https://github.com/tokio-rs/axum/blob/main/examples/templates/src/main.rs
And relevant discussion in Discord starting [here](https://discord.com/channels/824484045370818580/948673930909077524/1163620197689720893)

redshiftzero:
Closes #3220

This splits the web app into three pages.

## Main page

This currently displays the current phase, where the current phase info is determined by looking in the database to see if the transition between phase 1 and 2 has occurred. 

It also contains instructions for this phase, where the `--phase` CLI arg is populated using the information determined as above. 

Remaining TODOs:
- [-] Populate the text on the page with the COORDINATOR_URL and SUMMONER_ADDRESS - no longer needed as we will hardcode these in pcli to simplify the user's experience

## `/phase/1` and `/phase/2` pages

These are identical pages, one per phase, that show the:

- The number of contributions in that phase
- The most recent 5 contributions: short address, hash, time of contribution completed

Remaining TODOs:
- [x] Show most recent bid amount
- [x] If phase is currently active, show the number of folks connected

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3500,9 +3500,9 @@ dependencies = [
 
 [[package]]
 name = "indexmap"
-version = "2.0.2"
+version = "2.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8adf3ddd720272c6ea8bf59463c04e0f93d0bbf7c5439b691bca2987e0270897"
+checksum = "d530e1a18b1cb4c484e6e34556a0d948706958449fca0cab753d649f2bce3d1f"
 dependencies = [
  "equivalent",
  "hashbrown 0.14.2",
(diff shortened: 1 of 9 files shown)


cronokirby:
Instead of populating the text on page, we should simply make pcli have default arguments for these two in the code.

cronokirby:
Also, I don't see why we wouldn't just display all the contributions; thousands shouldn't be an issue.

PR #3264, opened 2023-10-31
fix(summonerd): colorized cli output, via clap v3

conorsch:
Downgrades clap in summonerd v4 -> v3, to match other CLI tooling. Essentially reverts 7bb3e800824a8cdc6fe866d6dbe146420ed4ca94. With clap v3, there's extra whitespace between long-arg descriptions, due to our use of long_about in summonerd help docs. A major downside to clap v4 is colorized output doesn't appear to be supported for derive-style (it may be for builder-style, but we don't use that).

There's upstream progress on stabling color support, but until then, let's just stick with v3 and put up with the whitespace.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -187,52 +187,10 @@ dependencies = [
 ]
 
 [[package]]
-name = "anstream"
-version = "0.6.4"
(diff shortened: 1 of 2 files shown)


cronokirby:
LGTM!

PR #3268, opened 2023-11-01
fix ibc client query: use revision number and correct path

avahowell:


conorsch:
Nice! Refs #3143, where I first tried to implement these. Looks like these query changes don't affect consensus, so we could conceivably release as a point release. Thoughts?

PR #3272, opened 2023-11-02
Further summonerd parallelization

cronokirby:
Because some of the circuits are smaller, you want a slightly more intrusive level of parallelization for some operations,
in order to more effectively distribute work. This PR does that in a couple places.

## Server Side

Deserialization, which needs to validate elements, now parallelizes the validation in each circuit. (The infamous scalar multiplications).

## Client Side

Parallelize phase1 contribution creation further, now it consistently utilizes all cores.

The client side part is less important, since we don't control it, but it's still nice if beefier contributors can finish making their contribution faster.

### DIFF
--- a/crates/bin/pcli/src/command/ceremony.rs
+++ b/crates/bin/pcli/src/command/ceremony.rs
@@ -81,9 +81,12 @@ pub enum CeremonyCmd {
     Contribute {
         #[clap(long)]
         phase: u8,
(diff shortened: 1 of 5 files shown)


redshiftzero (approved):
LGTM, I agree the deserialization methods are equivalent to the corresponding Arkworks ones

PR #3277, opened 2023-11-02
Ceremony: use hardcoded URL but not wallet address

cronokirby:
Migrate away from a dev / test URL, and also avoid hardcoding a wallet address, since this might change more readily.

### DIFF
--- a/crates/bin/pcli/src/command/ceremony.rs
+++ b/crates/bin/pcli/src/command/ceremony.rs
@@ -81,12 +81,9 @@ pub enum CeremonyCmd {
     Contribute {
         #[clap(long)]
         phase: u8,
-        #[clap(long, default_value = "https://summoning-dev.penumbra.zone")]
+        #[clap(long, default_value = "https://summoning.penumbra.zone")]
         coordinator_url: Url,
-        #[clap(
-            long,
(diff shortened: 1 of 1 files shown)


crates/bin/pcli/src/command/ceremony.rs
```diff
@@ -81,12 +81,9 @@ pub enum CeremonyCmd {
     Contribute {
         #[clap(long)]
         phase: u8,
-        #[clap(long, default_value = "https://summoning-dev.penumbra.zone")]
+        #[clap(long, default_value = "https://summoning.penumbra.zone")]
         coordinator_url: Url,
-        #[clap(
-            long,
-            default_value = "penumbra1zjd6p7d6xta200mzcxtcudqthe4yfsnm6cw7wpr6tdecj6gsdpjmjld6vezu7rdj3kr8u9rl9pkhgst99sxy8wlvl8wqsst0wjhy0wewsczx9d6ys96z84rhewdwrm537y6wwz"
```
redshiftzero, line 88: I guess we should update the `main.html` template to add `--coordinator-address penumbra1zjd6p7d6xta200mzcxtcudqthe4yfsnm6cw7wpr6tdecj6gsdpjmjld6vezu7rdj3kr8u9rl9pkhgst99sxy8wlvl8wqsst0wjhy0wewsczx9d6ys96z84rhewdwrm537y6wwz` to the invocation of the `pcli ceremony` command?

cronokirby: Yeah, we can go ahead and do that; we still need another pass along with styling to embed further instructions, so that an external reading of the guide isn't necessary, but doing this now is good.

cronokirby: a galaxy brained thing would be for the coordinator to dynamically fetch its own address lol

Issue #3279, opened 2023-11-02
ci: deployment workflow improvements


conorsch:
We performed a point release today (`v0.63.1`) for the first time in a while. Here are some notes on refinements for keeping things smooth:

- [x] `-l part-of` filter doesn't grab fullnode deployments, only the validators. Update the Deployment spec in the penumbra-node chart to ensure the labels are present. 
- [x] `helmfile apply` not supported, but is preferred over `helmfile sync`; consider using an action to set it up, e.g. https://github.com/mamezou-tech/setup-helmfile - https://github.com/penumbra-zone/penumbra/pull/3280
- [ ] release binaries _for linux_ built again! but the windows build failed, which caused the macos build to fail, too. let's test in a private repo to figure out what the broken dep is. let's also pin MSRV across workflows, so we catch breakage early.
- [x] bot container builds: the PENUMBRA_VERSION var is not making it through to the dispatched job; results in errors like `Failed to pull image "ghcr.io/penumbra-zone/osiris:penumbra-":`. manually rebuilding via the bot repo is enough to resolve, but let's automate that.

conorsch:
> release binaries for linux built again! but the windows build failed, which caused the macos build to fail, too. let's test in a private repo to figure out what the broken dep is. 

Resolved by https://github.com/penumbra-zone/penumbra/pull/3319. Linux and macOS builds are working again, as of v0.63.2: https://github.com/penumbra-zone/penumbra/releases/tag/v0.63.2

PR #3280, opened 2023-11-03
ci: use apply rather than sync for deploys

conorsch:
This reverts commit 155df60540ada18f8c4315d41b1a3bfad11dd76d, which itself reverted commit 25c4e6fd531136f97bbdee71aef269f05c35ccf0.

ci: use helmfile gha helper
For including helm-diff, required for running "helmfile apply"

Refs #3279.


conorsch:
This worked on an ad-hoc deploy here https://github.com/penumbra-zone/penumbra/actions/runs/6739478414/job/18321101736

PR #3282, opened 2023-11-03
Add styling to summonerd website

cronokirby:
image
image

Tries to match the web wallet styles.

Also, the pipeline for changing the styling here is very annoying, I basically manually copied over the output of running an actual Parcel build using tailwindcss.

I tried to figure out TowerHttp's ServeDir service for a bit, but wasn't getting great results, and also realized this made deployment more annoying cause we needed to now manage this static directory of files. Was easier to just bite the bullet and include a few files in the binary.


turbocrime:
i have some things to contribute to this, i'll add today

PR #3284, opened 2023-11-03
penumbra-ibc: rename ibc relay actions

### LINKED ISSUE #3209: ibc: rename ibc_actions proto field for legibiility
There's currently a field on the TransactionPlannerRequest proto called `ibc_actions`:

Counterintuitively, this field does _not_ accept an Ics20Withdrawal, but rather applies specifically to actions related to relaying, _not_ transaction planning. Let's rename the field to something like `ibc_relay_actions` and the corresponding type `IbcRelay`. 

See related discussion in #3108.

erwanor:
Close #3209, this PR:

- renames the ibc component `IbcAction` enum to `IbcRelay`
- the proto enum `IbcAction` to `IbcRelay`
- the fields in various composite proto messages (e.g. `TransactionPlannerRequest`) to `ibc_relay_action`


avahowell:
LGTM

Issue #3274, opened 2023-11-02
Mocking indexed-db to run wasm tests


Valentine1898:
We create an `IndexedDBStorage` instance when we initialize the view server. This means that we are guaranteed to get an error when running tests in an environment without `indexed-db `
I am considering two options 
- switch to dependency injection. This means that we should change the code so that we can inject `IndexedDBStorage` via constructor or setter. This will allow us to use mock for `IndexedDBStorage` in the tests 
- move the creation of `IndexedDBStorage` from `ViewServer::new()` to the place in the code where `IndexedDBStorage` is called. This will allow us to test most of the view server code, but we will miss some test cases that should lead to `IndexedDBStorage` calls

Issue #3238, opened 2023-10-30
[wasm] Implement unit testing suite


TalDerei:
Implement unit testing using the `wasm-bindgen-tests` attribute for the independent components involved in the build process including `TransactionPlan`, `Authorization`, `Witness`, and `Build`. 

Integration testing for the [View Server](https://github.com/penumbra-zone/penumbra/blob/main/crates/wasm/src/view_server.rs) comes with it's own set of complexities since it interacts with external services like IndexedDB. It's not recommended to use `wasm-bindgen-tests` here per https://github.com/penumbra-zone/penumbra/issues/3180.

grod220:
@Valentine1898 fyi

TalDerei:
Referenced in https://github.com/penumbra-zone/penumbra/pull/3289

Issue #3236, opened 2023-10-29
Parallelize the building process using web-workers 


TalDerei:
### Objective
The objective is to refactor the [WASM](https://github.com/penumbra-zone/penumbra/tree/main/crates/wasm) crate by Zpoken (https://github.com/penumbra-zone/penumbra/pull/2973) to support parallelization of the transaction building process, enabling each action in a `TransactionPlan` to be handled by a separate web-worker, on the TS side. Currently, each action is performed sequentially by the main execution thread in the browser. 

This requires diving into the internals of the [build](https://github.com/penumbra-zone/penumbra/blob/main/crates/wasm/src/tx.rs#L158) method (https://github.com/penumbra-zone/penumbra/pull/3289) and  decoupling the individual actions from the TransactionPlan. Based on @hdevalence's suggestions, I propose a `ActionPlan::Build(BuildPlan)` method. 

Once a parallel strategy has been implemented on the Rust side, the necessary internals in the WASM crate will be decorated with additional `wasm-bindgen` attributes, and compiled into typescript wrappers using `wasm-pack`.  

### Motivation 
From my understanding, there are currently two independent implementations of a **view server**:
- Typescript / WASM implementation in the web extension
- Pure rust implementation using an async runtime like Tokio 

**PCLI** leverages the view server as a library and invokes Tokio to spawn parallel tasks. Exposing this same level of parallelism to typescript requires creating a rust API that exposes the internal details of these top-level functions, and then stitching those lower-level components back together in a parallel manner on the TS side. Since we ultimately want to split up the work into separate worker threads in TS, this involves a refactoring of the Rust code in a such a way that it doesn't have a tight coupled dependency on Tokio. 

### Testing
I've mocked the build process in TS independently from the [web-code](https://github.com/penumbra-zone), since testing web-workers in a service-worker context is complicated. 
- https://github.com/penumbra-zone/webgpu/issues/6 
- https://github.com/penumbra-zone/webgpu/pull/8

Once the necessary modifications are made to the Rust code, I can simulate the building process in parallel using web-workers in this isolated environment.

- https://github.com/penumbra-zone/webgpu/pull/9

### Integration 
There will need to be a strategy for eventually integrating this with the web-code according to

- https://github.com/penumbra-zone/web/issues/146

### Cleanup
[...]

PR #3289, opened 2023-11-07
feat(wasm): split-up build

TalDerei:
References #3236 and #3274 and https://github.com/penumbra-zone/penumbra/issues/3238. The build method has been refactored to enable separate web-workers to operate on sub-components of build process in parallel in the web context. The appropriate [methods](https://github.com/penumbra-zone/penumbra/blob/tx-decompose/crates/wasm/src/build.rs#L37) will be exposed to wasm using the `wasm-bindgen` attribute in the wasm crate. 

The PR includes a preliminary wasm unit testing suite using `wasm-bindgen-test` that mocks indexDB calls using an interactive browser. This simplified integration testing rather than having to compile and test locally in my [repo](https://github.com/penumbra-zone/webgpu) every time a change was made. The command to run the test suite is `wasm-pack test --chrome -- --test test_build --target wasm32-unknown-unknown --release`.


TalDerei:
The parallel build process is split into two phases:

- **Phase 1:** Spawn a distinct web-worker for each `ActionPlan`, and generate an `Action` by invoking the `action_builder` wasm function. The action_builder is generic on the ActionPlan. Under the hood, it uses pattern matching to distinguish between different ActionPlan variants, and invokes [Build(BuildPlan)](https://github.com/penumbra-zone/penumbra/blob/tx-decompose/crates/core/transaction/src/plan/action.rs#L34) to construct the proper `Action`. Each web worker will be supplied with an instance of the private witness. The result of the first phase will a list of Actions `[Action]`. 

- **Phase 2:** A top-level build method, `build_parallel`, will be exposed for constructing the transaction. Among other fields, `build_parallel` will require the `TransactionPlan` and `[Action]` and `AuthorizationData`. Internally, it will call the `build_tx_parallel` and `authorize` methods for constructing the `Transaction` and blinding signature. 


Screenshot 2023-11-08 at 1 05 50 PM


I generally noticed as I expose more of the lower-level methods in the build process to WASM, some of the internal structs are not serializable since they're exported from external libraries (arkworks) that don't implement the serialize / deserialize traits (for instance the synthetic_blinding_factor field in  UnauthTransaction: https://github.com/penumbra-zone/penumbra/blob/main/crates/core/transaction/src/plan/build.rs#L434). This construction was designed around this limitation. In the future, we shouldn't have to define extra data formats unless we're specifically exposing features from an external library like Arkworks.

hdevalence:
@TalDerei I'm not totally sure I understand the phase 1 / phase 2 distinction. Are these phases of implementation, or different phases of using the same API? If we want to expose parallelism to the web context, the unit of parallelism has to be tasks assigned to different web workers, right? I'm also not sure about adding a new `BuildPlan` variant to the `ActionPlan`, I think building is something we want to be able to _do_ to any `ActionPlan`, rather than a type of `ActionPlan` itself.

It would be good to see if we could reduce some of the code duplication we currently have between the existing parallel and non-parallel Rust methods, as a side effect of doing this wasm work.  

What about something like the following:

- [ ] Add an `ActionPlan::build_unauth(&self, witness_data: &WitnessData) -> Action` method to the `ActionPlan` enum.  Internally, this method would match on the `ActionPlan` variant and call the action-specific build methods inside. Because it doesn't have access to the `AuthorizationData`, it should fill any required authorization data in with dummy values (like an all 0s array).
- [ ] Add a `TransactionPlan::build_unauth_with_actions(&self, witness_data: &WitnessData, actions: Vec<Action>) -> Transaction` method to the `TransactionPlan`.  This would work like the existing `TransactionPlan::build` method, but it would slot in the provided prebuilt `actions` instead of using the `ActionPlan`s in the `TransactionPlan`. Because it doesn't have access to the `AuthorizationData`, it would need to fill in the binding signature with dummy values (like an all 0s array).
- [ ] Add a `TransactionPlan::authorize(&self, transaction: &Transaction, auth_data: &AuthorizationData) -> Transaction` method that would use the provided `auth_data` to overwrite all the dummy values in all of the `transaction`'s actions, then use the blinding factors in the transaction plan (`&self`) to derive the synthetic blinding factor needed to compute the binding signature.

The use of dummy values is unappealing, but the big upside of this approach is that we can use it in the Typescript/WASM context without having to add any additional serialization formats, and I think we can also clean up the existing Rust code:

- We can replace the existing `TransactionPlan::build` with an all-in-one method `TransactionPlan::build(&self, witness_data: &WitnessData, auth_data: &AuthorizationData) -> Transaction` that (1) calls `ActionPlan::build_unauth` on each action, (2) passes those to `TransactionPlan::build_unauth_with_actions`, (3) slots in the auth data with `TransactionPlan::authorize`, and then returns the completed transaction.

- We can replace the existing `TransactionPlan::build_concurrent` with an all-in-one method `TransactionPlan::build_concurrent(&self, witness_data: &WitnessData, auth_data: &AuthorizationData) -> Transaction` that is exactly like `TransactionPlan::build` but it wraps each call to `ActionPlan::build_unath` in a `tokio::spawn`. This gives almost perfect code reuse between `build` and `build_concurrent`.

- The web code can farm out the calls to `ActionPlan::build_unauth` to individual web workers, then do the `build_unauth_with_actions` and `authorize` steps once it gets the `AuthorizationData`. This means we can start the expensive part (proving) before getting the auth signatures, but still only have one step once we get the `AuthorizationData`.

- We can eliminate the `UnauthTransaction` type, I'm not sure any of the Rust code is using the delayed authorization feature, and the reason it exists is for this use case, which this new design would solve better.

TalDerei:
I should have clarified, `phases 1 / 2` are different phases of the same API. I think refactoring this way makes a lot of sense, and seems to be more idiomatic, while reducing code duplication at the same time! 

The only thing I'm wondering is why `ActionPlan::build_unauth` needs to do anything with the AuthorizationData dummy values in the first place?

hdevalence:
Well, suppose ActionPlan::build_unauth is called on an ActionPlan::SpendPlan. This has to return an Action::Spend with a fully constructed Spend inside, but that Spend has an auth_sig field. That field needs to be set to something, and the AuthorizationData is not available, so we can set it to a placeholder value instead.

TalDerei:
I see now, the `auth_sig` placeholder value was set to resemble the code screaming `AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA`.

TalDerei:
Currently the memo field isn't being handled properly. There's a divergence in the `wrappedMemoKey` field in an `Action` between the **serial** and **concurrent** build methods.

TalDerei:
@hdevalence breaking API changes: 

**penumbra-wasm package**
- new exposed WASM method [action_build](build_concurrent) in `wasm_planner.rs`
- new exposed WASM method [build_parallel](https://github.com/penumbra-zone/penumbra/blob/tx-decompose/crates/wasm/src/tx.rs#L198) in `tx.rs`
- modified the internals of the serial [build](https://github.com/penumbra-zone/penumbra/blob/tx-decompose/crates/wasm/src/tx.rs#L159) in `tx.rs` to reduce code duplication based on your suggestions. 

**penumbra-transaction package**

- [build_unauth](https://github.com/penumbra-zone/penumbra/blob/tx-decompose/crates/core/transaction/src/plan/action.rs#L87C12-L87C24) method in `action.rs`
- [build_unauth_with_actions](https://github.com/penumbra-zone/penumbra/blob/tx-decompose/crates/core/transaction/src/plan/build.rs#L28) method in `build.rs`
- [authorize_with_auth](https://github.com/penumbra-zone/penumbra/blob/tx-decompose/crates/core/transaction/src/plan/build.rs#L169) method in `build.rs`
- [build](https://github.com/penumbra-zone/penumbra/blob/tx-decompose/crates/core/transaction/src/plan/build.rs#L251) method with modified internals to reduce code duplication in `build.rs`
- [build_concurrent](https://github.com/penumbra-zone/penumbra/blob/tx-decompose/crates/core/transaction/src/plan/build.rs#L301) with modified internals to reduce code duplication in `build.rs`
- removal of `UnauthTransaction` type

We now have a all-in-one [build](https://github.com/penumbra-zone/penumbra/blob/tx-decompose/crates/core/transaction/src/plan/build.rs#L251) method that internally calls `build_unauth_with_actions` and `authorize_with_auth` per your suggestions. The conflict with this approach is existing calls to the build method in [main](https://github.com/penumbra-zone/penumbra/blob/main/crates/core/app/src/action_handler/actions/submit.rs#L345) separate the `build` and `authorize` methods since the authorize methods takes an extra field for randomness. I suggest we strip out the `authorize_with_auth` from the build method to handle this?

### Testing
I've manually tested the equivalence of the refactored serial and parallel build methods against the original build method. They yield the same transaction payload and proofs, except with different blinding signatures as expected.

Additionally, I'll be checking for any failing unit / integration tests. I'm wondering if we can further test the resulting proof generation using a verifier?

### TODOs

- The methods currently support `Spend` and `Output` actions. Add support for `Swap`, `SwapClaim`, etc.
- Add protobuf definitions for new API functions
- Additionally, we need to update `transaction_signing.md` doc to reflect the removal of the`UnauthTransaction` type. 

### Relevant 
According to one of the comments in [effect_hash](https://github.com/penumbra-zone/penumbra/blob/tx-decompose/crates/core/transaction/src/effect_hash.rs#L143): `"TransactionPlan::build builds the actions sorted by type, so hash the actions in the order they'll appear in the final transaction."` The web-workers will need to implement some ordering mechanism to place the actions in the correct order.

hdevalence:
> According to one of the comments in [effect_hash](https://github.com/penumbra-zone/penumbra/blob/tx-decompose/crates/core/transaction/src/effect_hash.rs#L143): "TransactionPlan::build builds the actions sorted by type, so hash the actions in the order they'll appear in the final transaction." The web-workers will need to implement some ordering mechanism to place the actions in the correct order.

This is wrong (or should be wrong, we should check that the implementation actually doesn't do this).  At one point in the past we attempted to do this, and then changed course.

The `EffectHash` should be computed on the actions in the order they are in the plan. Otherwise (as we experienced), there are all kinds of possibilities for subtle mismatches of ordering.  Instead, choosing an ordering of the actions is the responsibility of the planner; once the `TransactionPlan` is generated, it should be used as-is.

hdevalence:
> Additionally, I'm currently running the existing unit / integration tests to check for failures. What's the idiomatic approach to testing correctness here?

It should be sufficient to run the Rust tests (either with `cargo test` or [`cargo nextest`](https://nexte.st/)) and to push a PR and see that the smoke test (running `pcli` against `pd`) passes.

hdevalence:
> Our all-in-one [build](https://github.com/penumbra-zone/penumbra/blob/tx-decompose/crates/core/transaction/src/plan/build.rs#L251) method internally calls build_unauth_with_actions and authorize_with_auth per your suggestions. The conflict with this approach is that existing calls to the build method in [main](https://github.com/penumbra-zone/penumbra/blob/main/crates/core/app/src/action_handler/actions/submit.rs#L345) separate the build and authorize calls since authorize takes an extra field for randomness. I suggest we strip out the authorize_with_auth from the build method, and call the authorize_with_auth method separately to handle this? The downside to doing this is we'll need to clone the TransactionPlan between the build and authorize_with_auth calls. Unless they can take a reference to self rather than taking ownership?

This is a good catch. I think, though, that we should keep the existing API, and figure out a way to make it work.

At a high level, the `TransactionPlan` is supposed to be a fully deterministic description of the transaction to be built. So why do we need new randomness at all? Especially, why do we need randomness _after_ the authorization signatures are already computed? Why does `build_dao_transaction` have to jump through hoops to add fake randomness?

Looking through the existing code, where is the randomness actually used? It's only used in the computation of the binding signature:
```rust
        // Compute the binding signature and assemble the transaction.
        let binding_signing_key = rdsa::SigningKey::from(synthetic_blinding_factor);
        let auth_hash = transaction.transaction_body.auth_hash();

        let binding_sig = binding_signing_key.sign(rng, auth_hash.as_bytes());
        tracing::debug!(bvk = ?rdsa::VerificationKey::from(&binding_signing_key), ?auth_hash);
```
Why does the `decaf377-rdsa` `sign` take an RNG? `decaf377-rdsa` is an EdDSA variant, so it should support deterministic signing. If we changed the `decaf377-rdsa` API to support deterministic signing, we wouldn't need any RNG at all, and then we wouldn't have this problem.

crates/core/transaction/src/plan/action.rs
```diff
@@ -74,6 +77,55 @@ pub enum ActionPlan {
...
 
 impl ActionPlan {
+    /// Builds a planned [`Action`] specified by
+    /// the [`ActionPlan`].
+    /// Arguments:
+    ///     &self: `ActionPlan`
+    ///     fvk: `FullViewingKey`
+    ///     witness_data: `&WitnessData`
+    ///     memo_key: `Option<PayloadKey>`
+    /// Returns: `Action`
```
hdevalence, line 87: nit: it's better to leave these out, because Rustdoc will render the arguments based on the actual Rust code, and these will get mashed together with the rest of the markdown.

crates/core/transaction/src/plan/build.rs
```diff
@@ -14,72 +17,35 @@ use crate::{
...
         witness_data: WitnessData,
-    ) -> Result<UnauthTransaction> {
-        let mut actions = Vec::new();
-        let mut synthetic_blinding_factor = Fr::zero();
-
+    ) -> Result<Transaction> {
         // Add the memo.
         let mut memo: Option<MemoCiphertext> = None;
-        let mut memo_key: Option<PayloadKey> = None;
+        let mut _memo_key: Option<PayloadKey> = None;
```
hdevalence, line 36: What happened to the memo key?

crates/core/transaction/src/plan/build.rs
```diff
@@ -14,72 +17,35 @@ use crate::{
...
+    /// [`ActionPlan`]s in the TransactionPlan.
+    /// Arguments:
+    ///     self: `TransactionPlan`
+    ///     fvk: `FullViewingKey`
+    ///     witness_data: `WitnessData`
+    /// Returns: `Transaction`
+    pub fn build_unauth_with_actions(
         self,
-        fvk: &FullViewingKey,
+        mut actions: Vec<Action>,
```
hdevalence, line 30: Shouldn't `actions` be immutable, since it's preconstructed data?

crates/core/transaction/src/plan/build.rs
```diff
@@ -14,72 +17,35 @@ use crate::{
...
-                memo_key.as_ref().unwrap_or(&dummy_payload_key),
-            )));
+            _memo_key = Some(memo_plan.key);
         }
 
         // Build the transaction's swaps.
         for swap_plan in self.swap_plans() {
-            synthetic_blinding_factor += swap_plan.fee_blinding;
-            actions.push(Action::Swap(swap_plan.swap(fvk)));
+            actions.push(Action::Swap(swap_plan.swap(&fvk)));
```
hdevalence, line 48: Wait, why are we building the action here, shouldn't the `actions` come preconstructed?

crates/core/transaction/src/plan/build.rs
```diff
@@ -188,273 +153,56 @@ impl TransactionPlan {
...
+    ///     rng: `&mut R`
+    ///     auth_data: `&AuthorizationData`
+    ///     transaction: `Transaction`
+    /// Returns: `Transaction`
+    pub fn authorize_with_auth<R: CryptoRng + RngCore + Debug>(
+        &self,
+        rng: &mut R,
+        auth_data: &AuthorizationData,
+        mut transaction: Transaction,
+    ) -> Result<Transaction> {
```
hdevalence, line 177: Rather than `authorize_with_auth`, could this just be `authorize`?

In terms of API design, the reason that I suggested putting this on the `TransactionPlan`, rather than the `Transaction`, is so that the method is only relevant when you're already building transactions. Because of that, I think it's fine to just call it `authorize`.

crates/core/transaction/src/plan/build.rs
```diff
@@ -188,273 +153,56 @@ impl TransactionPlan {
...
-            let memo_key = memo_key.as_ref().unwrap_or(&dummy_payload_key).clone();
-            in_progress_output_actions.push(tokio::spawn(async move {
-                output_plan.output(&ovk, &memo_key)
-            }));
         }
-
-        // Start building the transaction's swaps.
-        let mut in_progress_swap_actions = Vec::new();
-        for swap_plan in self.swap_plans().cloned() {
+        for swap_plan in self.swap_plans() {
```
hdevalence, line 197: We have a lot of loops over each type of action here. It would be easy to miss some. What if instead we added a method `ActionPlan::value_blinding()` whose body was a `match` over the possible kinds of `ActionPlan`, returning `Fr::zero()` for action plans that don't have a value blinding (because their effect on the value balance is public).

Then this could just be
```
for action_plan in &self.action_plans {
    synthetic_blinding_factor += action_plan.value_blinding();
}
```

crates/core/transaction/src/plan/build.rs
```diff
@@ -488,13 +235,149 @@ impl UnauthTransaction {
...
+            _memo = memo_plan.memo().ok();
+            memo_key = Some(memo_plan.key);
+        }
+
+        let mut actions: Vec<Action> = Vec::new();
+
+        // We build the actions sorted by type, with all spends first, then all
+        // outputs, etc.  This order has to align with the ordering in
+        // TransactionPlan::effect_hash, which computes the auth hash of the
+        // transaction we'll build here without actually building it.
```
hdevalence, line 277: We should just iterate over the list of actions in the plan in the order they're in the plan. I believed that this had already been done as part of earlier `EffectHash` changes, but that seems to not have been the case: https://github.com/penumbra-zone/penumbra/blob/main/crates/core/transaction/src/effect_hash.rs#L142-L146

We need to not compute the `EffectHash` that way for code simplicity and to allow constrained devices like hardware wallets to stream the `ActionPlan`s rather than process a full `TransactionPlan` all at once.

crates/core/transaction/src/plan/build.rs
```diff
@@ -488,13 +235,149 @@ impl UnauthTransaction {
...
+        }
+        for output_plan in self.output_plans() {
+            let output = ActionPlan::Output(output_plan.to_owned());
+            let action = output
+                .build_unauth(full_viewing_key, &witness_data, memo_key.clone())
+                .expect("Build output action failed!");
+            actions.push(action);
+        }
+
+        // TODO: Handle other actions (swaps, swap claims, etc.).
```
hdevalence, line 295: Rather than having to handle each action type specially, this should just run over the action plans in one go:
```
for action_plan in &self.action_plans {
    actions.push(action_plan.build_unauth(full_viewing_key, &witness_data, &memo_key));
}
```

crates/core/transaction/src/plan/build.rs
```diff
@@ -488,13 +235,149 @@ impl UnauthTransaction {
...
+        let mut _memo: Option<MemoCiphertext> = None;
+        let mut memo_key: Option<PayloadKey> = None;
+        if self.memo_plan.is_some() {
+            let memo_plan = self
+                .memo_plan
+                .clone()
+                .ok_or_else(|| anyhow!("missing memo_plan in TransactionPlan"))?;
+            _memo = memo_plan.memo().ok();
+            memo_key = Some(memo_plan.key);
+        }
```
hdevalence, line 270: Do we need to build the memo prior to building the actions? Or can we just access the memo key from the plan?

crates/core/transaction/src/plan/build.rs
hdevalence: I fixed this in c3ec960bd (cc @redshiftzero @cronokirby)

crates/core/transaction/src/plan/build.rs
hdevalence: Also, elsewhere in the effect hash code I noticed that if the memo is missing, there's a zero payload key used instead, maybe that could work here too (cc @redshiftzero)?

crates/core/transaction/src/plan/action.rs
TalDerei: removed

crates/core/transaction/src/plan/build.rs
TalDerei: In an earlier comment in this PR, I mentioned: **"I'm trying to think through the best way to handle the memo field. There's also code duplication happening between the [build_action](https://github.com/penumbra-zone/penumbra/blob/tx-decompose/crates/wasm/src/wasm_planner.rs#L98), [build_unauth_with_actions](https://github.com/penumbra-zone/penumbra/blob/tx-decompose/crates/core/transaction/src/plan/build.rs#L34) and [build](https://github.com/penumbra-zone/penumbra/blob/tx-decompose/crates/core/transaction/src/plan/build.rs#L258) methods with respect to the memo field that need to be addressed."**. 

The way we've constructed this, the `memo_key` field isn't required in the computation of [build_unauth_with_actions](https://github.com/penumbra-zone/penumbra/blob/tx-decompose/crates/core/transaction/src/plan/build.rs#L28), and the `memo` field isn't required in the computation of [build](https://github.com/penumbra-zone/penumbra/blob/tx-decompose/crates/core/transaction/src/plan/build.rs#L261) and [build_concurrent](https://github.com/penumbra-zone/penumbra/blob/tx-decompose/crates/core/transaction/src/plan/build.rs#L321). The [build_actions](https://github.com/penumbra-zone/penumbra/blob/tx-decompose/crates/wasm/src/wasm_planner.rs#L97) method in the wasm crate also requires the memo field. 

Do you have any suggestions on how to unify the memo field across these different implementations to reduce code duplication?

crates/core/transaction/src/plan/build.rs
TalDerei: I haven't handled other actions besides `spend` and `output` actions in [build_action](https://github.com/penumbra-zone/penumbra/blob/tx-decompose/crates/wasm/src/wasm_planner.rs#L123) and [build_unauth](https://github.com/penumbra-zone/penumbra/blob/tx-decompose/crates/core/transaction/src/plan/action.rs#L126) yet. The build method will also need account for these [changes](https://github.com/penumbra-zone/penumbra/blob/tx-decompose/crates/core/transaction/src/plan/build.rs#L295). This is planned on being removed.

crates/core/transaction/src/plan/build.rs
TalDerei: Right now, I'd call the `actions` **_semi_** preconstructed data. If we think about this from the perspective of a web-worker, they each call [build_action](https://github.com/penumbra-zone/penumbra/blob/tx-decompose/crates/wasm/src/wasm_planner.rs#L76) which internally matches on the `Action` type. `build_action` should only be called for actions that are computationally intensive to compute. It therefore only precomputes the `output, spend, swap, swap-claim, and delegator-vote` actions. 

If we don't mark it as mutable, how should we handle pushing other [actions](https://github.com/penumbra-zone/penumbra/blob/tx-decompose/crates/core/transaction/src/plan/build.rs#L79) to the transaction body?

hdevalence: We explicitly shouldn't handle that, we should require that all of the actions are built and passed in, rather than trying to manage things case by case.

As a first-pass implementation strategy, I think it would be a good idea to unconditionally spawn a new web worker for each ActionPlan, but even if we did something more sophisticated, it wouldn't change this interface.

In general we should try to avoid action-by-action special case handling and instead handle ActionPlans and Actions uniformly, as with some of the other PR suggestions.

TalDerei: The web-workers are heavy instances (rather than simply being lightweight threads) since they replicate the underlying VM state. Spawning unnecessary web-workers would degrade performance since it increases the message passing overhead.

crates/core/transaction/src/plan/build.rs
TalDerei: removed comment

crates/core/transaction/src/plan/build.rs
TalDerei: There are also 22 variants for an `ActionPlan`, which means we'd need to spawn more threads than the number of physical cores on the device.

crates/core/transaction/src/plan/build.rs
redshiftzero: We don't need to build the memo first per se, we can just access the memo key from the `MemoPlan`, or indeed use `[0; 32]`. We only need the memo key in order to add it to any outputs.

TalDerei: https://github.com/penumbra-zone/penumbra/blob/main/crates/wasm/src/planner.rs#L547 prevents us from accessing the memo key directly from the plan. We could go with @hdevalence earlier suggestion and instead take self and consume the planner. I'll play around with it.

hdevalence: I don't think it does, that code is in the _planner_, which is producing the plan, not in the returned plan itself. The `build` method is on the plan and the plan is immutable, so there's not a problem here.

hdevalence:
`decaf377` `0.9.0` is out now and includes `sign_deterministic`.

crates/core/transaction/src/plan/build.rs
TalDerei: That's a great idea, I switched to this pattern!

crates/core/transaction/src/plan/build.rs
TalDerei: I implemented your suggestion as a first-pass strategy. The pre-built actions are now _immutable_ by default. I'll see how the performance scales in terms of execution time and latency in the **webgpu** repository and report back.

crates/core/transaction/src/plan/build.rs
TalDerei: The other action variants have been handled, and `actions` have been made [immutable](https://github.com/penumbra-zone/penumbra/pull/3289/commits).

hdevalence:
I rebased this on top of `main`, but I preserved the original state of the branch as `tx-decompose_r0`.

crates/core/app/src/action_handler/transaction.rs
```diff
@@ -179,11 +179,9 @@ mod tests {
         };
         let mut rng = OsRng;
         let tx = plan
-            .build_concurrent(&mut rng, fvk, witness_data)
+            .build_concurrent(fvk, &witness_data, &auth_data)
             .await
-            .expect("can build transaction")
-            .authorize(&mut rng, &auth_data)
-            .expect("can authorize transaction");
+            .expect("can't build transaction");
```
hdevalence, line 184: Note: this should be `.expect("can build transaction")`, since the convention for `.expect` is that the string names the expected invariant.

crates/core/component/dex/src/lp/plan.rs
```diff
@@ -127,6 +127,7 @@ impl PositionRewardClaimPlan {
         todo!()
     }
 
+    // FIXME: why is there an auth sig here??
```
hdevalence, line 130: Splitting out into follow-up issue

hdevalence:
I'm worried about this work getting tripped up on other changes -- like I mentioned on Discord, I didn't realize that the change to remove the custom ordering logic from the `TransactionPlan`'s `EffectHash` implementation didn't happen over the summer, so the plan I suggested was more complex than I'd intended, and now this PR has a larger scope than we would have liked.

For that reason, I tried carrying it further, filling in some missing match arms and cleaning up some of the code.

hdevalence (approved):
Happy to merge this on green CI unless there are other changes you'd like to get in.

We should squash the commits when merging, the history is a mess because of the rebasing.

We can fix any straggler bits in follow-up PRs.

crates/bin/pcli/tests/network_integration.rs
```diff
@@ -166,11 +166,15 @@ fn transaction_send_from_addr_0_to_addr_1() {
 
     let tvp: ProtoTransactionView = serde_json::value::from_value(view_json).unwrap();
     let tv: TransactionView = tvp.try_into().unwrap();
+    // TODO: the first may no longer be a spend because of ordering changes.
+    // Let's not try to fix this at the moment. Later we can put a "canonical ordering" into the planner.
+    /*
```
hdevalence, line 171: As follow-up, we should add a `TransactionPlan::sort_actions(&mut self)` and change the `Planner` to call it just before returning the final `TransactionPlan`, then re-enable this test (which implicitly depended on action ordering)

TalDerei:
We still need to make the relevant updates to the docs, for instance updating `transaction_signing.md` to reflect the removal of the `UnauthTransaction` type.

crates/wasm/tests/test_build.rs
```diff
@@ -0,0 +1,469 @@
...
+        error::WasmError,
+        storage::IndexedDBStorage,
+        tx::{authorize, build, build_parallel, witness},
+        wasm_planner::WasmPlanner,
+    };
+
+    #[wasm_bindgen_test]
+    async fn mock_build_serial_and_parallel() {
+        // Limit the use of Penumbra Rust libraries since we're mocking JS calls
+        // that are based on constructing objects according to protobuf definitions.
```
grod220, line 39: Why can't we import the structs instead of re-defining them here?

crates/core/transaction/src/plan/build.rs
TalDerei: much cleaner!

crates/wasm/tests/test_build.rs
```diff
@@ -0,0 +1,469 @@
...
+            "position": "3204061134848",
+            "source": {
+                "inner": "oJ9Bo9v22srtUmKdTAMVwPOuGumWE2cAuBbZHci8B1I="
+            }
+        }
+        "#;
+
+        // Convert note to `SpendableNoteRecord`.
+        let spendable_note: SpendableNoteRecord =
+            serde_json::from_str(spendable_note_json).unwrap();
```
grod220, line 138: I wonder if there's a way we can use serde_wasm_bindgen with JsValue here instead of serde_json 🤔 . Maybe it doesn't matter though.

crates/wasm/src/tx.rs
```diff
@@ -172,11 +173,54 @@ pub fn build(
...
+/// Building a transaction may take some time,
+/// depending on CPU performance and number of actions in transaction_plan
+/// Arguments:
+///     actions: `Vec<Actions>`
+///     transaction_plan: `pb::TransactionPlan`
+///     witness_data: `pb::WitnessData`
+///     auth_data: `pb::AuthorizationData`
+/// Returns: `pb::Transaction`
+#[wasm_bindgen]
+pub fn build_parallel(
```
grod220, line 197: I can't really tell, but is this where web workers will be spawned later?

crates/wasm/tests/test_build.rs
```diff
@@ -0,0 +1,469 @@
...
+        let serial_transaction = build(
+            full_viewing_key,
+            transaction_plan.clone(),
+            witness_data.as_ref().unwrap().clone(),
+            authorization_data.clone(),
+        )
+        .unwrap();
+        console_log!("Serial transaction is: {:?}", serial_transaction);
+    }
+}
```
grod220, line 466: I didn't see any assertions in the test, should we add some?

crates/wasm/src/error.rs
```diff
@@ -10,6 +10,7 @@ use web_sys::DomException;
 use penumbra_tct::error::{InsertBlockError, InsertEpochError, InsertError};
 
 pub type WasmResult<T> = Result<T, WasmError>;
+pub type WasmOption<T> = Option<T>;
```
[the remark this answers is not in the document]
TalDerei: removed

crates/wasm/src/lib.rs
```diff
@@ -1,15 +1,13 @@
...
-mod view_server;
-mod wasm_planner;
+pub mod note_record;
+pub mod planner;
+pub mod storage;
+pub mod swap_record;
+pub mod tx;
+pub mod utils;
+pub mod view_server;
+pub mod wasm_planner;
```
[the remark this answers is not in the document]
TalDerei: In order to make the modules visible to `wasm_bindgen_test`, otherwise the test suite will fail since the modules are private.

crates/wasm/Cargo.toml
```diff
@@ -43,8 +43,9 @@ serde-wasm-bindgen       = "0.5.0"
 thiserror                = "1.0"
 wasm-bindgen             = "0.2.87"
 wasm-bindgen-futures     = "0.4.37"
-wasm-bindgen-test        = "0.3.0"
+wasm-bindgen-test        = "0.3.37"
 web-sys                  = { version = "0.3.64", features = ["console"] }
+serde_json               = "1.0.107"
```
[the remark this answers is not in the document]
TalDerei: moved to dev-dependencies

crates/wasm/src/storage.rs
```diff
@@ -36,11 +39,52 @@ pub struct IndexedDBStorage {
 
 impl IndexedDBStorage {
     pub async fn new(constants: IndexedDbConstants) -> WasmResult<Self> {
-        let db_req: OpenDbRequest = IdbDatabase::open_u32(&constants.name, constants.version)?;
+        let mut db_req: OpenDbRequest = IdbDatabase::open_u32(&constants.name, constants.version)?;
+
+        // Conditionally create object stores in the `IdbDatabase` database for testing purposes
```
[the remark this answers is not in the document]
TalDerei: Attempting to migrate this to the test suite has been extremely unsuccessful over the past couple of days. It boils down to **1.** our decision to use `IdbDatabase`, a wrapper for the `IndexDB` database, and **2.** the tight [coupling](https://github.com/penumbra-zone/penumbra/issues/3274) of the `IndexedDBStorage` instance when we initialize the `ViewServer` / `WasmPlanner`. Triggering the creation of an object store in the database requires a an `onupgradeneeded` event, which requires a connection handle to the database instance. Unfortunately, the handle is [consumed](https://github.com/penumbra-zone/penumbra/blob/tx-decompose/crates/wasm/src/storage.rs#L79) in the creation of the database. Anyways, we can't create the database in the test suite directly because any changes made to it won't affect the database instance specific to the `ViewServer` / `WasmPlanner`. 

Dependency injection via a constructor or setter would resolve this, but It doesn't seem to work in this environment since `Serialize` is not implemented for `indexed_db_futures::IdbDatabase`. 

For instance, something like
  ```
  pub async fn set_storage(&mut self, storage: JsValue) -> WasmResult<()>{
        let storage = serde_wasm_bindgen::from_value(storage)?;
        self.storage = Some(storage)
    }
```

would require somehow converting a `JsValue` into `IndexedDBStorage` or vice-versa. 

TLDR; the conditional inside the storage module is the simplest way to mock IndexDB calls and bypass the aforementioned limitations in this environment. I think we can keep it for now.

File: crates/wasm/src/tx.rs [hunk 21]
```diff
@@ -172,11 +173,54 @@ pub fn build(
...
+    let plan_proto: pb::TransactionPlan = serde_wasm_bindgen::from_value(transaction_plan)?;
+
+    let witness_data_proto: pb::WitnessData = serde_wasm_bindgen::from_value(witness_data)?;
+    let witness_data_: WitnessData = witness_data_proto.try_into()?;
+
+    let auth_data_proto: pb::AuthorizationData = serde_wasm_bindgen::from_value(auth_data)?;
+    let auth_data_: AuthorizationData = auth_data_proto.try_into()?;
+
+    let plan: TransactionPlan = plan_proto.try_into()?;
+    let actions_: Vec<Action> = serde_wasm_bindgen::from_value(actions)?;
```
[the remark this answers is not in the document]
TalDerei: there's no semantic meaning beyond distinguishing from the `actions` variable being passed in. Is there a better naming convention?

crates/core/transaction/src/plan/build.rs
```diff
@@ -413,48 +52,42 @@ impl TransactionPlan {
...
     ) -> Result<Transaction> {
         // Do some basic input sanity-checking.
-        let spend_count = self.inner.spends().count();
+        let spend_count = transaction.spends().count();
         if auth_data.spend_auths.len() != spend_count {
             anyhow::bail!(
                 "expected {} spend auths but got {}",
                 spend_count,
                 auth_data.spend_auths.len()
             );
```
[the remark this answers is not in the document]
TalDerei: It returns early with an error, but we can change it.

crates/wasm/src/wasm_planner.rs
```diff
@@ -55,6 +64,142 @@ impl WasmPlanner {
...
+        let transaction_plan_: TransactionPlan = transaction_plan_proto.try_into()?;
+
+        let witness_data_proto: pb::WitnessData = serde_wasm_bindgen::from_value(witness_data)?;
+        let witness_data_: WitnessData = witness_data_proto.try_into()?;
+
+        let action_proto: pb::ActionPlan = serde_wasm_bindgen::from_value(action_plan)?;
+        let action_plan_: ActionPlan = action_proto.try_into()?;
+
+        let full_viewing_key: FullViewingKey = FullViewingKey::from_str(full_viewing_key)
+            .expect("The provided string is not a valid FullViewingKey");
```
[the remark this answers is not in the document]
TalDerei: ah right, we're moving away from `unwrap` and `expect`. removed

crates/wasm/src/wasm_planner.rs
```diff
@@ -55,6 +64,142 @@ impl WasmPlanner {
...
+        utils::set_panic_hook();
+
+        let transaction_plan_proto: tp = serde_wasm_bindgen::from_value(transaction_plan.clone())?;
+        let transaction_plan_: TransactionPlan = transaction_plan_proto.try_into()?;
+
+        let witness_data_proto: pb::WitnessData = serde_wasm_bindgen::from_value(witness_data)?;
+        let witness_data_: WitnessData = witness_data_proto.try_into()?;
+
+        let action_proto: pb::ActionPlan = serde_wasm_bindgen::from_value(action_plan)?;
+        let action_plan_: ActionPlan = action_proto.try_into()?;
```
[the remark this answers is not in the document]
TalDerei: Nice! made the change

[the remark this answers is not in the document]
TalDerei: Nice! It doesn't work for `WitnessData` and `tct:Proof` don't implement the **Serialize** trait.

crates/wasm/src/wasm_planner.rs
```diff
@@ -55,6 +64,142 @@ impl WasmPlanner {
...
+        let mut memo_key: Option<PayloadKey> = None;
+        if transaction_plan_.memo_plan.is_some() {
+            let memo_plan = transaction_plan_
+                .memo_plan
+                .clone()
+                .ok_or_else(|| anyhow!("missing memo_plan in TransactionPlan"))?;
+            memo_key = Some(memo_plan.key);
+        }
+
+        //
```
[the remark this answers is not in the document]
TalDerei: removed

crates/wasm/tests/test_build.rs
TalDerei: we'd need to mark the fields on the `IndexedDbConstants` and `Tables` structs as **pub**, since the test suite can't access the private fields.

crates/wasm/src/storage.rs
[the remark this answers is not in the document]
grod220: Ah, indeed that is tricky. But if the db instance is expected, we couldn't create the object store? Like this in the test:

```rust
let database: *const IdbDatabase = storage_ref.get_database();
database.create_object_store(...);
```

However, if this is not possible, think it would be worth extracting this logic into a separate method (create_tables_for_test()) so the ::new() method isn't pull of this business logic.

crates/wasm/src/tx.rs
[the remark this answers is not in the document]
hdevalence: Yes, a better convention would be to shadow the original bindings:

let actions = ...;

crates/core/transaction/src/plan/build.rs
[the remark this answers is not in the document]
hdevalence: I'm confused — bail!() does return an error rather than panicking.

crates/wasm/src/wasm_planner.rs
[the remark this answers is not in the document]
hdevalence: Can you file an issue for this?

crates/wasm/src/tx.rs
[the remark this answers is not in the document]
grod220: Ah, I see what you mean. I suppose you could shadow the originals by doing `let auth_data = ...`, but maybe the underscore isn't a big deal. I just wasn't so aware if it was a convention for something.

crates/core/transaction/src/plan/build.rs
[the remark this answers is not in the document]
grod220: If it returns an Error type, it's ok. But I wonder if bail specifically panics/aborts. This is the behavior that isn't too safe and causes unexpected behavior given it can't be caught.

hdevalence:
Sounds good. I think it would be good to make some issues to track follow up work, but to merge this in the meantime for the avoidance of other possible conflicts.

crates/wasm/tests/test_build.rs
grod220: I think that would be a worthwhile tradeoff instead of re-writing the structs. What do you think?

crates/core/transaction/src/plan/build.rs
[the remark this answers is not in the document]
hdevalence: No, bail!() specifically doesn't panic — its purpose is to have something that's ergonomically like panic!() but returns an error instead.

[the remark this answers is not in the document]
grod220: Ah, just seeing Henry's comment. We are good then 👍

crates/wasm/src/storage.rs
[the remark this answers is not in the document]
TalDerei: This won't work since it'll complain that the creation of an object store requires an `upgradeneeded` event. 

I think a separate method for testing purposes would work well here.

crates/wasm/src/wasm_planner.rs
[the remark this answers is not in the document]
TalDerei: https://github.com/penumbra-zone/penumbra/issues/3400

crates/wasm/src/tx.rs
TalDerei: The web-workers will call the `build_action` method in the wasm planner, and then the prebuilt actions will be slotted into `build_parallel` where the final transaction can be assembled.

crates/wasm/tests/test_build.rs
TalDerei: The assertions here would be tough because we're essentially comparing if a serial transaction and parallel transaction match. The only difference in their transaction payload should be the binding signature since it's based on randomness. I've been instead manually inspecting if they match in the dev-console, but some kind of assertion would be nice.

crates/wasm/tests/test_build.rs
TalDerei: @hdevalence is there any issues with making these fields pub?

PR #3291, opened 2023-11-08
view: hack attempting to mitigate split-brain GRPC failures

hdevalence:
This isn't a permanent fix, and it's not the right way to solve the robustness issue (more details TK) but it may help with the immediate problem with `summonerd`, and seems unlikely to add new failure modes.

### DIFF
--- a/crates/view/src/worker.rs
+++ b/crates/view/src/worker.rs
@@ -1,6 +1,7 @@
 use std::{
     collections::BTreeSet,
     sync::{Arc, Mutex},
+    time::Duration,
 };
 
 use anyhow::Context;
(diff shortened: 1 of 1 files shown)


cronokirby:
What's with the protobuf lint failures?

hdevalence:
I think it's comparing the state of this branch against `main`, even though the PR is being made against a different branch. Probably the solution is to change the CI workflow to filter to only apply to PRs against `main`.

cronokirby:
Also, are we sure that one second is enough?

hdevalence:
> Also, are we sure that one second is enough?

We are not! But it seems likely, and fits in the middle of the confirmation delay we already put in `pcli`. If it isn't, we'll see it really clearly in the logs, because there will be a warning before an error. So I think it's probably a reasonable starting point.

cronokirby:
> > Also, are we sure that one second is enough?
> 
> We are not! But it seems likely, and fits in the middle of the confirmation delay we already put in `pcli`. If it isn't, we'll see it really clearly in the logs, because there will be a warning before an error. So I think it's probably a reasonable starting point.

Yeah, makes sense, if we still run into this we could do a few retries with exponential backoff, but I'm in favor of merging this as an obviously correct patch for now

cronokirby (approved):

Issue #3292, opened 2023-11-08
Restrict Buf Lint workflow to PRs against `main`


hdevalence:
> I think it's comparing the state of this branch against `main`, even though the PR is being made against a different branch. Probably the solution is to change the CI workflow to filter to only apply to PRs against `main`.

_Originally posted by @hdevalence in https://github.com/penumbra-zone/penumbra/issues/3291#issuecomment-1802176484_

PR #3293, opened 2023-11-08
Summonerd: add link to blog post in main page

cronokirby:


conorsch:
The protobuf CI check is failing because we're targeting a feature branch, and the CI check isn't smart enough to compare against the target branch, it always targets `main`.

conorsch:
> CI check isn't smart enough to compare against the target branch, it always targets main.

Already filed as https://github.com/penumbra-zone/penumbra/issues/3292

Issue #981, opened 2022-06-08
Implement ICS20 balance-of-payments or other circuit-breaker functionality.


hdevalence:
We can add more complex functionality later, but to start, we should just have *a* part of the code that does basic accounting and circuit-breaking:

- [x] Record the net amount of each token transferred between each counterparty chain (this may be partially accounted for by the generic supply tracking functionality in the shielded pool -- we already track the supply of all of the bearer tokens, so we just need to record how many tokens we've sent out to each counterparty chain);
- [x] For assets originating from Penumbra, never accept the return of more tokens than were transferred out.
- [x] For assets not originating from Penumbra, never transfer out more tokens than were transferred in.  (Our counterparties should be checking this anyways, since if we were Byzantine we could lie to them).

conorsch:
@avahowell suspects this work may be complete already. @avahowell can you comment with more details, and close if warranted?

conorsch:
Discussed in IBC sync today. Good news: 2/3 of these are already done! Leaving this issue open to track the remaining item.

erwanor:
This was done in #3294

PR #3294, opened 2023-11-08
ics20: implement balance of payments for non-penumbra origin assets

avahowell:
This should close the remaining balance of payments task in #981.

### DIFF
--- a/crates/core/component/shielded-pool/src/component/transfer.rs
+++ b/crates/core/component/shielded-pool/src/component/transfer.rs
@@ -381,7 +381,7 @@ async fn recv_transfer_packet_inner<S: StateWrite>(
             .await?
             .unwrap_or_else(Amount::zero);
 
-        let new_value_balance = value_balance + value.amount;
+        let new_value_balance = value_balance.saturating_add(&value.amount);
         state.put(
             state_key::ics20_value_balance(&msg.packet.chan_on_b, &denom.id()),
             new_value_balance,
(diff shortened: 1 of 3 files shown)


crates/core/component/shielded-pool/src/component/transfer.rs
```diff
@@ -335,6 +359,21 @@ async fn recv_transfer_packet_inner<S: StateWrite>(
...
+        // update the value balance
+        let value_balance: Amount = state
+            .get(&state_key::ics20_value_balance(
+                &msg.packet.chan_on_b,
+                &denom.id(),
+            ))
+            .await?
+            .unwrap_or_else(Amount::zero);
+
+        let new_value_balance = value_balance + value.amount;
```
erwanor, line 372: I have two questions:
- should this be a `saturating_add`? Since depositing is permissionless, I could spin-up a chain with infinite issuance and make this panic
- why is this in `chan_open_init_check` vs. `recv_transfer_packet_inner`? my understanding is that we want to tally all inbound Ics20 transfers, but this method is only executed during the channel creation handshake? is that correct?

conorsch:
@avahowell can you take another look, addressing @erwanor's comments? We'd like this to land for 64 (#3408).

crates/core/component/shielded-pool/src/component/transfer.rs
avahowell: This is in recv_transfer_packet_inner, i think it's just a rendering bug making it look like it's in open_init

hdevalence:
Is this ready to land in testnet 64? cc @avahowell

Issue #2136, opened 2023-03-10
Bug: fix client scanning to not record 0-value swap outputs


zbuc:
**Describe the bug**
The view client scanning is supposed to ignore 0-value notes, but this is not happening.

hdevalence:
Did we ever get more information about why it's not happening? I checked the code and the check is there.

Maybe the 0-value check is only applying to _scanned_ notes, and not ones we derive deterministically as swap outputs?

hdevalence:
I'm pretty sure that's exactly the problem. To fix this, we'll want to change the view server not to record the 0-valued outputs of a swapclaim.

hdevalence:
In the meantime, a client state can be fixed by opening the database in `sqlite3` and running
```sql
DELETE FROM spendable_notes WHERE ROWID IN (SELECT spendable_notes.ROWID FROM spendable_notes JOIN notes ON spendable_notes.note_commitment = notes.note_commitment WHERE notes.amount = 0);
```

hdevalence:
Looks like this is affecting testnet users: https://discord.com/channels/824484045370818580/915710988312002610/1087793758935978045

aubrika:
> not to record the 0-valued outputs of a swapclaim

Unsure how exactly to calculate this for swaps because I'm unfamiliar with that implementation - looking at the `BatchSwapOutputData`

`pub struct BatchSwapOutputData {
    pub delta_1: Amount,
    pub delta_2: Amount,
    pub lambda_1_1: Amount,
    pub lambda_2_1: Amount,
    pub lambda_1_2: Amount,
    pub lambda_2_2: Amount,
    pub height: u64,
    pub trading_pair: TradingPair,
}
`
Do we just want to exclude `SwapClaimPlan`s where delta_1 and delta_2 are both zero, or should a more complex calculation be done here?

hdevalence:
I think it won't be necessary to interpret the BSOD at all, at some point in the code the view service already computes the swap claim output notes in order for them to be recorded. We just need to trace through how that happens, and filter out the notes before we record them.

hdevalence:
On mobile but I think this is how they're getting in:

![image](https://user-images.githubusercontent.com/44879/227385177-769e5621-9585-48a5-b050-db18e9ce9270.jpeg)

![image](https://user-images.githubusercontent.com/44879/227385118-d64e5d8c-dade-4a13-b460-1538a9f1e409.jpeg)

[...]

PR #3299, opened 2023-11-09
Fix broken list-tx-hashes function for swap claims

### LINKED ISSUE #2547: `SwapClaim` transactions are not viewable by the view server

### LINKED ISSUE #3214: Bug: `pcli view balance` not properly consolidating denoms
As shown here:

```
❯ pcli -n https://grpc.testnet-preview.penumbra.zone view balance
Scanning blocks from last sync height 976 to latest height 976
 Account  Amount          
 0        20000gm         
 0        501000test_usd  
 0        100cube         
 0        10000pizza      
 0        1002700penumbra 
 0        20000gn         
 0        8000test_usd    
 0        6300penumbra
```

Obviously this is suboptimal, could be a query issue. To be fixed.

aubrika:
View: removes code to skip storage of output notes of value 0, adds code to include transaction ids for incoming swaps to filtered blocks.

Closes #2547
Closes #3214

### DIFF
--- a/crates/view/src/storage.rs
+++ b/crates/view/src/storage.rs
@@ -955,7 +955,13 @@ impl Storage {
                     }
                 }
                 let amount = record.note.amount();
-                output.push(record);
+
+                // Only display notes of value > 0
+
+                if amount.value() > 0 {
+                    output.push(record);
+                }
+
                 // If we're tracking amounts, accumulate the value of the note
                 // and check if we should break out of the loop.
                 if amount_cutoff {
@@ -1169,11 +1175,6 @@ impl Storage {
     }
 
     pub async fn give_advice(&self, note: Note) -> anyhow::Result<()> {
-        // Do not insert advice for zero amounts, simply return Ok because this is fine
-        if u128::from(note.amount()) == 0u128 {
-            return Ok(());
-        }
-
         let pool = self.pool.clone();
         let mut lock = pool.get()?;
         let dbtx = lock.transaction()?;
@@ -1254,7 +1255,7 @@ impl Storage {
         spawn_blocking(move || {
             pool.get()?
                 .prepare(&format!(
-                    "SELECT nullifier FROM (SELECT nullifier FROM spendable_notes UNION SELECT nullifier FROM swaps) WHERE nullifier IN ({})",
+                    "SELECT nullifier FROM (SELECT nullifier FROM spendable_notes UNION SELECT nullifier FROM swaps UNION SELECT nullifier FROM tx_by_nullifier) WHERE nullifier IN ({})",
                     nullifiers
                         .iter()
                         .map(|x| format!("x'{}'", hex::encode(x.0.to_bytes())))
--- a/crates/bin/pcli/src/command/view/balance.rs
+++ b/crates/bin/pcli/src/command/view/balance.rs
@@ -1,7 +1,7 @@
 use anyhow::Result;
 use comfy_table::{presets, Table};
 use penumbra_asset::{asset::Cache, Value};
-use penumbra_keys::{keys::AddressIndex, FullViewingKey};
+use penumbra_keys::FullViewingKey;
 use penumbra_view::ViewClient;
 #[derive(Debug, clap::Args)]
 pub struct BalanceCmd {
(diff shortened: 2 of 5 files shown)


crates/view/src/storage.rs
```diff
@@ -1169,11 +1169,6 @@ impl Storage {
     }
 
     pub async fn give_advice(&self, note: Note) -> anyhow::Result<()> {
-        // Do not insert advice for zero amounts, simply return Ok because this is fine
-        if u128::from(note.amount()) == 0u128 {
-            return Ok(());
-        }
```
redshiftzero, line 1175: wow really nice find! Two thoughts:
1. I think we'd need to apply this same change in the view impl over in `IndexedDBStorage::store_advice` in the wasm crate, such that the same fix is applied over there
2. The implication of this change is that #2136 is re-introduced, so maybe we do something like filter out 0-valued notes when querying for spendable notes (then users won't see behavior like https://discord.com/channels/824484045370818580/915710988312002610/1087796172078469281). What do you think?

aubrika: Confirmed that this reintroduces the issue mentioned previously in discord:

```Scanning blocks from last sync height 156681 to latest height 160295
[0s] ██████████████████████████████████████████████████    3614/3614    4133/s ETA: 0s
 Account  Amount      
 0        1gm         
 0        0gm         
 0        9gm         
 0        0gm         
 0        0gm         
 0        996test_usd 
 0        1test_usd   
 0        5penumbra   
 0        4penumbra   
 0        0gn         
 0        0gn         
 0        7gn         
 0        1gn         
 0        1gn         
 0        1gn         
 0        30penumbra  
aubrey@penpro penumbra % 
```

I like your suggestion - the balances query needs fixing anyway (it's showing several rows for one denom instead of properly consolidating denoms, as in #3214 ), so it seems like not including zero-valued notes as "spendable" is something we always should have been doing. Good catch on also doing this in the wasm crate - I'll dig into this tomorrow, and if we're lucky this will also shed some light on the existing balances query issue. Thank you for the review!

PR #3300, opened 2023-11-09
pd: add `--enable-expensive-rpc` flag

erwanor:
This PR adds a `--simulate-trade-service` flag to `pd` that activates the DEX trade simulation service, which allow users to send `SimulateTrade` RPCs. The service is disabled by default, because it can be onerous to run it against a rich orderbook, and therefore is a potential DoS vector.

### DIFF
--- a/crates/bin/pd/src/main.rs
+++ b/crates/bin/pd/src/main.rs
@@ -21,6 +21,7 @@ use pd::testnet::{
     join::testnet_join,
 };
 use pd::upgrade::{self, Upgrade};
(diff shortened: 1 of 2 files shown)


conorsch:
@erwanor mind if I tack on the necessary deployment changes to this PR, so that we can use this on preview once it's merged? I can also follow up in a separate PR if you'd prefer.

erwanor:
Not at all, thanks actually

hdevalence:
Not to bikeshed, but I'm wondering if this could be --enable-expensive-rpc or something similar, so that we could just have one flag and if we add some other similar RPC later we could have it all grouped into one setting.

PR #3302, opened 2023-11-09
specifying `rustc` version requirement in `pcli` installation docs

bekauz:
Just went through the process of installing `pcli` and got the following error:
> failed to select a version for `ibc-proto` which could resolve this conflict

Fixed it by updating the `rustc` version to `1.7.3`. Perhaps specifying it in the docs will save others some time.


hdevalence:
Thanks!!

Issue #3304, opened 2023-11-09
Summonerd sqlite schema falls down under load


hdevalence:
**Describe the bug**

`summonerd` stores contribution blobs inline with the contribution metadata, causing it to fall down under load.

**To Reproduce**

The service is currently down.

conorsch:
Load times on the phase 1 contrib page (https://summoning.penumbra.zone/phase/1) has dramatically improved post-migration. Closing this as resolved.

PR #3305, opened 2023-11-09
summonerd: add new schema separating blob data

hdevalence:
Towards #3304

### DIFF
--- a/tools/summonerd/src/storage.rs
+++ b/tools/summonerd/src/storage.rs
@@ -105,12 +105,12 @@ impl Storage {
         let tx = conn.transaction()?;
 
         tx.execute(
-            "INSERT INTO phase1_contributions VALUES (0, 1, NULL, NULL, ?1)",
-            (current_time_unix(),),
-        )?;
-        tx.execute(
             "INSERT INTO phase1_contribution_data VALUES (0, ?1)",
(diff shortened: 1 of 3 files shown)


redshiftzero:
Summonerd smoke test is passing as of commit 4456f85d8cfe32ef514f2971ec5a31022f7a2a31, the migration process has been tested by:

1. Checking out `summonerd-feature-branch`, starting a phase 1 ceremony locally and doing 1 phase 1 contribution. This results in two rows in the `phase1_contributions` table.
2. Checking out commit 4456f85d8cfe32ef514f2971ec5a31022f7a2a31, and running the migration via:

```
sqlite3 /tmp/summonerd/ceremony.db < old-to-new-migration.sql
```
3. Then starting the summoner again, verifying one can do a phase 1 contribution without issue ✅

redshiftzero:
There was a smoke test failure [here](https://github.com/penumbra-zone/penumbra/actions/runs/6814474555/job/18531419797) but it seems unrelated to the changes in this PR

redshiftzero (approved):

cronokirby (approved):
I looked at storage.rs in isolation, I think every query was migrated; also, looking at the proposed query plans in sqlite, I think the queries all avoid scanning the blob table as well.

Migrations ran and passed locally.

PR #3310, opened 2023-11-10
staking: reverse staking rate calculations (wip)

erwanor:
This is a WIP rebased branch. I think we will end up cherry picking commits because there are some gnarly conflicts on code generated serde files, among other things.


hdevalence:
Based on looking at the code, we determined it will be simpler to rewrite these changes from scratch (possibly using this code as a reference) than to rebase them.

Issue #3298, opened 2023-11-08
Windows builds are broken


conorsch:
**Describe the bug**
It's not currently possible to build Penumbra on Windows platforms. 

**To Reproduce**
1. Boot a Windows machine and install rust and git, and LLVM (for `clang`).
2. Run `cmd.exe` 
3. Clone Penumbra repo, `chdir` into it, check out `v0.63.1`, and run `cargo build --release`.
4. Observe error message:

```
error[E0793]: reference to packed field is unaligned
    --> C:\Users\conor\.cargo\registry\src\index.crates.io-6f17d22bba15001f\ntapi-0.3.7\src\ntexapi.rs:2783:52
     |
2783 |         *tick_count.QuadPart_mut() = read_volatile(&(*USER_SHARED_DATA).u.TickCountQuad);
     |                                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |
     = note: packed structs are only aligned by one byte, and many modern architectures penalize unaligned field accesses
     = note: creating a misaligned reference is undefined behavior (even if that reference is never dereferenced)
     = help: copy the field contents to a local variable, or replace the reference with a raw pointer and use `read_unaligned`/`write_unaligned` (loads and stores via `*p` must be properly aligned even when using raw pointers)

error[E0793]: reference to packed field is unaligned
    --> C:\Users\conor\.cargo\registry\src\index.crates.io-6f17d22bba15001f\ntapi-0.3.7\src\ntexapi.rs:2807:25
     |
2807 |         ((read_volatile(&(*USER_SHARED_DATA).u.TickCountQuad)
     |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |
     = note: packed structs are only aligned by one byte, and many modern architectures penalize unaligned field accesses
     = note: creating a misaligned reference is undefined behavior (even if that reference is never dereferenced)
     = help: copy the field contents to a local variable, or replace the reference with a raw pointer and use `read_unaligned`/`write_unaligned` (loads and stores via `*p` must be properly aligned even when using raw pointers)

   Compiling js-sys v0.3.65
   Compiling serde_qs v0.8.5
   Compiling infer v0.2.3
   Compiling encode_unicode v0.3.6
   Compiling async-process v1.8.1
   Compiling quote v1.0.33
   Compiling headers v0.3.9
   Compiling async-dup v1.2.2
[...]

PR #3319, opened 2023-11-14
ci: update cargo-dist for performing releases

conorsch:
In this commit we move from cargo-dist 0.0.4 to 0.4.2. The tooling updates are considerable, as the version number implies. To test, I regenerated the workflow YAML and run it many times in a private fork of the repo. With the notable change that we're no longer building binaries for Windows (#3298), everything else looks to work pretty well. Even better, the runtime for the workflow has decreased from ~60m to ~20m, with the biggest win skipping Windows builds.

Refs #3279, #3298.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,143 +1,183 @@
+# Copyright 2022-2023, axodotdev
+# SPDX-License-Identifier: MIT or Apache-2.0
+#
 # CI that:
 #
 # * checks for a Git Tag that looks like a release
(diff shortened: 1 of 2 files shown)


conorsch:
Given the extensive testing performed in a private fork, I'm happy with this logic, and so I'm merging it to main. After merge, I'd like to ship a point release, but doing so will require creating a feature branch from 0.63.1, since we've merged breaking changes into main, and cherry-picking over a few CI fixes, such as this one. I'll handle that work later this week.

PR #3325, opened 2023-11-14
Create an in-repo implementation of FROST for Decaf377-RDSA

cronokirby:
Featuring protobufs.

The reason I opted towards doing this was that I realized we would have otherwise needed to use the serde features of FROST, which I don't like because they involved silly things like:
- modifying decaf377-rdsa to implement Serialize for a dummy ciphersuite marker struct, for no reason
- putting json encoded as bytes inside a protobuf

Instead, we use protobufs all the way down, using the fact that frost-core will actually use the serialization types you define yourself. If you make these all `Vec<u8>` it's pretty convenient to marshall things to and from protobuf generated structs.

This also makes it very easy to embed the FROST protocol inside a larger more specific protocol, which we'll need to do for the threshold custody MVP. (I realized this when implementing that, and side stepped back to do this as a separate initial PR).


cronokirby:
(drafting to wait for improvements to make the API to be ergonomic, as I like at how we use it in the E2E test)

redshiftzero (approved):
LGTM although I'm not sure why [the smoke test is failing as it](https://github.com/penumbra-zone/penumbra/actions/runs/6867563336/job/18676191603?pr=3325) looks unrelated to the changes here

proto/penumbra/penumbra/crypto/decaf377_frost/v1alpha1/decaf377_frost.proto
```diff
@@ -0,0 +1,50 @@
+syntax = "proto3";
+
+package penumbra.crypto.decaf377_frost.v1alpha1;
```
hdevalence, line 3: Having this as a standalone package seems fine. I think though that we should decide where the _Penumbra_ threshold messages should go. (By this I mean messages where the input to the signing ceremony is a `TransactionPlan` and the output is an `AuthorizationData`, rather than the "raw" FROST where the input is an arbitrary message and the output is a signature).

Previously we'd stubbed these out as part of Narsil, but I think we may want to put them in the `custody` package and put the threshold custody implementation as part of the `custody` crate. WDYT?

cronokirby: I was going to put them in `custody.threshold`

redshiftzero:
@erwanor figured out the build failure here was due to https://github.com/cjriches/serde_unit_struct/issues/2. That release has been yanked, so a `cargo update` here should resolve

PR #3331, opened 2023-11-14
custody: fill in scope of CustodyService

hdevalence:
This fills in missing methods that allow the CustodyService to be the interface to a device like a hardware wallet, or to allow websites to request confirmation for displayed addresses.

Result of pairing with @redshiftzero and @cronokirby .


proto/penumbra/penumbra/custody/v1alpha1/custody.proto
```diff
@@ -19,16 +19,30 @@ import "penumbra/core/transaction/v1alpha1/transaction.proto";
 service CustodyProtocolService {
   // Requests authorization of the transaction with the given description.
   rpc Authorize(AuthorizeRequest) returns (AuthorizeResponse);
+
+  // Requests the full viewing key from the custodian.
+  //
+  // Custody backends should decide whether to honor this request, and how to
+  // control access to it.
+  rpc ExportFullViewingKey(ExportFullViewingKeyRequest) returns (ExportFullViewingKeyResponse);
```
hdevalence, line 27: @grod220 The web extension should implement this method by erroring out.

proto/penumbra/penumbra/custody/v1alpha1/custody.proto
```diff
@@ -19,16 +19,30 @@ import "penumbra/core/transaction/v1alpha1/transaction.proto";
...
+
+  // Displays an address to a user for confirmation.
+  //
+  // Custody backends with user interaction should present the address to the
+  // user and wait for explicit confirmation before returning.
+  //
+  // Non-interactive custody backends may return immediately.
+  rpc ConfirmAddress(ConfirmAddressRequest) returns (ConfirmAddressResponse);
```
hdevalence, line 35: @grod220 The web extension should implement this method by creating an extension popup.  The point of this is to allow web content to steer users to copy an address they'll send to someone else from trusted UI. Realistically, though, we won't be able to ensure that users don't ever copy an address from the web content, but we can give an option for trusted display.

grod220: So this will trigger a popup for a user to copy _any_ of their addresses? Sort of like this?
 
Screenshot 2023-11-15 at 10 11 32 AM

hdevalence: Yeah, the idea is to provide a way to confirm an address on a trusted display path controlled by the custodian. For a Ledger this is viewing the address on screen, for the web extension this would be some kind of popup trigger. I don't think it's a super high priority in the web extension but something that would be nice to have eventually

Issue #2712, opened 2023-06-14
add a `LazyProvingKey` and rework loading of proving keys


redshiftzero:
Currently the proving keys are only present in the `penumbra-proof-params` crate if the `proving-keys` feature is enabled: that feature uses the [`include_bytes!` macro](https://doc.rust-lang.org/std/macro.include_bytes.html) to bundle the generated proving key files.  The proving keys are each of the type `Lazy<ProvingKey<..>>`.

Instead we could always have the proving keys be present, i.e. `SPEND_PROOF_PROVING_KEY`, etc. will no longer be conditionally included if the `proving-keys` feature is enabled. The `proving-keys` feature would instead control whether the proving key object is loaded in the crate.

We'd create a `LazyProvingKey` type that internally holds a `OnceCell<ProvingKey>`. It behaves just like the `Lazy` type: it implements `Deref`, except if you dereference it without the `proving-keys` feature enabled or without loading the proving key from elsewhere, it will panic.

We’d have an API, e.g. a method on `LazyProvingKey::load` that takes a byte slice. If the `proving-keys` feature was not enabled, you must call this loading method. Internally it will call `get_or_init` on the `OnceCell`: https://docs.rs/once_cell/latest/once_cell/sync/struct.OnceCell.html#method.get_or_init.

Tl;dr: if you want everything bundled together, you enable the `proving-keys` feature as before, and everything works in exactly the same way. If you _don’t_ use that feature, then it's your responsibility to call the load method with the bytes of the proving key before doing anything with the proving keys, else the program panics. The benefit of this is that you have the option of loading from a file, and not having to balloon the size of the binary.

conorsch:
Some context: the dapp website was previously loading a ~90MB wasm file, which delayed TTFP. That's no longer the case, given that Zpoken retooled things so that the wasm logic is consumed exclusively by the extension, rather than by the dapp. It'd still be nice to resolve this problem, but let's wait until we have some dedicated web story support.

cronokirby:
Another additional thing we could do here is always encode the hash of the proving keys in the binary itself, that way the binary acts as a root of trust for this information. This way, the loading method can be trustless, like via a public API we control, or any other kind of CDN. The API proposed above would also work, just that the `load` method would also check that the hash of the bytes passed in matches.

hdevalence:
[...]

PR #3337, opened 2023-11-15
refactor: bundled-proving-keys

TalDerei:
References https://github.com/penumbra-zone/penumbra/issues/2712. A new `LazyProvingKey` type is introduced to represent the proving keys. The limitation with the former `Lazy` type was that `Lazy::new()` expects a closure. This ultimately creates a tight coupling with the declaration of a static lazy variable and its initialization. The `LazyProvingKey` enables this separation based on the conditional `bundled-proving-keys` feature. 

This will further reduce the wasm binary size to only include the verification keys, originally from **~90MB** to a few MB.


crates/crypto/proof-params/src/lib.rs
```diff
@@ -31,22 +73,32 @@ pub mod spend {
...
 
 /// Proving key for the output proof.
-#[cfg(feature = "proving-keys")]
-pub static OUTPUT_PROOF_PROVING_KEY: Lazy<ProvingKey<Bls12_377>> =
-    Lazy::new(proving_keys::output_proving_parameters);
+pub static OUTPUT_PROOF_PROVING_KEY: Lazy<LazyProvingKey> = Lazy::new(|| {
+    let mut output_proving_key = LazyProvingKey::new();
 
-/// Proving key for the spend proof.
+    #[cfg(all(feature = "proving-keys", not(target_arch = "wasm32")))]
```
hdevalence, line 79: Here and elsewhere I don't think we want to have any arch-specific `cfg`s.

Rather, we should either:

- have the `proving-keys` feature enabled, in which case the proving key is bundled into the binary (regardless of architecture)
- have the `proving-keys` feature not enabled, in which case either:
  - a byte slice with the proving key data is passed to an initialization function before proving, or
  - nothing is passed before proving, causing a panic.
  
  This behavior shouldn't depend on architecture, though, it should just be code configuration.

hdevalence: Also, maybe we should rename the feature to `bundled-proving-keys`?

hdevalence:
> This will further reduce the wasm binary size to only include the verification keys, originally from ~90MB to a few MB.

Nice!!!!

crates/crypto/proof-params/src/lib.rs
TalDerei: The `proving-keys` feature in the wasm crate is intrinsically enabled since its complex chain of dependencies require it. I'm trying to think through an optimal design approach here that caters to both configurations. 

Can we somehow make the behavior depend on the source crate compiling the code rather than the target architecture? For instance, if we're building the binary from the wasm crate, it'll detect that and conditionally skip the code block above. Something like `#[cfg(all(feature = "proving-keys", not(env = "wasm_crate")))]`. 

Marking the `proving-keys` feature not enabled with `#[cfg(not(feature = "proving-keys"))]` would require changing the existing calls (https://github.com/penumbra-zone/penumbra/blob/main/crates/bin/pcli/tests/proof.rs#L33) to something like `let pk = &SPEND_PROOF_PROVING_KEY.load(byte_slice)`. Ideally we shouldn't need to make code modifications outside the `proof-params` package.

hdevalence: Right now, various library crates need to enable the `proving-keys` feature, because it currently controls whether those items _exist at all_ in the source code. With the change I suggested, (let's assume we did the rename), we'd have a feature `bundled-proving-keys` that controls whether the proving keys are _pre-initialized with bundled data_.

In that world, there is no need for any library to toggle the feature on, and nothing that requires the wasm crate to have it on.  Rather, libraries would never enable the feature, and _binaries_ (the roots of the dependency graph) would add a dependency on the `proof-params` crate with the `bundled-proving-keys` feature enabled. Since features are unified across the entire dep tree, this allows individual binaries to opt in to bundled proving keys or not.

crates/crypto/proof-params/src/lib.rs
```diff
@@ -17,12 +19,52 @@ pub use traits::{
...
+#[cfg(feature = "proving-keys")]
+impl LazyProvingKey {
+    // Lazily construct proving key
+    pub fn new() -> Self {
+        LazyProvingKey { 
+            inner: OnceCell::new(),
+        }
+    }
+
+    pub fn load(&self, key: &str) -> ProvingKey<Bls12_377> {
```
redshiftzero, line 37: I think `load` should take a byte slice, and then attempt to deserialize into a ProvingKey in this method

crates/crypto/proof-params/src/lib.rs
```diff
@@ -17,12 +19,52 @@ pub use traits::{
...
-/// Verifying key for the spend proof.
+/// Note: Conditionally load the proving key objects in the crate
+/// unless the target is wasm. 
+
+/// Proving key for the spend proof.
+pub static SPEND_PROOF_PROVING_KEY: Lazy<LazyProvingKey> = Lazy::new(|| {
+    let mut spend_proving_key = LazyProvingKey::new();
+
+    #[cfg(all(feature = "proving-keys", not(target_arch = "wasm32")))]
+    spend_proving_key.inner.get_or_init(proving_keys::spend_proving_parameters);
```
redshiftzero, line 62: Here we can do this or we can call `load` with the bytes from e.g. `include_bytes!("gen/spend_pk.bin")`

redshiftzero (commented):
thanks for picking this up!

crates/crypto/proof-params/src/lib.rs
TalDerei: Yes I haven't changed the type signature yet!

TalDerei: The type signature for the load method has been changed accordingly.

redshiftzero (approved):
great!

crates/core/component/dex/src/swap/plan.rs
```diff
@@ -54,8 +54,8 @@ impl SwapPlan {
     }
 
     /// Construct the [`SwapProof`] required by the [`swap::Body`] described by this [`SwapPlan`].
-    #[cfg_attr(docsrs, doc(cfg(feature = "proving-keys")))]
-    #[cfg(feature = "proving-keys")]
+    // #[cfg_attr(docsrs, doc(cfg(feature = "proving-keys")))]
```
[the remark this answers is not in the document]
TalDerei: removed

crates/bin/pcli/Cargo.toml
```diff
@@ -106,7 +106,7 @@ tempfile = "3.3.0"
 regex = "1.6.0"
 penumbra-governance            = { path = "../../core/component/governance", default-features = false }
 penumbra-proof-params = { path = "../../crypto/proof-params", features = [
-    "proving-keys",
+    "bundled-proving-keys",
     "download-proving-keys",
```
hdevalence, line 110: should `download-proving-keys` imply `bundled-proving-keys`? probably, right? and in that case maybe it should be `download-bundled-proving-keys`?

crates/bin/pclientd/Cargo.toml
```diff
@@ -64,6 +65,10 @@ assert_cmd = "2.0"
 base64 = "0.20"
 ibc-types = { version = "0.7.0" }
 ibc-proto = { version = "0.37.0", default-features = false, features = ["server"] }
+penumbra-proof-params = { path = "../../crypto/proof-params", features = [
+    "bundled-proving-keys",
+    "download-proving-keys",
+] }
```
hdevalence, line 71: i think we don't want to unconditionally enable these features, if we have feature control of our own above. either we should enable them unconditionally and not have the features in this binary (seems preferable?) or we should only enable them conditionally.

crates/bin/pcli/Cargo.toml
hdevalence: Ah, never mind, I missed that this was in the dev-dependencies section.

crates/bin/pclientd/Cargo.toml
hdevalence: ditto missed that this was a dev feature

crates/crypto/proof-params/src/lib.rs
```diff
@@ -14,15 +15,55 @@ pub use traits::{
...
+        }
+    }
+
+    pub fn load(&self, bytes: &[u8]) -> ProvingKey<Bls12_377> {
+        self.inner
+            .get_or_init(|| {
+                ProvingKey::deserialize_uncompressed_unchecked(bytes)
+                    .expect("Failed to deserialize ProvingKey!")
+            })
+            .to_owned()
```
hdevalence, line 39: How does this `load` method relate to the existing `load_proving_parameters`? That method checks the provided data against hardcoded hashes of the proving key. Could we combine them?

hdevalence: I think we could do it if we made the `LazyProvingKey` take a `&'static str`!

hdevalence: Fixed in cb83595

hdevalence:
@TalDerei This looks great! I added another commit that merges the check-ID behavior from the existing `load_proving_keys` method so that we know that proving keys are always loaded with the correct values, if they're loaded at all.

PR #3338, opened 2023-11-15
Try restoring SQLite pragmas

### LINKED ISSUE #3329: Investigate potential Sqlite performance regression
**Describe the bug**

A user reported that syncing a fresh wallet was estimated to take 13min and was only scanning ~180blocks/sec, rather than the expected ~8000 I saw testing on a Macbook.

> strange, its on a linux threadripper box
> too many cores, 256gb of ram, everything on nvme

One idea about what might be happening is that the Rust view server does a sqlite transaction to write data after every nonempty block, and this may be triggering expensive fsyncs. This would explain the performance difference with MacOS, which improves fsync performance by lying to userspace about data persistence.

In the past, I remember personally writing code that changed the sqlite configuration about data consistency in order to get better performance, though I don't recall the exact parameters. 

However looking at the connect method here 

https://github.com/penumbra-zone/penumbra/blob/main/crates/view/src/storage.rs

I don't see related parameters. Maybe they were lost in the port off of the old sqlite bindings?

We should pull up the code as it was before porting to r2d2 and see what it did.

hdevalence:
Closes #3329

This restores configurations we had prior to migrating to rusqlite.

Even on MacOS this results in speed improvements syncing the chain with a newly generated wallet:
```
~/c/p/penumbra3 ((v0.63.1)|✔) $ cargo run --quiet --release --bin pcli -- --home /tmp/pcli-scratch-sync-test/ view sync
Scanning blocks from last sync height 0 to latest height 241904
[28s] ██████████████████████████████████████████████████  241904/241904  8429/s ETA: 0s
```
vs
```
~/c/p/penumbra2 (sqlite-pragmas|✔) $ cargo run --quiet --release --bin pcli -- --home /tmp/pcli-scratch-sync-test/ view sync
2023-11-15T05:52:24.778662Z ERROR r2d2: database is locked
2023-11-15T05:52:24.778666Z ERROR r2d2: database is locked
Scanning blocks from last sync height 0 to latest height 241981
[18s] ██████████████████████████████████████████████████  241981/241981  13035/s ETA: 0s
```

Note however that now there are errors. So this isn't a complete solution.  Nothing will conflict with this PR, so there's no rush to merge it.

### DIFF
--- a/crates/view/src/storage.rs
+++ b/crates/view/src/storage.rs
@@ -108,6 +108,10 @@ impl Storage {
                     OpenFlags::default() & !OpenFlags::SQLITE_OPEN_URI,
                 )
                 .with_init(|conn| {
+                    // "NORMAL" will be consistent, but maybe not durable -- this is fine,
+                    // since all our data is being synced from the chain, so if we lose a dbtx,
+                    // it's like we're resuming sync from a previous height.
+                    conn.execute_batch("PRAGMA journal_mode=WAL; PRAGMA synchronous=NORMAL;")?;
                     // We use `prepare_cached` a fair amount: this is an overestimate of the number
                     // of cached prepared statements likely to be used.
                     conn.set_prepared_statement_cache_capacity(32);
--- a/crates/view/src/storage.rs
+++ b/crates/view/src/storage.rs
@@ -117,7 +117,11 @@ impl Storage {
                     conn.set_prepared_statement_cache_capacity(32);
                     Ok(())
                 });
-            Ok(r2d2::Pool::new(manager)?)
+            Ok(r2d2::Pool::builder()
+                // We set max_size=1 to avoid "database is locked" sqlite errors,
+                // when accessing across multiple threads.
+                .max_size(1)
+                .build(manager)?)
         } else {
             let manager = SqliteConnectionManager::memory();
             // Max size needs to be set to 1, otherwise a new in-memory database is created for each


conorsch:
> I had to reset/re-sync several times before I was able to generate an error.

The errors occurred fairly reliably on db initialization, but not on sync resumption after init:

```
❯ pcli view sync
2023-12-01T02:14:56.231178Z ERROR r2d2: database is locked    
Scanning blocks from last sync height 0 to latest height 490283
[3s] █░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░   11113/490283  3068/s ETA: 2m
```

Adding a bunch of tracing statements showed that the errors were occurring specifically during the `r2d2::Pool::new(manager)?` call. Simply restricting the maximum number of concurrent connections in the pool to 1, same as we already do for the in-memory db config, resolved. I made sure to rerun a full sync (~500k blocks) both with and without the new commit, and observed no difference in sync time, neither positive nor negative. Sync is now absurdly fast for me again:

```
❯ time pcli view sync
Scanning blocks from last sync height 0 to latest height 490412
[1m] ██████████████████████████████████████████████████  490412/490412  6468/s ETA: 0s

real	1m16.440s
user	2m14.132s
sys	0m12.950s
```

@hdevalence if you agree these changes are solid, I'll squash, merge, and cherry-pick onto a point release.

hdevalence:
Seems good! I don't have particular insight into the sqlite internals but the explanation seems solid

cronokirby:
Ran this independently; can confirm no errors on sync, and a speedup on Linux.

cronokirby (approved):

Issue #3332, opened 2023-11-14
Panic instead of error when sending funds to invalid address


Valentine1898:
When I try to send funds to `penumbra1rtjzkgm553lyqufnr4sp368dp8rajlm2p96hmclld3r4wcqapezraav04xq5849ksq7qf8jvhvv37y4uqxsjv4s684au3vc2g203f8um2f0rwsgm8qjf7anvqgrgdf0safyxsr9` I get panic, even though I expect to get an invalid address error.

Related issue on web repo: https://github.com/penumbra-zone/web/issues/232

I sent this address to faucet channel and it seems to have broken it for a while 

```
thread 'main' panicked at crates/core/transaction/src/plan/clue.rs:33:64:
valid address: InvalidAddress
stack backtrace:
   0: rust_begin_unwind
             at /rustc/75b064d26970ca8e7a487072f51835ebb057d575/library/std/src/panicking.rs:645:5
   1: core::panicking::panic_fmt
             at /rustc/75b064d26970ca8e7a487072f51835ebb057d575/library/core/src/panicking.rs:72:14
   2: core::result::unwrap_failed
             at /rustc/75b064d26970ca8e7a487072f51835ebb057d575/library/core/src/result.rs:1653:5
   3: penumbra_transaction::plan::clue::CluePlan::clue
   4: penumbra_transaction::effect_hash::<impl penumbra_transaction::plan::TransactionPlan>::effect_hash
   5: penumbra_transaction::plan::auth::<impl penumbra_transaction::plan::TransactionPlan>::authorize
   6: penumbra_custody::soft_kms::SoftKms::sign
   7: <penumbra_custody::soft_kms::SoftKms as penumbra_proto::penumbra::custody::v1alpha1::custody_protocol_service_server::CustodyProtocolService>::authorize::{{closure}}
   8: <<penumbra_proto::penumbra::custody::v1alpha1::custody_protocol_service_server::CustodyProtocolServiceServer<T> as tower_service::Service<http::request::Request<B>>>::call::AuthorizeSvc<T> as tonic::server::service::UnaryService<penumbra_proto::penumbra::custody::v1alpha1::AuthorizeRequest>>::call::{{closure}}
   9: <penumbra_proto::penumbra::custody::v1alpha1::custody_protocol_service_server::CustodyProtocolServiceServer<T> as tower_service::Service<http::request::Request<B>>>::call::{{closure}}
[...]

PR #3340, opened 2023-11-15
Use result instead of expect for clues

grod220:
Closes https://github.com/penumbra-zone/penumbra/issues/3332

Uses of `unwrap()` are in tests, fyi.

Re-arrangement of imports seem to be a RustRover IDE thing. Think it's ok, but can revert if preferred otherwise.

### DIFF
--- a/crates/core/transaction/src/effect_hash.rs
+++ b/crates/core/transaction/src/effect_hash.rs
@@ -1,4 +1,6 @@
+use anyhow::Result;
 use blake2b_simd::Params;
+
 use decaf377_fmd::Clue;
 use penumbra_chain::EffectHash;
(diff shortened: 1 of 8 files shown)


crates/core/transaction/src/effect_hash.rs
```diff
@@ -635,7 +638,7 @@ mod tests {
 
         let plan_effect_hash = plan.effect_hash(fvk);
 
-        let auth_data = plan.authorize(rng, &sk);
+        let auth_data = plan.authorize(rng, &sk).unwrap();
```
grod220, line 641: In a test fyi

hdevalence:
This isn't really an issue of expect()s but something more fundamental we need to think through about how we work with clue keys — can discuss synchronously.

PR #3384, opened 2023-11-15
add view struct for memos with `AddressView` for return address

### LINKED ISSUE #2780: MemoView does not have an AddressView
**Describe the bug**

The MemoView's Visible variant has a MemoPlaintext, defined as

```
message MemoPlaintext {
	penumbra.core.crypto.v1alpha1.Address sender = 1;
	string text = 2;
}
``` 

This only has an Address, not an AddressView, but it would be better if software consuming a MemoView had an AddressView instead so that they can determine whether or not the address has known structure (eg rendering it as Account 3 or similar).

However, as viewing an address requires a full viewing key while decrypting the memo only requires a payload key, we will need to think through how to thread any necessary information through the decryption + TransactionPerspective + TransactionView stack.

redshiftzero:
Closes #2780

This PR:
* Modifies the `MemoView` to instead of having two variants, the `MemoPlaintext` and `MemoCiphertext`, to replace the `MemoPlaintext` with a `MemoPlaintextView`.
* The `MemoPlaintextView` contains the string memo data along with an `AddressView` (instead of an `Address`).
* When we construct `TransactionPerspective`s, we need to add the address view data for the memo return address.


aubrika (approved):
LGTM!

Issue #3333, opened 2023-11-14
Wasm planner does not support source control


Valentine1898:
Currently wasm planner does not support source control and adds to plan notes regardless of source 

Related web issue https://github.com/penumbra-zone/web/issues/161

PR #3387, opened 2023-11-15
added source control support to wasm planner

Valentine1898:
Closes https://github.com/penumbra-zone/penumbra/issues/3333

Changes this PR will entail in web repo -  https://github.com/penumbra-zone/web/pull/245

### DIFF
--- a/crates/wasm/src/storage.rs
+++ b/crates/wasm/src/storage.rs
@@ -1,6 +1,7 @@
 use indexed_db_futures::prelude::OpenDbRequest;
 use indexed_db_futures::{IdbDatabase, IdbQuerySource};
 use serde::{Deserialize, Serialize};
+use wasm_bindgen::JsValue;
 use web_sys::IdbTransactionMode::Readwrite;
 
 use penumbra_asset::asset::{DenomMetadata, Id};
@@ -10,7 +11,7 @@ use penumbra_proto::DomainType;
 use penumbra_sct::Nullifier;
 use penumbra_shielded_pool::{note, Note};
 
-use crate::error::WasmResult;
+use crate::error::{WasmError, WasmResult};
 use crate::note_record::SpendableNoteRecord;
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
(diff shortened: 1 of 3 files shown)


crates/wasm/src/storage.rs
grod220, line 1: What do you think about this? Attempting to simplify these nested closures.

```rust
pub async fn get_notes_v2(
        &self,
        request: NotesRequest,
    ) -> WasmResult<Vec<SpendableNoteRecord>> {
        let idb_tx = self
            .db
            .transaction_on_one(&self.constants.tables.spendable_notes)?;
        let store = idb_tx.object_store(&self.constants.tables.spendable_notes)?;

        let raw_values = store.get_all()?.await?;
        let parsed_notes = raw_values
            .into_iter()
            .filter_map(|js_value| self.parse_note(js_value, &request).ok())
            .collect();

        Ok(parsed_notes)
    }

    fn parse_note(
        &self,
        js_value: JsValue,
        request: &NotesRequest,
    ) -> WasmResult<SpendableNoteRecord> {
        let note: SpendableNoteRecord = serde_wasm_bindgen::from_value(js_value)?;

        let asset_id_matches = match &request.asset_id {
            Some(asset_id) => note.note.asset_id() == asset_id.clone().try_into()?,
            None => true,
        };

        let address_index_matches = match &request.address_index {
            Some(address_index) => note.address_index.eq(&address_index.clone().try_into()?),
            None => true,
        };

        if asset_id_matches && address_index_matches && note.height_spent.is_none() {
            Ok(note)
        } else {
            Err(WasmError::Anyhow(anyhow::anyhow!(
                "Note does not match the request"
            )))
        }
    }
```

crates/wasm/src/storage.rs
```diff
@@ -52,18 +52,35 @@ impl IndexedDBStorage {
...
-                    None => Some(note),
-                })
+                        None => Some(note),
+                    })
+                    .and_then(
+                        |note: SpendableNoteRecord| match request.address_index.clone() {
+                            Some(address_index) => {
+                                if note
+                                    .address_index
+                                    .eq(&address_index.try_into().expect("invalid address index"))
```
grod220, line 74: We should not use `.expect()` except for in tests. Causing panics cause unexpected behaviors in the frontend.

crates/wasm/src/storage.rs
Valentine1898: I agree, the code looks cleaner this way, and at the same time we get rid of `.expect()`

Valentine1898:
I was still able to test these changes on web yesterday, but now I need to update all buf dependencies on the web and fix any type compatibility issues.

grod220 (approved):

PR #3398, opened 2023-11-17
staking: rework epoch transition rate calculations

erwanor:
Follow-up to #2909, this PR creates a simple `DistributionManager` with stub methods, and rework the staking logic that is executed during epoch transitions.


crates/core/component/distributions/src/component.rs
```diff
@@ -17,21 +22,144 @@ pub struct Distributions {}
...
+                        let issuance: u64 = issuance.try_into().expect("total issuance is within `u64`");
+                        tracing::debug!(%component, ?issuance, "issuing tokens to component"
+                        );
+                        match component {
+                            Staking => state.set_staking_issuance(issuance),
+                            Dex => todo!("set dex issuance"),
+                        }
+                    }
+                }
+        */
```
hdevalence, line 137: This can probably be cut out for the same reasons as in #3310 right?

erwanor:
Will merge on green CI.

PR #3403, opened 2023-11-17
IBC Governance Action Support

### LINKED ISSUE #2173: IBC-related governance proposals
We don't want to forget to have governance proposals for modifying / mutating IBC state:

- [ ] Performing unplanned upgrades of an IBC connection
- [ ] Freezing IBC clients
- [ ] Unfreezing IBC clients

zbuc:
Closes #2173


crates/core/app/src/action_handler/actions/submit.rs
```diff
@@ -203,6 +223,30 @@ impl ActionHandler for ProposalSubmit {
             ProposalPayload::UpgradePlan { .. } => {
                 // TODO(erwan): no stateful checks for upgrade plan.
             }
+            ProposalPayload::UnplannedIbcUpgrade { .. } => {
+                // TODO: which checks happen here?
+                todo!();
```
erwanor, line 228: I suggest we ship this PR with Freeze/Unfreeze and leave out trying to handle arbitrary client modifications for now, and remove the `todo!()`. The reason for this is that we do not have a great idea yet of the both the scope of changes that need to happen there.

avahowell: Agree with this

avahowell (changes requested):

conorsch:
We are letting this slip from Testnet 64, and will return after shipping.

crates/core/component/governance/src/proposal.rs
```diff
@@ -202,6 +212,15 @@ pub enum ProposalKind {
     /// An upgrade proposal.
     #[cfg_attr(feature = "clap", clap(display_order = 500))]
     UpgradePlan,
+    /// An unplanned IBC connection modification.
+    #[cfg_attr(feature = "clap", clap(display_order = 600))]
+    UnplannedIbcUpgrade,
```
avahowell, line 217: should this be removed now?

crates/core/component/governance/src/component/view.rs
```diff
@@ -875,8 +879,30 @@ pub trait StateWriteExt: StateWrite {
...
+                let client_id = &ClientId::from_str(client_id)
+                    .map_err(|e| tonic::Status::aborted(format!("invalid client id: {e}")))?;
+                let current_height = self
+                    .get_block_height()
+                    .await
+                    .expect("block height should be set");
+                let client_state = self.get_client_state(client_id).await?;
+                let revision_number = self.get_revision_number().await?;
+
+                let freeze_height = IbcHeight::new(revision_number, current_height)?;
```
avahowell, line 892: I think this height is wrong, it should be the height at which the client was frozen, in terms of that client's height (not the host chain's height). So we should be pulling the height from the client_state here instead.

avahowell (changes requested):

crates/core/component/governance/src/component/view.rs
erwanor: Ah yes, makes sense

avahowell (approved):

Issue #3409, opened 2023-11-17
Change staking system to do fixed issuance per block


hdevalence:
**Is your feature request related to a problem? Please describe.**

As part of #3398 @erwanor and I cleaned up some of the staking rates implementation. The next step is to reverse the direction of the base rate computation, so that the rates are determined by issuance rather than vice versa.

To do this, we should pick the simplest possible issuance formula, then implement it in the distributions component. This gets the structure right, and then we can port the logic from the Cosmos SDK for computing the issuance later.

**Describe the solution you'd like**

- [ ] Add chain parameters to distributions component, initially with one parameter, `staking_issuance_per_block`. Set this to `1penumbra`. (This is enough to be measurable but probably not enough to really mess up the testnet if it's wrong).
- [ ] Add end-epoch logic to distributions component that computes the staking issuance as `num_blocks * staking_issuance_per_block` and writes this in a way the staking component can read.
- [ ] Change the staking component to read the issuance in its end-epoch logic and use the formulas in #2909 to compute the `next_base_rate` from the new issuance and the `prev_base_rate`.

**Additional context**

Later, we can port over the PID controller from the Cosmos SDK.

erwanor:
This is done in #3413 and #3415. Next steps are to implement a simple controller that performs issuance based on an inflation window and a target amount of bonded stake, similar to what the Cosmos SDK does in x/mint.

PR #3415, opened 2023-11-20
distributions: compute and distribute inflation per epoch

erwanor:
Part of #3409, this PR uses the new distributions component chain parameter (`staking_issuance_per_block`) to compute the per-epoch inflation budget and store it in the object store.

### DIFF
--- a/crates/core/component/chain/src/component/view.rs
+++ b/crates/core/component/chain/src/component/view.rs
@@ -53,6 +53,12 @@ pub trait StateReadExt: StateRead {
             .map(|params| params.epoch_duration)
     }
 
+    async fn get_epoch_for_height(&self, height: u64) -> Result<Epoch> {
+        self.get(&state_key::epoch_by_height(height))
+            .await?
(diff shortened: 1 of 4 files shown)


crates/core/component/stake/src/component.rs
```diff
@@ -324,22 +326,61 @@ pub(crate) trait StakingImpl: StateWriteExt {
...
+        // MERGEBLOCK(erwan): unsafe casting to `u64`. Should we use `Amount`s?
+        let next_base_rate = prev_base_rate.next(base_reward_rate.value() as u64);
+        tracing::debug!(
+            ?prev_base_rate,
+            ?next_base_rate,
+            ?base_reward_rate,
+            ?total_active_stake_previous_epoch,
+            ?issuance_budget_for_epoch,
+            "calculated base rate for the upcoming epoch"
+        );
```
hdevalence, line 383: I think this shouldn't be merge-blocking, but we should instead fold it into an issue about reworking the staking arithmetic (in general) to use the same `U128x128` codepaths as other arithmetic.

PR #3419, opened 2023-11-20
storage: implement ICS23 proofs for multistore

avahowell:
This PR

- Updates snapshot::get_with_proof to return an ics23 multi-proof, allowing (existence and nonexistence) proofs to substore keys
- Removes the AppHash construction, replaces it with the root hash of the multistore, and replaces the apphash proof spec with the multistore proof spec


crates/core/component/chain/src/component.rs
```diff
@@ -1,7 +1,22 @@
...
+use once_cell::sync::Lazy;
+
+// the commitment prefix for the IBC state in the penumbra chain.
+pub static PENUMBRA_IBC_COMMITMENT_PREFIX: Lazy<MerklePrefix> = Lazy::new(|| MerklePrefix {
+    key_prefix: "ibc/".as_bytes().to_vec(),
+});
+
+// the ICS23 proof spec for penumbra's IBC state; this can be used to verify proofs
+// for other substores in the penumbra state, provided that the data is indeed inside a substore
+// (as opposed to directly in the root store.)
```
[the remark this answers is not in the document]
erwanor: same thing on the merkle prefix const

avahowell:
Added prefixing to all IBC keys, and initialized pd with substore prefix for ibc

crates/core/component/ibc/src/prefix.rs
```diff
@@ -0,0 +1,34 @@
...
+pub static IBC_COMMITMENT_PREFIX: Lazy<MerklePrefix> = Lazy::new(|| MerklePrefix {
+    key_prefix: IBC_SUBSTORE_PREFIX.as_bytes().to_vec(),
+});
+
+/// the ICS23 proof spec for penumbra's IBC state; this can be used to verify proofs
+/// for other substores in the penumbra state, provided that the data is indeed inside a substore
+/// (as opposed to directly in the root store.)
+pub static IBC_PROOF_SPECS: Lazy<Vec<ics23::ProofSpec>> = Lazy::new(|| {
+    vec![
+        penumbra_storage::ics23_spec(),
```
avahowell, line 17: I'm wondering if we should move the proof spec to the storage crate, but this is a bit tricky since the correct proof spec to use now depends on the key (whether or not it's in a substore)

hdevalence: I don't think it should move to the storage crate, I think we should say that this is the proof spec for the IBC data only.

erwanor:
Merging optimistically, so that I can make some quick changes without creating conflicts. I am going to modify the internals of key routing and harden the API to disallow empty keys or overwriting substore roots. The reason for the former is that I have learned that even though ibc-go allows a lot of freedom in prefix definition, it is actually in fact considered an identifier (and not a path, like the proto message hints) by the ICS24 spec (see "constant byte string"). This means that we're restricted to alphanumeric characters for our prefixes. In practice, this is tolerated by both ibc-rs and ibc-go. Sorry for the churn, and thanks for the feedback

PR #3420, opened 2023-11-20
Update Grafana Dashboards

voteblake:
Two changes to address issues I had when reusing the Grafana dashboards for a full node and validator:

- The ABCI dashboard relies on a metric that CometBFT emits with a name that now includes the unit (seconds)
- The Sync, Transactions, and Penumbra dashboards have at least some hardcoded Prometheus plugin UID's. Turning this into a parameter makes the dashboard reusable, and is the approach the other dashboards take.
- The Validators dashboards looks for a `role` label with the value `consensus` on the validator counts that is not emitted
- The Penumbra dashboard looks for a `worker` label with the value `consensus` on the validator counts that is not emitted - it also creates a panel that performs a group operation over `workers` filtered to `consensus` value, which, if present, would duplicate the result of calling sum over the metric directly

### DIFF
--- a/deployments/config/grafana/dashboards/ABCI.json
+++ b/deployments/config/grafana/dashboards/ABCI.json
@@ -167,7 +167,7 @@
             "type": "prometheus",
             "uid": "${DS_PROMETHEUS}"
           },
-          "expr": "sum(rate(tendermint_abci_connection_method_timing_count[5m])) by (method)",
+          "expr": "sum(rate(tendermint_abci_connection_method_timing_seconds_count[5m])) by (method)",
           "refId": "A"
         }
       ],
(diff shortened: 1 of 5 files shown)


conorsch:
Huge, thank you, @voteblake! Along with #3322 these changes are worth considering for inclusion in a point release.

voteblake:
Looking at these together I actually see one other issue, I'm converting to a draft and will push an update today.

conorsch:
Word, holding off for now. Prior to merge, I plan to rebase this and squash it down into a single commit, simply to make a point-release easier. You can do that after making your changes, or I can do it as part of review, either way.

voteblake:
Alright, apologies for the force-push gore, I opened this a bit too soon. I was missing updates to two dashboards that my provisioning script was skipping. With dashboards running from this branch, I am down to only one panel missing data.

The 'Serialized TCT Size' panel on the 'Penumbra' dashboard references a metric `penumbra_storage_tct_size_bytes` that I do not see a current equivalent for. I'm not sure the history of that metric - if it is truly not emitted any more or just not present in my deployment.

hdevalence:
Ah, yep, that one is long gone, we haven't cleaned up the dashboards in a while.

PR #3422, opened 2023-11-21
use `prost::Name` trait instead of `TypeUrl`

### LINKED ISSUE #3330: Use prost's Name trait instead of our TypeUrl trait

We have a TypeUrl trait with manually written Protobuf type URLS in it, and these are not kept up to date and are now wrong. This is a problem because it breaks Any support and also because the type URLs are part of the consensus critical hashing.

Upstream landed support for auto generated names:

https://github.com/tokio-rs/prost/issues/926

Check that the upstream support made it into a release and use it to delete all the existing TypeUrl code and replace with auto generated Name impls

### LINKED ISSUE #3380: Audit effect hash constants
The effect hash depends on protobuf type URLs and these are currently hardcoded and may not (do not) correspond to the actual protos.

Effect hash changes are breaking changes used to sign transactions.

_What we need to implement_

- [ ] (Ideally) use the prost-build support for the `Name` trait to do this automatically.

redshiftzero:
Closes #3330
Closes #3380

@erwanor flagged that we also need https://github.com/cosmos/ibc-proto-rs/pull/163 to completely migrate away from TypeUrl

Remaining TODOs:
- [x] update effect hash docs
- [x] update remaining IBC protos after ibc_types is updated


erwanor:
`ibc-types@0.10` is released and removes the `TypeUrl` bound from `DomainType`, replacing it with `prost::Name`

cronokirby (approved):
One thing to note is that I think this is breaking because we lose the leading slash on the name.

Also, I haven't literally looked at every file, but it seems that the process here was just removing impl TypeUrl everywhere, and then using the prost::Name trait instead, so that seems fine.

hdevalence:
> One thing to note is that I think this is breaking because we lose the leading slash on the name.

Yes, also because a bunch of our URLs were outdated since I didn't fix them while rearranging all the protos.

PR #3427, opened 2023-11-23
fix: circular dependency cleanup

TalDerei:
@conorsch during your cleanup of the commit history, I think one of the dev dependencies in the `penumbra-stake` crate was unnecessarily kept in and snuck into master. To completely resolve the circular dependency bug, that needs to be removed as well.


TalDerei:
@conorsch can we merge?

Issue #3431, opened 2023-11-25
Modularize wasm crate logic 


TalDerei:
Structuring the `build_action` method within the `WasmPlanner` introduces a challenge when calling it from **web-workers**. It means that an instance of the `WasmPlanner` struct must be passed to the web-workers by either:

- **(1)** passing in the individual fields that comprise the `WasmPlanner` struct, and manually reconstructing an instance of the object within the web-worker, 
- **(2)** passing the complex object which internally invokes the [structured-clone algorithm](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm) to perform a deep-copy. 

These methods introduce unnecessary overhead and complexity. Instead, I propose relocating the `build_action` method to a dedicated [build](https://github.com/penumbra-zone/penumbra/blob/wasm-builder/crates/wasm/src/build.rs) file since its design doesn't depend on the `WasmPlanner`. This enables the `build_action` to be called directly from within the web-worker. 

As we expose more of the underlying proving system logic in wasm, the build file can be used to hold that logic.

hdevalence:
I'm not sure I follow — why do these interact at all?

The Planner code produces the TransactionPlan and doesn't need to be parallel at all, it's not expensive to run. Once it finishes, the TransactionPlan has all the information needed to build the transaction, so the build methods should only need access to the TransactionPlan and the WitnessData, as described in the previous PRs. Where does the planner come in?

hdevalence:
Oh, I see, is the point that the wasm code previously attached the building logic to the planner, even though planning and building are supposed to be separate phases?

TalDerei:
> Oh, I see, is the point that the wasm code previously attached the building logic to the planner, even though planning and building are supposed to be separate phases?

Yeah exactly, the building logic shouldn't be under the planner. Separating them is cleaner and simplifies the `build_action` calls on the web-side.

Issue #3400, opened 2023-11-17
`WitnessData` and `tct::Proof` structs serialization 


TalDerei:
[WitnessData](https://github.com/penumbra-zone/penumbra/blob/main/crates/core/transaction/src/witness_data.rs#L8) and [tct::Proof](https://github.com/penumbra-zone/penumbra/blob/main/crates/crypto/tct/src/proof.rs#L8) structs don't implement the `Serialize` trait.

PR #3432, opened 2023-11-25
refactor(wasm): modularize wasm crate logic

TalDerei:
References #3431 and https://github.com/penumbra-zone/penumbra/issues/3400.

### DIFF
--- a/crates/wasm/src/build.rs
+++ b/crates/wasm/src/build.rs
@@ -5,8 +5,10 @@ use wasm_bindgen::JsValue;
 
 use crate::utils;
 use penumbra_keys::FullViewingKey;
-use penumbra_proto::core::transaction::v1alpha1 as pb;
-use penumbra_transaction::{plan::ActionPlan, plan::TransactionPlan, WitnessData};
+use penumbra_transaction::{
+    plan::{ActionPlan, TransactionPlan},
+    WitnessData,
+};
 
 /// Builds a planned [`Action`] specified by
 /// the [`ActionPlan`] in a [`TransactionPlan`].
@@ -28,8 +30,7 @@ pub fn build_action(
     let transaction_plan: TransactionPlan =
         serde_wasm_bindgen::from_value(transaction_plan.clone())?;
 
-    let witness_data_proto: pb::WitnessData = serde_wasm_bindgen::from_value(witness_data)?;
-    let witness_data: WitnessData = witness_data_proto.try_into()?;
+    let witness: WitnessData = serde_wasm_bindgen::from_value(witness_data)?;
 
     let action_plan: ActionPlan = serde_wasm_bindgen::from_value(action_plan)?;
 
@@ -37,8 +38,8 @@ pub fn build_action(
 
     let memo_key = transaction_plan.memo_plan.map(|memo_plan| memo_plan.key);
 
-    let action = ActionPlan::build_unauth(action_plan, &full_viewing_key, &witness_data, memo_key)?;
+    let action = ActionPlan::build_unauth(action_plan, &full_viewing_key, &witness, memo_key)?;
 
-    let action_result_proto = serde_wasm_bindgen::to_value(&Some(action))?;
-    Ok(action_result_proto)
+    let result = serde_wasm_bindgen::to_value(&action)?;
+    Ok(result)
 }
--- a/crates/wasm/src/build.rs
+++ b/crates/wasm/src/build.rs
@@ -1,0 +1,113 @@
+use anyhow;
+use wasm_bindgen::JsValue;
+use wasm_bindgen::prelude::wasm_bindgen;
+use crate::error::WasmResult;
+use std::str::FromStr;
+
+use crate::utils;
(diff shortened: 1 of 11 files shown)


crates/wasm/src/build.rs
```diff
@@ -0,0 +1,107 @@
...
+    let witness_data_proto: pb::WitnessData = serde_wasm_bindgen::from_value(witness_data)?;
+    let witness_data: WitnessData = witness_data_proto.try_into()?;
+
+    let action_plan: ActionPlan = serde_wasm_bindgen::from_value(action_plan)?;
+
+    let full_viewing_key: FullViewingKey = FullViewingKey::from_str(full_viewing_key)?;
+
+    let memo_key = transaction_plan.memo_plan.map(|memo_plan| memo_plan.key);
+
+    let action = match action_plan {
```
hdevalence, line 41: Why do we need to match here? Can't we call the existing ActionPlan::build_unauth method?

TalDerei: good point, noticed that as well. I'll make the change.

TalDerei:
@hdevalence looking at `build_action` method again, I'm realizing it doesn't need to take the [transaction_plan](https://github.com/penumbra-zone/penumbra/blob/wasm-builder/crates/wasm/src/build.rs#L21) field. It's deserializing an entire `TransactionPlan` just to access the `memo_plan` which can simply be passed in from the web-side.

hdevalence:
Yeah, I thought the reason for that was that it would be simpler to only pass one object over FFI into wasm. Even if it's a little inefficient, that's fine if the code becomes simpler, as long as it's not the dominant performance factor.

crates/wasm/src/build.rs
```diff
@@ -0,0 +1,44 @@
...
+    full_viewing_key: &str,
+    witness_data: JsValue,
+) -> WasmResult<JsValue> {
+    utils::set_panic_hook();
+
+    let transaction_plan: TransactionPlan =
+        serde_wasm_bindgen::from_value(transaction_plan.clone())?;
+
+    let witness_data_proto: pb::WitnessData = serde_wasm_bindgen::from_value(witness_data)?;
+    let witness_data: WitnessData = witness_data_proto.try_into()?;
```
grod220, line 32: Can we just annotate `WitnessData` so we get the auto-try-into stuff?

```rust
#[derive(Clone, Debug, Serialize, Deserialize)]
#[serde(try_from = "pb::WitnessData", into = "pb::WitnessData")]
pub struct WitnessData {
```

crates/wasm/src/build.rs
```diff
@@ -0,0 +1,44 @@
...
+
+    let action_plan: ActionPlan = serde_wasm_bindgen::from_value(action_plan)?;
+
+    let full_viewing_key: FullViewingKey = FullViewingKey::from_str(full_viewing_key)?;
+
+    let memo_key = transaction_plan.memo_plan.map(|memo_plan| memo_plan.key);
+
+    let action = ActionPlan::build_unauth(action_plan, &full_viewing_key, &witness_data, memo_key)?;
+
+    let action_result_proto = serde_wasm_bindgen::to_value(&Some(action))?;
```
grod220, line 42: Does it have to be an Option? or can we pass it like:

```rust
serde_wasm_bindgen::to_value(&action)?
```

crates/wasm/src/wasm_planner.rs
```diff
@@ -190,6 +79,7 @@ impl WasmPlanner {
     /// Set gas prices
     /// Arguments:
     ///     gas_prices: `GasPrices`
+    #[wasm_bindgen]
```
grod220, line 82: for wasm_bindgen, it's not necessary adding annotations for struct methods. The rendered class will include them automatically.

grod220 (commented):
For the wasm-compilation script, left comments in the other PR 👍

crates/wasm/src/build.rs
TalDerei: We can, references https://github.com/penumbra-zone/penumbra/issues/3400. 

I modified the witness as `let witness: WitnessData = serde_wasm_bindgen::from_value(witness_data)?`

crates/wasm/src/build.rs
```diff
@@ -0,0 +1,44 @@
...
+    Ok(action_result_proto)
```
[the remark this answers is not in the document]
TalDerei: changed to simply `result`

crates/wasm/src/build.rs
TalDerei: Good catch, I'm not sure why this was marked as an option. The call to `build_unauth` expects a known `ActionPlan` to be specified. I made the change.

crates/wasm/src/wasm_planner.rs
TalDerei: removed

TalDerei:
This PR is blocked until #3429 is merged.

grod220 (approved):

PR #3438, opened 2023-11-27
add getters for `IbcPacket` fields

noot:
self explanatory - i need these fields for astria's `Ics20Withdrawal` implementation.


erwanor (approved):
Thanks!

PR #3449, opened 2023-11-29
Add threshold custody commands to pcli

### LINKED ISSUE #3343: Threshold Custody: Implement broadcast transport over the command line
The idea is that the custody backend assumes a sort of automated environment where we can broadcast messages out into the ether and somehow receive a fixed number of responses back. We need to instantiate this transport protocol with an instance that prints out textual encodings of messages, and then receives back the responses over the command line as well.

### LINKED ISSUE #3344: Threshold Custody: Create a pcli command which acts as a centralized dealer
We should have a pcli command that spits out several pre-filled in config files using a centralized dealer, for threshold custody.

I don't envision this being the main way people use this, instead people should use the decentralized key generation mode.

However, this is very useful for testing, and allows us to focus on the signing aspect first, before then implementing DKG as an orthogonal task.

### LINKED ISSUE #3345: Threshold Custody: Implement threshold signing in pcli
Once we've created a Custody backend that works, we can plug that in to pcli, along with a command to follow along with the coordinator.

cronokirby:
Closes #3343.
Closes #3344.
Closes #3345.

Probably want to merge #3383  first, then switch this PR to be based on main, and review it.

### DIFF
--- a/crates/bin/pcli/src/command/init.rs
+++ b/crates/bin/pcli/src/command/init.rs
@@ -1,6 +1,7 @@
 use std::{io::Read, str::FromStr};
 
 use anyhow::Result;
+use camino::Utf8PathBuf;
 use penumbra_keys::keys::{Bip44Path, SeedPhrase, SpendKey};
 use rand_core::OsRng;
 use url::Url;
@@ -30,6 +31,9 @@ pub enum InitSubCmd {
     /// Initialize `pcli` with a basic, file-based custody backend.
     #[clap(subcommand, display_order = 100)]
     SoftKms(SoftKmsInitCmd),
+    /// Initialize `pcli` with a manbual threshold signing backend.
+    #[clap(subcommand, display_order = 150)]
+    Threshold(ThresholdInitCmd),
     /// Initialize `pcli` in view-only mode, without spending keys.
     #[clap(display_order = 200)]
     ViewOnly {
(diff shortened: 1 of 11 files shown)


crates/bin/pcli/src/command/init.rs
```diff
@@ -30,6 +31,9 @@ pub enum InitSubCmd {
     /// Initialize `pcli` with a basic, file-based custody backend.
     #[clap(subcommand, display_order = 100)]
     SoftKms(SoftKmsInitCmd),
+    /// Initialize `pcli` with a manbual threshold signing backend.
```
redshiftzero, line 34:
```suggestion
    /// Initialize `pcli` with a manual threshold signing backend.
```

crates/bin/pcli/src/command/init.rs
```diff
@@ -106,8 +110,53 @@ impl SoftKmsInitCmd {
...
+            }
+        }
+    }
+}
+
 impl InitCmd {
     pub fn exec(&self, home_dir: impl AsRef<camino::Utf8Path>) -> Result<()> {
+        if let InitSubCmd::Threshold(cmd) = &self.subcmd {
+            cmd.exec(self.grpc_url.clone())?;
+            return Ok(());
```
redshiftzero, line 158: doesn't returning here mean that the `PcliConfig` doesn't get created? I think then later when we try to use `pcli threshold` we'll get an error as it's inspecting the contents of the `PcliConfig` to determine the configured custody backend

crates/bin/pcli/src/command/threshold.rs
```diff
@@ -0,0 +1,26 @@
+use anyhow::Result;
+
+use crate::{terminal::ActualTerminal, App};
+
+#[derive(Debug, clap::Subcommand)]
+pub enum ThresholdCmd {
+    /// Follow along with the threshold signing process
+    Follow,
```
hdevalence, line 8: Should this be `sign`? `pcli threshold sign` seems more intuitive than `pcli threshold follow` to me,  but this is a bit of a bikeshed to be sure.

crates/bin/pcli/src/command/init.rs
```diff
@@ -106,8 +110,53 @@ impl SoftKmsInitCmd {
...
+#[derive(Debug, clap::Subcommand)]
+pub enum ThresholdInitCmd {
+    /// Use a centralized dealer to create config files for each signer.
+    Deal {
+        /// The minimum number of signers required to make a signature (>= 2).
+        #[clap(short, long)]
+        threshold: u16,
+        /// One file name for each signer, where the dealt config will be saved.
+        #[clap(short, long, value_delimiter = ' ', multiple_values = true)]
+        out: Vec<Utf8PathBuf>,
```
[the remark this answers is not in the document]
cronokirby: The difference would be between running:
```
<command> dir1/config.toml dir2/config.toml ...
```
and
```
<command> dir1/ dir2/ ...
```
I guess it's probably good to avoid having to remember what the name of the config file is, yeah

cronokirby:
Is the failure in https://github.com/penumbra-zone/penumbra/actions/runs/7065721967/job/19236295269?pr=3449 spurious?

hdevalence:
I don't know, it doesn't seem obviously triggered by a failing expectation, we can re-run the job and find out.

I'm noticing, however, that the smoke test job is taking alarmingly long to run. It should be finishing within 5-10 minutes, not up to 30.

PR #3459, opened 2023-12-01
app: lift k/v rpcs into storage crate

### LINKED ISSUE #3130: Move key/value RPCs into `storage` crate
As part of #3091, I tried to move the key/value query RPCs into the storage crate. These RPCs are used by relayer implementations to get data from the chain with inclusion proofs. 

At the time, this turned out not to be possible, because of the complexities involved in our `AppHash` implementation, described in #3129. Implementing that approach would allow us to eliminate the custom Penumbra `AppHash`, and move the key/value query RPCs into the storage crate where they belong.

Then they can be used by Astria for their relayer implementation.

erwanor:
Close #3130, this PR:

- Add query service definitions for storage protos
- Remove the KV RPC definitions from the app query service
- Lifts the key value rpc implementations into `penumbra_storage::rpc`

One unfortunate collateral damage that I forgot about is that we can't do the chain id fail-early check in the storage component because that's an implementation detail of the penumbra application. Maybe we should have a service layer that does this sort of generally useful checks, maybe not.


erwanor:
Merging this optimistically, but would appreciate a review async.

Issue #3401, opened 2023-11-17
Update the docs related to building a transaction 


TalDerei:
Make relevant updates to the docs for changes to transaction building, for instance updating `transaction_signing.md` to reflect the removal of the `UnauthTransaction` type.

PR #3463, opened 2023-12-04
Update docs related to transaction build process

TalDerei:
References #3401 

`transaction_signing.md` doc is modified to describe the build process in more detail according to the changes in #3289. Additionally, since the `UnauthTransaction` type was removed, I propose changing the intermediate state of a transaction to `Unauthenticated Transaction` or `Pending Transaction` for improved clarity.


docs/protocol/src/crypto/transaction_signing.md
```diff
@@ -8,20 +8,23 @@ what they are signing.
...
 
-The building process takes the `TransactionPlan`, generates the proofs, and constructs a
-transaction with dummy signatures, that can be filled in once the
+The building process takes the `TransactionPlan`, generates the proofs, and constructs a fully-
+formed `Transaction`. This process is internally partitioned into three steps:
+1. Each `Action` is individually built based to its specification in the `TransactionPlan`. 
+2. The pre-built actions to collectively used to construct a transaction with placeholder dummy signatures, that can be filled in once the
 signatures from the `AuthorizationData` are ready[^1]. This intermediate state
-of the transaction without the full authorizing data is called the `UnauthTransaction`.
+of the transaction without the full authorizing data is referred to as the `UnauthenticatedTransaction`.
```
hdevalence, line 26: Here we should probably say "unauthenticated `Transaction`", right, since there's no distinct data type / message?

TalDerei:
> > Additionally, since the `UnauthTransaction` type was removed, I propose changing the intermediate state of a transaction to `UnauthenticatedTransaction` or `PendingTransaction` for improved clarity.
> 
> One thing that's a little awkward is that now we don't actually have a different state, it's just a transaction (with invalid data in it). Not sure the best way to handle this in the docs.

It might be better to describe it in terms of a state transition, where the `Transaction` internals change across the discrete build phases. I think we kind of already do that with our current description **"This intermediate state of the transaction without the full authorizing data..."**?

hdevalence (approved):

PR #3464, opened 2023-12-04
Threshold Custody: Implement Distributed Key Generation Protocol

### LINKED ISSUE #3346: Threshold Custody: Implement decentralized key generation wrapper protocol
We should create a protocol for establishing a key with FROST, but also any auxiliary metadata we need as part of the protocol.

### LINKED ISSUE #3347: Threshold Custody: Implement pcli initialization with decentralized key generation
Using the wrapped FROST protocol, we should have a way to initialize pcli without a centralized dealer, for the threshold custody setting.

cronokirby:
Closes #3346.
Closes #3347.


cronokirby:
One thing I think I forgot was to add a check in round2 that all the VKs are distinct and not equal to your own.

crates/custody/src/threshold.rs
```diff
@@ -108,6 +109,65 @@ pub async fn follow(config: &Config, terminal: &impl Terminal) -> Result<()> {
...
+    terminal.broadcast(&to_json(&round2_message)?).await?;
+    // Round 2 bottom
+    terminal
+        .explain(&format!(
+            "Round 2/2: Gather {expected_responses} messages from the other participants:"
+        ))
+        .await?;
+    let round2_replies = {
+        let mut acc: Vec<dkg::Round2> = Vec::new();
+        while acc.len() < expected_responses {
```
redshiftzero, line 159: Here we're checking that we're getting the right number of responses, should we also be checking at each phase that we have no duplicate vks and that our own vk is not provided in a response?

cronokirby: Yeah, we should do this internally, and also in signing

Issue #3456, opened 2023-11-30
staking: update `UndelegateClaimCircuit` to use `penumbra-num` amount and fixpoint arithmetic


redshiftzero:
@erwanor is taking a first pass at updating the penalty calculation in the staking component to use the `U128x128` fixpoint type and amounts throughout (ref #3453).

Next the `UndelegateClaimCircuit` will also need to be updated since it uses the penalty calculation (as in `Penalty::apply_to`). If the penalty representation changes from a u64 to a `U128x128` or amount, then the R1CS `PenaltyVar` representation will need to be updated accordingly. We then will need to update the constraints that assert that the `penalized_amount` - currently represented as an `AmountVar` and constrained to be 128 bits - has been computed correctly based on the penalty and amount being unbonded. The other parts of the circuit will be unchanged. 

Then we will determine if the circuit cost exceeds the size of the phase 1 output for the `UndelegateClaimCircuit`. If it exceeds the size of the phase 1 output which was the existing circuit size rounded up to the nearest power of 2 (we can add 1608 constraints), then we will either optimize the circuit cost to fit or specialize a phase 1 output of another circuit to this end (h/t to @cronokirby for this suggestion).

redshiftzero:
https://github.com/penumbra-zone/penumbra/pull/3466/commits/ab0fbe5cec3ed405320b7e444e4e0328a1c7d40f has a commit with the updated `UndelegateClaim` proof. The size of the `UndelegateClaim` circuit actually gets slightly smaller - from 14,776 to 14,423 constraints. My intuition on why this is so is that previously we were using the expensive `AmountVar::quo_rem` function in order to add constraints that the (witnessed from out of circuit computation) penalized amount had been calculated correctly, which is no longer necessary as we use the `U128x128Var::checked_mul` method to derive the penalized amount in-circuit directly.

PR #3466, opened 2023-12-04
staking: make `Penalty` a `U128x128` (discovery)

erwanor:
This PR is an experimental branch part of #3456, it serves as a way to discover requirements for a staking penalty rewrite. Once this exploratory phase is over and we have a clear idea of the magnitude of changes that are it causes, we will be able to fork the staking component. When we are done, we will switch the legacy staking component with the one containing the updated undelegate claim circuit. Another goal of this effort is to identify areas of the staking components that could have testing coverage _today_ (i.e. that don't require an end-to-end cometbft mocking story).


crates/core/component/stake/src/params.rs
```diff
@@ -31,8 +32,8 @@ impl TryFrom<pb::StakeParameters> for StakeParameters {
         Ok(StakeParameters {
             unbonding_epochs: msg.unbonding_epochs,
             active_validator_limit: msg.active_validator_limit,
-            slashing_penalty_downtime: msg.slashing_penalty_downtime,
-            slashing_penalty_misbehavior: msg.slashing_penalty_misbehavior,
+            slashing_penalty_downtime: todo!("convert proto defs"),
+            slashing_penalty_misbehavior: todo!("convert proto defs"),
```
hdevalence, line 36: do we want to change the parameters to encode a `U128x128`, or should we keep them as some kind of decimal fraction like they are now?

crates/core/component/stake/src/penalty.rs
```diff
@@ -12,52 +12,58 @@ use penumbra_asset::{
...
-        let one = 1_0000_0000u128;
-        let p1 = self.0 as u128;
-        let p2 = other.0 as u128;
-        let q = u64::try_from(one - (((one - p1) * (one - p2)) / 1_0000_0000))
-            .expect("value should fit in 64 bits");
+        /* Compounding:
+         * For two penalty rates p1 and p2, the compounded penalty rate q is:
+         *     `1 - q = (1-p1)(1-p2)``
+         *     `q = 1 - (1-p1)(1-p2)`
+         */
```
hdevalence, line 48: do we want to do this, if we're using a `U128x128`? what if we just multiplied the penalties together?

 the previous representation did the (1-q) thing i think in order to make representing the fixed-point values easier, but if we are using a `U128x128` why not just make the penalty be multiplicative? wouldn't that have the least loss of precision?

cronokirby: I think doing 1 - q vs q has no impact on precision at all in a fixed point representation

hdevalence: The three subtractions become unnecessary, though, and it's simpler to reason about.

cronokirby: Yeah, ultimately the 1 - is a matter of what the public interface for the component is, I agree that having rates be [0, 1) internally directly is simpler

hdevalence: All the circuit needs to be doing is multiplying a number between 0 and 1 against the amount of tokens. We need to do that in the simplest way possible so we can reduce the risk of bugs.

crates/core/component/stake/src/penalty.rs
```diff
@@ -12,52 +12,58 @@ use penumbra_asset::{
...
     pub fn apply_to(&self, amount: Amount) -> Amount {
-        let penalized_amount = (u128::try_from(amount).expect("amount should be a valid u128"))
-            * (1_0000_0000 - self.0 as u128)
-            / 1_0000_0000;
-        Amount::try_from(
-            u64::try_from(penalized_amount).expect("penalized amount should fit in u64"),
-        )
-        .expect("all u64 values should be valid Amounts")
+        let amount = U128x128::from(amount);
+        let penalized_amount = (amount * self.0).expect("cannot overflow");
```
hdevalence, line 62: This contradicts the computations above, which is probably why the tests are failing.

Issue #3435, opened 2023-11-26
Introduce a canonical ordering to the `TransactionPlan`


TalDerei:
`TransactionPlan::sort_actions(&mut self)` will be implemented which will be called by the `Planner` before returning the final `TransactionPlan`. This will standardize the actions in a transaction with a set ordering.

hdevalence:
We should implement a `TransactionPlan::sort_actions`. 

This should sort the actions by type, using the protobuf field number in the `Action` oneof as the comparator key. The field numbers can safely be hardcoded because they're never allowed to change in the protobufs. We don't need to do any sorting beyond the action type.

cronokirby:
Might be good to wait for #4081 for this.

TalDerei:
Unblocked by https://github.com/penumbra-zone/penumbra/pull/4312

Issue #3379, opened 2023-11-15
Audit transaction arity implications for client fingerprinting


aubrika:
Penumbra transactions reveal the arity of their actions: how many spends, outputs, etc. Prior to committing to the transaction format, we should audit it to see whether there are any opportunities to close information leaks. For instance, should we place restrictions on how actions are ordered? Or should that be a client-side concern?

_Why this feature is required pre-mainnet_

We need to do this prior to stabilizing the transaction format

_What we need to implement_

Unclear; this would primarily be a design review with presumably minor changes to the transaction format or transaction building.

TalDerei:
Tangentially related to #3435 and tracked by #3467

PR #3467, opened 2023-12-04
refactor(transaction): Implement canonical ordering in `TransactionPlan`

TalDerei:
References #3435 and https://github.com/penumbra-zone/penumbra/issues/3379


cronokirby:
We should also have tie breaking based on the effect hash, rather than just using whatever order between actions that happens to exist (or perhaps something more intricate). Because of this, it might be a good idea to instead shift the approach of this PR to adding an Eq and Ord instance to Action based on this philosophy.

TalDerei:
> effect hash

Why is tie breaking against the effect hash necessary for the same action type?

TalDerei:
This PR will remain as a draft until after the next release.

conorsch:
We are not planning to merge this for Testnet 64, so we'll return to it next week.

TalDerei:
Is this something we want done for the next testnet release? otherwise it can remain in the backlog

hdevalence:
Let's keep it in the backlog for now

TalDerei:
reopened since #3435 was tagged with `V1`

TalDerei:
blocked until https://github.com/penumbra-zone/penumbra/issues/4081

TalDerei:
unblocked by https://github.com/penumbra-zone/penumbra/pull/4312

crates/core/transaction/src/plan.rs
```diff
@@ -48,6 +49,20 @@ pub struct TransactionPlan {
...
+
+        actions
+    }
+
+    /// Sort the action plans by type, using the protobuf field number in the [`ActionPlan`].
+    pub fn sort_action_plans(&self, mut actions_plans: Vec<ActionPlan>) -> Vec<ActionPlan> {
+        actions_plans.sort_by_key(|action: &ActionPlan| action.variant_index());
+
+        actions_plans
+    }
```
hdevalence, line 64: I think we probably just want to have a single method
```
TransactionPlan::sort(&mut self) { ... }
```
right? Then we can sort the plan exactly once, immediately before the planner returns it, and then we won't have to change any of the other code.

TalDerei: yeah that's much cleaner

hdevalence (approved):
LGTM!

PR #3473, opened 2023-12-05
Fetch and store transactions from `pd` storage

### LINKED ISSUE #3451: pd: index `cometbft` block data
In Penumbra, chain upgrades are performed via "genesis migrations". The chain state is exported, and node operators produce an upgraded genesis file that specifies the new application hash and starting height for the post-upgrade chain.

Unfortunately, this process requires `cometbft` to delete all of the block data preceding the upgrade height. This is problematic because Penumbra clients require access to historical block data in order to synchronize their view service, or query transactions.

In particular, Penumbra clients need to:
- query transactions by block height ("fetch all txs that are present in block $h$")
- query transactions by tx hash ("fetch the transaction data corresponding to `tx_hash`")

To solve this, we need to index block data directly into the application, and make the `TendermintProxy` service route queries to this new local block storage.

In order to do this, we should:

1. Figure out a storage strategy and schema for the full block data

The general schema of our indices should allow for retrieval by block height (frequent) and retrieval by tx hash (sparser):
- `block_height` -> `Vec<block_data>`
- `tx_hash` -> (block_height, position_in_block)

In pseudo-code:

```rust=
fn get_tx(tx_hash) -> Tx {
    let (height, position) = lookup_index(tx_hash);
    let tx = get_block(height)[position];
    tx
}
```

2. Pull transactions handled in `ProcessProposal` and store them in `pd` block db
3. Modify the `TendermintProxy` service  to route `get_tx` and `get_block` RPCs to the local block store

This is part of #1804, and blocking for #3408

zbuc:
This removes `pcli`'s reliance on the `get_block_by_height` method since it can't persist data after a chain upgrade occurs. Since it was only using this to fetch transaction data during sync, the transaction data is all that we need to record.

All block transactions will be recorded in nonverifiable storage in a new `cometbft-data` substore.

Closes #3451


proto/penumbra/penumbra/core/app/v1alpha1/app.proto
```diff
@@ -9,13 +9,33 @@ import "penumbra/core/component/governance/v1alpha1/governance.proto";
...
+  // Returns the CometBFT transactions that occurred during a given block.
+  rpc TransactionsByHeight(TransactionsByHeightRequest) returns (stream TransactionsByHeightResponse);
 }
 
+// Requests the list of all transactions that occurred within a given block.
+message TransactionsByHeightRequest {
+  // The expected chain id (empty string if no expectation).
+  string chain_id = 1;
+  // The block heights to retrieve.
+  repeated uint64 block_heights = 2;
```
hdevalence, line 27: What's the motivation for making this be streaming rather than unary? It seems like it might be simpler if it was just for a single height.

zbuc: There are two separate things going on:

1. the request type supports asking for multiple block heights at the same time; I thought this could be useful though only a single block height is supplied by pcli in this PR
2. the responses are streaming to reduce response gRPC message size

As for 1. since it's not being used it could be removed, but most of our existing APIs are streaming so 2. matches the existing pattern in our APIs

proto/penumbra/penumbra/core/transaction/v1alpha1/transaction.proto
```diff
@@ -201,10 +200,15 @@ message ActionView {
   }
 }
 
+// An authorization hash for a Penumbra transaction.
+message EffectHash {
+  bytes inner = 1;
+}
```
hdevalence, line 206: Side note: I think we might want to split the effect hash code into its own crate, rather than putting it in the transaction crate. The reason is that all of the components need to be able to define their effecting data, but the overarching `transaction` crate depends on all the components.

hdevalence (approved):
This LGTM, thanks for getting it done.

PR #3474, opened 2023-12-05
Start pruning content out of the `penumbra-chain` crate

hdevalence:
This is a catch-all crate at the bottom of the component dependency tree, and it'd be clearer to get rid of it and move the code into places where it belongs.


hdevalence:
This is a reasonable start and doesn't depend on further changes.

crates/core/component/shielded-pool/src/genesis/allocation.rs
```diff
@@ -10,16 +14,38 @@ use serde::{Deserialize, Serialize};
     into = "pb::genesis_content::Allocation"
 )]
 pub struct Allocation {
-    pub amount: Amount,
-    pub denom: String,
+    pub raw_amount: Amount,
+    pub raw_denom: String,
```
hdevalence, line 18: I changed these to `raw_` to reflect that they might be in units that need adjustment. This is not a great pattern -- we should avoid using `Amount` for anything other than an amount of the base denom -- but it's probably okay here because it's only for the genesis data, which has to be trusted anyways.

crates/core/component/stake/src/component.rs
```diff
@@ -918,12 +919,15 @@ impl<T: StateWrite + StateWriteExt + ?Sized> StakingImpl for T {}
 
 #[async_trait]
 impl Component for Staking {
-    type AppState = GenesisContent;
+    type AppState = (GenesisContent, ShieldedPoolGenesisContent);
```
hdevalence, line 922: In this way, components can reference other components' genesis data.

PR #3478, opened 2023-12-06
Implement a minimal fix for Penalty calculation in UndelegateClaimCircuit

### LINKED ISSUE #3456: staking: update `UndelegateClaimCircuit` to use `penumbra-num` amount and fixpoint arithmetic

cronokirby:
Closes #3456.

I lost track of what exactly had changed in https://github.com/penumbra-zone/penumbra/pull/3466, so I instead just implemented this without basing myself off of that PR.

The difference in approach is that this PR makes the `Penalty` type completely opaque, and modifies no other typed values in the system to use Penalty or U128x128 except the internal representation of the Penalty type.

I think making Penalty opaque is the correct approach, because it makes it impossible for there to ever be any confusion about what the meaning of the u64s you're passing in to Penalty are.

Another difference with the other PR is that the circuit behavior should match the out of circuit behavior exactly now, whereas the other PR is missing a 1 - X operation in the circuit. This bumps up the constraint count to 15714. (We can golf this down with at least two strategies, but we don't really need to, it still fits).

One *important* thing that should be considered is the fact that this PR modifies the proto-encoding of penalty to now hold the full precision of U128x128 by encoding the 32 bytes that value consists of. I'm not quite sure of the ramifications of this change but it should be considered.

Update (2023-12-06T08:52:27+00:00): I've switched the Penalty to store the a fraction from [0, 1], representing the *kept* amount, rather than the removed amount. This simplifies compounding and the circuit, and also fixes the weird test failures I tried to debug for a couple hours.


redshiftzero:
Per discussion in today's sprint planning meeting, I'll add a commit later today that adds a test of the `U128x128Var::round_down_to_amount` functionality

crates/core/component/stake/src/penalty.rs
```diff
@@ -99,80 +132,17 @@ impl AllocVar<Penalty, Fq> for PenaltyVar {
...
-
-impl From<&PenaltyVar> for AmountVar {
-    fn from(penalty_var: &PenaltyVar) -> Self {
-        // `AmountVar`s must fit in 128 bits, but `PenaltyVar`s have already
-        // been constrained to fit in 64 bits, so we can safely
-        // construct an `AmountVar` from a `PenaltyVar`.
-        AmountVar {
-            amount: penalty_var.inner.clone(),
-        }
+        Ok(Self {
```
redshiftzero, line 135: Here in `AllocVar` as an additional guard we could return an error if `mode` is not `AllocationMode::Input`

redshiftzero (approved):

crates/core/component/stake/src/penalty.rs
```diff
@@ -12,52 +10,66 @@ use penumbra_asset::{
...
 };
-use penumbra_num::{fixpoint::bit_constrain, Amount, AmountVar};
+use penumbra_num::{
+    fixpoint::{U128x128, U128x128Var},
+    Amount, AmountVar,
+};
 
 /// Tracks slashing penalties applied to a validator in some epoch.
 ///
 /// The penalty is represented as a fixed-point integer in bps^2 (denominator 10^8).
```
erwanor, line 20: We should replace this description

erwanor: We can remove all mention of how it is represented internally since the type is opaque

crates/core/component/stake/src/penalty.rs
```diff
@@ -12,52 +10,66 @@ use penumbra_asset::{
...
-        )
-        .expect("all u64 values should be valid Amounts")
+    pub fn apply_to_amount(&self, amount: Amount) -> Amount {
+        self.apply_to(amount)
+            .round_down()
+            .try_into()
+            .expect("converting integral U128xU128 into Amount will succeed")
+    }
+
+    /// Apply this `Penalty` to an some fracton.
```
erwanor, line 70: nit: typos

hdevalence (approved):
This LGTM. Let's land this prior to the Testnet 64 release, and do another round of circuit assurance during Testnet 64.

PR #3479, opened 2023-12-06
feat(pd): custom voting periods testnet generate

conorsch:
During testing of upgrades, it's been handy to reduce the default, static value of "24h" for voting on a given proposal, so that a proposal can be made, voted on, and passed, all within e.g. 1h. This setting greatly accelerates testing on devnets.

Also includes customization var in deploy scripts for `epoch_duration`, which we're not currently overriding, but now we can if we want to.

Ideally we'd set these customizations by just overriding the `pd testnet generate` command, but given that the cmd requires a storage dir arg that's non-standard, it's more straightforward to stick with the clunky vars-based approach for now.

### DIFF
--- a/crates/bin/pd/src/main.rs
+++ b/crates/bin/pd/src/main.rs
@@ -183,6 +183,10 @@ enum TestnetCommand {
         /// Testnet name [default: latest testnet].
         #[clap(long)]
         chain_id: Option<String>,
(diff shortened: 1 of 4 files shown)


conorsch:
The proto lint failure is unrelated to the changes in this PR, see #3487 for attempted resolution on that. The rust test failure is legitimate, though, so I'll fix that and push up again.

PR #3487, opened 2023-12-06
fix(proto): remove asset import in chain protos

conorsch:
Follow up to https://github.com/penumbra-zone/penumbra/pull/3474, to satisfy buf linting CI check.


conorsch:
Weirdly I've removed the unnecessary import but CI is still flagging it:

ci-buf-lint-what

I've confirmed that `buf lint` doesn't report this error on this branch locally, while it does still flag it on `main`, and that's good enough for me, so I'm going to merge this once the smoke test passes (which is a useful sanity check here) and rebase other PRs.

PR #3488, opened 2023-12-06
ci: fix gke auth actions

conorsch:
The prior versions of our GHA helpers were throwing an error on deploy:

    The v0 series of google-github-actions/auth is no longer maintained.
    It will not receive updates, improvements, or security patches.
    Please upgrade to the latest supported versions

Patching both the /auth and /get-gke-credentials GHA helpers.


conorsch:
This was sufficient on an ad-hoc deploy https://github.com/penumbra-zone/penumbra/actions/runs/7120716781, modulo osiris, which is a separate problem.

PR #3491, opened 2023-12-07
pd(upgrade): create fresh validator state and accept `--genesis-start`

erwanor:
This PR makes the `SimpleUpgrade` migration script generate a fresh validator state file to populate comet's view of the validator data. It also adds an optional `--genesis-start` flag to `pd upgrade` to help testing with multinode setups.

### DIFF
--- a/crates/bin/pd/src/upgrade.rs
+++ b/crates/bin/pd/src/upgrade.rs
@@ -14,7 +14,7 @@ use crate::testnet::generate::TestnetConfig;
 pub enum Upgrade {
     /// No-op migration
     Noop,
-    /// Testnet 60 migration
+    /// A simple migration: adds a key to the consensus state.
     Testnet60,
 }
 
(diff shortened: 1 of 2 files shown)


crates/bin/pd/src/upgrade.rs
```diff
@@ -14,74 +14,95 @@ use crate::testnet::generate::TestnetConfig;
...
+                let genesis_json = serde_json::to_string(&genesis).expect("can serialize genesis");
+                tracing::info!("genesis: {}", genesis_json);
+                let mut genesis_path = path_to_export.clone();
+                genesis_path.push("genesis.json");
+                std::fs::write(genesis_path, genesis_json).expect("can write genesis");
+
+                let mut validator_state_path = path_to_export.clone();
+                validator_state_path.push("priv_validator_state.json");
+                let fresh_validator_state =
+                    crate::testnet::generate::TestnetValidator::initial_state();
```
conorsch, line 100: Beautiful, this'll solve our problems! As an aside, is there any kind of chain upgrade where we _wouldn't_ want the validator state reset? I can't think of one.

Issue #3460, opened 2023-12-02
Clean up Penumbra-defined ABCI Events to use Protobufs


hdevalence:
**Is your feature request related to a problem? Please describe.**

Currently, Penumbra's events are defined in an ad-hoc way: each component has an `event.rs` file with Rust functions returning ABCI events. These are constructed on-demand and emitted with the state's `.record()` method, then never used again.

There is no schema, no format definition, and all of the fields are encoded in different ways: sometimes hex, sometimes bech32, sometimes strings, etc. This makes it practically very difficult for any external tooling to consume, and the only reason we haven't noticed is that we haven't yet made a serious attempt at trying.

The SDK had a similar problem, and belatedly migrated to "typed events" defined using Protobufs. We should adopt a similar system, and migrate all of the Penumbra-defined events to be Protobuf-defined types, before external parties start trying to build on top of them

**Describe the solution you'd like**

- [ ] Define a system where events are defined by Protobuf messages. 

  The way the SDK does this can be [seen here](https://github.com/cosmos/cosmos-sdk/blob/main/types/events.go#L89-L156).  For our purposes, it doesn't seem necessary to implement a full domain-type/proto-type split like other proto messages. We don't have any Rust code that consumes the events, so we don't need to parse them, only generate them -- but we should ensure that tooling in _other_ languages can easily parse the event back to a Proto-generated type.  

  One possibility would be to define a trait in the proto crate, something like
  ```rust
  pub trait ProtoEvent: Message + Name {
      fn into_event(&self) -> abci::Event {
          // Use the `Name` trait to extract the proto type name
          // Serialize to JSON and munge strings as in Go code
      }
      fn from_event(event: &abci::Event) -> Result<Self> {
          // Use the `Name` trait to check the event type
          // Assemble k/v pairs into JSON, deserialize to proto
      }
  }
  ```
  
  Then, change the `StateWriteProto` trait in the `penumbra-proto` crate to add a method that does proto-event recording:
  ```rust
  pub trait StateWriteProto: StateWrite + Send + Sync {
      // ... (existing methods) ...
      
      /// Records a protobuf message as a typed ABCI event.
      fn record_proto<E: Message + Name>(&mut self, proto_event: E) {
          self.record(proto_event.into_event());
      }
  }
  ```
  
[...]

PR #3492, opened 2023-12-07
Add Protobuf typed/defined Penumbra ABCI Events 

ejmg:
PR for #3460.


crates/proto/src/event.rs
```diff
@@ -0,0 +1,73 @@
...
+pub trait ProtoEvent: Message + Name + Sized {
+
+    fn into_event(&self) -> abci::Event {
+
+        let kind  = Self::NAME;
+
+        // NOTE: Clarify if the serde derivation is pbjson/prost aware. Or, more specifically, if Penumbra generated serialization code for protobuf defs are more coherent than the 
+        //       3-4 different ways the cosmo-sdk chooses to serializes and deserialize its protobuf data in the reference code.
+        // TODO: Would it be preferred to return an empty(?) or error coded abci::Event?
+        let event_json = serde_json::to_vec(&self.encode_to_vec()).expect("ProtoEvent constrained values should be JSON serializeable.");
```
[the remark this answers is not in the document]
ejmg: this is clarifying. at some point I realized that I didn't know if there was a cost/benefit to relying on those traits being transitively resolved because of `Message`.

crates/proto/src/event.rs
```diff
@@ -44,18 +39,18 @@ pub trait ProtoEvent: Message + Name + Sized {
...
-                serde_json::to_vec(&attr.value).expect("EventAttribute values should be serializeable.")
-            );
-        }
+        let json = serde_json::to_value(attributes).expect("HashMap of String, serde_json::Value should be serializeable.");
 
-        // NOTE: Same thing here as into_event(), check what the equivalent to the jsonpb Unmarshaler is.
-        let json_bytes = serde_json::to_vec(&attributes_kv).expect("Hashmap of EventAttribute Keys and values should be serializeable.");
-        // let decoded_event = Self::decode(serde_json::from_slice(&json_bytes).expect("Serde derived serialized bytes should be deserializeable."));
+        // Hitting a lifetime error that requires further restriction on Self: Deserialize that conflicts with the current supertraits defined above.
+        // return Ok(serde_json::from_value(json)?)
```
ejmg, line 53: Ok, @hdevalence, all I'm left is figuring out how to get the final bit of equivalent unmarshalling/marshalling code in `from_event` method so I can do the actual test data/string munging of results. These lines are the rough equivalent to the final lines of the [go impl here](https://github.com/cosmos/cosmos-sdk/blob/8fb62054c59e580c0ae0c898751f8dc46044499a/types/events.go#L145-L155).

With the final bit twiddling, do I just want `serde_json` to give me back a value of `Result<Self>` like I have commented out above? If so, I'm hitting a Trait restriction compile error directly on `Self` that required an additional lifetime that had me re-writing the Trait's signature to the point I started leaking out the parameterized type variable into the `StateWriterTrait` trait.

I tried massaging things around but I want to make sure I'm not going on another snipe hunt again.

hdevalence: What about using https://docs.rs/serde/latest/serde/de/trait.DeserializeOwned.html ?

hdevalence:
This looks great! I added a few commits to carry it over the finish line. In particular:

- using `DeserializeOwned` avoids the lifetime bounds
- using the fully-qualified proto name will avoid potential conflicts between events in different modules.

I had to make some changes to the (de)serialization to make it work with the `EventOutput`, as an example of something with nested substructures.

In the original issue there was a suggestion to replace all of the existing events with proto-events. I don't think we should do that before merging this, it could be a separate issue. The events are not consensus-critical, so we can change them later (though it's slightly annoying to have different event formats in the same database or otherwise to re-execute all the blocks, etc)

PR #3497, opened 2023-12-08
Port summonerd 63 changes to main

### LINKED ISSUE #3446: Backport summonerd fixes for Testnet 64
During Testnet 63 we [launched phase 1 of the summoning ceremony](https://penumbra.zone/blog/summoning-ceremony-launch). Since then, we've made several hotfixes to summonerd, storing them on a feature branch, [`summonerd-feature-branch`](https://github.com/penumbra-zone/penumbra/tree/summonerd-feature-branch). Prior to releasing 64 (#3408), we must port those commits to main, e.g. via cherry-picking and PRing, so we can transition the ceremony to phase 2.

conorsch:
Closes #3446. This PR rebases `summonerd-feature-branch` on top of current `main`. A single commit was excluded due to conflicts: that from https://github.com/penumbra-zone/penumbra/pull/3291. We definitely want a plan for handling the split-brain, so we can either 1) separately apply that commit, resolving conflicts manually; or 2) configure summonerd to use a single, non-LB grpc endpoint on 64. In the interest of time, I suggest option 2.

I'm also open to squashing this down into a single commit, but up to you, @cronokirby.

### DIFF
--- a/crates/view/src/worker.rs
+++ b/crates/view/src/worker.rs
@@ -1,6 +1,7 @@
 use std::{
     collections::BTreeSet,
     sync::{Arc, Mutex},
+    time::Duration,
 };
 
 use anyhow::Context;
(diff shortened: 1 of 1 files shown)


cronokirby (approved):
This looks good to me. I agree with option 2 of using a non-load balanced endpoint being a good idea.

Issue #3475, opened 2023-12-05
Wasm bindgen tests in ci/cd


grod220:
We currently have one wasm_bindgen_test: https://github.com/penumbra-zone/penumbra/blob/main/crates/wasm/tests/test_build.rs#L38C7-L39 that extensively tests the wasm-crate's build methods.

We additionally need:
1. Documentation. There appears to be a specific command needed to run this test: https://github.com/penumbra-zone/penumbra/pull/3407. We should add some background info + instructions on running this test in the wasm-crate's readme.
2. Add test to CI/CD. Upon pull requests, we should be running this test as well. This will need the addition of a new github action workflow file.

TalDerei:
Additionally, implement assertion checks instead of manually comparing the equality of the transaction outputs as part of the CI/CD integration.

PR #3503, opened 2023-12-09
chore(wasm): add CI workflow 

TalDerei:
References #3475

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -85,22 +85,22 @@ jobs:
       - name: Load rust cache
         uses: astriaorg/buildjet-rust-cache@v2.5.1
 
-      - name: build wasm32 target
-        run: cargo build --release --target wasm32-unknown-unknown
-        # TODO: consider failing on warnings here. A few dead-code
-        # warnings, leaving as-is because wasm is undergoing rapid dev
-        # in support of web extension.
-        working-directory: crates/wasm
-
       # Download prebuilt binary for wasm-pack; faster than `cargo install`.
       - uses: jetli/wasm-pack-action@v0.4.0
         with:
           version: 'latest'
 
-      # `wasm-pack build` will automatically compile to target wasm32-unknown-unknown,
-      # but we run it as a separate task to isolate potential failures in the build pipeline.
+      # Ensure that wasm crate builds cleanly. Is this superfluous, given `wasm-pack test` below?
       - name: run wasm-pack build
         run: wasm-pack build
+        working-directory: crates/wasm
+
+      # Download firefox
+      - uses: browser-actions/setup-firefox@v1
+      - run: firefox --version
+
+      - name: Run headless browser tests
+        run: wasm-pack test --headless --firefox -- --test test_build --target wasm32-unknown-unknown --release --features "mock-database"
         working-directory: crates/wasm
 
       # Always show the results of `cargo tree`, even if the tests failed, since that
--- a/crates/wasm/README.md
+++ b/crates/wasm/README.md
@@ -14,3 +14,15 @@ wasm-pack build
 ```
 
 Artifacts can be found in `pkg/`.
+
+## Testing
+The WASM crate incorporates a unit testing suite that leverages `wasm-bindgen-test` to simulate a sample spend transaction.
+For testing purposes, the suite mocks `indexDB` database calls in an interactive browser accessible at `http://127.0.0.1:8000`.
+
+```
+wasm-pack test --chrome -- --test test_build --target wasm32-unknown-unknown --release --features "mock-database"
+```
+
+The transaction outputs are accessible in the browser's developer console.
+
+The wasm browser tests are also run headlessly in Penumbra CI. Review the relevant GHA workflow file for specifics.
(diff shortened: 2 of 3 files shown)


TalDerei:
Attempting to execute the workflow in `headless` mode seems to fail. Alternatively, the interactive mode (without the `headless` flag) stalls indefinitely since the CI environment can't interact with it. We may need to rethink our approach to testing the WASM crate.

conorsch:
Nice start! Looks like the new test isn't passing yet, so I'll check back when it's green. N.B. we already have a "wasm" job inside the rust workflow—perhaps this new test should be added around here: https://github.com/penumbra-zone/penumbra/blob/b56182336aff03e5d68dd4cfccc60f886b11d0ee/.github/workflows/rust.yml#L105 (I realize the parent ticket mentioned a new workflow, but I'm being persnickety. <3) Currently, the only wasm-related CI we run on this repo is "make sure the wasm builds OK." Is that check still useful if we're actually running tests? That is, doesn't running the tests involve building before running? 

Holler if you want more eyes on this.

conorsch:
> Tagging @conorsch if you have any suggestions?

Ha, I hadn't seen your comment when I wrote my message. Thanks for pinging. 

> Attempting to execute the workflow in headless mode seems to fail. Alternatively, the interactive mode (without the headless flag) compiles successfully, but stalls indefinitely since the CI environment can't interact with it. 

Can you teach me how to run this test locally? I see the new command added to both the wasm crate's README and the CI job: `wasm-pack test --chrome -- --test test_build --target wasm32-unknown-unknown --release --features "mock-database"` but when I run that locally, I get stuck on:

```
Set timeout to 20 seconds...
Interactive browsers tests are now available at http://127.0.0.1:8000

Note that interactive mode is enabled because `NO_HEADLESS`
is specified in the environment of this process. Once you're
done with testing you'll need to kill this server with
```

which is odd, because I definitely don't have `NO_HEADLESS` exported in my environment. Are you able to run these tests locally? If so, and you can teach me to do the same, I'm sure we can make CI happy.

TalDerei:
You're running it properly! Since it's an interactive browser, you need to open `http://127.0.0.1:8000` to start executing the test. Alternatively running it in `headless` mode locally fails.

conorsch:
> ... fails locally.

Yup, same here, that command fails. Likely we'll have to prepare a `webdriver.json` config or something. But my recommendation is to forget about the CI workflows until we have a testing command that executes reliably locally. For instance, this command _does_ pass for me locally:

```
❯ wasm-pack test --headless --firefox -- --test test_build --target wasm32-unknown-unknown --release --features "mock-database"
[...snip...]
Set timeout to 20 seconds...
Running headless tests in Firefox on `http://127.0.0.1:42199/`
Try find `webdriver.json` for configure browser's capabilities:
Not found
running 1 test                                    

test test_build::tests::mock_build_serial_and_parallel ... ok

test result: ok. 1 passed; 0 failed; 0 ignored
```

That's a start, at least!

TalDerei:
This was backlogged in favor of more higher priority web related work. @conorsch Strangely `wasm-pack test --headless --firefox -- --test test_build --target wasm32-unknown-unknown --release --features "mock-database"` passes locally in the `wasm-ci-cd` branch, but fails in the main branch. What would the reason for that be since this branch doesn't have any changes that should affect that?

TalDerei:
tagging @grod220

conorsch:
> Strangely wasm-pack test --headless --firefox -- --test test_build --target wasm32-unknown-unknown --release --features "mock-database" passes locally in the wasm-ci-cd branch, but fails in the main branch. What would the reason for that be since this branch doesn't have any changes that should affect that?

That's not clear to me, either, but FWIW I can definitely reproduce the behavior you describe. There _are_ changes to the wasm tests in this diff, but even copying `crates/wasm/tests/test_build.rs` on `main` and rerunning the tests still fails.

grod220:
Sadly, won't be able to dive into this before my holiday given other high-priority rpc changes before we freeze those. Think we can de-prioritize this and revisit later.

conorsch:
> fails in the main branch

As of merge of #3586, I'm able to run `wasm-pack test --headless --firefox -- --test test_build --target wasm32-unknown-unknown --release --features "mock-database"` on main and it passes! I'm going to rebase this branch on top of latest main and poke at the new test.

TalDerei:
@conorsch might need [webdriver.json](https://rustwasm.github.io/wasm-bindgen/wasm-bindgen-test/browsers.html) config to satisfy CI?

conorsch:
I've made sure I can run the wasm-pack headless test command on both my local workstation, and also on a truly headless setup (rented cloud box), and no problems in either place. So there's something different about the GHA context that we're overlooking. I'll dig into this more.

conorsch:
Finally sorted out the weirdness between local env and CI env: the WIP CI job didn't have lfs=true, so the proving keys being included in the tests weren't real. The error was occurring on https://github.com/penumbra-zone/penumbra/blob/9a5bb7fc18a03137938d5be3def146bf55033110/crates/wasm/tests/test_build.rs#L57 which gets data from https://github.com/penumbra-zone/penumbra/blob/9a5bb7fc18a03137938d5be3def146bf55033110/crates/wasm/tests/test_build.rs#L45

Then simply adding `lfs=true` to the checkout task worked. I've since reorganized a bit, adding the headless browser tests to the existing WASM workflow. When we next discuss breaking the wasm crate out of the workspace, we can rethink how best to map the tests to the testing coverage we want. For now, the headless browser tests are running just the `test_build` setup, against headless Firefox, which I declare Good Enough™ for merge.

TalDerei:
cc @grod220

grod220:
wow! amazing work 🎊

PR #3505, opened 2023-12-11
storage: remove `commit_in_place`

erwanor:
This PR removes `commit_in_place`, changing our imagined migration strategy from overwriting the pre-upgrade block, to synthesizing a phantom block at the upgrade boundary.


hdevalence:
Can we make an issue to discuss the pros and cons of this approach?

conorsch:
We have a detailed plan for next steps on the upgrade story in #3506. Closing this PR to focus on the work planning described there.

PR #3508, opened 2023-12-12
workspace: rename penumbra-storage to cnidarium

hdevalence:
The name penumbra-storage implies something penumbra-specific, when the storage
crate is not intended to be penumbra-specific.  Renaming the parts of our stack
that aren't specific to Penumbra helps us more clearly distinguish between the
parts of the stack that are specific to our use case and the parts that are
intended to be shared with others.

The word "cnidarium" is a portmanteau of "cnidaria" (the phylum containing
jellyfish) and "aquarium" (to reflect that it's a container for managing
jellyfish merkle trees through their lifecycle).


hdevalence:
7A077399-A17C-4633-AF39-5D3611405DF2_1_102_o

Issue #3504, opened 2023-12-11
Port all existing `event.rs` files to use proto events.


hdevalence:
Following the work in #3460 to define protobuf event handling, it would be good to convert the existing events:

> We should then replace all existing events defined in the Penumbra application with Protobuf-based events. These events should be defined in the Proto module corresponding to the component that emits the event. Currently, there are only 12 events currently defined (VS Code search fn in files named event.rs), so this shouldn't be a huge lift.
>
> Since any Protobuf message could in principle be encoded as an event this way, we should adopt a naming convention to distinguish messages defining events from messages defining other data. One possibility would be to name all of the event messages Event*. Using a prefix rather than a suffix means that all of the events will be grouped together when messages are sorted alphabetically.

In particular, porting over the DEX events will make them useful for trading bots (cf https://discord.com/channels/824484045370818580/1163684711667867678/1181748568135061514 )

erwanor:
Note to self, that we should also port the staking component events to use proto-event.

hdevalence:
Was this completed? We should ensure it's completed as part of V1 -- we can always add more events but changing events people are using will be difficult.

erwanor:
I ported over the DEX events, but got pulled in another direction before doing the rest of the codebase. I believe that @zbuc is tracking the staking events in #3797. I'm unassigning myself for now

zbuc:
The only place using non-protobuf event types is now the IBC code. This will require defining protobuf event types in `ibc-types`

conorsch:
We discussed during sprint-planning today, and we're calling this complete. We specifically _don't_ want to 

> defin[e] protobuf event types in ibc-types

as we're trying to keep that library chain-agnostic, and we don't want to emit Penumbra-specific events as part of that tooling.

PR #3513, opened 2023-12-12
dex(events): use `ProtoEvent` to record events

erwanor:
Part of #3504


proto/penumbra/penumbra/core/component/dex/v1alpha1/dex.proto
```diff
@@ -561,3 +561,55 @@ message SimulateTradeRequest {
...
 }
+
+message EventSwap {
+  // The trading pair to swap.
+  TradingPair trading_pair = 1;
+  // The amount for asset 1.
+  num.v1alpha1.Amount delta_1_i = 2;
+  // The amount for asset 2.
+  num.v1alpha1.Amount delta_2_i = 3;
+  // A commitment to a prepaid fee for the future SwapClaim.
```
hdevalence, line 572: Why do we care about this in the event? Indexers probably don't care about it at all right, there's no reason to have this information...

It does seem useful to have the _swap commitment_, so that it's possible to query the event DB by swap commitment?

erwanor:
I'll add the SCT events in a separate PR. I also want to work through a couple examples of DEX integrations to make sure that we record everything useful. But I will do that offline.

PR #3518, opened 2023-12-13
docs: expand node operator explanations

### LINKED ISSUE #2875: docs: add faqs for running a node
A common ask in the Discord channel is "How can I confirm that my node is running OK?" There are very standard checks a node operator can run, such as:

* poll tendermint rpc endpoint for catchup/block height
* compare tendermint rpc endpoint with public PL-run endpoint
* check logs for advancing blocks
* check pcli query for validator status (with caveats about delays)

Let's document those so it's easier for folks to make sense of what they're running.

conorsch:
Fleshing out the pd docs to address FAQs from new node operators:

  * hardware requirements
  * which services are tied to which ports
  * how is a validator distinct from a fullnode

Also moves the cometbft installation logic into a reusable script, which is exercised in CI via the smoke test.

Closes #2875.


conorsch:
Because this PR touches the summoner smoke test workflow, I've kicked off an ad-hoc run of that job to ensure it still passes: https://github.com/penumbra-zone/penumbra/actions/runs/7201677797

PR #3522, opened 2023-12-15
pd: allow bundling the web frontend into the fullnode

hdevalence:
This is an experiment towards bundling the web frontend code into the fullnode, with the goal that users who run `pd` nodes can self-host a complete Penumbra frontend, on their custom domain with auto-HTTPS, with just one command: `pd start`.

image

To do this, we

- add a new `zipserve` module in `pd`, which takes a `&'static [u8]` with archive data and produces an `axum::Router`;
- bundle an `assets/frontend.zip` (tracked with git-lfs) into the `pd` crate;
- use `axum::Router::merge` to combine the GRPC and frontend endpoints.

To make this work for real, we'd need to:

- [x] Fix up the auto-HTTPS support so that it can work directly with Axum, rather than Tonic;
- [x] Change the frontend code so that all URLs are relativizable to an arbitrary prefix (we want to have it at `/app` when served in `pd`, to allow a status page, but `/` when at `app.testnet.penumbra.zone` or other deployments);
- [x] Create a process to regenerate `assets/frontend.zip` from the `web` repo periodically and check the result into the repo (under LFS).

These seem pretty doable, and it'd be really cool to be able to have easily self-hosted frontends!

### DIFF
--- a/crates/bin/pd/src/main.rs
+++ b/crates/bin/pd/src/main.rs
@@ -230,12 +230,15 @@ async fn main() -> anyhow::Result<()> {
                 )));
             }
 
-            // Now we drop down a layer of abstraction, from tonic to axum.
-            //
-            // TODO(kate): this is where we may attach additional routes upon this router in the
-            // future. see #3646 for more information.
+            // Create Axum routes for the frontend app.
+            let frontend = pd::zipserve::router("/app/", pd::MINIFRONT_ARCHIVE_BYTES);
+            let node_status = pd::zipserve::router("/", pd::NODE_STATUS_ARCHIVE_BYTES);
+
+            // Now we drop down a layer of abstraction, from tonic to axum, and merge handlers.
             let router = grpc_server
                 .into_router()
+                .merge(frontend)
+                .merge(node_status)
                 // Set rather permissive CORS headers for pd's gRPC: the service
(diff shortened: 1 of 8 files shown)


crates/bin/pd/src/main.rs
```diff
@@ -458,6 +458,12 @@ async fn main() -> anyhow::Result<()> {
                 )));
             }
 
+            // TODO: will need a bit of finesse to make this work with auto-https,
+            // since auto-https support above works with the tonic `serve_with_incoming`
+            // method, but now we're dropping down to axum directly. just needs some realignment
+            // probably.
```
hdevalence, line 464: In the ~year since this code was written, there may be better options. https://docs.rs/tokio-rustls-acme/latest/tokio_rustls_acme/ seems promising

cratelyn:
opened #3646 to track landing this feature.

hdevalence:
Great! I have a rebase of this branch in progress, I'll pick it up next.

hdevalence:
Here's a video of this code in action:

https://github.com/penumbra-zone/penumbra/assets/44879/1697c681-4b09-4cf8-b655-c94356eb56fd

Note that sending a transaction fails because the web extension hasn't been updated to the new proving parameters. However, it otherwise seems to work!

PR #3527, opened 2023-12-19
Flexible provenance tracking for state commitments

### LINKED ISSUE #3349: Provenance Tracking
Clients need to be able to identify and track provenance of all of the funds they control. 

This is important to support use cases like “where did these funds come from” and “who sent me money”. Because Penumbra is a shielded chain, if we do not design this capability in, it may not be possible at all. We need to think through the end-to-end data flow before mainnet, because we may need to make changes to the chain state to be able to track all 
the necessary information.

_What we need to implement_

- [ ] Add source information to SpendableNoteRecord
- [ ] Determine how to represent the source address — the AddressView can only represent a Penumbra address, but the source might be an address on another chain that did an ICS20 transfer
- [ ] Determine how the view server can obtain the source information during scanning
- [ ] Re-evaluate the design of the NoteSource — should it really be a “stuffed” hash?

hdevalence:
Closes #3349

This PR removes the previous `NoteSource`, a stuffed transaction hash, with
a new, extensible `CommitmentSource`, and makes managing commitment sources the
responsibility of the SCT component that manages state commitments.  This
allows several important use cases:

- We can track provenance of funds as they enter the chain, by recording the
  sender address, packet number, and channel for `NotePayloads` created by the
  ICS20 handler. This is particularly important because there is no "transaction"
  to download to fetch extended information there.

- Clients have an easier time deciding whether they need to do an extended
  fetch of block data, they just need to see if any detected payload was from a
  `Transaction` source.

- We can save space in the compact block by "dehydrating" the transaction
  sources and removing the transaction hashes.  Clients _already_ have to fetch
  a full block's worth of transactions, because nullifiers don't have source info
  attached, so we gain nothing by including detailed source info in the `CompactBlock`.

- Clients have an easier time working with Penumbra data, since there's no more
  ad-hoc stuffing.
  
In addition, it does various other cleanup along the way.


crates/bin/pcli/src/command/query/shielded_pool.rs
```diff
@@ -13,16 +12,6 @@ pub enum ShieldedPool {
...
-    /// Queries the state commitment tree's block anchor for a given height.
-    BlockAnchor {
-        /// The height to query.
-        height: u64,
-    },
-    /// Queries the state commitment tree's epoch anchor for a given epoch index.
-    EpochAnchor {
-        /// The epoch to query.
-        epoch: u64,
-    },
```
hdevalence, line 25: This is functionality that should really be part of an indexer / explorer, rather than asking every full node to keep the data around in case someone wants to run this command against a random fullnode.

crates/core/app/src/tests/spend.rs
```diff
@@ -60,6 +62,7 @@ async fn spend_happy_path() -> anyhow::Result<()> {
     spend.check_stateless(transaction_context).await?;
     spend.check_stateful(state.clone()).await?;
     let mut state_tx = state.try_begin_transaction().unwrap();
+    state_tx.put_mock_source(1u8);
```
hdevalence, line 65: This has to be inserted now that the SCT logic tracks provenance, otherwise execution will fail, because in the mock setting, we're only running a single action in isolation, which wouldn't pick up the transaction-level source setting.

hdevalence:
Demo of this support in `pcli`:

PR #3529, opened 2023-12-21
fix(proto): update `DomainType::Proto` for `ClientState/ConsensusState`

noot:
Found this bug while testing the IBC implementation with Hermes. Previously it was writing an `Any` type to state at the client_state/consensus_state storage locations, but the type Hermes was expecting was the `ibc_proto` `ClientState`/`ConsensusState`. This resulted in a proto decoding error in these locations:
https://github.com/informalsystems/hermes/blob/master/crates/relayer/src/client_state.rs#L127
https://github.com/informalsystems/hermes/blob/master/crates/relayer/src/consensus_state.rs#L59

These are called in the `Endpoint` `query_client_state`/`query_consensus_state` methods (for example here https://github.com/informalsystems/hermes/blob/b98ca86991fc1cc6d3f6c595918251a34eb6f1db/crates/relayer/src/chain/cosmos.rs#L1238)

specifically the error I got from Hermes was `Error::decode_raw_client_state`:
```
ERROR failed while querying for client 07-tendermint-0 on chain id astria: other error: error decoding raw client state: error decoding buffer into message: failed to decode Protobuf message: Fraction.numerator: ClientState.trust_level: invalid wire type: LengthDelimited (expected Varint)
```

I can give more details on how I was running everything if needed!

also linted the protos cause CI complained 😭

### DIFF
--- a/crates/proto/src/protobuf.rs
+++ b/crates/proto/src/protobuf.rs
@@ -156,7 +156,6 @@ impl TryFrom<crate::core::keys::v1alpha1::ConsensusKey> for tendermint::PublicKe
 // IBC-rs impls
 extern crate ibc_types;
 
-use ibc_proto::google::protobuf::Any;
 use ibc_proto::ibc::core::channel::v1::Channel as RawChannel;
 use ibc_proto::ibc::core::client::v1::Height as RawHeight;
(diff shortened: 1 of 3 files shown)


hdevalence (approved):
LGTM! Thanks for the find.

Issue #3391, opened 2023-11-16
Tweak proto definitions for fees


hdevalence:
We should consider:

- [x] moving the fee into the `TransactionParameters`
- [ ] adding a `FeeView` to provide denom support.

hdevalence:
The first point was done in #3542, but there's no `FeeView`. I'm not sure if we should add it, we would have to make a `TransactionParametersView` and it doesn't do much because we only accept one asset type as the fee token.

PR #3539, opened 2023-12-26
proto: clean up tx format with breaking changes

hdevalence:
This makes a few cleanup changes that align things for the long term:

- [x] remove use of `bytes` nullifiers
- ~~[ ] move anchor into `TransactionParameters`~~ (this is a bad idea, it's witness data)
- [x] move fee into `TransactionParameters`
- [x] eliminate `Fee` type in favor of `Value`
- [x] align `TransactionPlan` with `Transaction`


hdevalence:
Closes #3391

Issue #3541, opened 2023-12-28
`TransactionPlan` proto messages are incorrect and were not kept in sync with `Transaction` changes


hdevalence:
**Describe the bug**

The layout of the `TransactionPlan` proto message does not match the `Transaction` message, because the `Transaction` was edited without corresponding changes to the `TransactionPlan`.

PR #3542, opened 2023-12-29
proto: transaction format cleanup

hdevalence:
This commit:

- Aligns the `TransactionPlan` with the `Transaction` data structure
- Moves the `fee` into the `TransactionParameters`

Now the transaction body consists solely of:

- a list of `Action`s describing changes to the chain state
- a set of `TransactionParameters` describing under what conditions those changes can be applied
- a set of `DetectionData` (extensible) for assisting with transaction detection
- a set of `MemoData` with encrypted information about the transaction's purpose.


hdevalence:
Closes #3391 #3541

hdevalence:
We have a lot of memo-related types.

crates/core/transaction/src/effect_hash.rs
```diff
@@ -55,27 +55,26 @@ impl TransactionBody {
...
-            state.update(
-                detection_data
-                    .expect("detection data is some")
-                    .effect_hash()
-                    .as_bytes(),
-            );
-        }
+        state.update(parameters_hash.as_bytes());
+        state.update(memo_hash.as_bytes());
+        state.update(detection_data_hash.as_bytes());
```
hdevalence, line 77: cc @redshiftzero ; in the case of a ledger or similar device, we can avoid having the ledger compute FMD clues by passing the detection data hash from the host device to the ledger

PR #3547, opened 2023-12-30
Correct naming to "Community Pool"

hdevalence:
The functionality is directly copied from the Cosmos SDK's community pool, and naming it as such is important to correctly convey expectations about what it is.


crates/bin/pcli/src/command/tx/proposal.rs
```diff
@@ -12,7 +12,7 @@ pub enum ProposalCmd {
         /// The file to output the template to.
         #[clap(long, global = true)]
         file: Option<camino::Utf8PathBuf>,
-        /// The kind of the proposal to template [one of: signaling, emergency, parameter-change, or dao-spend].
+        /// The kind of the proposal to template [one of: signaling, emergency, parameter-change, or Community Pool-spend].
```
hdevalence, line 15: s/r error

PR #3555, opened 2024-01-02
feat(pd): report bind errors on startup

conorsch:
A common problem for node operators is sorting out which ports are required to be available on the system prior to running. We've recently improved the documentation to be more explicit about this [0], but here we add explicit error messages when binds fail. This addition should make reasoning about network surface more straightforward, and eliminate guessing games to figure out which port needs to be overridden or freed.

[0] c0445f34752f64a253b672188f3a6d8f5f58c116


conorsch:
I made sure to make this diff compatible with `v0.64.1`, so that we can issue it as a point release. It's a minor change, and there's a lot of discussion in Discord about bind ports.

conorsch:
Backported to `v0.64.x` in  #3590.

PR #3556, opened 2024-01-02
Create a general convert circuit

### LINKED ISSUE #3523: Generalize `UndelegateClaimCircuit` to more general circumstances
One discussion that motivated this was wanting to implement restaking natively. A circuit for doing this would need to convert delegated tokens of one validator into staking tokens of another validator, by looking at the current exchange rates relative to the base staking token. In this case, the two validators are known, the rate is known (as a consequence), but the amount being converted is hidden. The result of the circuit produces a commitment to the amount consumed and produced, and proves that they're related by the exchange rate.

This is awfully similar to the `UndelegateClaimCircuit` we have now, which does the same thing, but with one asset being implicitly the staking token, and the exchange rate being the penalty.

A generalization of this circuit applicable to this case, and potentially others, would be a circuit with:
- Public assets `from`, `to`, and an exchange rate, a public balance commitment
- A private amount of `from`.
The circuit would prove that the balance commits to `-amount[from] + (rate * amount)[to]`, without revealing the amount.

Consensus rules using this circuit would then check that the assets are correct, and the rate is correct as well.

cronokirby:
Closes #3523.

This approach doesn't replace UndelegateClaimProof, but instead just changes the circuit that proof uses to be a more general one.


hdevalence:
Should the `ConvertProof` be part of the shielded pool component?

redshiftzero (approved):
LGTM

cronokirby:
> Should the `ConvertProof` be part of the shielded pool component?

This is a simple change that makes sense, imo, better to do now than have confusion later.

PR #3557, opened 2024-01-02
Remove wallet id filter from view service

grod220:
Closes https://github.com/penumbra-zone/web/issues/334

Only one wallet is active at a time for the view service. Therefore, we can remove all of these `wallet_id` filters for it.


proto/penumbra/penumbra/view/v1alpha1/view.proto
grod220, line 1: Critical thing to review. Proto messages have been updated.

crates/bin/pcli/src/command.rs
grod220, line 1: Think RustRover reformats the imports in this way to make publicly exposed exports at the top

hdevalence: kind of noisy but seems fine. there's a way to get cargo fmt to do this but it requires using nightly rustfmt and a custom config and that seemed like an unacceptable DX tax to put on every contributor, but the effect is that our imports don't stay tidy. so occasionally having them all cleaned up seems fine.

proto/penumbra/penumbra/view/v1alpha1/view.proto
```diff
@@ -21,9 +21,6 @@ import "penumbra/crypto/tct/v1alpha1/tct.proto";
 // transaction-related actions, to request data from a view service, which is
 // responsible for synchronizing and scanning the public chain state with one or
 // more full viewing keys.
-//
-// View protocol requests optionally include the wallet id, used to
-// identify which set of data to query.
```
hdevalence, line 26: If we did want to have a multi-tenant view service in the future, we could do so by putting the wallet ID in the request headers. This would be reasonably easy to implement on the Rust side. But not important for now, just noting that the option is open.

hdevalence (approved):
LGTM!

PR #3558, opened 2024-01-03
pcli: use a default timeout height for ibc withdrawals

### LINKED ISSUE #3206: pcli: timeout height for ibc transfers should have default value
The command `pcli tx withdraw` takes a `--timeout-height`, which is difficult for users to specify accurately. The "default" value is currently `0-0`, but this will never work, always yielding the error: `error: Invalid value "0-0" for '--timeout-height <TIMEOUT_HEIGHT>': attempted to parse an invalid zero height`.

The structure of that arg is a tuple of:

1. revision number on counterparty chain id (e.g., `osmo-test-5` -> `5`)
2. block height on counterparty chain

In order to provide a sane default value, given that we already require the user to specify a `--channel-id`, we can use that information to look up IBC client information, and infer the revision number from the last recorded height for the client. The height that the client stores in is in IBC format, meaning the revision number and block height are already specified, meaning we don't need a separate lookup.

Finally, inspired by the `rly` tool, let's change the flag from `--timeout-height` to `--timeout-height-offset`, so that expiry happens _n_ blocks in the future. As for a sane value for this, how about 1000.

avahowell:
Closes #3206 . Won't work until we have implemented the remaining IBC queries (currently all the channel ones are missing)

### DIFF
--- a/crates/bin/pcli/src/command/tx.rs
+++ b/crates/bin/pcli/src/command/tx.rs
@@ -9,7 +9,16 @@ use std::{
 use anyhow::{Context, Result};
 use ark_ff::UniformRand;
 use decaf377::{Fq, Fr};
+use ibc_proto::ibc::core::client::v1::{
+    query_client::QueryClient as IbcClientQueryClient, QueryClientStateRequest,
+};
+use ibc_proto::ibc::core::connection::v1::query_client::QueryClient as IbcConnectionQueryClient;
+use ibc_proto::ibc::core::{
+    channel::v1::{query_client::QueryClient as IbcChannelQueryClient, QueryChannelRequest},
+    connection::v1::QueryConnectionRequest,
+};
 use ibc_types::core::{channel::ChannelId, client::Height as IbcHeight};
+use ibc_types::lightclients::tendermint::client_state::ClientState as TendermintClientState;
 use rand_core::OsRng;
 use regex::Regex;
 
@@ -179,8 +188,8 @@ pub enum TxCmd {
         /// height, e.g. `5-1000000` means "chain revision 5, block height of 1000000".
         /// You must know the chain id of the counterparty chain beforehand, e.g. `osmosis-testnet-5`,
         /// to know the revision number.
-        #[clap(long, default_value = "0-0", display_order = 100)]
-        timeout_height: IbcHeight,
+        #[clap(long, display_order = 100)]
+        timeout_height: Option<IbcHeight>,
         /// Timestamp, specified in epoch time, after which the withdrawal will be considered
         /// invalid if not already relayed.
         #[clap(long, default_value = "0", display_order = 150)]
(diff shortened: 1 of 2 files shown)


conorsch:
> Won't work until we have implemented the remaining IBC queries (currently all the channel ones are missing)

Added in https://github.com/penumbra-zone/penumbra/pull/3572

crates/bin/pcli/src/command/tx.rs
```diff
@@ -858,13 +867,72 @@ impl TxCmd {
...
+                    None => {
+                        // look up the height for the counterparty and add 2 days of block time
+                        // (assuming 10 seconds per block) to it
+
+                        // look up the client state from the channel by looking up channel id -> connection id -> client state
+                        let mut ibc_channel_client =
+                            IbcChannelQueryClient::new(app.pd_channel().await?);
+
+                        let req = QueryChannelRequest {
+                            port_id: "transfer".to_string(),
```
conorsch, line 881: Side note: we hardcode `"transfer"` a lot of places, when referring to the IBC port. Should we not constify that value in the IBC crate somewhere, so we can tersely refer to it as the constant value it is?

crates/bin/pcli/src/command/tx.rs
```diff
@@ -858,13 +867,72 @@ impl TxCmd {
                     .full_viewing_key
                     .ephemeral_address(OsRng, AddressIndex::from(*source));
 
+                let timeout_height = match timeout_height {
+                    Some(h) => h.clone(),
+                    None => {
+                        // look up the height for the counterparty and add 2 days of block time
+                        // (assuming 10 seconds per block) to it
```
conorsch, line 874: Why assume 10s per block? Looking at the current Testnet 64, it looks like blocks actually take about 5s:

block-times

which I assume is coming from the cometbft value we set here: https://github.com/penumbra-zone/penumbra/blob/v0.64.1/testnets/cometbft_config_template.toml#L413

crates/bin/pcli/src/command/tx.rs
```diff
@@ -858,13 +867,72 @@ impl TxCmd {
...
+                            .client_state
+                            .ok_or_else(|| anyhow::anyhow!("client state not found"))?;
+
+                        let tm_client_state = TendermintClientState::try_from(client_state)?;
+
+                        let last_update_height = tm_client_state.latest_height;
+
+                        IbcHeight {
+                            revision_number: last_update_height.revision_number,
+                            revision_height: last_update_height.revision_height + 87400,
```
conorsch, line 925: Related to question above: where does `87400` come from? `24 * 60 * 60 = 86400` but the comment at the top of the match suggests it should be _two_ days, not one. Can we var this value and explain its construction in a comment?

conorsch (commented):
Requesting some clarification to aid my own understanding of the diff.

crates/bin/pcli/src/command/tx.rs
hdevalence: The number applies to counterparty chains, so we can't get it correct, since they all have different values, it'll just be an approximation. See 

https://twitter.com/larry0x/status/1742764484144320997/photo/1

all the cosmos chains have different block times

conorsch: Ahh, of course. That makes a lot more sense!

crates/bin/pcli/src/command/tx.rs
avahowell: We do this, it's `Portid::transfer()` in ibc-types. i'll update this pr to use that

crates/bin/pcli/src/command/tx.rs
avahowell: it should be ((24*60*60)/10) * 2, since we're assuming 10s per block

avahowell:
Updated with review comments

conorsch:
Awesome, thanks for explaining, @avahowell!

Issue #3549, opened 2023-12-31
Change View RPCs to allow streaming progress indicators during build


hdevalence:
> there's not a way to change this with the current view service rpc, right? should we change it so that it does streaming responses, for the different stages (returning a message when built, broadcasted, confirmed, etc)?

_Originally posted by @hdevalence in https://github.com/penumbra-zone/web/issues/324#issuecomment-1872452247_

Some questions to figure out:

- How should the streaming responses be structured? What's a way to allow progress indicators but be relatively agnostic about how the server implementation supplies them?
- Should this be a second RPC, like *WithProgress, so that clients that don't care about progress don't need to consume a stream?

grod220:
This would have to be converted to proto message format, but maybe something like this?

```rust


enum Status {
    Started,
    Completed,
}

enum BroadcastStatus {
    Started,
    Submitted(Id),
    Confirmed,
}

enum Stage {
    Plan(Status),
    Authorization(Status),
    Witness(Status),
    Build(Status),
    Broadcast(BroadcastStatus),
    Confirmed,
}

struct TxSubmitStreamRequest {
    req: TransactionPlannerRequest, // Should be the same as TxPlannerReq?
}

struct TxSubmitStreamResponse {
    stage: Stage,
}
```

I don't know if this extensible in the protobuf sense though 🤔 

Maybe this is even more generic

```rust
struct Stage {
    name: String,
    progress: u8,
    total: u8
}

struct TxSubmitStreamRequest {
    req: TransactionPlannerRequest,
}

struct TxSubmitStreamResponse {
    stage: Stage,
}
```

hdevalence:
The planning shouldn't be included, it's instant anyways. Similarly for the "witness" step. We should only care (if at all) about building and submitting the transaction.

For building the transaction, the status message should have a float for a progress value between 0 and 1, which the server can set arbitrarily. The message sequence should look like `repeated [status update] => complete` (the server can send as many status updates as it wants and then completes).

For broadcasting the transaction, we only care about two stages, "submitted" and "confirmed".

Do we want to support one-shot methods at all? Should we just change the existing methods to return streaming responses? We already make pretty heavy use of streaming methods, so maybe it's not unreasonable to ask clients to do "consume this stream until it completes".

PR #3559, opened 2024-01-03
Groundwork for Streaming Progress Updates from View Service

hdevalence:
Sketch of one possibility for #3549 for discussion. (No code changes as yet, only proto changes).


proto/penumbra/penumbra/view/v1alpha1/view.proto
```diff
@@ -115,8 +134,20 @@ message AuthorizeAndBuildRequest {
...
+  }
+  // Signals that the transaction is complete.
+  message Complete {
+    // The finished transaction.
+    penumbra.core.transaction.v1alpha1.Transaction transaction = 1;
+  }
+  oneof status {
+    BuildProgress build_progress = 1;
+    Complete complete = 2;
+  }
```
grod220, line 150: Looks like these can be unnested so it can be shared with WitnessAndBuildResponse too

proto/penumbra/penumbra/view/v1alpha1/view.proto
```diff
@@ -115,8 +134,20 @@ message AuthorizeAndBuildRequest {
 }
 
 message AuthorizeAndBuildResponse {
-  // The transaction that was built.
-  core.transaction.v1alpha1.Transaction transaction = 1;
+  // Signals that building is in progress.
+  message BuildProgress {
+    // An approximate progress of the build, from 0 to 1.
+    float progress = 1;
```
grod220, line 140: Does this correlate to the % of actions that have been built thus far?

grod220 (dismissed):
Just two comments from my end

hdevalence:
There's something off with the CI, this PR is green but it should definitely be red, it's half-finished and only changes protos, doesn't recompile them or propagate changes through the rest of the codebase. @conorsch any ideas?

proto/penumbra/penumbra/view/v1alpha1/view.proto
hdevalence: It's up to the implementation, I guess, in principle an implementation that had insight into how much proving was done as it was happening could give better updates. I believe there's a way to get Rayon to do this on the Rust side. I think at the proto level we should be as agnostic as possible, and initially, we should just try to expose info we have rather than adding progress plumbing into the proving internals.

proto/penumbra/penumbra/view/v1alpha1/view.proto
hdevalence: Sounds good, feel free to pick it up, we'll also have to propagate these changes through the rest of the source code.

conorsch:
Good catch, I think I broke this back in https://github.com/penumbra-zone/penumbra/pull/3297, which altered the behavior so that the buf checks were run only when the PR is first opened, but not every time it's changed. Updated the logic in #3575, once that's in let's rebase this one and confirm we see the buf checks again!

aubrika:
Blocked on #3621

conorsch:
CI is now passing on this PR, module the protobuf lint failure, which is to be expected, because the proto changes here are Breaking Changes™. Will continue investigation of sweep test in #3621. Proceeding with merge here to unblock web impl consuming this new interface.

PR #3562, opened 2024-01-03
ci: increase cuiloa testnet storage

conorsch:
The testnet fullnodes had their storage increased in [0], and the fullnode backing the testnet cuiloa instance should have the same allotment. Also pinning a specific version of penumbra container, so that newer (and incompatible) versions aren't pulled on restart.

[0] bff43e609e7a2a1730728ee7a310e4c8eeb603d4


conorsch:
Applied this diff manually, watching for fullnode to catch up, merging in the meantime.

PR #3566, opened 2024-01-03
chore: cargo lint

### LINKED ISSUE #3543: Fix unused import visibility and re-enable warnings in CI

In #3533 I noticed that changes to Rust had added new warnings causing the build to fail. I disabled the warnings step in CI. We should fix the new warnings (which appear to be about more precise tracking of visibility) then re-enable the check.

conorsch:
Satisfies the latest cargo lints. I took the easy way out, preserving the existing function sigs, rather than performing a deeper refactor. Ignored the checks on the re-export of metrics code across components.
    
Holding off on adding `--all-targets` to the check command, because there's some cleanup to do in the benchmarks, related to #3556.
    
Revert "ci: disable -D warnings"
This reverts commit 4e7c77b7404e69ab279d8d273a56935e7dba53c7.
    
Refs #3543.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -15,13 +15,14 @@ jobs:
         uses: taiki-e/install-action@nextest
       - name: Load rust cache
         uses: astriaorg/buildjet-rust-cache@v2.5.1
-      - name: Run cargo check
+      - name: Run cargo check, failing on warnings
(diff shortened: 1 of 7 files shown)


conorsch:
Downgrading to draft because @cronokirby noticed we weren't checking the benchmarks, and #3556 made some changes that now cause lints to fail.

Issue #3563, opened 2024-01-03
Refactor Proof APIs to take Public and Private input as separate structs


cronokirby:
See the `ConvertProof` introduced in https://github.com/penumbra-zone/penumbra/pull/3556 as an example of doing this. The idea is that each proof will have a related struct for the public inputs and private inputs to the circuit. We should also make the fields of the `Circuit` struct private, only allowing creating circuits via the `DummyWitness` trait.

Some advantages of this:
- It makes it clearer what the public and private inputs are.
- It removes duplication of parameters between proving and verifying.
- It makes the functions have fewer parameters, whereas now they have a lot, to the point that clippy complains.
- It makes writing a Rust version of the circuit satisfaction code easier.

*Circuits that need to be changed*:
- [x] `Spend`
- [ ] `Output`
- [x] `DelegatorVote`
- [ ] `UndelegateClaim`
- [ ] `Swap`
- [ ] `SwapClaim`
- [x] `NullifierDerivation`

PR #3571, opened 2024-01-04
refactor: define public, private structs for spend, nullifier derivation and delegator vote circuits

redshiftzero:
Three items towards #3563

### DIFF
--- a/crates/core/component/governance/src/delegator_vote/proof.rs
+++ b/crates/core/component/governance/src/delegator_vote/proof.rs
@@ -89,8 +89,7 @@ pub struct DelegatorVoteCircuit {
 }
 
 impl DelegatorVoteCircuit {
-    #[allow(clippy::too_many_arguments)]
-    pub fn new(
+    fn new(
         DelegatorVoteProofPublic {
             anchor,
             balance_commitment,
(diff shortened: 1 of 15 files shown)


crates/core/component/shielded-pool/src/spend/proof.rs
```diff
@@ -47,30 +76,33 @@ pub struct SpendCircuit {
...
-    pub nullifier: Nullifier,
+    nullifier: Nullifier,
     /// the randomized verification spend key.
-    pub rk: VerificationKey<SpendAuth>,
+    rk: VerificationKey<SpendAuth>,
 }
 
 impl SpendCircuit {
     #[allow(clippy::too_many_arguments)]
     pub fn new(
```
cronokirby, line 91: Can we make new private, relying on the existence of DummyWitness instead?

crates/core/component/governance/src/delegator_vote/proof.rs
```diff
@@ -45,33 +76,36 @@ pub struct DelegatorVoteCircuit {
...
-    pub rk: VerificationKey<SpendAuth>,
+    rk: VerificationKey<SpendAuth>,
     /// the start position of the proposal being voted on.
-    pub start_position: tct::Position,
+    start_position: tct::Position,
 }
 
 impl DelegatorVoteCircuit {
     #[allow(clippy::too_many_arguments)]
     pub fn new(
```
cronokirby, line 93: Similarly, we should try and make new private here if we can.

cronokirby (approved):
Looks good to me, a few minor bits of cleanup could be to make Circuit::new private, and to remove the clippy exceptions now that they should be ok.

PR #3572, opened 2024-01-04
ibc: implement missing query RPCs

avahowell:


erwanor (approved):
LGTM

conorsch:
Follow-up to  #3531. This gets us most of the way toward IBC query coverage. Still to come are:

* client_status
* client_params
* upgraded_client_state
* upgraded_consensus_state

All of which still return "unimplemented" in this PR. We'll follow up on those when we need them.

PR #3575, opened 2024-01-04
ci: fix buf workflow on PRs

### LINKED ISSUE #3570: Buf CI actions aren't running
**Describe the bug**

I noticed that the Buf CI actions aren't running. #3559 has green CI but is broken and incomplete.

conorsch:
Follow up to #3292, which incorrectly used `pull_request_target` rather than `pull_request` to restrict workflow runs to PRs targeting main. We still want to skip the buf workflows if the PR is targeting a feature branch.

Closes #3570.

### DIFF
--- a/.github/workflows/buf-pull-request.yml
+++ b/.github/workflows/buf-pull-request.yml
@@ -1,12 +1,12 @@
 name: Protobuf
 on:
   # Exclude feature branches, only run if the PR is targeting main.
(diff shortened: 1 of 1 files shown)


conorsch:
I amended the commit message, force-pushed, and confirmed that the buf lint action ran _again_ on this PR. Good to merge after CI passes.

PR #3579, opened 2024-01-05
Use "convert" as the name of the parameters for the proving keys

### LINKED ISSUE #3576: Rename `undelegateclaim_pk/vk` to `convert_pk/vk`
After merging https://github.com/penumbra-zone/penumbra/pull/3556, we should also change the name of this file to reflect the new circuit.

We also need to make a PR to https://github.com/penumbra-zone/web to change which file is referenced.

cronokirby:
Closes #3576.

(We'll also need to change the web crate)


redshiftzero (approved):
Seems straightforward, LGTM!

PR #3580, opened 2024-01-05
Fix unused variable / import warnings in pcli

cronokirby:
I don't quite know when this regression was introduced, but it's related to unused variables and imports.


hdevalence:
This wasn't a regression, rather, rustc changed how it emitted warnings and started noticing new things to warn against. (Good to fix anyways)

conorsch:
See https://github.com/penumbra-zone/penumbra/issues/3543, https://github.com/penumbra-zone/penumbra/pull/3566. I'll merge every passing lint fix. It's just the benchmarks that aren't updated yet, then we'll have full coverage across the repo.

conorsch:
> It's just the benchmarks that aren't updated yet

Looks like #3579 fixes that!

Issue #3564, opened 2024-01-03
Implement In-Rust Version of Proof Satisfaction


cronokirby:
This is a follow up to https://github.com/penumbra-zone/penumbra/issues/3563. After doing that, it becomes easy to write a function which implements the function of type `Public x Private -> 01` representing the relation the proof is checking. This then makes the test of circuit logic very simple, since we just need to check that for inputs we generate, the rust code agrees with the circuit.

We should add such a function to each circuit, and modify our tests to use it.

*Circuits that need to be changed*:
- [x] `Spend`
- [x] `Output`
- [x] `DelegatorVote`
- [ ] `UndelegateClaim`
- [x] `Swap`
- [x] `SwapClaim`
- [x] `NullifierDerivation`

PR #3583, opened 2024-01-05
test: add proptests of first three circuits vs an in rust impl

redshiftzero:
Three items towards #3564


cronokirby (approved):
Ok, these match the circuit from what I can tell, and the behavior makes sense.

One thing to think about is if we might have forgotten to check if something is not an identity point, but I think we've gotten that all.

PR #3585, opened 2024-01-06
MemoPlaintext string length enforced, TransactionPlan panic avoided

turbocrime:
identified a panic when working on web transactions.

added checks to verify text length is under maximum when deserializing from `pbt::MemoPlaintext` or `Vec<u8>`, to prevent the creation of impossible memos. perhaps `MemoPlaintext` should use private fields and a constructor to enforce format limitations that cannot be specified by protobuf, but this is a smaller change.

ultimately the source of the panic was an `expect` used in `TransactionPlan::effect_hash` to unwrap a potential error, this is now bubbled as `anyhow::Result`

### DIFF
--- a/crates/core/transaction/src/plan.rs
+++ b/crates/core/transaction/src/plan.rs
@@ -56,7 +56,7 @@ impl TransactionPlan {
     ///
     /// This method is not an [`EffectingData`] impl because it needs an extra input,
     /// the FVK, to partially construct the transaction.
-    pub fn effect_hash(&self, fvk: &FullViewingKey) -> EffectHash {
+    pub fn effect_hash(&self, fvk: &FullViewingKey) -> Result<EffectHash> {
         // This implementation is identical to the one for Transaction, except that we
         // don't need to actually construct the entire `TransactionBody` with
         // complete `Action`s, we just need to construct the bodies of the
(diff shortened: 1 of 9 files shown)


crates/core/transaction/src/plan.rs
```diff
@@ -68,18 +68,10 @@ impl TransactionPlan {
...
-                    .expect("can compute memo ciphertext")
-                    .effect_hash()
-            })
-            // If the memo is not present, use the all-zero hash to record its absence in
-            // the overall effect hash.
-            .unwrap_or_default();
+        let memo_hash = match self.memo {
+            Some(ref memo) => memo.memo()?.effect_hash(),
+            None => EffectHash::default(),
+        };
```
turbocrime, line 74: this is a significant change in style compared to the previous implementation, not sure if that's appropriate

cronokirby (approved):
I like this a lot; I feel like it would be worth it to refactor MemoPlaintext to have a guarded constructor, but I'll take care of that in a separate PR.

crates/core/transaction/src/plan.rs
cronokirby: I like the result of the change better, it's more succinct this way.

PR #3586, opened 2024-01-07
wasm(fix): failing unit test

TalDerei:
Fixes breaking changes from https://github.com/penumbra-zone/penumbra/pull/3527 and https://github.com/penumbra-zone/penumbra/pull/3579

### DIFF
--- a/crates/wasm/tests/test_build.rs
+++ b/crates/wasm/tests/test_build.rs
@@ -61,7 +61,7 @@ mod tests {
             serde_wasm_bindgen::to_value(&nullifier_derivation_key).unwrap();
         let swap_key_js: JsValue = serde_wasm_bindgen::to_value(&swap_key).unwrap();
         let swap_claim_key_js: JsValue = serde_wasm_bindgen::to_value(&swap_claim_key).unwrap();
-        let undelegate_claim_key_js: JsValue = serde_wasm_bindgen::to_value(&convert_key).unwrap();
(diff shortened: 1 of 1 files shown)


conorsch:
How does one confirm or disprove that this change fixes the test? If I ran `cargo test` within `crates/wasm/`, I see all tests passing on both this branch and on main.

TalDerei:
> How does one confirm or disprove that this change fixes the test? If I ran `cargo test` within `crates/wasm/`, I see all tests passing on both this branch and on main.

The wasm crate incorporates a unit testing suite that leverages `wasm-bindgen-test` to simulate a sample spend transaction. For testing purposes, the suite mocks indexDB database calls in an interactive browser accessible at `http://127.0.0.1:8000`. This requires executing `wasm-pack test --chrome -- --test test_build --target wasm32-unknown-unknown --release --features "mock-database"` rather than `cargo test` to check that particular test, and the transaction outputs are accessible in the browser's developer console. The serial and parallel transaction building were failing prior to these changes. 

https://github.com/penumbra-zone/penumbra/pull/3503 will eventually integrate this test, and I suspect a more expressive wasm unit testing suite, into the CI using the headless browser mode.

conorsch:
Nice, thanks for explaining, @TalDerei! I can now confirm that with the invocation you provided, tests fail on main, as expected, and pass on this branch.

PR #3590, opened 2024-01-08
feat(pd): report bind errors on startup

conorsch:
A common problem for node operators is sorting out which ports are required to be available on the system prior to running. We've recently improved the documentation to be more explicit about this [0], but here we add explicit error messages when binds fail. This addition should make reasoning about network surface more straightforward, and eliminate guessing games to figure out which port needs to be overridden or freed.

[0] c0445f34752f64a253b672188f3a6d8f5f58c116

(cherry picked from commit 8680e838f8afc074d8cb189e906c4d3334b76f67)

### DIFF
--- a/crates/bin/pd/src/main.rs
+++ b/crates/bin/pd/src/main.rs
@@ -6,6 +6,7 @@ use std::{net::SocketAddr, path::PathBuf};
 use console_subscriber::ConsoleLayer;
 use metrics_tracing_context::{MetricsLayer, TracingContextLayer};
 use metrics_util::layers::Stack;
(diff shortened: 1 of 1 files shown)


conorsch:
The Rust CI job [is failing](https://github.com/penumbra-zone/penumbra/actions/runs/7451780286/job/20273605715?pr=3590) on some unhandled warnings; that's because https://github.com/penumbra-zone/penumbra/issues/3543 hasn't been backported to the release branch. Perhaps we should pin a specific version of rust in a toolchain file to avoid this kind of false negative; I'll file a separate issue.

PR #3592, opened 2024-01-08
feat: 📝 implement mock-cometbft crate (work-in-progress)

cratelyn:
this commit introduces a new library, in `crates/test/`. this library contains a mock implementation of cometbft, for use in cargo integration tests.

* this does NOT add the `penumbra-mock-cometbft` crate to the list of crates included in the rust documentation in `deployments/scripts/rust-docs`. the `penumbra-tct-property-test` crate was also not included in that list at the time of writing.

/!\ ------------------------------------------------------- /!\
/!\ NOTE: this is a rolling work-in-progress.               /!\
/!\ this branch will be force-pushed frequently until it is /!\
/!\ ready for review. proceed accordingly!                  /!\
/!\ ------------------------------------------------------- /!\

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5477,6 +5477,10 @@ dependencies = [
 ]
 
 [[package]]
+name = "penumbra-mock-cometbft"
+version = "0.64.1"
+
+[[package]]
 name = "penumbra-num"
 version = "0.64.1"
 dependencies = [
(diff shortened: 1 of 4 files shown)


hdevalence:
Two high-level comments:

1. we should avoid using the word "mock" to prevent confusion with cometmock, which is a different piece of software
2. we should have a separate crate, maybe "tenderstub", that _just_ has the functionality of producing headers (and possibly abci functionality, but no penumbra-specific code) so that this component can be more easily reused in other parts of the ecosystem (eg https://x.com/hdevalence/status/1736860951000031672 )

PR #3598, opened 2024-01-10
ibc host chain interface v2

avahowell:


cratelyn:
let me know when this is ready for another round of review @avahowell! :mag: :slightly_smiling_face:

avahowell:
this should be ready for another round of review now!

crates/core/component/ibc/src/component/host_interface.rs
```diff
@@ -0,0 +1,10 @@
+use async_trait::async_trait;
+use cnidarium::StateRead;
+
+#[async_trait]
+pub trait HostInterface {
+    async fn get_chain_id<S: StateRead>(state: S) -> anyhow::Result<String>;
+    async fn get_revision_number<S: StateRead>(state: S) -> anyhow::Result<u64>;
+    async fn get_block_height<S: StateRead>(state: S) -> anyhow::Result<u64>;
+    async fn get_block_timestamp<S: StateRead>(state: S) -> anyhow::Result<tendermint::Time>;
```
erwanor, line 9: Do we need to add host defined client verification parameters like `max_time_per_block`?

crates/core/app/src/penumbra_host_chain.rs
```diff
@@ -0,0 +1,50 @@
...
+
+use anyhow::Context;
+use async_trait::async_trait;
+use ibc_types::core::connection::ChainId;
+use penumbra_chain::{component::StateReadExt, state_key};
+use penumbra_ibc::component::HostInterface;
+use penumbra_proto::StateReadProto;
+use tendermint::Time;
+
+// ibc-async HostChainInterface for the Penumbra chain
```
cratelyn, line 11: i think this comment refers to a symbol that was renamed? it would be nice, but not blocking, to document this `PenumbraHost` type lightly, since it is publicly exported by the `penumbra-app` crate.

cratelyn (approved):
this looks good to me! i spotted a stray comment that i believe is stale, but the mechanical changes here all seem well-written. great work @avahowell!

crates/core/component/ibc/src/component/host_interface.rs
avahowell: I was thinking of doing a second pass to extract all chain-specific calls into this trait, currently this is only extracting the penumbra-chain calls

cratelyn (approved):

Issue #3565, opened 2024-01-03
Add strategies for negative testing of proofs


cronokirby:
After https://github.com/penumbra-zone/penumbra/issues/3564, in addition to testing that the in-rust logic matches the circuit for valid inputs, we should also augment our strategies to test invalid inputs as well, expanding the search space.

*Circuits that need to be changed*:
- [x] `Spend`
- [x] `Output`
- [x] `DelegatorVote`
- [ ] `UndelegateClaim`
- [x] `Convert`
- [x] `Swap`
- [x] `SwapClaim`
- [x] `NullifierDerivation`

PR #3600, opened 2024-01-11
test: add negative proptest strategies for first three circuits

redshiftzero:
This adds negative proptest strategies for the `SpendCircuit`, `DelegatorVoteCircuit`, and `NullifierDerivationCircuit` towards #3565


cronokirby (commented):
Perhaps we should have some kind of comment as to how the generators produce invalid values? Right now it's not entirely clear as to how this works, as the code looks very similar, which can be confusing.

A different approach towards invalid values would be take a valid value (via the existing generator), and then pick a random value to tweak, replacing with a random one.

PR #3569, opened 2024-01-04
Add provenance tracking by sender address to `SpendableNoteRecord`.

hdevalence:
This changes the view service protos to include an optional `return_address` for each note. This allows showing the provenance of individual notes.

Also included is a terrifically ugly fix for #3568.

PR #3604, opened 2024-01-12
Provenance storage fix

### LINKED ISSUE #3568: NoteSource changes broke `notes_by_sender` and hence the summoner
**Describe the bug**

In #3527 I killed the `NoteSource` "stuffed" transaction hash approach in favor of an enum. However, this had an unintended effect of breaking `notes_by_sender`:
```
            JOIN tx ON spendable_notes.source = tx.tx_hash
```
Because the `source` column in SQLite is the proto encoding, it no longer matches up with the `hash` column of the `tx` table.

Presumably this is why the summoner smoke test has been failing.

aubrika:
Closes #3568


conorsch:
After much heroic wrangling with CI by @cronokirby (see https://discord.com/channels/824484045370818580/1044394105498771506/1199092033667743754 for context), we expect green across the board here, after which we can merge.

conorsch:
This PR resolves https://github.com/penumbra-zone/penumbra/issues/3568, right? Do we still want #3569, or does this PR supersede that one?

aubrika:
This supersedes that one, I made a new branch so I could isolate my changes but I'll close the prior PR

PR #3605, opened 2024-01-12
test: add proptests of remaining circuits vs an in rust impl

### LINKED ISSUE #3564: Implement In-Rust Version of Proof Satisfaction

redshiftzero:
Closes #3564

### DIFF
--- a/crates/core/component/dex/src/swap/plaintext.rs
+++ b/crates/core/component/dex/src/swap/plaintext.rs
@@ -146,6 +146,20 @@ impl SwapPlaintext {
         }
     }
 
+    pub fn delta_1_value(&self) -> Value {
+        Value {
+            amount: self.delta_1_i,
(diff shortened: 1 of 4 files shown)


cronokirby (approved):
LGTM

PR #3606, opened 2024-01-12
typo: PRF takes input 1 (not 0) when deriving `nk`

kwantam:
Quick typo fix in docs (surmised based on [the official implementation](https://github.com/penumbra-zone/penumbra/blob/main/crates/core/keys/src/keys/spend.rs#L67-L75)).

### DIFF
--- a/docs/protocol/src/protocol/addresses_keys/spend_key.md
+++ b/docs/protocol/src/protocol/addresses_keys/spend_key.md
@@ -55,7 +55,7 @@ personalization `label`, key `key`, and input `input`.  Define
 integer in little-endian order.  Then
 ```
 ask = from_le_bytes(prf_expand("Penumbra_ExpndSd", spend_key_bytes, 0)) mod r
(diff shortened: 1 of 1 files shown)


redshiftzero (approved):
thanks!

kwantam:
One other minor comment on the [spending key derivation document](https://protocol.penumbra.zone/main/protocol/addresses_keys/spend_key.html):

The output of BIP-32 child key derivation is an integer mod n (i.e., a secp256k1 scalar), but the document treats this value as a 32-byte sequence, so it should explicitly define the way that it converts the integer to bytes. [The code](https://github.com/penumbra-zone/penumbra/blob/5adb302246b2f1ad61189fcedc483ecc2bc4fa0d/crates/core/keys/src/keys/spend.rs#L126) appears to treat the integer as big-endian bytes, which is consistent with BIP-32's `ser256` function. But it would be great to be explicit!

PR #3608, opened 2024-01-13
ibc: encode client states and consensus states as any

avahowell:


noot:
confirming that this works for my hermes fork! can open connections and channels now.

PR #3609, opened 2024-01-13
stake: index consensus set based on minimum stake requirement

erwanor:
This is a work-in-progress, here is a high-level overview of what this PR does:
- Add a `min_validator_stake` stake parameters specifying the minimum amount of stake required for a validator to be `Inactive` (default is `100penumbra`).
- Add a `Defined` state for validator that is the initial state for all post-genesis validators
- Keep a single source-of-truth for validator definitions in the JMT.
- Create a `consensus_set` nonverifiable index that tracks the identity keys of validators that have reached a minimum stake threshold
- Remove usage of the `validator_list_*` methods that returns (and collect) every validator definition.[^1]

This is the current validator state machine diagram:
```
///                         ┌──────────────────────────────────────┐     
///         ┌────────────┐  │  ┌─────────────────────┐             │     
///         │            │  ▼  ▼                     │             │     
///         ▼          ┌─┴──────┐               ┌────────┐         │     
///    ┌─────────┐     │        ├──────────────▶│        │    ┌────────┐ 
///    │ Defined │────▶│Inactive│    ┌──────────│ Active │───▶│ Jailed ││
///    └─────────┘     │        │    │          │        │    └────────┘│
///         │  ▲       └────────┘    │          └────────┘         │    │
///         │  │            ▲        │               │             │    │
///         │  │            │        │               │             │    │
///         │  │            │        │               │             │    │
///         │  │            │        │               │             │    │
///         │  │            │        ▼               │             │    │
///         │  │            │  ┌──────────┐          ▼             │    │
///         │  │            └──┤          │    ┌──────────┐        │    │
///         │  │               │ Disabled │    │Tombstoned│◀───────┘    │
///         │  └───────────────┤          │    └──────────┘             │
///         │                  └──────────┘                             │
///         │                     ▲     ▲                               │
///         └─────────────────────┘     └───────────────────────────────┘
```

[^1]: The idea here is to keep a state key that stores all validator definitions, but abrogate direct iteration on that store (because it could be really large).


hdevalence:
I haven't looked at the changes in detail yet, but I wanted to make one initial comment:

> Remove usage of the validator_list_* methods that returns (and collect) every validator definition.

This is a non-goal of the change, the point of introducing the Defined state is that we _don't_ want to make extensive changes to the staking component state machine we already have, we just want to gate access to it.

erwanor:
Thanks for giving it a look, let me try to workshop an alternative.

Currently, we have:

- A `state_key::validator` that maps identity keys to validator definitions
- `StateReadExt::validator_list` which is a `prefix` query that collects all the validator definitions
- `StateReadExt::validator_identity_list` which is a `prefix_keys` query that collect all the validator identity keys

I don't think we can escape adding another index because even if we gate access to `state_key::validator` for post-Defined validators, we still have to track `Defined` validator definitions some place else. If that's alright, then we could:
- Add a `state_key::defined_validator` that tracks defined validators.
- Gate access to `state_key::validator` on being post-Defined
- Keep the `validator_list` and `validator_list_identity` methods, perhaps turning them into streams. 

The entry point for transitions in-out of the precursor `Defined` state:
- Executing a `Delegate` action
- Processing undelegations

erwanor:
We discussed this and will go with the original approach because the alternative approach outlined in the previous comment is actually more complicated. To understand why, walk through the case where a validator is `Defined` and we process a delegation that triggers a state transition to `Inactive`. We need to pull the validator's definition from storage. If we store the `Defined` validators separately this actually creates more I/O than promoting the `IdentityKey` of the validator to an entry in a `consensus_set` index, because we have to remove the definition from that store and add it to the list of validators.

erwanor:
There are some "architectural" TODOs that I plan to tackle as part of #3615.

PR #3618, opened 2024-01-17
refactor(protobuf): authorizeAndBuild method

TalDerei:
We're eventually deprecating `witnessAndBuild` in favor of the `authorizeAndBuild` RPC method, and it unnecessarily requires an `authorizationData` field in order to be invoked. This modification of the protobuf definition is a necessary precursor for https://github.com/penumbra-zone/web/pull/369. 

cc @grod220 @turbocrime


TalDerei:
@conorsch how can I satisfy the CI here?

hdevalence:
Why are we getting rid of this RPC? Just because the web extension isn't planning to use it doesn't mean we should lock users out of a split custody setup...

TalDerei:
@hdevalence should we proceed with this change?

hdevalence:
Yep, let's get rid of the erroneous argument I added via copy paste error, but leave the RPC in place.

conorsch:
> how can I satisfy the CI here? the breaking change is intentional

You cannot, the CI is red because it's telling us "this is a breaking change." We know that, and we accept the breaking change and its implications, so what I'll do it override the merge control and get it into main anyway. However, merge is also blocked due to conflicts, so this needs to be freshly rebased on main (at least #3559, merged today, will have introduced conflicts). Can you rebase and mark ready for review?

conorsch:
Spoke with @TalDerei out of band, I'm going to rebase and get this to green (modulo the proto lint).

conorsch:
CI is good here. I'm holding off on merge so that @TalDerei can perform a bit more testing locally.

conorsch:
Tal confirmed good to go, so I'm merging to unblock further work in web world.

PR #3629, opened 2024-01-18
docs: update dev sql advice

conorsch:
This change started as a simple addition to the dev docs, adding a handy sqlite3 json-aware blob-decoding example, suggested by @TalDerei. While mucking around in the docs, I discovered some old `sqlx` info that AFAICT is dead documentation, given #2372.


aubrika:
Great catch! Sqlx was deprecated some time ago in favor of rusqlite and it looks like the docs update for that got missed. LGTM!

aubrika (approved):

PR #3631, opened 2024-01-19
refactor: rename pd upgrade -> migrate

### LINKED ISSUE #3578: Standardize on "migrate" terminology for performing upgrades
In the context of performing chain upgrades (https://github.com/penumbra-zone/penumbra/issues/1804), we use the word "upgrade" a lot, so much that it's become overloaded and therefore ambiguous. To reduce confusion, let's use "upgrade" to refer to:

1. the set of operations required to move a node from one version of the chain to another, e.g. `penumbra-testnet-rhea` to `penumbra-testnet-titan` chain id;
2. the act of updating a specific software component from one version to another, e.g. `pd` from `0.63.3` to `0.64.0`.

However, we currently _also_ use the term "upgrade" in the subcommand `pd upgrade`, which migrates local pd state to a format compatible with the future chain. Instead, let's call that action `pd migrate` and update documentation accordingly. Doing so will allow us to differentiate in documentation between the act of "upgrading pd" and "migrating state", both of which are necessary subtasks to be performed by operators within the greater whole of "performing a chain upgrade".

conorsch:
Closes #3578.


erwanor (approved):
LGTM on green CI

PR #3634, opened 2024-01-19
fee: fix regression in #3624

hdevalence:
In #3624, the default gas prices were changed to have nonzero values, in order to test client functionality with nonzero fees. However, we *don't* want to make this the default until we've ported all the client software over, and even then, we should make the gas price setting be part of the `AppParameter`s, rather than changing the default value. (We want the default value for `GasPrices` to align with the proto default value).

### DIFF
--- a/crates/core/component/fee/src/gas.rs
+++ b/crates/core/component/fee/src/gas.rs
@@ -45,7 +45,11 @@ impl Sum for Gas {
 }
 
 /// Expresses the price of each unit of gas in terms of the staking token.
-#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+///
+/// These prices have an implicit denominator of 1,000 relative to the base unit
+/// of the staking token, so gas price 1,000 times 1 unit of gas is 1 base unit
+/// of staking token.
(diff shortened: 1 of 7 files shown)


crates/core/component/fee/src/gas.rs
```diff
@@ -55,22 +59,16 @@ pub struct GasPrices {
...
-    /// Calculates the price based on given gas schedule. Applies an implicit
-    /// denominator of 1,000 to the gas prices.
-    pub fn price(&self, gas: &Gas) -> Amount {
+    /// Use these gas prices to calculate the fee for a given gas vector.
+    pub fn fee(&self, gas: &Gas) -> Amount {
         Amount::from(
-            self.block_space_price * (gas.block_space * 1_000) / 1_000
-                + self.compact_block_space_price * (gas.compact_block_space * 1_000) / 1_000
-                + self.verification_price * (gas.verification * 1_000) / 1_000
-                + self.execution_price * (gas.execution * 1_000) / 1_000,
```
hdevalence, line 73: Since the motivation for the denominator is to allow more precision in prices, we want to apply the multiple to the parameters we eventually use, rather than doing the multiplication here.

Issue #3617, opened 2024-01-17
penumbra: remove the chain component 


erwanor:
This is a draft issue, collecting requirements to remove the `penumbra-chain` component. In particular, the SCT's timestamp management needs to be fleshed out. Feedback welcome. 

**Is your feature request related to a problem? Please describe.**
The `penumbra_chain` crate is a catch-all layer of our stack. Though referred to as a "component" it does not actually implement `cnidarium_component::Component`, nor does it provide functionality that could not be integrated in actual `Component`s of the penumbra application.

Moreover, having a chain crate that is both a leaf (downstream of our application components) and a tie-together crate (abstract access to multiple components) complicates our application architecture.

So, we should try to remove it entirely. A good place to start figuring out requirements, is inspecting the chain crate's state keys, or by peeking at the extension traits:

```rust
▾ StateReadExt : interface
    chain_halt_count(&self) -> Result<u64> : method
    chain_params_updated(&self) -> bool : method
    check_chain_id(&self, provided: &str) -> Result<()> : method
    epoch(&self) -> Result<Epoch> : method
    epoch_by_height(&self, height: u64) -> Result<Epoch> : method
    epoch_ending_early(&self) -> bool : method
    get_block_height(&self) -> Result<u64> : method
    get_block_timestamp(&self) -> Result<Time> : method
    get_chain_id(&self) -> Result<String> : method
    get_chain_params(&self) -> Result<ChainParameters> : method
    get_current_epoch(&self) -> Result<Epoch> : method
    get_current_fmd_parameters(&self) -> Result<FmdParameters> : method
    get_epoch_duration(&self) -> Result<u64> : method
    get_epoch_for_height(&self, height: u64) -> Result<Epoch> : method
    get_previous_fmd_parameters(&self) -> Result<FmdParameters> : method
    get_revision_number(&self) -> Result<u64> : method
    is_chain_halted(&self, total_halt_count: u64) -> Result<bool> : method
    is_upgrade_height(&self) -> Result<bool> : method

▾ StateWriteExt : interface
    put_block_height(&mut self, height: u64) : method
    put_block_timestamp(&mut self, timestamp: Time) : method
    put_chain_params(&mut self, params: ChainParameters) : method
    put_current_fmd_parameters(&mut self, params: FmdParameters) : method
    put_epoch_by_height(&mut self, height: u64, epoch: Epoch) : method
    put_previous_fmd_parameters(&mut self, params: FmdParameters) : method
[...]

PR #3636, opened 2024-01-19
fee: move fixed initial gas prices into a chain parameter

hdevalence:
This updates the protos, and also adds an RPC to query the current parameters.

Follow up work:

- [ ] Ensure that proposals can enact gas price changes
- [ ] Add CLI flag to `pcli testnet generate` that allows specifying a gas price list (could be like `100,200,300,400`), similar to how we have args for epoch length etc
- [ ] Change smoke test (only) to use that CLI flag to enable fees in smoke test but not yet on `preview`


zbuc (approved):

erwanor:
Looks like this slipped (?), this is not a huge deal but we must ship this before 65 as it is protobreaking. Since this doesn't rebase cleanly and more proto changes are imminent as we progress through #3617, I suggest we tackle this as a last step before our next tagged release.

The material set of changes in this PR is not big, but since it the changeset affects the application parameters it is likely to create rebase work for other streams.

conorsch:
Discussed in planning meeting today. As @erwanor says, we need this for 65 (#3554). @hdevalence has offered to rebase to remove conflicts and get it in.

PR #3640, opened 2024-01-20
staking: remove ad-hoc fixpoint arithmetic

### LINKED ISSUE #3453: staking: use `U128x128` for rate calculations
We should switch to using `Amount`s and `penumbra_num::fixpoint` in the staking component.

erwanor:
Close #3453, this PR systematize the rate calculation process:
- In public interfaces, rates are represented using `Amount`s with an implicit $10^8$ scaling factor
- For internal rate calculations, we use `penumbra_fixnum::U128x128`

In addition to that, this PR:
- Fix reward amount calculations to use the ending epoch's base rate and simplify its signature
- Fix the distribution component's issuance budget to apply the staking token unit denom amount
- Lay the groundwork to implement the funding reward logic in a distinct component


erwanor:
AFK, but hypothesis for why the smoke test is failing, these fixes and the block-based issuance of the distribution component means that we have actual dynamic validator exchange rates. So the amount of delegation tokens obtained is not exactly equal to the amount of staking tokens delegated.

hdevalence:
The smoke test absolutely shouldn't be testing the exact amounts, its purpose is to exercise the commands end to end, not to unit test the arithmetic. We should just delete those checks and have unit tests.

erwanor:
Ugh. That was it, merging to not block work on funding streams and modularization.

PR #3647, opened 2024-01-22
Add fee tiers to pcli

### LINKED ISSUE #3153: Add tiers for gas payments to pcli
Split from discussion in #3133

Branching off of https://github.com/penumbra-zone/penumbra/pull/3133/files#r1344827368, here's a further rabbithole to go down: how do gas prices affect transaction linkability?

Historically, one of the reasons that Zcash was resistant to having a fee market was that the fee is necessarily revealed in the clear, and so it's a vector to leak information about the client. For instance, if different clients have different logic for fee selection, it might be possible to infer which client was used to create a specific transaction, or to link two transactions by the same user (if that user overrode the fee amount in a unique way shared only among those transactions).

We're exposed to this because of the ability to set a custom fee amount. However, we do need to price resources, to avoid the DoS problems later experienced with Zcash (and with the perspective that a fee market will exist, regardless of whether we'd like it to or not), and so we should take a harm reduction approach.

One way that we could reduce the anonymity risk of setting a custom fee amount would be to use the fact that we have static gas amounts, and to have a few preset values for "low" "medium" "high" that would have different multiples of the base price (e.g., low = 1.05x, medium = 1.3x, high = 2x), rather than an arbitrary setting.

However, while we can quantize the multiple into a few settings, this still allows an observer who sees the transaction to infer what prices were used by that user's client, and hence learn information about when they started planning the transaction. Consider, for instance, the "randomized TWAP" use case, where a user wants to perform a swap over some time period without revealing their complete flow, by splitting their trade into sub-trades with randomized sub-amounts. The TransactionPlan mechanism allows all sub-trades to be planned and authorized by the user at once, storing a collection of pre-authorized transaction plans that can be proved just before submission so that the time of authorization isn't revealed by the transaction contents. But auto-adjusting gas prices potentially messes this up, since the plan has to include what fee it will pay, and computing the fee requires using a specific choice of gas prices.

Perhaps this suggests that we should try to change prices less often than every block, to have longer chunks of time in which gas prices don't change. On the other hand, that messes up the resource pricing goal. Another solution could be to apply some kind of quantization after computnig the fees, so that if gas prices are slightly different the fee would still be the same, although this also seems potentially problematic to define in advance (how to we know how much to quantize without knowing what the prices will be?)

Unfortunately, adding noise only to the RTWAP transactions doesn't really help, because then they're distinguishable from other transactions that would use quantized fee amounts. We could make all transactions have noisy fee levels, but this is also tricky, because that opens up a large surface area for implementation fingerprinting (what if different clients add noise in different ways that turn out to be statistically distinguishable?).

One possibility that might be workable is to stick with preset tiers for gas adjustments (e.g., "low" = 1.05x base, "medium" = 1.3x base, "high" = 2x base), but instead of using fixed multipliers, define specific distributions for clients to draw from to choose a multiplier. So setting "medium" would cause your client to sample a random value near 1.3, using the same distribution as other people's clients.

The hope would then be that slight variation in the gas prices used to compute the fee amount would hide in the noise of the randomized adjustment. This wouldn't be the case if prices changed very rapidly, but if they change only a little bit, it might work.

For a first pass, though, I think we should stick to fixed tiers and figure out randomization later.

### LINKED ISSUE #3369: Tiers for Fees
Since fees may change between transaction creation and submission, we should support tiers (low, medium, high) of fee payments so users can have a higher likelihood of execution. This could also eventually apply some noise (via server-side construction of fee tier distributions?) to avoid fingerprinting based on client configuration, as fees are public. 

This is not strictly necessary pre-mainnet, as it is only a client-side change. If automatic gas pricing does not land pre-mainnet, there is no fee market, so all client fees should be 0 anyways. If automatic gas pricing does land, then we can quickly implement simple tier multipliers client-side at the same time.

_What we need to implement_

- [x] https://github.com/penumbra-zone/penumbra/issues/3153 Completely client-side.
- [x] #3648

zbuc:
Closes #3153


cratelyn (approved):
:heavy_check_mark: looks good to me! nice work @zbuc.

proto/penumbra/penumbra/core/component/fee/v1alpha1/fee.proto
```diff
@@ -24,6 +24,18 @@ message GasPrices {
   uint64 execution_price = 4;
 }
 
+message FeeTier {
+  // The tier for the fee.
+  enum Tier {
+    TIER_LOW = 0;
```
cratelyn, line 30: it looks like there is a lint here that we should address before merging.

cratelyn (commented):
noting one failing action we should fix before merging...

Issue #3627, opened 2024-01-18
Clean up `pd` auto-https code


hdevalence:
> In the ~year since this code was written, there may be better options. https://docs.rs/tokio-rustls-acme/latest/tokio_rustls_acme/ seems promising

_Originally posted by @hdevalence in https://github.com/penumbra-zone/penumbra/pull/3522#discussion_r1429038043_

#3522, the initial experiment for serving HTTP content on the GRPC endpoint, is blocked by the fact that we would then need to use Axum directly, but our current auto-https support only works with Tonic specifically (which seems wrong, it's "all HTTP")

hdevalence:
We'll probably need to rebuild #3522 anyways, so if we could clean up the auto-https support to get something that works directly with Axum first, we would have an easier time rebuilding that work.

cratelyn:
i going to self-assign this. @hdevalence and i spent some time talking today, getting me up to speed with the motivation here. once i have made some more headway on the mock consensus engine, #3588, i will pivot over to work on this.

cratelyn:
starting work on this issue this week! i've also opened #3646 to track the original motivation (a self-hosted web interface) in #3646.

PR #3650, opened 2024-01-23
🩹 pd: fix auto-https default grpc address

cratelyn:
while working on #3627, i found that a conditional hinges on the wrong option.

this changes pd to follow what the comments and documentation state as the
intent of this code: change the default grpc address based on whether we
have an auto-managed https domain.

a second commit refactors the control flow for the default address. the `parse`
calls only refer to hardcoded values defined in the source. let's define our
defaults as constants, so that we do not have error-handling logic that only
refers to what ought to be known-good addresses.


conorsch:
Nice, thanks for fixing!

PR #3651, opened 2024-01-23
staking: validator state transition fix

erwanor:
This PR fixes an issue with a missing transition (`Active` -> `Defined`) and adds context for each transition case. With that said, writing this code makes me think that we might be better off by limiting the amount of possible transitions, removing `Active -> Defined` and `Jailed -> Defined` transition, and use end-epoch processing to "sweep" inactive validators into the precursor `Defined` state.

Since I am in the middle of an assurance blitz for #3640, I will abstain from making more changes. But this is currently top-of-mind for me. Moreover, manual testing seems to surface issues with Unbonded/Unbonding (observed on `v0.64.0` as well) so I plan to look into that as well.


erwanor:
Merging optimistically, because I need this to do more assurance of #3640.

PR #3656, opened 2024-01-24
ci: improve smoke test logging

erwanor:
This PR change the smoke test workflow to independently capture and display logs produced by `cometbft`, `pd`, and the `pcli`/`pclientd` integration tests. This makes it easy to visualize and isolate the source of a smoke test failure.

It also sets a baseline of `info` level debugging for tracing statements, and `debug` level for `network_integration` tests.

Finally, it offers the opportunity for users to run the integration tests with maximum logging level by detecting if debug mode is enabled for this action run (see screenshot).

Screenshot 2024-01-24 at 11 37 50 AM

### DIFF
--- a/.github/workflows/smoke.yml
+++ b/.github/workflows/smoke.yml
@@ -30,5 +30,22 @@ jobs:
         run: |
           export PATH="$HOME/bin:$PATH"
           ./deployments/scripts/smoke-test.sh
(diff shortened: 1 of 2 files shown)


conorsch (approved):
Big fan, thanks for taking the time!

PR #3657, opened 2024-01-24
ci: fix smoke test

erwanor:
I think I introduced an indentation error, this should fix it!

### DIFF
--- a/.github/workflows/smoke.yml
+++ b/.github/workflows/smoke.yml
@@ -22,30 +22,24 @@ jobs:
           toolchain: stable
       - name: Load rust cache
         uses: astriaorg/buildjet-rust-cache@v2.5.1
-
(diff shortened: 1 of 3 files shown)


conorsch:
Thanks @erwanor, I'll get this one to green. Appreciate the improvements to the smoke test! I'll make sure it passes here (oddly https://github.com/penumbra-zone/penumbra/pull/3656 did indeed pass) and also on main.

conorsch:
Passing: https://github.com/penumbra-zone/penumbra/actions/runs/7644866555/job/20830140204 Also confirmed it passes locally. I suspect we might need to fuss with the auto-log-output functionality for stdout/stderr, but that'll only be obvious once we actually have a failing test. I'm going to merge this to unbreak the test suite for other PRs.

PR #3658, opened 2024-01-24
[wasm] Compute position_id

Valentine1898:
Related to https://github.com/penumbra-zone/web/pull/367 
We need a function to compute `PositionId` since `PositionId` is not stored in `Position` (most likely to optimize on-chain storage). 
`PositionId` is needed as a unique position identifier and is also used to update `PositionState` ( see usage https://github.com/penumbra-zone/web/pull/367/files#diff-ccb0493c986b958cf859e773170457e5ff59a4a2e0d742698492247f6ae872da. )

discussion 
https://discord.com/channels/824484045370818580/938328380917559326/1199461586927112283

### DIFF
--- a/crates/wasm/src/dex.rs
+++ b/crates/wasm/src/dex.rs
@@ -1,0 +1,17 @@
+use crate::utils;
+use penumbra_dex::lp::position::Position;
+use serde_wasm_bindgen::Error;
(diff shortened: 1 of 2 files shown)


grod220 (approved):
I'll deploy an updated preview wasm npm package so it unblocks the other work

PR #3660, opened 2024-01-24
distributions: add dep to `penumbra_asset`

erwanor:
Adds `penumbra-asset` to the distributions component deps. It's not totally clear why CI didn't catch that, I'll circle back when I'm done with more urgent tasks.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5200,6 +5200,7 @@ dependencies = [
  "async-trait",
  "cnidarium",
  "cnidarium-component",
+ "penumbra-asset",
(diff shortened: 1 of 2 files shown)


erwanor:
... I'm not sure what's going on, without this change, my local repo does not compile.

But the remote was fine without (previous PR was full green), and crashes with it.

conorsch:
Oops, great minds think alike. Does https://github.com/penumbra-zone/penumbra/pull/3659 satisfy?

erwanor:
Thanks, yes it does

PR #3667, opened 2024-01-25
Display full address in summoner interface

### LINKED ISSUE #3626: Change summoner web interface to have full addresses

Many users are confused when they check the summoning page and see a different address. If the address was shown, they could copy it and do pcli v address to check that it was theirs. But only the short form is shown, so they can't.

Change the column ordering to

Number
Date
Hash
Full Address

cronokirby:
Closes #3626.

(there's still some debate in that issue, but that's orthogonal to writing code to do what that feature request asks)


hdevalence (changes requested):
IMO it's important to change the column ordering so that the address goes last:

Number
Date
Hash
Full Address

Now that it's longer this means the overflow will just run off and clip / scroll horizontally, so it's like a short form (until you want to look at it)

hdevalence (approved):

PR #3669, opened 2024-01-25
Add execution circuit breaker

### LINKED ISSUE #3303: dex: implement an execution circuit-breaker
In the DEX design, we put a good amount of effort into ensuring we always make forward progress. But forward progress could be very slow. We don’t think that it will have pathological behavior – but it’s possible we could have missed something that then turns into a DoS vector. Similarly to the #2986, it could be useful to have some kind of “execution breaker” that will bail out with incomplete fill after some execution limit. This means that if the DEX engine does have a DoS vector, it does not affect users of the shielded pool.

We need to implement a DEX-engine specific metering system that tracks the number of operations performed during routing execution e.g. filling against a position, or assembling a `FrontierTx` and concludes execution if a threshold is exceeded. The threshold could be a DEX chain parameter.

### LINKED ISSUE #3374: dex: implement an execution circuit breaker
In the DEX design, we put a good amount of effort into ensuring we always make forward progress. But forward progress could be very slow. We don’t think that it will have pathological behavior – but it’s possible we could have missed something that then turns into a DoS vector. Similarly to the circuit breaker, it could be useful to have some kind of “execution breaker” that will bail out with incomplete fill after some execution limit. This means that if the DEX engine does have a DoS vector, it does not affect users of the shielded pool.

This feature is important to have for mainnet because it guards against unknown DoS vectors in the DEX engine.

_What we need to implement_

- [ ] Some kind of tracking of how many “operations” we have done while executing the DEX functionality, in a way that we can bail out after the limit is hit.

zbuc:
This adds a circuit break mechanism to the DEX that bounds the maximum number of path searches and position executions for any given batch swap during a block, to prevent potentially unbounded looping.

Closes #3303

### DIFF
--- a/crates/core/component/dex/src/circuit_breaker/execution.rs
+++ b/crates/core/component/dex/src/circuit_breaker/execution.rs
@@ -1,0 +1,36 @@
+const MAX_PATH_SEARCHES: u32 = 64;
+const MAX_EXECUTIONS: u32 = 64;
+
+/// Holds the state of the execution circuit breaker.
+/// Responsible for managing the conditions of halting execution of
+/// a single batch swap. All execution circuit breaker triggers are
+/// non-fatal and will allow the swap to be partially fulfilled up
+/// to the search and execution limits managed by the circuit breaker.
+///
+/// The circuit breaker ensures the swap will not use unbounded time complexity.
+struct ExecutionCircuitBreaker {
+    /// The maximum number of times to perform path searches before stopping.
+    pub max_path_searches: u32,
+    /// The number of times path searches have been performed.
+    pub current_path_searches: u32,
+    /// The maximum number of times to execute against liquidity positions before stopping.
+    pub max_executions: u32,
+    /// The number of times liquidity positions have been executed against.
+    pub current_executions: u32,
+}
+
+impl ExecutionCircuitBreaker {
+    pub fn new() -> Self {
+        Self {
+            max_path_searches: MAX_PATH_SEARCHES,
+            current_path_searches: 0,
+            max_executions: MAX_EXECUTIONS,
+            current_executions: 0,
+        }
+    }
+
+    pub fn exceeded_limits(&self) -> bool {
+        self.current_path_searches > self.max_path_searches
+            || self.current_executions > self.max_executions
+    }
+}
--- a/crates/core/component/dex/src/circuit_breaker/execution.rs
+++ b/crates/core/component/dex/src/circuit_breaker/execution.rs
@@ -8,6 +8,7 @@ const MAX_EXECUTIONS: u32 = 64;
 /// to the search and execution limits managed by the circuit breaker.
 ///
 /// The circuit breaker ensures the swap will not use unbounded time complexity.
+#[derive(Debug, Clone)]
 pub(crate) struct ExecutionCircuitBreaker {
     /// The maximum number of times to perform path searches before stopping.
     pub max_path_searches: u32,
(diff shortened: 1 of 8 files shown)


erwanor:
Will do a review on the concrete implementation, but first high-level comment: this approach is clean and I think that it follows what @hdevalence outlined on discord: a top level counter that increments function calls to `path_search` and `fill_route`.

I think my preference would be for more intrusive metering, where we reach deep into `fill_route_inner` and track each position we fill against. The reason for this is that a bug in the constraint resolution logic, or in the `Frontier::load` code, will cause a chain hang (similar to Testnet 53's) and infinite execution that the current approach would miss. With that said, I will review the approach that we have decided on here.

crates/core/component/dex/src/component/router/route_and_fill.rs
```diff
@@ -145,12 +145,21 @@ pub trait RouteAndFill: StateWrite + Sized {
 
         let max_delta_1: Amount = MAX_RESERVE_AMOUNT.into();
 
+        let mut execution_circuit_breaker = ExecutionCircuitBreaker::new();
```
erwanor, line 148: If we make `route_and_fill` take an `ExecutionCircuitBreaker`, and make the upper bound for path search and fills a parameter of that struct, this will let us craft test a variety of test vectors for the circuit breaking code. We can keep the constants for the production code.

crates/core/component/dex/src/circuit_breaker/execution.rs
```diff
@@ -0,0 +1,36 @@
+const MAX_PATH_SEARCHES: u32 = 64;
```
erwanor, line 1: Can we make this a parameter of `ExecutionCircuitBreaker`? Keeping the constant to initialize the code

crates/core/component/dex/src/circuit_breaker/value.rs
```diff
@@ -0,0 +1,69 @@
+use std::collections::BTreeMap;
+
+use penumbra_asset::{asset, Value};
+use penumbra_num::Amount;
+use penumbra_proto::{core::component::dex::v1alpha1 as pb, DomainType, TypeUrl};
+
+#[derive(Debug, Clone, Default)]
+pub struct AssetTallies {
```
erwanor, line 8: This isn't used AFAICT?

crates/core/component/dex/src/circuit_breaker/execution.rs
```diff
@@ -0,0 +1,36 @@
...
+    /// The maximum number of times to execute against liquidity positions before stopping.
+    pub max_executions: u32,
+    /// The number of times liquidity positions have been executed against.
+    pub current_executions: u32,
+}
+
+impl ExecutionCircuitBreaker {
+    pub fn new() -> Self {
+        Self {
+            max_path_searches: MAX_PATH_SEARCHES,
```
erwanor, line 25: For some reason my brain didn't parse that we were already using the constant to initialize the struct.

crates/core/component/dex/src/circuit_breaker/execution.rs
erwanor: This is already the case

crates/core/component/dex/src/component/router/route_and_fill.rs
erwanor: That's already the case. I guess I would ditch the construct hiding those details though.

erwanor (commented):
Resolved some comments

crates/core/component/dex/src/circuit_breaker/value.rs
zbuc: This is preliminary for the value circuit breaker

PR #3670, opened 2024-01-25
Update wasm readme

grod220:
- Updated to include more information on publishing
- Updated package.json deps
- Ran `prettier` formatting


conorsch (approved):
Crystal clear, thanks for taking the time, @grod220!

PR #3677, opened 2024-01-26
app: fold genesis param init into their respective component

### LINKED ISSUE #3117: Remove initialization code from `App::init_chain`
Captured from https://github.com/penumbra-zone/penumbra/pull/3086

@hdevalence: I think ideally we shouldn't have the `App` do any initialization at all, and only have it call into specific components' `init_chain` methods. In order to do that, we'd need to have a clear assignment of responsibility of functionality per component, which we don't currently have (but should).

For instance, why is the epoch being written directly by the App, if the epoch handling is intertwined with the SCT? Why are the FMD parameters set by the App rather than by the shielded pool? etc.

erwanor:
Close #3117, prelude to #3617


erwanor:
Merging optimistically to proceed with the chain component work

Issue #3676, opened 2024-01-26
decaf377: propagate changes to circuit stack


TalDerei:

aubrika:
This was moved into the current sprint outside of sprint planning - can someone account for this? @cronokirby @TalDerei

aubrika:
That is incorrect - it has not been estimated, and was not prioritized an explicitly added during sprint planning, so it should not be being worked on without explicit admission into the current sprint iteration. Other work in the Sprint 0 milestone should take priority, unless myself and @hdevalence explicitly sign off on adding a ticket.

aubrika:
Moved back to backlog for sprint load reasons

PR #3678, opened 2024-01-26
[DO NOT MERGE] decaf version bumps decaf

TalDerei:
Experimenting with #3676


TalDerei:
continuing debugging

TalDerei:
Currently, proof verification is still failing. There are a few noteworthy observations to mention:

- Old backend (`v0.5`) generates and verifies proofs with actual prover / verifier keys and [dummy keys](https://github.com/penumbra-zone/penumbra/blob/59cdd7552dd4e50aea948328a79bc034c599fa1e/crates/crypto/proof-params/src/traits.rs#L49-L50),
- New backend (`v0.8`) generates and verifies proofs with dummy keys, but not with actual prover / verifier keys,
- The representation of the dummy keys, for instance for a spend circuit, for some reason differs between `v0.5` and `v0.8`. 

Interestingly, in `v0.8`, the `blinding factors, pk, and public and private` [inputs](https://github.com/penumbra-zone/penumbra/blob/c1bb046454b01d99c7253a4b750edfb14406ed6e/crates/core/component/shielded-pool/src/spend/proof.rs#L278) in `spend_proof_parameters_vs_current_spend_circuit` match up against the old backend, but proving generates a different proof result than expected and fails proof verification for spend proofs. Simply swapping the actual proving keys for the dummy keys succeeds. 

As an aside, I generated a proof with `v0.5`, and verified the proof string directly in `v0.8` which passes the verification checks with the actual verification key. This suggests there may be an issue instead with how we're serializing specifically our proving keys. 

@redshiftzero I created a minimal crate `old-backend` with calls to `generate_constraint_matrices` and `generate_test_parameters` which don't complain, which suggests our new backend is properly creating R1CS matrices. We should expand the tests in attempting to resolve the failing proofs.

TalDerei:
Closing in favor of updated #3806

PR #3663, opened 2024-01-25
penumbra: create a funding component

erwanor:
This PR introduce scaffolding for a new `funding` component responsible for disbursing staking rewards and eventually proposer tips.

PR #3682, opened 2024-01-26
funding: process staking streams at end of epoch

erwanor:
Follow-up to #3663, we now process validator funding streams in the funding component.


erwanor:
Note for visibility: this PR halts the chain if an inflation invariant is breached.

PR #3689, opened 2024-01-27
Implement view service restarts in summonerd

cronokirby:
This uses the database being stale as a source of truth as to a lousy view service

### DIFF
--- a/crates/view/src/service.rs
+++ b/crates/view/src/service.rs
@@ -9,10 +9,16 @@ use ark_std::UniformRand;
 use async_stream::try_stream;
 use camino::Utf8Path;
 use decaf377::Fq;
-use futures::stream::{StreamExt, TryStreamExt};
+use futures::{
+    stream::{StreamExt, TryStreamExt},
+    FutureExt,
+};
 use rand::Rng;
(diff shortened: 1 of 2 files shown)


crates/view/src/service.rs
```diff
@@ -107,17 +115,30 @@ impl ViewService {
...
+    /// Close this view service, stopping the background syncing task.
+    ///
+    /// This will have a spoooky action at a distance in that other cloned instances of this
+    /// service sharing the syncing task will also be silently broken, so be careful when using
+    /// this method.
+    ///
+    /// Also this is sort of a stop-gap for a more robust view service where this method won't be
+    /// necessary, so one should exercise skepticism that this method even exists.
+    pub async fn abort(&self) {
+        self.worker_handle.abort();
```
erwanor, line 139: Random observation, but if you want some stronger guarantees about the timing of the join handle cancellation, you can immediately await the worker handle and that will make this method return when those resources are actually released. It can be useful for testing.

hdevalence:
closing in favor of view-internal logic / health checks

conorsch:
Refs #3732.

PR #3690, opened 2024-01-29
Implement value circuit breaker

### LINKED ISSUE #2986: dex: implement a value circuit breaker
The DEX should never be able to output more assets than were deposited in positions. This invariant is often referred to as the "asset conservation law" in the literature. We should create a circuit breaker that halts the chain if an asset excess is detected. To do this, we should create a DEX account that tracks aggregate amounts of assets present in liquidity positions. This can be done as we write position state back to storage.

- [ ] create a dex account tracking aggregate amounts
- [ ] update the tallies when writing positions to state
- [ ] read the BSOD and check that the invariant is upheld at the end of the block

zbuc:
This adds a value circuit breaker to the `PositionManager` logic.

Whenever a position is stored, the total reserve amounts (across _all_ positions) are tracked by the `ValueCircuitBreaker`. If after storing a position the reserve balance ever becomes negative ("required"), this implies that more value was taken out of positions than was stored in them, and an error is bubbled up.

Closes #2986

### DIFF
--- a/crates/core/component/dex/src/circuit_breaker/value.rs
+++ b/crates/core/component/dex/src/circuit_breaker/value.rs
@@ -1,4 +1,5 @@
-use penumbra_asset::Balance;
+use penumbra_asset::{asset::Id, Balance, Value};
+use penumbra_num::Amount;
 use serde::{Deserialize, Serialize};
 
 #[derive(Debug, Clone, Serialize, Deserialize, Default)]
@@ -11,23 +12,38 @@ impl ValueCircuitBreaker {
         self.balance += balance;
     }
 
-    pub fn check(&self) -> anyhow::Result<()> {
+    pub fn assert_balance_invariant(&self) {
         // No assets should ever be "required" by the circuit breaker's
         // internal balance tracking, only "provided".
         if let Some(r) = self.balance.required().next() {
-            return Err(anyhow::anyhow!(
+            assert!(
+                false,
                 "balance for asset {} is negative: -{}",
-                r.asset_id,
-                r.amount
-            ));
+                r.asset_id, r.amount
+            );
         }
+    }
 
-        Ok(())
+    pub fn available(&self, asset_id: Id) -> Value {
+        self.balance
+            .provided()
+            .find(|b| b.asset_id == asset_id)
+            .unwrap_or(Value {
+                asset_id,
+                amount: Amount::from(0u64),
+            })
     }
 }
 
 #[cfg(test)]
 mod tests {
+    use std::sync::Arc;
+
+    use crate::{
+        component::{router::limit_buy, tests::TempStorageExt, PositionManager as _},
+        DirectedUnitPair,
+    };
+    use cnidarium::{ArcStateDeltaExt as _, StateDelta, TempStorage};
     use penumbra_asset::{asset, Value};
     use penumbra_num::Amount;
     use rand_core::OsRng;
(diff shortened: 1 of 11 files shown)


crates/core/component/dex/src/component/position_manager.rs
```diff
@@ -455,5 +460,139 @@ pub(super) trait Inner: StateWrite {
...
+            ),
+            None => ValueCircuitBreaker::default(),
+        };
+
+        // Add the delta to the value circuit breaker for assets A and B.
+        value_circuit_breaker.tally(delta_a);
+        value_circuit_breaker.tally(delta_b);
+
+        // Confirm that the value circuit breaker is still within the limits.
+        value_circuit_breaker.check()?;
```
erwanor, line 586: It's late so maybe I'm missing something, but my first read of this PR is that the balances are aggregated correctly, but the circuit breaker check happens at the wrong resolution. We do it every time we update a position, but that means we can't factor in the "outflow" (the swap output) and so it will detect benign state updates at value breaking.

If I try to run through an example:
Let's take a position $P$ with $p,q=1$, and reserves $(R_1, R_2) = (100, 0)$, suppose that the position is updated and "flips" completely on the other side such that we have $(R_1^*, R_2^*) = (0, 100)$.

By applying the above logic, we compute the aggregated balance of the value circuit breaker (`(new_a - old_a, new_b - old_b)`):

$T_A = 0 - 100$ of asset 1

$T_B = 100 - 0$ of asset 2.

So the circuit breaker check will fail.

erwanor: I could be missing something, but it seems to me we want to perform checks at two levels:
- when we update a position
- when we collect the batched swap outputs from the dex engine

erwanor (commented):
I have done a preliminary pass, but it will be easier for me to dig in when we resolve the serialization shenanigans. It's late but something seems off about the resolution of the balance check; will do a second pass tomorrow am.

crates/core/component/dex/src/component/position_manager.rs
zbuc: > By applying the above logic, we compute the aggregated balance of the value circuit breaker ((new_a - old_a, new_b - old_b)):

I don't think this matches how it's implemented, unless I am misunderstanding.

What I believe is implemented (which should be validated with integration tests) is that:

* When the position is initially stored with reserves `(100, 0)`, `delta_a = +100 asset_1` and `delta_b = 0 asset_2`. 
* These are tallied to the circuit breaker and thus it has balances of `asset_1: 100, asset_2: 0`
* When the position is filled against and the reserves flip, the `(new_a - old_a, new_b - old_b)` are assigned to `(delta_a, delta_b)`. `delta_a = -100 asset_1` and `delta_b = +100 asset_2`, and then we add `delta_a` to the aggregate balance (by calling `value_circuit_breaker.tally(delta_a)`) which should result in a balance of `asset_1: 0, asset_2: 100` being tracked

erwanor: Ah I see, that makes total sense

zbuc: As for where we perform checks, I _think_ this is okay -- at the time the batch swap outputs are minted, the positions that were filled against (should) have already all been updated to reflect the in/outflows. I suppose this could miss a case where `put_position` was not called on positions that were filled against, but I don't think it should be possible given the way the dex is implemented (unless we fail to store the updated positions to storage or something??)

crates/core/component/dex/src/circuit_breaker/value.rs
```diff
@@ -1,69 +1,103 @@
...
+        DirectedTradingPair,
+    };
+
+    use super::*;
+
+    // Ideally the update_position_aggregate_value in the PositionManager would be used
+    // but this is simpler for a quick unit test.
+
+    #[test]
+    fn value_circuit_breaker() {
```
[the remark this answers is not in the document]
zbuc: Good idea

crates/core/component/dex/src/component/position_manager.rs
```diff
@@ -455,5 +460,139 @@ pub(super) trait Inner: StateWrite {
...
+            .expect("able to retrieve value circuit breaker from nonverifiable storage")
+        {
+            Some(bytes) => bincode::deserialize(&bytes).expect(
+                "able to deserialize stored value circuit breaker from nonverifiable storage",
+            ),
+            None => ValueCircuitBreaker::default(),
+        };
+
+        // Add the delta to the value circuit breaker for assets A and B.
+        value_circuit_breaker.tally(delta_a);
```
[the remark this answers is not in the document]
zbuc: Hmm I meant delta in the mathematical sense but you have a point. Maybe `difference_a` is better?

crates/core/component/dex/src/component/position_manager.rs
zbuc: I think we can add one additional check to `handle_batch_swaps`:

* At the beginning of the method, fetch the aggregate balances for asset_1/2 prior to calling `route_and_fill`
* At the end of the method perform a check that the outflows do not exceed the aggregate balance for either asset

erwanor: That sounds good!

crates/core/component/dex/src/component/position_manager.rs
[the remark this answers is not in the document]
erwanor: Since it is the change in position balance after accounting for all increase/decrease, maybe `net_change_for_a`? difference also works

PR #3691, opened 2024-01-29
penumbra-ibc: ⏎  invert early return condition

cratelyn:
this diff is best viewed ignoring whitespace; use `-w`, or the "_hide whitespace_" option in github (the gear icon in the "_files changed_" tab).

this inverts our early return, so that we do not need to wrap the whole function body within a conditional block. this is a pure refactor that should have no visible changes to behavior.


erwanor (approved):

cratelyn:
re, lint failing, that is happening upstream on `main`. see https://github.com/penumbra-zone/penumbra/actions/runs/7698514268/job/20977960857

cratelyn:
lint was kindly fixed in #3692, i am going to merge this now!

PR #3694, opened 2024-01-29
staking: fix unbonding logic

### LINKED ISSUE #3683: staking: fix validator unbonding period tracking
When a validator bonding state is `Unbonded`, we should allow undelegation claim to be processed immediately.

erwanor:
Close #3683. This is a first pass at reworking the unbonding delay logic.

What it does:
- Refactor the unbonding end epoch methods
- Allow undelegation claims if the validator is unbonded.
- Allow claims if we have reached the unbonding epoch target (vs. waiting one more epoch; IMO - this should be switched to a time-based mechanism)


crates/core/component/stake/src/validator/bonding.rs
```diff
@@ -71,7 +75,9 @@ impl TryFrom<pb::BondingState> for State {
                 } else {
                     anyhow::bail!("unbonding epoch should be set for unbonding state")
                 };
-                Ok(State::Unbonding { unbonding_epoch })
+                Ok(State::Unbonding {
+                    unbonds_at_epoch: unbonding_epoch,
+                })
```
cratelyn, line 80: i think these expressions could be updated to follow the `unbonds_at_epoch` renaming, keeping the shorthand syntax from before. not a blocker.

crates/core/component/stake/src/validator/bonding.rs
```diff
@@ -44,12 +46,14 @@ impl From<State> for pb::BondingState {
             state: match v {
                 State::Bonded => pb::bonding_state::BondingStateEnum::Bonded as i32,
                 State::Unbonded => pb::bonding_state::BondingStateEnum::Unbonded as i32,
-                State::Unbonding { unbonding_epoch: _ } => {
-                    pb::bonding_state::BondingStateEnum::Unbonding as i32
-                }
+                State::Unbonding {
+                    unbonds_at_epoch: _,
+                } => pb::bonding_state::BondingStateEnum::Unbonding as i32,
```
cratelyn, line 51:
```suggestion
                State::Unbonding {
                    ..
                } => pb::bonding_state::BondingStateEnum::Unbonding as i32,
```

cosmetic. not a blocker.

crates/core/component/stake/src/validator/bonding.rs
```diff
@@ -19,15 +19,17 @@ pub enum State {
     /// The validator has been removed from the active set.
     ///
     /// All delegations to the validator will be unbonded at `unbonding_epoch`.
```
cratelyn, line 21: this comment should be updated!

PR #3679, opened 2024-01-26
pd: 📬 create tonic router in penumbra-app

cratelyn:
this addresses a todo comment found in the `pd` entrypoint. we are now using `tonic` 0.10 so we can have the `penumbra-app` crate assemble a `Routes`, sparing us the recitation of every service.

PR #3696, opened 2024-01-29
pd: ♎ move node entrypoint into library

### LINKED ISSUE #3695: pd: 🐣 move entrypoint logic into `app` library
in #3588, we are tracking work on a mock consensus engine. the benefit of this will be an improved ability to write in-memory `#[test]` functions of our `pd` daemon, rather than the smoke-tests. see #3588 for additional information about the motivation for that work.

on the flipside, we should also make some adjustments to facilitate writing tests against the `pd` server.

first, we should move the logic responsible for running a `RootCommand` from the `main.rs` binary artifact, to within the `pd` library. this follows the advice from Ch. 11 of The Rust Programming Language book:

> [**Integration Tests for Binary Crates**](https://doc.rust-lang.org/book/ch11-03-test-organization.html#integration-tests-for-binary-crates)
>
> If our project is a binary crate that only contains a `src/main.rs` file and doesn’t have a `src/lib.rs` file, we can’t create integration tests in the tests directory and bring functions defined in the `src/main.rs` file into scope with a `use` statement. Only library crates expose functions that other crates can use; binary crates are meant to be run on their own.
>
> This is one of the reasons Rust projects that provide a binary have a straightforward `src/main.rs` file that calls logic that lives in the `src/lib.rs` file. Using that structure, integration tests can test the library crate with use to make the important functionality available. If the important functionality works, the small amount of code in the `src/main.rs` file will work as well, and that small amount of code doesn’t need to be tested.

#3672 made some initial headway on this. let's move the rest of our (a) tracing subscriber initialization, and (b) command logic, into a function somewhere in the `pd` library so that we can exercise this logic in integration tests someday.

* #3672
* #3588

cratelyn:
fixes #3695.

:construction: TODO: write description


cratelyn:
this work was based on #3679, which caused a regression. see #3697. that PR was reverted, so i am going to close this rather than address conflicts here.

PR #3700, opened 2024-01-30
view: restore denom metadata queries to view service

hdevalence:
This was previously removed as being duplicative. But it isn't, because something might want to request specifically from a view server, which can respond with more opinionated denom metadata than the chain.


crates/proto/src/gen/proto_descriptor.bin.no_lfs
hdevalence, line 1: seems like something funky's happening here, not critical to fix now though

PR #3703, opened 2024-01-30
penumbra: remove chain component and move epoch management to SCT

erwanor:
Part of #3617, this PR is a work-in-progress:
- Dissolve the `ChainParameters` structure into `AppParameters` (chain id) and `SctParameters` (epoch window config)
- Make epoch/time management the responsibility of the SCT
- Remove the public interface of the chain component

This last step is critical, because it means that the final next iteration of this work can happen independently of other integration work streams.


erwanor:
This compiles, so this is a good time to pause for now. My remaining task:
- storing the chain id (app parameter) in the view service DB
- restore chain id checks in various query service request handlers
- doing a general review pass, and a bit of assurance

I am going to let the CI run overnight. Since the view service is broken any tests that exercise claiming should not pass. However, I think the rest of the CI should pass. We're so close!

erwanor:
Merging this large PR optimistically so that integration work can proceed and to minimize conflicts. Follow-up tasks are tracked in independent issues that we are tackling today. Also working on a release tracking list for testnet 65.

PR #3706, opened 2024-01-30
compactblock: add unary rpc, fix pcli query

hdevalence:
This allows us to inspect compact blocks in Buf Studio (by exposing a unary RPC) and fixes the broken query in `pcli q`.

As part of follow up work, we should eventually aim to replace all of the k/v queries in pcli with real RPCs that are actually useful by other things than just our Rust code.


hdevalence:
Running this against a local devnet:
```
~/c/p/penumbra (unary-compactblock-rpc|✔) $ cargo run --release --bin pcli -- --home /tmp/pcli/local q shielded-pool compact-block 0 | jq '.fmdParameters'
    Finished release [optimized] target(s) in 0.31s
     Running `target/release/pcli --home /tmp/pcli/local q shielded-pool compact-block 0`
{
  "asOfBlockHeight": "1"
}
```
It appears that the genesis compact block does include FMD parameters (cc @erwanor @turbocrime)

hdevalence:
merging to deploy to assist with web debugging

Issue #3336, opened 2023-11-15
Add a standalone node to testnet deployments


hdevalence:
**Is your feature request related to a problem? Please describe.**

We need `pd` to work out of the box. But we don't test this anywhere, because we wrap it up in a bunch of infrastructure that hides broken behavior: #3281

**Describe the solution you'd like**

Add a node to the deployment that runs a full node the way we expect users to be able to: `pd start --grpc-auto-https mydomain.com` and `cometbft start`, no load-balancing, no reverse proxies, etc.

conorsch:
This is a good idea. As background, I've actually been running a discrete fullnode on separate infra, but failed to do so for 64. Aside from testnets, it's important that we replicate this functionality for preview, so that we can catch bugs like https://github.com/penumbra-zone/penumbra/pull/3650 before release. However, it's complicated in that we must preserve the contents of the ACME cache, currently defined as `<pd_home>/rustls_acme_cache`: https://github.com/penumbra-zone/penumbra/blob/v0.64.2/crates/bin/pd/src/main.rs#L487-L488 Most of our preview logic is a full wipe and reset, and we'll need to be a bit more careful to avoid getting a domain banned from the ACME ratelimits.

conorsch:
Added a workflow for this on preview. I'm going to run it ad-hoc a few times, and if no problems—like ratelimit triggers—then I'll move it to prod ACME API and make it part of the automatic deployments.

conorsch:
The automatic preview deploy is triggering too early, before the newly created network's RPC endpoints are returning. For most of our deploys, that's not a problem, because they'll automatically retry until successful. For the "standalone" config, however, that doesn't leverage the same orchestration, we need to be more explicit. 

Two improvements come to mind: ensure that the RPC endpoints are honoring the readiness state of the fullnodes behind them: right now, any fullnode in the deployment is instantly added to the RPC backend pool, but we should gate admission into the pool on formal readiness, meaning the internal rpc endpoint is returning OK. We could also instruct the deploy flow to block until all pods are ready, which would resolve the problem of the standalone deploy firing too early, but not address the intermittent RPC downtime during chain resets on preview.

conorsch:
This work is basically complete, although I haven't documented the new endpoints anywhere. I'll stick those in the wiki before closing. 

[...]

PR #3709, opened 2024-01-31
ci: workflow for standalone pd

conorsch:
We want to exercise the pd https logic, but we can't naively run it from scratch on every deploy, because that'd be far too many API requests to reissue certs from ACME. Instead, let's preserve the ACME directory before wiping state, and reuse it before bouncing the service.

This setup requires always-on bxoes provisioned out of band.

Still TK:

  * use dedicated `ci` shell account
  * add GHA secrets for key material
  * use --acme-staging arg for first few runs
  * add dedicated workflow ad-hoc runs

Refs #3336.


conorsch:
> So far, this adds the base logic via a workflow. In order to get it running, I'll need to iterate on the workflow, but workflows must land on main prior to being available for ad-hoc execution.

I'm going to get this in as a necessary-but-not-sufficient base for further iteration.

PR #3710, opened 2024-01-31
Fix a few mistakes in the Penumbra Protocol docs

jessepinho:
Just a couple things I noticed while reading the Penumbra protocol.

### DIFF
--- a/docs/protocol/src/concepts/addresses_keys.md
+++ b/docs/protocol/src/concepts/addresses_keys.md
@@ -62,5 +62,4 @@ addresses, allowing fine-grained control of delegation.
 
 This diagram shows only the user-visible parts of the key hierarchy.
 Internally, each of these keys has different components, described in detail in
-the [Addresses and Keys](../protocol/addresses_keys.md) section of the
-[Cryptographic Protocol](../protocol.md) chapter.
+the [Addresses and Keys](../protocol/addresses_keys.md) chapter.
--- a/docs/protocol/src/concepts/transactions.md
+++ b/docs/protocol/src/concepts/transactions.md
@@ -35,10 +35,6 @@ pool](../stake/undelegation.md), consuming delegation tokens from the
 transaction's value balance and producing new notes recording the appropriate
 amount of unbonded stake;
 
-- **Commission** descriptions are used by validators to [claim commission on
-staking rewards](../stake/validator-rewards.md) into shielded notes,
-adding unbonded stake to the transaction's value balance;
-
 #### Governance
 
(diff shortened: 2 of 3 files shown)


docs/protocol/src/concepts/addresses_keys.md
```diff
@@ -62,5 +62,4 @@ addresses, allowing fine-grained control of delegation.
 
 This diagram shows only the user-visible parts of the key hierarchy.
 Internally, each of these keys has different components, described in detail in
-the [Addresses and Keys](../protocol/addresses_keys.md) section of the
-[Cryptographic Protocol](../protocol.md) chapter.
+the [Addresses and Keys](../protocol/addresses_keys.md) chapter.
```
jessepinho, line 65: Addresses and Keys is now its own chapter, rather than being under the Cryptographic Protocol chapter.

docs/protocol/src/concepts/transactions.md
```diff
@@ -61,7 +61,7 @@ description leaves the value balance unchanged.
 transaction's value balance, burning them, and producing a swap commitment for
 use in the second stage;
 
-- **Sweep** descriptions perform the second phase of
+- **SwapClaim** descriptions perform the second phase of
```
jessepinho, line 64: Reviewers: please confirm that this change is correct.

docs/protocol/src/concepts/transactions.md
```diff
@@ -35,7 +35,7 @@ pool](../stake/undelegation.md), consuming delegation tokens from the
 transaction's value balance and producing new notes recording the appropriate
 amount of unbonded stake;
 
-- **Commission** descriptions are used by validators to [sweep commission on
+- **Commission** descriptions are used by validators to [claim commission on
 staking rewards](../stake/validator-rewards.md) into shielded notes,
 adding unbonded stake to the transaction's value balance;
```
hdevalence, line 40: We actually got rid of these entirely, now the commission is paid out by the chain, so we could delete this bullet point.

docs/protocol/src/concepts/transactions.md
hdevalence: Yep! Another terminology change we didn't update

hdevalence (approved):
LGTM! Thanks for flagging this stuff, the spec has drifted a bit as we've focused on the protocol.

docs/protocol/src/concepts/transactions.md
jessepinho: Got it — removed.

PR #3717, opened 2024-02-01
ci: remove hermes build

conorsch:
The fork repo for hermes was reset, so the referenced workflow no longer exists. This was causing an HTTP 422 in the build logic. Removing hermes references to sidestep the problem and unbreak preview deploys.


conorsch:
Overridding CI to merge pronto.

PR #3722, opened 2024-02-01
Add support for fee tiers to TransactionPlannerRequest

jessepinho:
Currently, there is no way to set a fee tier in a `TransactionPlannerRequest`. This blocks https://github.com/penumbra-zone/web/issues/309 (supporting fee tiers on web), because the only way to plan a transaction on web is to pass a `TransactionPlannerRequest` to the extension's view service.

This PR adds support for fee tiers to `TransactionPlannerRequest`.

Reviewers might want to look at the [first commit](https://github.com/penumbra-zone/penumbra/pull/3722/commits/fd962a75680b69487d8dac2b1111ae7459b953d1) on its own, since that's where I changed the `.proto` file; the second commit is simply the result of running `protobuf-codegen`.

Note that linting fails in this PR due to breaking changes in the `TransactionPlannerRequest` protobuf definition.

Related issue: #3724


proto/penumbra/penumbra/view/v1alpha1/view.proto
```diff
@@ -185,13 +185,19 @@ message BroadcastTransactionResponse {
...
+    // The fee for the requested TransactionPlan, if any.
+    core.component.fee.v1alpha1.Fee fee = 2;
+    // The fee tier for the requested TransactionPlan, if any.
+    core.component.fee.v1alpha1.FeeTier fee_tier = 3;
+  };
+
   // The memo for the requested TransactionPlan.
   // The memo must be unspecified unless `outputs` is nonempty.
-  core.transaction.v1alpha1.MemoPlaintext memo = 3;
+  core.transaction.v1alpha1.MemoPlaintext memo = 4;
```
hdevalence, line 198: We should avoid renumbering fields unless really necessary, to avoid breaking changes: https://protobuf.dev/programming-guides/proto3/#assigning

proto/penumbra/penumbra/view/v1alpha1/view.proto
```diff
@@ -185,13 +185,19 @@ message BroadcastTransactionResponse {
...
   uint64 expiry_height = 1;
-  // The fee for the requested TransactionPlan, if any.
-  core.component.fee.v1alpha1.Fee fee = 2;
+
+  oneof fee_option {
+    // The fee for the requested TransactionPlan, if any.
+    core.component.fee.v1alpha1.Fee fee = 2;
+    // The fee tier for the requested TransactionPlan, if any.
+    core.component.fee.v1alpha1.FeeTier fee_tier = 3;
+  };
```
hdevalence, line 194: I think it would be good if this were changed to make it clearer that it's either a manual fee setting or an auto fee setting.

hdevalence:
Pairing with @jessepinho led to the discovery of a large amount of friction in actually using these parameters but also ideas on how we can radically simplify the planning code, once we split out the current wasm crate into a minimal bindings part and an extension implementation part.

jessepinho:
@hdevalence looks like I can't hit the approve button since I authored the PR, but 👍🏻 — feel free to self-merge

PR #3729, opened 2024-02-02
shielded-pool(component): simplify tracking of fmd parameters

### LINKED ISSUE #3713: shielded-pool: declutter fmd parameter init
Follow-up to #3703, that we are tackling today. This is a small change to the shielded pool's state read trait.

erwanor:
Close #3713, a follow-up to #3617. Now that we have shielded pool parameters, we do not want to have a separate code path to update the FMD parameters without also updating the shielded pool params.

### DIFF
--- a/crates/core/app/src/app/mod.rs
+++ b/crates/core/app/src/app/mod.rs
@@ -254,7 +254,9 @@ impl App {
                 state_tx.put_ibc_params(ibc_params);
             }
             if let Some(shielded_pool_params) = app_params.new.shielded_pool_params {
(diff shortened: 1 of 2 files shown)


hdevalence (changes requested):
I don't think we want to do this. 

Long term the FMD parameters shouldn't be in the chain parameters anyways, similar to gas prices. They need to be dynamically adjusted based on transaction volume. The values in the AppParameters are just fixed starting parameters that should eventually be replaced by whatever parameters are needed to tune the auto-adjuster.

erwanor:
I see - tbh, I had not realized that it was V1 tagged but it is.

PR #3730, opened 2024-02-02
view: auto-restart view worker on error

hdevalence:
Currently, any transient error in the view worker permanently stops the server forever. Instead, restart after a delay.

### DIFF
--- a/crates/view/src/worker.rs
+++ b/crates/view/src/worker.rs
@@ -384,22 +384,20 @@ impl Worker {
     }
 
     pub async fn run(mut self) -> anyhow::Result<()> {
-        self.run_inner().await.map_err(|e| {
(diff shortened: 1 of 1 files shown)


conorsch:
Refs https://github.com/penumbra-zone/penumbra/issues/3732

cronokirby:
We should revisit the use of `error_slot` and how the status of the view service is reported in other methods, otherwise I anticipate confusion when an upstream caller looks at the status during the sleep.

hdevalence:
Yeah, agreed that those are all tricky things. I think what we want to do is have a health check method that allows the caller to see if the service is healthy or not. There's one example of how to do an "internal" health check method in Hermes, though we might want something slightly different.

In the meantime I think we should merge this and track that separately.

Issue #3615, opened 2024-01-16
stake: refactor the staking component to use modular traits (draft)


erwanor:
**Context:**

The staking component is largely implemented by a `StakingImpl` trait which has three responsibilities:

- Manage the validator state machines: both triggering and implementing state transitions;
- Implement end-epoch logic: computing delegation changes, defining the base rate for the next epoch, updating validator state/power;
- Responsible for important but miscellaneous tasks like tracking validator uptime, processing misbehavior evidence, and materializing cometbft consensus set updates. 

To achieve this, the `StakingImpl` trait ties together three other interfaces:
- State extension traits (R/W) that handle validator indices, staking component parameters, rate data, delegation changes, and penalty tracking.
- `ValidatorUpdates` R/W that queues consensus set changes.

The validator state machines are highly coupled to the extension traits, which themselves have low cohesion, making it hard to implement new features without having to inspect every method in the the fits-all extension trait to make sure that things are correct. 

**Proposal:**

```rust
/// Implements the staking [`cnidarium_component::Component`] logic.
pub struct Staking {}

pub(crate) trait StakingImpl: StateWrite {
    /// Process the end of an epoch, delegation changes, rate data calculations
    /// and produce the resulting updated validator set.
    fn end_epoch(...) -> Option<TendermintUpdate>;
}

/// An internal trait responsible for the implementation of the validator
/// state machine.
pub(crate) trait ValidatorManager: StateWrite + ValidatorIndex {
    /// Add a validator definition, defining its initial state and rate data.
    fn add_validator(&mut self, validator: Validator);
    /// Add a genesis validator definition, defining it as [`State::Active`].
    fn add_validator_genesis(&mut self, validator: Validator);
    /// Apply a state transition to a validator.
    async fn set_validator_state(&mut self, validator: IdentityKey, new_state: validator::State) -> Result<()>;
    /// Update a validator's definition, potentially causing a state transition.
    async fn update_validator(&mut self, validator: IdentityKey, definition: Validator) -> Result<()>;
    /// Update the consensus set's uptime, jailing validators that have missed too many blocks.
    async fn track_validator_uptime(&mut self, last_commit: tendermint::LastCommitInfo) -> Result<()>;
}

/// An internal trait that mediate writes to staking data.
pub(crate) trait StakingData {
    fn push_delegation(&mut self, delegation: Delegate);
    fn push_undelegation(&mut self, undelegation: Undelegate);
}

/// A public interface to staking data.
[...]

PR #3737, opened 2024-02-05
staking: break out staking component into modular traits

erwanor:
Implements #3615.


erwanor:
Since this is largely following the #3615 design issue (modulo some renaming), I will merge this optimistically. Next up is adjusting the state-keys, and finally, fixing the issues surfaced during the validator state machine review.

PR #3742, opened 2024-02-05
feat(pcli): default coordinator address for ceremony

conorsch:
During Phase 1 of the summoning ceremony, some contributors were confused about the `--coordinator-address` flag, and incorrectly overrode it to a wallet address they control. Let's set a sane default value for that arg, and remove its display from the web page.


cronokirby (commented):
Seems like the summonerd changes were a mistake?

Let me do a quick modification of the commit, then we can merge.

cronokirby:
(ran rustfmt in an amended commit)

cronokirby (approved):
LGTM!

PR #3749, opened 2024-02-06
📌 fix(ci): pin nightly toolchain

cratelyn:
this ci action fails due to tkaitchuck/aHash#200.


cratelyn:
the rustdoc build script uses `+nightly`. we'll need to sort out how to thread that through from the github action for future nightly breakages like this.

conorsch:
> the rustdoc build script uses +nightly

I patched the rust-docs build script to use `nightly-2024-02-01` as a temporary workaround. Builds locally for me again, so CI should pass, too.

PR #3752, opened 2024-02-06
Emit ABCI events from governance-related actions

### LINKED ISSUE #3581: governance: record proto events when proposal pass
Before we ship 65, we should make sure that governance events are recorded using proto-events. 

Part of #3504.

zbuc:
Closes #3581


erwanor (approved):
LTGM on green ci

Issue #2921, opened 2023-08-11
staking: add a `min_validator_stake` threshold for validator indexing


plaidfinch:
This is blocked on completion of #2909, because it will be simpler to implement once that's done, and also because it will generate horrible merge conflicts if done concurrently.

## The problem

Currently, the staking component loops over all validators, every epoch end. This presents the possibility for an intentional or unintentional denial of service where the amount of I/O and computation per epoch end scales proportionately to the number of validators which have _ever_ been declared, regardless of their state. As a consequence, a malicious actor could declare a large number of validators, paying an up-front cost to do so as a per-definition fee, and permanently slow down the chain, increasing the resource usage required to operate a node every epoch end. Additionally, a long-lived chain will likely see many non-malicious declarations of validators, many of which will not be successful enough to attain a long-lived position in the active set. Because the continuous per-epoch cost to the chain currently scales with the total historical number of validators ever declared, it is currently a mismatch in deterrance for declaring a validator to assess a one-time fee on validator declarations.

## High-level solutions

There are several conceptual ways to fix this:

1. Impose an iterated fee on declared validators, somehow removing them entirely from the chain if this continuous cost is not met.
2. Change the internal implementation of the chain so that computation and I/O does not scale linearly per-epoch with the total historical number of validators.
3. Require a minimum delegation amount for a validator to be considered even to be `Inactive`, and ensure that validators below this minimum amount won't be iterated over. (This is essentially [what Namada does](https://namada.net/blog/namada-cubic-proof-of-stake).)

~~The second option is strongly preferable to the first, because it does not require additional operational concerns for penumbra validators, and it does not complicate the staking state model further by adding yet another state into which a validator can enter.~~ Edit: after further discussion with @hdevalence, we've decided to go with (3), since it addresses the issue while reducing the chance of implementation error now or in the future when modifying the staking implementation. **The revised proposed implementation and analysis is directly below, with the old design and analysis left as a comment to record the alternate proposal.**

As we will see, achieving (2) actually does impose an implicit continuous cost on any validator whose mere validator definition's existence causes a per-epoch cost. (Though this is no longer relevant because we are going with the third option).

## Why was it this way to begin with?

We store the current rate for each validator in the state:

```rust
pub fn current_rate_by_validator(id: &IdentityKey) -> String {
    format!("staking/validator_rate/current/{id}")
}
```

This key stores values of type `RateData`:

```rust
pub struct RateData {
    /// The validator's identity key.
[...]

PR #3761, opened 2024-02-07
staking: make multi-tombstoning a no-op and a state machine fix

erwanor:
This PR implement fixes that were surfaced during a validator state-machine
review:

- Double tombstoning: Previously, we disallowed the state transition
`Tombstoned -> Tombstoned` implicitly assuming that a validator
node can only commit byzantine behavior once.  This is, of course,
not a reasonable assumption. A malicious validator can equivocate
as many times as they want. Therefore, we need to be able to support
`T-> T` transitions

- Missing transitions: We did not implement the `Defined <-> Disabled`
transition. This is a bug in #2921.

### DIFF
--- a/crates/core/component/stake/src/component/epoch_handler.rs
+++ b/crates/core/component/stake/src/component/epoch_handler.rs
@@ -20,7 +20,7 @@ use tracing::{instrument, Instrument};
 use crate::{
     component::{
         stake::{ConsensusUpdateWrite, InternalStakingData, RateDataWrite},
-        validator_handler::{ValidatorDataRead, ValidatorManager, ValidatorStore},
+        validator_handler::{ValidatorDataRead, ValidatorManager, ValidatorDataWrite},
         SlashingData, FP_SCALING_FACTOR,
     },
     validator, CurrentConsensusKeys, DelegationToken, FundingStreams, IdentityKey, Penalty,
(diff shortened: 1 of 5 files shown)


hdevalence:
> This is, of course, not a reasonable assumption. A malicious validator can equivocate as many times as they want.

Hmm, I'm not sure the assumption is "malicious validators would only equivocate once", but rather "any equivocation is equally bad as arbitrary equivocation". A validator only needs to equivocate once to fault the chain and so should be punished accordingly.

I'm not opposed to merging, but I am not sure we should make/preserve this change without thinking it through separately, it's semantically different rather than just patching a gap.

My recollection from talking to @zmanian two years ago is that the tombstoning behavior in the SDK was chosen because of lessons from Game of Stakes, I think an issue where a validator was misconfigured and got slashed into oblivion while the operators (in another TZ) were asleep and no one could reach them. I don't recall exactly what happened there.

hdevalence (approved):
Approved, with followup for double tombstoning discussion

zmanian:
Context:

Around the time of the Cosmos launch, the Coinone Tezos validator was misconfigured and was generating equivocations every couple hours until the operator woke up.

The tombstoning design ended up having too effects. It protected delegators from having a continuous stream of faults. It was an also a dramatic penalty to a validator because it would reset their delegations from zero. This seems like a healthy balance in optimizing for network security.

erwanor:
Thanks for chiming in @zmanian @hdevalence , this is really helpful. I think there is probably a lot more we could do to polish this mechanism, but a fixed rate applied once and irrespective of stake size seems good enough for now. I'll amend the PR.

PR #3763, opened 2024-02-07
Make effect hash optional

grod220:
Allows web's `WitnessAndBuildRequest` to not pass `authorizationData` as a field. Needed for cases like `swapClaim` where no auth is required to building.

### DIFF
--- a/crates/core/app/src/action_handler/actions/submit.rs
+++ b/crates/core/app/src/action_handler/actions/submit.rs
@@ -4,27 +4,14 @@ use std::sync::Arc;
 use anyhow::{Context, Result};
 use ark_ff::PrimeField;
 use async_trait::async_trait;
-use cnidarium::{StateDelta, StateRead, StateWrite};
 use decaf377::Fq;
 use decaf377_rdsa::{VerificationKey, VerificationKeyBytes};
 use ibc_types::core::client::ClientId;
 use once_cell::sync::Lazy;
+
+use cnidarium::{StateDelta, StateRead, StateWrite};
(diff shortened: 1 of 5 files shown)


crates/core/app/src/action_handler/actions/submit.rs
```diff
@@ -4,34 +4,34 @@ use std::sync::Arc;
...
 use penumbra_sct::component::tree::SctRead;
 use penumbra_shielded_pool::component::SupplyWrite;
-
 use penumbra_transaction::plan::TransactionPlan;
 use penumbra_transaction::Transaction;
 use penumbra_transaction::{AuthorizationData, WitnessData};
 
 use crate::action_handler::ActionHandler;
 use crate::community_pool_ext::CommunityPoolStateWriteExt;
 use crate::params::AppParameters;
```
cronokirby, line 34: Just sorting imports, I see.

cronokirby (changes requested):
So, the changes in this PR are coherent, but one question I have about the approach is why we need to change `AuthorizationData`. For a transaction where no authorization is needed, shouldn't we instead not pass in `AuthorizationData`, rather than pass in an instance of the struct with no signatures or effect hash?

Also, why can't the web context create an `AuthorizationData` with the effect hash computed, but no signatures?

cronokirby (approved):
After some more context in Discord, I see that computing the effect hash in the frontend would require having WASM, which is a bit of an annoyance. Also, this effect hash being optional in the authorization data is fine, because the TransactionPlan binds it anyways, it's more of a sanity check thing.

TalDerei (approved):

PR #3767, opened 2024-02-07
ibc: add optional packet data memo field

### LINKED ISSUE #3762: ibc: add memo field to ICS20 FungibleTokenPacketData
Since the implementation of https://github.com/cosmos/ibc/issues/584, the standard packet data for ICS20 transfers adds a new field, `memo: string`, in order to support more complex uses of interchain transfers. Our ICS20 implementation was written before this change was widely adopted by the ecosystem (as it is now), and as such doesn't contain this memo field in `FungibleTokenPacketData`.

We should upgrade our IBC protos to the current interchain standard that includes the `memo` field. I don't think we need to populate it in our state machine, just updating the protos (and any associated domain types) should be sufficient. This change should be implemented in ibc-types.

avahowell:
Closes #3762


cratelyn (approved):

conorsch:
Needs rebase due to proto descriptor churn.

PR #3768, opened 2024-02-07
staking: add limit on number of validator funding streams

### LINKED ISSUE #2195: Limit validator definitions to eight funding streams

The chain mints reward notes at the end of each epoch, distributing validator commissions to the funding streams. These notes are minted by the chain, and bypass any normal fee mechanism. Allowing validators to have arbitrarily many funding streams is problematic, because they could have the chain mint many notes on their behalf.

The funding stream minting is especially problematic because it happens in EndBlock, which must be infallible, and there's a bound on the number of state commitments that can be added to the state commitment tree in each block.

We previously did a ballpark estimate suggesting that this bound couldn't be reached within reasonable block size limits. Keeping the limit small (8) helps us continue to avoid this problem; with a 100-validator limit, we only need 800 commitments' of buffer (out of 16k).

Add a stateless check to the validator definition ActionHandler that checks that there are at most 8 funding streams per validator.

erwanor:
Close #2195.

### DIFF
--- a/crates/core/component/stake/src/component/action_handler/validator_definition.rs
+++ b/crates/core/component/stake/src/component/action_handler/validator_definition.rs
@@ -31,6 +31,10 @@ impl ActionHandler for validator::Definition {
             anyhow::bail!("validator description must be less than 280 characters")
         }
 
+        if self.validator.funding_streams.funding_streams.len() > 8 {
+            anyhow::bail!("validators can declare at most 8 funding streams")
+        }
+
         // Then, we check the signature:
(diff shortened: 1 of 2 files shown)


conorsch:
Discussed the possibility of constifying this value with @erwanor, since it'd be a lot easier to reason about. We should still do that, but I'm not going to block merge on it: there are a bunch of related values like string length limits on the website fields that similarly should have a named const. Deferring that clean up for now, so we can focus on unblocking the upcoming 67 release.

PR #3770, opened 2024-02-08
staking: add missing effect for `Defined <> Inactive`

erwanor:
There is a bug in the current v0.65.0 release. `Defined` validators are not transitioning out of the precursor state into `Inactive`. The reason is that we are not actually effecting the validator state. This had been previously fixed in a follow-up #2921, but somehow slipped back in.

### DIFF
--- a/crates/core/component/stake/src/component/validator_handler/validator_manager.rs
+++ b/crates/core/component/stake/src/component/validator_handler/validator_manager.rs
@@ -109,12 +109,14 @@ pub trait ValidatorManager: StateWrite {
                 // the staking component.
                 tracing::debug!(identity_key = ?identity_key, "validator has reached minimum stake threshold to be considered inactive");
                 self.add_consensus_set_index(identity_key);
+                self.put(validator_state_path, Inactive);
             }
             (Inactive, Defined) => {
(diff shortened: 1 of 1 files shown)


hdevalence:
Great find, unfortunate bug. Since this would change the execution of previous blocks, I'm not sure we can easily merge this without causing newly syncing nodes to diverge. I suspect the best course of action here (cc @conorsch) is:

- Merge, rewrite the git tag, nuke the existing network and redeploy (it's only a day old)
- Discuss in a retro of how we could (a) mitigate root causes of the bug and (b) how we could make mitigation less painful.

I think this will be a useful learning experience.

conorsch:
> Merge, rewrite the git tag, nuke the existing network and redeploy

Slight change after discussion: we'll push up a fresh tag, to avoid breaking repos in the wild, `v0.66.0`. Then we'll indeed fully deploy, nuking all chain state and resetting to 0. We'll still use the same chain id, so the name "Deimos" still refers to the batch of changes we tried to ship yesterday.

conorsch:
Post-merge, I will deploy an ad-hoc validator on preview and confirm it can reach the active set.

Issue #3787, opened 2024-02-08
tests: 🎪 `App` can be instantiated with mock consensus


cratelyn:
see #3588.

#3781 added some test stubs for our mock engine that compile, but do not yet pass. now, write the basic functionality needed to instantiate the `App` in a test with the mock consensus engine.

cratelyn:
this was marked complete, and fixed by #3816 and #3807.

* #3816
* #3807

Issue #3588, opened 2024-01-08
Tracking issue: 🤝 mock consensus engine for `App` tests


cratelyn:
### 💭 background and motivation

currently, writing complete end-to-end integration tests for Penumbra is
difficult. moreover, we don't have a better alternative strategy for testing the core `App` type.

currently, the [`smoke-test.sh`][smoke-test] script performs the steps needed
to run integration tests of the system, including:
- generate testnet configuration
- start a `cometbft` process
- start a `pd` process
- run pclientd integration tests against the local test network
- run pcli integration tests against the local test network
- confirm the test network is still running

because these integration tests assume a test network including `pd` and
`cometbft` is running locally, these tests must run serially, _and_ must spend
time waiting for new blocks to be generated.

additionally, this means we must under-handedly make use of the `#[ignore]`
attribute to prevent these tests from running (_and failing_) alongside unit
tests when commands like `cargo test --workspace` are run.

see [CometMock] for a mock implementation of CometBFT implemented
in Go. see #2771 for previous work in this space integrating with `CometMock`.

### 📐 overview

this state of affairs would be improved if we had a mock implementation of the
consensus engine ([CometBFT]) that would allow `#[test]` functions to
generate ABCI messages. this mock implementation should function as a drop-in
replacement for CometBFT from the perspective of the `pd` daemon.

this mocking library should additionally expose some "meta" interfaces. to allow us
to write integration tests including things like:
- manipulating the timer used for generating blocks, so that edge-cases
  involving e.g. expiry may be exercised. 
- commit double-signing violations
- [...] (_this list is non-exhaustive at the time of writing_)

this will allow us to run integration tests faster, write them more easily,
and generally improve our ability to make assurances about the correctness
of our software.

### :triangular_ruler: requirements & design guidelines

(_this will go into crate-level documentation at some point_)

* the mock consensus engine should be application agnostic. that means it should not depend on `penumbra-*` crates. ([#3810](https://github.com/penumbra-zone/penumbra/pull/3810#discussion_r1488110547))

* the mock consensus engine is built to drive a consensus service `C: Service<ConsensusRequest, Response = ConsensusResponse, Error = BoxError>`

### :link: related work

* #3597
* #3695
  * #3672
  * #3789
  * #3794
* #3787
  * #3781
  * penumbra-zone/tower-abci#41
  * #3799
  * #3807
  * #3810
* #3816
  * #3902
  * #3922
* #3788
  * #3849
  * #3857
* #3792
[...]

PR #3799, opened 2024-02-09
tests(app): ⭐ use in-memory abci server

cratelyn:
see #3588, #3787.

related to penumbra-zone penumbra-zone/tower-abci#41.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8291,8 +8291,6 @@ dependencies = [
 [[package]]
 name = "tower-abci"
 version = "0.11.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0d4826f3df3e9a37083d978cae73f020bcdf6143956b7dfc1bd6050b4e16367c"
 dependencies = [
  "bytes",
  "futures",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8292,6 +8292,8 @@ dependencies = [
 [[package]]
 name = "tower-abci"
 version = "0.11.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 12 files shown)


hdevalence:
crossreferencing https://github.com/penumbra-zone/tower-abci/pull/41#issuecomment-1936860930 for visibility

cratelyn:
@hdevalence i will respond to your comment here, rather than in
penumbra-zone/tower-abci#41, if that's okay. that's a relatively small patch,
and we're focused on how this approach fits into a solution to #3588.

the tldr; up front: **i did this to write a "hello world" test.** this is a
non-commital choice, and a short-term stopgap prioritizing ease and speed of
implementation.

- using `tower-abci` allows me to write a "hello world" program by
sending other kinds of ABCI messages, namely an `Echo`. this is the
*primary* motivation. i am trying to stand the consensus engine up, and
this is the easiest way to start.

- this is a non-commital choice. if i start by writing tests against a
`Service<C, M, I, S>`, it is very easy to back out later into writing the
same tests against the `C` generic that specifically works with consensus
requests.

in your comment, you said:

> [bridges] the bespoke Tendermint socket protocol to a more reasonable, request/response-oriented API.

to your point, this is why i am reaching for tower-abci! it is a reasonable
request/response-oriented API that is pleasant to work with. the downside
of this approach is encoding/decoding requests and responses. the upside is
that we have a simpler, more ergonomic surface to work with: requests go in,
responses come out.

i agree that using `Consensus` in the long run is a more performant way of
running application tests. using `tower-abci` helps in the short-term however,
because i can write a "hello world" test and incrementally build the mock
consensus engine against a simple abstraction of the penumbra node.

cratelyn:
closing in favor of #3807.

PR #3803, opened 2024-02-12
staking: use validator exchange rate to compute rewards

### LINKED ISSUE #3802: staking: reward calculation exceeds issuance budget
Testnet 66 has halted because an invariant check on rewards has been breached:

```
penumbra_funding::component: the sum of staking rewards for the epoch has exceeded the issuance budget total_staking_rewards_for_epoch=2003279 staking_issuance_budget=2000000 reward_amount_for_stream=19571
```

In other words, more rewards have been minted than authorized by the distribution component.

In #3310, we ported the staking component's rate calculations to use our fixed point arithmetic crate (`U128x128`), in this process we found a couple bugs with how rates and rewards were computed, by and large, these were holdovers from a time when rates were computed on the first block of new epochs.

We missed however, that the reward amount formula does not account for historical commissions or penalties:

```
// The reward amount is computed as:
//   y_v * c_{v,e} * r_e * psi(e)
// where:
//   y_v = total delegation tokens for validator v
//   c_{v,e} = commission rate for validator v, at epoch e
//   r_e = base reward rate for epoch e
//   psi(e) = base exchange rate for epoch e
// The commission rate is the sum of all the funding streams rate, and is capped at 100%.
// In this method, we use a partial commission rate specific to `this` funding stream.
```

Zooming in, take note of how we translate the validator's delegation pool in units of staking tokens: $y_v \cdot \psi(e)$. In plain english, we multiply the delegation pool of validator $v$, by the base exchange rate for epoch $e$.

This means that we effectively assume that this validator has had (historically) a zero commission rate and no penalties. On the other hand, the chain's base reward rate is computed based on the correct amount of total active stake. As a result, there's a discrepancy between the reward rate ("quantity of issuance per unit of staking token") and the realized reward quantity ("amount of rewards that correspond to this pool").

I think we can fix this by using the validator's exchange rate, so that we account for historical commissions and penalties.

erwanor:
Submitted for independent review and testing, should fix #3802. 

Testing that this change work is a little bit tricky, here are the steps that I have taken just now:

1. I setup a devnet with short block time (300ms), and short epoch durations (50 blocks).
2. The default validator has 100% of the total active stake and a 100% commission rate.
3. On the `v0.66.0` tag, this experimental setups leads to a chain halt on epoch 2 (~ 25 seconds after start).

### DIFF
--- a/crates/core/component/funding/src/component.rs
+++ b/crates/core/component/funding/src/component.rs
@@ -1,6 +1,7 @@
 mod state_key;
 pub mod view;
 use penumbra_asset::{Value, STAKING_TOKEN_ASSET_ID};
+use penumbra_stake::component::validator_handler::ValidatorDataRead;
 pub use view::{StateReadExt, StateWriteExt};
 
 use std::sync::Arc;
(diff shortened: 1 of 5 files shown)


cronokirby (approved):
I gave this a thorough synchronous review with @erwanor, and am convinced that this is correct, and fixes other bugs beyond the circuit breaker, like a validator getting too much in commissions if slashed.

cratelyn (approved):
seconding the approval, as another participant in group review today. :star:

Issue #3784, opened 2024-02-08
Testnet 68: Deimos 3


conorsch:
# Testnet Release

Testnet name: Deimos 3
Release date: 2024-02-26
Testnet release manager: @erwanor 

The Phobos release should exercise chain upgrades.

# Testnet Release Manager Checklist

Preceding Friday (sprint planning day):

- [ ] Create GitHub project column, work with team to populate the milestone with tickets targeted for the release.

Tuesday (or after release of previous testnet):

- [ ] Construct the genesis data for the release:
  - [ ] Create new testnet directory with initial genesis allocations for this testnet by running `cd testnets && ./new-testnet.sh`
    - This genesis data will be used for `testnet-preview` with a randomized version of the future testnet's chain ID.
  - [ ] Make sure that the chain _name_ is preserved: it should be `penumbra-testnet-deimos-3`.

Thursday:

- [ ] Check in with team again in a release meeting and update the GitHub milestone to ensure it represents what will make it into the testnet.
- [ ] Draft an announcement for peer review to ensure major changes included are comprehensive.

Following Monday (release day):

- [ ] Verify that `testnet-preview.penumbra.zone` is operational; it is redeployed on every push to main, and is an exact preview of what is about to be deployed.
- [ ] Add an entry to the `COMPATIBILITY.md` 
- [ ] Bump the version number and push its tag, via [cargo-release](https://crates.io/crates/cargo-release):
    - [ ] Enable direct pushes to main, temporarily, by unchecking "Do not allow bypassing the above settings" on the [branch protection page](https://github.com/penumbra-zone/penumbra/settings/branch_protection_rules/22169356).
    - [ ] Run `cargo release minor` for a new testnet, or `cargo release patch` for a bugfix. For the latter, make sure you're on a dedicated release branch.
    - [ ] Push the commit and newly generated tag, e.g. `v0.51.0`, to the remote.
    - [ ] Re-enable branch protection on main.
- [ ] Wait for the ["Release" workflow](https://github.com/penumbra-zone/penumbra/actions/workflows/release.yml) to complete: it'll take ~20m, after which the release URL will be live.
- [ ] Edit the newly created (and published) release object, then click "Generate release notes" and save it. 
[...]

Issue #3779, opened 2024-02-08
Testnet 66: Deimos


conorsch:
# Testnet Release

Testnet name: Deimos
Release date: 2024-02-08
Testnet release manager: @conorsch

On 2024-02-07 we shipped Testnet 65 (#3554), but it included a showstopping bug, where non-genesis validators could not enter the active set. We've got a fix queued up for that here: https://github.com/penumbra-zone/penumbra/pull/3770

Merging that fix is a consensus-breaking change, so we're going to bump minor versions, but preserve the "deimos" (and chain id) for this long-running testnet. 

# Confirming resolution of problem

- [x] Merge #3770 to deploy to preview
- [x] Join a fullnode to the preview deployment
- [x] Submit validator definition for that fullnode to preview
- [x] Delegate to that validator
- [x] Wait requisite epochs, confirm non-genesis validator has entered active set

# Testnet Release Manager Checklist

Following Monday (release day):

- [x] Verify that `testnet-preview.penumbra.zone` is operational; it is redeployed on every push to main, and is an exact preview of what is about to be deployed.
- [x] Bump the version number and push its tag, via [cargo-release](https://crates.io/crates/cargo-release).
    - [x] Run `cargo release minor` for a new testnet, or `cargo release patch` for a bugfix. For the latter, make sure you're on a dedicated release branch.
    - [x] Push the commit and newly generated tag, e.g. `v0.51.0`, to the remote.
- [ ] Wait for the ["Release" workflow](https://github.com/penumbra-zone/penumbra/actions/workflows/release.yml) to complete: it'll take ~60m, most of which is the Windows builds.
- [ ] Edit the newly created (and published) release object, then click "Generate release notes." Cut and paste the generated text from the bottom to the top of the post, then save it.
- [ ] You must [manually review](https://docs.github.com/en/actions/managing-workflow-runs/reviewing-deployments) the [`Waiting` deployment in the GitHub Actions UI](https://github.com/penumbra-zone/penumbra/actions/workflows/deploy-testnet.yml) before the deployment will begin. Monitor the GitHub action to ensure it completes after it is approved.
- [ ] Delegate to the Penumbra Labs CI validators; use amounts of ~200k `penumbra` per validator.
- [ ] Open a position to provide some initial liquidity: `pcli tx lp replicate xyk penumbra:test_usd 20000penumbra --current-price 20`
- [ ] Update Galileo deployment, [following docs](https://github.com/penumbra-zone/galileo).
[...]

PR #3804, opened 2024-02-12
chore: create "deimos-2" chain id

conorsch:
Testnet 66 experienced a chain halt. To recover, we're going to release a new chain id, "penumbra-testnet-deimos-2", in an old-school nuke-and-redeploy scenario. The team discussed the opportunity to exercise chain upgrade logic, but decided it'd be better to prepare docs for validator operations ahead of time.

This change effectively reverts (or moots) 8c5cb5b8e088ca585c724c7c7f57a1f0f1dda997, which prematurely bumped the chain id on main to phobos. We're not using the phobos name yet, because the 65, 66, and now 67 versions are logically part of the deimos series.

Notably we don't increment APP_VERSION in the app crate, because doing so wouldn't be reflected in the generated chain id, given the current build logic. We should overhaul the chain id logic to append the app_version automatically, which is consistent with chain naming expectations in the ecosystem.


conorsch:
Refs https://github.com/penumbra-zone/penumbra/issues/3784, https://github.com/penumbra-zone/penumbra/issues/3779

PR #3807, opened 2024-02-13
tests(app): 🔗 send InitChain request to test node

cratelyn:


crates/core/app/tests/common/mod.rs
```diff
@@ -1,9 +1,32 @@
...
 use cnidarium::TempStorage;
 use penumbra_app::app::App;
 use penumbra_genesis::AppState;
 use std::ops::Deref;
 
+// Installs a tracing subscriber to log events until the returned guard is dropped.
+pub fn set_tracing_subscriber() -> tracing::subscriber::DefaultGuard {
+    use tracing_subscriber::filter::EnvFilter;
+
+    let filter = "debug,penumbra_app=trace,penumbra_mock_consensus=trace";
```
hdevalence, line 16: this may need to have some custom statements in order to not make arkworks explode, see how the tracing filters are set in pcli for reference

hdevalence (approved):
This looks great!

crates/test/mock-consensus/src/builder/init_chain.rs
```diff
@@ -0,0 +1,94 @@
...
+        C::Error: Sized,
+    {
+        use tendermint::v0_37::abci::response;
+
+        let request = Self::init_chain_request();
+        let service = consensus
+            .ready()
+            .await
+            .tap_err(|error| error!(?error, "failed waiting for consensus service"))
+            .map_err(|_| anyhow!("failed waiting for consensus service"))?;
```
hdevalence, line 37: can this use .context so we don't throw away the error info?

hdevalence: then i think we also wouldn't need `tap`, since all the context would be in the error itself

crates/test/mock-consensus/src/builder/init_chain.rs
```diff
@@ -0,0 +1,94 @@
...
+
+        let response::InitChain { app_hash, .. } = match service
+            .call(request)
+            .await
+            .tap_ok(|resp| debug!(?resp, "received response from consensus service"))
+            .tap_err(|error| error!(?error, "consensus service returned error"))
+            .map_err(|_| anyhow!("consensus service returned error"))?
```
hdevalence, line 44: ditto above

crates/test/mock-consensus/src/builder/init_chain.rs
```diff
@@ -0,0 +1,94 @@
...
+        {
+            ConsensusResponse::InitChain(resp) => resp,
+            response => {
+                error!(?response, "unexpected InitChain response");
+                bail!("unexpected InitChain response");
+            }
+        };
```
hdevalence, line 51: tioli: i think this could be slightly cleaner if it were a `let ... else { bail!() }` structure

crates/test/mock-consensus/src/builder/init_chain.rs
```diff
@@ -0,0 +1,94 @@
...
+
+        Ok(TestNode {
+            consensus,
+            last_app_hash: app_hash.as_bytes().to_owned(),
+        })
+    }
+
+    fn init_chain_request() -> ConsensusRequest {
```
hdevalence, line 59: probably this will want to be `&self` in the future so we can read stuff off the builder (assuming it's not like that for now to suppress unused var warnings)

crates/test/mock-consensus/src/lib.rs
```diff
@@ -1,45 +1,19 @@
...
-    }
-}
-
-impl Builder {
-    // TODO: add other convenience methods for validator config?
-
-    /// Creates a single validator with a randomly generated key.
-    pub fn single_validator(self) -> Self {
+impl<C> TestNode<C> {
+    pub fn next_block() -> tendermint::Block {
```
hdevalence, line 16: i think this could probably not be a `pub fn`, i'm not sure when we'll need to access the block itself (?)

crates/test/mock-consensus/src/builder/init_chain.rs
cratelyn: i opted for this form so we have the chance to log what we _did_ get in the unexpected case, though i'm a fan of `let .. else` is 99% of cases :slightly_smiling_face:

crates/test/mock-consensus/src/builder/init_chain.rs
cratelyn: :heavy_plus_sign: i agree, it's just like this for now until i do need those fields. lazy evaluation, so to speak

crates/core/app/tests/common/mod.rs
cratelyn: 3c878f7c1

cratelyn: for my future intuition, what does arkworks do with respect to our tests? i had seen that explicit disabling before, but figured that was due to pcli being part of the public interface with penumbra. does arkworks run our tests under some certain conditions?

crates/test/mock-consensus/src/lib.rs
cratelyn: 137eb1719

crates/test/mock-consensus/src/builder/init_chain.rs
cratelyn: i was having difficulty with this, i'd love to figure out how to recover this too.

`anyhow::Context` seems to want a `'static` bound that isn't included in `BoxError`. if i try to call `context`, i see:

```
error[E0599]: the method `context` exists for enum `Result<&mut C, Box<dyn Error + Send + Sync>>`, but its trait bounds were not satisfied
  --> crates/test/mock-consensus/src/builder/init_chain.rs:38:14

   = note: the following trait bounds were not satisfied:
           `Box<dyn std::error::Error + Send + Sync>: anyhow::context::ext::StdError`
           which is required by `Result<&mut C, Box<dyn std::error::Error + Send + Sync>>: anyhow::Context<&mut C, Box<dyn std::error::Error + Send + Sync>>`
```

crates/test/mock-consensus/src/builder/init_chain.rs
cratelyn: https://github.com/penumbra-zone/penumbra/pull/3807#discussion_r1487863623

crates/core/app/tests/common/mod.rs
hdevalence: Really we should file a bug upstream.

PR #3808, opened 2024-02-13
Add cometbft-addr to docker-compose.yml

AlbaRossa:
pd-node have to know correct address JSON-RPC of cometbft node to work correctly gRPC request to 8080 will fail if pd can't send request to cometbft node RPC port


conorsch:
We aren't using compose in our deployments, but glad the reference has been useful to you. Merging for consistency with the other CLI options, e.g. `tcp://pd-node0:26658`.

AlbaRossa:
Thanks. Well now you know that the docker-compose variant works fine too )

PR #3810, opened 2024-02-13
tests: ✨ provide app state in genesis request

cratelyn:
this is a follow-on to #3807. this makes the
`mock_consensus_can_send_an_init_chain_request` test pass.

see #3588.

add an app state field to the builder, thread that through to the abci request. log this after the response is received.

```
; RUST_LOG="penumbra-app=debug,info" cargo test --package penumbra-app init_chain -- --nocapture

running 1 test
  2024-02-13T15:18:01.770473Z  INFO cnidarium::storage: snapshot dispatcher task has started
    at crates/cnidarium/src/storage.rs:208

  2024-02-13T15:18:01.880291Z  INFO penumbra_app::server::consensus: genesis state is a full configuration
    at crates/core/app/src/server/consensus.rs:143

  2024-02-13T15:18:01.883007Z  INFO penumbra_app::server::consensus: finished init_chain, consensus_params: Params { block: Size { max_bytes: 1, max_gas: 1, time_iota_ms: 1 }, evidence: Params { max_age_num_blocks: 1, max_age_duration: Duration(1s), max_bytes: 1 }, validator: ValidatorParams { pub_key_types: [] }, version: Some(VersionParams { app: 1 }), abci: AbciParams { vote_extensions_enable_height: None } }, validators: [], app_hash: RootHash("5fc87d8a020365f0bc2cff7b1e8fc5dc3d14f862a7c860dc231e6d44fa847277")
    at crates/core/app/src/server/consensus.rs:153

  2024-02-13T15:18:01.883067Z  INFO mock_consensus: finished init chain, hash: [5F, C8, 7D, 8A, 2, 3, 65, F0, BC, 2C, FF, 7B, 1E, 8F, C5, DC, 3D, 14, F8, 62, A7, C8, 60, DC, 23, 1E, 6D, 44, FA, 84, 72, 77]
    at crates/core/app/tests/mock_consensus.rs:30

test mock_consensus_can_send_an_init_chain_request ... ok
```

### DIFF
--- a/crates/core/app/tests/mock_consensus.rs
+++ b/crates/core/app/tests/mock_consensus.rs
@@ -7,24 +7,27 @@ mod common;
 
 use cnidarium::TempStorage;
 use penumbra_app::server::consensus::Consensus;
+use penumbra_genesis::AppState;
 
 #[tokio::test]
-async fn mock_consensus_can_send_a_failing_init_chain_request() -> anyhow::Result<()> {
+async fn mock_consensus_can_send_an_init_chain_request() -> anyhow::Result<()> {
     // Install a test logger, and acquire some temporary storage.
     let guard = common::set_tracing_subscriber();
     let storage = TempStorage::new().await?;
 
     // Instantiate the consensus service, and start the test node.
-    use penumbra_mock_consensus::TestNode;
-    let consensus = Consensus::new(storage.as_ref().clone());
-    let engine = TestNode::builder()
-        .single_validator()
-        .app_state(() /*genesis::AppState::default()*/)
-        .init_chain(consensus)
-        .await;
+    let engine = {
+        use penumbra_mock_consensus::TestNode;
+        let app_state = AppState::default();
+        let consensus = Consensus::new(storage.as_ref().clone());
+        TestNode::builder()
+            .single_validator()
+            .app_state(app_state)
+            .init_chain(consensus)
+            .await?
+    };
 
-    // NB: we don't expect this to succeed... yet.
-    assert!(engine.is_err(), "init_chain does not return an Ok(()) yet");
+    tracing::info!(hash = %engine.last_app_hash_hex(), "finished init chain");
 
     // Free our temporary storage.
     drop(storage);
--- a/crates/core/app/tests/mock_consensus.rs
+++ b/crates/core/app/tests/mock_consensus.rs
@@ -6,6 +6,7 @@
 mod common;
 
 use cnidarium::TempStorage;
+use common::BuilderExt;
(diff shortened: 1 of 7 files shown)


crates/core/app/tests/mock_consensus.rs
```diff
@@ -7,24 +7,27 @@ mod common;
...
+        TestNode::builder()
+            .single_validator()
+            .app_state(app_state)
+            .init_chain(consensus)
+            .await?
+    };
 
-    // NB: we don't expect this to succeed... yet.
-    assert!(engine.is_err(), "init_chain does not return an Ok(()) yet");
+    tracing::info!(hash = %engine.last_app_hash_hex(), "finished init chain");
```
cratelyn, line 30: :tada:

crates/test/mock-consensus/src/builder.rs
```diff
@@ -26,9 +30,9 @@ impl Builder {
         self
     }
 
-    pub fn app_state(self, _: ()) -> Self {
-        // this does not do anything yet
-        self
+    pub fn app_state(self, app_state: AppState) -> Self {
```
hdevalence, line 33: This should be taking a Vec<u8>, so that the mock engine code doesn't build in Penumbra dependencies. Instead, we'll want to have a penumbra_app_state method in an extension trait like we sketched while pairing that does the serialization and calls this method with the resulting bytes.

hdevalence (changes requested):
We should be careful to keep the mock engine API penumbra-agnostic and keep the penumbra specific parts in the extension trait we scoped out while pairing.

crates/test/mock-consensus/src/builder.rs
cratelyn: sure!

to explain why this didn't include that trait originally, i had been hoping to add the extension trait in a distinct PR following this one. i'll go ahead and add it now, to preserve that agnosticism.

i also added a note about this as a design principle to #3588, to write that down somewhere. i'll mention it in some crate-level documentation once we're a little further along into porting tests.

see: 17cfd12

**NB:** using `Into<Bytes>` as a parameter ended up panning out a little more nicely, the commit message there goes into a little more detail.

hdevalence: Awesome, all good, it just seemed important to get the split in from the beginning rather than trying to tease it apart afterwards

hdevalence (approved):

crates/test/mock-consensus/src/builder.rs
cratelyn: agreed, it is difficult to undo that sort of coupling after the fact.

marking as resolved!

PR #3821, opened 2024-02-14
testnet: tmp feature branch to double check for no state breaking changes (do not merge)

erwanor:
There shouldn't be any, but as a matter of process, since the testnet sync takes less than an hour we might as well double check.


erwanor:
Syncing node - will report back in

erwanor:
I checked that main + PR syncs, and pass CI.

PR #3824, opened 2024-02-14
[wip] proto: change BalancesResponse to have Views

hdevalence:


grod220 (dismissed):

grod220:
There are various import re-arranging and code formatting that it appears RustRover did along the way. Can revert those if it's beyond the scope of this, but it makes sense to me.

hdevalence (commented):
LGTM

Issue #3777, opened 2024-02-08
dex(swap_claim): make proof verification faillible


erwanor:
**Is your feature request related to a problem? Please describe.**
When a swap claim is broadcasted to a node, stateless checks will be run during `CheckTx` to verify that it is a valid action before including it in the node's mempool.

In the case of `SwapClaim`s, we first verify that the proof it embeds is correct. In the `SwapClaimProof::verify` logic, [there are `expect` statements](https://github.com/penumbra-zone/penumbra/blob/v0.65.0-alpha.1/crates/core/component/dex/src/swap_claim/proof.rs#L390) enshrining assumptions about the public inputs of the proof. But since we don't check that those assumptions are true ahead of performing the verification, we run the risk of getting user submitted input be able to halt specific nodes.

redshiftzero:
Looking for similar issues, the same problem exists in:
* the `DelegatorVote` verification [here](https://github.com/penumbra-zone/penumbra/blob/main/crates/core/component/governance/src/delegator_vote/proof.rs#L334-L358) 
* the `SpendProof` verification [here](https://github.com/penumbra-zone/penumbra/blob/main/crates/core/component/shielded-pool/src/spend/proof.rs#L326) when we decompress the randomized verification key

The `Swap`, `Convert`, `NullifierDerivation`, and `Output` proofs are ok.

cratelyn:
#3829 addresses the `SwapClaimProof::verify` panics. i'll handle the others that @redshiftzero [found](https://github.com/penumbra-zone/penumbra/issues/3777#issuecomment-1934456026) in subsequent pr's.

cratelyn:
#3833 and #3834 remove panics from `DelegatorVoteProof::verify` and `SpendProof::verify`, respectively.

PR #3829, opened 2024-02-15
dex: 🏒  `SwapClaimProof::verify` is fallible

cratelyn:
see https://github.com/penumbra-zone/penumbra/issues/3777 for more information. this replaces the `Result::expect` statements with error variants that are propagated to the caller.

:heart_decoration: some gardening is done along the way to make this code a little more visually consistent. refactoring changes are intentionally kept in separate commits, to facilitate ease of review.


redshiftzero (approved):
looks great!

PR #3831, opened 2024-02-15
Fix delegation token regex to check boundaries

### LINKED ISSUE #3827: staking: delegation tokens parsing is too permissive
Zellic auditors have found that our delegation token parsing is too permissive. Indeed, since the regex does not specify `^` it will match anywhere in the denom string. This means that an IBC deposited asset that includes this denom string can then usurp voting on proposals, and possibly even more nefarious uses.

```rust
    let validator_identity =
        Regex::new("udelegation_(?P<data>penumbravalid1[a-zA-HJ-NP-Z0-9]+)")
             .expect("regex is valid")
             .captures(&base_denom.to_string())
             .ok_or_else(|| {
                 anyhow::anyhow!(
                     "base denom {} is not a delegation token",
                     base_denom.to_string()
                 )
             })?
```

**Solution**
Parsing should be stricter, both DRI and reviewer should use this as an opportunity to carefully check other load bearing regex in the codebase.

zbuc:
Closes #3827

### DIFF
--- a/crates/core/component/stake/src/delegation_token.rs
+++ b/crates/core/component/stake/src/delegation_token.rs
@@ -63,7 +63,7 @@ impl TryFrom<asset::Metadata> for DelegationToken {
         // Note: this regex must be in sync with both asset::REGISTRY
         // and VALIDATOR_IDENTITY_BECH32_PREFIX
         let validator_identity =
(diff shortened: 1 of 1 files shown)


hdevalence (approved):
LGTM

PR #3833, opened 2024-02-15
governance: 🎾 `DelegatorVoteProof::verify` is fallible

cratelyn:
see https://github.com/penumbra-zone/penumbra/issues/3777.

this changes `DelegatorVoteProof::verify` so that it is fallible. errors will be propagated to the caller, rather than inducing a panic.

* #3777
* #3829

as with #3829, i've applied some tlc to the code as i've addressed the core issue at hand (_panics_). noop code transformations are applied in distinct commits, to facilitate ease of review.

PR #3834, opened 2024-02-15
shielded-pool: 🏊 `SpendProof::verify` is fallible

cratelyn:
see https://github.com/penumbra-zone/penumbra/issues/3777.

this changes `DelegatorVoteProof::verify` so that it is fallible. errors will be propagated to the caller, rather than inducing a panic.

* #3777
* #3829
* #3833

as with #3829 and #3833, i've applied some tlc to the code as i've addressed the core issue at hand (_panics_). noop code transformations are applied in distinct commits, to facilitate ease of review.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5404,6 +5404,7 @@ dependencies = [
  "rand_core 0.6.4",
  "serde",
  "serde_json",
+ "tap",
  "tendermint",
  "thiserror",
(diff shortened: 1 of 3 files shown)


cratelyn:
conflict from #3835, i'll fix in the morning.

redshiftzero (approved):
LGTM!

PR #3836, opened 2024-02-15
docs: stipulate support for precompiled binaries

conorsch:
This is a common question in the Discord, from folks trying to install the precompiled binaries on an LTS OS like Ubuntu 20.04. We only build one set of binaries per architecture, and we don't want to start building per-OS, so let's just be clear about what we know works.


cratelyn (approved):
thank you for doing this, @conorsch!

PR #3857, opened 2024-02-20
mock-consensus: 💰 execute a spend via mock consensus

### LINKED ISSUE #3788: tests: 💰 perform a spend using mock consensus engine
replace the `spend.rs` test with an integration test that runs via the mock consensus engine.

* #3857
* #3866

#3857 added a test that (attempts to) spend a note.

#3866 rephrased our mock consensus spend test to use `TransactionPlan`. on `main` prior to that PR, and after that change, we see an error like this:

```

Caused by:
    Invalid signature.
```

let's track this down so that the spend action is performed successfully.

cratelyn:
fixes #3788.

this introduces a new `mock_consensus_can_send_a_spend_action` test which, as the name implies, shows that the mock consensus harness can send and execute a `Spend` action via our consensus service.

a `MockClient::with_sync_to_storage` helper (bikesheds are welcome here) is introduced, to facilitate syncing to the latest snapshot of the temporary storage.

* #3588
* #3788

### DIFF
--- a/crates/core/app/tests/mock_consensus.rs
+++ b/crates/core/app/tests/mock_consensus.rs
@@ -6,7 +6,10 @@
 mod common;
 
 use {
+    anyhow::anyhow,
     cnidarium::TempStorage,
+    penumbra_keys::test_keys,
+    penumbra_mock_client::MockClient,
     penumbra_sct::component::clock::EpochRead,
     tracing::{error_span, Instrument},
 };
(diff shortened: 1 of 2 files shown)


cratelyn:
> **hdevalence [penumbra]**
> i think the planner API won't be immediately helpful here, since we're using the mock client, not a full view server. i think i described some bits of this in the comment i left describing how the rewritten tests would look. 
>
> the only thing that seems suspicious is the chain id. otherwise: expiry height 0 means no expiry, and default fee is 0, which checks out?

for posterity: feedback from #status in discord.

we should use `TransactionPlan`. i'm going to handle that in a follow-on, and have altered the description of the pr so that it won't close out #3788. i've also opened #3864 as a little tweak, along the way.

PR #3864, opened 2024-02-22
transaction: 🚪 reëxport `TransactionPlan`

cratelyn:
as the crate's documentation points out, `TransactionPlan` is one of the critical types provided by this library.

let's `pub use` it in `lib.rs`, alongside `Transaction`.


cratelyn:
whoops, wrong label! should be `C-chore`.

PR #3866, opened 2024-02-22
mock-consensus: ⭐ use transaction plan to add spends

cratelyn:
* #3588.
* #3788.
* #3857


cratelyn:
i continued to push this along today.

```

Caused by:
Invalid signature.
```

for tomorrow: i'm currently seeing this error in my logs. noticeably, this doesn't propagate into an error. i'll investigate more tomorrow.

crates/core/app/tests/mock_consensus.rs
```diff
@@ -78,71 +82,40 @@ async fn mock_consensus_can_send_a_spend_action() -> anyhow::Result<()> {
...
-            binding_sig,
-            anchor: sct.root(),
-        };
-        <Transaction as penumbra_proto::DomainType>::encode_to_vec(&transaction)
-    };
+        .ok_or_else(|| anyhow!("mock client had no note"))?
+        .tap(|note| tracing::info!(?note, "mock client note commitment"));
+
+    // XXX(kate): this should probably move into the mock-client library, rather than living here
+    // as an extension trait. for now, this is defined here to avoid dealing with dependencies.
```
hdevalence, line 101: I think it'd be better to leave the plan creation inline here, and later evaluate how to use it with the `Planner` API.

I don't think we want to create a general-purpose `spend_tx` method on the `MockClient`; it's not possible to make a general purpose API like that and starting down that path will lead us to another half-baked implementation of a `Planner` that's only used for tests.

cratelyn: sounds good. the latest commit 81eea7755 moves this inline.

PR #3870, opened 2024-02-22
fix(pd): re-add cors headers

### LINKED ISSUE #3865: pd: cors headers not being served (regression)
**Describe the bug**
Back in https://github.com/penumbra-zone/penumbra/issues/3281 we modified `pd` to serve permissive cors headers directly. Some time in the intervening period that behavior was lost. First reported by @grod220 yesterday. 

**To Reproduce**
1. Run pd locally on your machine.
2. Run `curl -I http://localhost:8080`

**Expected behavior**
Displayed headers should include:

```
access-control-allow-headers: *
access-control-allow-methods: *
access-control-allow-origin: *
```

**Additional context**
We can temporarily patch the hosting config to re-apply these headers. That's what I've done as a stopgap to unblock the web team. Those ad-hoc changes will be dropped on the next CI deploy, though, so we should try to fix the problem in pd.

conorsch:
This change restores the serving of CORS headers in HTTP responses by pd. During the refactor of auto-https logic in #3652, we overlooked that the tower [CorsLayer](https://docs.rs/tower-http/0.4.4/tower_http/cors/struct.CorsLayer.html#) was no longer being carried through after type conversions between tower, tonic, and axum. Simply moving the layer attachment to the already-built axum router, rather than the previously constructed tonic router, resolves the problem.

We now include an integration test to check for the headers, so we don't inadvertently drop them again.

Closes #3865.

### DIFF
--- a/.github/workflows/smoke.yml
+++ b/.github/workflows/smoke.yml
@@ -31,9 +31,12 @@ jobs:
       - name: Display comet logs
         if: always()
         run: cat deployments/logs/comet.log
-      - name: Display pd logs
+      - name: Display pd runtime logs
         if: always()
         run: cat deployments/logs/pd.log
+      - name: Display pd test logs
+        if: always()
+        run: cat deployments/logs/pd-tests.log
       - name: Display pclientd logs
         if: always()
         run: cat deployments/logs/pclientd.log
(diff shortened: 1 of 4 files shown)


cratelyn (approved):
it was a pleasure to track down this mystery together today! nice work :slightly_smiling_face:

Issue #3859, opened 2024-02-21
shielded-pool: dummy spends can grief nullifiers


erwanor:
**Is your feature request related to a problem? Please describe.**
It is possible to mark arbitrary nullifiers as "spent" by creating a note with value 0. This will cause the `SpendCircuit` to treat it as a dummy spend which will lift make constraint check  to always pass ([should_enforce](https://github.com/penumbra-zone/penumbra/blob/efe8c79112416771db6f53c24ddeb68f97f77591/crates/core/component/shielded-pool/src/spend/proof.rs#L186-L187)). Since the stateful check pass, the action gets executed and the nullifier is marked as spent.

@hdevalence suggested that we modify the check to only skip the merkle integrity check if the note is a dummy spend.

Initial suggestion for how to fix this:

- [x] #3868
- [x] #3869

redshiftzero:
Initial suggestion for how to fix this:
1. `SpendPlan::dummy()` should use valid data instead of random,
2. the circuit should only short-circuit on the merkle proof verification,
3. the rust code that corresponds to the circuit code should also check everything except for the merkle path,
4. we need to get a spend auth sig for this 0-valued spend.

aubrika:
adding to Sprint 0 as unplanned work, targeted for 2/23

avahowell:
The underlying issue is fixed as of #3871

PR #3871, opened 2024-02-23
shielded-pool(spend): fix validation of dummy spends in spend proof

avahowell:
Instead of short circuiting all checks in the spend circuit in the case of a dummy spend (a note with value 0), this PR changes the spend circuit so that it only skips the SCT inclusion proof check. This addresses bug reported by Zellic in #3859 and ensures the integrity of spends with zero value.

I also added a test that reproduces the bug in main, and verifies that the proof verifier rejects dummy spends with invalid proofs.


redshiftzero (approved):
looks great!

cronokirby (approved):

Issue #3876, opened 2024-02-23
mock-consensus: 1️⃣ report non-zero status codes as errors


cratelyn:
in #3866, i noticed that a non-zero status code in the response is not reported as an error in tests. see this log:

```
2024-02-23T02:55:09.755882Z  INFO penumbra_app::server::consensus: deliver_tx failed, e: binding signature failed to verify

Caused by:
Invalid signature.
at crates/core/app/src/server/consensus.rs:210
in penumbra_mock_consensus::abci::deliver_tx
in penumbra_mock_consensus::block::execute with height: 1, time: 1708656909
```

we should make sure that this results in an error when calling `BlockBuilder::execute`.

hdevalence:
No, I think that's the correct behavior — DeliverTx is fallible. Blocks can include transactions that fail to execute (and this has to be the case, because prior to PrepareProposal there's no way for the application to control transaction inclusion).

cratelyn:
do you think that this should include the `BlockBuilder::execute` helper though? what i'm worried about with this issue is that, as part of my work on #3866, i noticed that the spend transaction was failing silently.

i expect that in the majority of "happy path" cases, we would want to know about a transaction that failed to execute, right?

for test cases that do want to send an invalid transaction to the app, they could still rely on the [`abci`](https://github.com/penumbra-zone/penumbra/blob/main/crates/test/mock-consensus/src/abci.rs) submodule, and send the `deliver_tx` request themselves, inspecting the response to confirm that e.g. it was ruled invalid as expected.

hdevalence:
I'm not sure — and I'm not sure we should overthink it, since if we move to current ABCI this interface all changes anyways.

One question I have is, why is it important for a test to know whether the transaction succeeded or not? Wouldn't we want to be writing assertions about the _effects_ of the transaction, rather than reliance on a self described success code? And in that case do we need this anyways?

For instance, in the spend case, wouldn't we want to write an assertion that, following the block, the nullifier is present in the state, that the MockClient detects the new output note, etc?

cratelyn:
> [..] why is it important for a test to know whether the transaction succeeded or not? Wouldn't we want to be writing assertions about the _effects_ of the transaction, rather than reliance on a self described success code? And in that case do we need this anyways?
> 
> For instance, in the spend case, wouldn't we want to write an assertion that, following the block, the nullifier is present in the state, that the MockClient detects the new output note, etc?

[...]

PR #3891, opened 2024-02-26
app: 💸 a mock consensus spend is performed

### LINKED ISSUE #3788: tests: 💰 perform a spend using mock consensus engine

cratelyn:
fixes #3788.

this branch builds upon the work in #3857 and #3866, filling out the remaining holes in the transaction plan. this now performs a spend, using the mock consensus engine.

### DIFF
--- a/crates/core/app/tests/mock_consensus.rs
+++ b/crates/core/app/tests/mock_consensus.rs
@@ -8,15 +8,22 @@ mod common;
 use {
     anyhow::anyhow,
     cnidarium::TempStorage,
+    penumbra_fee::Fee,
     penumbra_keys::test_keys,
     penumbra_mock_client::MockClient,
+    penumbra_mock_consensus::TestNode,
     penumbra_proto::DomainType,
     penumbra_sct::component::clock::EpochRead,
-    penumbra_shielded_pool::SpendPlan,
-    penumbra_transaction::TransactionPlan,
+    penumbra_shielded_pool::{OutputPlan, SpendPlan},
+    penumbra_transaction::{
+        memo::MemoPlaintext,
+        plan::{CluePlan, DetectionDataPlan, MemoPlan},
+        ActionPlan, Transaction, TransactionParameters, TransactionPlan, WitnessData,
+    },
+    rand_core::OsRng,
     std::ops::Deref,
     tap::Tap,
-    tracing::{error_span, Instrument},
+    tracing::{error_span, info, Instrument},
 };
 
 /// Exercises that a test node can be instantiated using the consensus service.
@@ -80,13 +87,13 @@ async fn mock_consensus_can_send_a_spend_action() -> anyhow::Result<()> {
     let guard = common::set_tracing_subscriber();
     let storage = TempStorage::new().await?;
     let mut test_node = common::start_test_node(&storage).await?;
-    let mut rng = <rand_chacha::ChaChaRng as rand_core::SeedableRng>::seed_from_u64(0xBEEF);
 
     // Sync the mock client, using the test account's full viewing key, to the latest snapshot.
     let (viewing_key, spend_key) = (&test_keys::FULL_VIEWING_KEY, &test_keys::SPEND_KEY);
     let client = MockClient::new(viewing_key.deref().clone())
         .with_sync_to_storage(&storage)
-        .await?;
+        .await?
+        .tap(|c| info!(client.notes = %c.notes.len(), "mock client synced to test storage"));
 
     // Take one of the test account's notes...
     let (commitment, note) = client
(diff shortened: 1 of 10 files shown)


crates/core/app/tests/mock_consensus.rs
```diff
@@ -80,54 +87,90 @@ async fn mock_consensus_can_send_a_spend_action() -> anyhow::Result<()> {
...
+        .await?
+        .tap(|c| info!(client.notes = %c.notes.len(), "mock client synced to test storage"));
 
-    client.notes.get(&commitment).unwrap();
+    // Show that we performed the spend as expected.
+    assert_eq!(
+        client_after_spend.notes.len(),
+        client.notes.len() + 1,
+        "a new note should exist after performing the spend",
+    );
```
cratelyn, line 173: cc: @hdevalence

in our [previous conversation](https://github.com/penumbra-zone/penumbra/issues/3876#issuecomment-1961951033) in #3876, we came to the decision that we should make some assertions here to show that the spend was performed properly.

i'm not entirely familiar with the underlying model w.r.t `Note`s. i have _an_ assertion, above, but is there more that we should be asserting here?

crates/core/app/tests/mock_consensus.rs
```diff
@@ -80,54 +87,90 @@ async fn mock_consensus_can_send_a_spend_action() -> anyhow::Result<()> {
...
-        let spend = SpendPlan::new(&mut rng, note.clone(), position);
+            .witness(commitment.clone())
+            .ok_or_else(|| anyhow!("commitment is not witnessed"))?;
+        let note = note.clone();
+        let position = proof.position();
+        SpendPlan::new(&mut OsRng, note, position).into()
+    };
+
+    // Build a transaction spending this note.
+    let tx: Transaction = {
```
cratelyn, line 117: a passing thought, but this part of the API is pretty long-winded. we talked a bit about this in our sprint planning meeting today, but i'd love to see a more convenient way to go about this for future test authors.

hdevalence: The API we'd like to use is the `Planner` in the view crate. We'd need to be able to pass the `MockClient` to its plan() method: https://rustdoc.penumbra.zone/main/penumbra_view/struct.Planner.html#method.plan

cratelyn: x-ref #3896, i think?

hdevalence, line 117: Rather than doing this, I think it would be cleaner to write
```
let plan = TransactionPlan {
...
};
let tx = plan.build_concurrent().await;
```
The focus should be on the plan, the transaction is a side effect.
```

crates/core/app/tests/mock_consensus.rs
hdevalence: Added in 54c0b35d8 :

- we create pre- and post- snapshots of the chain, and check that the nullifiers were correctly revealed
- we check the mock client for the exact output note commitment we're expecting

crates/core/app/tests/mock_consensus.rs
hdevalence: tidied in 54c0b35d8 , it seemed easier to do it by example than to do line-by-line

hdevalence (approved):
This looks great! I think it would be good to deduplicate a bit with the tests we already have, after we get this merged.

crates/core/app/tests/mock_consensus.rs
hdevalence: Clarification post 54c0b35d8 , where the transaction plan creation becomes much simpler: my point is that for simple tests, where we can write down the complete transaction plan explicitly, we should do so. In this case, we can, which is what 54c0b35d8 does.

 If we need to do more complex test logic, test logic that is complex enough that we cannot simply write down the plan, then we should decide on a way to use the `Planner` API to plan transactions. This may involve pulling in the full rust view server implementation, or continuing to use the `MockClient`. But for simple cases it's important to be able to test the chain logic without all the complexity of the rust view server.

crates/core/app/tests/mock_consensus.rs
cratelyn: you resolved this in https://github.com/penumbra-zone/penumbra/pull/3891/commits/54c0b35d82802c08b3a48cc2a1008c69951eaa9b.

crates/core/app/tests/mock_consensus.rs
cratelyn: you resolved this in https://github.com/penumbra-zone/penumbra/pull/3891/commits/54c0b35d82802c08b3a48cc2a1008c69951eaa9b.

Issue #3894, opened 2024-02-26
Incorrect feature flagging broke wasm build


hdevalence:
**Describe the bug**

Currently, the wasm code in the extension fails to build, with errors like:

![image](https://github.com/penumbra-zone/penumbra/assets/44879/0dfbb618-616d-413f-8222-30f51b15c58f)

I believe the problem here is that since we moved _all_ of the wasm code, we are no longer properly checking that our code builds correctly. 

Two pieces of context to recall:

#### Component separation

The structure of the Penumbra application code is roughly as follows. Application logic is divided into distinct crates called "components", akin to Cosmos SDK "modules" (but we don't use the m-word, since it already has a meaning in Rust). In each crate, some parts of the code are used both by clients and by the fullnode, while other logic is only used by the fullnode. For instance, the action definitions are used by the client, while the actual action handler is only used on the fullnode. 

Each component crate has a `component` submodule feature-gated by a `component` feature. All of the async code must live in the `component` submodule. This way, the crate can be built _without_ the component feature and without pulling in any Tokio deps, async code, etc.

#### Wasm Cleanup

Our wasm code was originally supposed to be "just bindings", building our set of crates and exposing functions to JS/TS users.  However, as we built the web extension, we also ended up having to implement extension-specific logic. This caused problems because we had extension code living outside of the extension repo, and testing and changing it became painful and difficult.

To fix this, we moved all of the wasm code out of the core protocol repo and into the web repo.  (Along the way, there was also discussion of splitting up the wasm code into "extension logic" and "just bindings", but this didn't happen, potentially for good reasons, I don't know).

#### Unexpected Effects

The effect of this change, however, was that we now aren't checking that the core Penumbra crates will build correctly when the `component` feature is disabled.

There are two problems to solve here, an instant-term (now) problem and a longer-term problem:

- [ ] Fix the current compile errors and allow the code to be compiled in wasm.
  To do this, I think the workflow would look like:
  - Check out web repo
  - Check out core repo in an adjacent directory
  - Change web repo's wasm crate to have `path = "../.. ..."` specifiers to the local checkout of core
  - Change core repo until web repo compiles (use the web repo as the build mechanism)
  - Change version string in workspace `Cargo.toml` to `0.68.0-alpha.2`
  - Merge PR with resulting changes into core
  - Push `v0.68.0-alpha.2` tag to core
  - In web repo, change deps from `path = ..` to `tag = ...` with the new tag
[...]

PR #3895, opened 2024-02-26
staking: remove `component::FP_SCALING_FACTOR` dup

erwanor:
This should fix the wasm build unless something else is wrong. The problem was that we defined the same constant (`FP_SCALING_FACTOR`) twice, once correctly at the crate-level, and another one in `penumbra_stake::component`. During development, the latter one was used and created a discreet dependency on the component module.

x-ref #3894


hdevalence:
Before merging, can we

- [ ] Check that this works with the extension's wasm build (#3894) cc @turbocrime 
- [ ] Bump the workspace version so we can immediately do a new tag

turbocrime:
made one minor change, building successfully with a dep on this branch, will PR in web repo in a sec

turbocrime:
Screenshot 2024-02-26 at 12 15 57

turbocrime:
Screenshot 2024-02-26 at 12 22 25
jk it works

hdevalence:
@turbocrime oh great, is that screenshot also indicating that the updated proof params work with `testnet-preview`?

turbocrime:
> @turbocrime oh great, is that screenshot also indicating that the updated proof params work with `testnet-preview`?

they work enough to make a send tx, yes

Issue #3838, opened 2024-02-16
Transaction Perspective + View support for equivalent values


hdevalence:
**Is your feature request related to a problem? Please describe.**

Recently, we changed the `ValueView` proto to add support for "equivalent values": https://github.com/penumbra-zone/penumbra/pull/3824

This allows a view server to give out view data that includes numeraire information inside each `ValueView`, letting us estimate a USD amount of tokens or express a delegation token in terms of the staking token, etc.

However, although this is in the `ValueView`, we didn't fill in any of the data pipeline needed to actually use it. Prices vary over time, so we want a solution where a price is input in one place and filled out across an entire `ValueView`. The natural place for this is in the `TransactionPerspective`.

**Describe the solution you'd like**

- [ ] Extend the `TransactionPerspective` to allow specifying prices. Off the top of my head, I'd suggest something like
  ```proto
  message TransactionPerspective {
      message Price {
          AssetId priced_asset = 1;
          AssetId numeraire = 2;
          // Multiply units of the priced asset by this value to get the value in the numeraire.
          double numeraire_per_unit = 3;
      }
      repeated Price prices = ...;
   ```
   Motivation: we're rendering values, not trying to compute exact prices, so a double is actually appropriate here and much easier to work with for callers. We distinguish the `priced_asset` from the `numeraire` for clarity and so that the price is "directional": we want to show GM in terms of USDC but not the other way around.
- [ ] Change the transaction view generation code so that every time we construct a `ValueView` (by augmenting an `AssetId` with the appropriate `Metadata` from the perspective), we also use the perspective's `prices` list to fill in equivalent values: build a map of `priced_asset => (numeraire, numeraire_per_unit)` and fill in all relevant prices.
- [ ] Extend the `TransactionPerspective` to allow specifying `extended_metadata` using [an `Any` type](https://docs.rs/pbjson-types/latest/pbjson_types/struct.Any.html)
- [ ] Change the transaction view generation code to fill in [the `extended_metadata`](https://buf.build/penumbra-zone/penumbra/docs/main:penumbra.core.asset.v1#penumbra.core.asset.v1.ValueView.KnownAssetId).

hdevalence:
Context on `extended_metadata` and why it's useful: it allows use cases like

- adding `ValidatorInfo` into a delegation token
- adding position reserves into an LPNFT

PR #3898, opened 2024-02-27
Add support for equivalent values to `ValueView`

hdevalence:
I'm not sure if the `EquivalentPrice` should also be defined in the asset proto package rather than alongside the TxP? I filled in an implementation that inserts equivalent values into `TransactionView`s but I realized that because it relies heavily on the TxP it may not be directly useful for the web extension use cases, which so far seem mostly about balances (?).

However, while doing that I realized that in the context of viewing a _transaction_, it doesn't really make sense to just have an "equivalent value" on its own, so I added an `as_of_height` to allow indicating when the prices were relevant.


hdevalence:
Closes #3838 whenever we work out what the right design is

grod220 (approved):
I think this makes sense. As we are scanning through blocks, we'll use the new `prices: Vec<EstimatedPrice>` available in the transaction and save these to our DB. 

Then, upon request for balances, we'll fill out the ValueViews by querying the DB & applying some filtering (specific numeraire, recency/confidence, etc) and making the conversion like you did in the `view_value` method.

hdevalence:
> I think this makes sense. As we are scanning through blocks, we'll use the new `prices: Vec<EstimatedPrice>` available in the transaction (the node provides this right?) and save these to our DB.

Well, I'm not 100% sure if we want to do this, since it gets to the problem that asks for an `as_of_height`: when viewing a transaction, what are the right prices to display? The historical ones? The current ones? How would we convey the time of the price in the UI? 

I think this would need a bit more product thinking than we've done so far, it seems like a bit of a rabbithole, so we might be better off not doing anything for now and only focusing on the other parts of the RPC (like balances) which are inherently "current" data. But that's exactly the use case the current state of this PR *doesn't* address, since it only works on transactions.

> Then, upon request for balances, we'll fill out the ValueViews by querying the DB & applying some filtering (specific numeraire, recency/confidence, etc) and making the conversion like you did in the `view_value` method.

Right, the issue I was pointing to is that (in the current state of this branch) this isn't possible, since the equivalent values can only be generated in the context of a `Transaction`.

hdevalence:
I updated the PR to make it possible to use the estimated prices independently of the `TxP` mechanism. I think it's good to go now.

The breaking change notifications from the proto lint is a false positive, it's comparing to the previous run rather than comparing to the pre-PR state.

grod220:
> I updated the PR to make it possible to use the estimated prices independently of the TxP mechanism. I think it's good to go now.

On the web side of things, it seems we'd just need to adjust the price sourcing (versus relying upon tx). For example, we'd have to do something like:
- Scan through blocks
- If `swapOutputs` detected, loop through them and save swap to database
   - Table: "SWAP_OUTPUTS"
   - Fields:
     - in: Amount (delta1 - lambda1)
     - out: Amount (delta2 - lamda2)
     - height: BigInt
     - tradingPair: TradingPair
- On Balances rpc call, query db

hdevalence:
> > I updated the PR to make it possible to use the estimated prices independently of the TxP mechanism. I think it's good to go now.
> 
> On the web side of things, it seems we'd just need to adjust the price sourcing (versus relying upon tx). For example, we'd have to do something like:
> 
>     * Scan through blocks
> 
>     * If `swapOutputs` detected, loop through them and save swap to database
>       
>       * Table: "SWAP_OUTPUTS"
>       * Fields:
>         
>         * in: Amount (delta1 - lambda1)
>         * out: Amount (delta2 - lamda2)
>         * height: BigInt
>         * tradingPair: TradingPair
> 
>     * On Balances rpc call, query db

@grod220 I think https://github.com/penumbra-zone/web/issues/556 is a more general design

PR #3900, opened 2024-02-27
app: 🤠 address warning

cratelyn:
addresses this compiler warning:

```
warning: argument must be of the form: `expected = "error message"`
  --> crates/core/app/tests/spend.rs:97:1
   |
97 | #[cfg_attr(
   | ^
   |
   = note: errors in this attribute were erroneously allowed and will become a hard error in a future release
```

### DIFF
--- a/crates/core/app/tests/spend.rs
+++ b/crates/core/app/tests/spend.rs
@@ -96,7 +96,7 @@ async fn spend_happy_path() -> anyhow::Result<()> {
 #[tokio::test]
 #[cfg_attr(
     debug_assertions,
-    should_panic("assertion failed: cs.is_satisfied().unwrap()")
+    should_panic = "assertion failed: cs.is_satisfied().unwrap()"
 )]
 async fn invalid_dummy_spend() {
     let mut rng = rand_chacha::ChaChaRng::seed_from_u64(1312);


erwanor (commented):
```rust
    // 3. Simulate execution of the Spend action
    spend
        .check_stateless(transaction_context)
        .await
        .unwrap_err()
        .to_string()
        .contains("spend proof did not verify");
```

This should be wrapped in an `assert!` right?

cratelyn:
> ```rust
>     // 3. Simulate execution of the Spend action
>     spend
>         .check_stateless(transaction_context)
>         .await
>         .unwrap_err()
>         .to_string()
>         .contains("spend proof did not verify");
> ```
> 
> This should be wrapped in an `assert!` right?

i'm not sure. i didn't write this test, it was added in #3871. i just wanted to address the warning i'm seeing when building `main`

crates/core/app/tests/spend.rs
```diff
@@ -96,7 +96,7 @@ async fn spend_happy_path() -> anyhow::Result<()> {
 #[tokio::test]
 #[cfg_attr(
     debug_assertions,
-    should_panic("assertion failed: cs.is_satisfied().unwrap()")
+    should_panic = "assertion failed: cs.is_satisfied().unwrap()"
```
erwanor, line 99: Not a huge deal, but reading the last state of the proof test, I am not sure I understand why we expect this to panic? the attribute makes me think it might be a debug assert nested deep in arkworks? If so we should probably add a comment

erwanor:
Ah yeah, np I'll ask @avahowell

crates/core/app/tests/spend.rs
avahowell: This is cursed, if you try to verify an invalid proof _in debug mode_, arkworks will panic with this error. So we expect this to panic in debug mode, in release mode it should error with an invalid spend proof. The way the test is written currently with contains() is a bug

PR #3597, opened 2024-01-09
feat: 📝 implement cometstub crate (work-in-progress)

cratelyn:
see #3588.

```
/!\ ------------------------------------------------------- /!\
/!\ NOTE: this is a rolling work-in-progress.               /!\
/!\ this branch will be force-pushed frequently until it is /!\
/!\ ready for review. proceed accordingly!                  /!\
/!\ ------------------------------------------------------- /!\
```

PR #3902, opened 2024-02-27
mock-consensus: 🌱 single genesis validator

### LINKED ISSUE #3816: tests: ⭐ mock consensus can `InitChain` with a single validator
see #3588.

with #3787 complete, we can proceed to the next phase of work on the mock consensus testing framework: we should implement the logic to patch an `AppState` with the information for a single test validator.

see this todo comment: https://github.com/penumbra-zone/penumbra/blob/7d61991df55ad0039b433af237a8e2080ff65a34/crates/core/app/tests/common/mod.rs#L73-L77

cratelyn:
fixes #3816.

this provides extension facilities to the mock consensus test node
builder, which allows penumbra-app tests to define a single validator,
and subsequently retrieve it from the chain state.

see `mock_consensus_can_define_a_genesis_validator`. when run with
`--no-capture` enabled, these logs will be visible:

```
2024-02-28T23:00:43.751036Z DEBUG penumbra_stake::component::stake: adding validator identity to consensus set index, validator: penumbravalid172v76yyqwngcln2dxrs8ht0sjgswer3569yyhezgsz6aj97ecvqqyf3h9h
at crates/core/component/stake/src/component/stake.rs:533
in penumbra_stake::component::stake::staking

[...]

2024-02-28T23:00:43.776880Z  INFO penumbra_app::server::consensus: genesis state is a full configuration
at crates/core/app/src/server/consensus.rs:145

2024-02-28T23:00:43.780436Z DEBUG penumbra_app::app: finished committing state, jmt_root: RootHash("46dc0e9561f17eee61a2c13f517036d4d0a4c77c60362cb6cc165083675dcaf7")
at crates/core/app/src/app/mod.rs:592
```

logging facilities are provided so that helper warnings should be given
to users that forget to call `with_penumbra_single_validator`, or
provide an `AppState` object whose validator list would be overwritten.

the `serde_json` dependency is removed from the mock consensus library,
it is no longer used.

a warning is added to the mock consensus library to note to future
contributors that other penumbra dependencies should be avoided in that
library.

a new `http::Extensions` field is added to the builder. it is not used
internally, but provides extension traits a place to hold additional
state.

* #3588
* #3816


crates/test/mock-consensus/Cargo.toml
```diff
@@ -10,6 +10,12 @@ license.workspace = true
 [dependencies]
 anyhow = { workspace = true }
 bytes = { workspace = true }
+decaf377-rdsa = { workspace = true }
+ed25519-consensus = { workspace = true }
+penumbra-proto = { workspace = true }
+penumbra-keys = { workspace = true }
```
cratelyn, line 16: note to self: remove these tomorrow!

crates/core/app/tests/common/mod.rs
```diff
@@ -83,19 +87,134 @@ impl TempStorageExt for TempStorage {
...
+
+impl<C> TestNodeExt for TestNode<C> {
+    //  TODO(kate): `penumbra_stake::IdentityKey` could be `From<[u8; 32]>`
+    fn penumbra_identity_key(&self) -> penumbra_stake::IdentityKey {
+        use penumbra_proto::core::keys::v1 as pb;
+        use prost::Message;
+        pb::IdentityKey::decode(self.identity_key().as_slice())
+            .unwrap()
+            .try_into()
+            .unwrap()
```
cratelyn, line 218: this is where i am getting a panic rn,

`called `Result::unwrap()` on an `Err` value: DecodeError { description: "invalid tag value: 0", stack: [] }`

i need to do this dance slightly differently, probably by referring directly to the `decaf377` type instead of using `prost::Message`. i'll figure that out tomorrow!

crates/core/app/tests/mock_consensus.rs
```diff
@@ -37,6 +40,32 @@ async fn mock_consensus_can_send_an_init_chain_request() -> anyhow::Result<()> {
...
+    let identity_key = test_node.penumbra_identity_key();
+    match storage
+        .latest_snapshot()
+        .get_validator_state(&identity_key)
+        .await?
+        .ok_or_else(|| anyhow!("genesis validator state was not found"))?
+    {
+        penumbra_stake::validator::State::Active => {}
+        other => panic!("unexpected genesis validator state, found: {other}"),
+    }
```
cratelyn, line 60: small question for @erwanor: do genesis validators start in `Active`?

crates/test/mock-consensus/src/builder.rs
```diff
@@ -5,12 +5,17 @@
...
+    crate::TestNode, bytes::Bytes, penumbra_proto::penumbra::core::component::stake::v1 as pb,
+    rand_core::OsRng,
+};
 
 /// A buider, used to prepare and instantiate a new [`TestNode`].
 #[derive(Default)]
 pub struct Builder {
-    app_state: Option<Bytes>,
+    pub app_state: Option<Bytes>,
+    pub validators: Vec<pb::Validator>,
```
cratelyn, line 17: todo: this should be a type that isn't `pb::Validator`. it also might melt into the extension trait, since this is feeding `with_penumbra_auto_app_state`.

in any case, another problem to sort out tomorrow!

crates/core/app/tests/mock_consensus.rs
```diff
@@ -37,6 +39,33 @@ async fn mock_consensus_can_send_an_init_chain_request() -> anyhow::Result<()> {
...
+    match storage
+        .latest_snapshot()
+        .get_validator_state(&identity_key)
+        .tap(|_| info!(?identity_key, "getting validator state"))
+        .await?
+        .ok_or_else(|| anyhow!("genesis validator state was not found"))?
+    {
+        penumbra_stake::validator::State::Active => info!("genesis validator is active"),
+        other => panic!("unexpected genesis validator state, found: {other}"),
+    }
```
cratelyn, line 60: :rocket:

crates/core/app/tests/mock_consensus.rs
hdevalence: I think we could make it always the case, it seems simplest. There's no reason to define a validator in the genesis file if it's not going to be immediately participating, after all.

crates/core/app/tests/common/mod.rs
```diff
@@ -47,7 +51,7 @@ pub async fn start_test_node(storage: &TempStorage) -> anyhow::Result<PenumbraTe
     let app_state = AppState::default();
     let consensus = Consensus::new(storage.as_ref().clone());
     TestNode::builder()
-        .single_validator()
+        .with_penumbra_single_validator()
         .with_penumbra_auto_app_state(app_state)?
```
hdevalence, line 55: What's the relation between these two methods?

crates/core/app/tests/common/mod.rs
```diff
@@ -83,19 +87,130 @@ impl TempStorageExt for TempStorage {
...
+                ik: bytes.to_vec().clone(),
+            }),
+            governance_key: Some(GovernanceKey {
+                // NB: for now, we will use the same key for governance. See the documentation of
+                // `GovernanceKey` for more information about cold storage of validator keys.
+                gk: bytes.to_vec().clone(),
+            }),
+            consensus_key: {
+                let signing_key = ed25519_consensus::SigningKey::new(OsRng);
+                signing_key.as_bytes().as_slice().to_vec()
```
hdevalence, line 189: This is publishing the validator's signing (private) key, rather than their verification (public) key. Presumably this isn't causing problems at the moment because we're not actually producing valid headers, but we should rethink this API to better handle key management.

hdevalence:
I think we should take the opportunity to iterate on the API a little bit and figure out how we want to manage validator keys. It might be good to do that in chat or a slightly more synchronous setting than GH issues.

crates/core/app/tests/common/mod.rs
hdevalence: xref https://github.com/penumbra-zone/penumbra/pull/3902#issuecomment-1971674791

crates/core/app/tests/common/mod.rs
hdevalence: xref https://github.com/penumbra-zone/penumbra/pull/3902#issuecomment-1971674791

cratelyn:
today we held a [discussion](https://discord.com/channels/824484045370818580/1212231915923570778/1212822461973004308) in the #testing channel about this PR.

this comment strives to record the guiding design principles outlined in that
conversation.

### 🤯 why doesn't the original plan work?

> what the relation is between `BuilderExt::with_penumbra_single_validator()`
> and `BuilderExt::with_penumbra_auto_app_state()`?
> 
> I think we should take the opportunity to iterate on the API a little bit and
> figure out how we want to manage validator keys. It might be good to do that
> in chat or a slightly more synchronous setting than GH issues.

\- <https://github.com/penumbra-zone/penumbra/pull/3902#issuecomment-1971674791>
\- <https://github.com/penumbra-zone/penumbra/pull/3902#discussion_r1507968266>

our [original][pair-1] [idea][pair-2] was that a `TestNode::single_validator()`
method would generate a single set of validator keys, while
`with_penumbra_auto_app_state` would:

> 1. read out list of abci/comet validators from the builder
> 2. define a penumbra validator for each one
> 3. inject that into the penumbra app state
> 4. serialize to json and then call `with_app_state_bytes`

#3597 was drafted during a very early stage of this work. we have since made a
lot of progress and crucially, an important design goal has since come into
focus: the mock consensus engine should not depend on other `penumbra-*`
crates, to avoid circular dependencies and to offer a reusable testing tool.

that last point in particular complicates things. it is difficult to store a
value inside of a library without allowing that library to know what you are
providing it.

because of that, the builder must rely on a complex [extension
trait][builder-ext] in order to store, and later retrieve, a list of penumbra
validators in `BuilderExt::with_penumbra_auto_app_state()` while avoiding the
addition of a `penumbra-proto` dependency.

while this "works" it is solving a problem that we can avoid causing.

[pair-1]: https://github.com/penumbra-zone/penumbra/pull/3597/files#diff-4b1c924f57b5a531a0be51a9975a35280b53ba9c3f8019a01839859a171aa9a1
[pair-2]:https://github.com/penumbra-zone/penumbra/pull/3597/files#diff-ad36d14674139876ba9059526375edc55e2718509b01e913c64c574a78ba10b9R30-R33
[builder-ext]: https://github.com/penumbra-zone/penumbra/blob/3499c2d3d320263f3a56767d6f048dad7dc7c94c/crates/core/app/tests/common/mod.rs#L102-L203

### 💭 brief review, useful notes

first, some background:

* penumbra's `AppState` contains validator definitions. these definitions
  control the tendermint consensus keys, and sets the voting power based on the
  number of delegation tokens each validator has.

* we can retrieve the validator definitions at any point, but we cannot
  retrieve the validators' identity keys.

* there are two classes of `penumbra-mock-consensus` users; **(1)** only
  applies to the staking and governance components, **(2)** is the common case.
  * **(1)** users who **do** care about simulating validator operator actions
  * **(2)** users who do **not** need to act as a validator operator

* penumbra and cometbft have distinct notions of a "validator". none of the
  penumbra application code uses the ed25519 consensus keys; these are used by
  cometbft. the internal representation of a penumbra validator includes
  identity and governance keys.

### 😌 the new plan

#### 🔨 generate definitions in tests...

the builder does not need to store validators; users in group **(1)** can
construct the app state and pass it into the engine, using helper functions
to generate validator definitions.

this should be easy to achieve. `with_penumbra_single_validator` is already
in effect a `Fn() -> Validator`, it does not need to be defined as an
extension trait method.

#### 🔑 ...then define a keyring

initially, we will discard the consensus keys of our generated validator(s).
in order to test the staking and governance components however, we will need
to provide a keyring, so that the engine can sign blocks.

```rust
type Keyring = BTreeMap<VerificationKey, SigningKey>;

impl<C> TestNode<C> {
    pub fn keyring(&self) -> &Keyring { todo!() }
    pub fn keyring_mut(&self) -> &mut Keyring { todo!() }
}
```

this should be in service to tests of the staking component that:

> 1. create the validator definition for the new validator, generating identity
>    and consensus keys
> 1. insert the consensus signing key into the engine’s keyring
> 1. simulate a tx that posts that validator definition
> 1. when the app reports the new validator to the consensus engine, the engine
>    can look up its key in the keyring and start using it for signing blocks

### 🤠 yeehaw!

tomorrow i will take these points, and use them to refine our API. i am going
to mark this PR as a draft for now, and mark it as ready for review once more
when these changes are complete.

i sense some unknowns involving the fine details of how the tendermint/penumbra
views of validator generation will interact, but i'll save further questions
there for the moment.

hdevalence:
Awesome, this is a great summary of the discussion, thanks for writing it up!

One point of potential clarification (or else overcommunication): with regard to

> our original idea was that a TestNode::single_validator()
method would generate a single set of validator keys, while
with_penumbra_auto_app_state would: ...

We should still do this as the (immediate) next steps of the current plan to serve use case (2), and we may not even need any helper methods — we can just generate and insert the validator definitions in a for loop inside the auto_app_state method, exactly as in the original plan. In other words, the takeaway from the requirements discovery is that we don't need to change the existing plan for the use case we are currently concerned with.

Later, we can discover requirements for use case (1) by actually writing some of those tests.

The mock engine's keyring won't be needed until we actually sign blocks, so if we're not currently doing that, we won't need to do it immediately.

crates/test/mock-consensus/src/builder.rs
```diff
@@ -5,12 +5,16 @@
...
+    crate::{keyring::Keys, TestNode},
+    bytes::Bytes,
+};
 
 /// A buider, used to prepare and instantiate a new [`TestNode`].
 #[derive(Default)]
 pub struct Builder {
-    app_state: Option<Bytes>,
+    pub app_state: Option<Bytes>,
+    pub keys: Option<Keys>,
```
cratelyn, line 17: NB: im avoiding defining a map, and the other keyring related interfaces, for now. i hope to save that for a follow-on PR.

crates/core/app/tests/common/test_node_builder_ext.rs
cratelyn, line 1: i avoided decomposing the rest of `common/mod.rs` for now, but decided it was time to pull this out into its own lil' submodule.

crates/core/app/tests/common/test_node_builder_ext.rs
```diff
@@ -0,0 +1,120 @@
...
+            Some(validator) => app_state_with_validators(app_state, validator)?,
+            None => app_state,
+        };
+
+        // Serialize the app state into bytes, and add it to the builder.
+        serde_json::to_vec(&app_state)
+            .map_err(Self::Error::from)
+            .map(|s| self.app_state(s))
+    }
+}
```
cratelyn, line 43: :point_up: here's the meat of it

crates/test/mock-consensus/Cargo.toml
```diff
@@ -7,10 +7,15 @@ repository.workspace = true
 homepage.workspace = true
 license.workspace = true
 
+# NB: to avoid circular dependencies: do not add other `penumbra-*` crates
+# as dependencies, below. we should provide interfaces in terms of generic
+# types, and provide penumbra-specific helper functions via extension traits.
```
cratelyn, line 12: added a note here, just so this does not live in our respective minds :brain:

crates/test/mock-consensus/src/keyring.rs
```diff
@@ -0,0 +1,24 @@
+pub struct Keys {
+    pub consensus_signing_key: ed25519_consensus::SigningKey,
+    pub consensus_verification_key: ed25519_consensus::VerificationKey,
+}
```
cratelyn, line 4: i avoided the map (see [above](https://github.com/penumbra-zone/penumbra/pull/3902/files#r1509543713)) so for now i'm stuffing the signing/verification keys into a structure. we can revisit this when we want multiple validators etc.

crates/core/app/tests/common/mod.rs
cratelyn: marking this as resolved via https://github.com/penumbra-zone/penumbra/pull/3902#issuecomment-1972239643 and the referenced conversation in #testing.

we found a better way forward.

(_and also importantly, the verification key is used. thanks for calling that out!_)

crates/core/app/tests/common/mod.rs
cratelyn: xref https://github.com/penumbra-zone/penumbra/pull/3902#issuecomment-1972239643

we talked this through and found a better way to do this!

crates/test/mock-consensus/src/keyring.rs
cratelyn: also of note: identity keys are stubs atm, but eventually the identity keys should be put here too.

crates/core/app/tests/common/test_node_builder_ext.rs
```diff
@@ -0,0 +1,120 @@
...
+    Keys {
+        consensus_verification_key,
+        ..
+    }: &Keys,
+) -> PenumbraValidator {
+    /// A temporary stub for validator keys.
+    ///
+    /// NB: for now, we will use the same key for governance. See the documentation of
+    /// `GovernanceKey` for more information about cold storage of validator keys.
+    const BYTES: [u8; 32] = [0; 32];
```
hdevalence, line 87: This isn't a valid key, which may cause problems. 

We should instead generate a random signing key, get its verification key, use that for the identity key, and throw the signing key away after.

crates/core/app/tests/mock_consensus.rs
```diff
@@ -37,6 +38,42 @@ async fn mock_consensus_can_send_an_init_chain_request() -> anyhow::Result<()> {
     Ok(())
 }
 
+/// Exercises that the mock consensus engine can provide a single genesis validator.
+#[tokio::test]
+async fn mock_consensus_can_define_a_genesis_validator() -> anyhow::Result<()> {
+    // Install a test logger, acquire some temporary storage, and start the test node.
+    let guard = common::set_tracing_subscriber();
+    let storage = TempStorage::new().await?;
+    let _test_node = common::start_test_node(&storage).await?;
```
hdevalence, line 47: We should avoid doing this, so that it's possible to reason about each test in isolation. If the initialization code is so complicated we can't copy paste it between tests we should try to simplify it so that it can be.

crates/test/mock-consensus/src/keyring.rs
hdevalence: Per previous discussions about the scope of auto_app_state, I don't think we'll ever need to keep those keys — tests that want to have access to the identity keys can construct their own app state.

crates/test/mock-consensus/src/keyring.rs
```diff
@@ -0,0 +1,24 @@
...
+
+impl Keys {
```
hdevalence, line 6: We should really try to avoid creating something called just "keys", we have lots of different types of keys and it's important to be clear about what they are and what they do.

crates/test/mock-consensus/src/builder.rs
hdevalence: In this case it seems unclear what we're saving, not defining a map means having to define a custom Keys structure; if this were a BTreeMap<VerificationKey, SigningKey> there would be no extra structure to define and it would also be the data we actually want to use.

crates/core/app/tests/common/test_node_builder_ext.rs
hdevalence: It's a little hard for me to reason about what the code is doing. We want to:

1. Generate a list of penumbra validator definitions corresponding to the consensus keys specified in the builder. There must be at least one.

2. Inject those validator definitions into the provided AppState

3. Create allocations of delegation tokens for each validator, controlled by some address. Otherwise the validator won't have any voting power and will be immediately deleted.

Could we make it more explicit?

crates/core/app/tests/common/test_node_builder_ext.rs
```diff
@@ -0,0 +1,120 @@
...
+        }
+    }
+}
+
+/// Generates a [`Validator`][PenumbraValidator] given a set of consensus [`Keys`].
+fn generate_penumbra_validator(
+    Keys {
+        consensus_verification_key,
+        ..
+    }: &Keys,
```
hdevalence, line 81: Can we pass the consensus key directly here?

crates/core/app/tests/common/test_node_builder_ext.rs
cratelyn: e66628697 among other things, rewrites this function. i think now that should more clearly reflect steps (1) and (2)!

crates/core/app/tests/common/test_node_builder_ext.rs
cratelyn: addressed in 91bc6e72c

crates/core/app/tests/common/test_node_builder_ext.rs
cratelyn: i addressed this in 718fecff073986ab58a5de02ecc4f3676bef2124, in the form of some comments. i agree this will cause problems, but i am explicitly interested in having a failing test in hand before i add the identity key.

crates/test/mock-consensus/src/builder.rs
cratelyn: e66628697 addresses this

crates/test/mock-consensus/src/keyring.rs
cratelyn: e66628697 addresses this

crates/test/mock-consensus/src/keyring.rs
cratelyn: e66628697 addresses this

cratelyn (commented):
:speech_balloon: added some commits to address review comments. most importantly, `Keys` is replaced with a `Keyring`. in doing so, we also have addressed (most of) #3934!

PR #3903, opened 2024-02-27
Check value balance during IBC packet timeout

### LINKED ISSUE #3773: ics20: value balance is not updated after timeout
**Problem description**
The timeout logic doesn't update the ICS20 value balance in one of the timeout handling branches (see: [shielded-pool(component): `transfer.rs:L474`](https://github.com/penumbra-zone/penumbra/blob/58f14d3fe08517a95ae9cdeddb4623e349f81927/crates/core/component/shielded-pool/src/component/transfer.rs#L474-L488)).

A malicious user could potentially drive the value balance to zero and block egress from the host chain.

zbuc:
Once the mock test framework is further along, it would be good to add some tests around this logic.

Closes #3773

### DIFF
--- a/crates/core/component/shielded-pool/src/component/transfer.rs
+++ b/crates/core/component/shielded-pool/src/component/transfer.rs
@@ -261,7 +261,7 @@ async fn recv_transfer_packet_inner<S: StateWrite>(
     //
     // https://github.com/cosmos/ibc/tree/main/spec/app/ics-020-fungible-token-transfer (onRecvPacket)
     //
-    // NOTE: spec says proto but thsi is actualy JSON according to the ibc-go implementation
+    // NOTE: spec says proto but this is actually JSON according to the ibc-go implementation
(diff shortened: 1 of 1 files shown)


crates/core/component/shielded-pool/src/component/transfer.rs
```diff
@@ -478,6 +477,14 @@ async fn timeout_packet_inner<S: StateWrite>(mut state: S, msg: &MsgTimeout) ->
             new_value_balance,
         );
     } else {
+        let value_balance: Amount = state
+            .get(&state_key::ics20_value_balance(
+                &msg.packet.chan_on_a,
```
avahowell, line 482: this is correct, the channel should be the same as the other branch since channel `a` or `b` refers to what side of the bidirectional state machine you're on. In this case chain `a` is the one that sent a packet and is now receiving a timeout, and chain `b` is the one that never committed the packet.

avahowell:
lgtm

PR #3907, opened 2024-02-28
fix: replace atty with `std::io::IsTerminal`

kevinji:
Resolves GHSA-g98v-hv3f-hcfr.

`std::io::IsTerminal` has been present since [Rust 1.70](https://blog.rust-lang.org/2023/06/01/Rust-1.70.0.html).


cratelyn (approved):
thank you for doing this @kevinji!

PR #3909, opened 2024-02-28
Add intra-transaction duplicate position-id check

### LINKED ISSUE #3855: dex: treat position `Id`s like nullifiers
The DEX stateful check are too permissive, they allow anyone to create [the same position multiple times in the same transaction](https://github.com/penumbra-zone/penumbra/blob/efe8c79112416771db6f53c24ddeb68f97f77591/crates/core/component/dex/src/component/action_handler/position/open.rs#L32-L37), which in turns would grant multiple `LpNft`s for the same position Id. It is a really severe bug that has been demonstrated to be useful in mounting a vuln chain that leads to DEX inflation. 

<blockquote>
I don't think this needs any redesign of the execution model. We can already handle this in execute, which is fallible for this reason — and is how IBC actions are handled.

In the case of this specific bug, there are two possible things that could have been done:

- Treat position IDs as akin to nullifiers (check all in parallel and then also check no overlaps within transactions).

- Have the "add a new position" method check that the position does not exist when writing it out.

In terms of a general solution, I think what we need is to have written justifications for each of the checks we do and why stateless/stateful/execution is the right place for them to live.
</blockquote>

_Originally posted by @hdevalence in https://github.com/penumbra-zone/penumbra/issues/3847#issuecomment-1955121794_

Linked issue contains helpful context

cronokirby:
Closes #3855.

Also adds an extra sanity check that opened positions are in the `Opened` state.

### DIFF
--- a/crates/core/component/dex/src/component/action_handler/position/open.rs
+++ b/crates/core/component/dex/src/component/action_handler/position/open.rs
@@ -32,14 +32,13 @@ impl ActionHandler for PositionOpen {
         Ok(())
     }
 
-    async fn check_stateful<S: StateRead + 'static>(&self, state: Arc<S>) -> Result<()> {
-        // Validate that the position ID doesn't collide
-        state.check_position_id_unused(&self.position.id()).await?;
-
+    async fn check_stateful<S: StateRead + 'static>(&self, _state: Arc<S>) -> Result<()> {
         Ok(())
     }
 
     async fn execute<S: StateWrite>(&self, mut state: S) -> Result<()> {
+        // Validate that the position ID doesn't collide
+        state.check_position_id_unused(&self.position.id()).await?;
         state.put_position(self.position.clone()).await?;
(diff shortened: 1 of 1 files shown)


cronokirby:
(I have a test related to this, but pending planner support, unless someone feels like manually doing gm and gn spends to create reserves)

hdevalence:
Can we add commentary to the check_stateful calls that justifies their presence in that method? 

The suggestion in the linked issue was to do the simple thing of moving the check into execute(), if we want to do a performance optimization here instead, we should add commentary to the check_stateful() impl justifying its structure and alerting future devs to the fact that the additional check is load-bearing.

cronokirby:
Oh, I didn't realize that this was a performance optimization for spends; I guess given that I expect no honest party will actually open conflicting positions, it seems fine to do this in execute; will amend.

erwanor (approved):

PR #3917, opened 2024-03-01
Add unfilled field to SimulateTradeResponse

grod220:
See: https://github.com/penumbra-zone/web/issues/566

Would be nice if clients did not have to calculate this on their own by doing the math manually.


crates/core/component/dex/src/component/rpc.rs
```diff
@@ -541,7 +543,13 @@ impl SimulationService for Server {
             .await
             .map_err(|e| tonic::Status::internal(format!("error simulating trade: {:#}", e)))?;
 
+        let unfilled = Value {
+            amount: input.amount - swap_execution.input.amount,
```
erwanor, line 547: This all looks good but can we make this use `checked_sub` and log the error, just in case we turn up bugs in `SwapExecution`.

erwanor (approved):

crates/core/component/dex/src/component/rpc.rs
grod220: Ah, good call. Will do.

PR #3923, opened 2024-03-01
staking: use block height to enforce unbonding delay

### LINKED ISSUE #3738: staking: rework the unbonding delay mechanism
Delegation pools follow an unbonding schedule determined by the chain to provide a window of opportunity for the consensus layer to propagate evidence of misbehavior (e.g., validator equivocation) back to the application so that it can be punished by the staking component ("byzantine slashing").

Currently, the unbonding schedule is defined in terms of epochs (`StakingParameters::unbonding_epochs`) that must elapse before an undelegation can be claimed or for a pool to become `Unbonded`.

However, since changes in the active validator set trigger epoch changes, it is possible for a consortium of active-set validators to expedite the unbonding period (in terms of the wall clock) by forcing early epoch changes.

**Potential solution.**

1. Make the unbonding schedule a number of blocks:
    - Penalty calculations work the same as before but we do an extra lookup to fetch the end block's epoch index
2. Add a minimum wall clock duration (`X blocks must have elapsed AND Y hours`)
3. Add a synchrony check to the chain to neutralize $f+1$ BFT time manipulation (#3668) 

** Other solutions you have considered **

1. Increase the `unbonding_duration` chain parameter
2. Rely completely on time-based duration (cons: a combination of logical and wall clock is more robust)

erwanor:
Close #3738, this PR makes the unbonding mechanism based on block delay rather than epochs. In practice, this means that a user will wait X blocks for a validator pool to unbond (rather than Y epochs).

To achieve this, this PR:
- defines `StakeParamater::unbonding_delay` measured in blocks
- parameterize unbonding tokens based on a `start_height`
- deprecate `start_epoch_index` fields in delegate/claim actions
- strip `epoch_index` from validator `RateData`s
- generalize the unbonding token denom format: 
`2.013718unbonding_start_at_1540_penumbravalid1` (vs. `_epoch_XXX`)

Important point about the mechanism: we bind tokens to the starting height of the epoch that they belong to. This let us avoid binding transactions to specific block heights.


cratelyn (approved):
i should not be a definitive approval, but i have ready through this and believe that the changes appear to be sound. i really appreciate how you have divided the changes here into distinct commits, it made reviewing this much easier. :star:

erwanor:
Thanks for the review @cratelyn, rebased conflicts, then merging this so we can start work on integration.

cratelyn:
> Thanks for the review @cratelyn, rebased conflicts, then merging this so we can start work on integration.

perfect timing, i was about to ask when this might land. i'm in the neighborhood via #4001. :slightly_smiling_face:

cratelyn:
merged on your behalf :star:

PR #3924, opened 2024-03-02
Add undelegate claims to TransactionPlannerRequest

jessepinho:
Undelegate claims are currently unsupported in `TransactionPlannerRequest`s. This PR fixes that. Relevant changes are [here](https://github.com/penumbra-zone/penumbra/pull/3924/files#diff-03b7341d5bf81ab9c8d8542d220d5ba4ae122bd3837531309742b685d7ec1619); everything else is generated by the protobuf codegen script.

Note that the `UndelegateClaim` message leaves out the `blinding` fields, so that the RPC method implementation can fill those in on behalf of the client. (In the case of the web app, this means that the WASM planner will fill those in, rather than the web app.)

I'm less familiar with the core repo, so please let me know if I've missed anything.


hdevalence (approved):
LGTM!

PR #3929, opened 2024-03-03
refactor(workspace dependency): add feature

TalDerei:
missing feature resulted in failing to call [try_load](https://github.com/penumbra-zone/penumbra/blob/katie/cometstub/crates/crypto/proof-params/src/lib.rs#L55) before attempting to dereference the keys, causing some test failures.


TalDerei:
this should actually just be a dev-dependency for a single crate, otherwise some tests will fail following the recent decaf377 changeset.

Issue #3920, opened 2024-03-01
ics20: improve native asset tracking


erwanor:
**Is your feature request related to a problem? Please describe.**
Native asset deposits cause the ics20 value balance to decrease and token supply to increase. However, withdrawals only increment the ics20 value balance without updating the token supply.

This means that anyone can inflate the token supply of a native asset as much as they want. This is problematic because among native assets are delegation tokens validator voting power is measured in terms of the size of its delegation pool.

We should fix this issue, and generally reconsider how we do token supply tracking since this is error-prone. An alternative approach could be that we build a `Bank` extension trait that components can compose in order to manage their own supply of tokens instead of delegating it to the shielded pool.

hdevalence:
xref https://github.com/penumbra-zone/penumbra/pull/3938#issuecomment-1977323894

The deeper issue here is not just about the ICS-20 accounting but that there is a conflation of two different kinds of token supply: (a) tracking the _current_ supply of tokens within the shielded pool for cross-chain transfers, and (b) tracking the _total_ supply of delegation tokens, used for staking power calculations.

This should be fixed by separating these two responsibilities, rather than just tweaking the ICS20 handler.

erwanor:
Closing in favor of #4018

PR #3938, opened 2024-03-04
ics20: decrease token supply in withdrawal of native assets

avahowell:
This addresses the finding in #3920


hdevalence:
I'm not sure this is safe. The token supply numbers are also used by the staking component right?

I think the problem is the existence of the shielded pool's value tracking mechanism at all. We should have a design meeting about what to do here.

erwanor:
Closing this in favor of having component-scoped supply tracking. Thanks for your work on this, the original ticket should have been clearer.

PR #3952, opened 2024-03-05
pd: 📦 add a tracing event with cargo version

cratelyn:
this adds a small tracing event including the version of pd that is running, before proceeding to handle the CLI command.

this should help operators identify when they are running a stale version of pd.

### DIFF
--- a/crates/bin/pd/src/main.rs
+++ b/crates/bin/pd/src/main.rs
@@ -66,6 +66,7 @@ async fn main() -> anyhow::Result<()> {
         registry.init();
     }
 
+    tracing::info!(?cmd, version = env!("CARGO_PKG_VERSION"), "running command");
     match cmd {
         RootCommand::Start {
             home,


hdevalence:
Maybe I'm missing something — this doesn't actually do what we want, right? That environment variable is set by Cargo when running Cargo, not when running pd...

cratelyn:
> Maybe I'm missing something — this doesn't actually do what we want, right? That environment variable is set by Cargo when running Cargo, not when running pd...

> CARGO_PKG_VERSION — The full version of your package.

https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates

to my understanding, this would log versions like `0.68.1` or `0.68.2` etc., based on the version of the workspace/crate. this _is_ set at compile-time like you describe, but that'd resolve to effectively `let version = "1.2.3";` in the resulting binary.

in the interest of disambiguating whether a stale `pd` is running, that seems like at least a small step in the right direction. i'm happy to revert this though. what do you think @hdevalence?

hdevalence:
My read of that page is that it's set by cargo, not embedded in the binary. But an easy way to test would be to run the binary ./target/release/pd and see.

cratelyn:
> My read of that page is that it's set by cargo, not embedded in the binary. But an easy way to test would be to run the binary ./target/release/pd and see.

i confirmed this works properly with a small test program:

```
; eza -T --ignore-glob target
.
├── Cargo.lock
├── Cargo.toml
└── src
   └── main.rs
; bat Cargo.toml | grep version
version = "0.1.0"
; ./target/debug/env-playground
0.1.0
; sed -i -e 's/0.1.0/1.2.3/' Cargo.toml
; cargo build --quiet
; ./target/debug/env-playground
1.2.3
```

hdevalence:
Ah great! This is much simpler than I'd expected then, thanks for getting it merged in!

PR #3953, opened 2024-03-05
ibc: show client expiry status in channel query

avahowell:
Improves `ibc q channel(s)` so that it will show the status as 'CLIENT EXPIRED' if the underlying client for a channel has expired.


cratelyn (approved):

conorsch:
For comparison's sake, here's what the change looks like.

### Before:

before

###After:
after

Note that the `State` column shows the expiry info.

PR #3906, opened 2024-02-28
pcli: improve pcli ibc channel queries

avahowell:
Closes #3905 

This PR makes the pcli channel queries more useful to the end user, displaying relevant channel metadata (like chain ID) instead of just dumping the protocol level struct to console as JSON.

Before:

Screenshot 2024-02-27 at 17 59 10

After:

Screenshot 2024-02-27 at 18 00 44

PR #3954, opened 2024-03-05
docs: update deimos ibc channel for osmosis

conorsch:
We had a few chain halts on the `v0.68.x` release series. As of `v0.69.0`, "channel-0" is the correct value for relaying to Osmosis testnet. Added mention of the slick new `pcli query ibc channels` to the docs.

Also overhauled the dev docs for how to test IBC functionality interactively, since that knowledge has mostly been communicated via pairing session, but not yet fully written down. We no longer use `rly`, neither for running relayer instances nor for ad-hoc CLI transfers.  Instead, we can use `hermes` for both roles.

Linked out to wiki docs on the high-churn hermes relayer config steps, since we're updating those frequently. Therefore review should also include eyes on https://github.com/penumbra-zone/penumbra/wiki/Updating-Hermes

### DIFF
--- a/docs/guide/book.toml
+++ b/docs/guide/book.toml
@@ -24,3 +24,6 @@ additional-js =["theme/js/mermaid.min.js", "theme/js/mermaid-init.js"]
 # [output.linkcheck]
 # follow-web-links = false
 # warning-policy = "ignore"
+#
+# The buf.build website doesn't support HTTP HEAD calls, returning 405.
+# exclude = ['buf\.build']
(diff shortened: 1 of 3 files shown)


conorsch:
TIL about https://github.com/penumbra-zone/penumbra/pull/3906. We haven't shipped that for 68 yet, but probably should. Regardless, this docs PR should be updated to use that new functionality, because that's the current state of play.

cratelyn (approved):

Issue #3955, opened 2024-03-05
view: 🗼 `Planner::plan_with_spendable_and_votable_notes()` does not reset fee tiers


cratelyn:
https://github.com/penumbra-zone/penumbra/blob/98c42d8f09200ed41771c4dad91e5997bb4a0cab/crates/view/src/planner.rs#L45-L54

there is a comment on `Planner` that notes that new fields should be reset when a plan is produced. this came from aa942e699a in #3647.

we should reset that here: https://github.com/penumbra-zone/penumbra/blob/98c42d8f09200ed41771c4dad91e5997bb4a0cab/crates/view/src/planner.rs#L676-L683

hdevalence:
We should not do this, and do #3896 instead, it's not worthwhile to fix up code we are planning to remove.

PR #3956, opened 2024-03-05
view: 🔭 exhaustive pattern in `Planner` reset

cratelyn:
see #3955.

`Planner::plan_with_spendable_and_votable_notes()` resets all of its fields, **except for** `fee_tier`.

this does not address that "bug", per conversation in discord:

> **kate [penumbra]** i noticed that Planner::plan_with_spendable_and_votable_notes() does not clear the fee tier field. i believe that is a dormant bug. would a PR to fix that be welcomed?
> **hdevalence [penumbra]** i'd hold off on it, i would prefer to throw that code away
> **hdevalence [penumbra]** there should be no clearing of fields at all, for instance

to prevent other such bugs creeping up in the meantime, we add an _exhaustive_ pattern destructuring the `Planner`, so that future fields added to the planner must be contended with.

### DIFF
--- a/crates/view/src/planner.rs
+++ b/crates/view/src/planner.rs
@@ -673,12 +673,26 @@ impl<R: RngCore + CryptoRng> Planner<R> {
 
         tracing::debug!(plan = ?self.plan, "finished balancing transaction");
 
-        // Clear the planner and pull out the plan to return
-        self.balance = Balance::zero();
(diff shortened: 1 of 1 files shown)


cratelyn:
https://github.com/penumbra-zone/penumbra/issues/3955#issuecomment-1979699326

PR #3962, opened 2024-03-06
component: eliminate TOCTOU bug factory

hdevalence:
This renames the previous `check_stateful` method to `check_historical`,
renames `execute` to `check_and_execute`, and by default moves all of the
content of `check_stateful` into `check_and_execute`, unless there's an obvious
and LOCAL reason why it's safe to do so that can be written into the body of
the method.

IMO it's still worth it to have the method in the trait. We may wish to take
advantage of it later, when we can spend more time carefully reasoning about
performance.

### DIFF
--- a/crates/cnidarium-component/src/action_handler.rs
+++ b/crates/cnidarium-component/src/action_handler.rs
@@ -44,33 +44,45 @@ pub trait ActionHandler {
     /// be run in parallel across all transactions in a block.
     async fn check_stateless(&self, context: Self::CheckStatelessContext) -> Result<()>;
 
-    /// Performs all of this action's stateful validity checks.
+    /// Performs those stateful validity checks that can be performed against a
+    /// historical state snapshot.
     ///
     /// This method provides read access to a snapshot of the `State` prior to
-    /// transaction execution.  Because it does not give write access, it's safe
-    /// to run the `ActionHandler` implementations for all actions in a given
-    /// transaction in parallel.
+    /// transaction execution.  It is intended to be run in parallel across all
+    /// actions within a transaction.
     ///
-    /// As much work as possible should be done in `check_stateful`, as it can
-    /// be run in parallel across all actions within a transaction.
+    /// # Warning
+    ///
+    /// Misuse of this method creates TOCTOU vulnerabilities. Checks performed
+    /// in this method must be valid if they are performed against a _prior_
+    /// state, as another action in the same transaction may execute first and
+    /// change the state.
+    ///
+    /// Checks performed in this phase should have a justification for why they
+    /// are safe to run in parallel with other actions in the same transaction,
+    /// and the default behavior should be to perform checks in
+    /// [`ActionHandler::check_and_execute`].
     ///
     /// # Invariants
     ///
     /// This method should only be called on data that has been checked
     /// with [`ActionHandler::check_stateless`].  This method can be called
     /// before [`Component::begin_block`](crate::Component::begin_block).
-    async fn check_stateful<S: StateRead + 'static>(&self, state: Arc<S>) -> Result<()>;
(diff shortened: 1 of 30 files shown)


crates/core/app/src/action_handler/transaction.rs
```diff
@@ -30,11 +30,14 @@ impl ActionHandler for Transaction {
...
         valid_binding_signature(self)?;
-        no_duplicate_spends(self)?;
-        no_duplicate_votes(self)?;
         num_clues_equal_to_num_outputs(self)?;
         check_memo_exists_if_outputs_absent_if_not(self)?;
 
+        // These should be redundant given the choice to move these checks into check_and_execute
+        // but are left here for reference and as a defense in depth.
+        no_duplicate_spends(self)?;
+        no_duplicate_votes(self)?;
```
hdevalence, line 39: Slightly ambivalent on this point -- should we just delete?

cronokirby: point in favor of deleting: people (me) might be tempted to add other such checks here

hdevalence: sounds good, looks like our test suite is checking the exact error message, so we'll need to patch up the tests

crates/core/app/src/action_handler.rs
```diff
@@ -16,26 +16,34 @@ mod transaction;
...
 #[async_trait]
-pub trait ActionHandler {
+pub trait AppActionHandler {
     type CheckStatelessContext: Clone + Send + Sync + 'static;
     async fn check_stateless(&self, context: Self::CheckStatelessContext) -> Result<()>;
-    async fn check_stateful<S: StateRead + 'static>(&self, state: Arc<S>) -> Result<()>;
-    async fn execute<S: StateWrite>(&self, state: S) -> Result<()>;
+    async fn check_historical<S: StateRead + 'static>(&self, _state: Arc<S>) -> Result<()> {
+        return Ok(());
+    }
```
cratelyn, line 31: nice :+1:

crates/core/app/src/action_handler.rs
```diff
@@ -16,26 +16,34 @@ mod transaction;
...
 /// we can delete this trait entirely.
+///
+/// Currently, there are only three impls, all of which are entangled with app-level data:
+///
+/// - ProposalSubmit (which is entangled with the whole-application state)
+/// - Action (which needs to slot in the PenumbraHost for IBC action handling)
+/// - Transaction (which depends on the above)
 #[async_trait]
-pub trait ActionHandler {
+pub trait AppActionHandler {
```
cratelyn, line 26: appreciate the disambiguating name here :+1:

crates/cnidarium-component/src/action_handler.rs
```diff
@@ -44,33 +44,45 @@ pub trait ActionHandler {
...
+    ///
+    /// Misuse of this method creates TOCTOU vulnerabilities. Checks performed
+    /// in this method must be valid if they are performed against a _prior_
+    /// state, as another action in the same transaction may execute first and
+    /// change the state.
+    ///
+    /// Checks performed in this phase should have a justification for why they
+    /// are safe to run in parallel with other actions in the same transaction,
+    /// and the default behavior should be to perform checks in
+    /// [`ActionHandler::check_and_execute`].
```
cratelyn, line 64: appreciate the additional information added in the comments here :heavy_check_mark:

cratelyn (approved):
this is great work :tada:

crates/core/app/src/action_handler/transaction.rs
cronokirby: We probably don't want to just change the error message, but instead modify those tests to actually have the right FMD clues, to actually exercise duplicate nullifiers.

PR #3963, opened 2024-03-06
add compat addresses that use bech32 (non-m) for compatibility with noble IBC

avahowell:
Bech32 (non-m) encoded addresses are the standard in the broader cosmos ecosystem. [Noble requires that withdrawal addresses be encoded as bech32](https://github.com/noble-assets/noble/blob/757b3111d15ef2ee8640639af60edef1b7712c6d/app/ante.go#L155), and attempting to transfer to a bech32m address (such as a penumbra address) results in a decoding error. This issue is acknowledged by the broader ecosystem. In the mean time, we can fix this on our end and enable testing with Noble or other similarly constrained cosmos chains by introducing a 'compat' address encoding `penumbracompat1`, that uses Bech32(non-m). 

This PR adds the compat encoding, adds a pcli flag to `view address` to allow you to see it, and changes the FromStr impl for address to detect and use the compat encoding.

### DIFF
--- a/crates/bin/pcli/src/command/view/address.rs
+++ b/crates/bin/pcli/src/command/view/address.rs
@@ -15,6 +15,9 @@ pub struct AddressCmd {
     /// Output in base64 format, instead of the default bech32.
     #[clap(long)]
     base64: bool,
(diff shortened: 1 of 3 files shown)


hdevalence:
Let's try this on preview and then potentially revert it if needed.

PR #3970, opened 2024-03-07
chore: update chain id to deimos-5

conorsch:
Corresponds with software v0.69.0, to resolve the chain halt caused by poorly coordinated release of `v0.68.3`.


cratelyn (approved):
thank you for doing this! :pray:

Issue #3913, opened 2024-02-29
Design a way to use the Rust view server with mock consensus integration tests


hdevalence:
**Is your feature request related to a problem? Please describe.**

Currently we are trying to expand the capability of our state machine tests, going depth-first: writing a few, targeted tests that incrementally expand the capability of the test harness and allow us to discover requirements.

These tests are intended to exercise functionality end to end, including client logic. Currently we do this with the MockClient, a minimal reference implementation of key Penumrba client logic.

However, a benefit and limitation of this approach is that we cannot automatically generate transaction plans, because the MovkClient cannot be used with a Planner. This is a benefit because we can write tests that are very precise about their state changes and do not rely on a much larger mass of code, but a limitation because more complex test logic would prefer to use the planner.

Additional context here: https://github.com/penumbra-zone/penumbra/pull/3891#discussion_r1503536584

**Describe the solution you'd like**

There are a lot of moving pieces and a lot of context to consider to figure out the right design. @cratelyn and I should pair to explore the best path forward.

#### :link: related work

* #3958
* #3971
* #3973
* #3980
* #3996
* #4349
* #4434
* #4435
* #4447
* #4460
* #4468
* #4469
* #4470
* #4471

cratelyn:
@hdevalence i prepared a scaffold in #3958, taking a small preparatory step towards using `MockClient` with `Planner::plan()`.

cratelyn:
#3986 is loosely related to this work.

cratelyn:
#4081 has been fixed by #4312, so i am planning to pick this work back up.

PR #3971, opened 2024-03-07
view: 🌯 remove frivolous `Option::expect` call

cratelyn:
here, we are within an `if let Some(_)` clause that has already bound the contents of the storage path. so, let's just pass that to `Self::load` instead of unwrapping it again.


cratelyn:
cherry-picked out of a forthcoming branch, re: #3913.

Issue #3972, opened 2024-03-07
refactor(bench): expand our benchmarking suite with finite field and elliptic curve ops


TalDerei:
Expand benchmarks (sections **V and VI** in https://docs.google.com/spreadsheets/d/1DonNNCN6-Nm3SxSpOymkGYOQi2wAK0l3_CfuDPq4vQc/edit#gid=0) with finite field and curve ops comparing the performance between decaf377 and arkworks.

PR #3976, opened 2024-03-07
refactor(bench): field and curve benchmarks

TalDerei:
references #3972


TalDerei:
this will become useful for comparing performance when the 32-bit arkworks backend changes land

TalDerei:
@hdevalence I'll merge this, just flagging for visibility

PR #3973, opened 2024-03-07
mock-consensus: 🪟 use view server in a test

cratelyn:
work in progress.

* #3913
* #3588
* #3980
* #3996

Issue #3697, opened 2024-01-29
pd: regression in http trailers


conorsch:
We observed a breaking change in the local dev env for [web extension](https://github.com/penumbra-zone/web) development this morning, caused by merge of https://github.com/penumbra-zone/penumbra/pull/3679. Although that PR passed smokes and even manual testing via pcli, it broke the web extension, due to the refactor no longer supporting HTTP trailers. 

For now, we'll revert that commit, to unbreak web development immediately. We should also investigate how to improve integration testing between the penumbra monorepo and the latest tip of extension code. 

## Misc
@hdevalence isolated the [failing commit via git bisect](https://discord.com/channels/824484045370818580/930529119165362196/1201639267668074587).

cratelyn:
hyperium/tonic#455 notes two unfinished issues:

> * Decide how to better integrate this functionality with the transport. For now, the GrpcWeb type is designed to decorate services but it doesn't have to be. A different design might not even expose the service and wire everything up internally.
> * **How to better integrate with tonic's router.** To multiplex services, it is necessary to wrap them individually. This is either a feature or a bug.

(_emphasis mine_)

i suspect that the change to use a `Routes` in #3679 was adversely affected by this.

i'll take a look tomorrow, but i plan to keep a tight timebox on attempting to land that PR again, as written. it was addressing a `TODO` that i found while working on auto-https. i may remove that comment, or rewrite it to point here. 

it would be nice to test pd against a grpc-web client, if we do not already. that feels outside the scope of this issue.

thanks for opening this and for reverting that commit, @conorsch! (and thank you to everybody else that helped track this regression :bug: :mag:)

hdevalence:
Another possibility is that we could file an issue upstream, and then keep the commit around to pull up again if it gets resolved -- the new code is certainly nicer but i'm not sure it's critical to land it at the moment.

cratelyn:
note: i filed #3704 to track the addition of integration tests exercising grpc-web compatibility, to keep that conversation separate from this specific regression.

PR #3980, opened 2024-03-08
refactor(pd): 🍱 pull grpc endpoint into a standalone crate

cratelyn:
cherry-picked out of #3973. see #3913, and #3588.

we would like to use the rust view server in mock consensus tests. in order to run the `penumbra_view::ViewServer` however, we need to spin up the corresponding grpc endpoint for it to connect to.

this branch performs a bit of code motion, moving the `grpc_server` out of `pd` and into a standalone library that may be used within our mock consensus tests.

there will likely be other functional changes to the code in question before we can use it in those tests, but this PR is interested in moving that code into a place where our tests can rely upon it.


crates/bin/pd/src/main.rs
```diff
@@ -134,102 +129,14 @@ async fn main() -> anyhow::Result<()> {
...
+                tonic_web::enable(TendermintProxyServiceServer::new(tm_proxy.clone())),
+            );
             if enable_expensive_rpc {
-                grpc_server = grpc_server.add_service(we(SimulationServiceServer::new(
-                    DexServer::new(storage.clone()),
-                )));
+                use penumbra_dex::component::rpc::Server as DexServer;
+                grpc_server = grpc_server.add_service(tonic_web::enable(
+                    SimulationServiceServer::new(DexServer::new(storage.clone())),
+                ));
```
cratelyn, line 139: for simplicity i left these both here. i don't have reason to believe yet that they will be needed by our tests, but they'd be easy to move over later if needed.

cratelyn:
smoke tests are failing:

```
2024-03-08T21:19:33.340400Z ERROR penumbra_view::worker: view worker error e=status: Unknown, message: \"Service was not ready: transport error\", details: [], metadata: MetadataMap { headers: {} }
Error: status: Unknown, message: \"grpc-status header missing, mapped from HTTP status code 405\", details: [], metadata: MetadataMap { headers: {\"content-length\": \"0\", \"access-control-allow-origin\": \"*\", \"vary\": \"origin\", \"vary\": \"access-control-request-method\", \"vary\": \"access-control-request-headers\", \"access-control-expose-headers\": \"*\", \"allow\": \"GET,HEAD\", \"date\": \"Fri, 08 Mar 2024 21:19:32 GMT\"} }
```

i will track that down after the weekend. i believe i know how to fix that, i'll pick it up when i'm back on monday.

crates/bin/pd/src/main.rs
cratelyn: https://github.com/penumbra-zone/penumbra/pull/3980/#issuecomment-1986436183

cratelyn: https://github.com/penumbra-zone/penumbra/pull/3980#issuecomment-1986436183

cratelyn: the order of `add_service` is load-bearing

hdevalence:
Moving this code out of the `pd` crate seems good, but I'm wondering why we want to create a new crate, rather than putting it inside the `app` crate. Didn't we attempt to do that before, and ran into a Tonic bug? Can we avoid that bug while still moving the GRPC services into the `app` crate?

cratelyn:
#### :clapper: re: deja vu

> Didn't we attempt to [move this] before, and ran into a Tonic bug? Can we avoid that bug while still moving the GRPC services into the `app` crate?

good memory, i am glad this sounds eerily familiar to you too! the difference is subtle.

#3679 was trying at the time to address [this](https://github.com/penumbra-zone/penumbra/pull/3679/files#diff-fbc4204ceb976c8cb30ed06168e2476700bae21bfd803e26281b2d026194d430L175-L182) `TODO` comment.

the difference between the two is that #3679 would have `penumbra-app` create a `Routes`, that we would [add](https://github.com/penumbra-zone/penumbra/pull/3679/files#diff-fbc4204ceb976c8cb30ed06168e2476700bae21bfd803e26281b2d026194d430R204) to the builder (_which stays in `pd`_).

here, i'm not trying to make that cut between `Router` and `Routes`, and am attempting to hoist the whole thing out of `pd`, without making changes to how we interact with `tonic`. my aim is for us to be able to move this, without running into that bug (#3697) again.

cratelyn:
closing in favor of #3996! :clapper:

PR #3984, opened 2024-03-09
view(rpc): add `DelegationsByAddressIndex` query

jessepinho:
## Describe your changes
Right now, the `StakingService` has a method, `validatorInfo`, that returns a series of `ValidatorInfo`s.

As of https://github.com/penumbra-zone/web/pull/604, we've refactored the minifront staking page to treat validators as tokens that an address may have a balance of. To achieve that, [we're fetching validator infos, then generating `ValueView`s from those validators](https://github.com/penumbra-zone/web/pull/604/files#diff-9bd971ea2f5dadcc555a9a848d1cea2a0d6df1a0069b9e5e490218058ac5adbc) that include the address's balance of those tokens.

This logic should be moved into a new RPC method on the `ViewService` called `DelegationsByAddressIndex`, since it's a common enough use case that clients shouldn't all have to do it themselves.

Relevant changes are to [`view.proto`](https://github.com/penumbra-zone/penumbra/pull/3984/files#diff-03b7341d5bf81ab9c8d8542d220d5ba4ae122bd3837531309742b685d7ec1619) and [`ViewService`](https://github.com/penumbra-zone/penumbra/pull/3984/files#diff-36188e6ab5083e8be9d29039370cc523cfcfede683b223327b2d90e3f75d40c7); the rest is just codegen changes.

Closes https://github.com/penumbra-zone/web/issues/612

- [x] (n/a) If this code contains consensus-breaking changes, I have added the "consensus-breaking" label.


crates/view/src/service.rs
```diff
@@ -1627,4 +1633,11 @@ impl ViewService for ViewServer {
             denom_metadata: metadata.map(Into::into),
         }))
     }
+
+    async fn delegations_by_address_index(
+        &self,
+        _request: tonic::Request<pb::DelegationsByAddressIndexRequest>,
+    ) -> Result<tonic::Response<Self::DelegationsByAddressIndexStream>, tonic::Status> {
+        unimplemented!("delegations_by_address_index")
```
zbuc, line 1641: Looks like this PR isn't ready yet?

jessepinho: Well, the idea is that it could be implemented by other implementations of `ViewService`, such as in the web app. (Note that `fn authorize_and_build` above is also unimplemented.) I had to add this method here or else the Rust compiler would complain that this method was missing.

jessepinho:
@erwanor I'd accidentally left this as a draft PR — that's prob why! :D

Issue #3813, opened 2024-02-13
Support separate custody (including multisig) for validator definition and validator voting


plaidfinch:
**Is your feature request related to a problem? Please describe.**

The signing authority which controls the identity of a Penumbra validator is exactly a Penumbra spendkey. In the simplest custody setup (a single software wallet), whichever wallet is used to _broadcast_ a validator definition is also the same spend authority used to _authorize_ that update to the validator.

The protocol does not, however, require this conflation; indeed, the design of the validator system for Penumbra is deliberately made so that a validator operator can practice rigorous separation of concerns: a validator's non-custodial funding streams need not remit funds to addresses controlled by the key which controls the validator definition.

One possibly desirable validator configuration would be separated custody for the validator's _funding_ (from reward streams, and optionally self-delegation) and the validator's _identity_. Additionally, it should be possible to custody the validator's _identity_ just as securely (or not) as any other Penumbra spend key, with the choice of custody backend (online software wallet, airgapped software wallet, threshold multisig, hardware wallet) informed by the operator's requirements, not by the limitations of client software.

The above affordances are all permitted by the protocol but not the current client tooling. This issue describes a sequence of steps to resolve all of these problems and permit validator identity custody to be as first-class as any other type of custody in the system.

**Describe the solution you'd like**

In order to close this gap, I propose the following augmentations to the client tooling. Note that none of them require changing any protobuf or RPC definitions in a way that is backwards-incompatible.

1. The first step is to permit separated signing and broadcasting of validator-related actions:
  - [x] The command `pcli validator definition upload` should take an additional argument `--signature` which should be attached to the validator definition, in place of the default locally-generated signature using this `pcli`'s own custody backend.
  - [x] A new command `pcli validator definition sign` should take the same arguments as `upload` (excepting the new addition above), but instead use the _local_ custody backend to produce a signature for that validator definition and output it to the command line.
  - [x] The command `pcli validator vote` should be turned into a subcommand `pcli validator vote cast`, to make room in the hierarchy of commands for other vote actions.
  - [x] The command `pcli validator vote cast` should take an additional argument `--signature` which should be attached to the validator's vote, in place of the default locally-generated signature using this `pcli`'s own custody backend.
  - [x] A new command `pcli validator vote sign` should take the same arguments as `cast` (excepting the new addition above), but instead use the _local_ custody backend to produce a signature for that vote and output it to the command line.

With the above implemented, validator custody can now function across an airgap.

[...]

PR #3985, opened 2024-03-10
Permit validator voting across an airgap

plaidfinch:
## Describe your changes

This addresses the first changeset described in #3813, permitting validator definitions and votes to be signed separately and those signatures to be broadcast from a potentially-unrelated penumbra account.

This is a breaking change to the CLI for all workflows that use the `validator vote` subcommand, because it redefines that command to `validator vote cast` in order to make room for `validator vote sign`. It is purely a client-side change, and as such is not consensus-breaking.

#3813

- [X] If this code contains consensus-breaking changes, I have added the "consensus-breaking" label.

### DIFF
--- a/docs/guide/src/pcli/governance.md
+++ b/docs/guide/src/pcli/governance.md
@@ -67,12 +67,16 @@ received when voting in real life at your polling place.
 ### Voting As A Validator
 
 If you are a validator who was active when the proposal started, you can vote on it using the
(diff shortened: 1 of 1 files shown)


erwanor (approved):
LGTM!

conorsch:
Nit: the docs at https://guide.penumbra.zone/main/pcli/governance.html#voting-as-a-validator should be updated to reflect this change. 

Also, when we do land this, please squash-merge so that it's cherry-pickable. Important as we move toward stable testnets!

PR #3986, opened 2024-03-10
Permit separated governance keys from spend keys

plaidfinch:
## Describe your changes

These changes are meant to be stacked on top of #3985, which implements the first set of three change sets defined in #3813.

The changes in this PR separate out a distinct governance subkey, which can be initialized after main initialization in `pcli` by using `pcli init validator-governance-subkey {soft-kms, threshold}`. This command will edit in place the existing configuration file to add a section describing the custody of a separate governance key, which may be custodied differently than the main key. These changes support soft-kms and threshold custody for governance key.

The generated distinct governance key will then, after initialization, be used instead of the key implied by the spend key, when signing and casting validator votes.

Subsequently, when generating a validator definition template, this key is the one inserted into the template, and it can be fetched directly using the `pcli validator governance-key` command.

Being merely client-side, this set of changes does not break consensus.

#3813

- [X] If this code contains consensus-breaking changes, I have added the "consensus-breaking" label.

### DIFF
--- a/crates/bin/pcli/src/command/init.rs
+++ b/crates/bin/pcli/src/command/init.rs
@@ -11,14 +11,14 @@ use rand_core::OsRng;
 use url::Url;
 
 use crate::{
-    config::{CustodyConfig, PcliConfig},
+    config::{CustodyConfig, GovernanceCustodyConfig, PcliConfig},
     terminal::ActualTerminal,
 };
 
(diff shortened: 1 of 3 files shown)


erwanor:
I think the smoke test should be fixed by rebasing onto main once #3985 gets merged.

PR #3989, opened 2024-03-10
tests(pcli): 🔇 make large tracing event trace-level

cratelyn:
there are two issues at hand:
- we log `undelegation_result` twice
- `undelegation_result` can be many thousands of lines long

this means that the `delegate_and_undelegate` pcli network integration test thrashes the test output somewhat severely, and test logs aren't fully viewable in e.g. tmux, when running this locally.

https://github.com/penumbra-zone/penumbra/actions/runs/8210700115/job/22458602964#step:11:18

as of `main` at the time of writing, this CI job logs an event that is...

```
; xclip -o -selection clipboard | wc --bytes
243775
```

243Kb is disruptively long. (_let alone 486Kb!_)

so, let's make this field an opt-in field to display. we only log it once, and do so at the trace level.

### DIFF
--- a/crates/bin/pcli/tests/network_integration.rs
+++ b/crates/bin/pcli/tests/network_integration.rs
@@ -323,8 +323,7 @@ fn delegate_and_undelegate() {
             .timeout(std::time::Duration::from_secs(TIMEOUT_COMMAND_SECONDS));
         let undelegation_result = undelegate_cmd.assert().try_success();
 
-        tracing::debug!("undelegation done - output");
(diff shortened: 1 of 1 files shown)


erwanor (approved):
+1, in fact we could remove it completely, it was used to debug spurious errors

cratelyn:
> +1, in fact we could remove it completely, it was used to debug spurious errors

nice! i removed it in 85fa01b, thank you for the suggestion!

PR #3999, opened 2024-03-11
ci: public ingress for postgres db

conorsch:
External developers on e.g. the DEX explorer need a database with testnet events that they can develop against. Rather than instructing them how to set up a fullnode on their dev workstations, we'll provide an external URL and creds so that they can use the cloud-hosted db directly in their local dev workflows.

To support this, a few changes were necessary to the configs:

  * optionally reference pre-existing Secret for db creds
  * optionally reference pre-existing Secret for TLS certs

Additionally, IPs and DNS were configured out of band, one per environment (e.g. preview & testnet).

### DIFF
--- a/deployments/charts/penumbra-node/templates/statefulset.yaml
+++ b/deployments/charts/penumbra-node/templates/statefulset.yaml
@@ -93,6 +93,17 @@ spec:
               - key: "postgres-cometbft-schema.sql"
                 path: "postgres-cometbft-schema.sql"
       {{ end }}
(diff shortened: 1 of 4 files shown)


conorsch:
Developer confirmed out of band that the new setup is working on their machine, so I'm merging this. Note that we'll need to cherry-pick this commit onto a release branch if we decide to point-release on the `v0.69.x` series, otherwise there'll be a regression in event indexing.

Issue #3995, opened 2024-03-11
tests: 🧨 mock consensus exercises staking component


cratelyn:
in #3588, we are building testing facilities for our `App`. we should write tests using the mock consensus engine that exercise the staking component.

we should write a test (or collection thereof) that performs the following:
- [x] 1. start with a single validator, using the auto_app_state flow
- [x] 2. run the chain for 1000 blocks
- [x] 3. make a tx that defines a new validator, register its consensus signing key with the engine
- [x] 4. after executing (3), check that:
  - [x] 4a. the validator appeared on chain in the defined state
    - _"this is a negative test: we want to make sure that just because someone defined a validator, we don't actually do anything in the staking component's logic_"
  - [x] 4b. the validator set did not change ("defined" validators don't affect staking logic)
- [x] 5. make a tx that delegates more than the minimum amount (where is this defined?) to the new validator
- [x] 6. after executing (5), check that:
  - [x] 6a. no changes to the validator set occur immediately
    - "_this is another negative test, we want to be checking that (un)delegations are queued for the epoch transition rather than being effected immediately_"
  - [x] 6b. **kate:** the amount of delegation tokens looks correct
- [x] 7. run to the end of the epoch (can hardcode or check epoch length), then check
  - [x] 7a. the validator becomes Active (it will probably skip Inactive because only one other validator)
  - [x] 7b. the two validators have different exchange rates
    - "_this is also a negative test, we want to make sure that the validator exchange rate mechanism is actually doing stuff, we expect that the two validators would have different rates because the second validator is created later and the chain has already "warmed up" and run through a few epochs_"
- [x] 8. run 1000 blocks into the future, then check
  - [x] 8a. the new validator's uptime is being tracked correctly
    - "_check that the mock engine is actually feeding correct-enough information to the application. the validator uptime is recorded by the app based on what validators were recorded as signing the block. i would expect there would be failures here when we first write this, and that resolving them would require co-development of the mock engine and the test. **this will be a site of requirements discovery**_"
- [x] 9. make a tx that undelegates some of the delegation tokens
- [x] 10. after executing (9), check that:
  - [x] 10a. the client gets unbonding tokens
  - [x] 10b. the validator's voting power does not change immediately
  - [x] 10c. **kate:** the amount of undelegation tokens looks correct
- [x] 11. run to the end of the epoch, then check
  - [x] 11a. the validator's voting power has decreased
[...]

Issue #3845, opened 2024-02-19
Tracking issue: Staking/Distribution QA and testing


erwanor:
### Staking/Distribution QA and testing

Main tracking issue: #3452
 

In the course of building the staking and distribution components, we frequently uncovered/created bugs in the implementation, either in the form of straight up dysfunctions (e.g. missing state transitions) or in harder to detect invariant violations (e.g. reward overspending).

In both cases, these mistakes could have been caught by leveraging our `App` test harness.

This issue is a first pass at inventorizing bugs that have occured in the process of [re]developing these components. For now, its purpose is to serve as a basis to guide future assurance work. Later, it will serve as a regular tracking issue that describes roadmap and completion of a project.

### Historical bug inventory:
**Validator state-machine**
-  https://github.com/penumbra-zone/penumbra/pull/3770
- https://github.com/penumbra-zone/penumbra/pull/3761

**Staking end-epoch handling**
-  https://github.com/penumbra-zone/penumbra/issues/3814

**Rate calculations**
- https://github.com/penumbra-zone/penumbra/issues/3802
- https://github.com/penumbra-zone/penumbra/pull/3803

**Stake bonding mechanisms**
- https://github.com/penumbra-zone/penumbra/pull/3803
- https://github.com/penumbra-zone/penumbra/issues/3738
- https://github.com/penumbra-zone/penumbra/issues/4056
 
**Validation**
- https://github.com/penumbra-zone/penumbra/pull/3820
- https://github.com/penumbra-zone/penumbra/pull/3831
- https://github.com/penumbra-zone/penumbra/pull/3768
- https://github.com/penumbra-zone/penumbra/issues/3858

### Testing tracker:

**Tombstoning**
- #4149 

**Enshrining action invariants**
- #4183
- https://github.com/penumbra-zone/penumbra/pull/4182

**Rate calculations**
- [ ] staking(tests): add rate calculation scenarios

**Validator state-machine**
- [ ] staking(tests): cover validator state machine transitions

PR #4001, opened 2024-03-12
tests: 🏃 mock consensus can exercise staking component

### LINKED ISSUE #3908: tests: 🐦 mock consensus supports post-genesis validators
the mock consensus engine should support the ability to add a validator after genesis.

this will be a part of adding support for tests of the staking component.

see #3588.

### LINKED ISSUE #3966: mock-consensus: 🪙 `auto_app_state` assigns delegation tokens
in #3902, we [discussed](https://github.com/penumbra-zone/penumbra/pull/3902#discussion_r1513620298) that `auto_app_state` should:

1. Generate a list of penumbra validator definitions corresponding to the consensus keys specified in the builder. There must be at least one.

2. Inject those validator definitions into the provided AppState

3. Create allocations of delegation tokens for each validator, controlled by some address. Otherwise the validator won't have any voting power and will be immediately deleted.

we should address point (3), so that the validator is not immediately deleted.

* #3588

cratelyn:
fixes #3966. fixes #3908. fixes _part of_ #3995.

this branch introduces the first steps towards mock consensus (#3588) testing of the staking component (#3845).

this defines a validator after genesis, and then shows that it does _not_ enter the consensus set. #3966 is addressed in this branch so that the existing genesis validator correctly enters the consensus set, and so that we can successfully progress to the second epoch.

subsequent changes will exercise delegating to this validator in the `mock_consensus_can_define_and_delegate_to_a_validator`.

#### :sparkles: changes

* alters `with_penumbra_auto_app_state` so that it adds an allocation of delegation tokens to the shielded pool component's genesis content.

* extends `generate_penumbra_validator` so that it generates a real spend key, and returns an `Allocation` for the generated validator. _(see #3966)_

* adds a new `mock_consensus_can_define_and_delegate_to_a_validator` test that defines a post-genesis validator. _(see #3908)_

* defines a new `ConsensusIndexRead::get_consensus_set()` method, which collects all of the identity keys returned by `consensus_set_stream`.

* lowers the events in `penumbra_mock_consensus::block::Builder::execute()` to trace-level events.

* `penumbra_mock_consensus::builder::Builder` will now log a warning if values may be errantly rewritten by the builder methods.

* `TestNode::fast_forward` sets its `i` span field to `1..n`, rather than `0..n-1`.

---

#### :link: related

* #4009
* #4010
* #4011
* #4017
* #4027
* #4028
* #4029
* #3966
* #3908
* #3588

### DIFF
--- a/crates/core/app/tests/common/test_node_builder_ext.rs
+++ b/crates/core/app/tests/common/test_node_builder_ext.rs
@@ -54,7 +54,7 @@ impl BuilderExt for Builder {
 }
 
 /// Generates a [`Validator`][PenumbraValidator] given a consensus verification key.
-pub fn generate_penumbra_validator(
+fn generate_penumbra_validator(
     consensus_key: &ed25519_consensus::VerificationKey,
 ) -> (PenumbraValidator, Allocation, penumbra_keys::keys::SpendKey) {
     use decaf377_rdsa::VerificationKey;
--- a/crates/core/app/tests/common/test_node_builder_ext.rs
+++ b/crates/core/app/tests/common/test_node_builder_ext.rs
@@ -5,6 +5,8 @@ use {
         core::keys::v1::{GovernanceKey, IdentityKey},
         penumbra::core::component::stake::v1::Validator as PenumbraValidator,
     },
+    penumbra_shielded_pool::genesis::Allocation,
+    tracing::trace,
 };
 
 /// Penumbra-specific extensions to the mock consensus builder.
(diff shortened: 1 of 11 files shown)


cratelyn:
cherry-picked logging additions out into #4009.

cratelyn:
`shielded-pool` changes are going to conflict with #4020 when that lands.

crates/core/app/tests/mock_consensus.rs
```diff
@@ -98,6 +98,7 @@ async fn mock_consensus_can_send_a_sequence_of_empty_blocks() -> anyhow::Result<
 }
 
 #[tokio::test]
+#[ignore]
```
cratelyn, line 101: do not merge this kate!! (_this was helpful when working in a `cargo watch` loop_)

crates/core/app/tests/mock_consensus_staking.rs
```diff
@@ -132,10 +126,10 @@ async fn mock_consensus_can_delegate_to_a_validator() -> anyhow::Result<()> {
                 chain_id: TestNode::<()>::CHAIN_ID.to_string(),
                 ..Default::default()
             },
-        };
-        plan.populate_detection_data(OsRng, 0);
-        client.witness_auth_build(&plan).await?
+        }
     };
+    plan.populate_detection_data(rand_core::OsRng, 0);
+    let tx = client.witness_auth_build(&plan).await?;
```
cratelyn, line 132: unrelated change, pardon me!

crates/core/app/tests/mock_consensus_staking.rs
```diff
@@ -101,29 +101,23 @@ async fn mock_consensus_can_delegate_to_a_validator() -> anyhow::Result<()> {
...
-            let seed = SpendKeyBytes(OsRng.gen());
-            SpendKey::from(seed).spend_auth_key().sign(OsRng, &bytes)
-        };
+        let bytes = new_validator_pb.encode_to_vec();
+        let auth_sig = new_validator_spend_key.spend_auth_key().sign(OsRng, &bytes);
         let action = ActionPlan::ValidatorDefinition(validator::Definition {
-            validator,
+            validator: new_validator.clone(),
             auth_sig,
         });
```
cratelyn, line 119: :point_up: right here, we need the spend auth key to sign the validator definition

crates/core/app/tests/mock_consensus_staking.rs
```diff
@@ -154,19 +148,20 @@ async fn mock_consensus_can_delegate_to_a_validator() -> anyhow::Result<()> {
...
+        // consensus yet.
+        assert_eq!(
             snapshot
                 .get_validator_state(&new_validator.identity_key)
-                .await?
-                .is_none(),
-            "validator definition with a bad signature should not be defined"
+                .await?,
+            Some(State::Defined),
+            "new validator definition should be defined but not active"
```
cratelyn, line 158: ...and then we see that transaction is accepted!

crates/core/app/tests/common/test_node_builder_ext.rs
```diff
@@ -56,7 +56,7 @@ impl BuilderExt for Builder {
 /// Generates a [`Validator`][PenumbraValidator] given a consensus verification key.
 pub fn generate_penumbra_validator(
     consensus_key: &ed25519_consensus::VerificationKey,
-) -> (PenumbraValidator, Allocation) {
+) -> (PenumbraValidator, Allocation, penumbra_keys::keys::SpendKey) {
```
hdevalence, line 59: We don't want to return the spend key here, since this method shouldn't be used in any situation where we need access to the spend key.

This method should be purely internal to the `with_penumbra_auto_app_state` method -- where we discard the validator's signing keys post-setup -- and not exposed for external use.

crates/core/app/tests/mock_consensus_staking.rs
```diff
@@ -101,29 +101,23 @@ async fn mock_consensus_can_delegate_to_a_validator() -> anyhow::Result<()> {
 
     // Make a transaction that defines a new validator.
     let new_validator_vk = node.keyring_mut().add_key();
-    let (new_validator_pb, _) = self::common::generate_penumbra_validator(&new_validator_vk);
+    let (new_validator_pb, _, new_validator_spend_key) =
+        self::common::generate_penumbra_validator(&new_validator_vk);
```
hdevalence, line 105: This structure has inverted control flow to what we want. We don't want to be using the engine to generate the keys. We want to be defining a validator and passing it to the engine.

hdevalence: In particular, we **don't** want to be using a common helper method to define the new validator we're adding. It should be possible to reason about the semantics of the test function just from reading its code, but if we reuse an internal helper method of the `auto_app_state` implementation, we're coupling the semantics of the test to the implementation of some other part of the test harness. This is actively undesirable.

crates/core/app/tests/common/test_node_builder_ext.rs
```diff
@@ -21,75 +23,59 @@ impl BuilderExt for Builder {
...
-    /// - get its verification key
-    /// - use that for the identity key
-    /// - throw the signing key away
-    ///
-    /// NB: for now, we will use the same key for governance. See the documentation of
-    /// `GovernanceKey` for more information about cold storage of validator keys.
-    const INVALID_KEY_BYTES: [u8; 32] = [0; 32];
+pub fn generate_penumbra_validator(
+    consensus_key: &ed25519_consensus::VerificationKey,
+) -> (PenumbraValidator, Allocation, penumbra_keys::keys::SpendKey) {
```
hdevalence, line 59: this should remain a private API, and not return the spend key

crates/core/app/tests/mock_consensus_staking.rs
```diff
@@ -0,0 +1,173 @@
...
+    penumbra_mock_consensus::TestNode,
+    penumbra_stake::component::validator_handler::ValidatorDataRead as _,
+    tap::Tap,
+    tracing::{error_span, info, Instrument},
+};
+
+mod common;
+
+#[tokio::test]
+async fn mock_consensus_can_delegate_to_a_validator() -> anyhow::Result<()> {
```
hdevalence, line 18: maybe `define_and_delegate_to_a_new_validator`?

crates/core/app/tests/mock_consensus_staking.rs
```diff
@@ -0,0 +1,173 @@
...
+        assert_eq!(end, Some(State::Active));
+    }
+
+    async fn get_consensus_set(
+        snapshot: &cnidarium::Snapshot,
+    ) -> anyhow::Result<Vec<penumbra_stake::IdentityKey>> {
+        use futures::TryStreamExt;
+        use penumbra_stake::component::ConsensusIndexRead;
+        snapshot.consensus_set_stream()?.try_collect().await
+    }
```
hdevalence, line 88: this could usefully be promoted to the `ConsensusIndexRead` trait as a helper method there?

crates/core/app/tests/mock_consensus_staking.rs
hdevalence: example here: ceb5d6283

crates/core/app/tests/mock_consensus_staking.rs
cratelyn: yes! a6584c64d

crates/core/app/tests/mock_consensus.rs
cratelyn: 9f569c4a0

crates/core/app/tests/mock_consensus_staking.rs
cratelyn: ceb5d62

crates/core/app/tests/common/test_node_builder_ext.rs
cratelyn: fixed in ceb5d62

crates/core/app/tests/mock_consensus_staking.rs
cratelyn: > we don't want to be using a common helper method to define the new validator we're adding. It should be possible to reason about the semantics of the test function just from reading its code, but if we reuse an internal helper method of the auto_app_state implementation, we're coupling the semantics of the test to the implementation of some other part of the test harness. This is actively undesirable.

definitely. i agree. it was a mistake to reuse that helper function here. i have cherry-picked ceb5d62 into this branch.

crates/core/app/tests/common/test_node_builder_ext.rs
cratelyn: i cherry-picked ceb5d62 into this branch.

cratelyn (commented):
i've cherry-picked changes from `staking-mock-consensus-tweaks` into this branch. it resolves most of the comments in discussion above, and i've pushed a6584c64d, 733b9a123, and 9f569c4a0 to address the other remaining threads.

crates/core/app/tests/common/test_node_builder_ext.rs
cratelyn: 3ca8c462e

crates/core/app/tests/common/test_node_builder_ext.rs
cratelyn: 3ca8c462e removes the spend key from the return value.

cratelyn:
opened #4039 to remove the `keyring` submodule, using a type alias rather than a newtype, and moving key generation logic out to call-sites.

crates/core/app/tests/mock_consensus_staking.rs
```diff
@@ -0,0 +1,196 @@
...
+        // Keyring should just be a BTreeMap rather than creating a new API
+        .insert(validator_consensus.clone(), validator_consensus_sk);
+     */
+
+    // Now define the validator's configuration data.
+    let new_validator = Validator {
+        identity_key: new_validator_id.clone(),
+        // TODO: upstream a direct conversion from ed25519_consensus
+        consensus_key: tendermint::PublicKey::from_raw_ed25519(&new_validator_consensus.to_bytes())
+            .expect("consensus key is valid"),
```
cratelyn, line 124: filed https://github.com/informalsystems/tendermint-rs/pull/1401 to address this `TODO` :slightly_smiling_face:

PR #4003, opened 2024-03-12
mock-consensus: 🍐 clarify builder kind in error

cratelyn:
there is a test node builder and a block builder. let's clarify which builder we mean in our error message. cherry-picked out of #4001.


cratelyn:
this ends up being unnecessary due to another branch i'll have up shortly, since this pattern will no longer be refutable. closing this!

PR #4004, opened 2024-03-12
mock-consensus: 👷 block data is implicitly empty, can be appended

### LINKED ISSUE #3936: tests: 🧃 mock consensus block builder has empty data by default, `add_tx` to append tx's
https://github.com/penumbra-zone/penumbra/blob/4fca2ac10bca8e601c8d6481d7182d30a724c39b/crates/test/mock-consensus/src/block.rs#L27-L28

this doesn't need to be an `Option<T>`.

https://github.com/penumbra-zone/penumbra/blob/4fca2ac10bca8e601c8d6481d7182d30a724c39b/crates/test/mock-consensus/src/block.rs#L48-L54

this should be accompanied by an `add_tx` method to add a _single_ entry.

cratelyn:
fixes #3936. based upon #4002. see #3588.

this changes the type of the block builder's `data`, allowing it to be implicitly kept empty. this is useful when e.g. fast forwarding a number of blocks _(see #4002)_. this spares us the need to call `with_data(vec![])`, and additionally means that the block builder is never in an uninitialized state _(making #4003 needless)._

* #3936
* #4002
* #3588

### DIFF
--- a/crates/test/mock-consensus/src/block.rs
+++ b/crates/test/mock-consensus/src/block.rs
@@ -49,6 +49,12 @@ impl<'e, C> Builder<'e, C> {
         Self { data, ..self }
     }
 
+    /// Appends the given tx to this block's data.
+    pub fn add_tx(mut self, tx: Vec<u8>) -> Self {
+        self.data.push(tx);
+        self
+    }
+
     /// Sets the evidence [`List`][evidence::List] for this block.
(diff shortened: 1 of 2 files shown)


cratelyn:
force pushed and rebased off of `main`, now that #4002 is merged.

PR #4008, opened 2024-03-12
pd: save work streaming compact blocks by avoiding needless deserialization

hdevalence:
## Describe your changes

This pulls a proto type out of the state and passes it to the RPC directly, rather than deserializing through a domain type (which involves another round of allocations and some crypto operations to parse field elements). This should help reduce load for RPC servers streaming compact blocks to clients.

We could in principle go further and pull bytes directly out of the state and pass them directly to the client (avoiding the bytes>proto>bytes conversion) but that would be more work (I'm not sure exactly how or if Tonic supports that), and this is lower-hanging fruit.

No issue, just an observation after I looked at Grafana and took a few minutes to change

  > This only changes the RPC implementation, not any consensus logic; it should be safe to cherry-pick and deploy on a release branch.


hdevalence:
Looks like my changes in #3993 broke the PR template (cc @aubrika)

PR #4013, opened 2024-03-12
Link nullifier key to claim address in swap claim

### LINKED ISSUE #3978: dex: swap claim should check diversified address integrity

You should only be able to claim swap outputs once. Instead, an attacker can construct a valid SwapClaim multiple times for the same swap. 

Iin the swap claim, we are demonstrating the integrity of the revealed nullifier via `nullifier = hash3(nk, position, swap commitment)`.  Elsewhere in the circuit we have the transmission key `pk_d`, and the diversified base `B_d`  in order to demonstrate swap commitment integrity. However, `pk_d` and `nk` are _not_ checked to be associated with one another, so an attacker can witness multiple values for `nk` in order to generate multiple nullifiers that will each be accepted as valid. 

PoC: https://github.com/penumbra-zone/penumbra/commit/f21d8b512db5a3af2530cf8f833493e5c9500422

```
     Running `target/debug/pcli tx swap 1test_usd --into upenumbra`
Scanning blocks from last sync height 3025 to latest height 3025
building transaction...
finished proving in 13.828 seconds [3 actions, 3 proofs, 2373 bytes]
broadcasting transaction and awaiting confirmation...
transaction broadcast successfully: 3fbe3561656f9b8bd4394c077003b58a360aae9966edd15362e26196eadaee13
transaction confirmed and detected: 3fbe3561656f9b8bd4394c077003b58a360aae9966edd15362e26196eadaee13 @ height 3029
Swap submitted and batch confirmed!
You will receive outputs of 0test_usd and 49.999mpenumbra. Claiming now...
building transaction...
finished proving in 8.715 seconds [1 actions, 1 proofs, 570 bytes]
broadcasting transaction and awaiting confirmation...
transaction broadcast successfully: d8bc3575a646a5825a3e9ebcd17194076d07a9ebcb2c15b8ab280703be7570c2
transaction confirmed and detected: d8bc3575a646a5825a3e9ebcd17194076d07a9ebcb2c15b8ab280703be7570c2
PoC: claiming swap outputs again
building transaction...
finished proving in 7.258 seconds [1 actions, 1 proofs, 570 bytes]
broadcasting transaction and awaiting confirmation...
transaction broadcast successfully: 159b7294bf3dc7629e42659d5ca29277d27f7865d9d54af615f72966c8994c70
transaction confirmed and detected: 159b7294bf3dc7629e42659d5ca29277d27f7865d9d54af615f72966c8994c70
```

1. Witness `ak`
2. Using 1 and the existing variables in circuit for pk_d and nk, check `pk_d = [ivk] B_d` deriving the ivk from ak and nk
3. Check diversified base B_d is not identity

cronokirby:
Closes #3978

Checking that the ak is not the identity may not be necessary? It's what spend does, but I don't think you need it here.


avahowell:
since this contains circuit changes, i assume we need to regenerate proving keys?

crates/core/component/dex/src/swap_claim/proof.rs
```diff
@@ -211,6 +239,18 @@ impl ConstraintSynthesizer<Fq> for SwapClaimCircuit {
...
         nullifier_var.enforce_equal(&claimed_nullifier_var)?;
 
+        // Connection between nullifier key and address
+        let ivk = IncomingViewingKeyVar::derive(&nk_var, &ak_var)?;
+        let computed_transmission_key =
+            ivk.diversified_public(&swap_plaintext_var.claim_address.diversified_generator)?;
+        computed_transmission_key
+            .enforce_equal(&swap_plaintext_var.claim_address.transmission_key)?;
+        // Check the diversified base is not identity.
+        let identity = ElementVar::new_constant(cs.clone(), decaf377::Element::default())?;
```
erwanor, line 249: I think that it's missing from the spec

hdevalence (changes requested):
Can we update the spec in tandem with these changes?

redshiftzero (approved):

docs/protocol/src/dex/action/swap_claim.md
```diff
@@ -64,6 +65,20 @@ using the witnessed values above and where `ds` is a constant domain separator:
 
 as described in [Nullifiers](../notes/nullifiers.md).
 
+### Nullifier Key Linking
+
+The zk-SNARK certifies that the diversified address $pk_d$ associated with the note being spent was derived as:
```
hdevalence, line 70: This doesn't seem right -- a `SwapClaim` doesn't spend any notes, so there is no address associated with the note being spent.

What we need to describe here is:

- Which `nk` is permitted to derive the nullifier for the state commitment consumed by the `SwapClaim`?
- How does the proof statement certify that the revealed nullifier was correctly derived using the permitted `nk`?

cronokirby: Yeah, this was a mistake, there's an address associated with the swap being claimed.

docs/protocol/src/sct.md
```diff
@@ -1 +1,52 @@
...
+Establishing that a state fragment was previously validated by the chain is not
+enough, however.  Clients also need to establish that the state fragments their
+transaction consumes have not _already_ been consumed by another transaction, to
+prevent double-spend attacks.
+
+This is accomplished by the use of _nullifiers_, described in detail in the
+[Nullifiers](./sct/nullifiers.md) section.  Each state commitment has an
+associated nullifier.  Transactions reveal the nullifiers of state fragments
+they consume, and the chain checks that they have not been previously revealed.
+Nullifiers can only be derived by the actor with authority over the state
```
cronokirby, line 49:
```suggestion
Nullifiers can only be derived by the actor with viewing authority over the state
```

I think this is more accurate, because you don't need spend authority, just viewing capability (nullifier key is in the FVK)

hdevalence: I was trying to be a little obscure about _which_ authority exactly, since it might differ by state fragment, and leave the details for the nullifiers section. For instance, one thing I want us to resolve through discussion is which nk should have nullification authority over a swap commitment, which we never explicitly described. 

I would prefer to not say that there is a notion of viewing authority over a swap commitment, for instance, because it means we have to define exactly what that term means, so far we've only defined "spend authority" and that's not relevant to a swap commitment.

cronokirby: The relevant authority for a swap commitment would whoever has viewing authority over the claim address in the swap then, I guess.

PR #4016, opened 2024-03-13
dex: hotfix to avoid discarding EventPositionExecution

### LINKED ISSUE #4007: `PositionExecution` events do not appear in indexer
**Describe the bug**

The DEX is supposed to emit `PositionExecution` events as it executes. However, these events don't appear in the indexer.

**To Reproduce**

On `deimos-5`, I created a liquidity position `plpid1cr5cxr7dt8qwwjx8lmgfjnxea3fmxn7kzlw46pnk4pt5c384nw6qajqzr0` that was intentionally mispriced. Its initial reserves were 232.897367penumbra, 0test_usd. Creation transaction is here: https://cuiloa.testnet.penumbra.zone/transaction/360cf15d5fedf532f1c2dfc2d3245354780ef3ba8218eb8b46a9c10ecf0dfede

In block 64883, this was submitted against and then arb'd to have reserves 700.935094560192616373test_usd

The ProtoJSON for that LPID is `{"inner":"wOmDD81ZwOdIx/7QmUzZ7FOzT9YX3V0GdqhXTET1m7Q="}`.

After closing and withdrawing, I saw
```
penumbra=# SELECT *
FROM events
WHERE events.rowid IN (
  SELECT event_id
  FROM attributes
  WHERE value = '{"inner":"wOmDD81ZwOdIx/7QmUzZ7FOzT9YX3V0GdqhXTET1m7Q="}'
);
 rowid | block_id | tx_id |                         type
-------+----------+-------+------------------------------------------------------
 17255 |     3576 |   400 | penumbra.core.component.dex.v1.EventPositionOpen
 24719 |     5322 |   528 | penumbra.core.component.dex.v1.EventPositionWithdraw
 24695 |     5316 |   527 | penumbra.core.component.dex.v1.EventPositionClose
(3 rows)
```
in the events DB. showing that there were other events but not `EventPositionExecution`. None of those seem to appear in the DB:
```
penumbra=# select * from events where type = 'penumbra.core.component.dex.v1.EventPositionExecution';
 rowid | block_id | tx_id | type
-------+----------+-------+------
(0 rows)
```
Looking for block-scoped events, we see
```
penumbra=# SELECT e.*
FROM events e
JOIN blocks b ON e.block_id = b.rowid
WHERE b.height = 64883 AND e.tx_id IS NULL;
 rowid | block_id | tx_id |                       type
-------+----------+-------+--------------------------------------------------
 17220 |     3576 |       | block
 17221 |     3576 |       | penumbra.core.component.dex.v1.EventArbExecution
 17222 |     3576 |       | penumbra.core.component.sct.v1.EventAnchor
 17223 |     3576 |       | penumbra.core.component.sct.v1.EventBlockRoot
(4 rows)
```
so while we see the `EventArbExecution` we don't see any `EventPositionExecution`.

The events are supposed to be emitted here: https://github.com/penumbra-zone/penumbra/blob/main/crates/core/component/dex/src/component/router/fill_route.rs#L410

**Expected behavior**

The emitted events should appear in the event database. Where are they being dropped?

**Additional context**

Blocks progress on the dex explorer

hdevalence:
Closes #4007

The bug results from a hazardous API I created. The `apply` method does _not_ apply the events to the underlying state, it extracts them and returns them to the caller, who can silently discard them.

A better fix is outlined in the comment:

```
// We have to manually extract events and push them down to the state to avoid losing them.
// TODO: in a commit not intended to be cherry-picked, we should fix this hazardous API:
// - rename `StateDelta::apply` to `StateDelta::apply_extracting_events`
// - add `StateDelta::apply_with_events` that pushes the events down.
// - go through all uses of `apply_extracting_events` and determine what behavior is correct
```

However, I didn't do this in this commit, because my hope is that we can have something easily cherry-pickable into a point release, which can be deployed to unblock the DEX work. After applying it to both `main` and a release branch, we can then do the fix for the bug class.

### DIFF
--- a/crates/core/component/dex/src/component/router/fill_route.rs
+++ b/crates/core/component/dex/src/component/router/fill_route.rs
@@ -255,7 +255,16 @@ async fn fill_route_inner<S: StateWrite + Sized>(
     tracing::debug!(?swap_execution, "returning swap execution of filled route");
 
     // Apply the state transaction now that we've reached the end without errors.
-    this.apply();
+    //
+    // We have to manually extract events and push them down to the state to avoid losing them.
+    // TODO: in a commit not intended to be cherry-picked, we should fix this hazardous API:
+    // - rename `StateDelta::apply` to `StateDelta::apply_extracting_events`
+    // - add `StateDelta::apply_with_events` that pushes the events down.
+    // - go through all uses of `apply_extracting_events` and determine what behavior is correct
+    let (mut state, events) = this.apply();
+    for event in events {
+        state.record(event);
+    }
 
     let fill_elapsed = fill_start.elapsed();
     metrics::histogram!(metrics::DEX_ROUTE_FILL_DURATION).record(fill_elapsed);


conorsch (approved):
Looks great. I'll test this on preview prior to point-release, and also file an issue to track more comprehensive changes to the API.

PR #4031, opened 2024-03-15
fix(pd): rename unbonding epochs -> delay

conorsch:
Follow-up to changes made in [0], where we settled on "unbonded delay" rather than "unbonding epochs" to describe how many blocks must elapse until undelegated stake becomes claimable. The `pd` CLI options are only used by developers, so I'm not worried about breaking operator scripts in the wild.

[0] https://github.com/penumbra-zone/penumbra/pull/3923


erwanor (approved):
Ah yes, good catch

PR #4035, opened 2024-03-16
Connect swaps/swap claims

jessepinho:
Despite the `SwapView` and `SwapClaimView` types supporting fields that allow swaps and swap claims to reference each other, we haven't actually been populating those fields yet.

This PR adds a couple keys to the `TransactionPerspective` Protobuf type to make it possible to populate the `claim_tx` and `swap_tx` fields in `SwapView` and `SwapClaimView`, respectively. Hopefully, in the future, we can do the same with undelegates <-> undelegate claims.


proto/penumbra/penumbra/core/transaction/v1/transaction.proto
```diff
@@ -114,6 +114,12 @@ message TransactionPerspective {
   repeated asset.v1.Metadata denoms = 5;
   // The transaction ID associated with this TransactionPerspective
   txhash.v1.TransactionId transaction_id = 6;
+  // A map of base64-encoded nullfiers to the IDs of the transactions containing
+  // the outputs that they nullify.
+  map<string, txhash.v1.TransactionId> transaction_ids_by_nullifier = 7;
+  // A map of base64-encoded commitments to the IDs of the transactions
+  // containing those commitments.
+  map<string, txhash.v1.TransactionId> transaction_ids_by_commitment = 8;
```
jessepinho, line 122: Reviewers, please let me know what you think of this approach. It's the only way I could think of to easily allow views with links to other transactions to be generated from perspectives.

In the future, we'll also need to figure out a way to stuff output 1 and output 2 from the swap claim into the `TransactionPerspective` of a swap. But this work has already been taking me quite a while, so I wanted to at least get this merged first.

crates/core/transaction/src/view/transaction_perspective.rs
```diff
@@ -40,6 +40,12 @@ pub struct TransactionPerspective {
     pub denoms: asset::Cache,
     /// The transaction ID associated with this TransactionPerspective
     pub transaction_id: TransactionId,
+    // A map of base64-encoded nullfiers to the IDs of the transactions
+    // containing the outputs that they nullify.
+    pub transaction_ids_by_nullifier: HashMap<String, TransactionId>,
```
redshiftzero, line 45: should these fields also be getting populated in `ViewServer::transaction_info_by_hash` in the `penumbra-view` crate?

jessepinho: Probably, but I figured I'd leave that up to core devs. These fields aren't a hard requirement — they're just _available_ (after this PR is merged) to use for generating richer views. We'll be populated these fields in the web app, and now the core devs can too if they choose to.

hdevalence: Yeah, I think it's fine for the Rust view server to lag a bit, until there's a concrete use case (like a web frontend pulling from it).

crates/core/component/dex/src/swap/view.rs
```diff
@@ -55,18 +65,21 @@ impl From<SwapView> for pb::SwapView {
...
-                    // Swap claim crossreferencing is not yet supported in the Rust stack.
-                    claim_tx: None,
+                    output_1,
+                    output_2,
+                    claim_tx,
+
                     // These fields are also not yet supported in the Rust stack.
                     asset_1_metadata: None,
                     asset_2_metadata: None,
                     batch_swap_output_data: None,
```
hdevalence, line 82: Don't we want this metadata too? We'd need it to render the assets (symbols, icons, etc) for a `Swap` without a corresponding `SwapClaim` right?

proto/penumbra/penumbra/core/transaction/v1/transaction.proto
hdevalence: Two comments:

1. I think the comment should read "transactions creating the commitments that they nullify" -- this is a minor nit but it's more technically correct since a swap commitment is a different kind of commitment than an output note

2. Why do we do base64 encoding of the nullifiers or commitments? This would force all of the consumers to add in special base64 parsing to interpret the perspective, right? Could we instead encode a list of `(Nullifier, TransactionId)` or `(Commitment, TransactionId)` pairs?

crates/core/transaction/src/is_action.rs
```diff
@@ -349,13 +361,20 @@ impl IsAction for SwapClaim {
         // Get the advice notes for each output from the swap claim
         let output_1 = txp.advice_notes.get(&self.body.output_1_commitment);
         let output_2 = txp.advice_notes.get(&self.body.output_2_commitment);
+        let nullifier = self.body.nullifier;
+        let swap_tx: Option<TransactionId> =
+            match txp.transaction_ids_by_nullifier.get(&nullifier.to_string()) {
```
hdevalence, line 366: Does this work? I'd be surprised if `nullifier.to_string()` returned a Base64 string. There's in general no guarantee that the `ToString` impl on the Rust domain type has any relation to the protojson encoding. For instance `Address`'s `ToString` does bech32.

proto/penumbra/penumbra/core/transaction/v1/transaction.proto
hdevalence: I think there are restrictions on what types Proto3 allows to be used in maps -- like they have to be scalar types (?) -- so this would mean defining two sub-messages, each modeling one of the pairs, and then having a `repeated PairType` (which is basically the way that a `map` will be encoded on the wire anyways). Then we'd change the domain type to have a `BTreeMap<A, B>` and deserialize the proto-gen-type's `Vec<PairType>` into `BTreeMap<A, B>` using `.collect`.

crates/core/transaction/src/is_action.rs
```diff
@@ -329,10 +329,44 @@ impl IsAction for Swap {
...
+                    // per trading pair per block and we trust the TxP provider not to lie about it.
+                    .find(|bsod| bsod.trading_pair == swap_plaintext.trading_pair);
+
+                let (output_1, output_2) = match bsod.map(|bsod| swap_plaintext.output_notes(bsod))
+                {
+                    Some((output_1, output_2)) => {
+                        (Some(txp.view_note(output_1)), Some(txp.view_note(output_2)))
+                    }
+                    None => (None, None),
+                };
```
hdevalence, line 348: The output notes of the swap are determined _entirely_ by (1) the `SwapPlaintext` (committed to by the swap commitment and (2) the `BatchSwapOutputData`.  This means that to get the output notes, we should supply a BSOD in the perspective, then recompute them.

Compared to trying to insert the output notes manually, this has two advantages:

1. It correctly handles the case where the swap tx has been accepted on chain but the claim has not yet been submitted. In that case, even though we don't know the swap claim tx (it doesn't exist), we can still show the output notes.
2. It allows rendering extra information (if we want), like the % share of the batch the user's swap was.

crates/core/transaction/src/is_action.rs
```diff
@@ -329,10 +329,44 @@ impl IsAction for Swap {
...
+                        .cloned(),
+                    batch_swap_output_data: bsod.cloned(),
+                    asset_1_metadata: txp
+                        .denoms
+                        .get(&swap_plaintext.trading_pair.asset_1())
+                        .cloned(),
+                    asset_2_metadata: txp
+                        .denoms
+                        .get(&swap_plaintext.trading_pair.asset_2())
+                        .cloned(),
```
hdevalence, line 367: Populating the `asset_1_metadata` and `asset_2_metadata` is important to be able to view the `Swap` action pre-submission, where there won't be any output notes.

proto/penumbra/penumbra/core/transaction/v1/transaction.proto
```diff
@@ -124,6 +124,31 @@ message TransactionPerspective {
...
+  message NullificationTransactionIdByCommitment {
+    crypto.tct.v1.StateCommitment commitment = 1;
+    txhash.v1.TransactionId transaction_id = 2;
+  }
+  repeated NullificationTransactionIdByCommitment nullification_transaction_ids_by_commitment = 50;
+
+  // Any relevant BatchSwapOutputData to the transaction.
+  //
+  // This can be used to fill in information about swap outputs.
+  repeated component.dex.v1.BatchSwapOutputData batch_swap_output_data = 60;
```
hdevalence, line 151: This needs to be `repeated` since there could be multiple different `Swap`s in the same tx.

hdevalence (approved):

hdevalence:
Rebased on main to resolve conflicts in the generated proto descriptors.

crates/core/transaction/src/view/transaction_perspective.rs
hdevalence: Filed #4083

crates/core/component/dex/src/swap/view.rs
hdevalence: xref https://github.com/penumbra-zone/penumbra/pull/4035#discussion_r1535134495

grod220:
Nice job! This was trickier than I expected.

PR #4037, opened 2024-03-16
Support threshold custody for validators

plaidfinch:
This PR finishes #3813, adding support for signing validator definitions and votes to the threshold custody backend, as well as widening the general interface for custody backends to permit any backend to support these operations.

**Status:** @cronokirby and I walked through this and determined that it does work after all, the UX is just a little unfortunate when multiple calls to threshold custody backend are required to broadcast a single transaction (as in the case with validator custody). But it works!


cratelyn:
> Ideally, we would have an integration test for each, exercising all of:
> 
>     * send a normal transaction 
>     * vote as a validator
>     * update a validator definition
> 
> @cratelyn how feasible do you think it would be to set up such tests?

@plaidfinch that sounds _very_ feasible! check out https://github.com/penumbra-zone/penumbra/blob/main/crates/core/app/tests/mock_consensus_staking.rs#L29-L52

that test creates a validator, and (un)delegates to it.

#3913 and #3973 track the work to integrate the view server into these tests, which should further streamline the ergonomics of creating transactions in tests. you can check out the [`penumbra-mock-consensus`](https://github.com/penumbra-zone/penumbra/tree/main/crates/test/mock-consensus) library if you'd like to take a look around. it's still a work-in-progress (#3588) but this sounds absolutely like a use-case we should cover.

i'm happy to talk sometime next week, or if you'd like, feel free to file a ticket and i can take a look as part of my ongoing work in #3588.

cronokirby (approved):
Paired on this with @plaidfinch.

hdevalence:
De-activated, merged, then re-activated the rule. Thanks!

PR #4053, opened 2024-03-19
cnidarium: use incremental jmt migration

### LINKED ISSUE #3506: upgrades: revisit our migration strategy

To perform a chain upgrade, a node operator needs to halt their node, export and migrate the node's chain state. Frequently, but not always, the migration will alter the chain's consensus state. In that case, we must have a process that is designed to be as least intrusive as possible, specifically we want:
- block continuity across the halt boundary
- the JMT state version and block height remain equal
- the post-upgrade application hash is known prior to performing an initialization handshake with comet

Our current approach, is to overwrite the JMT at the pre-upgrade version, as follow:
- node operator halts `pd` at height `h-1`
- node operator export their chain state
- node operator runs a migration, overwriting the JMT at version `h-1`
- node operator starts their node again, the chain resumes from height `h`

A basic implementation of this migration flow is captured by the [`SimpleUpgrade`](https://github.com/penumbra-zone/penumbra/blob/dbe8abf368b99f620ccf7a338ec92ce9032cbceb/crates/bin/pd/src/upgrade.rs#L32) migration script. The flow relies on the ability to append data to the JMT "in-place" i.e. without increasing its state version. This process, implemented by `Storage::commit_in_place` is currently broken, and using it will result in dangling JMT nodes that are inaccessible if queried via `get_with_proof`. So, while the data for that state version is available in our backing store, we cannot generate proofs for it. This is problematic. 

A correct implementation of `commit_in_place`, we will have to build on top of the `jmt::restore` API, consuming an existing tree state, and using a variation of the `add_chunk` method to append to the existing logical structure of the tree, rather than completely overwriting it. The `jmt::restore` API is also needed to implement JMT pruning (#1806) and state pruning is probably a key part of every realistic chain upgrade scenario. The main difference is that the current API requires callers to specify a post-restoration root hash, which we would not be able to provide before the fact.

Alternatively, we could rework the chain upgrade process to include an "offline" block `h`:
- node operator halts `pd` at height `h-1`
- node operator export their chain state
- node operator runs a migration, creating a block `h` and incrementing the JMT version number
- node operator starts their node again, the chain resumes from height `h+1`

This would require the offline step to setup an `App` that would be fed synthetic `Begin/EndBlock`s. This approach has the benefit of not requiring any JMT change, but creates more testing surface, and higher risk of wrongful validator slashing. For example it would require tracking a "fake" uptime for each validator, using a made up timestamp inserted into to the `BeginBlock` contents.

erwanor:
Close #3506, this PR:
- use `jmt@0.10` which includes the `migration` feature
- bump `borsh` to `1.3.0` to be able to serialize jmt nodes
- revamp the substore migration tests to include basic tests and a proptest strategy
- notably, the proptest integration tests lay the groundwork for deeper/more targeted strategies

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1409,6 +1409,7 @@ dependencies = [
  "parking_lot",
  "pbjson",
  "pin-project",
+ "proptest",
  "prost",
  "regex",
  "rocksdb",
@@ -1417,6 +1418,7 @@ dependencies = [
  "smallvec",
  "tempfile",
  "tendermint",
+ "test-strategy",
  "tokio",
  "tokio-stream",
  "tonic",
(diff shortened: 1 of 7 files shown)


erwanor:
`rustc` segfault? 😈

conorsch (approved):
LGTM. Can we squash on merge?

erwanor:
Yeah for sure, waiting for CI to pass

PR #4055, opened 2024-03-20
feat(pd): support archives for migrate and join

conorsch:
Enables opt-in archive generation when performing:

* pd export
* pd migrate

The goal is to provide a standardized bottling-up of pd state, specifically the rocksdb directory. In the context of upgrades, only the "pd migrate" functionality change is what we care about: we want the archived dir to contain both rocksdb data and the modified genesis file.

Accordingly, `pd testnet join` is modified to support an optional archive URL. If set, the remote tar.gz archive will be downloaded and extracted, clobbering the cometbft config. A remote bootstrap node is still contacted, to learn about peers, otherwise the newly created node wouldn't be able to talk to the network.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2351,6 +2351,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "filetime"
+version = "0.2.23"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1ee447700ac8aa0b2f2bd7bc4462ad686ba06baa6727ac149a2d6277f0d240fd"
+dependencies = [
+ "cfg-if",
+ "libc",
+ "redox_syscall",
+ "windows-sys 0.52.0",
+]
+
+[[package]]
 name = "fixed-hash"
 version = "0.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


erwanor:
I think what we want here is to give the ability for `pd join` to specify an archive url or path, and have the `pd` cli logic "install" that archive in the correct data directories. We could have `pd export` generate an archive, but I don't think we need that option for `pd migrate`?

conorsch:
> I think what we want here is to give the ability for pd join to specify an archive url or path, and have the pd cli logic "install" that archive in the correct data directories. 

This is implemented! (I recently force-pushed, so the code just changed.) What's _not_ yet implemented in clobbering of the genesis file. In my local testing, `pd migrate` throws an error, so I cannot generate a modified genesis file and test the round-trip of export -> migrate -> host -> join. 

> We could have pd export generate an archive

That's implemented and working. I don't think it really helps in the upgrade case, but it's easy enough to add the flag, since it reuses a function. What it does leave the door open to is creating archives of node state for fast-sync, so a node can bootstrap itself from height `n` if an archive URL is provided.

> I don't think we need that option for pd migrate?

I think we do! It's `pd migrate` that emits a directory containing 1) modified rocksdb state and 2) modified genesis file. Those two artifacts should be packed up in an archive, and that archive is what should be hosted for new-joining nodes to consume via `pd testnet join --archive-url <URL>`. Crucially, the new-joining node will still contact a remote node and fetch its peer info, which I think is a sane decision.

erwanor (approved):

conorsch:
Merging optimistically to continue with upgrade testing, ahead of #4087.

PR #4056, opened 2024-03-20
staking: handle param change and unbonding edge case

erwanor:
This fix a bug in the claim logic that would cause the node to crash if a parameter change had occurred, or if an undelegation was marked to have taken place after the unbonding height specified in `Unbonding { unbonds_at_height }`. The latter can happen if a validator gets ejected from the consensus set for inactivity, and is equivalent to an `Unbonded` state.


erwanor:
Tested this manually, merging optimistically so that staking test work can be done against an up-to-date staking component implementation.

conorsch:
Thank you for labeling this as consensus-breaking :pray:. I noticed an apphash mismatch when running slightly outdated pd code against preview, and sure enough, it was already labeled.

PR #4065, opened 2024-03-21
ibc: add withdrawal timestamp rounding chain rule

### LINKED ISSUE #4064: ibc: enforce rounding to the nearest even second for timeouts on withdrawals
IBC withdrawal timeouts have two components: a height, and a timestamp. The timestamp is in nanoseconds. Clients could be identified by their clock skew at that timestamp resolution, a privacy leak. So, we should do two things:

- [ ] Add a chain rule that enforces that timestamps are within the nearest even second (for withdrawals only)
- [ ] Amend the `pcli withdraw` client command to apply this rounding, and ensure all other clients (like minifront) do the same.

avahowell:
**this is a consensus-breaking change**

Fixes #4064

### DIFF
--- a/crates/bin/pcli/src/command/tx.rs
+++ b/crates/bin/pcli/src/command/tx.rs
@@ -1051,23 +1051,20 @@ impl TxCmd {
                     }
                 };
 
-                // get the current time on the local machine (rounded to the nearest even second)
-                let mut current_time_s = SystemTime::now()
+                // get the current time on the local machine
+                let current_time_ns = SystemTime::now()
                     .duration_since(UNIX_EPOCH)
                     .expect("Time went backwards")
-                    .as_secs() as u64;
-
-                if current_time_s % 2 != 0 {
-                    current_time_s += 1
-                }
-
-                let current_time_ns = current_time_s * 1_000_000_000;
+                    .as_nanos() as u64;
 
                 let mut timeout_timestamp = *timeout_timestamp;
                 if timeout_timestamp == 0u64 {
                     // add 2 days to current time
                     timeout_timestamp = current_time_ns + 1.728e14 as u64;
                 }
+
+                // round to the nearest even second
+                timeout_timestamp += timeout_timestamp % 2_000_000_000;
 
                 fn parse_denom_and_amount(value_str: &str) -> anyhow::Result<(Amount, Metadata)> {
                     let denom_re = Regex::new(r"^([0-9.]+)(.+)$").context("denom regex invalid")?;
--- a/crates/bin/pcli/src/command/tx.rs
+++ b/crates/bin/pcli/src/command/tx.rs
@@ -1063,8 +1063,8 @@ impl TxCmd {
                     timeout_timestamp = current_time_ns + 1.728e14 as u64;
                 }
 
-                // round to the nearest even second
-                timeout_timestamp += timeout_timestamp % 2_000_000_000;
+                // round to the nearest second
+                timeout_timestamp += timeout_timestamp % 1_000_000_000;
 
                 fn parse_denom_and_amount(value_str: &str) -> anyhow::Result<(Amount, Metadata)> {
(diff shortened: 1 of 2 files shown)


crates/core/component/shielded-pool/src/ics20_withdrawal.rs
```diff
@@ -77,6 +77,15 @@ impl Ics20Withdrawal {
             anyhow::bail!("timeout time must be non-zero");
         }
 
+        // in order to prevent clients from inadvertantly identifying themselves by their clock
+        // skew, enforce that timeout time is rounded to the nearest even second
+        if self.timeout_time % 2_000_000_000 != 0 {
```
hdevalence, line 82: What's the motivation for rounding to 2-second intervals rather than 1-second ones?

I might have miscommunicated in Discord, I was thinking of "even second" as meaning "on the second boundary" -- but looking at this I'm wondering what the right resolution should be. 1s is an "easy" answer, but what about 10 or 60, or greater?

avahowell: I don't have a good intuition for 'what amount of rounding covers the skew of >99% of machines', but I think single seconds might not be enough. From a user's standpoint quantizing withdrawal timeouts to minutes seems fine, so maybe 1 minute would be better

hdevalence: Let's set two rules:

- in consensus, require that timeout timestamps are truncated to one-minute boundaries
- in our clients, set the default timeout timestamp to be truncated to 1h boundaries

This leaves the possibility for someone who has a pressing need for finer grained timestamps to do so while setting a good default

cronokirby: For posterity, what's the reasoning behind baking a privacy setting like this into consensus? Is it to prevent mistakes with alternative clients?

hdevalence: Yeah, the "obvious" thing to do is set the timestamp as Utc.now() + timeout, and the existence of this rule means that that code will immediately break.

avahowell:
Updated the chain rule to require rounding to 1 minute, updated pcli to round to 1 hour

hdevalence:
@valardragon comments:

> Dev, [Mar 21, 2024 at 10:55:18 PM]:
> That make sense to me, but I personally wouldn't set a 1hr timeout as the default
> 
> I'd recommend 5m-15m, 1 hour is quite long

avahowell:
Changed the pcli default to round to nearest 10 minute, per suggestion

crates/core/component/shielded-pool/src/ics20_withdrawal.rs
```diff
@@ -77,6 +77,15 @@ impl Ics20Withdrawal {
             anyhow::bail!("timeout time must be non-zero");
         }
 
+        // in order to prevent clients from inadvertantly identifying themselves by their clock
+        // skew, enforce that timeout time is rounded to the nearest minute
+        if self.timeout_time % 60_000_000_000 != 0 {
+            anyhow::bail!(
+                "withdrawal timeout timestamp {} is not rounded to one second",
```
erwanor, line 84: s/second/minute

crates/bin/pcli/src/command/tx.rs
```diff
@@ -1052,17 +1052,20 @@ impl TxCmd {
...
 
                 let mut timeout_timestamp = *timeout_timestamp;
                 if timeout_timestamp == 0u64 {
                     // add 2 days to current time
-                    timeout_timestamp = current_time_u64_ms + 1.728e14 as u64;
+                    timeout_timestamp = current_time_ns + 1.728e14 as u64;
                 }
 
+                // round to the nearest 10 minutes
+                timeout_timestamp += timeout_timestamp % 600_000_000_000;
```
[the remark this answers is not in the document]
erwanor: ah yes, saw the comment

PR #4066, opened 2024-03-21
dex: add DexParameters, remove RoutingParams::default()

hdevalence:
This commit adds a `DexParameters` section to the app parameters and removes the previous `RoutingParams::default()`.

The DEX's path search considers two kinds of candidates during routing: dynamically-chosen candidate assets indexed by liquidity, and fixed candidates that provide the "spine" of the liquidity graph.  Previously, these fixed candidates were literally hardcoded into the test code, which wouldn't work on mainnet. Now, they're configurable parameters.


aubrika (approved):
LGTM

conorsch:
Added consensus-breaking label, for bookkeeping purposes. Please remove if I'm wrong about that.

Issue #4067, opened 2024-03-21
staking: add a delay before re-enabling a definition


erwanor:
Zellic found a particularly cheap way to do epoch-grinding. With an `Active` validator, it consists in stuffing a transaction with two actions:
- Disable identity -> This triggers an early epoch-end
- Re-enable identity -> This will set the validator to `Inactive`

When the block executes, the epoch handler will run and set the validator back to `Active`.

Zellic reported they were able to:
1. Slow down block production
2. Make the TCT epoch index overflow 

The core issue was known (see #3738) but this is a particularly cheap and effective way to abuse it.

**Proposed solution**
Add a block delay to slow down the grinding and increase coordination costs on the attackers' part.

1. When a validator is disabled write the current height to `"staking/validators/data/last_disabled/{id}"` to NV storage (this is okay to do because the effect of this consensus rule is eventually captured in the jmt, in the form of the `validator::State`)
2. When a validator definition is uploaded, add a check that ensures that at least N blocks have elapsed

Justification:
1. We have refactored the end-epoch handler since Zellic reviewed the codebase. In #3874, we determined that the new end-epoch handler didn't show in application profiles.
2. This makes it substantially harder to reach a TCT overflow

PR #4068, opened 2024-03-21
staking: add delay to reenable a validator

erwanor:
Sketch solution to #4067, this gives a starting point to provide a justification why this solve epoch grinding and TCT index overflows.


cratelyn (commented):
this seems like a reasonable fix. should we also write a mock consensus test to ensure that this doesn't regress in the future, and that it works properly? i wouldn't block on that ask, a separate issue to prioritize afterwards would be fine with me.

erwanor:
> this seems like a reasonable fix. should we also write a mock consensus test to ensure that this doesn't regress in the future, and that it works properly? i wouldn't block on that ask, a separate issue to prioritize afterwards would be fine with me.

That's a great idea, I've been looking for an opportunity to write one so sgtm

PR #4070, opened 2024-03-21
tests(app): 🧪 test that uptime tracks signed blocks

### LINKED ISSUE #4040: tests: 🖋️ mock consensus engine can test validator uptime measurements
stub issue, see #3588. see #3995.

currently, the consensus keys in the keyring are not used by either the test node or block builder to sign blocks. we should introduce the ability to sign blocks, in order to test e.g. the staking component.

see section 8a in #3995. we will likely need to do some requirements discovery as part of that work, which should drive the fix to this issue.

* #4061
* #4070
* #3995
* #3588
* #4099

cratelyn:
**NB:** this is based on #4099.

this adds test coverage, complementary to the work in #4061, which asserts that we properly track the _affirmative_ case of validators signing blocks.

fixes #4040. see #3995.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5282,6 +5282,7 @@ dependencies = [
  "bytes",
  "ed25519-consensus",
  "rand_core 0.6.4",
+ "sha2 0.10.8",
  "tap",
  "tendermint",
  "tower",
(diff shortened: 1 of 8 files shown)


cratelyn:
rebased now that #4099 has landed.

PR #4086, opened 2024-03-22
dex: redesign value circuit breaker

### LINKED ISSUE #4025: dex: move value circuit breaker accounting to dex engine boundaries
Currently the DEX value circuit breaker is implemented within `put_position`, however this has two drawbacks:

1. Misuse resistance -- what if some code modified positions without calling `put_position`? I audited the codebase for any patterns like this during the initial implementation and couldn't find any, however it could appear in the future. However that seems like a "dark pattern" in general as `put_position` has various other state machine checks and handles updating indices and liquidity
2. More importantly, an implementation at the boundaries will be significantly more efficient, as not every position update requires updating and checking the value circuit breaker

Instead, there are only two places that value flows into the dex and two places value flows out:

**IN:**
* swap inputs
* opened liquidity positions

**OUT:**
* swap outputs
* withdrawn positions

If we move the value circuit breaker management to the entry points for those operations, we are guaranteed to check that dex inflows > outflows and get the same assurance with fewer CPU cycles.

hdevalence:
Closes #4025

This commit sketches a new mechanism but does not fix the existing tests; it needs to be picked up and pushed over the finish line. (We should be testing this).


crates/core/component/dex/src/component/dex.rs
```diff
@@ -209,12 +209,26 @@ pub trait StateWriteExt: StateWrite + StateReadExt {
...
+    async fn set_output_data(
         &mut self,
         output_data: BatchSwapOutputData,
         swap_execution_1_for_2: Option<SwapExecution>,
         swap_execution_2_for_1: Option<SwapExecution>,
-    ) {
+    ) -> Result<()> {
+        // Debit the DEX for the swap outflows.
+        // Note that since we credited the DEX for _all_ inflows, we need to debit the
+        // unfilled amounts as well as the filled amounts.
```
hdevalence, line 220: this would be good to cover in a test

crates/core/component/dex/src/component/circuit_breaker/value.rs
```diff
@@ -1,42 +1,46 @@
...
+            .checked_sub(&value.amount)
+            .ok_or_else(|| anyhow!("underflowed balance while debiting value circuit breaker"))?;
+        self.put(state_key::value_balance(&value.asset_id), new_balance);
+        Ok(())
     }
 }
 
+impl<T: StateWrite + ?Sized> ValueCircuitBreaker for T {}
+
+/*
```
zbuc, line 42: What happens if the tests are uncommented?

crates/core/component/dex/src/component/position_manager.rs
```diff
@@ -570,32 +569,6 @@ pub(crate) trait Inner: StateWrite {
             "updating position assets' aggregate balances"
```
zbuc, line 569: the update of the aggregate balance was for debugging the value circuit breaker; this tracing message and the `net_change_*` variables are likely no longer needed

crates/core/component/dex/src/component/circuit_breaker/value.rs
```diff
@@ -1,42 +1,46 @@
...
-    pub fn available(&self, asset_id: Id) -> Value {
-        self.balance
-            .provided()
-            .find(|b| b.asset_id == asset_id)
-            .unwrap_or(Value {
-                asset_id,
-                amount: Amount::from(0u64),
-            })
+    /// Debits a deposit from the DEX.
+    async fn vcb_debit(&mut self, value: Value) -> Result<()> {
```
zbuc, line 27: should we have some debug tracing messages here for the value circuit breaker credits/debits?

crates/core/component/dex/src/component/dex.rs
```diff
@@ -209,12 +209,26 @@ pub trait StateWriteExt: StateWrite + StateReadExt {
...
+        self.vcb_debit(Value {
```
zbuc, line 221: is the idea that the debit will underflow and cause an error to be returned in the value circuit breaker triggered case? it might be good to explicitly call that out

zbuc (approved):
Looks good overall, except the reduction in test coverage

crates/core/component/dex/src/component/circuit_breaker/value.rs
hdevalence: They don't compile, the tests need to be updated / rewritten, that's the remaining work on the PR that needs to be picked up.

crates/core/component/dex/src/component/circuit_breaker/value.rs
hdevalence: Sure, or we could also / instead fire off an ABCI event, so that it's possible to build a dashboard of inflows and outflows.

crates/core/component/dex/src/component/dex.rs
hdevalence: Yep

erwanor:
This breaks a few tests, as expected because we moved the vb credits to the action handlers:

```
        FAIL [   0.053s] penumbra-dex component::router::tests::best_position_route_and_fill
        FAIL [   0.116s] penumbra-dex component::router::tests::multi_hop_route_and_fill
```

I think we can fix them without too much effort by doing:

```rust
    let infinite_gm = Value {
        asset_id: gm.id(),
        amount: Amount::from(100000u128) * gm.unit_amount(),
    };

    let infinite_gn = Value {
        asset_id: gn.id(),
        amount: Amount::from(100000u128) * gn.unit_amount(),
    };

    let infinite_penumbra = Value {
        asset_id: penumbra.id(),
        amount: Amount::from(100000u128) * penumbra.unit_amount(),
    };

    state_tx.vcb_credit(infinite_gm).await?;
    state_tx.vcb_credit(infinite_gn).await?;
    state_tx.vcb_credit(infinite_penumbra).await?;
 ```

We will be able to replace them with something much better but in the meantime it seems useful to keep them around as basic checks.

erwanor (approved):
ACK on general implementation, merge on green CI

PR #4094, opened 2024-03-25
cnidarium: use `WriteBatch` when writing jmt nodes

erwanor:
This PR is a prelude to implementing "staged" writes to our storage layer (cc @noot #4095 ). Specifically, it:
- inline the `write_node_batch` logic inside of `SubstoreStorage::commit`
- make sure that nodes are inserted into the `rocksdb::WriteBatch` rather than immediately written to disk

The last point was a bug, we want to commit to a set of writes atomically, but the tree changes were written ahead of the rest of the transaction. By inserting a `panic!` statement between the tree write and the `write(node_batch` statement and restarting the full node I was able to observe a consistency discrepancy. This PR fixes this.


erwanor:
I am merging this optimistically ahead of a pairing session on #4095. I tested this manually in different directions (simulating a disk failure, running proptests for migrations).

Issue #1812, opened 2023-01-06
upgrades: create a template guide 


erwanor:
We should create a template that specifies the different steps that one has to take in order to upgrade their full node. This document should recommend best practices such as creating multiple backups, checking the hash of migrated state, handling common setups (e.g. systemd etc.)

erwanor:
This is not immediately actionable since we have yet to begin work on #1804, but this could be a good first issue once the `on-hold` tag is removed, maybe worth bookmarking if that's interesting for you!

erwanor:
@conorsch has been leading/driving the work on this front with the [wiki](https://github.com/penumbra-zone/penumbra/wiki/Performing-upgrades) entry

conorsch:
Was tracking this work as part of #3841, see details there. This will be resolved as soon as #4097 lands.

conorsch:
This is done, BUT we have changes to make as a result of feedback. Will close when those additions lands.

conorsch:
> This is done, BUT we have changes to make as a result of feedback. Will close when those additions lands.

Picking this up, will have a revised version of the docs up by tomorrow.

Issue #4087, opened 2024-03-22
Testnet 70, Deimos 6, The Upgrading


conorsch:
# Testnet Release

We are planning two consecutive releases, v0.70.0 and v0.71.0, in order to exercise chain upgrade logic (#1804). Accordingly, we will need to vote on a upgrade proposal to halt the chain, and provide documentation to node operators on the specific upgrade steps.

Testnet name: Deimos 6
Release date: 2024-03-26
Testnet release manager: @conorsch

Testnet name: Deimos 6 [sic]
Release date: 2024-03-28
Testnet release manager: @conorsch

# Testnet Release Manager Checklist

Friday before release:

- [x] Construct the genesis data for the release:
  - [x] Create new testnet directory with initial genesis allocations for this testnet by running `cd testnets && ./new-testnet.sh`
    - This genesis data will be used for `testnet-preview` with a randomized version of the future testnet's chain ID.

- [x] Check in with team again in a release meeting and update the GitHub milestone to ensure it represents what will make it into the testnet.
- [x] Draft an announcement for peer review to ensure major changes included are comprehensive.

Following Tuesday (release day):

- [x] Verify that `testnet-preview.penumbra.zone` is operational; it is redeployed on every push to main, and is an exact preview of what is about to be deployed.
- [x] Bump the version number and push its tag, via [cargo-release](https://crates.io/crates/cargo-release).
    - [x] Run `cargo release minor` for a new testnet, or `cargo release patch` for a bugfix. For the latter, make sure you're on a dedicated release branch.
    - [x] Push the commit and newly generated tag, e.g. `v0.51.0`, to the remote.
- [x] Wait for the ["Release" workflow](https://github.com/penumbra-zone/penumbra/actions/workflows/release.yml) to complete: it'll take ~60m, most of which is the Windows builds.
- [x] Edit the newly created (and published) release object, then click "Generate release notes." Cut and paste the generated text from the bottom to the top of the post, then save it.
- [x] You must [manually review](https://docs.github.com/en/actions/managing-workflow-runs/reviewing-deployments) the [`Waiting` deployment in the GitHub Actions UI](https://github.com/penumbra-zone/penumbra/actions/workflows/deploy-testnet.yml) before the deployment will begin. Monitor the GitHub action to ensure it completes after it is approved.
- [x] Delegate to the Penumbra Labs CI validators; use amounts of ~200k `penumbra` per validator.
[...]

PR #4097, opened 2024-03-25
docs: chain upgrade procedure, for operators

### LINKED ISSUE #3841: pd: support state tarball for joining nodes
When chain upgrades are performed (#1804), pd state may be collapsed by a migration, such that late-joining nodes (i.e. nodes that join the network after the upgrade boundary has passed) will not be able to verify historical state. To support late-joining nodes, we must provide the capability for `pd testnet join` to accept compressed archives of historical pd state, and use them during bootstrapping.

Proposal: add new optional flag `--archive-url=<URL>` to `pd testnet join`. Doing so will allow late-joining nodes to pull down a compressed archive from a remote URL, and extract that archive as starting state for pd. 

Specifically, this requires:

* [x] Defining archive format and structure (e.g. "all files/directories should be extracted to `~/.penumbra/testnet_data/node0/pd`"). My understanding is we'll need at least 1) rocks db and 2) genesis file in all cases.
* [x] Providing hosting capability for future snapshots (ideally community validators will assist with this process, but we still need to host snapshots we create somewhere)
* [x] Write logic for `pd testnet join --archive-url <url>` - https://github.com/penumbra-zone/penumbra/pull/4055
* [ ] Write user-facing documentation for using the flag - https://github.com/penumbra-zone/penumbra/pull/4097
* [x] Write developer-facing documentation for storing and updating snapshots.

conorsch:
Adds documentation for the specific steps to be performed by a node operator in order to participate in a chain upgrade. These docs are largely based on the existing wiki notes [0], adapted for a generalized Penumbra setup.

Similarly, sketched out some corresponding changes to the `pd testnet join` docs, but left them commented out for now: those docs are only relevant for joining a chain that has already been upgraded. If things go well with #4087, we'll uncomment those docs and start using them.

Finishes and therefore closes #3841.

[0] https://github.com/penumbra-zone/penumbra/wiki/Performing-upgrades

### DIFF
--- a/docs/guide/src/pd/chain-upgrade.md
+++ b/docs/guide/src/pd/chain-upgrade.md
@@ -1,0 +1,73 @@
+# Performing chain upgrades
+
+When consensus-breaking changes are made to the Penumbra protocol,
+node operators must coordinate upgrading to the new version of the software
+at the same time. Penumbra uses a governance proposal for scheduling upgrades
+at a specific block height.
+
+## Upgrade process abstractly
+
+At a high level, the upgrade process consists of the following steps:
+
+1. Governance proposal submitted, specifying explicit chain height `n` for halt to occur.
+2. Governance proposal passes.
+3. Chain reaches specified height `n-1`, nodes stop generating blocks.
+4. Manual upgrade is performed on each validator and fullnode:
+  1. Prepare migration directory via `pd export`.
+  2. Install the new version of pd.
+  3. Apply changes to node state via `pd migrate`.
+  4. Copy a few files and directories around, clean up CometBFT state.
+  5. Restart node.
+
+After the node is restarted on the new version, it should be able to talk to the network again.
+Once enough validators with sufficient stake weight have upgraded, the network
+will resume generating blocks.
+
+
+## Genesis time
+
+In order for the chain to start again after the upgrade, all nodes must be using the same genesis information,
+including the timestamp for the genesis event. While the `pd migrate` command will create a new `genesis.json` file,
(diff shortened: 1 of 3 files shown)


conorsch:
Also closes #1812.

docs/guide/src/pd/chain-upgrade.md
```diff
@@ -0,0 +1,56 @@
...
+4. Download the latest version of `pd` and install it. Run `pd --version` and confirm you see `{{ #include ../penumbra_version.md }}` before proceeding.
+<!--
+An example log message emitted by `pd migrate` without providing `--genesis-start`:
+
+    pd::upgrade: no genesis time provided, detecting a testing setup now=2023-12-09T00:08:24.225277473Z`
+
+The value after `now=` is what should be copied. In practice, for testnets, Penumbra Labs will advise on a genesis time
+and provide that value in the documentation. Or should we just pick a genesis start ahead of time, and use that for all?
+-->
+5. Apply the migration: `pd migrate --genesis-start "GENESIS_TIME" --target-directory ~/.penumbra/testnet_net/node0/pd-exported-state/ --migrate-archive ~/.penumbra/testnet_data/node0/pd-migrated-state-{{ #include ../penumbra_version.md }}.tar.gz`.
```
erwanor, line 44: I think I would add a section about genesis time and explain that this is decided by validators/node runners, possibly via the governance proposal itself (non-binding, opt-in). And describe how genesis time works: if it is set in the future, the node will start and wait for the time of genesis to be reached before it can start producing blocks (and peering?).

conorsch:
Great comments, @erwanor, thanks for the deep read. I've amended the docs according to your suggestions.

PR #4098, opened 2024-03-25
dex: streamline `PositionManager`

hdevalence:
This implements an API suggestion originally made by @cronokirby to expose context-specific methods for updating positions.

We maintain a single `put_position` method on the `Inner` trait so that we're
able to have a single point where we edit any necessary indexes, but this is
now not exposed to callers. Instead they call a scope-specific method that
calls `put_position` internally.

Currently there's still room to improve the API safety but we'll want to
potentially change the signature of `put_position` to ensure that this remains
efficient. It would be ideal if it were possible to do all reads up front and
concurrently. Further investigation on this point would be desirable.

This commit has a security-critical change in that the logic for position
withdrawals' balance commitment changes, and it'd be good to get review on that
part in particular.

### DIFF
--- a/crates/core/component/dex/src/component/action_handler/position/withdraw.rs
+++ b/crates/core/component/dex/src/component/action_handler/position/withdraw.rs
@@ -1,16 +1,11 @@
-use anyhow::{anyhow, Result};
+use anyhow::Result;
 use ark_ff::Zero;
 use async_trait::async_trait;
 use cnidarium::StateWrite;
 use cnidarium_component::ActionHandler;
 use decaf377::Fr;
-use penumbra_proto::StateWriteProto;
 
-use crate::{
-    component::{PositionManager, PositionRead, ValueCircuitBreaker},
-    event,
-    lp::{action::PositionWithdraw, position, Reserves},
-};
+use crate::{component::PositionManager, lp::action::PositionWithdraw};
 
 #[async_trait]
 /// Debits a closed position NFT and credits a withdrawn position NFT and the final reserves.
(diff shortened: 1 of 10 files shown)


crates/core/component/dex/src/component/position_manager.rs
```diff
@@ -203,9 +203,24 @@ pub trait PositionManager: StateWrite + PositionRead {
...
+            if position.reserves.r1 == 0u64.into() || position.reserves.r2 == 0u64.into() {
+                tracing::debug!(
+                    id = ?position.id(),
+                    r1 = ?position.reserves.r1,
+                    r2 = ?position.reserves.r2,
+                    "marking position as closed due to close-on-fill"
+                );
+                position.state = position::State::Closed;
+            }
+        }
```
hdevalence, line 219: @erwanor this can be done safely now that we're distinguishing position execution from other position handling

crates/core/component/dex/src/component/tests.rs
```diff
@@ -257,6 +258,59 @@ async fn single_limit_order() -> anyhow::Result<()> {
...
+async fn check_close_on_fill() -> anyhow::Result<()> {
+    let storage = TempStorage::new().await?.apply_minimal_genesis().await?;
+    let mut state = Arc::new(StateDelta::new(storage.latest_snapshot()));
+    let mut state_tx = state.try_begin_transaction().unwrap();
+
+    let gm = asset::Cache::with_known_assets().get_unit("gm").unwrap();
+
+    let mut position_1 = SellOrder::parse_str("100gm@1gn")?.into_position(OsRng);
+    position_1.close_on_fill = true;
+    let position_2 = SellOrder::parse_str("100gm@1.1gn")?.into_position(OsRng);
```
hdevalence, line 273: this is much nicer for writing tests than many of the existing methods (which were made prior to this api)

crates/core/component/dex/src/component/position_manager.rs
erwanor: This looks great, I think we should add a guard just to be paranoid against indexing/validation failures. The reasoning is that the inflation bug was two-pronged: you could create positions with duplicate IDs, and position execution/handling was commingled.

crates/core/component/dex/src/component/position_manager.rs
```diff
@@ -128,9 +178,141 @@ pub trait PositionManager: StateWrite + PositionRead {
...
+
+        self.record_proto(event::position_execution(&position));
+        self.update_position(position).await?;
+
+        Ok(())
+    }
+
+    /// Withdraw from a closed position, incrementing its sequence number.
+    ///
+    /// Updates the position's reserves and rewards to zero and returns the withdrawn balance.
```
erwanor, line 229: It doesn't actually zero the rewards, right?

crates/core/component/dex/src/component/action_handler/position/withdraw.rs
```diff
@@ -27,25 +22,22 @@ impl ActionHandler for PositionWithdraw {
...
-
+        //
         // Note: since this is forming a commitment only to the reserves,
         // we are implicitly setting the reward amount to 0. However, we can
         // add support for rewards in the future without client changes.
-        let expected_reserves_commitment = metadata
-            .reserves
-            .balance(&metadata.phi.pair)
-            .commit(Fr::zero());
+        let expected_reserves_commitment = actual_reserves.commit(Fr::zero());
```
erwanor, line 40: ACK: I am convinced that this is correct because `position_withdraw` returns the withdrawn reserves

erwanor (approved):

PR #4100, opened 2024-03-25
compact block: add epoch index to compact block

### LINKED ISSUE #2689: Include the epoch index in the `CompactBlock` as well as the block height.

Computing a nullifier for a state fragment requires knowing the position of the fragment.  Client implementations need to compute nullifiers of state fragments as they're detected, since nullifiers are used to detect when those state fragments are consumed.

However, our current compact block definition doesn't allow computing the position of the fragments it contains without access to other state (namely, a synchronized SCT instance).  Why? The position has three components: (commitment index, block index, epoch index). The commitment index is implicitly specified by the order of the fragment within the message. The block index is just the height. But the epoch index is not included in the compact block, so a client can only get it by cross-referencing with their SCT instance.

If we include the epoch index in the `CompactBlock`, however, we can statelessly compute positions of any fragment contained in the compact block.

Add the epoch index to the compact block with a low (sub-15) field number (to save space, since it will always be present).

aubrika:
closes #2689


proto/penumbra/penumbra/core/component/compact_block/v1/compact_block.proto
```diff
@@ -28,6 +28,8 @@ message CompactBlock {
   bool app_parameters_updated = 9;
   // Updated gas prices, if they have changed.
   fee.v1.GasPrices gas_prices = 10;
+  // The epoch index
+  uint64 epoch_index = 11;
```
conorsch, line 32: It would be grand to have a more descriptive docstring here, incorporating some of the context from #2689, that speaks to what this value represents, as well as why and how it might be used. Unfortunately, I'm a bit at a loss to suggest specific phrasing.

aubrika: I was also at a loss on this one ... maybe @hdevalence or perhaps @plaidfinch could shed some light on the "epoch index" for future readers of our documentation strings?

erwanor: This is the counter that identifies/index each epoch similar to the height of a block

erwanor: but for epochs

erwanor:
@conorsch I am not sure that this is consensus breaking; I could be wrong, but `CompactBlock`s are stored in nonverifiable storage and, in terms of interfaces, the change looks forward compatible because it is just adding a field to the proto definitions

erwanor (approved):

PR #4101, opened 2024-03-26
ibc: use outbound_ics20_transfers_enabled chain parameter

avahowell:
this uses the existing outbound_ics20_transfers_enabled to gate ics20 withdrawals


conorsch:
Discussed with @avahowell & @erwanor : we're going to merge this today, so it's included in v0.70.0 (#4087).

PR #4104, opened 2024-03-26
app: check asset id in fee stateful check

### LINKED ISSUE #4106: app: fees can be paid in any asset
Zellic noticed that our [`fee_greater_than_base_fee`](https://github.com/penumbra-zone/penumbra/blob/58f14d3fe08517a95ae9cdeddb4623e349f81927/crates/core/app/src/action_handler/transaction/stateful.rs) implementation does not check that the fee is paid using the staking token.

erwanor:
Fix #4106.

@hdevalence I remember discussing with you making the `fee` method return a `Value`, but I don't recall why we decided against it.

### DIFF
--- a/crates/core/app/src/action_handler/transaction/stateful.rs
+++ b/crates/core/app/src/action_handler/transaction/stateful.rs
@@ -1,4 +1,4 @@
-use anyhow::Result;
+use anyhow::{ensure, Result};
 use cnidarium::StateRead;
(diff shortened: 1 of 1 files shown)


hdevalence (approved):

hdevalence:
Hmm, I didn't remember that discussion (could very well have happened though). Fix looks good!

PR #4107, opened 2024-03-26
dex: move `SwapExecution`s to NV storage

### LINKED ISSUE #4084: Remove `SwapExecution`s from consensus state

We store all of the execution traces in the `SwapExecution` in consensus-verifiable storage. This data is never used by the application and the only reason we put it there was that at the time we developed the feature we didn't have an indexer.

Move it out to nonverifiable storage, so that we could remove it later if desired without breaking the app hash.

**Additional context**
See `state_key::swap_execution`

erwanor:
Close #4084, please do not merge.


erwanor:
This is ready to review and receive ACKs, but we should not merge it.

conorsch (approved):
Approving. This will shortly go out as part of `v0.71.0`, in which we ship a migration ship to perform a chain upgrade (#4087).

PR #4108, opened 2024-03-26
upgrades: write a migration script for #4084

erwanor:
This implements a migration for #4084 and makes it the default command in `pd migrate`.

### DIFF
--- a/crates/bin/pd/src/main.rs
+++ b/crates/bin/pd/src/main.rs
@@ -13,7 +13,7 @@ use cnidarium::{StateDelta, Storage};
 use metrics_exporter_prometheus::PrometheusBuilder;
 use pd::{
     cli::{Opt, RootCommand, TestnetCommand},
-    migrate::Migration::SimpleMigration,
+    migrate::Migration::Testnet70,
     testnet::{
         config::{get_testnet_dir, parse_tm_address, url_has_necessary_parts},
         generate::TestnetConfig,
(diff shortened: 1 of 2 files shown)


erwanor:
Something to note is that in this case, we can get away with using a prefix query because we're not trying to capture a range, but in the future we might to run more complex migrations that operate over a specific block window. We could do prefix+filtering but this might end up being much more expensive than a range query still. 

I am running a test on a private chain, if it works we can leave it as is. Otherwise, we'll fallback on the `SimpleMigration` strategy when we do the next pd release (in addition to incrementing the `TOTAL_HALT_COUNTER`).

crates/bin/pd/src/main.rs
```diff
@@ -13,7 +13,7 @@ use cnidarium::{StateDelta, Storage};
 use metrics_exporter_prometheus::PrometheusBuilder;
 use pd::{
     cli::{Opt, RootCommand, TestnetCommand},
-    migrate::Migration::SimpleMigration,
+    migrate::Migration::Testnet70,
```
conorsch, line 16: Nit: should this be Testnet71? That would make it clearer that we expect it to ship with `v0.71.0`. It's not exposed to operators, so we just need to keep the terminology straight ourselves.

conorsch (approved):
LGTM. Intentionally _not_ merging so that we can coordinate timing carefully.

crates/bin/pd/src/main.rs
erwanor: SG, my reasoning was 70 -> 71, but the other way around works too

PR #4017, opened 2024-03-13
staking: 🎁 add additional tracing telemetry

cratelyn:
this is a group of improvements to the telemetry in the staking component, cherry-picked out of #4001. this improves the logs seen in tests, and additionally takes the time to make some other small gardening tweaks: consolidating imports, adding documentation comments, etc. this is another branch in the spirit of #4009.

the changes here are itemized into distinct, small code transformations to facilitate review, and demonstrate that these are noop changes. you can start [here](https://github.com/penumbra-zone/penumbra/pull/4017/commits/f7f78b91e7b368377b55cdfba1e84ea7459609a9).

* f33196474 - feat(staking): ❗ traces for invalid or missing parameters 
* d33598d94 - docs(staking): 🤝 add a doc comment 
* f6c60a8bb - staking: 🙂 remove needless `Ok(_)` 
* fd4b8da09 - feat(staking): 💬 log when delegation changes are not found 
* fe597ff16 - feat(staking): 🔍 trace-level spans for `StateReadExt` methods 
* 04476004b - refactor(staking): 🚡 use `self::` prefix in reexports 
* 67893c92d - refactor(staking): 🙂 remove import from reexports 
* 043222e89 - refactor(staking): 🎁 consolidate public submodules 
* 637679619 - refactor(staking): 🎁 consolidate submodules 
* b5bd24070 - refactor(staking): 👯 component flag reexports StateWriteExt 
* d57beaaa5 - chore(staking): 🥡 consolidate component reexports 
* aa52f6d8c - chore(staking): 🧹 tidy epoch_handler imports 
* 765ed75d6 - refactor(compact-block): 🧹 tidy imports 
* 7cb93f7d0 - feat(staking): 🌐 log why epoch ended 
* 081d041aa - refactor(staking): add context to `state.end_epoch` call 
* 570129519 - feat(app): 🦠 tracing telemetry for `App::end_block` 
* f7f78b91e - docs(staking): 📕 add doc comment for `end_epoch`

Issue #3961, opened 2024-03-06
Use error codes in error messages that will reach the frontend


jessepinho:
**Is your feature request related to a problem? Please describe.**
There are a number of errors that can occur in core that get surfaced to the minifront during failed RPC calls or when broadcasting a transaction.

For example:
- If a user prepares an undelegation plan for a given epoch, but doesn't submit the transaction until a later epoch, core will throw [this error](https://github.com/penumbra-zone/penumbra/blob/v0.68.2/crates/core/component/stake/src/component/action_handler/undelegate.rs#L34): `undelegation was prepared for next epoch {} but the next epoch is {}`
- If a user prepares an undelegation claim before the unbonding tokens are eligible for claiming, core will throw [this error](https://github.com/penumbra-zone/penumbra/blob/v0.68.2/crates/core/component/stake/src/component/action_handler/undelegate_claim.rs#L44): `cannot claim unbonding tokens before the end epoch (current epoch: {}, unbonding epoch: {})`
- etc.

By the time these errors reach the frontend code (via Tendermint logs), it has text that looks like this: `ConnectError: [unknown] Tendermint: failed to deliver transaction: check_stateful failed: undelegation was prepared for next epoch 16 but the next epoch is 23` We obviously don't want to show this text to the user, but we also don't want to parse the text into something human-readable.

At the moment, the way we deal with this error on the frontend is to maintain [an array of known errors](https://github.com/penumbra-zone/web/blob/v4.2.0/packages/query/src/queriers/tendermint.ts#L8-L13), which is brittle and will break if anyone on the core team changes the text of an error. And then we show the full text of the error to the user.

**Describe the solution you'd like**
I'm proposing that we create a standardized set of error codes for all the different error cases. They could be numerical (`penumbra(1234)`) or strings (`UNDELEGATION_PREPARED_FOR_WRONG_EPOCH`) or whatever other style people prefer, as long as they're consistent. They could be included in the existing error messages in a standardized way — e.g., `[UNDELEGATION_PREPARED_FOR_WRONG_EPOCH] undelegation was prepared for next epoch {} but the next epoch is {}`. This way, on the frontend, we could just look for the machine-readable error code, and then decide how to display it to the user.

hdevalence:
The `CheckTx` method actually provides a numeric error code.

[...]

PR #4110, opened 2024-03-26
tests(app): ✨ add integration test coverage for undelegation

cratelyn:
this introduces additional test coverage for the staking component, examining the mock client's notes during delegation, the unbonding period, and after claiming an undelegation.

* #4115
* #3995

### DIFF
--- a/crates/core/app/tests/app_can_undelegate_from_a_validator.rs
+++ b/crates/core/app/tests/app_can_undelegate_from_a_validator.rs
@@ -1,0 +1,374 @@
+mod common;
+
+use {
+    self::common::BuilderExt,
+    anyhow::anyhow,
+    ark_ff::UniformRand,
+    cnidarium::TempStorage,
+    penumbra_app::server::consensus::Consensus,
+    penumbra_genesis::AppState,
+    penumbra_keys::test_keys,
+    penumbra_mock_client::MockClient,
+    penumbra_mock_consensus::TestNode,
+    penumbra_proto::DomainType,
+    penumbra_sct::component::clock::EpochRead as _,
+    penumbra_stake::{
+        component::validator_handler::ValidatorDataRead as _, validator::Validator,
+        UndelegateClaimPlan,
+    },
+    penumbra_transaction::{
+        memo::MemoPlaintext, plan::MemoPlan, TransactionParameters, TransactionPlan,
+    },
+    rand_core::OsRng,
+    tap::Tap,
+    tracing::{error_span, info, Instrument},
+};
+
+/// The length of the [`penumbra_sct`] epoch.
+///
+/// This test relies on many epochs turning over, so we will work with a shorter epoch duration.
+const EPOCH_DURATION: u64 = 2;
+
+/// The length of the [`penumbra_stake`] unbonding_delay.
+const UNBONDING_DELAY: u64 = 4;
+
+#[tokio::test]
+async fn app_can_undelegate_from_a_validator() -> anyhow::Result<()> {
+    // Install a test logger, acquire some temporary storage, and start the test node.
+    let guard = common::set_tracing_subscriber();
+    let storage = TempStorage::new().await?;
+
+    // Helper function to get the latest block height.
+    let get_latest_height = || async {
+        storage
+            .latest_snapshot()
+            .get_block_height()
+            .await
+            .expect("should be able to get latest block height")
+    };
+
+    // Helper function to get the latest epoch.
+    let get_latest_epoch = || async {
+        storage
+            .latest_snapshot()
+            .get_current_epoch()
+            .await
+            .expect("should be able to get curent epoch")
+    };
+
+    // Configure an AppState with slightly shorter epochs than usual.
+    let app_state = AppState::Content(penumbra_genesis::Content {
+        sct_content: penumbra_sct::genesis::Content {
+            sct_params: penumbra_sct::params::SctParameters {
+                epoch_duration: EPOCH_DURATION,
+            },
+        },
+        stake_content: penumbra_stake::genesis::Content {
+            stake_params: penumbra_stake::params::StakeParameters {
+                unbonding_delay: UNBONDING_DELAY,
+                ..Default::default()
+            },
+            ..Default::default()
+        },
+        ..Default::default()
+    });
+
+    // Start the test node.
+    let mut node = {
+        let consensus = Consensus::new(storage.as_ref().clone());
+        TestNode::builder()
+            .single_validator()
+            .with_penumbra_auto_app_state(app_state)?
+            .init_chain(consensus)
+            .await
+    }?;
+
+    // Retrieve the validator definition from the latest snapshot.
+    let Validator { identity_key, .. } = match storage
+        .latest_snapshot()
+        .validator_definitions()
+        .tap(|_| info!("getting validator definitions"))
+        .await?
+        .as_slice()
+    {
+        [v] => v.clone(),
+        unexpected => panic!("there should be one validator, got: {unexpected:?}"),
+    };
+    let delegate_token_id = penumbra_stake::DelegationToken::new(identity_key).id();
+
+    // Sync the mock client, using the test wallet's spend key, to the latest snapshot.
+    let mut client = MockClient::new(test_keys::SPEND_KEY.clone())
+        .with_sync_to_storage(&storage)
+        .await?
+        .tap(|c| info!(client.notes = %c.notes.len(), "mock client synced to test storage"));
+
+    // Now, create a transaction that delegates to the validator.
+    //
+    // Hang onto the staking note nullifier, so we can interrogate whether that note is spent.
+    let (plan, staking_note_nullifier) = {
+        use {
+            penumbra_shielded_pool::{OutputPlan, SpendPlan},
+            penumbra_transaction::{
+                memo::MemoPlaintext, plan::MemoPlan, TransactionParameters, TransactionPlan,
+            },
+        };
+        let snapshot = storage.latest_snapshot();
+        let note = client
+            .notes_by_asset(*penumbra_asset::STAKING_TOKEN_ASSET_ID)
+            .cloned()
+            .next()
+            .expect("should get staking note");
+        let rate = snapshot
+            .get_validator_rate(&identity_key)
+            .await?
+            .ok_or(anyhow!("validator has a rate"))?
+            .tap(|rate| tracing::info!(?rate, "got validator rate"));
+        let spend = SpendPlan::new(
+            &mut rand_core::OsRng,
+            note.clone(),
+            client
+                .position(note.commit())
+                .expect("note should be in mock client's tree"),
+        );
+        let staking_note_nullifier = spend.nullifier(&client.fvk);
+        let delegate = rate.build_delegate(
+            storage.latest_snapshot().get_current_epoch().await?,
+            note.amount(),
+        );
+        let output = OutputPlan::new(
+            &mut rand_core::OsRng,
+            delegate.delegation_value(),
+            *test_keys::ADDRESS_1,
+        );
+        let mut plan = TransactionPlan {
+            actions: vec![spend.into(), output.into(), delegate.into()],
+            // Now fill out the remaining parts of the transaction needed for verification:
+            memo: MemoPlan::new(&mut OsRng, MemoPlaintext::blank_memo(*test_keys::ADDRESS_0))
+                .map(Some)?,
+            detection_data: None, // We'll set this automatically below
+            transaction_parameters: TransactionParameters {
+                chain_id: TestNode::<()>::CHAIN_ID.to_string(),
+                ..Default::default()
+            },
+        };
+        plan.populate_detection_data(rand_core::OsRng, 0);
+        (plan, staking_note_nullifier)
+    };
+    let tx = client.witness_auth_build(&plan).await?;
+
+    // Show that the client does not have delegation tokens before delegating.
+    assert_eq!(
+        client.notes_by_asset(delegate_token_id).count(),
+        0,
+        "client should not have delegation tokens before delegating"
+    );
+
+    // Execute the transaction, applying it to the chain state.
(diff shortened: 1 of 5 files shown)


crates/core/app/tests/app_can_undelegate_from_a_validator.rs
```diff
@@ -0,0 +1,344 @@
...
+        }
+        client.sync_to_latest(storage.latest_snapshot()).await?;
+        assert_eq!(
+            count_unbonding_tokens(&client),
+            0,
+            "client should still not have unbonding tokens in final block of unbonding period"
+        );
+    }
+
+    // Step ahead one block, and witness that we now should have an unbonding token.
```
cratelyn, line 318: having trouble getting this assertion to pass. i'll track this down next.

cratelyn:
```
  2024-03-26T20:19:57.017320Z  INFO penumbra_app::server::consensus: deliver_tx failed, e: failed to deliver transaction                                  
                                                                                                                                                          
Caused by:                                                                                                                                                
    0: executing transaction                                                                                                                              
    1: undelegation was prepared for epoch 0 but the current epoch is 1                                                                                   
    at crates/core/app/src/server/consensus.rs:207                                                                                                        
    in penumbra_mock_consensus::block::execute with height: 4, time: 1711484396                                                                           
    in app_can_undelegate_from_a_validator::executing block with undelegation transaction   
```

crates/core/app/tests/app_can_undelegate_from_a_validator.rs
```diff
@@ -0,0 +1,375 @@
...
+    };
+    let tx = client.witness_auth_build(&plan).await?;
+
+    // Execute the transaction, applying it to the chain state.
+    node.block()
+        .add_tx(tx.encode_to_vec())
+        .execute()
+        .instrument(error_span!("executing block with undelegation claim"))
+        .await?;
+    let post_claim_snapshot = storage.latest_snapshot();
```
cratelyn, line 342:
```
  2024-03-27T01:50:24.883510Z  INFO penumbra_app::server::consensus: deliver_tx failed, e: failed to deliver transaction                                  
                                                                                                                                                          
Caused by:                                                                                                                                                
    0: executing transaction                                                                                                                              
    1: nullifier 4b8e1e728ad20b137bd1252cd456e6446d0c32de98d093e3620769fd7f559712 was already spent in "be166bd47e7cace1434b122d2eb08095dd0caad4e621bc34c2
f82bedc85972f1"                                                                                                                                           
    at crates/core/app/src/server/consensus.rs:207                                                                                                        
    in penumbra_mock_consensus::block::execute with height: 9, time: 1711504224                                                                           
    in app_can_undelegate_from_a_validator::executing block with undelegation claim 
```

crates/core/component/stake/src/component/action_handler/undelegate.rs
```diff
@@ -56,12 +63,18 @@ impl ActionHandler for Undelegate {
             })?;
         let expected_unbonded_amount = rate_data.unbonded_amount(u.delegation_amount);
 
-        ensure!(
```
erwanor, line 59: What do you think of `check_and_execute(...).map_err(|e| error!(?e))` instead? That way we don't need to add individual tracing statements and we just need to focus on making the error descriptive, which we will have to do as a prelude to using error codes in `CheckTx/DeliverTx` ([#3961](https://github.com/penumbra-zone/penumbra/issues/3961))

erwanor: And we can keep the short `ensure!(invariant, msg)` statements

crates/core/app/tests/app_can_undelegate_from_a_validator.rs
```diff
@@ -0,0 +1,374 @@
...
+        assert_eq!(
+            client.notes_by_asset(undelegate_token_id).count(),
+            1,
+            "client should still have undelegation notes"
+        );
+        assert_eq!(
+            client.notes_by_asset(delegate_token_id).count(),
+            1,
+            "client should still have delegation notes"
+        );
```
erwanor, line 366: Just two ideas, it would be good to check:
- the rate at which delegation <> undelegation token gets converted (by dividing the amounts and comparing it against the stated validator rate)
- the rate at which undelegation <> staking token gets converted (so that later we can pull that logic to check that slashing events effect that rate correctly)

crates/core/app/tests/app_can_undelegate_from_a_validator.rs
```diff
@@ -0,0 +1,374 @@
...
+        .await?;
+    let post_undelegate_snapshot = storage.latest_snapshot();
+
+    // Compute the height we expect to see this unbonding period finish.
+    let expected_end_of_unboding_period_height = {
+        // TODO(kate): right now the unbonding delay is calculated by using the start of the epoch
+        // as a beginning. so rather than...
+        // post_undelegated_height + UNBONDING_DELAY
+        // ...we write:
+        pre_undelegated_epoch.start_height + UNBONDING_DELAY
```
erwanor, line 278: There's a `compute_unbonding_height` routine maybe we should create a stateless version of it that we can reuse in tests and other places? This works too

erwanor (approved):
+1 on structure / scope, added some comments

crates/core/app/tests/app_can_undelegate_from_a_validator.rs
cratelyn: i've pushed 84516c6c34bec87ba85542ba007c1690e283ac58 as an initial step in that direction. i have read through the protocol specification, but i am still not very familiar with the rates/arithmetic involved here.

would you be available to pair on this sometime tomorrow (the 28th) to step through this?

erwanor: Yeah for sure!

crates/core/component/stake/src/component/action_handler/undelegate.rs
cratelyn: i have some conviction that this is a better pattern, and i'd like to try and convince you of why!

emitting a tracing event at the call-site means that we capture all of the span context that was present, precisely at the moment that an invariant was violated. this means that the tracing subscriber can log an error that looks like:

```
  2024-03-28T13:22:12.703847Z ERROR penumbra_stake::component::action_handler::undelegate: undelegation was prepared for a different epoch, from: Epoch { index: 0, start_height: 0 }, current: Epoch { index: 1, start_height: 2 }                                                                                                                                                                                                             
    at crates/core/component/stake/src/component/action_handler/undelegate.rs:32                                                                                                                                        
    in penumbra_transaction::action::Undelegate with idx: 2                                                                                                                                                             
    in penumbra_mock_consensus::block::execute with height: 3, time: 1711632132                                                                                                                                         
    in app_can_undelegate_from_a_validator::executing block with undelegation transaction
```

this context is lost when we emit tracing error events outside of the function from which the error in question originates. often, the errors propagated back up in the consensus service result in a test failure that looks like this:

```
thread 'app_can_undelegate_from_a_validator' panicked at /home/katie/work/penumbra/crates/core/component/stake/src/component/action_handler/undelegate.rs:32:13:
the service panicked
Error: consensus service returned error
test app_can_undelegate_from_a_validator ... FAILED
```

when this happens we have only the error message itself to go off of, and have to rebuild that context ourselves by finding that error and manually tracking down call-sites.

_zooming out:_ in my experience operating other rust services, there is another benefit to emitting tracing events: tracing is a ["scoped, **structured** logging and diagnostics system."](https://docs.rs/tracing/latest/tracing/#), emphasis mine. emitting structured logs allows operators to rely on key-value extraction provided by [various](https://www.splunk.com/en_us/blog/tips-and-tricks/key-value-pair-extraction-definition-examples-and-solutions.html) [popular](https://help.sumologic.com/docs/search/search-query-language/parse-operators/parse-keyvalue-formatted-logs/) [observability](https://docs.datadoghq.com/logs/log_configuration/parsing/?tab=matchers#key-value-or-logfmt) [platforms](https://www.elastic.co/guide/en/logstash/current/plugins-filters-kv.html). that allows operators to query historical logs over time, and to write more fine-grained alerting rules.

_zooming back in:_ the downside of being a little more verbose is real :pensive: but i found myself needing these to debug the `app_can_undelegate_from_a_validator.rs` test; a helpful error message can tell you what went wrong, but a scoped diagnostics system can tell you precisely when and why something went wrong. i could tell that i'd violated an invariant, but it was hard to tell which part of the test had tripped it without adding these.

this, in tandem with test code that uses `tracing::Instrument`, can give us more helpful test failures locally that precisely state what part of the test code cause an error (_see the "executing block with undelegation transaction" span above_) and all of the intermediate steps that led us to an error. 

this is all a more verbose telling of https://github.com/penumbra-zone/penumbra/pull/4017/commits/fd4b8da093aa8742daf959d5c60f0f3e8864a49e, from #4017, but i hope it sells you a bit on this :slightly_smiling_face:

crates/core/app/tests/app_can_undelegate_from_a_validator.rs
cratelyn: yeah! that's a good idea. i waffled :waffle: on whether or not to compute this here or use that method, but you're right that this is probably easier. fixed in d938eae!

crates/core/component/stake/src/component/action_handler/undelegate.rs
cratelyn: to your point though, because brevity is a noble thing, what do you think of 232ee82? i also like the idea of `check_and_execute` being a succinct list of invariants we enforce.

i tried pulling that logic out into two short helpers at the bottom of the file, and now we have a nice `check_and_execute` that looks like

```
async fn check_and_execute<S: StateWrite>(&self, mut state: S) -> Result<()> {
    /* ----- checks ------ */
    // NB: These checks all formerly happened in the `check_historical` method, if profiling
    // shows that they cause a bottleneck we could (CAREFULLY) move some of them back.

    // # Invariant: the undelegation must have been prepared for the current epoch.
    self.check_that_undelegate_was_prepared_for_current_epoch(&state)
        .await?;

    // # Invariant: The unbonded amount must be correctly computed.
    self.check_that_unbonded_amount_is_correctly_computed(&state)
        .await?;

    /* ----- execution ------ */

    // ...

    Ok(())
}
```

...and that feels potentially in the same spirit as the `ensure!(invariant, msg)` we had prior?

erwanor: This makes sense, thanks for the write up. I think we lose something important by formulating invariants negatively, rather than how they would be written in a spec e.g. "the penalties MUST match". But I agree the observability benefits are worth it. Maybe in the future we will get both, until then this looks good to me.

hdevalence: I strongly that we should not do this refactor, for a few different reasons.

1. We should avoid doing restructuring of the consensus critical application logic at this stage of development without a strongly motivated reason.

2. We should not move action handling logic out of the action handlers.

3. We should not be coupling our application logic to the way we emit tracing events. This is an internal implementation detail.

crates/core/component/stake/src/component/action_handler/undelegate.rs
```diff
@@ -76,3 +44,83 @@ impl ActionHandler for Undelegate {
...
+/// Interfaces for enforcing [`ActionHandler`] invariants.
+impl Undelegate {
+    /// Returns `Ok(())` if this delegation is from the current epoch.
+    ///
+    /// This ensures that the unbonding delay is enforced correctly. Additionally, it helps us
+    /// provide a more helpful error message if an epoch boundary was crossed.
+    async fn check_that_undelegate_was_prepared_for_current_epoch(
+        &self,
+        state: &impl StateWrite,
+    ) -> Result<()> {
```
hdevalence, line 57: We should not structure the code this way.

It spreads the consensus critical part of the ActionHandler out into different places, rather than than having it in one place where it can be inspected locally.

And it leaks the internals of that implementation into a public interface, with methods that may or may not appear depending on compile settings, because the Undelegate type must be independent of the component implementation.

hdevalence (changes requested):
We should not restructure the ActionHandler impls

crates/core/component/stake/src/component/action_handler/undelegate.rs
cratelyn: sounds good. reverted in 7bcd43c56.

crates/core/component/stake/src/component/action_handler/undelegate.rs
cratelyn: https://github.com/penumbra-zone/penumbra/pull/4110#discussion_r1543180355

cratelyn: this was reverted in 7bcd43c

cratelyn: https://github.com/penumbra-zone/penumbra/pull/4110/commits/b513ddbcfc7bd44117eeffbb4815403b8306d108

i've reverted this in b513ddbcfc7bd44117eeffbb4815403b8306d108

crates/core/component/stake/src/component/action_handler/undelegate.rs
cratelyn: reverted in b513ddbcfc7bd44117eeffbb4815403b8306d108

crates/core/app/tests/app_can_undelegate_from_a_validator.rs
cratelyn: okay we now have some additional assertions to check that validator rates look correct :tada:

crates/core/component/stake/src/component/action_handler/undelegate.rs
cratelyn: ...i'm going to mark this as resolved, now that this change has been reverted.

PR #4111, opened 2024-03-26
pcli: fix ics20 withdrawal timeout rounding

avahowell:
The rounding math was wrong previously (it wasn't actually rounding)

### DIFF
--- a/crates/bin/pcli/src/command/tx.rs
+++ b/crates/bin/pcli/src/command/tx.rs
@@ -1064,7 +1064,7 @@ impl TxCmd {
                 }
 
                 // round to the nearest 10 minutes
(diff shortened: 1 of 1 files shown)


conorsch:
Refs https://github.com/penumbra-zone/penumbra/pull/4065. @avahowell can you add some language to the PR submission text explaining the fix? I agree, we can ship this as a point release soon.

hdevalence:
@conorsch Can we fix the PR template so it pre-fills the description?

erwanor (approved):

conorsch:
Template PR rendering fixed in https://github.com/penumbra-zone/penumbra/pull/4124, confirmed in https://github.com/penumbra-zone/penumbra/pull/4125.

PR #4114, opened 2024-03-27
fee: gate `penumbra-proto/rpc` on component feature

erwanor:
After deploying testnet 70, we discovered that an OS specific crate popped up in the dependency tree of the `wasm` crate (used by the web extension). We typically use the `component` feature to prevent this from happening (e.g. mio relies on a variety of platform specific syscalls that simply don't exist in a wasm environment) but one slipped through.


cratelyn (approved):
nice find!

Issue #3940, opened 2024-03-04
Remove transaction-wide byte costs


zbuc:
As identified by @hdevalence here: https://github.com/penumbra-zone/web/pull/663#issue-2166001644 if we were to remove transaction-wide byte costs and instead have transaction costs be solely the summation of the actions included in the transaction's costs.

While looking over the work necessary to accomplish this, I noticed that we aren't even including any transaction-wide block cost at the moment:

```rust
impl GasCost for Transaction {
    fn gas_cost(&self) -> Gas {
        self.actions().map(GasCost::gas_cost).sum()
    }
}
```

## Steps

- [ ] Determine block cost per action and set in the various `GasCost` impls

aubrika:
Currently, action execution costs are all hard-coded at `10` - the checkbox here makes sense, but what is the proposed method for determining a more detailed/varied block cost per action? Or asked another way: how should per-action costs ideally be differentiated?

hdevalence:
The block space has cost should be the number of bytes in the proto encoding of the action. 

This can be computed on the Action by doing encode_to_vec().len() and on the ActionPlan by doing the same thing (if the action doesn't require building) or by figuring out what the byte size of the action will be once it is constructed (without actually doing so)

hdevalence:
For each action, we should:

- identify if it has variable size or fixed size (or close enough to fixed size, e.g., a `uint64` will have variable-length encoding but it's close enough to fixed size for these purposes)
- if fixed size, pick a representative size and hardcode that in the gas schedule
- if variable size, the gas schedule should be `.encode_to_vec().len()`

I think we don't have any variable-size actions that need "building". For instance a `Spend` needs building, but we need to compute a gas cost from a `SpendPlan` before building it, but a spend is fixed-size so we can use a hardcoded value. Or an `IbcRelay` has variable length, but does not need building, the plan is the action itself, so there's no problem. We just want to avoid having to put `action.encode_to_vec()` on an action type where there's a plan.

aubrika:
> For each action, we should:

> * identify if it has variable size or fixed size (or close enough to fixed size, e.g., a `uint64` will have variable-length encoding but it's close enough to fixed size for these purposes)

is this done by determining if there are any variable-size data types (e.g. Vec<>) or via some other means?

> * if fixed size, pick a representative size and hardcode that in the gas schedule

what does "representative size" mean here?

> I think we don't have any variable-size actions that need "building". 

[...]

PR #4116, opened 2024-03-27
fees: add per-action block space calculations & representative values

aubrika:
Draft example of current best interpretation for #3940 - any Action with a bespoke ActionPlan is currently given a block space value of "64" as a placeholder, but obviously that needs to change, just unsure how based on previous ticket comments.


aubrika:
Added some approximate block space estimates based on the proto fields, but for some data types (e.g. bytes, string) of variable size, I'm unsure if the chosen values (64 bytes?) are good approximations - do you have thoughts @hdevalence ?

Also, are there any actions for which this draft is currently computing the length of the encoding which should be hard-coded? I chose the set of actions with bespoke associated `ActionPlans`, but I just want to confirm I understood the prior breakdown correctly.

crates/core/transaction/src/gas.rs
```diff
@@ -31,9 +33,8 @@ pub trait GasCost {
 
 pub fn spend_gas_cost() -> Gas {
     Gas {
-        // Each [`Action`] has a `0` `block_space` cost, since the [`Transaction`] itself
-        // will use the encoded size of the complete transaction to calculate the block space.
-        block_space: 0,
+        // Spends are fixed size, so we hardcode the block space cost.
+        block_space: 352,
```
hdevalence, line 37: Here (& elsewhere) it'd be good to have a comment with a rationale for where the 352 comes from.

Ignoring protobuf framing, I count:

192 byte proof + 64 byte auth_sig + 3 *32 byte body

crates/core/transaction/src/gas.rs
```diff
@@ -47,9 +48,7 @@ pub fn spend_gas_cost() -> Gas {
 
 pub fn output_gas_cost() -> Gas {
     Gas {
-        // Each [`Action`] has a `0` `block_space` cost, since the [`Transaction`] itself
-        // will use the encoded size of the complete transaction to calculate the block space.
-        block_space: 0,
+        block_space: NOTEPAYLOAD_SIZE + 352,
```
hdevalence, line 51: Not sure where this came from since there's no comment, but I count (ignoring proto framing)

NOTEPAYLOAD_SIZE + balance_commitment (32 bytes) + wrapped_memo_key (48 bytes) + ovk_wrapped_key (48 bytes) + zk proof (192) 

which would be + 320 rather than + 352.

The proto message for OutputBody has a comment saying the ovk_wrapped_key is 80 bytes but this seems to be incorrect from looking at the source code and should probably be corrected: https://buf.build/penumbra-zone/penumbra/docs/main:penumbra.core.component.shielded_pool.v1#penumbra.core.component.shielded_pool.v1.OutputBody

hdevalence (commented):
The numbers seem plausible, it would be good to have comments clarifying:

- what is being added up
- why we are not concerned about the exact byte size for the fixed-size actions (e.g., why it's ok to ignore protobuf framing), namely that it makes only a small difference in the action's gas cost and simplifies the accounting

PR #4121, opened 2024-03-27
Improve `update_position` API and position execution events.

### LINKED ISSUE #3844: dex: change `EventPositionExecution` to have old reserves as well as new ones

The [`EventPositionExecution`](https://buf.build/penumbra-zone/penumbra/docs/main:penumbra.core.component.dex.v1#penumbra.core.component.dex.v1.EventPositionExecution) only has the new reserves, not the old reserves, so it's not possible to see the fill when considering the event in isolation.

- [ ] Add `old_reserves_1` and `old_reserves_2` fields to the event with the pre-execution reserves
- [ ] Do needful plumbing to ensure the old reserves can be emitted as part of the event

**Additional context**

This may be useful/needed for the DEX explorer.

hdevalence:
> This commit changes `update_position` so that the caller is responsible for
fetching the previous state and then passing it in. While this adds misuse
potential into the API, the method is only used by the `PositionManager` and it
allows the high-level modification methods to do their own invariant checks
without blocking on extra reads.

> This commit improves event handling for PositionExecution:
>
> - we add the previous balances to the event
> - we add a `context` to the event that allows connecting the micro-scale execution with a high-level context
> - we fix a bug where events were emitted multiple times
> - we avoid emitting events for no-op executions.

I believe that both of these changes are _not_ state-breaking or consensus-breaking but would appreciate review of that.

Closes #3844 (cc @philipjames44)


crates/core/component/dex/src/component/position_manager.rs
```diff
@@ -164,18 +176,13 @@ pub trait PositionManager: StateWrite + PositionRead {
...
+    async fn close_queued_positions(&mut self) -> Result<()> {
         let to_close = self.pending_position_closures();
         for id in to_close {
-            match self.close_position_by_id(&id).await {
-                Ok(()) => tracing::debug!(?id, "position closed"),
-                // The position was already closed, which in and of itself is not an error.
-                // It's possible that the position was closed by the engine, for example
-                // because it was a limit-order.
-                Err(e) => tracing::debug!(?id, "failed to close position: {}", e),
-            }
```
hdevalence, line 176: I believe this logic was incorrect, it would suppress _any_ errors, not just an error about a closed position.

Now `close_position_by_id` enforces its own state transitions internally (allows no-op closures).

proto/penumbra/penumbra/core/component/dex/v1/dex.proto
```diff
@@ -651,6 +651,12 @@ message EventPositionExecution {
   num.v1.Amount reserves_1 = 3;
   // The reserves of asset 2 of the position after execution.
   num.v1.Amount reserves_2 = 4;
+  // The reserves of asset 1 of the position before execution.
+  num.v1.Amount prev_reserves_1 = 5;
+  // The reserves of asset 2 of the position before execution.
+  num.v1.Amount prev_reserves_2 = 6;
+  // Context: the end-to-end route that was being traversed during execution.
+  DirectedTradingPair context = 7;
```
hdevalence, line 659: I made this a single-field `DirectedTradingPair` rather than two fields as in discord with the thought that it would make it easier to make queries like "show me all the position executions that occurred in the context of the swap X=>Y at height ABC"

crates/core/component/dex/src/component/position_manager.rs
```diff
@@ -188,38 +195,81 @@ pub trait PositionManager: StateWrite + PositionRead {
...
+        &mut self,
+        mut new_state: Position,
+        context: DirectedTradingPair,
+    ) -> Result<()> {
+        let prev_state = self
+            .position_by_id(&new_state.id())
+            .await?
+            .ok_or_else(|| anyhow::anyhow!("withdrew from unknown position {}", new_state.id()))?;
+
+        if !matches!(&prev_state.state, position::State::Opened) {
```
erwanor, line 235: I like to use `anyhow::ensure!(invariant, msg)` for guards

crates/core/component/dex/src/component/position_manager.rs
erwanor: That match statement was awkward AND !buggy!, good riddance

crates/core/component/dex/src/component/position_manager.rs
```diff
@@ -311,30 +361,31 @@ pub(crate) trait Inner: StateWrite {
...
         // Only index the position's liquidity if it is active.
-        if position.state == position::State::Opened {
-            self.index_position_by_price(&position);
+        if new_state.state == position::State::Opened {
+            self.index_position_by_price(&new_state);
         }
 
         // Update the available liquidity for this position's trading pair.
-        self.update_available_liquidity(&position, &prev).await?;
+        self.update_available_liquidity(&new_state, &prev_state)
```
erwanor, line 384: This PR actually improves on the previous impl because prev_state is `Option<Position>` but noting that the parent method signature is:
```rust
async fn update_position(
         &mut self,
         prev_state: Option<Position>,
         new_state: Position,
    ) -> Result<()>;
```
And this method is:

```rust
async fn update_available_liquidity(
        &mut self,
        position: &Position,
        prev_position: &Option<Position>,
    ) -> Result<()>;
```

but since the types are different this is probably fine

erwanor (approved):
ACK / I left some smaller suggestions take it or leave it. I'm convinced this is correct and not consensus breaking.

crates/core/component/dex/src/component/position_manager.rs
hdevalence: Good idea, updated

crates/core/component/dex/src/component/position_manager.rs
hdevalence: I'd rather streamline that code in a later PR that implements eviction, but good spotting. Added a TODO.

PR #4122, opened 2024-03-27
cnidarium: implement deferred commits via `StagedWriteBatch`

### LINKED ISSUE #4095: cnidarium: implement deferred commit
The ABCI++ release in `v0.38` introduces two large changes to its interface: support for vote extensions (VEs) and an important refactor of block execution into a unique [`FinalizeBlock`](https://docs.cometbft.com/v0.38/spec/abci/abci++_methods#finalizeblock) message. This method requires applications to produce a state root capturing the effect of block execution, but without actually committing to that root just yet.

Our current `Storage` implementation judiciously defers all hashing/merkleization to the last possible moment, when a `StateDelta` is [committed](https://rustdoc.penumbra.zone/main/cnidarium/struct.Storage.html#method.commit) to the backing store and returns a `cnidarium::RootHash`. But it does not yet support decoupling between "observing the effects" (via a state root) and "persisting to a backing store" that the ABCI++ interface requires. 

Under the hood, our storage layer weaves writes to substores/main store into a single atomic transaction (via a `rocksdb::WriteBatch`). This is done by passing around the batch and performing writes against it; eventually, the batch is on disk.

The API is extremely simple and consists of a single method:

```rust
    /// Commits the provided [`StateDelta`] to persistent storage as the latest
    /// version of the chain state.
    pub async fn commit(&self, delta: StateDelta<Snapshot>) -> Result<crate::RootHash>;
```

We propose to keep this API as it is and extend the `Storage` impl with two methods, which, when composed together, are strictly equivalent to a single call to `commit`:

```rust
/// Stage [`StateDelta`] changes into a DB transaction.
pub fn prepare_commit(delta: StateDelta) -> Result<StagedWriteBatch>;

/// Persist a [`StagedWriteBatch`] to storage.
/// # Errors
/// This method errors if the `batch` version is stale.
pub fn commit_staged(batch: StagedWriteBatch) -> Result<RootHash>;

pub fn commit(delta: StateDelta) -> Result<RootHash> {
    let batch = self.prepare_commit(delta);
    self.commit_staged(batch)
}
```

This would give downstream users of `cnidarium` the ability to manage staged writes to the DB, caching them during after `ProcessProposal` or `FinalizeBlock` without increasing the complexity for ordinary usage of the storage functionality, or design changes to the rest of the system (e.g., snapshot cache, or state change subscribers).

In terms of data modeling, we need to think through the `StagedWriteBatch` structure, but we can imagine something along these lines:

```rust
pub struct StagedWriteBatch {
    /// The write batch to commit to RocksDB.
    pub(crate) write_batch: rocksdb::WriteBatch,
    /// The new version of the chain state.
    pub(crate) version: jmt::Version,
    /// The new versions of each substore.
    pub(crate) multistore_versions: multistore::MultistoreCache,
    /// The root hash of the chain state corresponding to this set of changes.
    pub(crate) root_hash: crate::RootHash,
    /// A lightweight copy of the changeset, this is useful to provide
    /// a stream of changes to subscribers.
    pub(crate) changes: Arc<Cache>,
}
```

Managing the lifecycle of write batches is the responsibility of downstream users. They have to be careful about two things:
- `WriteBatch` is a handle to internal rocksdb buffers, there is memory overhead, and they need to be destroyed.
- `Storage::commit_staged` will reject stale write batches (version mismatch between main/substores in batch and storage).

noot:
refactor `commit` into `prepare_commit` and `commit_batch`. this allows a `StagedWriteBatch` to be prepared via `prepare_commit` and the actual writing the `StagedWriteBatch` to disk to occur in `commit_batch`. this allows us to get the root hash of the changes without requiring the data to be fully committed.

closes  #4095

### DIFF
--- a/crates/cnidarium/src/lib.rs
+++ b/crates/cnidarium/src/lib.rs
@@ -70,6 +70,7 @@ mod store;
 mod tests;
 mod utils;
 mod write;
+mod write_batch;
 
 #[cfg(feature = "metrics")]
 pub use crate::metrics::register_metrics;
@@ -81,6 +82,7 @@ pub use read::StateRead;
 pub use snapshot::Snapshot;
 pub use storage::{Storage, TempStorage};
 pub use write::StateWrite;
+pub use write_batch::StagedWriteBatch;
 
 pub mod future;
 
--- a/crates/cnidarium/src/snapshot.rs
+++ b/crates/cnidarium/src/snapshot.rs
@@ -124,14 +124,15 @@ impl Snapshot {
     }
 
     pub fn prefix_version(&self, prefix: &str) -> Result<Option<jmt::Version>> {
-        let config = self
+        let Some(config) = self
             .0
             .multistore_cache
             .config
-            .find_substore(prefix.as_bytes());
-        if prefix != config.prefix {
-            anyhow::bail!("requested substore (prefix={prefix}) does not exist")
-        }
+            .find_substore(prefix.as_bytes())
+        else {
+            anyhow::bail!("rquested a version for a prefix that does not exist (prefix={prefix})")
+        };
+
         Ok(self.substore_version(&config))
     }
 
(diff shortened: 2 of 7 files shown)


erwanor (changes requested):
Thanks for taking this on, this is great work and it unlocks a lot of cool things besides 0.38 support. There are a couple things we want to change before merging but overall ACK. I have put together a couple unit tests that I will add before merging, once we're done with review

noot:
@erwanor thanks for the review, comments addressed!

erwanor (approved):

erwanor:
Added some simple tests scenarios, will merge on green CI. This is really exciting! Thanks again for your work on this.

PR #4127, opened 2024-03-28
app: increment halt counter

erwanor:
## Describe your changes

This increments the `TOTAL_HALT_COUNTER` for `v0.71.0`

  >  This is an internal counter.

### DIFF
--- a/crates/core/app/src/lib.rs
+++ b/crates/core/app/src/lib.rs
@@ -19,7 +19,9 @@ pub use crate::{
 
 use once_cell::sync::Lazy;
 
-pub const APP_VERSION: u64 = 1;
+/// Representation of the Penumbra application version. Notably, this is distinct
+/// from the crate version(s). This number should only ever be incremented.
+pub const APP_VERSION: u64 = 2;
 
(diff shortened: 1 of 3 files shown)


crates/core/app/src/lib.rs
```diff
@@ -19,7 +19,7 @@ pub use crate::{
 
 use once_cell::sync::Lazy;
 
-pub const APP_VERSION: u64 = 1;
```
conorsch, line 22: nit: the `TOTAL_HALT_COUNT` var has a friendly docstring. Can we add one here? Especially since this value isn't actually used anywhere (yet). But describing when and why it should be incremented is super important.

conorsch: Done, and squashed in.

conorsch:
Refs:

* https://github.com/penumbra-zone/penumbra/pull/4107
* https://github.com/penumbra-zone/penumbra/pull/4108
* #4087

crates/core/app/src/lib.rs
erwanor: Thanks!!

PR #4129, opened 2024-03-28
docs: provide genesis-start for testnet 71

conorsch:
Refs #4087. Also lists an archive URL for `pd testnet join`.

  > docs-only change


conorsch:
Merging via override because these changes are docs-only, and I want them live for operators to use.

PR #4132, opened 2024-03-28
Update minifront/node-status apps

jessepinho:
## Describe your changes
Pulls in the build outputs of the node-status and minifront apps.

PR #4139, opened 2024-03-29
fix(pd): minifront asset directory subpath

jessepinho:
## Describe your changes
The previous .zip files had a subdirectory of the same name as the file itself. These updated files don't.

We should be creating zip files via a script — see https://github.com/penumbra-zone/web/issues/857. This will avoid issues like this in the future.

  > No code related to consensus


conorsch:
Thank you @jessepinho!

conorsch:
Refs https://github.com/penumbra-zone/penumbra/pull/4132#issuecomment-2027543305

conorsch:
Tacked on a test. LGTM!

conorsch (approved):

Issue #4081, opened 2024-03-22
Restructure Rust planner to be stateful, using simpler strategy used in wasm planner


hdevalence:
**Is your feature request related to a problem? Please describe.**

The planner implementation is very complicated, because of a desire to share code that didn't end up working out. See https://github.com/penumbra-zone/web/pull/663 for context.

**Describe the solution you'd like**

Apply the same strategy as in https://github.com/penumbra-zone/web/pull/663 to the Rust planner in the view service, eliminating all of the intermediate serializable one-pass requests-for-data and making it compute fees precisely.

Should follow #3940

PR #4141, opened 2024-04-01
refactor(planner): simplify transaction planner

TalDerei:
References https://github.com/penumbra-zone/penumbra/issues/4081 to simplify and fix transaction planner logic in the context of **non-zero** fees. Additionally expands scope to consume #4155. 

Blocked until fee restructuring https://github.com/penumbra-zone/penumbra/issues/3940 is merged. 

https://github.com/penumbra-zone/penumbra/issues/4154 and https://github.com/penumbra-zone/penumbra/issues/3435 will follow.


TalDerei:
@hdevalence the planner logic has been refactored to account for the presence of non-zero gas fees, following the gas schedules changes that added block space costs for every action in https://github.com/penumbra-zone/penumbra/issues/3940. There's a few things to note here. 

1. we introduced `prioritize_and_filter_spendable_notes`, enabling a custom note prioritization scheme based on **(1)** one-time addresses, or **(2)** largest value. In the case of **(2)**, the order of the notes being consumed seemed to matter in the original planner logic. For instance, for a specific user balance and non-zero gas fee, consuming smaller note  `N1` and larger note `N1` (where `N0 + N1 > Fee`) failed, whereas consuming larger note N1 and smaller note N0 (where `N0 + N1 > Fee`) succeeded. I modified the planner to be agnostic to the order of notes in order to balance the transaction (which is how it should be). Currently adding a few non-zero gas fee tests that check this **commutativity** property. 

2. We need to double check the correctness of the modified planner to be _absolutely_ sure that we're not double counting fees between iterations in the `while let` loop. For instance, in the case where if we can't sufficiently balance the transaction in the first iteration of the for loop, requiring another iteration to add additional spend notes. In that case, we're calling `adjust_change_for_fee` multiple times, subtracting fees from the change notes. The fee is a function of the actions / output change in the planner. If the first iteration has **output** (amount we want to spend), **spend** (note contributing to that spend), and **output** (change note)...and we iterate again to add another note, we now have **output, spend, spend, output**. Does that mean we're double counting the fees being subtracted from the change note? Further investigation is required to ensure to avoid making transactions unnecessarily more expensive than they need to be. 

3. currently delegate voting functionality isn't properly balanced, causing smoke test failures in pcli. Additionally, how should IBC actions be handled? Another set of eyes here would be really helpful.

4. After **1, 2, and 3** are accounted for, we can proceed with #4154 and check how non-zero gas fees fare in our smoke tests.

crates/view/src/planner.rs
```diff
@@ -79,12 +83,129 @@ impl<R: RngCore + CryptoRng> Planner<R> {
...
+        // so this gas estimate will be an underestimate, but since the tx-bytes contribution
+        // to the fee is ideally small, hopefully it doesn't matter.
+        let mut gas = Gas::zero();
+        for action in &self.actions {
+            // TODO missing AddAssign
+            gas = gas + action.gas_cost();
+        }
+        for action in self.change_outputs.values() {
+            // TODO missing AddAssign
+            // TODO missing GasCost impl on OutputPlan
```
zbuc, line 135: In the original implementation, none of the `*Plan` structs had `GasCost` impls, but that can probably change

zbuc:
>  3. the delegate voting functionality isn't being properly balanced, causing smoke test failures in pcli. Additionally, how should IBC actions be handled? Another set of eyes here would be really helpful.

Why are these presenting issues? Is there a reason they need to be handled differently from other actions?

TalDerei:
> **1.** we introduced `prioritize_and_filter_spendable_notes`, enabling a custom note prioritization scheme based on **(1)** one-time addresses, or **(2)** largest value. In the case of **(2)**, the order of the notes being consumed seemed to matter in the original planner logic.

the prioritization scheme was modified to consume largest notes, which makes this a moot point.

TalDerei:
> * **4.** brainstorm other edge cases

@zbuc after reviewing the modified planner logic, can you think of any other edge cases we may not be considering?

TalDerei:
> * **3.** the delegate voting functionality isn't being properly balanced, causing smoke test failures in pcli. Additionally, how should IBC actions be handled? Another set of eyes here would be really helpful.

@hdevalence can you check this out?

TalDerei:
> * **2.** we need to double check the correctness of the modified planner to be _absolutely_ sure we're not double counting fees.

I believe I'm currently accounting for this case as well. If for any iteration the note spent has a smaller notional value than the amount being spent, it will yield zero change outputs. This means fees won't be subtracted from the change notes until the final loop iteration when the equation balances.

crates/view/src/planner.rs
```diff
@@ -79,12 +83,129 @@ impl<R: RngCore + CryptoRng> Planner<R> {
...
+
+    fn calculate_balance_with_fees(&self, base_fee_estimation: Fee) -> Balance {
+        let mut balance = Balance::zero();
+        for action in &self.actions {
+            balance += action.balance();
+        }
+
+        for action in self.change_outputs.values() {
+            balance += action.balance();
+        }
```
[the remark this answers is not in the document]
TalDerei: I changed it to reduce code duplication

crates/view/src/planner.rs
TalDerei: adding the `GasCost` impls

TalDerei:
> Why are these presenting issues? Is there a reason they need to be handled differently from other actions?

I'll defer to @hdevalence for this. I added the same voting functionality we previously had, but it's not balancing properly. I haven't had a chance to inspect this too closely.

TalDerei:
planner logic does **_not_** account for the auction logic, and will need to be updated accordingly in a separate PR.

cratelyn:
removing my review request for now, while the smoke tests get sorted out, but ping me when this is ready for a look!

TalDerei:
a message to all reviewers: this has ballooned in scope into what now resembles frankensteins monster, falling out of parity with the web extension. In favor of maintaining that parity and reducing scope / making it easier to review, the port will be consumed by https://github.com/penumbra-zone/penumbra/pull/4300.

PR #4145, opened 2024-04-02
Enhance Nix packaging to build the crates and CometBFT

plaidfinch:
## Describe your changes

This PR enhances the Nix flake with additional derivations to build Penumbra's binaries and also CometBFT at the correctly compatible version.

Do `nix build` to build everything. Targets are: `.#cometbft`, `.#pd`, `.#pcli`, and `.#pclientd`.

Do `nix develop` to jump into a shell with everything set up (requires flake support to be turned on in your nix install).

  > This is just adding a Nix derivation for building Penumbra.


.envrc
```diff
@@ -0,0 +1 @@
+use flake
```
cratelyn, line 1: how nice! i should try out direnv, shouldn't i? :slightly_smiling_face:

cratelyn (commented):
this was such a pleasure to review. thank you for doing this, @plaidfinch. this is really lovely work that will likely pay some dividends for ease of deployment, and it's almost certainly going to make my day-to-day life better too.

i have some assorted notes below, but the point about pinning to a specific rust toolchain version is the one i am most interested in answering. the other notes are mostly requests for comments; flake files feel a little magical sometimes, and it'd be good to be particularly explicit about intent/purpose.

.envrc
conorsch: The `.envrc` file should instead be `.envrc.example`: otherwise, anyone with direnv installed will be forced to run `use flake` even if they don't use nix.

.gitignore
```diff
@@ -12,6 +12,12 @@ book
 # Used for deployments
 firebase-tmp/
 
+# Nix build output goes here
+result
+
+# Direnv files
+.direnv
+
```
conorsch, line 20: Matching comment above, add `.envrc` to the gitignore—but _not_ `.envrc.example`, which will be version-controlled.

conorsch (changes requested):
Nice! Thanks for taking the time. I'm not yet using a nix shell, but more comprehensive coverage like this makes it enticing.

cratelyn (approved):
this is tremendous work, thank you! :tada:

cratelyn:
```
   thread 'main' panicked at crates/crypto/proof-params/build.rs:37:18:
  called `Result::unwrap()` on an `Err` value: "error downloading proving key: can download proving key: can get bytes from file: error decoding response body: operation timed out: operation timed out"
```

this seems like it is likely to be spurious, but otherwise this all LGTM!

conorsch (approved):

PR #4151, opened 2024-04-02
feat(staking): 🎽 `IdentityKey` is internally represented as bytes

### LINKED ISSUE #2304: Change the `IdentityKey` domain type to store bytes rather than a decoded group element.
> One sub-case of this that could be worth doing sooner is the `IdentityKey`, which is currently a significant fraction of the total sync time.

_Originally posted by @hdevalence in https://github.com/penumbra-zone/penumbra/issues/1957#issuecomment-1497989146_

cratelyn:
work in progress.

fixes #2304.

### DIFF
--- a/crates/core/component/stake/src/identity_key.rs
+++ b/crates/core/component/stake/src/identity_key.rs
@@ -1,3 +1,7 @@
+// TODO(kate):
+// decompressed: std::sync::OnceLock<VerificationKey<SpendAuth>>,
+
+use decaf377_rdsa::{SigningKey, SpendAuth, VerificationKey, VerificationKeyBytes};
 use penumbra_proto::{
     core::component::stake::v1::CurrentValidatorRateRequest,
     // TODO: why is this not in the keys crate?
@@ -6,8 +10,7 @@ use penumbra_proto::{
     DomainType,
 };
 use serde::{Deserialize, Serialize};
-
-use decaf377_rdsa::{SpendAuth, VerificationKey};
+use std::sync::OnceLock;
 
 /// The root of a validator's identity.
 ///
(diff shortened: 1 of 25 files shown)


cratelyn:
**NB:** while we should eventually move the `IdentityKey` type to `penumbra-keys`, that change is avoided for now. thus, `feat(staking)`.

hdevalence:
> **NB:** while we should eventually move the `IdentityKey` type to `penumbra-keys`, that change is avoided for now. thus, `feat(staking)`.

We shouldn't do this, the identity key is staking-specific, which is why it's in the staking component. Also, since the proto lives in the staking proto package, this isn't possible without deprecating the proto type and leaving a duplicate around. So we can plan to not do it.

crates/core/component/stake/src/identity_key.rs
```diff
@@ -1,3 +1,7 @@
+// TODO(kate):
+// decompressed: std::sync::OnceLock<VerificationKey<SpendAuth>>,
```
hdevalence, line 2: ?

crates/core/component/stake/src/identity_key.rs
```diff
@@ -18,11 +21,83 @@ use decaf377_rdsa::{SpendAuth, VerificationKey};
...
-#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
+#[derive(Clone, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(try_from = "pb::IdentityKey", into = "pb::IdentityKey")]
-pub struct IdentityKey(pub VerificationKey<SpendAuth>);
+pub struct IdentityKey {
+    /// The *compressed* bytes of the identity verification key.
+    bytes: VerificationKeyBytes<SpendAuth>,
+    // The *decompressed* identity verification key.
+    key: OnceLock<DecompressionResult>,
+}
```
hdevalence, line 31: This should just be a wrapper around the `VerificationKeyBytes`:
```rust
pub struct IdentityKey(pub VerificationKeyBytes<SpendAuth>)
```
All we want to be doing here is replacing `VerificationKey` with `VerificationKeyBytes`, nothing more or less.

crates/core/component/stake/src/identity_key.rs
```diff
@@ -18,11 +21,83 @@ use decaf377_rdsa::{SpendAuth, VerificationKey};
...
+
+impl IdentityKey {
+    /// Decompresses this identity key, returning a [`VerificationKey`].
+    ///
+    /// This is idempotent and can be called repeatedly, and the bytes will only decompressed
+    /// once.
+    pub fn key(&self) -> Result<VerificationKey<SpendAuth>, decaf377_rdsa::Error> {
+        let Self { bytes, ref key } = *self;
+        key.get_or_init(|| VerificationKey::try_from(bytes)).clone()
+    }
```
hdevalence, line 89: We shouldn't be exposing this as part of the public API, it's not necessary for callers to have a convenient method for.

crates/core/component/stake/src/identity_key.rs
```diff
@@ -18,11 +21,83 @@ use decaf377_rdsa::{SpendAuth, VerificationKey};
...
+        Self { bytes, key }
+    }
+}
+
+impl<'vk> From<&'vk VerificationKey<SpendAuth>> for IdentityKey {
+    /// An [`IdentityKey`] can be constructed from a reference to a [`decaf377_rdsa::VerificationKey`].
+    fn from(vk: &'vk VerificationKey<SpendAuth>) -> Self {
+        Self::from(*vk)
+    }
+}
```
hdevalence, line 79: I'm not sure it's desirable to have all of these impls. We didn't have them before, do we need them to change the internal representation?

hdevalence (changes requested):
This PR adds a lot of code changes beyond the scope of the issue, which is changing `VerificationKey` to `VerificationKeyBytes`.

erwanor:
> We shouldn't do this, the identity key is staking-specific, which is why it's in the staking component. Also, since the proto lives in the staking proto package, this isn't possible without deprecating the proto type and leaving a duplicate around. So we can plan to not do it.

@hdevalence The `IdentityKey` proto definition already lives in `core.keys` so the current situation is a bit awkward fwiw

hdevalence:
Ah OK, in that case we should align it with the protos.

cratelyn:
this was an exploratory draft, and was not intended or ready to be reviewed. i'm going to close this in favor of #4152.

cratelyn:
> **NB:** while we should eventually move the `IdentityKey` type to `penumbra-keys`, that change is avoided for now. thus, `feat(staking)`.

it seems we came to agreement on this point. i'll file an issue about it sometime this week.

PR #4156, opened 2024-04-03
ci: use nightly rust for mdbook installation

conorsch:
## Describe your changes

Docs builds broke after merge of [0], not because its settings were incompatible, but because its inclusion of a `rust-toolchain.toml` file busted the CI cache, and caused the `mdbook` dep to be freshly installed, rather than retrieved from cache. We don't version-lock mdbook and friends, and the latest version evidently bumped MSRV to 1.74, which is newer than what we set for MSRV. We should probably bump MSRV too, but I'll do that separately. This commit is to unbreak the docs builds.

[0] https://github.com/penumbra-zone/penumbra/pull/4145

  > CI-only, docs-building logic.

### DIFF
--- a/.github/workflows/notes.yml
+++ b/.github/workflows/notes.yml
@@ -30,7 +30,8 @@ jobs:
       # Ostensibly building from source, but the cache-loading above
       # ensures we don't need to rebuild frequently.
       - name: Install mdbook dependencies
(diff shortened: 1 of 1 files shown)


conorsch:
Ad-hoc CI run on this branch succeeded https://github.com/penumbra-zone/penumbra/actions/runs/8546682283/job/23417520649 OK to merge after CI passes.

PR #4160, opened 2024-04-04
feat(utils): implement v038 for tower-trace

noot:
## Describe your changes

implement support for ABCI v038 for the tower-trace util crate. required for eventual upgrade to ABCI v038 :)

n/a 

it only affects the tower-trace util crate


erwanor (approved):
LGTM, every consensus request variant is covered and this is using the right `v0_38` messages. I will merge on green CI.

PR #4164, opened 2024-04-05
proto: add extra metadata fields to `SwapView.Opaque`

hdevalence:
Many of the "opaque" variants of our view messages don't have any extra data other than the raw action. This turns out to be insufficient, because often there's additional data that's practically necessary to understand the public parts of the action.

For instance, in this case, the `SwapView`, we want to correctly and nicely render the public parts of the swap, which include:

- the input amounts, which are transparent;
- the pro-rata output amounts, which are computable given public information.

However, this information can't be easily rendered just from the underlying `Swap`, because:

- it doesn't contain the asset `Metadata` about the values, only the amount and asset IDs;
- it doesn't contain the `BatchSwapOutputData` with the (public) results of the batch swap;
- it doesn't have precomputed output values, so the rendering would have to understand and replicate the pro-rata computations.

Instead, we should make use of the perspective/view mechanism to augment the `SwapView.Opaque` with that data.

To complete this PR, we'll need to:

- [x] Regenerate protos and fix any compile errors in the Rust code.
- [x] Ensure that the tx+perspective => view transformation uses data from the perspective to fill in the new fields (this should be similar to the existing code, but for opaque swaps rather than just visible ones).

The web frontend's code for displaying simulated "public views" of the user's transactions can probably make use of this functionality after the first step (since it constructs the simulated view by attenuating the "full view" rather than by regenerating with a fresh perspective), but we should fill in the perspective generation for completeness and to ensure that the data we're adding actually makes sense and is possible to generate from the perspective.

  > This is only a change to the client-side view code.

### DIFF
--- a/crates/core/transaction/src/is_action.rs
+++ b/crates/core/transaction/src/is_action.rs
@@ -367,9 +367,65 @@ impl IsAction for Swap {
                         .cloned(),
                 }
             }
-            None => SwapView::Opaque {
-                swap: self.to_owned(),
-            },
+            None => {
+                // If we can find a matching BSOD in the TxP, we can use it to compute the output notes
+                // for the swap.
+                let bsod = txp
+                    .batch_swap_output_data
+                    .iter()
+                    // This finds the first matching one; there should only be one
+                    // per trading pair per block and we trust the TxP provider not to lie about it.
+                    .find(|bsod| bsod.trading_pair == self.body.trading_pair);
+
+                // We can get the denom metadata whether we get a BSOD or not
+
+                let denom_1 = txp.denoms.get(&self.body.trading_pair.asset_1()).cloned();
+                let denom_2 = txp.denoms.get(&self.body.trading_pair.asset_2()).cloned();
+
+                match bsod {
+                    None => {
+                        // If we can't find a matching BSOD, we can't compute the output notes
+                        // for the swap.
+                        // TODO: is defaulting to 'None' the right behavior here, or is not matching on a BSOD as above an error?
+                        SwapView::Opaque {
+                            swap: self.to_owned(),
+                            batch_swap_output_data: None,
+                            output_1: None,
+                            output_2: None,
+                            asset_1_metadata: denom_1.clone(),
+                            asset_2_metadata: denom_2.clone(),
+                        }
+                    }
+                    Some(bsod) => {
+                        // If we can find a matching BSOD, use it to compute the output notes
+                        // for the swap.
+
+                        let (lambda_1_i, lambda_2_i) =
+                            bsod.pro_rata_outputs((self.body.delta_1_i, self.body.delta_2_i));
+
+                        SwapView::Opaque {
+                            swap: self.to_owned(),
+                            batch_swap_output_data: Some(bsod.clone()),
+                            asset_1_metadata: denom_1.clone(),
+                            asset_2_metadata: denom_2.clone(),
+                            output_1: Some(
+                                Value {
+                                    amount: lambda_1_i,
+                                    asset_id: self.body.trading_pair.asset_1(),
+                                }
+                                .view_with_cache(&txp.denoms),
+                            ),
+                            output_2: Some(
+                                Value {
+                                    amount: lambda_2_i,
+                                    asset_id: self.body.trading_pair.asset_2(),
+                                }
+                                .view_with_cache(&txp.denoms),
+                            ),
+                        }
+                    }
+                }
+            }
         })
     }
 }
--- a/crates/core/transaction/src/is_action.rs
+++ b/crates/core/transaction/src/is_action.rs
@@ -378,7 +378,6 @@ impl IsAction for Swap {
                     .find(|bsod| bsod.trading_pair == self.body.trading_pair);
 
                 // We can get the denom metadata whether we get a BSOD or not
-
                 let denom_1 = txp.denoms.get(&self.body.trading_pair.asset_1()).cloned();
                 let denom_2 = txp.denoms.get(&self.body.trading_pair.asset_2()).cloned();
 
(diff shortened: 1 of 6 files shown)


hdevalence:
cc @aubrika @jessepinho

aubrika:
Met with @/jessepinho to discuss this changeset - the work seems straightforward & we will revisit once the initial proto/perspective generation changes are made, but these proposed changes seem straightforward to consume on the web side, as expected.

crates/core/transaction/src/is_action.rs
```diff
@@ -367,9 +367,40 @@ impl IsAction for Swap {
...
-                swap: self.to_owned(),
-            },
+            None => {
+                // If we can find a matching BSOD in the TxP, use it to compute the output notes
+                // for the swap.
+                let bsod = txp
+                    .batch_swap_output_data
+                    .iter()
+                    // This finds the first matching one; there should only be one
+                    // per trading pair per block and we trust the TxP provider not to lie about it.
```
jessepinho, line 377: > we trust the TxP provider not to lie about it

That's not actually a security concern, right? i.e., if the TxP provider (such as a wallet extension) were to put inaccurate data into the TxP, that'd just mean that the transaction would render as a view incorrectly, right? Just making sure I understand the stakes.

crates/core/transaction/src/is_action.rs
```diff
@@ -367,9 +367,40 @@ impl IsAction for Swap {
...
+                let denom_2 = txp.denoms.get(&self.body.trading_pair.asset_2()).cloned();
+
+                SwapView::Opaque {
+                    swap: self.to_owned(),
+                    batch_swap_output_data: bsod.cloned(),
+                    asset_1_metadata: denom_1.clone(),
+                    asset_2_metadata: denom_2.clone(),
+                    output_1: Some(
+                        Value {
+                            amount: self.body.delta_1_i,
```
jessepinho, line 390: In the `SwapPlaintext` Protobuf type, `delta_1_i` and `delta_2_i` refer to the _inputs_ of the swap, not the outputs. So I _believe_ this is incorrect.

That said, I don't know how you would get outputs on an opaque swap. For a visible swap, you can call `swap_plaintext.output_notes(bsod)`, but `swap_plaintext` isn't available here. So I'm not sure how to calculate the outputs correctly 😬

jessepinho (commented):
One main concern about how outputs are calculated. I think someone else will have to advise on how to do that, though — or if it's even possible!

crates/core/transaction/src/is_action.rs
hdevalence: Yes. One use case of the TxP that we are vaguely aiming at but haven't really explored in detail yet is as a selective disclosure mechanism to third parties (rather than untrusted first-party agents).

In the current scenario, we're doing selective disclosure to a first-party agent (the frontend) who's not trusted to have key material.

In a future scenario, we'd like to use the same mechanism to do disclosure to third parties, who may not trust the disclosing party to be honest. Thus, instead of consuming a view directly, they'd receive a TxP from the user, and then use that to generate their own view. This is mostly already possible. In this context, they'd probably want to fill in the public "bonus data" themselves, or check it against the chain.

crates/core/transaction/src/is_action.rs
hdevalence: Yes, this seems incorrect. $\Delta_{1,i}$ and $\Delta_{2,i}$ are the inputs. But since the outputs are the pro-rata shares of the batch output, which is public, and the inputs are public, it's possible to compute the output amounts.

It's indeed not possible to compute the output _notes_, since they rely on secret information: a note is a value plus a control address plus a blinding factor, and the control address and blinding factors are private. That's why in the `SwapClaimView` the outputs are `NoteView`s, but here they're mere `ValueView`s. 

(Side note: is that actually true? It looks like this is constructing a `Value`, it should be a `ValueView`...)

The method for computing the output amounts (used internally by `output_notes`) is `BatchSwapOutputData::pro_rata_outputs`: https://rustdoc.penumbra.zone/main/penumbra_dex/struct.BatchSwapOutputData.html#method.pro_rata_outputs

crates/core/transaction/src/is_action.rs
```diff
@@ -367,9 +367,65 @@ impl IsAction for Swap {
...
+                // We can get the denom metadata whether we get a BSOD or not
+
+                let denom_1 = txp.denoms.get(&self.body.trading_pair.asset_1()).cloned();
+                let denom_2 = txp.denoms.get(&self.body.trading_pair.asset_2()).cloned();
+
+                match bsod {
+                    None => {
+                        // If we can't find a matching BSOD, we can't compute the output notes
+                        // for the swap.
+                        // TODO: is defaulting to 'None' the right behavior here, or is not matching on a BSOD as above an error?
```
cratelyn, line 389: this is a question i do not know the answer to!

hdevalence: Not finding a BSOD in the TxP means that whoever generated the transaction perspective didn't include that data, either because they didn't have it (eg the transaction was created but never confirmed by the chain and executed) or because they didn't support it in their implementation.

We should handle that by graceful degradation, just not rendering the BSOD-requiring data if we don't have it.

cratelyn: to prompt discussion: we end up on this branch if the swap ciphertext could not be decrypted with the perspective's payload key associated with this swap's commitment, _and_ no batch swap output data in the perspective matched this trading pair.

if so, does it make sense to alternatively return an error (_this function would return an `anyhow::Result<ActionView>`_), return nothing at all (_this function would return an `Option<ActionView>`_) or does filling `None` into these fields as-is make sense?

cratelyn: (pardon me, didn't see the new message when drafting this.)

cratelyn: pushed 5d3533f to remove this `TODO` comment.

cratelyn (approved):
this is great work @aubrika! this PR looks good to me.

PR #4166, opened 2024-04-05
ci: add per-crate checks via cargo-hack

conorsch:
# Describe your changes
Suggested by @erwanor, based on the Astria CI config [0]. The goal is to ensure that our default feature sets actually work, by compiling each crate in the workspace individually.

[0] https://github.com/astriaorg/astria/blob/6cc8e2b828f8f5ee65e03c2b3383c4252b4c6b81/.github/workflows/test.yml#L64-L68

  > The changes to crate content are explicitly declaring which imports should be activated based on feature sets. The rest of the diff is CI-only logic.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -62,3 +62,18 @@ jobs:
       - name: Load rust cache
         uses: astriaorg/buildjet-rust-cache@v2.5.1
       - run: cargo fmt --all -- --check
+
+  check:
+    name: Check all crate features
+    runs-on: buildjet-8vcpu-ubuntu-2204
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+      - name: Load rust cache
+        uses: astriaorg/buildjet-rust-cache@v2.5.1
+      - name: install cargo-hack
+        run: cargo install cargo-hack
+      # Build each crate separately, to validate that the feature-gating is working.
+      # This is a lighter-weight version of `cargo check-all-features --workspace --release`.
+      - run: ./deployments/scripts/check-crate-feature-sets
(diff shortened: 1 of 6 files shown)


crates/core/component/compact-block/src/component/manager.rs
```diff
@@ -1,6 +1,7 @@
 use anyhow::{Context, Result};
 use async_trait::async_trait;
 use cnidarium::StateWrite;
+#[cfg(feature = "component")]
```
erwanor, line 4: nice catch!

.github/workflows/rust.yml
```diff
@@ -62,3 +62,18 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+      - name: Load rust cache
+        uses: astriaorg/buildjet-rust-cache@v2.5.1
+      - name: install cargo-hack
+        run: cargo install cargo-hack
+      # Build each crate separately, to validate that the feature-gating is working.
+      # This is a lighter-weight version of `cargo check-all-features --workspace --release`.
```
cratelyn, line 78:
```suggestion
      # This is a lighter-weight version of `cargo check --all-features --workspace --release`.
```

nit: i believe we meant `cargo check` here?

cratelyn (approved):
:sparkles: lovely! spotted one comment nitpick, otherwise i am happy to see some extra assurance that individual crates build. :package:

.github/workflows/rust.yml
conorsch: Weirdly, no! See context in https://github.com/frewsxcv/cargo-all-features & https://github.com/penumbra-zone/penumbra/issues/2210

cratelyn: thank you for the background!

Issue #4077, opened 2024-03-22
Add liquidity-based position eviction to the DEX


hdevalence:
**Is your feature request related to a problem? Please describe.**

The DEX engine does not bucket individual positions into bins with quantized ticks. Each position is filled against separately and can be whatever price it wants. This poses a potential problem: what if someone created a large number of very small positions, and the DEX then had to fill against all of them in sequence, forever? 

Our answer to this has been to (1) make filling against a position very fast (2) to have nonzero gas fees required to create a position (3) to ensure that the DEX engine only considers the tip of the order book. For volatile pairs (where the price moves), this seems sufficient: the price of the position is fixed, but the price of the pair moves, so dust positions cannot remain at the tip of the order book without active management (causing fees on the part of the position creator). However, this doesn't work for stable pairs, where the price doesn't move. We've put off dealing with this issue for a while, but we should add a mitigation before mainnet.

**Describe the solution you'd like**

Add a chain parameter that sets a maximum number of positions per pair.  A position can remain open only if it is in the top `N` positions, sorted by amount of reserves `R_1` OR if it is in the top `N` positions, sorted by amount of reserves `R_2`.  (This means that in principle there could be up to `2*N` positions per pair).

This should be implemented in such a way that we don't do expensive iteration over all positions, if possible.

hdevalence:
Replaces https://github.com/penumbra-zone/penumbra/issues/2928

aubrika:
This needs design work - meeting for the team forthcoming

aubrika:
@erwanor can you share here the current status of work on this?

erwanor:
Yes, sorry well overdue update - this ticket was split into two smaller PRs that I completed before switching gear towards the auction effort:

- [x] #4167 
- [x] #4172 

I just opened a draft PR to complete the last bit, it should be ready for review tomorrow:
- [ ] #4511

erwanor:
This is done on the protocol front, I have an app test that I would like to include before closing the issue.

PR #4167, opened 2024-04-05
dex: track total number of positions per pair

erwanor:
## Describe your changes

Working towards achieving #4077.

This PR adds a `PositionCounter` extension trait which implements a `TradingPair` scoped position counter. I am not totally sold on a `u16` for the counter type, yet, but it's useful for testing anyway.

  > This is consensus-breaking because it introduces a new implicit validation rule: a hard-limit on the number of position opened for each trading pair (via counter overflow). It could also be consensus breaking if there was a bug in the DEX that allowed closed positions to be updated, but my assumption is that it's not the case.

### DIFF
--- a/crates/core/component/dex/src/component/action_handler/position/open.rs
+++ b/crates/core/component/dex/src/component/action_handler/position/open.rs
@@ -13,7 +13,6 @@ use crate::{
 impl ActionHandler for PositionOpen {
     type CheckStatelessContext = ();
     async fn check_stateless(&self, _context: ()) -> Result<()> {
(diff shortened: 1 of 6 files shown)


erwanor:
I am going to merge this because the full eviction mechanism relies on it, but feel free to review asynchronously (please). I have three other separate PRs:
- one that adds a simple inventory index (#4172),
- one that adds a protocol specification amendment,
- one that adds that ties everything together, implementing the end-to-end eviction mechanism.

cratelyn (commented):
:heavy_check_mark:

PR #4172, opened 2024-04-08
dex: add an eviction index based on position reserves

erwanor:
## Describe your changes

This PR implements an inventory index, together with #4167, this PR works toward #4077 which contain the full eviction mechanism along with the adequate protocol specification update.

The state key added:

- $\text{I}_{A \rightarrow B} \coloneqq \text{prefix} \Vert BE(R_A) \Vert \text{id}$
- $\text{I}_{B \rightarrow A} \coloneqq \text{prefix} \Vert BE(R_B) \Vert \text{id}$

each corresponding to an index of position `Id`s ordered by reserves (ascending).

I plan to immediately follow-up this PR with a proposal to refactor the inner `PositionManager` index implementations.

  > It adds a state key to nonverifiable storage, and absent the actual eviction mechanism, this isn't actually consensus breaking.

### DIFF
--- a/crates/core/component/dex/src/component/position_manager.rs
+++ b/crates/core/component/dex/src/component/position_manager.rs
@@ -442,13 +442,15 @@ pub(crate) trait Inner: StateWrite {
         let pair_ab = DirectedTradingPair::new(canonical_pair.asset_1(), canonical_pair.asset_2());
         let inventory_a = position
             .reserves_for(pair_ab.start)
-            .expect("infaillible (remove it?)");
+            .expect("the directed trading pair is correct");
         let key_ab = eviction_queue::inventory_index::key(&pair_ab, inventory_a, id).to_vec();
         self.nonverifiable_put_raw(key_ab, vec![]);
 
         // B -> A
         let pair_ba = pair_ab.flip();
-        let inventory_b = position.reserves_for(pair_ba.start).expect("infaillible");
+        let inventory_b = position
+            .reserves_for(pair_ba.start)
+            .expect("the directed trading pair is correct");
         let key_ba = eviction_queue::inventory_index::key(&pair_ba, inventory_b, id).to_vec();
         self.nonverifiable_put_raw(key_ba, vec![]);
     }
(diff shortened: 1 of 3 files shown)


erwanor:
I want to follow-up with a refactor to the structure of the indices, mainly to make the `PositionManager` inner trait easier to read. In particular, we want to be able to audit the position update methods and its facades with a simple glance.

In #4167, I noticed that the liquidity index can double-count position contributions if a position is closed multiple times (e.g. auto-closing, and queued for closure), so we should harden the implementation and maybe streamline it.

crates/core/component/dex/src/component/position_manager.rs
```diff
@@ -428,6 +432,53 @@ pub(crate) trait Inner: StateWrite {
...
+    fn index_position_by_inventory(&mut self, position: &position::Position, id: &position::Id) {
+        tracing::debug!("indexing position by inventory");
+        let canonical_pair = position.phi.pair;
+        // A position is bound to an unordered trading pair: A <> B.
+        // We want to index the position by inventory for each direction:
+        // A -> B
+        let pair_ab = DirectedTradingPair::new(canonical_pair.asset_1(), canonical_pair.asset_2());
+        let inventory_a = position
+            .reserves_for(pair_ab.start)
+            .expect("infaillible (remove it?)");
```
cratelyn, line 445: should we address these expect statements here? would the infallible `reserves_1` and `reserves_2` methods be useful for this?

cratelyn (commented):
i have one question about whether we might be able to sidestep some phantom fallibility in retrieving reserves for a position's related assets.

otherwise, this seems pretty reasonable to me!

crates/core/component/dex/src/component/position_manager.rs
erwanor: Good catch, using `reserves_for` here is more defensive because we are dealing with two levels of semantics wrt ordering:
- a `TradingPair` has a canonical ordering based on the random field elements that make up its two asset `Id`s
- a `DirectedTradingPair` is an ordered trading pair i.e. one that has both a canonical ordering and an application-level one


The `reserves_1` and `reserves_2` getters return the reserves relative to the canonical ordering, but it might not correspond to the one we want. We could write the method so that they do always coincide, I have considered it, but this breaks silently during a refactor - so more brittle than an extra branch.

cratelyn: ah! thanks for the explanation 🙂sounds good to me

Issue #3846, opened 2024-02-19
staking: 🎏 remove `validator_definitions` method and refactor `rpc::validator_info` to use `consensus_set_stream`


erwanor:
The staking component query service provides a `ValidatorInfo` RPC that returns the list of registered validators, optionally filtering `Inactive` validators.

The way this method is implemented is by using an internal `ValidatorDataRead::validator_definitions` methods which returns every single definition ever submitted to the chain. This is problematic because this list will inevitably grow quite large, turning usage of a benign-looking method into an onerous source of I/O and expensive rendering.

A better solution is to instead use `ConsensusIndexRead` and filter through validator definitions that are in the consensus set (i.e. validators that are in the active set, or eligible to be promoted into it).

On the other hand, this method is also used by validators to check that their submissions have been successfully recorded by the chain. To address this valid use case, we could create a `pcli query validator status` that `Defined` validator can use to inquire about their validator state. 

### :ballot_box_with_check: to do:
- [ ] Remove `ValidatorDataRead::validator_definitions` (the state accessor doing the prefix query, not the underlying store)
  * #4262
- [ ] Rework `penumbra_stake::component::rpc::validator_info` to use `ConsensusIndexRead ::consensus_set_stream`
  * #4262
- [ ] pcli(query): add `Status` to `ValidatorCmd`
  * #4263
  * https://github.com/penumbra-zone/penumbra/pull/4263#discussion_r1582437217

erwanor:
We must make this change transparent to consumers of the service (web extension).

hdevalence:
We should not aim to preserve API compatibility, we should instead just make it do the right thing, restricting to the list of validators above the minimum limit.

cratelyn:
following up on this, after it came up in #4173:

`validator_definitions` returning a vector of every definition is useful (_and sometimes required_) for test logic, and saves test cases the need to repeatedly deal with polling an asynchronous stream to get a `&[Validator]` or `&[IdentityKey]` to make assertions upon.

i don't have strong opinions about how we present this information via RPC, but i'd hope we can preserve the friendly ergonomics in tests!

erwanor:
We should assign this during sprint planning @cratelyn

cratelyn:
> The staking component query service provides a `ValidatorInfo` RPC that returns the list of registered validators, optionally filtering `Inactive` validators.
> 
[...]

PR #4173, opened 2024-04-08
tests(app): 😌 deduplicate staking test boilerplate

cratelyn:
this branch adds some small tweaks to cut down on common boilerplate code found
in many of the staking component tests. this allows us to (a) easily set the
epoch duration, (b) fast forward to the end of an epoch, and (c) obtain the
identity keys of the currently defined validators. 

#### f3ecf6d5806e9b99de9fa98e0e0b85076840bef0 tests(app): 🔶 add genesis content setters
    
this is some common glue code in integration tests for the staking component.
in order to smooth the inspection of component logic that runs at epoch
boundaries, the defaults for `AppState`'s inner settings for e.g. epoch
duration are overriden to shorter values.

as is, this requres stating object literals that are somewhat verbose, even
with `..T::default()` elision. this commit adds some small `with_` helpers to
allow test cases to manipulate these settings succinctly.

#### bf8f4d14253878a38bdd1589fb15d414f4745094 tests(app): 🏃 add fast forward to next epoch helper
    
this adds an extension method that can be used to fast forward to the next
epoch. this is a very common part of test logic that exercises the staking
component.

this allows test cases to invoke an extension method to jump ahead, without
needing to import the clock component and write a loop inspecting the current
epoch after creating and executing an empty block.

#### 1c5b90a23 tests(app): 🔑 use `validator_identity_keys` accessor

this adds a `ValidatorDataRead::validator_identity_keys()` trait method.

frequently, test logic needs to keep track of validators by their identity
keys. this led to a bit of common glue code that would get the latest snapshot,
the set of all validator definitions, assert that only one validator is
present, and then take the identity key of that single validator.

now we can just get the keys, bind that single entry to a variable, or return
an error if something went wrong.

---

- [x] If this code contains consensus-breaking changes, I have added the
  "consensus-breaking" label. Otherwise, I declare my belief that there are not

> this is only additive changes, or adjusting test code.

### DIFF
--- a/crates/core/app/tests/app_tracks_uptime_for_validators_only_once_active.rs
+++ b/crates/core/app/tests/app_tracks_uptime_for_validators_only_once_active.rs
@@ -4,7 +4,10 @@ use {
     self::common::BuilderExt,
     cnidarium::TempStorage,
     decaf377_rdsa::{SigningKey, SpendAuth, VerificationKey},
-    penumbra_app::{genesis::AppState, server::consensus::Consensus},
+    penumbra_app::{
+        genesis::{self, AppState},
+        server::consensus::Consensus,
+    },
     penumbra_keys::test_keys,
     penumbra_mock_client::MockClient,
     penumbra_mock_consensus::TestNode,
@@ -31,14 +34,8 @@ async fn app_tracks_uptime_for_validators_only_once_active() -> anyhow::Result<(
     let storage = TempStorage::new().await?;
 
     // Configure an AppState with slightly shorter epochs than usual.
-    let app_state = AppState::Content(penumbra_app::genesis::Content {
-        sct_content: penumbra_sct::genesis::Content {
-            sct_params: penumbra_sct::params::SctParameters {
-                epoch_duration: EPOCH_DURATION,
-            },
-        },
-        ..Default::default()
-    });
+    let app_state =
+        AppState::Content(genesis::Content::default().with_epoch_duration(EPOCH_DURATION));
 
     // Start the test node.
     let mut node = {
--- a/crates/core/app/tests/app_tracks_uptime_for_validators_only_once_active.rs
+++ b/crates/core/app/tests/app_tracks_uptime_for_validators_only_once_active.rs
@@ -1,7 +1,5 @@
-mod common;
-
 use {
-    self::common::BuilderExt,
+    self::common::{BuilderExt, TestNodeExt},
     cnidarium::TempStorage,
     decaf377_rdsa::{SigningKey, SpendAuth, VerificationKey},
     penumbra_app::{
@@ -21,6 +19,8 @@ use {
     tap::Tap,
     tracing::{error_span, info, Instrument},
 };
+
+mod common;
 
 #[tokio::test]
 async fn app_tracks_uptime_for_validators_only_once_active() -> anyhow::Result<()> {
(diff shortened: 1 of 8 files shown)


crates/core/app/tests/app_tracks_uptime_for_validators_only_once_active.rs
```diff
@@ -73,19 +70,12 @@ async fn app_tracks_uptime_for_validators_only_once_active() -> anyhow::Result<(
...
-            [Validator { identity_key, .. }] => *identity_key,
-            unexpected => panic!("there should be one validator, got: {unexpected:?}"),
-        }
-    };
+    let [existing_validator_id] = storage
+        .latest_snapshot()
+        .validator_identity_keys()
+        .await?
+        .try_into()
+        .map_err(|keys| anyhow::anyhow!("expected one key, got: {keys:?}"))?;
```
cratelyn, line 78: an example of the benefit of being able to request a `Vec<IdentityKey>` out of the temporary storage's latest snapshot.

crates/core/app/tests/app_tracks_uptime_for_validators_only_once_active.rs
```diff
@@ -237,19 +227,10 @@ async fn app_tracks_uptime_for_validators_only_once_active() -> anyhow::Result<(
...
-            .await?
-            .tap(|start| tracing::info!(?start, "fast forwarding to next epoch"));
-        let next = loop {
-            node.block().execute().await?;
-            let current = get_epoch().await?;
-            if current != start {
-                break current;
-            }
-        };
-        tracing::info!(?start, ?next, "finished fast forwarding to next epoch");
```
cratelyn, line 252: an example of the benefit of `fast_forward_to_next_epoch`.

crates/core/app/tests/common/mod.rs
```diff
@@ -89,3 +89,46 @@ impl TempStorageExt for TempStorage {
...
+    ) -> Result<penumbra_sct::epoch::Epoch, anyhow::Error> {
+        use {penumbra_sct::component::clock::EpochRead, tap::Tap};
+
+        let get_epoch = || async { storage.latest_snapshot().get_current_epoch().await };
+        let start = get_epoch()
+            .await?
+            .tap(|start| tracing::info!(?start, "fast forwarding to next epoch"));
+
+        loop {
+            self.block().execute().await?;
```
erwanor, line 126: This completely works, and I think the helper method makes sense. For later, I think we will want to have a way to fast-forward to the next epoch as dynamically selected by the application. This requires only an extra look-up and gives us the capability to enshrine staking component invariant in tests (e.g. an easy one is that `Tombstoning` MUST trigger an epoch transition)

crates/core/component/stake/src/component/validator_handler/validator_store.rs
```diff
@@ -235,6 +235,15 @@ pub trait ValidatorDataRead: StateRead {
             .try_collect()
             .await
     }
+
+    /// Returns a list of **all** known validators identity keys.
+    async fn validator_identity_keys(&self) -> Result<Vec<IdentityKey>> {
+        self.prefix(state_key::validators::definitions::prefix())
```
erwanor, line 241: x-ref #3846 so that it shows up on the issue

erwanor (approved):

crates/core/app/tests/common/mod.rs
cratelyn: > For later, I think we will want to have a way to fast-forward to the next epoch as dynamically selected by the application. 

x-ref #4149, where we can address this :slightly_smiling_face:

PR #4174, opened 2024-04-08
mock-consensus: 🫛 `two_validators` adds two keys

### LINKED ISSUE #3937: tests: ➁ mock consensus can `InitChain` with two validators
see:
* #3588
* #3902
* #3816
* #3934

in #3902 we introduced the ability to start a test node with a single set of validator keys. from conversation with @/erwanor, it would be helpful when writing staking tests to provide another interface, allowing us to start a test node with four validators.

cratelyn:
fixes #3937.

* #3937
* #3588

this adds a `two_validators` method to the test node builder, so that tests may set up a test node that has two validator keys.

  > only adjusts test facilities

### DIFF
--- a/crates/test/mock-consensus/src/builder.rs
+++ b/crates/test/mock-consensus/src/builder.rs
@@ -8,7 +8,6 @@ mod init_chain;
 use {
     crate::{Keyring, TestNode},
     bytes::Bytes,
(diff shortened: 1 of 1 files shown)


erwanor (commented):
LGTM

erwanor (approved):
Woops.

Issue #4170, opened 2024-04-07
Force final validator definitions to "disabled" to reduce incentive to steal validator signing key


plaidfinch:
Presently, if a validator signing key is compromised, the original operator can as a final action use their key to broadcast a definition with a sequence number of `u32::MAX` which disables the validator permanently, thereby preventing the attacker from continuing to profit from validator rewards they might steal using the compromised key. It is therefore in the interest of the attacker to beat the original operator to the punch by broadcasting a new definition with a sequence number of `u32::MAX`, a new consensus key which the attacker controls, a new funding streams which take 100% commission and send it to the attacker, and which keeps the validator enabled.

If the attacker can manage to do this before the operator can broadcast the permanent disabling message, then the attacker can steal all validator rewards that continue to accumulate during the time period when the validator is still in the active set (until eventually enough people will undelegate due to the attack, at which point the validator will stop earning rewards).  The compromised original operator would be helpless to stop this except by publicizing the attack, because the sequence number of `u32::MAX` would prevent them from further updating the on-chain validator definition to disable it.

It would reduce the payoff for an attacker if the original validator operator could always have the last say, even if compromised, so that the attacker would profit only minimally from stolen validator rewards before the operator responded by disabling the validator forever. If this were the case, attackers would have less incentive to attempt validator signing key compromise in the first place.

This can be achieved very simply: add a rule that if the sequence number of a validator definition is `u32::MAX`, then the definition must set the validator to the disabled state.

If this rule was added to consensus, an attacker would be forced to either leave the validator active but allow the original operator the ability to shut it down forever, or to shut it down forever themself; it would be impossible for any party to lock a validator in an enabled state permanently.

hdevalence:
Seems reasonable, want to make a PR?

PR #4178, opened 2024-04-08
Add consensus check preventing end-of-life enabled validators

plaidfinch:
## Describe your changes

This PR adds a **new consensus rule** which forces validators with `sequence_number = u32::MAX` to be in the state `enabled = false`. This reduces the incentive to steal validator identity signing keys, because it means that it's impossible to "lock" a validator in the `enabled` state; the operator can always permanently disable their validator, cutting off profits for the attacker.

This is an implementation of issue #4170; see its description for more details about the risk this mitigates.

  > **This is a consensus-breaking change**: it rejects validator definitions which would previously have been accepted by the chain.


erwanor (approved):
Good idea for the test, I'll do it later

Issue #3856, opened 2024-02-20
Depending on Penumbra requires building with tokio-unstable


plaidfinch:
Experience report depending on penumbra libraries (`-proto` and `-governance`) externally to the mono-repo: in order to avoid build errors related to use of `tokio::task::Builder` in `cnidarium` (and possibly elsewhere?) I have to create a `.cargo/config.toml` with:

```toml
[build]
rustflags = ["--cfg", "tokio_unstable"]
```

Ideally I wouldn't have to do this, and the use of the task builder interface would be feature-gated so that I just lose out on the ability to get good task introspection in tokio-console if I don't do this, rather than getting a build-time failure in a dependency.

conorsch:
> Ideally I wouldn't have to do this, and the use of the task builder interface would be feature-gated

Word. Unfortunately, that's not an option for us, because the [builder interface uses unstable APIs](https://docs.rs/tokio/1.37.0/tokio/task/struct.Builder.html#), and the use of [unstable APIs in tokio mandates use of rustflags](https://docs.rs/tokio/1.37.0/tokio/index.html#unstable-features):

> To enable these features, the --cfg tokio_unstable argument must be passed to rustc when compiling. This serves to explicitly opt-in to features which may break semver conventions, since Cargo does not yet directly support such opt-ins

All this isn't worth the trouble. What we get for using the Builder interface is nicely "named" tasks, which would allow the tokio console output to be a lot more meaningful during debugging. Given the downstream friction this creates, it's not worthwhile, so let's just move away from the Builder interface entirely, reconsidering it if and when it stabilizes. 

There's a bit of [context in discord](https://discord.com/channels/824484045370818580/883455571003576320/1225915353671008376).

conorsch:
@erwanor has pointed out OOB that it _is_ technically possible that we could feature-gate this functionality, but we all agree we don't want to perform those gymnastics, and we're happier just ripping it out.

conorsch:
Done in https://github.com/penumbra-zone/penumbra/pull/4169

PR #4179, opened 2024-04-08
penumbra: completely remove `console-subscriber`

erwanor:
## Describe your changes

This completely removes `console-subscriber` and the associated flag, on second thought it is worse to leave it around if it requires knowing about a secret compile flag to actually exercise. This also fix the summoner smoke test... 

  > Internal change.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1484,43 +1484,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "console-api"
-version = "0.6.0"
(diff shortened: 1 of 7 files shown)


conorsch:
Follow-up to https://github.com/penumbra-zone/penumbra/pull/4176. We haven't been using this functionality, so it's definitely fine to remove. Better still, yanking it out should unblock progress on https://github.com/penumbra-zone/penumbra/issues/3856

conorsch (approved):

PR #4182, opened 2024-04-09
tests(app): 🙅 app rejects invalid auth signatures

### LINKED ISSUE #4050: tests: 🙅 mock consensus test that validator definition with invalid auth sig are rejected
a validator definition tx that has an invalid authentication signature should be rejected, and not allowed to enter the `Defined` state. add test coverage for this.

cratelyn:
fixes #4050. see also #3588.

this introduces a test case that demonstrates that transactions' validator definition actions must have a valid authentication signature, or the validator will not be added to the set of known validators.

  > this is a test case.


erwanor (approved):

erwanor:
This is a good test to have and a pretty slick example of using mock consensus to write app-level tests.

PR #4188, opened 2024-04-11
dex: refactor internal indexing into ext traits

erwanor:
## Describe your changes

This PR:
- break out indexing methods from the `Inner` position manager trait into submodules with crate visibility
- refactor `update_position` to delegate indexing checks to `update_*_index` methods
- add a redundant guard against invalid transitions in `PositionManager::update_position`
- streamline the base liquidity index, fixing a bug with double counting closed positions

This PR contain changes to critical parts of the DEX engine internals.

  > This is technically consensus breaking because we fix a bug in the base liquidity index logic, which could trickle down into different DEX execution in some rare cases.

### DIFF
--- a/crates/core/component/dex/src/component/mod.rs
+++ b/crates/core/component/dex/src/component/mod.rs
@@ -19,8 +19,6 @@ pub(crate) use arb::Arbitrage;
 pub use circuit_breaker::ExecutionCircuitBreaker;
 pub(crate) use circuit_breaker::ValueCircuitBreaker;
 pub use dex::{Dex, StateReadExt, StateWriteExt};
-// TODO(erwan): exposing a DEX interface to other components
-// is useful but maybe we should restrict it to open/queue/close positions
 pub use position_manager::{PositionManager, PositionRead};
 pub use swap_manager::SwapManager;
 #[cfg(test)]
--- a/crates/core/component/dex/src/component/mod.rs
+++ b/crates/core/component/dex/src/component/mod.rs
@@ -11,16 +11,16 @@ mod arb;
 pub(crate) mod circuit_breaker;
 mod dex;
 mod flow;
-pub(crate) mod position_counter;
-pub(crate) mod position_manager;
+mod position_manager;
 mod swap_manager;
 
 pub use self::metrics::register_metrics;
-pub use arb::Arbitrage;
+pub(crate) use arb::Arbitrage;
 pub use circuit_breaker::ExecutionCircuitBreaker;
 pub(crate) use circuit_breaker::ValueCircuitBreaker;
 pub use dex::{Dex, StateReadExt, StateWriteExt};
-pub use position_manager::{PositionManager, PositionRead};
+pub(crate) use position_manager::PositionManager;
+pub use position_manager::PositionRead;
 pub use swap_manager::SwapManager;
 #[cfg(test)]
 pub(crate) mod tests;
--- a/crates/core/component/dex/src/component/mod.rs
+++ b/crates/core/component/dex/src/component/mod.rs
@@ -19,8 +19,9 @@ pub(crate) use arb::Arbitrage;
 pub use circuit_breaker::ExecutionCircuitBreaker;
 pub(crate) use circuit_breaker::ValueCircuitBreaker;
 pub use dex::{Dex, StateReadExt, StateWriteExt};
-pub(crate) use position_manager::PositionManager;
-pub use position_manager::PositionRead;
+// TODO(erwan): exposing a DEX interface to other components
+// is useful but maybe we should restrict it to open/queue/close positions
+pub use position_manager::{PositionManager, PositionRead};
 pub use swap_manager::SwapManager;
 #[cfg(test)]
 pub(crate) mod tests;
--- a/crates/core/component/dex/src/component/position_manager/base_liquidity_index.rs
+++ b/crates/core/component/dex/src/component/position_manager/base_liquidity_index.rs
@@ -1,0 +1,174 @@
+use anyhow::Result;
+use cnidarium::StateWrite;
+use penumbra_num::Amount;
+
+use crate::lp::position::{Position, State};
+use crate::state_key::engine;
+use crate::DirectedTradingPair;
+use penumbra_proto::DomainType;
+
+pub(crate) trait AssetByLiquidityIndex: StateWrite {
+    /// Updates the nonverifiable liquidity indices given a [`Position`] in the direction specified by the [`DirectedTradingPair`].
+    /// An [`Option<Position>`] may be specified to allow for the case where a position is being updated.
+    async fn update_liquidity_index(
+        &mut self,
+        pair: DirectedTradingPair,
+        position: &Position,
+        prev: &Option<Position>,
+    ) -> Result<()> {
+        tracing::debug!(?pair, "updating available liquidity indices");
+
+        let (new_a_from_b, current_a_from_b) = match (position.state, prev) {
+            (State::Opened, None) => {
+                // Add the new position's contribution to the index, no cancellation of the previous version necessary.
+
+                // Query the current available liquidity for this trading pair, or zero if the trading pair
+                // has no current liquidity.
+                let current_a_from_b = self
+                    .nonverifiable_get_raw(&engine::routable_assets::a_from_b(&pair))
+                    .await?
+                    .map(|bytes| {
+                        Amount::from_be_bytes(
+                            bytes
+                                .try_into()
+                                .expect("liquidity index amount can always be parsed"),
+                        )
+                    })
+                    .unwrap_or_default();
+
+                // Use the new reserves to compute `new_position_contribution`,
+                // the amount of asset A contributed by the position (i.e. the reserves of asset A).
+                let new_position_contribution = position
+                    .reserves_for(pair.start)
+                    .expect("specified position should match provided trading pair");
+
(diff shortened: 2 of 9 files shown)


erwanor:
This is the general idea, probably needs workshopping but i think it's directionally what we want: a safe public DEX api with a small footprint, a private inner implementation that mediate state updates (with a second line of defense against invalid transitions), and a trust boundary behind which we have a variety of internal indices that only filter for relevant state transitions:

```rust
///                                                                         
///    ┌─────────────────────────────────────────┐   public DEX component api    
///    │             PositionManager             │   that are wrappers around   
///    │ public interface to the DEX component   │   `update_position` and manage
///    │                   │                     │   the high-level side effects
///    └───────────────────┼─────────────────────┘   like ProtoEvent records   
///                        │                                                
///                        ▼                                                
///  ┌─────────────────────────────────────────────┐                        
///  │   private Inner impl mediate state update   │  update_position with  
///  │                     │                       │    redundant checks    
///  └─────────────────────┼───────────────────────┘                        
///  ■━━━━━━━━━━━━━━━━━━━━━╋━━━━━━━━━━━━━━━━━━━━━━━━━━■    trust          
///                        ▼                              boundary          
///     ┌────────────────────────────────────┐                              
///     │pub(super) trait Index {            │                              
///     │      fn update_index(prev, new, id)│    update_index filters             
///     │                  │                 │    for relevant state transitions      
///     └──────────────────┼─────────────────┘    and deindex/index accordingly           
///                        ▼                                 
///               ┌────────────────┐                         
///               │ internal Inner │  implementation details of the              
///               └────────────────┘  indexing/deindexing logic         

```

crates/core/component/dex/src/component/mod.rs
```diff
@@ -11,15 +11,16 @@ mod arb;
...
 mod swap_manager;
 
 pub use self::metrics::register_metrics;
-pub use arb::Arbitrage;
+pub(crate) use arb::Arbitrage;
 pub use circuit_breaker::ExecutionCircuitBreaker;
 pub(crate) use circuit_breaker::ValueCircuitBreaker;
 pub use dex::{Dex, StateReadExt, StateWriteExt};
+// TODO(erwan): exposing a DEX interface to other components
+// is useful but maybe we should restrict it to open/queue/close positions
```
hdevalence, line 23: Isn't that what the `PositionManager` is?

crates/core/component/dex/src/component/position_manager/base_liquidity_index.rs
```diff
@@ -0,0 +1,167 @@
...
+use crate::state_key::engine;
+use crate::DirectedTradingPair;
+use penumbra_proto::{StateReadProto, StateWriteProto};
+
+pub(crate) trait AssetByLiquidityIndex: StateWrite {
+    /// Update the base liquidity index, used by the DEX engine during path search.
+    ///
+    /// # Overview
+    /// Given a directed trading pair `A -> B`, the index tracks the amount of
+    /// liquidity available to convert the quote asset B, into a base asset A.
```
hdevalence, line 16: @erwanor @zbuc maybe this comment could be clarified / expanded in light of today's discussion in the dex channel on discord, explaining that although the index may seem "backwards", we think it's a heuristic that's both useful and also hard to forge?

crates/core/component/dex/src/component/mod.rs
erwanor: yes, i was think we might want to restrict it further for example wrt closing positions, and force consumers to use the queue

crates/core/component/dex/src/component/position_manager/base_liquidity_index.rs
```diff
@@ -0,0 +1,167 @@
...
+    ) -> Result<()> {
+        match prev_state {
+            Some(prev_state) => match (prev_state.state, new_state.state) {
+                // We only want to update the index when we process active positions.
+                (Opened, Closed) => {}
+                (Opened, Opened) => {}
+                _ => return Ok(()),
+            },
+            None => {}
+        }
```
hdevalence, line 72: Not sure this logic is correct. We do want to update the index when a new position is added, no? Can we refactor this into a match statement like
```rust
match (prev_state.map(|prev| prev.state), new_state.state) {
    ...
}
```
or (if it's cleaner) eliminate that match statement entirely in favor of something where we compute two values:

- a previous contribution to the total, to be subtracted (this is 0 if the position is newly opened)
- a current contribution to the total, to be added (this is 0 if the position is newly closed)

if both are 0 it can short circuit without state changes

crates/core/component/dex/src/component/position_manager/counter.rs
```diff
@@ -24,13 +25,45 @@ pub(crate) trait PositionCounter: StateWrite {
             return 0;
         };
 
-        // This is safe because we only increment the counter via a [`Self::increase_position_counter`].
+        // This is safe because we only increment the counter via [`Self::increase_position_counter`].
```
hdevalence, line 28: would the extra two bytes for a u32 really kill us here? we spend more than that on a dex prefix. what if we want to have more than 100k positions? or what if the limit is something close, like 60k, and a bunch of new positions are opened within the block, and then there's an overflow (since we allow the position count to float above the limit during the block before closing positions to restore it)

crates/core/component/dex/src/component/position_manager/counter.rs
```diff
@@ -24,13 +25,45 @@ pub(crate) trait PositionCounter: StateWrite {
...
+                (Opened, Closed) => {
+                    let _ = self.decrement_position_counter(&trading_pair).await?;
+                }
+                _ => {}
+            },
+
+            None => {
+                let _ = self.increment_position_counter(&trading_pair).await?;
+            }
+        }
```
hdevalence, line 60: IMO this match statement would be clearer if there was a bit of prep work with a map clause, like
```rust
match (prev_state.map(|p| p.state), new_state.state) {
    // Increment the counter whenever a new position is opened
    (None, Opened) => self.increment_position_counter(&trading_pair).await?,
    // Decrement the counter whenever an opened position is closed
    (Some(Opened), Closed) => self.decrement_position_counter(&trading_pair).await?,
    // Other state transitions don't affect the opened position counter
    _ => {}
}
```

crates/core/component/dex/src/component/position_manager/inventory_index.rs
```diff
@@ -0,0 +1,84 @@
...
+use crate::{
+    lp::position::{self, Position},
+    state_key::eviction_queue,
+    DirectedTradingPair,
+};
+
+use anyhow::Result;
+use position::State::*;
+
+pub(super) trait PositionByInventoryIndex: StateWrite {
```
hdevalence, line 12: Can we add docs to this index describing where it's used? This is the index used for the position eviction logic, but there's another index of _assets_ by inventory that's used for routing heuristics, and we should disambiguate between those.

crates/core/component/dex/src/state_key.rs
```diff
@@ -103,52 +103,61 @@ pub(crate) mod engine {
...
         use penumbra_num::Amount;
 
         use super::*;
 
-        /// `A || be_bytes(A_from_B) => B` this will be an ordered encoding of every asset `B` directly routable to from `A`.
-        /// `a_from_b` represents the amount of `A` that can be bought with `B`.
-        /// The prefix query includes only the `A` portion, meaning the keys will be returned in order of liquidity.
-        pub(crate) fn prefix(from: &asset::Id) -> [u8; 39] {
+        /// A prefix key that takes a start asset `A` (aka. base asset) and surface adjacent
+        /// assets `B` (aka. quote asset), in ascending order of the base liquidity available.
```
hdevalence, line 113: I think this description is actually a little less clear than the preceding comments

crates/core/component/dex/src/component/mod.rs
hdevalence: Ah, well, the auction component as designed doesn’t allow that. So i would guess we probably wouldn’t

crates/core/component/dex/src/component/position_manager/counter.rs
erwanor: Good point, there wasn't a great reason to set it to two bytes

crates/core/component/dex/src/component/position_manager/inventory_index.rs
erwanor: Agreed, though I'd prefer to document this with the eviction queue PR

crates/core/component/dex/src/component/position_manager/base_liquidity_index.rs
erwanor: I reworked the structure based on your feedback, now we compute contributions upfront:
- compute `(prev_a, prev_b)` and clamps each value to zero if the position is closed/withdrawn
- compute `(new_a, new_b)` and clamp each value to zero if the position is closed/withdrawn
- short-circuit if all values are zero (e.g. `Closed -> Closed` or other "inactive" transitions)

crates/core/component/dex/src/component/position_manager/base_liquidity_index.rs
erwanor: I think we could do this in #4193

erwanor: I think the entire index description could be workshopped, we should put some examples

hdevalence (approved):

Issue #4189, opened 2024-04-11
dex: use big endian of complement in the base liquidity index 


erwanor:
**Is your feature request related to a problem? Please describe.**
@zbuc and I noticed that the base liquidity heuristic for path search surface the worst asset first. This is because the key is encoded as `prefix || base_asset || BE(base_liquidity)` which transfer the numerical ordering over into a lexicographic ordering. But that's not what we want! 

**Describe the solution you'd like**
Instead of doing we should compute `BE(!base_liquidity)` i.e. get the big-endian encoding of the **complement** so that the position with the most base liquidity get surfaced first.

**Describe alternatives you've considered**
It is tempting to extend `cnidarium` to allow prefix queries to specify a custom iteration order. But I think that fixing the state-key is the better solution because: 1/ a unique iteration order means you don't have to remember which one is the correct for the particular query you are making 2/ this is an easy migration and a useful exercise for us to do.

Random idea, a tiny schema library could provide out-of-the-box encoding wrappers e.g. `DescendingOrder(Amount) -> BE(!amount)`.

**Additional context**
We should have a couple simple unit-tests that assert this works.

hdevalence:
Along the way to this, it'd be great to have an isolated unit test that the ordering encoding is what we expect

PR #4193, opened 2024-04-11
Fix routable asset ordering

zbuc:
## Describe your changes

Fixes the ordering of routable assets by liquidity as well as adds tests.

Closes https://github.com/penumbra-zone/penumbra/issues/4189


zbuc:
@erwanor we should be good to clean this up and work on a migration in a pairing session when you've got time

erwanor (approved):
LGTM, we could add a couple lines to the `AssetByLiquidityIndex` trait about the intuition around manipulation-resistance and why we expect it to be directionally correct even if it doesn't factor the price

PR #4197, opened 2024-04-12
feat(ci): 🔖 add `needs-refinement` label to new issues

cratelyn:
the Penumbra team meets regularly to refine and discuss our backlog of tasks. this adds a workflow so that newly opened issues are accordingly labeled `needs-refinement`, so that they can be easily filtered in our refinement meeting.

this workflow is heavily inspired by the example workflow shown here: https://docs.github.com/en/actions/managing-issues-and-pull-requests/adding-labels-to-issues

  > only changes github workflows


conorsch (approved):
Love it! Let's try it!

PR #4198, opened 2024-04-12
chore: 🪹 remove empty `event.rs` files

cratelyn:
while looking around through files named `event.rs`, per the description of #3504, i found that some of these files are empty. this commit removes these empty files.

  > only removes empty files


conorsch (approved):

cratelyn:
https://discord.com/channels/824484045370818580/935270886091673660/1228349853390868536

we talked about this in discord, and came to the decision that we should keep these files around. while they are empty now, they should soon be updated to contain things, as part of addressing #3504. the existence of these is a signpost for where to place events in the future.

Issue #4196, opened 2024-04-12
Tracking issue: Dutch auction support for Penumbra


erwanor:
This issue is tracking completion of Dutch Auction for Penumbra:

**Design doc and context:** https://hackmd.io/@hdevalence/B1RjyYJgR

**Implementation plan:**
We break the implementation into the following milestones:

| Milestone | Deliverable |
| -------- | -------- |
| Component M0 | Functional action handlers + basic RPC |
| Component M1 | Value balance accounting /  Full RPC + Events |
| `pcli` M0 | We can record / execute / retire auctions in CLI    |
| `pcli` M1 | We can pass high-level parameters we want users to specify (time pref) and generate multiple DAs |
| Prax M0     | Update protos, react views for signing, view server updates |
| Minifront M0     | UI for gradual Dutch auctions     |
| Explorer M0    | Page displaying a DA     |
| Explorer M1    | Listing DAs / integrate into pair view |

with the following dependency relations:

```mermaid
graph TD
  CM0(Component M0) --> CM1
  CM1(Component M1) --> EM0
  EM0(Explorer M0) --> EM1
  EM1(Explorer M1)
  CM0 --> PM0
  PM0(pcli M0) --> PM1
  PM1(pcli M1)
  PM0 --> PrM0
  PrM0(Prax M0)
  PrM0 --> MM0
  MM0(Minifront M0)
  PM1 -.- MM0

  style CM0 fill:#f9d4e5,stroke:#333,stroke-width:2px
  style PM0 fill:#f9d5e5,stroke:#333,stroke-width:2px
  style PrM0 fill:#f6cd61,stroke:#333,stroke-width:2px
  style MM0 fill:#fe8a71,stroke:#333,stroke-width:2px
```

Tracking issue:

## Component
**Component M0: to build a protocol MVP for DAs**
- [x] auction: create component crate scaffolding: events.rs / state_key.rs
- [x] auction: define proto scaffolding
- [x] #4206
    - [x] `AuctionId`
    - [x] `AuctionNft`
    - [x] `DutchAuctionDescription`
    - [x] `DutchAuctionState`
    - [x] `DutchAuction`
    - [x] `ActionDutchAuctionSchedule`
    - [x] `ActionDutchAuctionEnd`
    - [x] `ActionDutchAuctionWithdraw`
    - [x] `ActionDutchAuctionWithdrawView`
    - [x] `ActionDutchAuctionScheduleView`
- [x] #4212
[...]

PR #4205, opened 2024-04-14
auction: create component scaffolding

erwanor:
## Describe your changes

This create scaffolding for an auction component, including structure for future `state_key`s, `event`s, and `rpc` definition.

Once #4204 is merged, I will rebase against it and do a second pass on the state keys. 

Part of #4196 

  > This is not consensus breaking (yet) because the component is not used by the app.


cratelyn (approved):
🏗️ 🔜

PR #4207, opened 2024-04-15
fix(asset): 🔊 `Metadata: Debug` includes all fields

### LINKED ISSUE #4190: `asset::Metadata` `Debug` impl throws away all data except base denom
**Describe the bug**

The `Debug` impl for `asset::Metadata` throws away all data except for the base denom. This is extremely inconvenient for debugging.

cratelyn:
fixes #4190.

> The `Debug` impl for `asset::Metadata` throws away all data except for
the base denom. This is extremely inconvenient for debugging.

this commit updates `Metadata`'s impl of `std::fmt::Debug` so that it includes all of the fields in the `Inner` structure.

note for review: #4190 didn't explicitly mention which fields to include, so i have included all of them as a starting point. if we should find a middle ground, i am open and interested in hearing about that!

#### issue ticket number and link

#4190

  > debug representations should not affect consensus.


hdevalence (approved):
LGTM!

If we notice, as a result of this change, that there are places in the code where we're putting a ton of info in a tracing event (because we did `?metadata` or something, relying on current behavior), we should change those to do the base denom or ID as appropriate

PR #4215, opened 2024-04-15
Update guide for pcli threshold

hdevalence:
## Describe your changes

Restructures `guide.penumbra.zone` to make it more useful and to open space for `pcli` threshold docs.

  > only changes to the guide


cronokirby (approved):

conorsch (changes requested):
There's breakage here where the dynamic import for version/tag: 

```
Error updating "{{ #include ../penumbra_version.md }}", Could not read file for link {{ #include ../penumbra_version.md }}
```

Solution is just to update the relpaths. I'll handle that as part of review. Ideally CI would have fallen down on this, too.

conorsch:
>  Ideally CI would have fallen down on this, too.

Evidently that's a thing, see https://github.com/rust-lang/mdBook/issues/1094. Fixed the links for our docs throughout.

conorsch (approved):
Great reorg. :sunglasses:

Issue #4212, opened 2024-04-15
auction: build `penumbra-transaction` support 


erwanor:
We need to implement `IsAction` and `GasCosts` prior to including the actions in `penumbra-transaction`:

**Implementing `IsAction`**
- [x] `ActionDutchAuctionSchedule`
- [x] `ActionDutchAuctionEnd`
- [x] `ActionDutchAuctionWithdraw`

**Implementing `GasCosts`**
- [x] `ActionDutchAuctionSchedule`
- [x] `ActionDutchAuctionEnd`
- [x] `ActionDutchAuctionWithdraw`

We should be particularly attentive to the value balance calculations and getting the gas cost vectors right.

erwanor:
This isn't done yet, we need to finish integration in `penumbra-transaction`

PR #4218, opened 2024-04-16
auction: define action balances

erwanor:
## Describe your changes

This PR:
- add `auctionnft`s to the asset registry
- augment `AuctionNft` with denom metadata
- implements `DutchAuctionDescription::id`
- implements value `balance` for each action
- implements computing the balance commitment for the withdraw action

This is a prelude to #4212 

  > This isn't connected to the rest of the app yet.

### DIFF
--- a/crates/core/component/auction/src/auction/dutch.rs
+++ b/crates/core/component/auction/src/auction/dutch.rs
@@ -9,7 +9,7 @@ use crate::auction::AuctionId;
 
 pub mod actions;
 
-pub const DUTCH_AUCTION_DOMAIN_SEP: &[u8] = b"penumbra_dutch_auction_nft";
+pub const DUTCH_AUCTION_DOMAIN_SEP: &[u8] = b"penumbra_DA_nft";
 
 /// A deployed Dutch Auction, containing an immutable description
 /// and stateful data about its current state.
--- a/crates/core/component/auction/src/auction/dutch.rs
+++ b/crates/core/component/auction/src/auction/dutch.rs
@@ -5,6 +5,8 @@ use penumbra_num::Amount;
 use penumbra_proto::{core::component::auction::v1alpha1 as pb, DomainType};
 use serde::{Deserialize, Serialize};
 
+use crate::auction::AuctionId;
+
 pub mod actions;
 
 /// A deployed Dutch Auction, containing an immutable description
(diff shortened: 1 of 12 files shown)


crates/core/component/auction/src/auction/dutch/actions/withdraw.rs
```diff
@@ -15,6 +17,44 @@ pub struct ActionDutchAuctionWithdraw {
...
+    ///
+    /// More context: [Actions and Value balance][protocol-spec]
+    /// [protocol-spec]: https://protocol.penumbra.zone/main/transactions.html#actions-and-value-balance
+    pub fn balance_commitment(&self) -> balance::Commitment {
+        let prev_auction_nft = Balance::from(Value {
+            amount: 1u128.into(),
+            // The sequence number should always be >= 1, because we can
+            // only withdraw an auction that has ended (i.e. with sequence number `>=1`).
+            // We use a saturating operation defensively so that we don't underflow.
+            asset_id: AuctionNft::new(self.auction_id, self.seq.saturating_sub(1)).asset_id(),
```
erwanor, line 44: My rationale to do this is that we check the binding signature before we run any action validation checks, so even if this MUST be `seq >= 1`, we can't actually rely on that yet.

crates/core/component/auction/src/auction/nft.rs
```diff
@@ -1,15 +1,31 @@
...
+    pub seq: u64,
+    /// The metadata corresponding to the nft denom.
+    pub metadata: asset::Metadata,
+}
+
+impl AuctionNft {
+    pub fn new(id: AuctionId, seq: u64) -> AuctionNft {
+        let metadata = asset::REGISTRY
+            .parse_denom(&format!("auctionnft_{seq}_{id}"))
+            .expect("auction nft denom is valid");
```
erwanor, line 22: I think I want to do a second pass on both auction and lp nfts to make their constructor fallible.

crates/core/component/auction/src/auction/dutch.rs
```diff
@@ -65,6 +69,28 @@ pub struct DutchAuctionDescription {
     pub nonce: [u8; 32],
 }
 
+impl DutchAuctionDescription {
+    /// Compute the unique identifier for the auction description.
+    pub fn id(&self) -> AuctionId {
+        let mut state = blake2b_simd::Params::default()
+            .personal(DUTCH_AUCTION_DOMAIN_SEP)
```
redshiftzero, line 76: I think the personalization bytes should be maximum 16 bytes: https://docs.rs/blake2b_simd/latest/blake2b_simd/struct.Params.html#method.personal

redshiftzero (changes requested):
Value balance changes LGTM! one small comment inline

crates/core/component/auction/src/auction/dutch.rs
erwanor: oh great catch

redshiftzero (approved):

Issue #4211, opened 2024-04-15
auction: define action handlers and validation rules


erwanor:
Once we have #4206, we can start filling in their validation/execution handlers:
- [x] `ActionDutchAuctionSchedule`
- [x] `ActionDutchAuctionEnd`
- [x] `ActionDutchAuctionWithdraw`

erwanor:
For the validation rules, we should try to use `anyhow::ensure!(INVARIANT, MSG_INVARIANT_CONTEXT)` as much as possible, so that we can formulate validation rules positively (like they would be in a natural language spec) e.g.

```rust
// Spec: The sequence number must be zero.
ensure!(seq == 0, "opened auction nft MUST have sequence set to zero (got {bad_seq})");
```

TalDerei:
can this issue be closed?

erwanor:
Not yet, we have to finish the component impl and then #4228

PR #4220, opened 2024-04-16
auction: `ActionDutchAuction*` validation and component impl

erwanor:
## Describe your changes

This PR adds validation logic for:
- `ActionDutchAuctionSchedule`
- `ActionDutchAuctionEnd`
- `ActionDutchAuctionWithdraw`

and a component implementation, on a model similar-but-simpler to the DEX:

- a `DutchAuctionManager` extension trait defines a crate-level API to effect a validated action
- an `Inner` internal implementation manages writing to the state
- internal data can be accessed via a general `AuctionStoreRead` trait which manage tentatively deserializing raw auction state (wkt `Any`) into a useful domain type.

#4211 

  > Those actions are not integrated in our transaction system, our application impl.

### DIFF
--- a/crates/core/component/auction/src/component/action_handler/dutch/end.rs
+++ b/crates/core/component/auction/src/component/action_handler/dutch/end.rs
@@ -5,6 +5,8 @@ use cnidarium_component::ActionHandler;
 
 use crate::auction::dutch::ActionDutchAuctionEnd;
 use crate::component::AuctionStoreRead;
+use crate::component::DutchAuctionManager;
+
 use anyhow::{bail, Context};
 
 #[async_trait]
@@ -14,7 +16,7 @@ impl ActionHandler for ActionDutchAuctionEnd {
         Ok(())
     }
 
-    async fn check_and_execute<S: StateWrite>(&self, state: S) -> Result<()> {
+    async fn check_and_execute<S: StateWrite>(&self, mut state: S) -> Result<()> {
         let auction_id = self.auction_id;
 
         let auction_state = state
@@ -32,6 +34,8 @@ impl ActionHandler for ActionDutchAuctionEnd {
             "auction MUST have a sequence number set to opened (0) or closed (1) (got: {})",
             auction.state.sequence
         );
+
+        state.close_auction(auction).await?;
         Ok(())
     }
 }
--- a/crates/core/component/auction/src/component/action_handler/dutch/end.rs
+++ b/crates/core/component/auction/src/component/action_handler/dutch/end.rs
@@ -1,9 +1,11 @@
-use anyhow::Result;
+use anyhow::{ensure, Result};
 use async_trait::async_trait;
 use cnidarium::StateWrite;
 use cnidarium_component::ActionHandler;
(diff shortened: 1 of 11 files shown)


crates/core/component/auction/src/component/action_handler/dutch/end.rs
```diff
@@ -12,7 +16,26 @@ impl ActionHandler for ActionDutchAuctionEnd {
...
+            .await
+            .context("the auction associated with this id is not a dutch auction")?;
+
+        let Some(auction) = auction_state else {
+            bail!("no auction found for id {auction_id}")
+        };
+
+        // Check that the sequence number for the auction state is 0 (opened) or 1 (closed).
+        ensure!(
+            auction.state.sequence <= 1,
```
cratelyn, line 33: tioli: i might write this to directly check that `sequence == 0 || sequence == 1`, or `matches!(sequence, 0 | 1)`

crates/core/component/auction/src/component/action_handler/dutch/schedule.rs
```diff
@@ -1,18 +1,98 @@
...
+        );
+
+        // Check that the step count is less than 1000.
+        ensure!(
+            step_count <= 1000,
+            "the dutch auction step count MUST be less than 1000 (got: {step_count})",
+        );
+
+        // Check that height delta is a multiple of `step_count`.
+        let block_window = end_height - start_height;
```
cratelyn, line 66: if we wrote this with `checked_sub` one of the invariant checks above would be addressed naturally here. i would "like" that but it is fine as written :)

crates/core/component/auction/src/state_key.rs
```diff
@@ -8,11 +8,11 @@ pub mod parameters {
     }
 }
 
-pub mod store {
+pub mod auction_store {
     use crate::auction::id::AuctionId;
 
     pub fn prefix() -> &'static str {
-        "auction/store/"
+        "auction/auction_store/"
```
cratelyn, line 15: i might mark this as `consensus-breaking` to be formal, even if the auction component isn't technically live yet

cratelyn (approved):
this is very well done, as evidenced by only having rather subjective comments to offer.

i've marked optional "_take it or leave it_" comments as such, and noted one place where we are making a consensus-breaking change.

great job!

crates/core/component/auction/src/state_key.rs
erwanor: i hear you, but i think the tag is more useful if it lets us get a filtered view of what is actually consensus compatible vs what isn't

crates/core/component/auction/src/component/action_handler/dutch/schedule.rs
erwanor: i think having both is better, good idea

crates/core/component/auction/src/component/action_handler/dutch/end.rs
erwanor: sg

PR #4221, opened 2024-04-16
dex: add checks to disable swaps and opening positions when the dex is disabled

### LINKED ISSUE #4078: Implement an enabled switch for the DEX

Along the way to #4066, we added an `is_enabled` field to the new `DexParameters`. However, this field is currently not used at all.

Add checks to the following actions so that they are auto rejected unless `is_enabled == true` at the time of execution:

- [x] `Swap`
- [x] `PositionOpen`

These are the "inflows" to the DEX. We do _not_ want to reject the following methods, which are "outflows":

- `SwapClaim`
- `PositionClose`
- `PositionWithdraw`

This ensures that the `is_enabled` field just turns on / off trading but does not interfere with people getting funds they already have. Otherwise, we'd have a situation where a governance controlled parameter could effectively lock funds.

This could be combined with #4025.

aubrika:
## Describe your changes

Closes #4078 

Unsure if `anyhow::bail!` is the ideal behavior for swaps & position opens when the dex is disabled, but it seems right to display an explanatory message as to why the swap or open does not occur.

https://github.com/penumbra-zone/penumbra/issues/4078

  > This will alter the dex behavior across the change boundary (acceptance/rejection of swaps & position opens) so it should break consensus between versions.

### DIFF
--- a/crates/core/component/dex/src/component/action_handler/swap.rs
+++ b/crates/core/component/dex/src/component/action_handler/swap.rs
@@ -1,4 +1,4 @@
-use anyhow::Result;
+use anyhow::{ensure, Result};
 use async_trait::async_trait;
 use cnidarium::StateWrite;
 use cnidarium_component::ActionHandler;
@@ -34,6 +34,14 @@ impl ActionHandler for Swap {
     }
 
     async fn check_and_execute<S: StateWrite>(&self, mut state: S) -> Result<()> {
+        // Only execute the swap if the dex is enabled in the dex params.
+        let dex_params = state.get_dex_params().await?;
+
+        ensure!(
+            dex_params.is_enabled,
+            "Dex MUST be enabled to process swap actions."
+        );
+
         let swap_start = std::time::Instant::now();
         let swap = self;
 
--- a/crates/core/component/dex/src/component/action_handler/position/open.rs
+++ b/crates/core/component/dex/src/component/action_handler/position/open.rs
@@ -1,10 +1,10 @@
-use anyhow::Result;
+use anyhow::{ensure, Result};
 use async_trait::async_trait;
 use cnidarium::StateWrite;
 use cnidarium_component::ActionHandler;
 
 use crate::{
-    component::PositionManager,
+    component::{PositionManager, StateReadExt},
     lp::{action::PositionOpen, position},
 };
 
@@ -28,6 +28,14 @@ impl ActionHandler for PositionOpen {
     }
 
     async fn check_and_execute<S: StateWrite>(&self, mut state: S) -> Result<()> {
+        // Only open the position if the dex is enabled in the dex params.
+        let dex_params = state.get_dex_params().await?;
+
+        ensure!(
+            dex_params.is_enabled,
+            "Dex MUST be enabled to open positions."
+        );
+
         state.open_position(self.position.clone()).await?;
         Ok(())
     }


crates/core/component/dex/src/component/action_handler/swap.rs
```diff
@@ -34,32 +34,39 @@ impl ActionHandler for Swap {
...
-        let swap_start = std::time::Instant::now();
-        let swap = self;
+        // Only execute the swap if the dex is enabled in the dex params.
+        let dex_params = state.get_dex_params().await?;
 
-        // All swaps will be tallied for the block so the
-        // BatchSwapOutputData for the trading pair/block height can
-        // be set during `end_block`.
-        let mut swap_flow = state.swap_flow(&swap.body.trading_pair);
+        if dex_params.is_enabled {
```
erwanor, line 40: would prefer a single:

`ensure!(dex_params.is_enabled, "dex MUST be enabled to process actions");`

This is a nice pattern because it makes the invariant easy to read and will concisely resolve to an early `return Err("... must be ...")` similar to `bail!`.

erwanor: we should also add it to other position actions

erwanor (changes requested):

crates/core/component/dex/src/component/action_handler/swap.rs
aubrika: Oh rad, yeah, that ensure pattern was the kind of suggestion I was looking for, I like that more. :)

However, as for adding it to the other actions, Henry noted that we don't want to disable the "outflow" actions:

```
We do not want to reject the following methods, which are "outflows":

SwapClaim
PositionClose
PositionWithdraw
This ensures that the is_enabled field just turns on / off trading but does not interfere with people getting funds they already have. Otherwise, we'd have a situation where a governance controlled parameter could effectively lock funds.
```

erwanor: for auctions, we will also check this but i can add it later (in addition to auction-level switches)

erwanor: good point, makes sense

erwanor (approved):
I think we could move the check in `check_historical` (default to a no-op, but relevant in cases when we have an invariant that depends on state that the transaction cannot effect, like in this case a component parameter), but IMO no need to block the PR on this.

aubrika:
> I think we could move the check in `check_historical` (default to a no-op, but relevant in cases when we have an invariant that depends on state that the transaction cannot effect, like in this case a component parameter), but IMO no need to block the PR on this.

if we move the check to `check_historical`, will that induce a delay in dex activation, and is that desirable behavior? would the reasoning for moving the check be to avoid the computational overhead of `check_historical` where that would occur in other actions?

erwanor:
You can parallelize checks performed in `check_historical` with a big caveat that it risks introducing TOCTOU vulnerabilities if, in order to be sound, a transaction *action check* must be able to observe the effect of preceding action execution (within the txs). `check_and_execute` is a safe, albeit less "optimal" default, because the check is folded with the execution of the action itself

erwanor:
Tbc, the overhead here is minuscule, but I thought I would bring it up as an example of `check_historical` usage that's sound.

Issue #2105, opened 2023-03-06
Add governance state machine tests


hdevalence:
**Is your feature request related to a problem? Please describe.**

Now that we have governance support, we should write tests that exercise the governance code.  These tests should be written as state machine tests, since what we care about is the nooks and crannies of the state machine.

The `StateDelta` system may be helpful, since we can exercise different codepaths with different forks.

conorsch:
Closely related to #1345. We're optimistic this'll ship with Testnet 51.

erwanor:
This is still in-progress, but deprioritized in favor of the distributions modules.

hdevalence:
Blocked by #3588.

cratelyn:
i believe this is no longer blocked, so i'm going to update the status of this issue.

PR #4222, opened 2024-04-16
feat(governance): 🧊 test that community pool spend proposals can be disabled

### LINKED ISSUE #4079: Add governance parameter for freezing/unfreezing the community pool

To minimize risk, we want to reduce the consequences of a governance-related exploit.  Except for community pool transactions, governance proposals have primarily signaling value.  This would be useful to enshrine in the form of a chain parameter that disables all community pool transactions.

Add a `bool community_pool_is_frozen` parameter to the governance parameters. When set:

- [ ] Proposals for community pool transactions are not accepted
- [ ] Community pool transactions cannot execute

cratelyn:
fixes #4079. related to, but does not finish #2105.

~~this introduces a new governance parameter that can be used to freeze or unfreeze the community pool. this branch additionally introduces a few mock consensus tests to exercise the governance component, to show that it works when unfrozen, and is properly disabled when frozen.~~

this introduces test coverage to show that community pool spend proposals can be disabled via a chain parameter.

### "the point"

~~the core mechanical change of this branch is made in these commits:~~ see https://github.com/penumbra-zone/penumbra/pull/4222#issuecomment-2064056971 below.

* ~~f88a98130 feat(app): ✅ no spend proposals when pool is frozen~~
* ~~512901254 feat(app): 🧊 community pool actions not accepted when frozen~~
* ~~51e70630f feat(proto): 🏂 add `community_pool_is_frozen` to governance params~~

~~**NB:** this is a consensus-breaking change, as a parameter will now block community pool transactions when this new parameter is active.~~

### related changes

some changes made, related to writing tests for the governance component:

* 6b3be5892 refactor(governance): 💶 `Value` accessors for proposal actions
* de16dace5 feat(tx): 🌮 `ActionPlan: From` for community pool actions
* 4e6b874ed feat(amount): ➖ `Amount: std::ops::SubAssign`

### tests

this addresses part of #2105. we should add tests that exercise the other
tallying logic before that issue is complete. tests are added in the following
commits:

* fedfa8ded tests(app): 💭 test a community pool spend
* 201fe4873 tests(app): 🏊 test a community pool deposit
* 8543869d4 tests(app): 🥶 test that community pool can be frozen


crates/core/app/tests/app_can_propose_community_pool_spends.rs
```diff
@@ -0,0 +1,296 @@
...
+
+    // Execute the transaction, applying it to the chain state.
+    test_node
+        .block()
+        .with_data(vec![tx.encode_to_vec()])
+        .execute()
+        .instrument(error_span!("executing block with governance proposal"))
+        .await?;
+    let post_proposal_snapshot = storage.latest_snapshot();
+    panic!("stop here"); // XXX this transaction does not execute successfuly, panic here for now.
```
cratelyn, line 231: **question:** this currently fails with:

```
  2024-04-16T19:48:50.055373Z  INFO penumbra_app::server::consensus: deliver_tx failed, e: failed to deliver transaction

Caused by:
    0: check_stateless failed
    1: binding signature failed to verify
    2: Invalid signature.
    at crates/core/app/src/server/consensus.rs:207
    in penumbra_mock_consensus::block::execute with height: 2, time: 1713296930
    in app_can_propose_community_pool_spends::executing block with governance proposal
```

i believe this is because i should also be https://protocol.penumbra.zone/main/governance.html#proposal-nfts generating the `proposal_1_submitted` nft as an output(?) alongside the governance proposal itself.

hdevalence:
> #### :speech_balloon: a tertiary discussion point
> 
> 
> 
> i see the following parameter: https://github.com/penumbra-zone/penumbra/blob/main/crates/core/component/community-pool/src/params.rs#L10-L13
> 
> 
> 
> my understanding is that this is similar, but not completely the same as, `community_pool_is_frozen`. i'd like to briefly gesture at it though, in case this sounds incorrect.

Hmm, it looks pretty much the same to me

hdevalence:
maybe we don't need both?

cratelyn:
> > #### 💬 a tertiary discussion point
> > i see the following parameter: https://github.com/penumbra-zone/penumbra/blob/main/crates/core/component/community-pool/src/params.rs#L10-L13
> > my understanding is that this is similar, but not completely the same as, `community_pool_is_frozen`. i'd like to briefly gesture at it though, in case this sounds incorrect.
> 
> Hmm, it looks pretty much the same to me

cool, i am glad i asked! to state the difference explicitly, this new parameter also prevents community pool deposits when the pool is frozen. whether that is a useful property, i'm unsure.

we probably don't need both though, i think that is a fair point nonetheless. in that case @hdevalence, do you think that we should keep the existing option, or remove/deprecate it in favor of this option?

hdevalence:
I think we should just have one parameter, it was an oversight on my part to not check that the work was already completed. The important thing is making sure that there's a chain parameter that prevents any community pool transactions from occurring.

cratelyn:
> I think we should just have one parameter, it was an oversight on my part to not check that the work was already completed. The important thing is making sure that there's a chain parameter that prevents any community pool transactions from occurring.

sounds good. i'll suggest the following course:

* **1.** revert the commits here introducing a new parameter.
* **2.** rephrase the tests to work with the existing parameter.

at risk of being pedantic again, the existing `community_pool_spend_proposals_enabled` only blocks specifically the proposal of community pool spend proposals, and transactions depositing into the pool are still allowed when that flag is on. that is, `community_pool_spend_proposals_enabled` lets funds go _into_ the pool, but not _out_.

i think that is fine, but i want to be sure that it serves our needs, re: "_prevents any community pool transactions from occurring._" if so, this pr doesn't need to be consensus-breaking, and we can just treat this as the addition of more tests.

crates/core/component/governance/src/proposal_submit/action.rs
```diff
@@ -30,21 +30,30 @@ impl EffectingData for ProposalSubmit {
...
-            amount: self.deposit_amount,
-            asset_id: *STAKING_TOKEN_ASSET_ID,
-        };
-
-        let proposal_nft = Value {
-            amount: Amount::from(1u64),
-            asset_id: ProposalNft::deposit(self.proposal.id).denom().into(),
-        };
+        let deposit = self.deposit_value();
+        let proposal_nft = self.proposal_nft_value();
```
erwanor, line 34: ACK. LGTM

crates/core/component/governance/src/proposal_withdraw/action.rs
```diff
@@ -33,19 +33,29 @@ impl From<ProposalWithdraw> for pb::ProposalWithdraw {
 }
 
 impl ProposalWithdraw {
-    /// Compute a commitment to the value contributed to a transaction by this proposal submission.
+    /// Compute a commitment to the value contributed to a transaction by this proposal withdrawal.
     pub fn balance(&self) -> Balance {
-        let voting_proposal_nft = Value {
+        let voting_proposal_nft = self.voting_proposal_nft_value();
+        let withdrawn_proposal_nft = self.withdrawn_proposal_nft();
```
erwanor, line 39: ACK. LGTM

erwanor (approved):

PR #4223, opened 2024-04-16
auction: add `GasCost` impls for actions

TalDerei:
## Describe your changes
Add `GasCost` vectors for DA actions, add`Action` / `ActionPlan` skeletons, and make gas fee costs consistent with protobuf encodings (https://github.com/penumbra-zone/penumbra/issues/3940)

References #4212

### DIFF
--- a/crates/core/transaction/src/gas.rs
+++ b/crates/core/transaction/src/gas.rs
@@ -45,10 +45,9 @@ pub fn spend_gas_cost() -> Gas {
         // penumbra.crypto.decaf377_rdsa.v1.SpendVerificationKey `rk`     = 32 bytes
         // penumbra.crypto.decaf377_rdsa.v1.SpendAuthSignature `auth_sig` = 64 bytes
         // ZKSpendProof `proof`                                           = 192 bytes
-        // ````````````````````````````` 352 bytes ``````````````````````````````````
 
         // The block space measured as the byte length of the encoded action.
-        block_space: 160 + ZKPROOF_SIZE,
+        block_space: 160 + ZKPROOF_SIZE, // 352 bytes
         // The compact block space cost is based on the byte size of the data the [`Action`] adds
         // to the compact block. For a `Spend`, this is the byte size of a `Nullifier`.
         compact_block_space: NULLIFIER_SIZE,
@@ -66,10 +65,9 @@ pub fn output_gas_cost() -> Gas {
         // wrapped_memo_key `wrapped_memo_key`                           = 48 bytes
         // ovk_wrapped_key `ovk_wrapped_key`                             = 48 bytes
         // ZKOutputProof `proof`                                         = 192 bytes
-        // ````````````````````````````` 560 bytes ``````````````````````````````````
 
         // The block space measured as the byte length of the encoded action.
-        block_space: 128 + NOTEPAYLOAD_SIZE + ZKPROOF_SIZE,
+        block_space: 128 + NOTEPAYLOAD_SIZE + ZKPROOF_SIZE, // 560 bytes
         // The compact block space cost is based on the byte size of the data the [`Action`] adds
         // to the compact block.
         compact_block_space: NOTEPAYLOAD_SIZE,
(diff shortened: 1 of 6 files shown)


TalDerei:
@erwanor @zbuc should we make the DA fees be counted towards the `block_space` instead of the `execution`?

zbuc:
> @erwanor @zbuc should we make the DA fees be counted towards the `block_space` instead of the `execution`?

Do you mean the part multiplied by the step count? I think there is both a per-step block space and per-step execution time cost that should be accounted for. Doesn't each step incur a block space cost of storing liquidity positions?

TalDerei:
> > @erwanor @zbuc should we make the DA fees be counted towards the `block_space` instead of the `execution`?
> 
> Do you mean the part multiplied by the step count? I think there is both a per-step block space and per-step execution time cost that should be accounted for. Doesn't each step incur a block space cost of storing liquidity positions?

yeah that's true, would the `block_space` therefore be something like `proto encoding cost of action + (dutch_action_schedule.description.step_count * 8)`?

zbuc:
> `proto encoding cost of action + (dutch_action_schedule.description.step_count * 8 bytes / step)`

I am operating under the assumption (that I am not 100% sure of) that every step, a LP will be closed (and withdrawn?) and a new one opened.

So I think it should be something like `proto encoding cost of action + dutch_action_schedule.description.step_count * (PositionOpen block space cost + PositionClose block space cost + PositionWithdraw block space cost)`

hdevalence:
I don't think I follow — how does the chain opening and closing positions on its own add data to a future block?

zbuc:
> I don't think I follow — how does the chain opening and closing positions on its own add data to a future block?

I suppose it's not adding new `Action`s to the block and instead using only additional storage space in the JMT for the position data, and we don't account for JMT space usage in gas pricing.

crates/core/transaction/src/gas.rs
```diff
@@ -216,21 +227,71 @@ fn delegator_vote_gas_cost() -> Gas {
...
+        // penumbra.core.asset.v1.Value `input` = 48 bytes
+        // penumbra.core.asset.v1.AssetId `output_id` = 32 bytes
+        // penumbra.core.num.v1.Amount `max_output` = 16 bytes
+        // penumbra.core.num.v1.Amount `min_output` = 16 bytes
+        // uint64 `start_height` = 8 bytes
+        // uint64 `end_height` = 8 bytes
+        // uint64 `step_count` = 8 bytes
+        // bytes `nonce` = 32 bytes
+
+        // TODO: Multidimentional fees will enable gas fees to dynamically be based on both `block_space`
```
erwanor, line 258: I would remove this comment, it doesn't seem (?) directly relevant to the schedule action, and I think it's technically off because what we're missing is the dynamic price  adjustments based on usage/utilization targets.

crates/core/transaction/src/gas.rs
```diff
@@ -216,21 +227,71 @@ fn delegator_vote_gas_cost() -> Gas {
...
+        // while block space is proportional to the space the action takes up in a block.
+        // We can therefore separately price the resource of “work performed during execution”
+        // and “bandwidth between nodes”.
+        //
+        // Currently we make the execution cost for DA actions proportional to the number of steps
+        // and costs of position open/close in dutch action.
+        block_space: 168, // 168 bytes
+        compact_block_space: 0,
+        verification: 50,
+        execution: 2 * dutch_action_schedule.description.step_count * 20,
```
erwanor, line 269: Could we add a comment to explain where `20` is coming from, and another reminder in  the `PositionOpen` and `PositionClose` gas vectors to update this value if they change.

crates/core/transaction/src/gas.rs
```diff
@@ -216,21 +227,71 @@ fn delegator_vote_gas_cost() -> Gas {
...
+
+fn dutch_auction_withdraw_gas_cost() -> Gas {
+    Gas {
+        // AuctionId `auction_id` = 32 bytes
+        // uint64 `seq`= 8 bytes
+        // penumbra.core.asset.v1.BalanceCommitment `reserves_commitment` = 32 bytes
+        block_space: 72, // 72 bytes
+        compact_block_space: 0,
+        verification: 0,
+        execution: 10,
```
erwanor, line 291: Ditto, can we just mention this is the cost ~ of position withdraw and add a reminder there to update this value.

erwanor (approved):
Requesting changes, but good to merge as soon as we have addressed the comments + green CI. I would be happy to pair on a second pass that makes the dependency between DA actions and the execution costs of position actions a little better (rather than having reminders in comments). But it seems good enough for now. Thanks for taking this on!

crates/core/transaction/src/gas.rs
TalDerei: forgot to remove this!

crates/core/transaction/src/gas.rs
TalDerei: added comments for clarification

crates/core/transaction/src/gas.rs
TalDerei: added comments for clarification

PR #4224, opened 2024-04-16
dex(base_liq): add `async_trait` to `BaseLiquidityIndex`

erwanor:
## Describe your changes

This slipped when doing #4188. I'm not completely sure yet why CI didn't catch this. 

  > Internal


aubrika (approved):
LGTM

Issue #3417, opened 2023-11-20
ci: publish versioned urls for docs


conorsch:
Our docs always live under `main` in the URL route for the guide: https://guide.penumbra.zone/main/pd/build.html However, we don't update those docs in line with releases, so the information on `main` is not correct for testnet users. See recent changes in https://github.com/penumbra-zone/penumbra/pull/3395, for instance: the docs were updated to remove mention of `--home`, but that's only true of the `main` branch, not of the currently running `v0.63.2` release.

We should continue to update docs on merge into main, but also tag new versions during releases, e.g. `https://guide.penumbra.zone/v0.63.2/pd/build.html`. That will help us communicating about docs.

hdevalence:
xref https://github.com/penumbra-zone/penumbra/issues/6

conorsch:
I still want this problem solved. From the historical issue, looks like Firebase can't support this use case as described. Any objections to simplifying the URLs in the meantime? I'd rather see `https://guide.penumbra.zone/` in my browser than `https://guide.penumbra.zone/main/index.html`, especially if I cannot navigate to `https://guide.penumbra.zone/v0.71.0/index.html`. We can specify redirects in the Firebase config so that historical URLs out in the wild don't break for people—that's what led to #4232 anyway.

It's no problem to migrate to historical tags later on: we can always check out versions and export the html and stick it in a subdir if we care about doing that.

PR #4232, opened 2024-04-17
docs: add a few redirects

conorsch:
## Describe your changes

Follow-up to #4215, in which we reorganized the guide for readability. Some of the existing content got shuffled into subroutes, and already one person in Discord has pointed out that we broken URLs. Tacking on some http redirects via mdbook [0].

[0] https://rust-lang.github.io/mdBook/format/configuration/renderers.html#outputhtmlredirect

  > docs-only, no code changes.

### DIFF
--- a/docs/guide/book.toml
+++ b/docs/guide/book.toml
@@ -27,3 +27,17 @@ warning-policy = "error"
 # The buf.build website doesn't support HTTP HEAD calls, returning 405.
 # Only relevant when `follow-web-links=true`.
 exclude = ['buf\.build']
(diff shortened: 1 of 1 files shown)


erwanor (approved):

conorsch:
This didn't work as well as I'd hoped: the mdbook settings are html-level redirects, so they display a warning in Firefox. Worse, the redirects conflict with the `/` -> `/main/index.html` redirect specified in the Firebase config. Going to follow up on this in #3417.

Issue #4226, opened 2024-04-17
auction: define rpc service in the protos and integrate in `pd`


erwanor:
Once we have completed our first pass at protocol support for auctions, we will want to expose the associated chain state via a **minimal** query service:

```protobuf
// Query operations for the auction component.
service QueryService {
  // Get the current state of an auction by ID.
  rpc AuctionStateById(AuctionStateByIdRequest) returns (AuctionStateByIdResponse);
  // Get the current state of a group of auctions by ID.
  rpc AuctionStateByIds(AuctionStateByIdsRequest) returns (stream AuctionStateByIdsResponse);
}

message AuctionStateByIdRequest {
    AuctionId id = 1;
}

message AuctionStateByIdResponse {
    // If present, the state of the auction. If not present, no such auction is known.
    google.protobuf.Any auction = 2;
    // The state of any DEX positions relevant to the returned auction.
    //
    // Could be empty, depending on the auction state.
    repeated core.component.dex.v1.Position positions = 3;
}

message AuctionStateByIdsRequest {
    // The auction IDs to request. Only known IDs will be returned in the response.
    repeated AuctionId id = 1;
}

message AuctionStateByIdsResponse {
    // The auction ID of the returned auction.
    AuctionId id = 1;
    // The state of the returned auction.
    google.protobuf.Any auction = 2;
    // The state of any DEX positions relevant to the returned auction.
    //
    // Could be empty, depending on the auction state.
    repeated core.component.dex.v1.Position positions = 3;
}
```

In a separate PR, we should implement these methods in the `penumbra_auction::component::rpc` module.

erwanor:
This is mostly done, we will fold the auction query service in the pd service stack with #4228

PR #4233, opened 2024-04-17
auction: pd service rpc

TalDerei:
## Describe your changes
Define minimal auction pd rpc service in the protos. 

References https://github.com/penumbra-zone/penumbra/issues/4226


proto/penumbra/penumbra/core/component/auction/v1alpha1/auction.proto
```diff
@@ -14,8 +14,24 @@ message GenesisContent {
...
+  rpc AuctionStateById(AuctionStateByIdRequest) returns (AuctionStateByIdResponse);
+}
+
+message AuctionStateByIdRequest {
+    AuctionId id = 1;
+}
+
+message AuctionStateByIdResponse {
+    // If present, the state of the auction. If not present, no such auction is known.
+    DutchAuctionState auction = 2;
```
erwanor, line 31: We should change this to an `Any` type and let clients inspect/deserialize to the concrete auction that the type mentions. The goal of this RPC is to be a minimal/general purpose "get me the auction state".

crates/core/component/auction/src/component/rpc.rs
```diff
@@ -22,4 +25,12 @@ impl Server {
...
 
 #[tonic::async_trait]
-impl QueryService for Server {}
+impl QueryService for Server {
+    #[instrument(skip(self, request))]
+    async fn auction_state_by_id(
+        &self,
+        request: tonic::Request<AuctionStateByIdRequest>,
+    ) -> Result<tonic::Response<AuctionStateByIdResponse>, Status> {
+        todo!()
```
erwanor, line 79: Once we have modified the proto method signature to have `google.protobuf.Any`, we can: 
1. Pull a snapshot (read-only view of the chain state): `let storage = storage.latest_snapshot();`
2. Use `AuctionStoreRead::get_raw_auction` to get an `Option<Any>`
3. Inform the client if nothing was found, or
4. Prepare a tonic response if there was a match

I think one missing piece of context here, is that we want to have a minimal RPC that work for a variety of auction types and defer to clients to support (or not) serialization into a concrete type.

erwanor (changes requested):

proto/penumbra/penumbra/core/component/auction/v1alpha1/auction.proto
TalDerei: changed to `google.protobuf.Any`

crates/core/component/auction/src/component/rpc.rs
TalDerei: I modified `get_raw_auction` to use `pbjson_types::Any` instead of `prost_types::Any` since the tonic response `AuctionStateByIdResponse` expects that, otherwise it throws mismatched type error. I think we can keep the streaming-based `auction_state_by_ids` unimplemented for now and circle back once we've sufficiently tested `auction_state_by_id`. 

I think having a minimal rpc where the client filters / serializes the raw data into a concrete type is a good design choice.

erwanor (approved):
Merge on green CI, the change to use the `pbjson_types` override seems fine, as far as I can tell it "just" consumes the `prost-types` WKTs and augments it with ser/des. We can revisit later anyway, good work ship it

PR #4235, opened 2024-04-18
pcli: fix terminal input in pcli threshold

### LINKED ISSUE #4230: `pcli threshold sign` hangs when pasting input into the terminal
**Describe the bug**

`pcli threshold sign` hangs when pasting a message in (on MacOS, using ITerm2 and `fish`, if that turns out to matter)

**To Reproduce**
1. Run `pcli threshold sign`
2. Paste the coordinator's message
3. It hangs

I'm expecting this is something about the way we're reading data from the terminal.

hdevalence:
## Describe your changes

When reading input from the terminal in `pcli threshold` commands, circumvent OS-imposed restrictions on line length (required for interactive editing and notably low on MacOS) by dropping into "raw mode", reading the input a byte at a time and manually printing out the read byte to the user, then return to "cooked mode" (yes, that's what it's called) after detecting `\n` or `\r`.

Closes #4230, which I confirmed through my own testing.

  > only changes to `pcli` terminal handling

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3669,6 +3669,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "libredox"
+version = "0.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3af92c55d7d839293953fcd0fda5ecfe93297cfde6ffbdec13b41d99c0ba6607"
+dependencies = [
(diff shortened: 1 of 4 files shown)


hdevalence:
@conorsch can we pull this into the bugfix release or a soon-subsequent patch release?

Issue #4227, opened 2024-04-17
view: extend view service for auctions


erwanor:
We want to amend the view service definition to add an `Auctions` rpc that the transaction planner will be able to use:

```protobuf
service ViewService {
    // ...

    // Gets the auctions controlled by the user's wallet.
    rpc Auctions(AuctionsRequest) returns (stream AuctionsResponse);

    // ...
}

message AuctionsRequest {
    // If present, filter balances to only include the account specified by the `AddressIndex`.
    core.keys.v1.AddressIndex account_filter = 1;
    // If present, include inactive auctions as well as active ones.
    bool include_inactive = 2;
    // If set, query a fullnode for the current state of the auctions.
    bool query_latest_state = 3;
}

message AuctionsResponse {
    core.component.auction.v1.AuctionId id = 1;
    // The note recording the auction NFT.
    SpendableNoteRecord note_record = 4;

    // The state of the returned auction.
    //
    // Only present when `query_latest_state` was provided.
    google.protobuf.Any auction = 2;
    // The state of any DEX positions relevant to the returned auction.
    //
    // Only present when `query_latest_state` was provided.
    // Could be empty, depending on the auction state.
    repeated core.component.dex.v1.Position positions = 3;
}
```

TalDerei:
should we wait to close this until a storage impl is filled into the stub?

erwanor:
@TalDerei  Created #4251 to track impl.

PR #4236, opened 2024-04-18
auction: extend view service 

TalDerei:
## Describe your changes
adds `Auctions` rpc interface that the transaction planner can query.

References https://github.com/penumbra-zone/penumbra/issues/4227


TalDerei:
@erwanor what else should this PR encapsulate if we don't have access to a `self.storage.request_auction_by_id` storage implementation for instance?

erwanor (approved):
@TalDerei We can leave it as a stub for now, let's merge on green CI

Issue #4118, opened 2024-03-27
Add protos for TCT crate structs


grod220:
The TCT crate has a few important structs that do not have protobuf equivalents. The one the web repo uses the most is `Updates` and all of its fields:
https://github.com/penumbra-zone/penumbra/blob/6897901b37f5c72d4b5c593167f0e0b933dd2eb5/crates/crypto/tct/src/storage.rs#L279-L290

Without protobuf equivalents, we don't have types in the web repo for these entities. So we [type them ourselves](https://github.com/penumbra-zone/web/blob/5dbca7b1d0b4aded1d2ba1cc80fcb594467b3de5/packages/types/src/state-commitment-tree.ts#L47-L53) w/ zod. It is a runtime validation and not fully extensive (kinda our best effort). Also, it will not stay in sync if things change with those structs.

### Proposed solution

Add protos and attributes for these (likely needs the impls too?)

```rust
#[serde(into = "pb::TctUpdates", try_from = "pb::TctUpdates")]
pub struct Updates {
...
```

This way, our bufbuild tooling will automatically spit out the Typescript types that other codebases can use for their block processing.

hdevalence:
Do these need to all have protos? So far, we haven't defined a stable data interchange format for the TCT internals on purpose.

What's the benefit of having runtime type validation for the TCT internals? Aren't they going to be produced by the same code that's consuming them?

grod220:
> Do these need to all have protos?

Ideally, there would be some way that non-rust systems consuming these can have native types that represent them and are packaged + versioned. In the past, the bufbuild proto type generation seems to be the canonical route.

> What's the benefit of having runtime type validation for the TCT internals?

- Runtime validation
  - This is more related to web repo specifics. Given TCT does not have types, we have to manually type them all. We could write them in typescript, but we do not get any indication on a breaking change (aka TCT struct field name change). We'll just get a weird error that we have to manually trace. If we type them with zod, we'll receive runtime validation that specifically tells us if the types end up breaking (and where).
- TCT internals
  - I think my main point is that these types are not internal. They are used within our block processor and a number of rpc methods.

> Aren't they going to be produced by the same code that's consuming them?

Yes, at the moment, every non-rust external consumer of TCT types has to manually type them.

PR #4239, opened 2024-04-18
Correct swap claim height check

cronokirby:
## Describe your changes

This changes the circuit check in swap claims to check that the exact epoch and block where a swap was made match that output data for the batch swap of that epoch and block. Previously were effectively only checking that the intra-block value was the same, but not the epoch.

TODO: Write migration for batch swap output datas

  > A migration will be needed to augment batch swap output data


cronokirby:
Not quite sure why the tests aren't passing, will pick back up tomorrow morning.

conorsch:
These changes should be based on `v0.71.0`, so that the fix is compatible with the existing testnet. Because the changes are consensus-breaking, we should bump the minor version in order to release. To that end, I've pushed a `release/v0.72.x` branch, branched from the `v0.71.0`, that should be targeted with this PR.

proto/penumbra/penumbra/core/component/dex/v1/dex.proto
```diff
@@ -227,6 +227,8 @@ message BatchSwapOutputData {
   TradingPair trading_pair = 8;
   // The starting block height of the epoch for which the batch swap data is valid.
   uint64 epoch_starting_height = 9;
+  // The prefix (epoch, block) of the position where this batch swap occurred.
+  uint64 sct_position_prefix = 10;
```
hdevalence, line 231: We should deprecate the old field in the protos and remove it from the domain types, the `sct_position_prefix` is what `epoch_starting_height` should have been and we shouldn't carry the old field forward.

crates/core/component/dex/src/component/router/route_and_fill.rs
```diff
@@ -23,21 +24,14 @@ use super::fill_route::FillError;
...
-    ))]
+    #[instrument(skip(self, trading_pair, batch_data, block_height, epoch, params))]
     async fn handle_batch_swaps(
         self: &mut Arc<Self>,
         trading_pair: TradingPair,
         batch_data: SwapFlow,
         // TODO: why not read these 2 from the state?
         block_height: u64,
-        epoch_starting_height: u64,
+        epoch: Epoch,
```
hdevalence, line 34: This seems dangerous to me -- I realize that the code is already like this, but given the importance of getting the values right, I think it would be good to eliminate these arguments and to instead have the correct values read out of the state right when the BSOD is created.

crates/core/component/dex/src/batch_swap_output_data.rs
```diff
@@ -142,9 +155,9 @@ pub struct BatchSwapOutputDataVar {
     pub lambda_2: U128x128Var,
     pub unfilled_1: U128x128Var,
     pub unfilled_2: U128x128Var,
-    pub height: FqVar,
     pub trading_pair: TradingPairVar,
-    pub epoch_starting_height: FqVar,
+    pub epoch: FqVar,
+    pub block: FqVar,
```
hdevalence: What's the rationale behind making these an `FqVar` rather than an alternate representation?

Also, could we give them different names to clarify that they're not standalone values, but components of a `Position`? Like
```
pub position_epoch_component: FqVar,
pub position_block_component: FqVar,
```
or some other name? I'm just worried that calling it `block` could suggest it's the block height, which it isn't.

cronokirby: Changed to store a PositionVar now

hdevalence: No I meant, what's the rationale for the representation? I'm not opposed to using an FqVar, I'm just wondering how we decide one or the other. Can we write up a brief rationale for why we picked one or the other?

crates/core/component/dex/src/component/router/route_and_fill.rs
cronokirby: Changed to read from the state inside this method.

The block height is read from the block in the end_block handler, so that is left unchanged for now.

crates/core/component/dex/src/batch_swap_output_data.rs
redshiftzero: A `PositionVar` has a position stored internally as both an `FqVar` and a series of 48 bit constraints - 16 bits each for the epoch, block, and commitment segment of the position. Since we need to compare the epoch and block from the position from the state commitment proof with the epoch and block from the BSOD, and we're already bit constraining to make two FqVars for the epoch and block in both cases, another (I think cheaper in constraints) option is to compare as bits directly. Comparing as `FqVar`s also works though.

hdevalence: Hmm, in that case wouldn't making the two prefix components FqVars be cheaper? Since we already have the bit-constrained components of the actual position, we can check that they're equal to the provided prefix components without needing to bit-constrain the prefix components, right?

redshiftzero: yep, you're right. That seems like the way to go to save the additional bit constraints

proto/penumbra/penumbra/core/component/dex/v1/dex.proto
cronokirby: Done

crates/core/component/dex/src/batch_swap_output_data.rs
cronokirby: Not sure that comparing bits is cheaper in constraints: checking that (x0 + 2 * x1 + ...) =? (y0 + 2 * y1 + ...) can be done in a single R1CS constraint. Checking x0 =? y0, x1 =? y1 would require several

hdevalence: Yeah, that seems like a good reason to use FqVars as in the original design, I think it's clearer than using a Position because it means we don't add any instance variables that don't have meaning but could be (mis)used in the circuit.

redshiftzero (approved):
I know there is a migration to do, but the other changes LGTM

crates/core/component/dex/src/component/router/route_and_fill.rs
```diff
@@ -95,16 +88,23 @@ pub trait HandleBatchSwaps: StateWrite + Sized {
...
             trading_pair,
             delta_1,
             delta_2,
             lambda_1,
             lambda_2,
             unfilled_1,
             unfilled_2,
+            sct_position_prefix: (
+                u16::try_from(epoch.index).expect("epoch index should be small enough"),
+                u16::try_from(block_height - epoch.start_height)
```
erwanor, line 103: Can we add a comment about relative block conversion?

cronokirby: Done

conorsch:
Took a stab at testing this in a pairing session with @cronokirby today. A few commits were pushed as part of that exploration. We want to do some more testing before we're ready to ship this.

crates/bin/pd/src/migrate/testnet72.rs
```diff
@@ -0,0 +1,206 @@
...
+    let (migration_duration, post_upgrade_root_hash) = {
+        let start_time = std::time::SystemTime::now();
+        let ctx = Context::new(storage.clone());
+
+        // Translate inside dex storage.
+        translate_dex_storage(ctx.clone(), &mut delta).await?;
+        // Translate inside compact block storage.
+        translate_compact_block_storage(ctx.clone(), &mut delta).await?;
+
+        delta.put_block_height(0u64);
```
hdevalence, line 153: Why do we reset the block height here?

erwanor: This is so we can do an `InitChain` handshake with comet and allow it to initialize its internal view before block sync/proposals. The flow looks like this:

```
pd <-   info query   -- comet
pd --  app height=0  -> comet
pd <- init chain req -- comet
pd -- init chain resp -> comet
```

unfortunately, the CometBft spec seem to have been reorganized again, so it will take me a minute to fish out the link

crates/bin/pd/src/migrate/testnet72.rs
```diff
@@ -0,0 +1,206 @@
...
+async fn translate_compact_block_storage(
+    ctx: Context,
+    delta: &mut StateDelta<Snapshot>,
+) -> anyhow::Result<()> {
+    let mut stream = delta.prefix_raw("compactblock/");
+    while let Some(r) = stream.next().await {
+        let (key, compactblack_bytes): (String, Vec<u8>) = r?;
+        let block = pb::compact_block::v1::CompactBlock::decode(compactblack_bytes.as_slice())?;
+        let block = ctx.translate_compact_block(block).await?;
+        delta.put_proto(key, block);
```
erwanor, line 122: Here the structure is correct, but we need to be reading from/writing to nonverifiable storage (the key looks correct).

crates/bin/pd/src/migrate/testnet72.rs
```diff
@@ -0,0 +1,206 @@
...
+
+/// Translate all of the compact block storage to hold the new BSOD data inside the compact blocks.
+async fn translate_compact_block_storage(
+    ctx: Context,
+    delta: &mut StateDelta<Snapshot>,
+) -> anyhow::Result<()> {
+    let mut stream = delta.nonverifiable_prefix_raw("compactblock/".as_bytes());
+    while let Some(r) = stream.next().await {
+        let (key, compactblack_bytes) = r?;
+        let block = pb::compact_block::v1::CompactBlock::decode(compactblack_bytes.as_slice())?;
```
conorsch, line 120: Nit: s/black/block/ in var names here.

conorsch:
Most recent round of testing on a private fork of the network showed good results. :+1: Let's make sure to squash this patch into a single commit, as we'll need to cherry-pick it back onto main post-release.

crates/bin/pd/src/migrate/testnet72.rs
hdevalence: Interesting. So the idea is that setting it to `0` signals that the app is ready for genesis, then once `pd` sends the init chain message, it will get set to `h+1` as set in the genesis file?

erwanor: That's right, it will consume the genesis as a result of reading a zero app height, and then pull the one specified in the file (`initialHeight`)

erwanor (approved):
ACK, thanks for your work on this @cronokirby @conorsch

crates/core/component/dex/src/component/router/route_and_fill.rs
grod220: I'd like to re-pitch adding tct protos: https://github.com/penumbra-zone/penumbra/issues/4118

so other systems like the web code doesn't have to manually re-create this serialization logic happening in the .into()'s

PR #4246, opened 2024-04-19
feat(app): 🎋 add missing `Sct` component hooks

### LINKED ISSUE #3739: app: 🎋 missing `Sct` component hooks
while reading through the `App` implementation, i noticed that the `Sct` component only ever has its `init_chain` and `begin_block` hooks called.

this isn't currently a bug, since its `end_block` and `end_epoch` hooks are no-ops, but could become one in the future if more logic was introduced to those trait methods.

this is an easy fix, but i'm filing this separately as an issue for reference in a tracking issue elsewhere.

cratelyn:
fixes #3739.

the `App` only ever invokes the Sct component's `init_chain` and `begin_block` hooks.

this isn't currently a bug, since its `end_block` and `end_epoch` hooks are noöps, but could easily lead to a surprising outcome in the future, if more logic was introduced to those trait methods.

  > while this **does** change the control flow of consensus-critical logic, this does **not**
 change the execution of incoming data. the hooks introduced here are noöps.

### DIFF
--- a/crates/core/app/src/app/mod.rs
+++ b/crates/core/app/src/app/mod.rs
@@ -382,6 +382,7 @@ impl App {
 
         tracing::debug!("running app components' `end_block` hooks");
         let mut arc_state_tx = Arc::new(state_tx);
+        Sct::end_block(&mut arc_state_tx, end_block).await;
         ShieldedPool::end_block(&mut arc_state_tx, end_block).await;
         Distributions::end_block(&mut arc_state_tx, end_block).await;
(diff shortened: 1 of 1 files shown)


cratelyn:
https://github.com/penumbra-zone/penumbra/blob/c5e30cf93217e664ed99aafe2b476f590fa7602c/crates/core/component/sct/src/component/sct.rs#L59-L69

erwanor (approved):

PR #4247, opened 2024-04-19
feat(app): 🎣 add missing `Dex` component hook

cratelyn:
see #3739, and #4246.

while wiring up a pair of missing component hooks for the `Sct` component, i found another (noöp) hook that we do not include in the `App`'s control flow.

this isn't currently a bug, since `<Dex as Component>::begin_block()` is a noöps but could easily lead to a surprising outcome in the future, if more logic was introduced to that trait method.

  > while this **does** change the control flow of consensus-critical logic, this does **not**
 change the execution of incoming data. the hooks introduced here are noöps.

### DIFF
--- a/crates/core/app/src/app/mod.rs
+++ b/crates/core/app/src/app/mod.rs
@@ -257,6 +257,7 @@ impl App {
             begin_block,
         )
         .await;
+        Dex::begin_block(&mut arc_state_tx, begin_block).await;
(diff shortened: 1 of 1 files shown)


cratelyn:
https://github.com/penumbra-zone/penumbra/blob/3b62d64b60182cb8ba07ae2d478c5f0c2ab6369a/crates/core/component/dex/src/component/dex.rs#L42-L47

for reference:

erwanor (approved):

Issue #4242, opened 2024-04-19
pcli: support for auction interactions


erwanor:
We should add support for, the work can be divided in two phases:

Phase 1: add CLI scaffolding for:
- [x] `pcli tx auction dutch schedule`
- [x] `pcli tx auction dutch end`
- [x] `pcli tx auction dutch withdraw`

Phase 2:  fill in the specific implementations:
- [x] `pcli tx auction dutch schedule`
- [x] `pcli tx auction dutch end`
- [x] `pcli tx auction dutch withdraw`

The item to prioritize is `tx auction dutch schedule`, so that we can work towards viewing capabilities.

TalDerei:
pcli implementations are currently only stubbed with planner-related boilerplate.

TalDerei:
phase 1 is tracked by https://github.com/penumbra-zone/penumbra/pull/4249. phase 2 will be completed in a separate PR.

PR #4248, opened 2024-04-19
auction(planner): add actions to planner

TalDerei:
## Describe your changes
Add planner support for auctions

## Issue ticket number and link
References https://github.com/penumbra-zone/penumbra/issues/4241

## Checklist before requesting a review

- [x] If this code contains consensus-breaking changes, I have added the "consensus-breaking" label. Otherwise, I declare my belief that there are not consensus-breaking changes, for the following reason:

Issue #4241, opened 2024-04-19
view: add auction support to planner / planner rpc


erwanor:

TalDerei:
there's an on-going planner refactor in https://github.com/penumbra-zone/penumbra/issues/4081 that will need to fold these changes. I think this issue should scope adding the relevant planner functions, and handle the specific planner logic separately once the blocking PR is merged? cc @erwanor

TalDerei:
can this be closed? or is there anything else this issue is intended to capture

PR #4249, opened 2024-04-19
auction(pcli): auction support in pcli

TalDerei:
## Describe your changes
Auction support in pcli

References https://github.com/penumbra-zone/penumbra/issues/4242


crates/bin/pcli/src/command/tx.rs
```diff
@@ -316,6 +367,10 @@ impl TxCmd {
             TxCmd::CommunityPoolDeposit { .. } => false,
             TxCmd::Position(lp_cmd) => lp_cmd.offline(),
             TxCmd::Withdraw { .. } => false,
+            // TODO: check this.
```
erwanor, line 370: 👍

crates/bin/pcli/src/command/tx.rs
```diff
@@ -218,15 +215,69 @@ pub enum TxCmd {
...
+    DutchAuctionSchedule {
+        /// Only spend funds originally received by the given account.
+        #[clap(long, default_value = "0", display_order = 100)]
+        source: u32,
+        /// The amount the seller wishes to auction.
+        #[clap(long, default_value = "0", display_order = 200)]
+        amount: String,
+        /// The denomination of the target asset the seller wishes to acquire.
+        #[clap(long, display_order = 300)]
+        recieve: String,
```
erwanor, line 236:
```suggestion
        receive: String,
```

crates/bin/pcli/src/command/tx.rs
```diff
@@ -1280,6 +1335,105 @@ impl TxCmd {
...
+                auction_id: _,
+                source,
+                fee_tier,
+            } => {
+                let mut planner = Planner::new(OsRng);
+                planner
+                    .set_gas_prices(gas_prices)
+                    .set_fee_tier((*fee_tier).into());
+
+                // TODO: add `ActionDutchAuctionEnd` to planner (depends on #4248).
```
erwanor, line 1424: FWIW, we can do away with the boilerplate and make those branches `todo!()`, your call.

TalDerei:
this will additionally encapsulate necessary changes to the pcli docs.

crates/bin/pcli/src/command/tx.rs
```diff
@@ -1280,6 +1335,105 @@ impl TxCmd {
...
                 replicate_cmd.exec(app).await?;
             }
+            TxCmd::DutchAuctionSchedule {
+                source,
+                amount,
+                recieve: _,
+                auction_duration,
+                starting_price,
+                reserve_price,
+                fee_tier,
```
erwanor, line 1345: The goal of these commands is to be able to exercise the DA logic, IMO it would be clearer to have them take the exact fields that an auction description specifies i.e.

- a start height
- an end height
- a step count
- an input value
- a minimum output
- a maximum output

ditto for the rest of the commands. We are the customers for this command so being able to easily specify custom descriptions is a feature

crates/bin/pcli/src/command/tx.rs
```diff
@@ -1280,6 +1335,105 @@ impl TxCmd {
...
+                // Use the specified time interval to calculate the auction's `start_height` and `end_height`,
+                // implicitely defining the speed of the auction.
+                let view: &mut dyn ViewClient = app
+                    .view
+                    .as_mut()
+                    .context("view service must be initialized")?;
+                let start_height = view.status().await?.full_sync_height;
+
+                // TODO: make this a valid calculation.
+                let end_height = start_height + auction_duration;
```
erwanor, line 1358: Let's make this a flag `--end-height`

crates/bin/pcli/src/command/tx.rs
```diff
@@ -218,15 +215,69 @@ pub enum TxCmd {
...
+        #[clap(short, long, value_enum, default_value_t, display_order = 300)]
+        fee_tier: FeeTier,
+    },
+    /// Ends a Dutch auction.
+    #[clap(display_order = 800)]
+    DutchAuctionEnd {
+        /// Identifier of the auction.
+        #[clap(long, default_value = "0", display_order = 100)]
+        auction_id: u32,
+        /// Source address withdrawing from auction.
```
erwanor, line 274: minor point:
```suggestion
        /// Source address withdrawing from the auction.
```

TalDerei:
flagging this workstream also has a dependency on having planner support https://github.com/penumbra-zone/penumbra/issues/4241

crates/bin/pcli/src/command/tx.rs
```diff
@@ -218,15 +215,69 @@ pub enum TxCmd {
...
         #[clap(long, default_value = "0", display_order = 200)]
         source: u32,
-
         /// The selected fee tier to multiply the fee amount by.
         #[clap(short, long, value_enum, default_value_t)]
         fee_tier: FeeTier,
     },
+    /// Schedule a Dutch auction, a tool to help accomplish price discovery.
+    #[clap(display_order = 600)]
+    DutchAuctionSchedule {
```
erwanor, line 227: Could we refactor the `DutchAuction*` variants into an `AuctionCmd` enum so that we can have these commands:

```rust
pcli auction dutch schedule ...
```

rather than:

`pcli dutch-auction-schedule`

🙏

erwanor (changes requested):
Thank you for the work you have put in this, I added a few comments that I think will make testing the core mechanism easier for us. I think at this point we just need to get the basic interface right, so if you want to replace the command impls with `todo!()` that would be ok.

crates/bin/pcli/src/command/tx.rs
TalDerei: fixed typo

crates/bin/pcli/src/command/tx.rs
TalDerei: what if we merge #4248 first?

crates/bin/pcli/src/command/tx.rs
TalDerei: ah good call, I was thinking of the customer in the context of the end user rather than us. It makes sense since we're still in dev mode. I'll make the necessary changes

crates/bin/pcli/src/command/tx.rs
TalDerei: fixed

crates/bin/pcli/src/command/tx.rs
TalDerei: I'll make the change. I originally did that, but then reverted since I noticed the issue description used `pcli tx auction`

crates/bin/pcli/src/command/tx.rs
TalDerei: I think the only boilerplate we should keep is the planner related code since we'll be passing all the necessary CLI parameters anyways.

crates/bin/pcli/src/command/tx.rs
erwanor: sorry I missed the tx in the command but we definitely want that

erwanor: the ask here is folding the auction commands into an AuctionCmd enum

crates/bin/pcli/src/command/tx.rs
erwanor: Sounds good

crates/bin/pcli/src/command/tx.rs
erwanor: so just to clarify: pcli tx auction dutch ...

TalDerei: yeah noted, was just clarifying!

crates/bin/pcli/src/command/tx.rs
TalDerei: modified fields to match design doc

crates/bin/pcli/src/command/tx.rs
TalDerei: @erwanor there seems to be something wrong with the clap ordering, can you check it out?

erwanor: @TalDerei i'm afk, but can take a look later, what's the problem exactly?

TalDerei: when we run `pcli transaction auction dutch...`, the commands don't display the proper flags

TalDerei: I figured it out, using clap's `default_value = "0"` prevents the flag from being displayed in the CLI.

erwanor:
LGTM, when we fill in the implementations we will want to ditch the nonce flag and use `OsRng + getrandom` to create and fill the buffer on behalf of the user. Let's do this when we get to it though!

erwanor (approved):

PR #4250, opened 2024-04-19
sct(component): Add `anchor_by_height` rpc method

### LINKED ISSUE #4153: sct(component): add a method to query anchor roots by height
Multiple downstream implementation broke when we updated the SCT anchor state key in #3723.
The problem is that the SCT query service does not offer a way to safely query the anchors, forcing consumers to perform raw KV queries against the chain state.

We have stabilized the state key schemas so this is not as urgent, nevertheless, we should avoid this and fix this gap in the RPC APIs. 

Add a `get_sct_anchor_by_height` method to the `SctQueryService`

**Additional context**
Both the extension and `relayertest` broke as a result of this change.

Prospective contributors: I am available to help you ship this, just holler at me here or on discord [@erwanor](https://discord.com/channels/824484045370818580/894628521429839983/1225106264074354774).

phasewalk1:
## Describe your changes
This PR adds a `anchor_by_height` method to the [`SctQueryService`](https://github.com/penumbra-zone/penumbra/tree/main/crates/core/component/sct/src/component/rpc.rs).

Closes #4153

  > This is an added rpc method for consumer convenience. It prevents consumers from being forced to perform raw KV queries against the chain state.


crates/core/component/sct/src/component/rpc.rs
```diff
@@ -35,4 +38,21 @@ impl QueryService for Server {
...
+    async fn anchor_by_height(
+        &self,
+        request: tonic::Request<AnchorByHeightRequest>,
+    ) -> Result<tonic::Response<AnchorByHeightResponse>, Status> {
+        let state = self.storage.latest_snapshot();
+
+        let anchor = state
+            .get_anchor_by_height(request.get_ref().height)
+            .await
+            .map_err(|e| tonic::Status::unknown(format!("could not get anchor for height: {e}")))?;
```
erwanor, line 52: minor point but this will display the error message, not the height, good idea to include both

phasewalk1: agreed. so something like this?
```rust
let height = request.get_ref().height;
let anchor = state
    .get_anchor_by_height(height)
    .await
    .map_err(|e| { tonic::Status::unknown(format!("could not get anchor for height {height}: {e}"))})?;
```

erwanor: that would be great 🙏

phasewalk1: do you prefer --fixup commits or should I add the changes separately?

erwanor: I'll squash so whatever works for you, adding the changes separately sounds good

phasewalk1:
rebasing this to fix conflicts with the proto descriptor

Issue #4251, opened 2024-04-20
view: implement auction support


erwanor:
**Is your feature request related to a problem? Please describe.**
Once #4227 is completed, we will be able to implement support for auctions in the Rust view server:

1. generate metadata for each `auctionnft_`
2. add a sqlite table to track auction state
3. use this to the `AuctionsRequest` rpc

We can model this implementation after the `Position*` actions logic in the view server, and corresponding `LpNft` handling.

PR #4252, opened 2024-04-22
view: add support for auctions

TalDerei:
## Describe your changes
View server implementation for auction.

References https://github.com/penumbra-zone/penumbra/issues/4251


crates/view/src/worker.rs
```diff
@@ -303,21 +303,72 @@ impl Worker {
                                 // Update the position record
                                 self.storage.update_position(position_id, state).await?;
                             }
+                            penumbra_transaction::Action::ActionDutchAuctionSchedule(
```
TalDerei, line 306: why do we need these action checks?

TalDerei:
I reviewed this earlier this morning on an outdated change-set, but haven't dived deeply into the `storage.rs` module.

crates/view/src/worker.rs
erwanor: On advice, we scan transactions and their actions, and record auction metadata by their asset id. This is useful to be able to display relevant asset metadata when presented with an asset id (among other things).

erwanor:
I this this is ready for review, the discoverability / rendering of the commands needs to improve, but seem good enough for an mvp. 

A few lines about the state modeling in the view server, this PR:

1. Adds an `auctions` table in the rust view server:

```sql
-- This table records the user's own auction state, using the
-- auction id as a primary key. An extra-column is available
-- to cross-reference note commitments that is associated with
-- the entry.
CREATE TABLE auctions (
     auction_id             BLOB PRIMARY KEY NOT NULL,
     auction_state          BIGINT NOT NULL,
     note_commitment        BLOB
);
```

2. The sync worker will record auction metadata as part of its block scanning routine (itself contingent on detecting activity)

3. As part of recording SNR asset data, it will update the auctions view state with an associated note commitment

4. This note commitment is used to cross reference spendable notes and cleartext notes from the `spendable_notes` and `notes` tables

erwanor:
I have tested that I am able to:
- query my auction state
- go through the lifecycle of a DA: scheduling, terminating, and withdrawing

crates/view/src/worker.rs
hdevalence: For a little bit more context: `Value`s in the shielded pool are an `Amount` and an `AssetId`. The asset ID (an $\mathbb F_q$ value) is the currently always a hash of a denom string, and client tooling looks up `asset.Metadata` via the asset ID in an asset registry.

The chain maintains one minimal asset registry (just containing denoms), but this registry does _not_ include entries for all of the LPNFTs (and now ANFTs), because this would bloat the chain state. Instead, the client's view service is responsible for locally deriving and caching the metadata of the bearer tokens it creates. This saves data for all the other clients, who don't want to have to maintain knowledge of everyone else's position NFTs.

erwanor (approved):
Optimistically merging this, async reviews welcome

Issue #4240, opened 2024-04-19
auction: `end_block`implementation


erwanor:

erwanor:
Done in #4254

PR #4254, opened 2024-04-22
auction: process triggers/execute DAs

erwanor:
## Describe your changes

This PR implements #4240, in particular it adds the "end block" DA auction logic, and:
- extend the auction with a `HandleDutchTriggers` trait, responsible for streaming auction state recorded for execution at a `trigger_height`
- implement `execute_dutch_auction`, which encapsulate auction LP logic and state updates
- add price interpolation logic
- create `penumbra_dex::lp::Position` to take a user specified random nonce (computed based off the auction description, step index, and a counter, in order to evade id-squatting) 
- encapsulate `TriggerData` into its own submodule (DA step function utils)

  > This is consensus breaking


hdevalence:
> create penumbra_dex::lp::Position to take a user specified random nonce (specified in the auction description)

Interesting find, I didn't think about the position nonces during the design. Can we clarify this a little bit?

One case we should think through is what happens if the position ID is already used. Since the auction data is public, any position ID that the chain can deterministically derive from it can also be predicted by an external party. Someone could attempt to claim the position ID first and grief the auction that way.

1. What happens in the current code if the position ID is already used? If this triggers an error, that's a potential DoS vector and should be avoided.

2. Can we create position IDs in some other way that would be robust to position-claiming attacks? For instance, what if the nonce for the position was something like `position_nonce = H( auction_nonce || step_number || counter )`, starting with `counter = 0` and incrementing `counter` until the resulting `position_id` was unused? Then creating the position would always be infallible, and the fact that it's infallible removes the incentive to grief (so `counter = 0` would ~always work in practice).

erwanor:
@hdevalence Thanks for the suggestion, I implemented it if you want to take a look - the current logic has an ugly failure mode because the position manager returns coarse errors (i.e. it's not possible to distinguish between a duplicate position id and an incoherent position state). I will fix it myself as part of M1 so that we only retry opening if the position id is squatted, and either terminate the auction, or bubble up the error, if a more serious error occurs.

hdevalence:
I think as a first pass we could ignore the error details and retry unconditionally

erwanor:
@hdevalence could you give this a +1?

hdevalence:
Will do as soon as I'm back at the computer

crates/core/component/auction/src/component/dutch_auction.rs
```diff
@@ -51,11 +64,183 @@ pub(crate) trait DutchAuctionManager: StateWrite {
...
+        let balance_input_asset = lp_reserves
+            .provided()
+            .filter(|v| v.asset_id == input.asset_id)
+            .map(|v| v.amount)
+            .sum::<Amount>();
+        let balance_output_asset = lp_reserves
+            .provided()
+            .filter(|v| v.asset_id == output_id)
+            .map(|v| v.amount)
+            .sum::<Amount>();
```
hdevalence, line 163: This feels a bit dangerous. If for some reason the asset IDs were mismatched, we wouldn't notice and we'd just pass through a zero, right? Or do we have to do that because of the way `Balance` works (does it record zero entries?)

crates/core/component/auction/src/component/dutch_auction.rs
```diff
@@ -51,11 +64,183 @@ pub(crate) trait DutchAuctionManager: StateWrite {
...
+
+        new_dutch_auction.state.current_position = None;
+        // We overwrite the input amount
+        new_dutch_auction.state.input_reserves = balance_input_asset;
+        new_dutch_auction.state.output_reserves += balance_output_asset;
```
hdevalence, line 168: Why is one of these overwritten but the other added? What happens to the value we destroy by overwriting?

crates/core/component/auction/src/component/dutch_auction.rs
```diff
@@ -51,11 +64,183 @@ pub(crate) trait DutchAuctionManager: StateWrite {
...
+            // position_nonce = H(auction_nonce || step_index || attempt_counter)
+            // until the resulting position id (based on the nonce) is unique and accepted
+            // by the DEX.
+            let mut attempt_counter = 0u64;
+            loop {
+                let pair = DirectedTradingPair::new(auction_input_id, auction_output_id);
+                let lp_reserves = Reserves {
+                    r1: new_dutch_auction.state.input_reserves,
+                    r2: Amount::zero(),
+                };
```
hdevalence, line 206: This seems to be double-counting the tokens. Either they're in the LP position (in which case they should not be in the auction) or they're in the auction object (in which they should not be in the LP reserves).

erwanor: Oh yeah indeed, this is wrong

hdevalence: xref https://github.com/penumbra-zone/penumbra/pull/4254/files#r1575464539

I think that the input reserves of the auction should be zero if the tokens are in an opened LP position

crates/core/component/auction/src/component/dutch_auction.rs
```diff
@@ -51,11 +64,183 @@ pub(crate) trait DutchAuctionManager: StateWrite {
...
+            // The nonce must be chosen so that the resulting position id is unique:
+            // `PositionManager::open_position` will reject duplicates.
+            //
+            // To do this, we keep track of our number of attempts at opening a position,
+            // and compute:
+            // position_nonce = H(auction_nonce || step_index || attempt_counter)
+            // until the resulting position id (based on the nonce) is unique and accepted
+            // by the DEX.
+            let mut attempt_counter = 0u64;
+            loop {
```
hdevalence, line 201: structure suggestion: could this be broken out into an `allocate_position` method that returns the position?

crates/core/component/auction/src/component/dutch_auction.rs
```diff
@@ -51,11 +64,183 @@ pub(crate) trait DutchAuctionManager: StateWrite {
...
+                match self.open_position(lp).await {
+                    Ok(_) => break,
+                    Err(e) => {
+                        // TODO(erwan): this coarsely assumes that all errors
+                        // are id collisions which obviously does not hold.
+                        // for now it's impossible to do finer grained control flow.
+                        tracing::error!(?e, attempt_counter, ?id, "failed to open position");
+                        attempt_counter += 1;
+                    }
+                }
```
hdevalence, line 233: Why don't we just check that the position ID is unused before opening it? Then we can be precise about our control flow, at the cost of duplicating a single state read (on average one time).

hdevalence: Note: we'll need to make sure that the control flow doesn't leak value, if we error creating a position for whatever reason we should try to leave the state consistent (put value back into the auction object or something), or panic

crates/core/component/auction/src/component/dutch_auction.rs
```diff
@@ -184,6 +367,58 @@ pub(crate) trait DutchAuctionManager: StateWrite {
...
+pub(crate) trait HandleDutchTriggers: StateWrite {
+    /// Process the trigger height for a [`DutchAuction`],
+    async fn process_trigger(&mut self, trigger_height: u64) -> Result<()> {
+        use futures::StreamExt;
+        // TODO(erwan): somewhat self-defeating, but this is a first pass.
+        let auction_ids: Vec<AuctionId> = self
+            .stream_dutch_ids_by_trigger(trigger_height)
+            .await
+            .collect()
+            .await;
```
hdevalence, line 380: Hmm, assuming the self-defeating is about getting all the IDs and then processing them in sequence, I don't think this is avoidable, we don't allow concurrent updates to the DEX...

crates/core/component/auction/src/component/dutch_auction.rs
```diff
@@ -184,6 +367,58 @@ pub(crate) trait DutchAuctionManager: StateWrite {
 
 impl<T: StateWrite + ?Sized> DutchAuctionManager for T {}
 
+#[async_trait]
+pub(crate) trait HandleDutchTriggers: StateWrite {
+    /// Process the trigger height for a [`DutchAuction`],
+    async fn process_trigger(&mut self, trigger_height: u64) -> Result<()> {
```
hdevalence, line 373: should this be `process_triggers` (plural)?

crates/core/component/auction/src/component/dutch_auction.rs
```diff
@@ -184,6 +367,58 @@ pub(crate) trait DutchAuctionManager: StateWrite {
...
+        let auction_ids: Vec<AuctionId> = self
+            .stream_dutch_ids_by_trigger(trigger_height)
+            .await
+            .collect()
+            .await;
+
+        for auction_id in auction_ids.into_iter() {
+            let _ = self
+                .execute_dutch_auction(auction_id, trigger_height)
+                .await?;
```
hdevalence, line 385: this looks like we're discarding a result but on closer inspection it seems like `_ = ()`? if so could we drop the let _ = ... so that we don't mislead a reader into thinking that that's the case, or accidentally drop an error we care about in the future?

crates/core/component/auction/src/component/dutch_auction.rs
```diff
@@ -202,155 +437,43 @@ trait Inner: StateWrite {
...
-        let data = TriggerData {
-            start_height: 100,
-            end_height: 200,
-            step_count: 5,
-            current_height: 121,
-        };
-        assert_eq!(compute_next_trigger(data).unwrap(), Some(140));
-    }
+/// TODO(erwan): prove that this works. More work needed to work out validation
+/// rules that let us operate within the 112 bits constraint.
```
hdevalence, line 460: it feels like this should be something we can check upfront when we parse the auction description for the first time?

crates/core/component/auction/src/component/dutch_auction.rs
erwanor: So, the auction state has a bunch of reserves $\Delta, \Lambda$, at each price point $p_1, p_2... p_i$ it auctions off a portion of that input, when we roll out the position (close -> open) we deduct the portion that has been sold (`balance_input_asset`) and accumulate what the LP has been able to trade:

```rust
        new_dutch_auction.state.input_reserves = balance_input_asset;
        new_dutch_auction.state.output_reserves += balance_output_asset;
```

By the moment we build those reserves, we should have closed and withdrawn the LP and credited the auction state with the balance from that position, what am I missing?

crates/core/component/auction/src/component/dutch_auction.rs
erwanor: It can be, in my mind this was "process a trigger height", but it can be read as "process dutch triggers"

crates/core/component/auction/src/component/dutch_auction.rs
erwanor: Yes for sure, basically the same validation rules for position apply for a DA description.

crates/core/component/auction/src/component/dutch_auction.rs
erwanor: We're getting this `Balance` from withdrawing the LP, but would adding a check that:
- there are no negative entries
- there are all from the correct asset ids
alleviate the concern here?

crates/core/component/auction/src/component/dutch_auction.rs
erwanor: https://github.com/penumbra-zone/penumbra/pull/4254#discussion_r1575475646

crates/core/component/auction/src/component/dutch_auction.rs
hdevalence: if we require `input_amount`, `output_amount` < 52 bits and `step_count < 256` then we are guaranteed that `input_amount * output_amount * step_count` < 112 bits.

erwanor: I put a TODO to think the case of large denom units (e.g. $10^{18} ~ 60 bits$)

crates/core/component/auction/src/component/dutch_auction.rs
```diff
@@ -202,155 +437,43 @@ trait Inner: StateWrite {
...
+
+    let q = auction_description
+        .input
+        .amount
+        .checked_mul(&step_count)
+        // TODO(erwan): this isn't true, this must part of 112 bits extra validation pass todo.
+        .expect("checked in validation");
+
+    let p = (step_count - step_index) * auction_description.max_output
+        + step_index * auction_description.min_output;
```
hdevalence, line 476: I think this could be a little clearer if it were factored like
```
// The target output, scaled up by `step_count` to avoid divisions.
// Linearly interpolate between `max_output` at `step_index = 0`
//                          and `min_output` at `step_index = step_count`.
let target_output_scaled = (step_count - step_index) * max_output + step_index * min_output;
// The input, scaled up by `step_count` to match.
let input_scaled = step_count * input.amount;

// The trading function interpolates between (input, 0) and (0, target_output)
let p = target_output_scaled;
let q = target_input_scaled;
```

hdevalence: Also doesn't this mean we need to iterate up to and INCLUDING step_count?

erwanor: Thanks, sg

crates/core/component/auction/src/component/dutch_auction.rs
```diff
@@ -51,11 +64,183 @@ pub(crate) trait DutchAuctionManager: StateWrite {
...
+
+        // Compute the current step index, between 0 and `step_count`.
+        let step_index = auction_trigger
+            .compute_step_index(trigger_height)
+            .expect("trigger data is validated");
+
+        if step_index >= step_count {
+            // If the termination condition has been reached, we set the auction
+            // sequence to 1 (Closed).
+            new_dutch_auction.state.sequence = 1;
```
hdevalence, line 179: Not sure this termination condition is correct, I think it won't ever hit the reserve price, should it be `step_index > step_count`?

crates/core/component/auction/src/component/dutch_auction.rs
hdevalence: we could exclude those from auctions, seems fine

crates/core/component/auction/src/component/dutch_auction.rs
erwanor: That's right it should be strictly greater

crates/core/component/auction/src/component/dutch_auction.rs
hdevalence: I'm not sure it's an important concern, if the `Balance` isn't creating zero entries it may not have entries for our assets anyways.

crates/core/component/auction/src/component/dutch_auction.rs
erwanor: Erratum, it auctions *all* of that input but only a portion gets sold. The unfilled part gets returned.

hdevalence: The auction object can own tokens in two ways: directly, in its own structure, or indirectly, via the LP position.

On line 118, value is moved out of the LP position (if it exists) and onto the stack (in `lp_reserves`, destructured to `balance_input_asset` and `balance_output_asset`.

On line 168, `balance_output_asset` is consumed by accumulating it into `state.output_reserves`. I agree this is correct.

On line 167, any previous value of `state.input_reserves` is burned, and `balance_input_asset` is consumed by overwriting it. Then on line 204, the value is duplicated into the position if one is created. This would be an inflation, except that any future pass through _this method only_ would happen to burn one of the copies when it re-executes line 167. However, this is not the only possibility -- if the auction is closed, that runs a different codepath, which accumulates the position reserves together with the auction's reserves, on line 313. I believe this would trigger an inflation bug.

Instead, we should avoid this problem by ensuring that we don't duplicate value. The auction's `input_reserves` should be zeroed out when the value moves into an LP position and restored when value is returned.

erwanor: Got it, makes sense, thanks

TalDerei:
currently out for holiday, but will have a chance (and I want to for my own benefit) to review this tonight.

crates/core/component/auction/src/component/dutch_auction.rs
erwanor: I just thought about it, and I think we need to change the interpolation so that:
- the minimum price is hit at the penultimate trigger: `step_count - 1`
- the auction is closed at the ultimate trigger: `step_count` (= `end_height`)

This means an auction has at least 2 steps, that seems like a reasonable tradeoff and already baked in with the idea of a `min` and `max` outputs (two points).

crates/core/component/auction/src/component/dutch_auction.rs
erwanor: Ended up leaving it as is and changed the price interpolation logic so that we reach the minimum at `step_count - 1` (x-ref:  https://github.com/penumbra-zone/penumbra/pull/4254#discussion_r1575591551). I think this makes more sense to have the auction terminate at the specified `end_height`.

erwanor:
I did a round of improvements, thank you for the engaged review and feedback, I think this PR is much better as a result of your attention. In particular, I want to call out the following changes:
1. we systematically zero out the action input state as its deployed in a PCLP
2. the answer to value leakage on failure/state incoherence is that it doesn't happen because we panic (for now), and we will tackle fallibility in a dedicated value breaker effort.
3. we interpolate to a minimum price at `step_count - 1`
4. we add a termination condition clause if the input reserves are exhausted

crates/core/component/auction/src/auction/dutch/actions/schedule.rs
```diff
@@ -5,6 +5,8 @@ use penumbra_proto::{core::component::auction::v1alpha1 as pb, DomainType};
 use penumbra_txhash::{EffectHash, EffectingData};
 use serde::{Deserialize, Serialize};
 
+pub const MAX_AUCTION_AMOUNT_RESERVES: u128 = (1 << 52) - 1;
```
TalDerei, line 8: can you add a comment clarifying the design choice behind the precision?

TalDerei (approved):

crates/core/component/auction/src/auction/dutch/actions/schedule.rs
erwanor: good idea

crates/core/component/auction/src/component/action_handler/dutch/schedule.rs
```diff
@@ -50,16 +79,17 @@ impl ActionHandler for ActionDutchAuctionSchedule {
...
+            step_count >= 2,
+            "step count MUST be at least two (got: {step_count})"
         );
 
-        // Check that the step count is less than 1000.
+        // Check that the step count is less than 255.
         ensure!(
-            step_count <= 1000,
-            "the dutch auction step count MUST be less than 1000 (got: {step_count})",
+            step_count <= 255,
```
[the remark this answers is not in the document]
erwanor: this is related to the limit above, we don't gain much by having a larger step count capacity since the end game is to have a collection of DAs anyway, but this costs us precious precision space when we perform the price interpolation

crates/core/component/auction/src/component/dutch_auction.rs
```diff
@@ -51,11 +63,147 @@ pub(crate) trait DutchAuctionManager: StateWrite {
...
+        let balance_output_asset = lp_reserves
+            .provided()
+            .filter(|v| v.asset_id == output_id)
+            .map(|v| v.amount)
+            .sum::<Amount>();
+
+        // After consuming the LP, we reset the state, getting ready to either
+        // execute another session, or retire the auction.
+        new_dutch_auction.state.current_position = None;
+        new_dutch_auction.state.input_reserves = balance_input_asset;
```
hdevalence, line 167: Shouldn't this also be `+=`? There's no situation where we want to burn existing input reserves, right? So this code is only correct if `input_reserves` is already zero, which it probably will be but might not be.

If we change it to `+=` I think the code will be locally correct without having to do global reasoning.

crates/core/component/auction/src/component/dutch_auction.rs
```diff
@@ -51,11 +63,147 @@ pub(crate) trait DutchAuctionManager: StateWrite {
...
+        new_dutch_auction.state.output_reserves += balance_output_asset;
+        new_dutch_auction.state.next_trigger = 0;
```
hdevalence, line 169: This is a little confusing, I think it would be cleaner to model the data as `next_trigger: Option<NonZeroU64>` so that we can have `Some(height)` or `None` in our code and we don't have to think about the 0 special case (even though `Option<NonZeroU64>` will specialize to exactly the same representation).

crates/core/component/auction/src/component/dutch_auction.rs
```diff
@@ -51,11 +63,147 @@ pub(crate) trait DutchAuctionManager: StateWrite {
...
+        let balance_input_asset = lp_reserves
+            .provided()
+            .filter(|v| v.asset_id == input.asset_id)
+            .map(|v| v.amount)
+            .sum::<Amount>();
+        let balance_output_asset = lp_reserves
+            .provided()
+            .filter(|v| v.asset_id == output_id)
+            .map(|v| v.amount)
+            .sum::<Amount>();
```
hdevalence, line 162: Rather than defining these as new variables, if the expressions are folded into the update below, like
```rust
new_dutch_auction.state.input_reserves += lp_reserves.provided()....
```
then we're protected against accidental misuse of the value stored in the variable (it won't be accessible outside of the expression updating the reserves)

crates/core/component/auction/src/component/dutch_auction.rs
erwanor: FWIW I think it's less confusing that way because doing a cumulative sum suggests that there's some remainder reserves in the auction state, but it ought to never be the case (as you pointed out), but will amend to `+=`

erwanor:
@hdevalence this is ready for another round of review

crates/core/component/auction/src/component/dutch_auction.rs
```diff
@@ -184,7 +322,113 @@ pub(crate) trait DutchAuctionManager: StateWrite {
...
+                .update(&auction_nonce)
+                .update(&step_index.to_le_bytes())
+                .update(&attempt_counter.to_le_bytes())
+                .finalize();
+            let mut tough_nonce = [0u8; 32];
+            tough_nonce[0..32].copy_from_slice(&full_hash.as_bytes()[0..32]);
+
+            let mut lp =
+                Position::new_with_nonce(tough_nonce, pair.clone(), 0u32, p, q, lp_reserves);
+            lp.close_on_fill = true;
```
hdevalence, line 413: nit: it'd be nice to have a comment explaining why this is important

crates/core/component/auction/src/component/trigger_data.rs
```diff
@@ -0,0 +1,156 @@
...
+    pub end_height: u64,
+    pub step_count: u64,
+}
+
+impl TriggerData {
+    /// Compute the step index of the supplied height.
+    ///
+    /// # Errors
+    /// This method errors if the block interval is not a multiple of the
+    /// specified `step_count`, or if it operates over an invalid block
```
hdevalence, line 16: The method doesn't actually error in this case though right? The `step_size` will be rounded down. I'm not sure it matters here but it doesn't quite match the description.

crates/core/component/auction/src/component/trigger_data.rs
```diff
@@ -0,0 +1,156 @@
...
+            start_height,
+            end_height,
+            step_count,
+        } = self;
+
+        let block_interval = end_height - *start_height;
+
+        // Compute the step size, based on the block interval and the number of
+        // discrete steps the auction specifies.
+        let step_size = block_interval / *step_count;
```
hdevalence, line 62: Why is it OK to assume that `step_count != 0` here but not above? The behavior should be consistent, right? Maybe we should make `TriggerData` have private fields and enforce the invariants we want on construction?

hdevalence (approved):
This looks great! My comments are non-blocking, and could be addressed in follow-up PRs if we want to get to them.

erwanor:
Thanks a lot for the review, I think this is mergeable but I agree with your comments about `TriggerData`. I think we want:
1. Make `TriggerData` inner field private
2. Enforce parameter validation via a constructor
3. Use that constructor in the action handlers
4. Drop checked operation (since we operate behind a validation boundary)

I will do that in a followup PR, and open a ticket in the meantime

PR #4257, opened 2024-04-22
Add ADR-001 for Dutch Auctions

cronokirby:
## Describe your changes

This creates a new folder for storing architecture design records, as part of our documentation. This adds the recent Dutch Auction design description here. This should encourage people to write more of them as there's a clear place to put them now.

  > Just documentation.


conorsch:
Out of band, we've discussed storing ADRs in a separate repo, or even appending directly to the protocol docs. FWIW, I agree with the very simple approach here: just stick them in their own directory. At any point in the future, we can wrangle the files elsewhere.

conorsch (approved):
Thanks for taking the time to record this, @cronokirby!

cratelyn (approved):

hdevalence:
Can we call these PADR or something so they're not confused with cosmos ADRS?

cronokirby:
> Can we call these PADR or something so they're not confused with cosmos ADRS?

Addressed

Issue #4154, opened 2024-04-03
add a flag to `testnet-generate` to start a testnet with non-zero gas prices


aubrika:
We should add this flag and use it in the smoke tests so that we're exercising non-zero fees. Otherwise, fees can break and we won't know.

aubrika:
This may have happened & may already be done, but we should make sure of this & also consider adding additional tests.

conorsch:
Checked on main, and it's not done. Updated OP accordingly.

TalDerei:
- [ ] Make a local devnet with non-zero fees Fees can be arbitrary, as long as they are there.
 Exercise basic transactions, using pcli to start.
- [ ] Currently, pcli uses the gas prices in the compactblock to set a fee, and we want to exercise that this actually works.

conorsch:
Adding the flag to modify genesis params is one thing, but updating the test suite to accommodate for tx fees is a can of worms. Given all the regex-based balance checks in the smoke tests, it'll be a bear to massage the existing test suite to accommodate for fees. Unclear whether it's worthwhile to port that functionality immediately to the mock-consensus—even if we do that, we'll still need to inspect the balances very precisely.

TalDerei:
@conorsch what's the current status of this? I'm currently experimenting with supporting nonzero prices in the pcli smoke tests.

conorsch:
Oops, lost track of this. I'll throw something simple together to unblock.

PR #4258, opened 2024-04-22
feat(pd): add --gas-price-simple flag to generate

conorsch:
## Describe your changes

Adds a new CLI flag for `pd testnet generate`, to opt in to non-zero gas
prices. It's a bit of a shortcut to reuse the same numeric value across
all of them, but we can always add more CLI flags later to finetune.

Closes https://github.com/penumbra-zone/penumbra/issues/4154.

  > Adds opt-in behavior, and modifies CI runs, but does not touch application code.
  
## Testing/review

The simplest way to validate the new behavior here is:

0. check out this feature branch
1. `cargo run --bin pd -- testnet generate --gas-price-simple 5`
2. start cometbft/pd to serve the local devnet
3. configure pcli to use http://locahost:8080 as grpc_addr
4. run `pcli --home <path> view balance` and confirm you have multiple denoms
5. run `pcli --home <path> tx send --to <different_addr_in_same_wallet> 1test_usd`
6. run `pcli --home <path> view balance` and confirm that 1) the sent denom moved; and 2) a wee bit of penumbra was burned to pay the fee.

### DIFF
--- a/crates/bin/pd/src/cli.rs
+++ b/crates/bin/pd/src/cli.rs
@@ -172,6 +172,13 @@ pub enum TestnetCommand {
         /// can be voted on.
         #[clap(long)]
         proposal_voting_blocks: Option<u64>,
+        /// The fixed gas price for all transactions on the network.
+        /// Described as "simple" because the single value will be reused
+        /// for all gas price types: block space, compact block space, verification, and execution.
+        /// The numeric value is one-thousandths of the staking token,
+        /// so `--gas-price-simple=1000` means all transactions will have a cost of 1penumbra.
+        #[clap(long)]
+        gas_price_simple: Option<u64>,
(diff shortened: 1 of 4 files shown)


TalDerei:
LGTM! 

so the trick here is to use another denom? I'm looking forward to seeing how our modified planner logic fares in this smoke test environment.

TalDerei (approved):

conorsch:
> so the trick here is to use another denom? I'm looking forward to seeing how our modified planner logic fares in this smoke test environment.

To my surprise, I didn't need to modify the smoke tests at all, because the tx send tests already use a non-staking token TEST_ASSET, so the addition of fees doesn't affect the balance checks of TEST_ASSET.

Issue #4228, opened 2024-04-17
app: connect the application with the auction component


erwanor:

erwanor:
This was done in #4259

PR #4259, opened 2024-04-23
auction: connect auction component to application

erwanor:
## Describe your changes

This PR adds:
- auction genesis content to the application state
- auction parameters to the application state (and governance update logic)
- connects the app ABCI implementation to the auction's component interface

#4228 and #4252 

  > This is consensus breaking


cratelyn (approved):
nice!

erwanor:
This was state-breaking because the auction component adds its own set of parameters to the app state!

PR #4262, opened 2024-04-23
staking: 🥞 `get_validator_info()` uses consensus set

cratelyn:
see #3846.

## :mag: changes

**NB:** this branch performs a refactor away from `validator_definitions()` and subsequently moves the definition of that method. this is done in distinct commits, which reviewers are encouraged to examine independently.

### staking: 🥞 `get_validator_info()` uses consensus set

> The staking component query service provides a ValidatorInfo RPC that returns
> the list of registered validators, optionally filtering Inactive validators.
>
> The way this method is implemented is by using an internal
> ValidatorDataRead::validator_definitions methods which returns every single
> definition ever submitted to the chain. This is problematic because this list
> will inevitably grow quite large, turning usage of a benign-looking method
> into an onerous source of I/O and expensive rendering.

this refactors the `validator_definitions()` method to only return validators
included within the consensus set. this will prevent the list returned from
becoming too large over time.

### app: 🧙 `ValidatorDataReadExt::validator_definitions()`

these methods are tremendously useful for tests, because they save us the
trouble of repeatedly hassling with a `Stream`. to discourage future use of
them in production code however, we can define these in a test-specific
extension.

this commit introduces `ValidatorDataReadExt`. `validator_identity_keys()` is
also moved, as it is subject to the same performance considerations.

  > this only changes the behavior of an RPC endpoint

### DIFF
--- a/crates/core/component/stake/src/component/rpc.rs
+++ b/crates/core/component/stake/src/component/rpc.rs
@@ -1,8 +1,7 @@
 use std::pin::Pin;
 
-use async_stream::try_stream;
 use cnidarium::Storage;
-use futures::{StreamExt, TryStreamExt};
+use futures::StreamExt;
 use penumbra_proto::{
     core::component::stake::v1::{
         query_service_server::QueryService, CurrentValidatorRateRequest,
@@ -13,10 +12,10 @@ use penumbra_proto::{
     DomainType,
 };
 use tonic::Status;
-use tracing::instrument;
+use tracing::{error_span, instrument, Instrument, Span};
 
-use super::{validator_handler::ValidatorDataRead, SlashingData};
-use crate::validator;
+use super::{validator_handler::ValidatorDataRead, ConsensusIndexRead, SlashingData};
+use crate::validator::{Info, State};
 
 // TODO: Hide this and only expose a Router?
 pub struct Server {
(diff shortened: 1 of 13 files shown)


cratelyn:
### :question: question for reviewers

there is a third to-do item mentioned in #3846, regarding a `pcli query validator status` command. i'd like to do that in a subsequent PR if possible, to keep this branch focused specifically on the changes to `ValidatorDataRead`.

is that reasonable, or is the need to inquire about validator state acute enough that i should do that here?

crates/core/component/stake/src/component/rpc.rs
```diff
@@ -39,38 +38,67 @@ impl QueryService for Server {
...
 
-        Ok(tonic::Response::new(
-            s.map_ok(|info| ValidatorInfoResponse {
-                validator_info: Some(info),
-            })
-            .map_err(|e: anyhow::Error| {
-                tonic::Status::unavailable(format!("error getting validator info: {e}"))
-            })
-            // TODO: how do we instrument a Stream
-            //.instrument(Span::current())
```
cratelyn, line 71: addressed this while i was here :relaxed:

cratelyn:
x-ref: asked a question about next steps here https://github.com/penumbra-zone/penumbra/issues/3846#issuecomment-2073517294

cratelyn:
@erwanor just force-pushed some typo fixes, sorry to thrash any review state :grimacing: :adhesive_bandage:

cratelyn:
> ### ❓ question for reviewers
> 
> there is a third to-do item mentioned in #3846, regarding a `pcli query validator status` command. i'd like to do that in a subsequent PR if possible, to keep this branch focused specifically on the changes to `ValidatorDataRead`.
> 
> is that reasonable, or is the need to inquire about validator state acute enough that i should do that here?

i've opened #4263. the changes to pcli ended up being a straightforward change. it is still kept distinct from the core changes here to facilitate review.

crates/core/component/stake/src/component/validator_handler/validator_store.rs
```diff
@@ -227,23 +227,6 @@ pub trait ValidatorDataRead: StateRead {
...
-    }
-
-    /// Returns a list of **all** known validators identity keys.
-    async fn validator_identity_keys(&self) -> Result<Vec<IdentityKey>> {
-        self.prefix(state_key::validators::definitions::prefix())
-            .map_ok(|(_key, validator)| validator)
-            .map_ok(|validator: Validator| validator.identity_key)
-            .try_collect()
-            .await
-    }
```
cratelyn, line 246: these are removed, and now live in a place where they are explicitly only for use in tests. that'll help prevent future regressions, but keeps the friendly ergonomics around for test assertions' sake.

crates/core/component/stake/src/component/rpc.rs
cratelyn, line 1: the unified diff here ends up looking a little funky, the "side-by-side" view is recommended here. you can click the little :gear: at the top of the page to get that.

cratelyn (commented):
:baguette_bread: breadcrumbs for review.

crates/core/component/stake/src/component/validator_handler/validator_store.rs
erwanor: great

erwanor (approved):
Awesome

cratelyn:
rebased onto main at dc8eff979 :recycle:

PR #4270, opened 2024-04-25
view(proto): remove reserves from auction withdrawal `TransactionPlannerRequest`

erwanor:
## Describe your changes

~~Accidentally defined the `TransactionPlannerRequest` auction withdraw submessage to specify the user reserves in `Amount`. But we need those to compute a balance commitment, and this requires the asset id.~~

@hdevalence noticed that we can defer populating this data to the view service (which is where we would have gotten it from in the first place!)

  > This is proto-breaking.


erwanor:
This is a protobuf breaking change, so I will want a +1 and ACK before I go on to override the lint and merge.

hdevalence:
Backing up a bit, why does the caller need to specify these in the request at all? Certainly they need to be known to plan the transaction, but should it be the caller's responsibility to supply the correct values, or should that be the responsibility of the view service that's processing the request?

cronokirby (approved):

TalDerei (approved):

hdevalence:
Let's remove the fields entirely, per @cronokirby's suggestion in discord

PR #4271, opened 2024-04-25
docs: chain upgrade for testnet 72

conorsch:
## Describe your changes
Bumps the application version to v0.72.0, which is required because the changes in #4293 were merged into a release branch, and thus didn't land on main naturally. We'll backport that patch shortly.

More relevantly, the chain-upgrade docs now reference the "current" migration, which is 71 -> 72, with its own genesis time.

## Issue ticket number and link

Refs #4239. 

## Checklist before requesting a review

- [x] If this code contains consensus-breaking changes, I have added the "consensus-breaking" label. Otherwise, I declare my belief that there are not consensus-breaking changes, for the following reason:

  > docs only

PR #4272, opened 2024-04-25
docs: update archive url for joining testnet

conorsch:

Refs #4239, #4271.

  > docs-only

## Testing

Try to join the network by running the testnet join command in the docs. You must use `v0.72.0` in order for it to work, otherwise your local copy of pd won't be compatible. Then verify you can actually join the network and grab blocks.


conorsch:
I made sure to test this myself, and was able to join the newly restarted testnet.

PR #4274, opened 2024-04-26
chore: bump chain id for testnet 73

conorsch:
## Describe your changes

We mistakenly broke testnet 72 on chain upgrade, by leaving a forked network running that caused tombstoning. To resolve, we're going to reset chain state and use a new chain id. This commit ensures that the chain id is bumped `penumbra-testnet-deimos-6` ->
`penumbra-testnet-deimos-7`. Not strictly necessary, but for posterity, copied over the same configs for the intervening 71 & 72 testnets.

  >  This PR updates the chain id and will result in a new network, so it's definitely consensus-breaking.
  
 ## Testing and review
 
 Run `cargo run --release --bin pd -- testnet generate` locally and confirm that you see `penumbra-testnet-deimos-7`, not `-6`, in the output.


conorsch:
Self-merging to get this out the door.

Issue #3797, opened 2024-02-09
staking: inventorize useful staking events to index


zbuc:
The staking component has minimal event indexing (delegate and undelegate as of now).

Similarly to #3645 we should enumerate useful events to index (state transitions and such) and implement.

zbuc:
Events to index:

- [x] Delegate
- [x] Undelegate
- [ ] Undelegate claim
- [ ] Validator definition change
- [ ] Validator state changes
- [ ] Validator rate/power/delegation token supply changes during end_epoch
- [ ] Validator missed block
- [ ] Funding stream reward payouts
- [ ] Slashing penalties applied
- [ ] Bonding state change

Issue #3746, opened 2024-02-06
metrics: surface validator status, e.g. jailed


conorsch:
During Testnet 64 folks in Discord have reported an uptick in jailing rates. It's difficult to reason concretely about that observation because we don't have good visibility into rates of banning. We do emit metrics such as `penumbra_stake_validators_jailed` in pd, but aren't visualizing those anywhere out of the box. We also try to log the status change via a span: https://github.com/penumbra-zone/penumbra/blob/v0.64.2/crates/core/component/stake/src/component.rs#L125-L135 but oddly I'm not seeing that span label return results in our 64 logs, even though we have debug logs on for the validators we run.

conorsch:
From current testnet:

```
❯ curl -s http://localhost:9000/metrics | rg ^penumbra_stake | rg -v missed_blocks
penumbra_stake_validators_inactive 363
penumbra_stake_validators_jailed -23
penumbra_stake_validators_disabled 29
penumbra_stake_validators_active 0
penumbra_stake_validators_defined 64
```

There are a few problems in there, such as active=0 and "jailed" being a negative value. Let's inspect the transition logic in the staking component and make sure gauges are incremented _and_ decremented correctly.

conorsch:
Similarly, the missed blocks metric is not accurate:

```
❯ curl -s http://localhost:9000/metrics | rg ^penumbra_stake_missed_blocks | sort -k 2 -n -r | head -n 20
penumbra_stake_missed_blocks{identity_key="penumbravalid14e54ukuppxs3f6dsg6flx9tsmqcv363tvvzrdku8fgmagjd435fqexu7zs"} 148359350
penumbra_stake_missed_blocks{identity_key="penumbravalid1u60u8405jg8mlz0epmhqfslqqkjurx6l4pm8n3j0ves66sztwqrqctw4f2"} 137059677
penumbra_stake_missed_blocks{identity_key="penumbravalid1qrxgv0nqm8a5edffr5t5ykhjuchkmwkejsgfhtc38t2y0kf33qrsql244n"} 133699997
penumbra_stake_missed_blocks{identity_key="penumbravalid1g3wfpjkplhktcq5d3t3l5ud4u34qwh6kxacwrkd8uacvh5jlcqqqzqnhxt"} 130544894
[...]

PR #4277, opened 2024-04-26
staking: emit event when tombstoning active/jailed/inactive vals

erwanor:
## Describe your changes

This PR adds a `EventTombstoneValidator` message that is emitted on the *first* validated evidence message for validators that are `Active | Inactive | Jailed`.

The evidence event is defined as follow:

```proto
message EventTombstoneValidator {
  // The height at which the offense occurred.
  uint64 evidence_height = 1;
  // The height at which the evidence was processed.
  uint64 current_height = 2;
  // The validator identity key.
  keys.v1.IdentityKey identity_key = 4;
  // The validator's Comet address.
  bytes address = 5;
  // The voting power for the validator.
  uint64 voting_power = 6;
}
```

My first instinct was to reach for storing the evidence in nonverifiable storage, hoping it would make it easier for node runners to query it. But I later realized, that comet only gives us a filtered version that does not include the actual double signed headers, or offending signatures.

  > Event emission


erwanor:
We'll need to port and complete the event coverage for the staking component as part of #3504 and #3797

conorsch:
This looks great, @erwanor, and provides us with a nice shape for adding the other state transitions, as described in https://github.com/penumbra-zone/penumbra/issues/3746#issuecomment-2049976212. The only thing I'm unclear on is why the emission is gated on transitions out of `Active | Inactive | Jailed`. The state machine diagram makes it clear that Defined and Disabled validators can also be Tombstoned—why not also emit the event in those scenarios?

erwanor:
Good point, my reasoning is that those transitions are less relevant because they don't affect the consensus view at all. Something else that I had in mind is that generating valid equivocation evidence rounds down to "free", hence why we only record the first offense, and tie it to 1/ a validator definition upload 2/ having reached the minimum stake threshold 3/ being enabled.

conorsch (approved):

conorsch:
Thanks, that's clearly explained!

erwanor:
There's a good overview in the spec: https://protocol.penumbra.zone/main/concepts/validators.html, happy to chat about it offline as well!

cratelyn (approved):

PR #4279, opened 2024-04-27
Add an RPC for querying the uptime of a single validator

plaidfinch:
## Describe your changes

This adds a new RPC method to the staking query service which gets the `Uptime` of a validator. This is useful for validator operators who want to know when their nodes have been signing blocks (or importantly, not doing so!). This PR also adds a query command `pcli query validator uptime` which exercises this RPC and uses it to print human-readable information about a given validator's uptime.

**Status:**I have tested this and it works, at least on a one-validator devnet. I have not tested in a scenario where actual downtime is possible, but I have simulated output to check that it looks good. Output looks like this for a 100% up validator:

```
Active validator: as of block 2466
Unmissed signing: 100.00% = 10000/10000 most-recent blocks
Required signing:   5.00% =   500/10000 most-recent blocks
Salient downtime:   0.00% =     0/10000 most-recent blocks ~ 0s cumulative downtime
Unexpended grace: 100.00% =  9500/10000 forthcoming blocks ~ 13h 11m 40s at minimum before penalty
Downtime penalty:   1.00% - if downtime exceeds 95.00%, penalty will be applied to all delegations
```

Discussion rooted here: https://discord.com/channels/824484045370818580/979785365437165658/1233538125909983242

  > This is a purely additive change to expose a new RPC query method and a client behavior that uses it.

### DIFF
--- a/crates/bin/pcli/src/command/query/validator.rs
+++ b/crates/bin/pcli/src/command/query/validator.rs
@@ -262,7 +262,7 @@ impl ValidatorCmd {
                 let missed_blocks = uptime.num_missed_blocks();
                 let window_len = uptime.missed_blocks_window();
 
-                let mut downtime_ranges: Vec<RangeInclusive<u64>> = Vec::new();
+                let mut downtime_ranges: Vec<RangeInclusive<u64>> = vec![];
                 for missed_block in uptime.missed_blocks() {
                     if let Some(range) = downtime_ranges.last_mut() {
                         if range.end() + 1 == missed_block {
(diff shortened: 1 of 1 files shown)


crates/core/component/stake/src/uptime.rs
```diff
@@ -142,6 +170,64 @@ mod tests {
...
+            let missed_blocks: Vec<_> = uptime.missed_blocks().collect();
+
+            // Retain only the heights from the simulated `VecDeque` that were not signed:
+            simulated_uptime.retain(|(_, signed)| !signed);
+            let simulated_missed_blocks: Vec<_> =
+                simulated_uptime.into_iter().map(|(height, _)| height).collect();
+
+            prop_assert_eq!(missed_blocks, simulated_missed_blocks);
+        }
+    }
```
cratelyn, line 229: what a delightful test!

flake.nix
```diff
@@ -107,7 +103,7 @@
           devShells.default = craneLib.devShell {
             inherit LIBCLANG_PATH;
             inputsFrom = [ penumbra ];
-            packages = [ cargo-watch ];
+            packages = [ cargo-watch cargo-nextest protobuf ];
```
cratelyn, line 106: :+1: nice

flake.nix
```diff
@@ -23,10 +23,6 @@
         let
           # Define versions of Penumbra and CometBFT
           penumbraRelease = null; # Use the local working copy
-          # penumbraRelease = { # Use a specific release
-          #   version = "0.71.0";
-          #   sha256 = "sha256-2mpyBEt44UlXm6hahJG9sHGxj6nzh7z9lnj/vLtAAzs=";
-          # };
```
cratelyn, line 29: i found this previous comment of yours to be a friendly invitation to tinkering, but i don't mind removing this either! :woman_shrugging:

cratelyn (approved):
this is very nicely done. thank you for doing this! :pager:

PR #4282, opened 2024-04-29
ci: update cargo-dist 0.5.0 -> 0.13.3

### LINKED ISSUE #4195: release dist-manifest.json describes artifacts that do not exist
`source.tar.gz` and `source.tar.gz.sha256` are not uploaded as release artifacts, though they are generated by `cargo dist` and described in `dist-manifest.json`

yes there is a `v0.71.0.tar.gz` and such published by github, but github does not provide a checksum for that file

### LINKED ISSUE #4255: `pcli` installation instructions don't have copy-pastable code for MacOS
**Describe the bug**

`pcli` is more difficult to install on MacOS because the instructions aren't copy-pastable: only Linux commands are provided.

It would be better if the guide had a separate block with copy-pastable MacOS instructions.

### LINKED ISSUE #4256: `pcli` install instructions suggest `unxz`, which is not available by default on MacOS

The `pcli` install instructions suggest using `unxz` to extract the xz archive, but this command is not included by default on MacOS. I never noticed because I already had it installed.

Figure out an alternative extraction method or archive format.  I don't think that the compression choice will make a huge difference, since most of our data is (cryptographically) random and uncompressible.

conorsch:
## Describe your changes

Bumps the cargo-dist dependency, and regenerates the relevant CI files. Tested these changes in a private fork, so we can be confident they won't break builds.

Motivated by deprecation warnings in cargo-dist declaring that pinning rust versions should use `rust-toolchain.toml`, updated ours to pin 1.75. Standardizing on the toolchain file lets us remove repeated versions from a few different places, such as the containerbuild. Removed custom action helpers from Rust CI workflows, because GHA runners already come with rustup, and will honor the toolchain file. This change ensures that our dev, build, and test environments all use the same version of rust, which should minimize surprises like we saw in #4231.

Included in the release artifacts now is a complete source tarball with associated checksum, as requested in #4195. We're also distributing binaries compressed as .tar.gz, rather than .tar.xz, to satisfy #4256, and opting into a custom install script to satisfy #4255. Therefore:

* closes #4195
* closes #4255
* closes #4256

  > docs and ci only, no code changes

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,5 +1,0 @@
-[build]
-# Enable Tokio's `tracing` support for `tokio-console`
-# rustflags = ["--cfg", "tokio_unstable"]
-# Note(erwan): We decided to disable it for the time being,
-# I'm keeping this around to be able to reactivate it on a whim.
--- a/.github/workflows/buf-pull-request.yml
+++ b/.github/workflows/buf-pull-request.yml
@@ -54,9 +54,6 @@ jobs:
         with:
           lfs: true
 
-      - name: Install rust toolchain
-        uses: dtolnay/rust-toolchain@stable
-
       - uses: bufbuild/buf-setup-action@v1
         with:
           buf_api_token: ${{ secrets.BUF_TOKEN }}
--- a/.github/workflows/docs-lint.yml
+++ b/.github/workflows/docs-lint.yml
@@ -13,9 +13,9 @@ jobs:
         with:
           lfs: false
 
-      - name: Install rust toolchain
+      - name: Install nightly rust toolchain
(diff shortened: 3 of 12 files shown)


conorsch:
Post-merge, I'd like cherry-pick this commit onto the release branch for the current testnet, `release/0.73.x`, in preparation for a point release containing a pcli fix: https://github.com/penumbra-zone/penumbra/pull/4235

conorsch:
> in preparation for a point release containing a pcli fix: https://github.com/penumbra-zone/penumbra/pull/4235

https://github.com/penumbra-zone/penumbra/pull/4284

erwanor (approved):

Issue #4123, opened 2024-03-28
Auto-closing positions don't emit `EventPositionClose`


hdevalence:
**Describe the bug**

When the DEX auto-closes a position inside `PositionManager::position_execution`, it doesn't emit an `EventPositionClose` the same way as if the position were closed manually.

**To Reproduce**

I made a mispriced position `tx position order sell 100penumbra@1gm --limit-order` then viewed in the explorer: https://dex-explorer.testnet.penumbra.zone/lp/plpid1d79ghpr2v6qlaytcky8m57sxjdea99cccxlreca245nnr3qkym6s07yu4c

<img width="768" alt="Screenshot 2024-03-27 at 8 52 27 PM" src="https://github.com/penumbra-zone/penumbra/assets/44879/85795770-31a4-4e99-a7ea-be4751dd2fb0">

There's no bar showing when the position was closed, because the event was never emitted. (Also, the execution events seem potentially out of order?)

To fix this, we'll need to change `position_execution` to emit the event.

hdevalence:
One thing I'm realizing while writing up the issue is that we don't have a consistent story about when the closure actually happens.

If the `EventPositionClose` is emitted when the position is actually updated (in `EndBlock`), there's no way to tie it back to the transaction that closed the position, because the event will be block scoped. Maybe we should have two events, one that's emitted when the position is actually closed, and a new `EventQueuePositionClose` or something that's emitted in the context of the transaction that enqueues the closure?

WDYT @erwanor @philipjames44 ?

PR #4283, opened 2024-04-29
dex: fix frontier cache incoherence

erwanor:
**Description**

This PR fix a bug in the `Frontier` position cache that caused a position execution guard to be hit.

**Context**

During routing execution, the DEX engine assemble frontiers of positions to execute against. It maintains an internal cache with a position stream for each hop in the path, along with the ids and state of the current loaded set. The position cache is persisted in its entirety when a round of routing execution has completed (via `Frontier::save`), and partially, if a position needs to be replaced (`Frontier::replace_position`).

**Context**

The bug occurs when the following conditions align:
- a position needs to be replaced
- that position is auto-closing, and set to `Closed` when recording its execution
- no replacements are available

In that case, since the position rotation logic does not update the internal state of the position, it will attempt to record execution of a stale position when saving the frontier.

**Fix**

The fix is two-fold, first, `PositionManager::position_execution` should make it clear that it can mutate its input and return the updated position state to allow callers to inspect its state. Once that's done, the position rotation logic will be able to immediately update its internal cache with the actual position state so that there are no incoherences between its internal view and the actual chain state.

  > This is consensus breaking.


conorsch:
> This is consensus breaking.

Added label. This means we'll need to write a migration in order to release this, and provide validators with instructions, right?

erwanor:
Ooops thanks I missed that, no migration needed, it's an internal structure and only used during execution

crates/core/component/dex/src/component/position_manager.rs
```diff
@@ -253,22 +253,53 @@ pub trait PositionManager: StateWrite + PositionRead {
...
             .ok_or_else(|| anyhow::anyhow!("withdrew from unknown position {}", new_state.id()))?;
 
+        // Optimization: it's possible that the position's reserves haven't
+        // changed, and that we're about to do a no-op update. This can happen
+        // when saving a frontier, for instance, since the FillRoute code saves
+        // the entire frontier when it finishes.
+        //
+        // If so, skip the write, but more importantly, skip emitting an event,
+        // so tooling doesn't get confused about a no-op execution.
+        if prev_state == new_state {
```
zbuc, line 294: Nit: `state` is a bit overridden here and it took me a couple reads to understand that `prev_state`/`new_state` aren't instances of the `position::State` enum but instead `Position`s.

hdevalence:
Can we fix https://github.com/penumbra-zone/penumbra/issues/4123 as part of these changes?

hdevalence (changes requested):
LGTM but we should fix https://github.com/penumbra-zone/penumbra/issues/4123 while we're making these changes.

erwanor:
@hdevalence added a fix/event for #4123

hdevalence (approved):

PR #4285, opened 2024-04-29
keys: 🥰 `Address` defers computation of diversified base

### LINKED ISSUE #1957: Change domain types for keys and addresses to store bytes and defer expensive crypto operations until use.

Our domain types for cryptographic keys perform some crypto operations during parsing.  For instance, verification keys decompress the group element, and our `Address` type computes the diversified basepoint.

This was originally done as part of a "parse, don't validate" approach, but it imposes real costs, because it means that every time we pass back and forth between bytes and domain types, we pay for the precomputation, even if we'll never use it. For instance, the only time we actually use an `IdentityKey` as an identity key is when we verify updated validator configuration data, and at all other times it's just used as an identifier. But we pay the cost at all times.

In other cases, this is even more wasteful: our binding and spend authorization signatures only ever use the same verification key once, so there's no possible benefit to precomputation, rather than waiting to decompress the verification key while processing the signature.

However, it is good to prevent fallibility from propagating through the codebase, so if not doing some check in decompression would result in operations that are currently infallible becoming fallible, we shouldn't defer it. (As a counterexample, deferring decompression of a verification key doesn't introduce any extra fallibility, because signature verification can already return an error).

cratelyn:
### 💬 describe your changes

fixes #1957.

the last remaining item mentioned in #1957 is that currently, while `Address`
_caches_ the diversified base point, it eagerly computes it when
`Address::from_components()` is called.

we can defer this action until the base point is needed, using `OnceLock<T>`.
in order to do that, we must revoke the `Copy` derivation. essentially, this
branch makes the following change(s) to `Address`:

```diff
-#[derive(Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Clone, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(try_from = "pb::Address", into = "pb::Address")]
 pub struct Address {
     d: Diversifier,
     /// cached copy of the diversified base
-    g_d: decaf377::Element,
+    g_d: OnceLock<decaf377::Element>,
 
     /// extra invariant: the bytes in pk_d should be the canonical encoding of an
     /// s value (whether or not it is a valid decaf377 encoding)
     /// this ensures we can use a PaymentAddress to form a note commitment,
     /// which involves hashing s as a field element.
     pk_d: ka::Public,
     /// transmission key s value
     transmission_key_s: Fq,
 
     ck_d: fmd::ClueKey,
 }
```
out of that, a few other types also lose their `Copy` property:
* AddressView
* FundingStream
* Recipient

..which leads to the addition of explicit `.clone()` calls in various places.

largely, that tends to affect test code, but it is a more invasive change than
expected, and i want to draw attention to that. it should _not_ be a change in
behavior, however. these are the same copies as previously made, but are now
performed explicitly.

for more, see the `Copy` documentation:
https://doc.rust-lang.org/stable/std/marker/trait.Copy.html#whats-the-difference-between-copy-and-clone

some extra documentation, a small compile-time check to enshrine thread-safety
properties (`Send` and `Sync`), and an extra `From<T>` implementation are added
too.

#### ➕ changes

to facilitate review and to make this easier to reason about, this branch is
divided into distinct steps:

* keys: 🔖 add more `Address` documentation
* keys: 🙂 add compile-time assertions about addresses
* keys: 📫 `pb::Address: From<&Address>`
* keys: 👯 `Address` is not `Copy`
* keys: 🥰 defer diversified base using `OnceLock<T>`

...reviewers are encouraged to review this step-by-step. :mag: 

**NB:** using `OnceLock<T>` is the core change at play here.

### issue ticket number and link

* #1957

  > this does not change the behavior of addresses, it only defers the
  > computation of the diversified base point.

### DIFF
--- a/crates/bin/pcli/src/command/ceremony.rs
+++ b/crates/bin/pcli/src/command/ceremony.rs
@@ -139,7 +139,7 @@ impl CeremonyCmd {
                     bid, address
                 );
 
-                handle_bid(app, *coordinator_address, index, bid).await?;
+                handle_bid(app, coordinator_address.clone(), index, bid).await?;
 
                 println!("connecting to coordinator...");
                 // After we bid, we need to wait a couple of seconds just for the transaction to be
--- a/crates/bin/pcli/src/command/tx.rs
+++ b/crates/bin/pcli/src/command/tx.rs
@@ -31,7 +31,7 @@ use penumbra_asset::{asset, asset::Metadata, Value, STAKING_TOKEN_ASSET_ID};
 use penumbra_dex::{lp::position, swap_claim::SwapClaimPlan};
 use penumbra_fee::Fee;
 use penumbra_governance::{proposal::ProposalToml, proposal_state::State as ProposalState, Vote};
-use penumbra_keys::keys::AddressIndex;
+use penumbra_keys::{keys::AddressIndex, Address};
 use penumbra_num::Amount;
 use penumbra_proto::{
     core::component::{
@@ -346,7 +346,7 @@ impl TxCmd {
                     .map(|v| v.parse())
                     .collect::<Result<Vec<Value>, _>>()?;
                 let to = to
-                    .parse()
+                    .parse::<Address>()
                     .map_err(|_| anyhow::anyhow!("address is invalid"))?;
 
                 let return_address = app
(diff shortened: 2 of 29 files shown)


cratelyn:
ci is failing, but i understand why. https://doc.rust-lang.org/stable/src/std/sync/once_lock.rs.html#500-505, `OnceLock<T>`'s `PartialEq` uses `get()` rather than `get_or_init()`. so, an uninitialized value will be considered not equal to an initialized value, even if they would resolve to the same value once initialized.

i'll fix this shortly.

redshiftzero (approved):
LGTM, the diversified base computation is infallible, and the choice of OnceLock seems reasonable for thread safety

Issue #3790, opened 2024-02-09
Investigate arb execution


hdevalence:
**Describe the bug**

Look at the following output:
```
~/c/p/penumbra3 ((v0.66.0)|✔) $ cargo run --release --bin pcli --  q dex arb-execution --height 4645
    Finished release [optimized] target(s) in 0.28s
     Running `target/release/pcli q dex arb-execution --height 4645`
Scanning blocks from last sync height 4675 to latest height 4675
[0s] ██████████████████████████████████████████████████       0/0       0/s ETA: 0s
46.566735penumbra => 6.342551penumbra via:
    Trace                                                                                                                                                    Subprice
    4.801446penumbra =>   6.241878888888888889test_usd =>   5.617691penumbra =>             5.617691gm =>          7.02211375test_usd =>  5.61769penumbra    854.702mpenumbra/1penumbra
    4.801446penumbra =>   6.241878888888888889test_usd =>   5.617691penumbra =>             7.02211375test_usd =>  5.61769gm =>           5.61769penumbra    854.702mpenumbra/1penumbra
    4.801446penumbra =>   6.241878888888888889test_usd =>   5.617691penumbra =>             5.617691gm =>          7.02211375test_usd =>  5.61769penumbra    854.702mpenumbra/1penumbra
    4.801446penumbra =>   6.241878888888888889test_usd =>   5.617691penumbra =>             7.02211375test_usd =>  5.61769gm =>           5.61769penumbra    854.702mpenumbra/1penumbra
    4.256778penumbra =>   5.533811344444433872test_usd =>   4.98043penumbra =>              4.98043gm =>           6.2255375test_usd =>   4.980429penumbra   854.702mpenumbra/1penumbra
    4.98043penumbra =>    6.2255375test_usd =>                                                                                            5.602983penumbra   888.89mpenumbra/1penumbra
    637.261mpenumbra =>   637.261mgm =>                     796576250000000000wtest_usd =>                                                716.918mpenumbra   888.89mpenumbra/1penumbra
    5.617691penumbra =>   5.617691gn =>                     7.02211375test_usd =>                                                         6.319902penumbra   888.889mpenumbra/1penumbra
    11.868791penumbra =>  14.242548888888888889test_usd =>                                                                                12.818294penumbra  925.926mpenumbra/1penumbra
```

**Expected behavior**

The summary `46.566735penumbra => 6.342551penumbra` seems wrong?

erwanor:
[...]

PR #4287, opened 2024-04-29
Add test for and fix arbitrage swap execution output amount

zbuc:
## Describe your changes

This extends the existing arbitrage test to ensure that the created `SwapExecution` has the expected trace and input/output values, as well as fixes the bug described in #3790 where `SwapExecution`s were created with incorrect `Output` values.

#3790

## Additional Follow-up

This will require a migration to update the existing `SwapExecution` structures. The migration should be as simple as adding the `input` to the `output` for each of them.

### DIFF
--- a/crates/bin/pd/src/migrate/testnet74.rs
+++ b/crates/bin/pd/src/migrate/testnet74.rs
@@ -1,0 +1,166 @@
+//! Contains functions related to the migration script of Testnet72
+
+use anyhow;
+use cnidarium::{Snapshot, StateDelta, StateRead, StateWrite, Storage};
+use futures::StreamExt as _;
+use jmt::RootHash;
+use penumbra_app::app::StateReadExt as _;
+use penumbra_app::SUBSTORE_PREFIXES;
+use penumbra_dex::SwapExecution;
+use penumbra_proto::core::component::sct::v1::query_service_server::QueryService;
+use penumbra_proto::penumbra::core::component as pb;
+use penumbra_proto::StateWriteProto;
+use penumbra_sct::component::clock::{EpochManager, EpochRead};
+use penumbra_sct::component::rpc::Server as SctServer;
+use penumbra_tct::Position;
+use prost::Message;
+use std::path::PathBuf;
+use std::sync::Arc;
+use tonic::IntoRequest;
+
+use crate::testnet::generate::TestnetConfig;
+
+/// Updates arb execution output amounts to include the input amount instead
+/// of reporting only profit (see #3790).
+async fn fix_arb_execution_outputs(delta: &mut StateDelta<Snapshot>) -> anyhow::Result<()> {
+    let mut stream = delta.prefix_raw("dex/arb_execution/");
+    while let Some(r) = stream.next().await {
+        let (key, swap_ex_bytes) = r?;
+        let mut swap_ex: SwapExecution =
+            pb::dex::v1::SwapExecution::decode(swap_ex_bytes.as_slice())?.try_into()?;
+        swap_ex.output = swap_ex.output + swap_ex.input;
+        delta.put(key, swap_ex);
+    }
+    Ok(())
+}
+
+/// Update the ordering of liquidity position indices to return in descending order (see #4189)
+async fn update_lp_index_order(delta: &mut StateDelta<Snapshot>) -> anyhow::Result<()> {
+    let prefix_key = "dex/ra/".as_bytes();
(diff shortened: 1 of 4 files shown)


crates/core/component/dex/src/component/arb.rs
```diff
@@ -114,7 +114,7 @@ pub trait Arbitrage: StateWrite + Sized {
                 amount: filled_input,
             },
             output: Value {
-                amount: arb_profit,
+                amount: filled_input + arb_profit,
```
erwanor, line 117: makes sense to me

crates/core/component/dex/src/component/tests.rs
```diff
@@ -866,6 +866,9 @@ async fn basic_cycle_arb() -> anyhow::Result<()> {
 /// The issue was that we did not treat the spill price as a strict
 /// upper bound, which is necessary to ensure that the arbitrage logic
 /// terminates.
+///
+/// This test also ensures that the created `SwapExecution` has the
+/// correct data.
```
erwanor, line 871:
```suggestion
///
/// *Arbitrage execution record bug:*
/// This test also ensures that the created `SwapExecution` has the
 /// correct data. (See #3790).
```

Good idea to test this but can we add context that it's separate from the testnet 53 issue

erwanor (approved):

crates/bin/pd/src/migrate/testnet74.rs
```diff
@@ -0,0 +1,159 @@
...
+use penumbra_sct::component::clock::{EpochManager, EpochRead};
+use prost::Message;
+use std::path::PathBuf;
+
+use crate::testnet::generate::TestnetConfig;
+
+/// Updates arb execution output amounts to include the input amount instead
+/// of reporting only profit (see #3790).
+async fn fix_arb_execution_outputs(delta: &mut StateDelta<Snapshot>) -> anyhow::Result<()> {
+    let mut stream = delta.prefix_raw("dex/arb_execution/");
```
erwanor, line 22: Could offload deserialization to `prefix_proto` would straighten up the rest of the migration

crates/bin/pd/src/migrate/testnet74.rs
```diff
@@ -0,0 +1,159 @@
...
+    while let Some(r) = stream.next().await {
+        let (key, swap_ex_bytes) = r?;
+        let mut swap_ex: SwapExecution =
+            pb::dex::v1::SwapExecution::decode(swap_ex_bytes.as_slice())?.try_into()?;
+        swap_ex.output = swap_ex
+            .input
```
erwanor, line 28: ACK this works because it's a cycle

crates/bin/pd/src/migrate/testnet74.rs
```diff
@@ -0,0 +1,159 @@
...
+
+    Ok(())
+}
+
+/// Run the full migration, given an export path and a start time for genesis.
+pub async fn migrate(
+    path_to_export: PathBuf,
+    genesis_start: Option<tendermint::time::Time>,
+) -> anyhow::Result<()> {
+    // Lookups for liquidity positions based on starting asset were ordered backwards
```
erwanor, line 75: This migration actually does two things right? IMO it would help readability to  lead with a couple bullets and have a subsection specifies `state_key` + `storage_location` and link to github tickets for more context.

I think it's better because we can get a local view of what the migration "spec" is, and can hit up github for more context.

crates/bin/pd/src/migrate/testnet74.rs
```diff
@@ -0,0 +1,159 @@
...
+        // Delete the old incorrectly ordered key:
+        delta.nonverifiable_delete(old_key.clone());
+
+        // Store the correctly formatted new key:
+        delta.nonverifiable_put_raw(new_key.to_vec(), asset_id);
+        tracing::info!(
+            new_key = ?EscapedByteSlice(&new_key),
+            ?old_key,
+            "updated liquidity index"
+        );
```
erwanor, line 64: ACK, read through the code and it look correct to me, we:

1. Create a (K, V) stream
2. Pull the aggregate amount from the key @ `32 + 7..32 + 7 + 16`
3. Compute the complement, encode it as big endian
4. Delete the old key
5. Insert the new key

erwanor (approved):

crates/bin/pd/src/migrate/testnet74.rs
```diff
@@ -0,0 +1,159 @@
+//! Contains functions related to the migration script of Testnet72
```
cratelyn, line 1:
```suggestion
//! Contains functions related to the migration script of Testnet74
```

PR #4291, opened 2024-04-30
Include`DelegatorVoteView` as an `ActionView` with `Visible` and `Opaque` Variants

ejmg:
## Describe your changes

Small change to `transaction.proto` that moves `DelegatorVoteView` to be included in the section of `ActionView`s that are explicitly documented as having `Opaque` and `Visible` variants. Beyond generated protobuf code from `scripts/protobuf-codegen`, there are no other changes. 

  > This PR only changes the order of a Protobuf Message definition for the purposes of documentation.


erwanor:
ACK, the comments are a bit unclear in the proto message, but grouping the `*View` is fine, will merge on green ci (conflict to address first).

cratelyn:
rebased on the latest main, but we'll do that again pending two other forthcoming pr's. (cc: @erwanor)

conorsch:
We intend to release this as part of Testnet 74. It'll need one more rebase after https://github.com/penumbra-zone/penumbra/pull/4368 lands, then should be good to go.

conorsch (approved):

cratelyn:
thank you for doing this @ejmg! this looks great.

Issue #4177, opened 2024-04-08
dex: surface swap/position assets during candidate selection


erwanor:
In order to decide which part of the search-space to explore, the DEX uses an extended candidate set that comprise:
- fixed assets that are "safe defaults", as well as,
- dynamic candidates, surfaced using liquidity heuristics

In #4066, @hdevalence noted that there are straightforward ways to make the routine cover more ground, by including:
- assets that form a trading pair that had swaps in the block
- assets that form a trading pair for which positions were opened in the block.

More context on dynamic candidate selection is available in the web of issues: #2750

zbuc:
Code complete, merging pending review

PR #4292, opened 2024-04-30
dex: Extend arbitrage routing

zbuc:
## Describe your changes

This PR extends arbitrage routing to route against assets involved in swaps and positions opened during the current block.

#4177 

### DIFF
--- a/crates/core/component/dex/src/component/dex.rs
+++ b/crates/core/component/dex/src/component/dex.rs
@@ -17,7 +17,7 @@ use crate::{
 
 use super::{
     router::{HandleBatchSwaps, RoutingParams},
-    Arbitrage, PositionManager, ValueCircuitBreaker,
+    Arbitrage, PositionManager, PositionRead as _, ValueCircuitBreaker,
 };
 
 pub struct Dex {}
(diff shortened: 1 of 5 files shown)


crates/core/component/dex/src/component/dex.rs
```diff
@@ -83,12 +83,22 @@ impl Component for Dex {
...
+        // Extend the fixed candidate set to include recently accessed assets, to have
+        // more arbitrage execution against newly opened positions.
+        let fixed_candidates = Arc::new(
+            routing_params
+                .fixed_candidates
+                .iter()
+                .cloned()
+                // Limit the inclusion of recently accessed assets to 10 to avoid
+                // potentially blowing up routing time.
+                .chain(state.recently_accessed_assets().iter().take(10).cloned())
```
hdevalence, line 95: This could have duplicates right? should we accumulate into a BTreeSet instead until its size increases by the max limit?

zbuc: The `recently_accessed_assets` set is also incredibly likely to feature assets that are _already_ in the fixed list, perhaps they should be filtered out on insertion as well.

crates/core/component/dex/src/component/position_manager.rs
```diff
@@ -244,13 +254,53 @@ pub trait PositionManager: StateWrite + PositionRead {
...
+        // Add the asset IDs from the new position's trading pair
+        // to the candidate set for this block.
+        let routing_params = self.routing_params().await?;
+        self.add_recently_accessed_asset(
+            position.phi.pair.asset_1(),
+            routing_params.fixed_candidates.clone(),
+        );
+        self.add_recently_accessed_asset(
+            position.phi.pair.asset_2(),
+            routing_params.fixed_candidates,
```
erwanor, line 266: It was a minor surprise that this wasn't internalized by `add_recently_accessed_asset` but I guess we technically save a lookup.

erwanor (approved):
LGTM, tested manually on a devnet

Issue #4219, opened 2024-04-16
auction: `TxV/TxP` support to populate the views


erwanor:

erwanor:
All the scaffolding exists to make this happen, we just need to connect do the plumbing so that the views get populated. This is somewhat lower priority, but if anyone wants to pick it up feel free to.

erwanor:
We stubbed the withdraw view, this needs some addition to transaction perspectives, so leaving this open, until I replace it with a more detailed plan of action.

cronokirby:
Shelved for now, this turns out to be a bit more subtle than we realized, and also requires a bit of a different processing scheme than other views.

Basically, unlike other views, we need access to the _historical_ value of the auction reserves at the time the withdrawal happens, which is unlike how the rest of the view services are designed.

We could add this in the future with the addition of an event for withdrawing actions, and then indexing those events to populate the view of withdrawal actions appropriately.

PR #4295, opened 2024-04-30
auction: add views to the `ActionView` definition

erwanor:
## Describe your changes

This PR adds `ActionDutchAuctionScheduleView` and `ActionDutchAuctionWithdrawView` to the `ActionView` message, it also replaces the `IsAction` perspective impls with concrete implementation.

Populating a view for the withdraw action without extending the `TxP` structure has proven squirrelly, and it will require a follow-up, although the data is "here" in the view server so I don't think it will be too complicated.

Part of #4219 

  > Breaking changes are made to `ActionView`: the `action_dutch_auction_schedule` and `action_dutch_auction_withdraw` fields have had their types changes from `ActionDutchAuctionSchedule` and `ActionDutchAuctionWithdraw` to `ActionDutchAuctionScheduleView` and `ActionDutchAuctionWithdrawView`, respectively.


proto/penumbra/penumbra/core/transaction/v1/transaction.proto
```diff
@@ -223,9 +223,9 @@ message ActionView {
     component.governance.v1.CommunityPoolOutput community_pool_output = 51;
     component.governance.v1.CommunityPoolDeposit community_pool_deposit = 52;
     // Dutch auctions
-    component.auction.v1alpha1.ActionDutchAuctionSchedule action_dutch_auction_schedule = 53;
+    component.auction.v1alpha1.ActionDutchAuctionScheduleView action_dutch_auction_schedule = 53;
     component.auction.v1alpha1.ActionDutchAuctionEnd action_dutch_auction_end = 54;
-    component.auction.v1alpha1.ActionDutchAuctionWithdraw action_dutch_auction_withdraw = 55;
+    component.auction.v1alpha1.ActionDutchAuctionWithdrawView action_dutch_auction_withdraw = 55;
```
erwanor, line 228: @jessepinho oversight on our end, but this should fix it.

conorsch:
Does this break protos that were released, or just what's on main?

conorsch (approved):

erwanor:
we publish to the BSR on main right? the blast damage should only be a branch in the web repo though

cratelyn:
checking my understanding: this is _not_ a `consensus-breaking` change because it only performs breaking changes to `ActionView`, rather than `Action`.

cratelyn (approved):
this looks good to me! :hammer:

erwanor:
@cratelyn that's right, however this would be a pretty bad client breaking change

erwanor:
Let's break some protos, shall we 😎

PR #4296, opened 2024-05-01
penumbra-proto: fix comments on `StateReadProto` nv methods

erwanor:
## Describe your changes

This fix incorrect comments on `StateReadProto` nv methods.

  > Comments


cratelyn (approved):
thanks for doing this!

PR #4298, opened 2024-05-01
docs: add security email info to readme

conorsch:
## Describe your changes
We have an email alias set up for team leads, so that external parties have a single entrypoint to disclose security-related issues. Once we settle on language here, we can duplicate this notice in other Penumbra-related repositories.

  > docs-only


cratelyn (commented):
:heavy_check_mark: this looks good, from a change perspective.

i'll let @aubrika own the formal approval, as this is a process change.

PR #4300, opened 2024-05-01
view: port wasm planner to core

TalDerei:
## Describe your changes

This PR adds a port of the wasm planner, integrating it within the `penumbra-view` crate.

**`SwapClaim` handling**
In the process of porting this PR, we found a bug in the handling of `SwapClaim` actions with the wasm planner.

`SwapClaim` actions produce surplus value from their prepaid fees) that must be released so that the transaction balance to zero. In principle, it should be possible to release this value into an output note. This isn't the approach we took with this PR. Instead, we release it to the transaction fee, providing a base for us to improve on later.

**Voting regression**
The original WASM planner did not handle delegator/validator voting. This PR does not yet attempt to do so either. The pcli integration test for delegator voting is commented out.

References https://github.com/penumbra-zone/penumbra/issues/4081

> Client code


crates/view/src/planner.rs
```diff
@@ -550,172 +657,92 @@ impl<R: RngCore + CryptoRng> Planner<R> {
...
+                .push(SpendPlan::new(&mut OsRng, note.note, note.position).into());
+
+            // Recompute the change outputs, without accounting for fees.
+            self.refresh_change(change_address.clone());
+            // Now re-estimate the fee of the updated transaction and adjust the change if possible.
+            let fee = self.fee_estimate(&self.gas_prices, &self.fee_tier);
+            self.adjust_change_for_fee(fee);
+
+            // Need to account to balance after applying fees.
+            self.balance = self.balance_with_fee_estimate(&self.gas_prices, &self.fee_tier);
```
erwanor, line 720: ^

TalDerei: can we remove this?

crates/view/src/planner.rs
```diff
@@ -550,172 +693,116 @@ impl<R: RngCore + CryptoRng> Planner<R> {
...
+                )
+                .into());
+            };
+            self.actions
+                .push(SpendPlan::new(&mut OsRng, note.note, note.position).into());
+
+            // Recompute the change outputs, without accounting for fees.
+            self.refresh_change(change_address.clone());
+            // Now re-estimate the fee of the updated transaction and adjust the change if possible.
+            let fee = self.fee_estimate(&self.gas_prices, &self.fee_tier);
```
TalDerei, line 752: shouldn't this just be `fee =`

TalDerei:
I reviewed the surplus logic again, and checks out to me. Merge on your approval ~

PR #4301, opened 2024-05-01
docs(guide): simplify url routes

conorsch:
## Describe your changes
Modifies the firebase hosting config to remove the `/<version>/` prefix from the guide URLs, which has only ever been `/main/`, due to #6. Also includes rewrites so that historical URLs with the `/main/` prefix will still be honored, and redirected to the appropriate location that's live in the docs.

Squirreling away this one-liner for generating a "sitemap" of all URLs:

    curl -s http://localhost:5000 \
        | rg -i 'class="chapter"' \
        | perl -npE 's/href=/\nhref=/g ' \
        | perl -nE  '/href="([\w.\/]+)"/g and say $1' > sitemap.txt

which was helpful in creating the redirects.

Removes the mdboook-only meta-refresh redirects added in #4232, because the browser UX was atrocious. Refs #3417.

  > docs/ci-only, no code changes
  
## Testing

In order to aid in review, I've deployed this new config from the CLI, so it's live right now. Of particular importance is ensuring that historical URLs published in various places like the README and Discord are still working. Some examples to the old-style URLs:

* https://guide.penumbra.zone/main/pd.html
* https://guide.penumbra.zone/main/pcli.html
* https://guide.penumbra.zone/main/pcli/update.html

All of those should get redirected to the correct locations in the docs, without the `/main/` prefix.

### DIFF
--- a/.github/workflows/notes.yml
+++ b/.github/workflows/notes.yml
@@ -36,15 +36,8 @@ jobs:
         run: cargo +nightly install mdbook mdbook-katex mdbook-mermaid mdbook-linkcheck
 
       - name: Build software guide
-        run: cd docs/guide && mdbook build
-
-      - name: Move software guide to subdirectory
-        run: |
-          cd docs/guide
-          rm -rf firebase-tmp
-          mkdir firebase-tmp
-          mv book/html firebase-tmp/${{ github.event.inputs.image_tag || github.ref_name }}
-          tree firebase-tmp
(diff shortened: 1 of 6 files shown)


hdevalence (approved):
The original rationale was to be able to generate copies of the docs for each version. But we never did this, and it's probably confusing to have multiple (outdated) versions of the guide.

PR #4315, opened 2024-05-03
Add stateful checks for chain ID and expiry height

### LINKED ISSUE #4308: Consensus does not check transaction chain ID or expiry
**Describe the bug**

A transaction which contains a mismatched chain ID or an expired expiry height is currently accepted by the chain.

**To Reproduce**

I have not reproduced this on a testnet, but there is currently, to my ability to determine, no code which checks these properties. It should be in the historical check method right next to the FMD parameters check, per @erwanor.

**Expected behavior**

If a transaction is submitted with **exactly the current chain ID**, it should be accepted. Otherwise, it should be rejected. Note that the empty chain ID string `""` *should not* be treated specially here. The empty chain ID should always cause a transaction to fail, because there will never be a chain with an empty chain ID. This is distinct from the current RPC behavior of "failing open" and accepting requests with an empty chain ID, so that you can request information without first acquiring the chain ID.

If a transaction is submitted with a zero expiry height or an expiry height greater than the current height, it should be accepted. Otherwise, it should be rejected.

**Additional context**

Expiry height was at one point going to be converted to time-based, but [this was closed out because of proto stabilization](https://github.com/penumbra-zone/penumbra/issues/1915). It's my opinion that we should enforce the field, rather than leaving it in a state that is documented as being enforced, but not enforced. The alternative would be to change the documentation to deprecate it.

This is technically a consensus-breaking change because it makes previously-valid transactions invalid, although no client tooling currently produces such transactions.

plaidfinch:
## Describe your changes

Resolves #4308.

**This is a consensus-breaking change.**

### DIFF
--- a/crates/core/app/tests/app_can_define_and_delegate_to_a_validator.rs
+++ b/crates/core/app/tests/app_can_define_and_delegate_to_a_validator.rs
@@ -35,8 +35,11 @@ async fn app_can_define_and_delegate_to_a_validator() -> anyhow::Result<()> {
     let storage = TempStorage::new().await?;
 
     // Configure an AppState with slightly shorter epochs than usual.
-    let app_state =
-        AppState::Content(genesis::Content::default().with_epoch_duration(EPOCH_DURATION));
+    let app_state = AppState::Content(
+        genesis::Content::default()
+            .with_epoch_duration(EPOCH_DURATION)
+            .with_chain_id(TestNode::<()>::CHAIN_ID.to_string()),
+    );
 
     // Start the test node.
     let mut node = {
--- a/crates/core/app/src/action_handler/transaction.rs
+++ b/crates/core/app/src/action_handler/transaction.rs
@@ -13,7 +13,10 @@ use super::AppActionHandler;
 mod stateful;
 mod stateless;
 
-use self::stateful::{claimed_anchor_is_valid, fee_greater_than_base_fee, fmd_parameters_valid};
+use self::stateful::{
+    chain_id_is_correct, claimed_anchor_is_valid, expiry_height_is_valid,
+    fee_greater_than_base_fee, fmd_parameters_valid,
+};
 use stateless::{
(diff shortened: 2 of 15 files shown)


erwanor:
Thanks a lot for noticing this, and taking the time to make a PR to fix it, I am very appreciative that this was caught. ACK, I'm going to push this across the finish line by fixing the test suite and destructuring `TransactionParameters` so that something breaks loudly if an addition is made.

cratelyn (approved):

crates/core/app/tests/app_can_define_and_delegate_to_a_validator.rs
```diff
@@ -35,8 +35,11 @@ async fn app_can_define_and_delegate_to_a_validator() -> anyhow::Result<()> {
     let storage = TempStorage::new().await?;
 
     // Configure an AppState with slightly shorter epochs than usual.
-    let app_state =
-        AppState::Content(genesis::Content::default().with_epoch_duration(EPOCH_DURATION));
+    let app_state = AppState::Content(
+        genesis::Content::default()
+            .with_epoch_duration(EPOCH_DURATION)
+            .with_chain_id(TestNode::<()>::CHAIN_ID.to_string()),
```
hdevalence, line 41: Just for clarification: was this necessary to make the change? It seems desirable to allow the chain ID to be `""` for testing, and it seems like there's no downside since any real network _will_ have a chain ID (which wherefore won't match).

hdevalence (approved):

hdevalence:
Let's wait to merge this until after resolving our current planner issues.

crates/core/app/tests/app_can_define_and_delegate_to_a_validator.rs
erwanor: It's not necessary, and an empty chain id would work!

Issue #4155, opened 2024-04-03
fees: add tests to validate that non-zero fees are working


TalDerei:
Create unit tests to verify correctness of **non-zero** fees in the context of properly balancing transactions. There are a few cases to consider:

- single large note > amount + cumulative fees --> success
- single small note < amount + cumulative fees --> failure
- multiple smaller notes > amount + cumulative fees --> success
- multiple smaller notes < amount + cumulative fees --> failure
- exact note for amount + another large note > amount + cumulative fees --> success
- exact note for amount + another small note < amount + cumulative fees --> failure

TalDerei:
tagging https://github.com/penumbra-zone/web/pull/690 for visibility.

PR #4316, opened 2024-05-04
fees: add unit tests

TalDerei:
## Describe your changes
I modified the unit tests for fees according to the updated planner logic. 

References https://github.com/penumbra-zone/penumbra/issues/4155

### DIFF
--- a/crates/view/src/planner.rs
+++ b/crates/view/src/planner.rs
@@ -843,3 +843,178 @@ impl<R: RngCore + CryptoRng> Planner<R> {
         Ok(())
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use penumbra_asset::STAKING_TOKEN_ASSET_ID;
+    use penumbra_keys::keys::{Bip44Path, SeedPhrase, SpendKey};
+    use rand_core::OsRng;
+
+    #[test]
+    fn test_sufficient_funds_for_non_zero_fees_in_transaction() {
+        let rng = OsRng;
+        let seed_phrase = SeedPhrase::generate(rng);
+        let sk = SpendKey::from_seed_phrase_bip44(seed_phrase, &Bip44Path::new(0));
+        let fvk = sk.full_viewing_key();
+        let (sender, _dtk) = fvk.incoming().payment_address(0u32.into());
+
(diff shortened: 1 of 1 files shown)


crates/view/src/planner.rs
```diff
@@ -843,3 +843,178 @@ impl<R: RngCore + CryptoRng> Planner<R> {
...
+            );
+
+            // Recompute the change outputs, without accounting for fees.
+            planner.refresh_change(sender.clone());
+
+            // Now re-estimate the fee of the updated transaction and adjust the change if possible.
+            let fee = planner.fee_estimate(&planner.gas_prices, &planner.fee_tier);
+            planner.adjust_change_for_fee(fee);
+
+            if planner.balance().is_zero() {
```
erwanor, line 1008: Isn't this cheating? The plan should balance once we run out of required notes if we have correctly refreshed the change notes

crates/view/src/planner.rs
```diff
@@ -843,3 +843,178 @@ impl<R: RngCore + CryptoRng> Planner<R> {
...
+        let amount = Value {
+            amount: 3u64.into(),
+            asset_id: *STAKING_TOKEN_ASSET_ID,
+        };
+
+        planner.output(amount, reciever);
+        planner.actions = planner.plan.actions.clone();
+
+        let mut iterations = 0usize;
+        while let Some(_required) = planner
```
erwanor, line 906: I think adding tests is a great idea, but my sense is that rather than replicating the logic (which has changed since this PR was opened) we should either:
- refactor the core planning logic into a pure function
- check in on #3913 to see if the plan is to be able to build a mock view client at some point, in which case we can use `Planner::plan` directly

erwanor (changes requested):
Thanks for adding tests, requesting changes because the planner api has changed since this PR was opened so this would break CI.

crates/view/src/planner.rs
erwanor: My intuition is that it's probably best to wait a little longer until we try to test this, we'll definitely have less turbulence in the planner in a couple weeks time or less

TalDerei:
cc @cratelyn closing out since it's fallen out of sync with the latest round of refactors. This can be a reference for eventually integrating into the mock consensus work-stream.

PR #4317, opened 2024-05-04
app: fold ibc stateful/historical check into their AHs

erwanor:
## Describe your changes

This is a mechanical refactor to fold the ibc/ics20 historical checks into a more visible place (their action handler implementations).

  > Mechanical refactor


avahowell (approved):
LGTM, we could have another check to verify that the `outbound_ibc_enabled` param is independent of the component being enabled, but this is much harder to test currently.

PR #4319, opened 2024-05-04
gov: simplify governance-controlled parameter changes

hdevalence:
## Describe your changes

This commit refactors the governance component to dramatically simplify the way
that governance-controlled parameter changes are performed.  Previously, the
governance component duplicated all of the parameter data of all of the other
components, which had to be kept in sync in a complicated and error-prone way.

Now, parameter changes are specified and performed as key-value edits:

- A parameter change proposal specifies a list of `(component, key, value)` pairs
- The governance component can apply these without knowledge of other components' structures by operating on a `serde_json::Value`
- The application can validate the changed parameters by deserializing a complete `AppParameters` from ProtoJSON.

The logic around applying these changes is also dramatically simplified. The
top-level app is responsible for all parameter changes. Components do not need
to track their own individual "dirty bit" for parameter changes. The governance
component can report to any other component when a parameter change was
scheduled to occur.

Another crucial improvement is that this code allows **different** parameters to be changed by **different** proposals, because changes are made to individual parameters, not to entire subsections.

In followup work, we should change the `pcli` interface to work with JSON
proposal data and delete the `ProposalToml` entirely.

None

  > This commit changes how the governance system works. However, no migrations are necessary, because it ignores any previous data and writes new data in different places.


cratelyn (approved):
pending ci, this looks good to me. great work! i'm looking forward to talking about this more next week.

hdevalence:
It's not actually possible to make a delegator vote to test it, however, so it's not yet good to merge. The planner needs to be fixed to make this possible, this was a recent regression. I am working on that now.

hdevalence:
I restructured the `Planner` implementation in this repo to match the refactoring in the web repo:

The part of the web planner refactoring we wanted to bring over to the core
repo was the creation of a separate `ActionList` modeling a list of actions
that may or may not be balanced.  This commit pulls the planner into that
shape, and extends the `ActionList` design found in the web code in a few ways:

- The `ActionList` tracks a target fee internally, rather than leaving that to
  its caller. This allows us to handle cases like a swap claim, where we want
  to put the prepaid fee released by a swap claim towards the transaction fee,
  and not create dust outputs.

- The `ActionList` has the groundwork for supporting multi-asset fee payments,
  by not assuming that the fee is always in the staking token when accounting
  for fees. Later we can add an asset ID to the `GasPrices` and it should be
  a relatively drop-in extension.

- The `ActionList` can be used to create a `TransactionPlan` directly. In the
  future, we should consider restructuring the `view` crate to have most of
  its current contents behind a `svc` flag, allowing us to depend on the `view`
  crate from the wasm code and have a common `ActionList` impl between them,
  just different code that manipulates it.

Various API cleanups were necessary along the way.

cc @TalDerei -- might be interesting to you, sorry we had confusion about the goals there.

hdevalence:
I tested this code on a local devnet and was able to freely adjust parameters (and had all actions have correctly priced fees, too!).

erwanor:
Delegator/validator voting is supported by the planner thanks to @TalDerei quick follow-up

erwanor:
Ah I hadn’t looked at the planner rerefactor, so I actually don’t know

hdevalence:
> Delegator/validator voting is supported by the planner thanks to @TalDerei quick follow-up 

When I tried using it to exercise parameter changes there were balance errors, but I was able to build further on the cleaned-up code and get through it reasonably easily, so it was a great groundwork.

TalDerei:
Can I propagate the planner changes to the web? Currently reviewing the changeset in this PR

hdevalence:
No, I’d rather make a plan of how to share code and get the plan right first, before we do more work.

Issue #4313, opened 2024-05-03
planner: allow prepaid fees to be collected in outputs


TalDerei:
There's parity between the wasm planner and the core planner, but there are some additional design considerations that have been discussed, and will be noted here for future reference. 

Currently, we release the surplus prepaid swap-claim fee directly into the transaction fee, making the transaction unnecessarily more expensive than it needs to be. We may consider releasing the excess value as a change output note. 

Additionally, there's been discussion of allowing for a specific change address to be specified, and change notes to be paid in any asset denomination. This would be a [UX boost](https://discord.com/channels/824484045370818580/883455571003576320/1233994930620334150) because the current design is more [restrictive](https://discord.com/channels/824484045370818580/883455571003576320/1233957899156586546) because you need `UM` to do anything on the network.

hdevalence:
These are all done except for #4322

PR #4320, opened 2024-05-05
fee: v0 support for multi-asset fees

hdevalence:
## Describe your changes

This commit implements a v0 of multi-asset fees (on the protocol side).
Multi-asset fees are important because they allows users who send tokens into
Penumbra over IBC to start using the chain immediately.

It does this by extending the `GasPrices` struct to include the asset ID of the
fee token the prices are for.  This allows natural handling of multi-asset fees
by just having multiple sets of `GasPrices`.

In the past, we've laid the groundwork for this (by allowing other assets in
the `Fee` structure, even if it defaults to the staking token) but haven't
prioritized it, due to concerns about some of the details of the
implementation, particularly:

A: How does the chain know that the prices of non-native fee tokens are correct?
B: What happens to the non-native fee tokens?
C: What are the privacy implications for users paying with non-native fee tokens?

This design punts on (A) by only allowing non-native fee tokens with prices
explicitly specified using governance. In the future, Penumbra could own DEX as
a price oracle, but this would be better to do after it's live and we have real
data about it, rather than worrying about edge cases now (e.g., if someone can
temporarily manipulate a price for their own trades, that's not a big deal in
itself, but hooking it to the fee mechanism used for resource usage could make
it one).

The approach to (C) is that these explicitly specified prices should be
substantially higher (10x) than those of the native token.  This means that
regardless of price variability, it should remain the case that the native fee
token is cheaper, so that the majority of users will use the same fee token and
their transactions will not be distinguishable.  On the other hand, letting it
be _possible_ to use non native fee tokens means that users who send IBC tokens
into Penumbra can access the protocol, and even if one of their first acts is
to swap some de minimis amount of the native token, they can do that within the
system.

This implementation currently does not properly handle (B), in that it silently
burns all fees, regardless of token.  It is not appropriate to burn any tokens
other than the native token, whose supply is managed by the chain. Instead, the
transaction-level fee check should move into `check_and_execute`, and call a
method on the fee component that records the fee amount in the object store. At
the end of the block, all non-native fees should be swapped into the native
token and then burned.  Emitted events should record:

- an `EventFee` for each transaction, with the transaction's gas cost, the gas
  prices used to compute the base fee, the base fee, and the actual fee;
- an `EventFeeBurn` at the end of the block, with the total amount of native
  fees burned as well as, for each alt fee token used to pay fees in that
  block, how much of the alt token was paid and how much of the native token it
  was traded for.

Client support is still mostly missing, though I have manually tested this on a
local devnet, by

1. Creating the devnet (with a short voting period, configured with `pd testnet generate`)
2. Executing a parameter change proposal to set `dexParams.fixedAltGasPrices` to `"[{\"assetId\":{\"altBaseDenom\":\"ugm\"},\"blockSpacePrice\":\"80000\",\"compactBlockSpacePrice\":\"80000\",\"verificationPrice\":\"80000\",\"executionPrice\":\"80000\"}]"`
3. Hardcoding at the top of `TxCmd::exec` a change to the `gas_prices` to add the `gm` asset ID and multiplying all the prices by 10.

To properly support clients, we need to

- Record the `gas_prices` data for alt fee tokens that is newly included in the
  `CompactBlock` in the view server's storage (probably in a new `gas_prices`
  table that maps a fee token asset ID to the latest price);

- Use that data to populate the view server's gas prices RPC response (instead
  of the current `Vec::new` stub that reports any alt fee tokens as
  non-existing);

- Change the `pcli tx` command so that the existing `--fee-tier` and a new
  `--fee-token` parameters move to `pcli tx`, and apply to all subcommands
  (i.e., `pcli tx --fee-tier ... --fee-token gm send ...`) so that `TxCmd::exec`
  can pull the right `GasPrices` out of the RPC response.

  > this is consensus-breaking, but not state-breaking


TalDerei:
references https://github.com/penumbra-zone/penumbra/issues/4313

crates/core/component/fee/src/component/view.rs
```diff
@@ -20,11 +16,24 @@ pub trait StateReadExt: StateRead {
...
+    }
+
+    /// Gets the current gas prices for alternative fee tokens.
+    async fn get_alt_gas_prices(&self) -> Result<Vec<GasPrices>> {
+        // When we implement dynamic gas pricing, we will want
+        // to read the prices we computed. But until then, we need to
+        // read these from the _fee params_ instead, since those are
+        // the values that will get updated by governance.
+        let params = self.get_fee_params().await?;
+        Ok(params.fixed_alt_gas_prices)
```
hdevalence, line 36: This fixes a critical bug in the original fee component logic, which would prevent fees from ever being changed post-genesis.

erwanor (approved):
ACK, this looks great, testing/stretching manually

PR #4327, opened 2024-05-06
fix(metrics): validator missed blocks, inc -> set

conorsch:
## Describe your changes

We were inappropriately `increment`ing a gauge to track missed blocks, when instead we should have been clobbering the value via `set`, as we know the precise number from the state machine. This fixes a broken metric for missed blocks that was growing too fast, emitting wrong values.

Refs #3746.

_Not_ included in this small patch is a rework of the other validator-related metrics. As described in #3746, we should migrate those metrics to use Events instead. 

  > metrics-only, no logic changes


cratelyn (approved):
thank you for fixing this!

Issue #4323, opened 2024-05-06
ci: automated testing of migration compatibility


conorsch:
**Is your feature request related to a problem? Please describe.**
When preparing a chain upgrade, manual testing of upgrades is an arduous process. 

**Describe the solution you'd like**
We should have an integration test that runs a devnet based on the currently-active testnet, via the most recently released tag, runs smoke tests against it to generate txs, then stops the network, runs the migration, restarts the network, and reruns the smoket ests.

**Describe alternatives you've considered**
Alternatives are manual testing, which is both slow and error-prone. Longer-term we want a capable "sudo mode" for testing upgrades, which is tracked in #4265.

**Additional context**
 We'll need to be careful about endpoint compatibility: we cannot run an older version of `pd` and run the most recent smoke tests against it, because the view server implementations will not be compatible. We can sidestep this by running the smoke tests from the tagged release. Ideally, we'd be able to swap out the path to binaries within the tests via an env var to make it a bit more to test prior versions, without rebuilding from source every time.

conorsch:
There's a spike on an overhaul of the smoke-test logic here https://github.com/penumbra-zone/penumbra/pull/4324 which provides a nice shape to extend into migration-testing.

hdevalence:
> We'll need to be careful about endpoint compatibility: we cannot run an older version of pd and run the most recent smoke tests against it, because the view server implementations will not be compatible. 

Can you elaborate on this a little? Isn't our expectation that clients should work across upgrade boundaries?

conorsch:
>  Can you elaborate on this a little? Isn't our expectation that clients should work across upgrade boundaries?

If you try to run a client from current main against a public testnet endpoint, you'll see in incompatibility message related to ongoing auction work:

```
❯ git rev-parse HEAD
7854a5fc561e2e3f514421c3ea97c80cea5a673e

❯ cargo run -q --release --bin pcli -- view sync
Error: proto response missing auction params
```

Those same dependencies carry over into the integration tests.

conorsch:
Pushed a draft PR with a spike on local testing of migration logic, that can be promoted to a CI job once it's solid. Got surprised by a proto incompat error that may be spurious, so I'm going to run through the upgrade process manually to sanity-check that the scripting order is sound.

hdevalence:
@conorsch https://github.com/penumbra-zone/penumbra/pull/4339 will simplify things a great deal

hdevalence:
[...]

Issue #4265, opened 2024-04-24
Implement `sudo` mode tooling for testing


aubrika:
It would be cool to have a docker command to set up a validator set with known consensus keys for testing purposes.

Questions

- How do we establish continuity between the previous validator set and the generated one?

hdevalence:
We don't have bandwidth to do this pre-mainnet. This is not a substitute for the missing upgrade smoke test.

PR #4337, opened 2024-05-06
tests(pd): automated migration testing

conorsch:
## Describe your changes

Adds intra-CI standalone testing of migration behavior. This can be run locally on developer workstations, and also in CI. The job is currently taking >20m, and requires a lot of disk space, but it's worth it for the assurance.

Building on the smoke-test rewrite to use process-compose, let's script the migration process, so that we can test
current HEAD of the monorepo against a prior tagged version, and validate that necessary migrations are in place.

One possible approach is to fetch prebuilt binaries from uploaded artifacts on Github. That's fine for `pd`, but doesn't work for running the smoke tests, due to client/server incompatibility. Therefore we'll clone the entire repo in a git-ignored subdir, and build the old binaries there. Heavy, but reliable.

Adds a new rust crate, strictly for running the migration-test suite of integration tests, which is very similar in nature to the already-existing network-integration tests AKA smoke tests. Copy/pastes a lot of code from the smokes, which we can always factor out into reusable utils, but not bothering with that right now.

Refs #4323.

  > Testing-only, no changes to application logic.
  
## Review, running locally

Check out this branch, and run `just migration-test`. It'll take a while to build! The test as written duplicates the local repo, so plan on ~100GB of disk space utilization.


hdevalence:
> Confoundingly, I'm getting a proto incompatible error, about missing AuctionParams.

I think this was an actual bug #4338. We shouldn't assume the migrations actually work until we do the testing. See also #4340

conorsch:
Marking as ready for review, so I can stop rebasing it. :sweat_smile: The immediate consequence of merging this PR is that CI runs will take >20m again, but I'm happy to dial that back and rely on the cron schedule. For now, though, having it run on every PR will give us confidence about the state of migrations as we push toward #4497.

deployments/compose/process-compose-migration-test-1.yml
cratelyn, line 1: it's a bit difficult to understand what this is testing. my understanding is that this is running the smoke tests on the v0.76.0 release? i don't think i am connecting the dots to how this is exercising migration logic.

it'd be helpful to (a) name these files more commucatively (_possibly via putting them in a folder together_), and (b) putting a comment at the top of the file, in place of the boilerplate there right now. this isn't a configuration for running migration tests, ~~it is _a_ migration test, aiui.~~

deployments/compose/process-compose-migration-test-2.yml
cratelyn, line 1: ditto, comment on `deployments/compose/process-compose-migration-test-1.yml`

crates/test/migration-test/tests/network_integration.rs
```diff
@@ -0,0 +1,403 @@
...
+}
+
+// Needed functions:
+//
+//   get_current_height()
+//   generate_proposal
+//   submit_proposal
+//
+//
+
```
cratelyn, line 185:
```suggestion
```

this looks like a dev comment that snuck through

crates/test/migration-test/tests/network_integration.rs
```diff
@@ -0,0 +1,403 @@
...
+            "--source",
+            "2",
+        ])
+        .timeout(std::time::Duration::from_secs(TIMEOUT_COMMAND_SECONDS));
+    tracing::info!(?delegate_cmd, "running delegation command");
+    delegate_cmd.assert().success();
+    // Wait for bonding, otherwise delegation isn't active.
+    tracing::info!("waiting for stake to bonding...");
+    thread::sleep(*EPOCH_DURATION);
+}
```
cratelyn, line 218: :grimacing: is this depending on cargo test running tests in a certain order? that sounds like a recipe for very confusing errors in the future, i'm very hesitant to follow that course if so. if another test is dependent on side-effects of this test first, we should consolidate them into a single test.

crates/test/migration-test/tests/network_integration.rs
```diff
@@ -0,0 +1,403 @@
...
+// Needed functions:
+//
+//   get_current_height()
+//   generate_proposal
+//   submit_proposal
+//
+//
+
+/// Look up top validator and delegate `TEST_DELEGATION` to it.
+#[cfg_attr(not(feature = "migration-test"), ignore)]
```
cratelyn, line 187: the line as is, says "_conditionally apply the `#[ignore]` attribute if the `migration-test` feature is **not** enabled._"

we can instead say "_this is active if the `migration-test` feature is enabled_"

```suggestion
#[cfg(feature = "migration-test")]
```

crates/test/migration-test/tests/network_integration.rs
```diff
@@ -0,0 +1,403 @@
...
+            "tx",
+            "undelegate-claim",
+        ])
+        .timeout(std::time::Duration::from_secs(TIMEOUT_COMMAND_SECONDS));
+    tracing::info!(?undelegate_claim_cmd, "claiming");
+    undelegate_claim_cmd.assert().success();
+    tracing::info!("success!");
+    sync(&tmpdir).expect("sync is successful");
+}
+
```
cratelyn, line 283:
```suggestion
```

this is not used, so it should be removed.

crates/test/migration-test/tests/network_integration.rs
```diff
@@ -0,0 +1,403 @@
...
+/// TOML for an "upgrade-plan" governance proposal.
+// Intentionally avoiding importing this type to adhere to strict
+// CLI interfacts for the pcli binaries.
```
cratelyn, line 286:
```suggestion
// CLI interfacts for the pcli binaries.
```

i think a sentence got garbled here!

crates/test/migration-test/tests/network_integration.rs
```diff
@@ -0,0 +1,403 @@
...
+#[derive(Deserialize, Serialize)]
+struct UpgradePlan {
+    id: u64,
+    title: String,
+    description: String,
+    kind: String,
+    height: u64,
+}
+
+#[cfg_attr(not(feature = "migration-test"), ignore)]
```
cratelyn, line 296:
```suggestion
#[cfg(feature = "migration-test")]
```

ditto, another conditional compilation attribute we can simplify

deployments/compose/process-compose-migration-test-1.yml
cratelyn: ...as i've read this more, i see these are phases. i'll mark this as resolved, but i did find coming to this understanding to be difficult. :sweat:

deployments/scripts/migration-test
```diff
@@ -0,0 +1,169 @@
...
+#   * [x] make sure devnet has very fast voting
+#   * [x] runs smoke tests against devnet, to generate txs
+#   * [x] initialize smoke-test pcli wallet locally
+#   * [x] submit governance proposal for chain halt
+#   * [x] vote for proposal to pass
+#   * [x] wait for halt
+#   * [x] stops devnet ~ ends phase 1
+#   * [x] migrates devnet via latest (i.e. HEAD) pd version ~ begins phase 2
+#   * [x] restarts devnet
+#   * [ ] re-runs smoke tests, from latest version, to validate
```
cratelyn, line 15: perhaps we can pick a subset of tests, so we don't end up running the entire smoke test suite twice. _or_, perhaps we should do it the other way around? we know previous releases work, but we're most interested in tests passing after the migration.

deployments/scripts/generate-config-for-bootstrap-local-node
```diff
@@ -0,0 +1,38 @@
...
+
+
+# Set default values for common pd args, nudging them up 100 ports.
+# This is to avoid the standard defaults, so that we can run multiple
+# nodes on localhost at the same time.
+PENUMBRA_PD_GRPC_BIND="${PENUMBRA_PD_GRPC_BIND:-127.0.0.1:8180}"
+PENUMBRA_PD_ABCI_BIND="${PENUMBRA_PD_ABCI_BIND:-127.0.0.1:26758}"
+PENUMBRA_PD_TM_RPC_BIND="${PENUMBRA_PD_TM_RPC_BIND:-127.0.0.1:26757}"
+PENUMBRA_PD_TM_P2P_BIND="${PENUMBRA_PD_TM_P2P_BIND:-127.0.0.1:26756}"
+PENUMBRA_PD_METRICS_BIND="${PENUMBRA_PD_METRICS_BIND:-127.0.0.1:9100}"
```
cratelyn, line 14: thought for another day, but it seems like we would get a lot of mileage out of putting these sorts of things on unix domain sockets, in a temp directory, so we aren't left doing math about which ports are available for which nodes.

that's totally not a blocker, just noting that thought now.

crates/test/migration-test/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+serde_with = {workspace = true, features = ["hex"]}
+toml = {workspace = true, features = ["preserve_order"]}
+tracing = {workspace = true}
+tracing-subscriber = {workspace = true, features = ["env-filter", "ansi"]}
+
+[dev-dependencies]
+assert_cmd = {workspace = true}
+predicates = "2.1"
+regex = {workspace = true}
+tempfile = {workspace = true}
```
cratelyn, line 35:
```suggestion
[dev-dependencies]
anyhow = { workspace = true }
assert_cmd = { workspace = true }
directories = { workspace = true }
once_cell = { workspace = true }
penumbra-keys = { workspace = true, default-features = false }
predicates = "2.1"
regex = { workspace = true }
serde = { workspace = true, features = ["derive"] }
serde_json = { workspace = true }
serde_with = { workspace = true, features = ["hex"] }
tempfile = { workspace = true }
toml = { workspace = true, features = ["preserve_order"] }
tracing = { workspace = true }
tracing-subscriber = { workspace = true, features = ["env-filter", "ansi"] }
```

these should all be [`dev-dependencies`](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#development-dependencies), since this crate only contains tests.

otherwise, `cargo build --package migration-test` will mean building all of these crates, despite the `main.rs` being empty and the `--features migration-test` not being provided

crates/test/migration-test/tests/network_integration.rs
cratelyn: since we aren't making any assertions about unbonding, we can save time by not waiting around.

```suggestion
}
```

cratelyn (approved):
i'm going to somewhat cautiously hit the "_approve_" button here.

we should have automated migration testing, and i want that sooner than later, but.. i'm a little nervous about how these are shaped. they feel somewhat heavy on boilerplate yaml, and there's a lot of shell glue surrounding the actual test.

i think a lot of the logic _here_ could also be (_and is, in the case of delegating_) covered by mock consensus tests, but the broader vision of having **integration** tests that run a real CometBFT node feel topically relevant for migrations.

i've left a smattering of comments below. i'm happy to see this merge before those things are addressed, being acutely aware that this life easier while managing upgrades :heart: ...but maybe file a tracking ticket(s) for following up on these points.

conorsch:
Downgrading to draft until I address the review comments. Overall, the shape of this approach is rather heavy. Will continue to discuss testing options with @cratelyn, so that we maximize what's possible with mock-consensus, but still retain integration tests for migration behavior. Notably, this PR does _not_ catch bugs in restarting a multi-validator setup post-upgrade, e.g. #4508.

conorsch:
This draft work successfully caught several early-stage issues with migrations, but only of the simpler variety: when migrations fail to apply to halted state, for instance. More complex situations such as the network failing to come back up in multi-validator setups due to app logic problems were not caught by these changes, and instead required a more complicated setup that leveraged an online multi-validator setup.

Given the CI changes in #4678, we should think hard about what an adequate testing environment looks like for this environment. For the near future, I recommend we stick to creating devnets and performing candidate migrations against them, ensuring that they behave well when restarted across the upgrade boundary. Longer-term, there's detail in #4323 and also #4265 for how to approach this problem comprehensively.

PR #4338, opened 2024-05-07
pd(migrate): add auction parameters to chain state for `Testnet74`

erwanor:
## Describe your changes

@hdevalence noticed that we were missing a migration for the `AuctionParameters`. This PR adds a migration to write the parameters to the app state.

This was a miss in #4259

  > This is a migration


hdevalence (approved):
Happy to merge optimistically.

PR #4339, opened 2024-05-07
pd: improve migration UX

hdevalence:
## Describe your changes

Just have one command that does the right thing.

  > not consensus-breaking, changes upgrade process


cratelyn (commented):
:heavy_check_mark: i don't want to take approval rights from @conorsch, but this looks like a great ergonomic improvement. it's exciting to see a sharp edge (_genesis time_) sanded down. LGTM!

conorsch (approved):
Excellent, now that https://github.com/penumbra-zone/penumbra/pull/4338 is both confirmed and fixed, I'll continue with upgrade testing using this new UX. It's a huge improvement over the old workflow!

Issue #4332, opened 2024-05-06
Move `ActionList` from `view` crate into `transaction` crate


hdevalence:
**Is your feature request related to a problem? Please describe.**

The `ActionList` introduced in #4319 represents an in-progress list of actions that will be turned into a `TransactionPlan`, and provides helper methods for balancing the value of those actions, making it the core of a planner implementation. It should be shared between both planners.

**Describe the solution you'd like**

Remove the current `view/src/planner/action_list.rs`, redistributing its contents:
- the `prioritize_and_filter_spendable_notes` function should move up into the parent module of the `view` crate
- everything else should move into `transaction/src/action_list.rs` and be exposed as an `action_list` submodule of the `transaction` crate

Then update the existing code to match the changed imports.

**Additional context**

This allows the web code to use the same `ActionList`.

hdevalence:
@TalDerei the call to canonicalize the action ordering would naturally fit at the end of the `ActionList::into_plan` method

TalDerei:
> @TalDerei the call to canonicalize the action ordering would naturally fit at the end of the `ActionList::into_plan` method xref #3467

agreed, I added it right after the `ActionList::into_plan` in that PR, but I'll move it into the method.

hdevalence:
Sounds great. If we do this part, the next time the web repo updates its rust deps, we'll be able to commonize the fee handling behavior between the two planner implementations.

TalDerei:
@hdevalence this needs to be prioritized to fix swap-claim bug identified in https://github.com/penumbra-zone/web/issues/1054. Then we can propagate the changes in https://github.com/penumbra-zone/web/issues/876.

hdevalence:
sounds great, want to take it on?

PR #4341, opened 2024-05-07
fix(dex): migration for dex trait proto encodings

### LINKED ISSUE #4340: Chain halt during swap execution on post-upgraded devnet
**Describe the bug**

On a local devnet on which I performed a chain upgrade (this may or may not be relevant), I saw a chain halt while swapping:
```
2024-05-07T03:02:26.618921Z  INFO abci:EndBlock{height=704}: penumbra_app::server::consensus: ending block height=704
thread 'tokio-runtime-worker' panicked at /Users/hdevalence/code/penumbra/penumbra/crates/core/component/dex/src/component/router/fill_route.rs:502:14:
writing to storage should not fail: could not decode proto from bytes

Caused by:
    failed to decode Protobuf message: invalid tag value: 0
thread 'tokio-runtime-worker' panicked at /Users/hdevalence/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tower-abci-0.11.1/src/v037/server.rs:179:75:
called `Result::unwrap()` on an `Err` value: ActorHungUp
```

hdevalence:
## Describe your changes

Fixes a chain halt due to changes introduced in #4188.

Previously, aggregate liquidity lookups in non-verifiable storage (i.e. "how much liquidity is available for trading pair `(A, B)`?") were stored as raw big-endian encoded values. In #4188 the DEX state accessors were refactored and part of that was adjusting the accessors for aggregate liquidity lookups to use `nonverifiable_put`/`nonverifiable_get` instead of the `_raw` methods they were using previously.

## Testing

**Manually testing the values were changed appropriately is actually quite difficult currently as `pcli` doesn't support querying values from non-verifiable storage, and returned values need to be parsed correctly into the `Amount` data type.**

- Create a devnet on v0.73.1
- Watch for changes to the aggregate liquidity prefix: `cargo run --release --bin pcli -- query watch --nv-key-regex 'dex/ab/.*'`
- Submit some liquidity positions. Note down the aggregate liquidity data
- Perform migration to this branch: delegate, propose, vote, halt, migrate, restart network
- The chain will halt if the migration doesn't properly change the encoding of the values. The first test is simply to ensure that the chain can run post-migration.
- A further test is to verify that the aggregate liquidity values remain the same post-migration. Query again and compare

Closes #4340

### DIFF
--- a/crates/bin/pd/src/migrate/testnet74.rs
+++ b/crates/bin/pd/src/migrate/testnet74.rs
@@ -37,11 +37,40 @@ async fn fix_arb_execution_outputs(delta: &mut StateDelta<Snapshot>) -> anyhow::
     Ok(())
 }
 
+/// Update base liquidity index values to be proto-encoded. Previously they were stored as big-endian
+/// encoded amounts, but in https://github.com/penumbra-zone/penumbra/pull/4188 they were changed
+/// to be proto-encoded.
+///
+/// This will rewrite all values under the `dex/ab/` prefix to be proto-encoded.
+async fn rewrite_base_liquidity_indices(delta: &mut StateDelta<Snapshot>) -> anyhow::Result<()> {
+    let prefix_key = "dex/ab/".as_bytes();
+    tracing::trace!(prefix_key = ?EscapedByteSlice(&prefix_key), "updating base liquidity index values");
+    let mut liquidity_stream = delta.nonverifiable_prefix_raw(&prefix_key).boxed();
+
+    while let Some(r) = liquidity_stream.next().await {
+        let (key, raw_amount): (Vec<u8>, Vec<u8>) = r?;
+        tracing::info!(?key, raw_amount = ?EscapedByteSlice(&raw_amount), "migrating base liquidity index entry");
+
+        let amount = Amount::from_be_bytes(raw_amount.as_slice().try_into()?);
+
+        // Store the correctly formatted new value:
+        delta.nonverifiable_put(key.clone(), amount);
+        tracing::info!(
+            key = ?EscapedByteSlice(&key),
+            raw_amount = ?EscapedByteSlice(&raw_amount),
+            ?amount,
+            "updated base liquidity index"
+        );
+    }
+
+    Ok(())
+}
+
 /// Update the ordering of liquidity position indices to return in descending order (see #4189)
 ///
 /// Lookups for liquidity positions based on starting asset were ordered backwards
 /// and returning the positions with the least liquidity first. This migration
-/// needs to modify the keys stored under the JMT `dex/ra/` prefix key to reverse
+/// needs to modify the keys stored under the nonverifiable `dex/ra/` prefix key to reverse
 /// the ordering of the existing data.
 async fn update_lp_index_order(delta: &mut StateDelta<Snapshot>) -> anyhow::Result<()> {
     let prefix_key = "dex/ra/".as_bytes();
@@ -81,11 +110,13 @@ async fn update_lp_index_order(delta: &mut StateDelta<Snapshot>) -> anyhow::Resu
 /// This migration script is responsible for:
 ///
 /// - Updating the ordering of liquidity position indices to return in descending order (see #4189)
-///     * JMT: `dex/ra/`
+///     * nonverifiable: `dex/ra/`
 /// - Updating arb execution output amounts to include the input amount instead of reporting only profit (see #3790)
 ///     * JMT: `dex/arb_execution/`
 /// - Add `AuctionParameters` to the chain state
 ///     * JMT: `dex/auction_parameters`
+/// - Update the base liquidity index values to be proto-encoded (see #4188)
+///     * nonverifiable: `dex/ab/`
 pub async fn migrate(
     path_to_export: PathBuf,
     genesis_start: Option<tendermint::time::Time>,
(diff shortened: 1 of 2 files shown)


conorsch:
We merged #4339, so this can be rebased on latest main.

conorsch:
Lots of churn on main, including a fix for https://github.com/penumbra-zone/penumbra/issues/4344. Let's keep this rebased. Locally I'll be cross-checking via #4332. Looks like we're within striking distance on a passing migration test end-to-end.

hdevalence:
To clarify my intent with this PR wasn't that it should become a fix, we should throw this code away — I just wanted to share context while writing up the problem for others to pick up.

conorsch:
Fair, we could just as easily open a new PR for the fix commit. But since it's already here, I think "taking over" this PR and using it to close is good enough. Up to @zbuc: if the debugging commit is no longer useful, we can snip it out pronto.

zbuc:
> Fair, we could just as easily open a new PR for the fix commit. But since it's already here, I think "taking over" this PR and using it to close is good enough. Up to @zbuc: if the debugging commit is no longer useful, we can snip it out pronto.

I removed the debug trace statements, I say using this PR and squashing the commits is fine.

hdevalence:
sounds good, just wanted to clarify intent

crates/bin/pd/src/migrate/testnet74.rs
```diff
@@ -37,11 +37,40 @@ async fn fix_arb_execution_outputs(delta: &mut StateDelta<Snapshot>) -> anyhow::
...
+        tracing::info!(
+            key = ?EscapedByteSlice(&key),
+            raw_amount = ?EscapedByteSlice(&raw_amount),
+            ?amount,
+            "updated base liquidity index"
+        );
+    }
+
+    Ok(())
+}
```
conorsch, line 67: Is this migration fn idempotent? We've been telling folks our migrations are idempotent, but not sure that's the case for this one. Zooming out, the changes in #4339 definitely aren't idempotent, so not blocking on this.

crates/bin/pd/src/migrate/testnet74.rs
```diff
@@ -117,6 +148,9 @@ pub async fn migrate(
         // Write auction parameters
         write_auction_parameters(&mut delta).await?;
 
+        // Rewrite base liquidity indices as proto-encoded
+        rewrite_base_liquidity_indices(&mut delta).await?;
+
```
conorsch, line 153: When we circle back to addressing whether or not we want to enforce that all migrations are idempotent, we can sprinkle in some anyhow::Context usage on these calls. The tracing spans are already quite helpful in debugging what failed in a migration, but more context is never a bad thing IMO.

conorsch (approved):
After following the test plan, post-migration, I no longer see a chain halt, nor do I see any error messages about invalid tags. Post-migration, I was able to submit and claim several swaps. LGTM.

hdevalence:
I also just tested this on the exact same state that caused the failure previously and confirmed that it worked fine. Great work!

crates/bin/pd/src/migrate/testnet74.rs
zbuc: Good catch. No, it is not idempotent. I wasn't aware that was a requirement for migrations; I will follow-up

erwanor: How is this not idempotent? if you run it multiple times you will get the same result, no? why not?

erwanor: is it because the encoding changes? that's completely fine imo

zbuc: > How is this not idempotent? if you run it multiple times you will get the same result, no? why not? 

Looking more closely, I am not sure at the moment. If we attempt to decode protobuf encoded Amount as a BE-encoded Amount it could fail silently and the Amount would change. However, this requires the protobuf byte encoding to be the same length as the big-endian encoding.

erwanor: Honestly not completely sure this is important but we can try to first decode it as a proto amount and return early if it works, and proceed if it correctly returns an error. This way it is resistant to accidental double runs

zbuc: There are some edge cases:

```rust
        let a: Amount = 4951760157141521099596496895u128.into();

        println!("amount 1: {:?}", a);

        let proto_a_bytes: Vec<u8> = penumbra_proto::core::num::v1::Amount::from(a).encode_to_vec();

        let amount = Amount::from_be_bytes(proto_a_bytes.as_slice().try_into()?);

        println!("amount 2: {:?}", amount);
```

```console
amount 1: 4951760157141521099596496895

amount 2: 11963051962064242856133983240072462207
```

so it could _theoretically_ still munge some values and be non-idempotent even if we took that approach if everything aligned and the stored `Amount`s were in a specific range and order.

But I'm also not convinced it's worth trying to fix... migration idempotency seems better to solve at a higher level than within individual functions modifying data if it's a design goal, as there are likely other similar cases that will come up in migration code.

zbuc: Ah there's some Discord discussion about this already: https://discord.com/channels/824484045370818580/1052302055680790568/1228444060113698938

Right now, migrations are not idempotent. This may require publishing the post-upgrade root hash, however that doesn't account for migrations affecting non-verifiable storage.

hdevalence: We should not tell people migrations are idempotent.

The right way to handle idempotency, later, is probably to use the ABCI application version and compare against a version number in the state.

PR #4346, opened 2024-05-07
docs: update guide for validator custody

hdevalence:
## Describe your changes

Tweak the description of validator key management to clarify the roles of the different keys

  > docs changes


conorsch (approved):
Looks great. I added one change: cross-linking the new pcli docs from the "validator requirements" page, so it's easy to discover based on context.

PR #4349, opened 2024-05-07
view: 👀 introduce more tracing instrumentation

cratelyn:
## describe your changes

this introduces some additional `tracing` instrumentation, used while debugging
integration with the view server in mock consensus tests. because these events
may be useful more broadly, and aren't strictly tied to testing, this is
cherry-picked out of that work for separate consideration.

## issue ticket number and link

* #3913

this branch only makes changes to tracing spans and events. as this is
only altering telemetry, this is not consensus-breaking.

### DIFF
--- a/crates/view/src/storage.rs
+++ b/crates/view/src/storage.rs
@@ -11,10 +11,12 @@ use r2d2_sqlite::{
     SqliteConnectionManager,
 };
 use sha2::{Digest, Sha256};
+use tap::{Tap, TapFallible};
 use tokio::{
     sync::broadcast::{self, error::RecvError},
     task::spawn_blocking,
 };
+use tracing::{error_span, Instrument};
 use url::Url;
 
 use penumbra_app::params::AppParameters;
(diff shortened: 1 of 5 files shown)


crates/view/src/storage.rs
```diff
@@ -75,20 +77,37 @@ pub struct Storage {
...
+                tracing::debug!(?path, "database exists");
                 return Self::load(path).await;
+            } else {
+                tracing::debug!(?path, "database does not exist");
             }
         };
 
-        let mut client = AppQueryServiceClient::connect(node.to_string()).await?;
+        let mut client = AppQueryServiceClient::connect(node.to_string())
+            .instrument(error_span!("connecting to endpoint"))
```
hdevalence, line 102: nit: could this be a snake_case name so that the span stack doesn't have spaces in it?

crates/view/src/storage.rs
```diff
@@ -75,20 +77,37 @@ pub struct Storage {
...
+            .await
+            .tap_err(|error| {
+                tracing::error!(?error, "failed to connect to app query service endpoint")
+            })?
+            .tap(|_| tracing::debug!("connected to app query service endpoint"));
         let params = client
             .app_parameters(tonic::Request::new(AppParametersRequest {}))
+            .instrument(error_span!("getting app parameters"))
```
hdevalence, line 110: ditto nit

hdevalence (approved):
LGTM!

crates/view/src/storage.rs
cratelyn: happily, also committing this take to memory for future span names. :+1:

cratelyn: addressed in 6189d376fd157e1fe06b106fcbd089318470aea9 :hammer:

crates/view/src/storage.rs
cratelyn: addressed in 6189d376fd157e1fe06b106fcbd089318470aea9 :hammer:

PR #4351, opened 2024-05-08
auction(rpc): support `grpc-web` in the auction query service

erwanor:
## Describe your changes

This adds missing middleware to the auction query service

  > query server change

### DIFF
--- a/crates/core/app/src/rpc.rs
+++ b/crates/core/app/src/rpc.rs
@@ -81,9 +81,9 @@ pub fn router(
         .add_service(we(StorageQueryServiceServer::new(StorageServer::new(
             storage.clone(),
         ))))
-        .add_service(AuctionQueryServiceServer::new(AuctionServer::new(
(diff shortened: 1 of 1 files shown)


hdevalence (approved):

cratelyn (approved):

crates/core/app/src/rpc.rs
```diff
@@ -81,9 +81,9 @@ pub fn router(
         .add_service(we(StorageQueryServiceServer::new(StorageServer::new(
             storage.clone(),
         ))))
-        .add_service(AuctionQueryServiceServer::new(AuctionServer::new(
+        .add_service(we(AuctionQueryServiceServer::new(AuctionServer::new(
```
cratelyn, line 84: oh no :weary:

jessepinho: My thoughts exactly

PR #4353, opened 2024-05-08
auction: define and record events

### LINKED ISSUE #4303: auction: define and emit events
We want to define three types of events
- [ ] `EventDutchAuctionScheduled`
- [ ] `EventDutchAuctionUpdated`
- [ ] `EventDutchAuctionEnded`

Each tracking sufficient data so that an explorer can cross-ref auction activity with position management happening on the dex.

erwanor:
## Describe your changes

This PR adds three messages to the `v1alpha1` auction interface:
- `EventDutchAuctionScheduled`
- `EventDutchAuctionUpdated`
- `EventDutchAuctionEnded`

Fixes #4303.

  > this only makes additive changes, recording events in the auction component.


cratelyn:
i'll mark this as "ready for review" since reviews were requested. feel free to move it back to a draft if there's still work to be done here!

proto/penumbra/penumbra/core/component/auction/v1alpha1/auction.proto
```diff
@@ -173,3 +173,30 @@ message ActionDutchAuctionWithdrawView {
...
+
+message EventDutchAuctionUpdated {
+  AuctionId auction_id = 1;
+  DutchAuctionState state = 2;
+}
+
+message EventDutchAuctionEnded {
+  // The reason the auction ended.
+  enum Reason {
+    REASON_UNSPECIFIED = 0;
```
cratelyn, line 190: i see we never create this variant in `crates/core/component/auction/src/event.rs`. should we remove it? we could always add new variants in the future if needed.

cratelyn (approved):
this looks great!

proto/penumbra/penumbra/core/component/auction/v1alpha1/auction.proto
cratelyn: answered this question myself, via `rg 'enum \w* {' -C 10 proto/penumbra`.

i see this is an existing pattern, so i'll mark this as resolved.

hdevalence: https://buf.build/docs/lint/rules#enum_value_prefix

All enums should have a zero value. proto3 doesn't differentiate between set and unset fields, so if an enum field isn't explicitly set, it defaults to the zero value. If an explicit zero value isn't part of the enum definition, this defaults to the actual zero value of the enum. For example, if you had the following .proto file, any Uri with scheme not explicitly set would default to SCHEME_FTP:

cratelyn:
@erwanor **NB** i edited the description to use `Fixes #4303` so that this will automatically close that issue as completed. this appears to address all of the events outlined in that issue.

proto/penumbra/penumbra/core/component/auction/v1alpha1/auction.proto
cratelyn: thanks for the background @hdevalence & @erwanor! i appreciate hearing the underlying reason for these `FOO_UNSPECIFIED` variants.

TalDerei:
LGTM!

PR #4361, opened 2024-05-09
Better handle regex in `pcli query watch`

### LINKED ISSUE #4360: Fix regex handling in `pcli query watch`
**Describe the bug**
The regex handling for `pcli query watch --nv-key-regex` doesn't seem to be working as expected:

```console
$ cargo run --release --bin pcli -- query watch --nv-key-regex '^dex/ab/.*'
Running `sign-and-run target/release/pcli query watch --nv-key-regex '^dex/ab/.*'`

192018 NVKV c3Rha2luZy92YWxpZGF0b3JzL2RhdGEvdXB0aW1lL3BlbnVtYnJhdmFsaWQxejZ3OGx3Nmg4dnZtaHBhcGNhNXB6dzU0aDVuZGFzdGg3bWF0OHRkYXJseXdnZDh2cDV4cTl3MGwwag== -> ...

$ echo -n "c3Rha2luZy92YWxpZGF0b3JzL2RhdGEvdXB0aW1lL3BlbnVtYnJhdmFsaWQxejZ3OGx3Nmg4dnZtaHBhcGNhNXB6dzU0aDVuZGFzdGg3bWF0OHRkYXJseXdnZDh2cDV4cTl3MGwwag==" | base64 -D
staking/validators/data/uptime/penumbravalid1z6w8lw6h8vvmhpapca5pzw54h5ndasth7mat8tdarlywgd8vp5xq9w0l0j
```

~I am also seeing unexpected behavior with the `--key-regex` parameter:~

```console
$ RUST_LOG=debug cargo run --release --bin pcli -- query watch --key-regex '.*'
2024-05-08T23:22:41.236673Z DEBUG pcli::command::query: req=WatchRequest { key_regex: ".*", nv_key_regex: "" }
Error: status: InvalidArgument, message: "invalid key_regex: Compiled regex exceeds size limit of 1024 bytes.", details: [], metadata: MetadataMap { headers: {"access-control-allow-credentials": "true", "access-control-allow-origin": "*", "access-control-expose-headers": "*", "content-type": "application/grpc", "date": "Wed, 08 May 2024 23:22:41 GMT", "vary": "origin", "vary": "access-control-request-method", "vary": "access-control-request-headers", "content-length": "0"} }
```

EDIT: this looks to be intended based on the docs for the `RegexBuilder::size_limit` method, but I am not sure if our current size limit (1024) is reasonable:

```rust
// It may surprise you how big some seemingly small patterns can
// be! Since \w is Unicode aware, this generates a regex that can
// match approximately 140,000 distinct codepoints.
```

In practice this seems ok:

`cargo run --release --bin pcli -- --home ~/.local/share/pcli-localhost query watch --key-regex 'dex/'` works fine.

zbuc:
## Describe your changes
This changes the `pcli query watch` command to fix a bug and improve usability.

Previously, the `nv_key_regex` was implemented incorrectly. This PR also makes the watch command disable watching the other storage type if only a single regex is supplied.

There is a slight drawback in the current implementation in that you cannot enable filtering the watch command for one of the regexes while supplying an empty regex for the other to view all matches, however changing this would require a change to the proto interface: `rpc Watch(WatchRequest) returns (stream WatchResponse);`

Closes #4360

  > Only affects RPC return

### DIFF
--- a/crates/cnidarium/src/rpc.rs
+++ b/crates/cnidarium/src/rpc.rs
@@ -28,6 +28,7 @@ use crate::rpc::proto::v1::{
     WatchResponse,
 };
 use futures::{StreamExt, TryStreamExt};
+use regex::Regex;
 use tokio_stream::wrappers::ReceiverStream;
 use tonic::Status;
 use tracing::instrument;
@@ -121,17 +122,29 @@ impl QueryService for Server {
 
         const MAX_REGEX_LEN: usize = 1024;
 
-        let key_regex = regex::RegexBuilder::new(&request.key_regex)
-            .size_limit(MAX_REGEX_LEN)
-            .build()
-            .map_err(|e| Status::invalid_argument(format!("invalid key_regex: {}", e)))?;
+        let key_regex = match request.key_regex.as_str() {
+            "" => None,
+            _ => Some(
+                regex::RegexBuilder::new(&request.key_regex)
+                    .size_limit(MAX_REGEX_LEN)
+                    .build()
+                    .map_err(|e| Status::invalid_argument(format!("invalid key_regex: {}", e)))?,
+            ),
+        };
 
         // Use the `bytes` regex to allow matching byte strings.
-        let nv_key_regex = regex::bytes::RegexBuilder::new(&request.key_regex)
-            .size_limit(MAX_REGEX_LEN)
-            .unicode(false)
-            .build()
-            .map_err(|e| Status::invalid_argument(format!("invalid nv_key_regex: {}", e)))?;
+        let nv_key_regex = match request.nv_key_regex.as_str() {
+            "" => None,
+            _ => Some(
+                regex::bytes::RegexBuilder::new(&request.nv_key_regex)
+                    .size_limit(MAX_REGEX_LEN)
+                    .unicode(false)
+                    .build()
+                    .map_err(|e| {
+                        Status::invalid_argument(format!("invalid nv_key_regex: {}", e))
+                    })?,
+            ),
+        };
 
         let (tx, rx) = tokio::sync::mpsc::channel::<Result<WatchResponse, tonic::Status>>(10);
 
@@ -148,8 +161,8 @@ impl QueryService for Server {
 
 async fn watch_changes(
     storage: Storage,
-    key_regex: regex::Regex,
-    nv_key_regex: regex::bytes::Regex,
+    key_regex: Option<regex::Regex>,
+    nv_key_regex: Option<regex::bytes::Regex>,
     tx: tokio::sync::mpsc::Sender<Result<WatchResponse, tonic::Status>>,
 ) -> anyhow::Result<()> {
     let mut changes_rx = storage.subscribe_changes();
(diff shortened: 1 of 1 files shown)


crates/cnidarium/src/rpc.rs
```diff
@@ -121,17 +122,29 @@ impl QueryService for Server {
...
-            .size_limit(MAX_REGEX_LEN)
-            .unicode(false)
-            .build()
-            .map_err(|e| Status::invalid_argument(format!("invalid nv_key_regex: {}", e)))?;
+        let nv_key_regex = match request.nv_key_regex.as_str() {
+            "" => None,
+            _ => Some(
+                regex::bytes::RegexBuilder::new(&request.nv_key_regex)
+                    .size_limit(MAX_REGEX_LEN)
+                    .unicode(false)
```
cratelyn, line 141: maybe we should add a comment here, remarking why the non-verifiable storage keys may not be valid unicode, compared to the verifiable storage keys?

crates/cnidarium/src/rpc.rs
```diff
@@ -160,31 +173,43 @@ async fn watch_changes(
...
-                        key: key.clone(),
-                        value: value.as_ref().cloned().unwrap_or_default(),
-                        deleted: value.is_none(),
-                    })),
-                }))
-                .await?;
+        if key_regex.is_some() || nv_key_regex.is_none() {
+            for (key, value) in changes.unwritten_changes().iter() {
+                if key_regex
+                    .as_ref()
```
cratelyn, line 179: i think the `Option<T>` changes here have ended up convoluting the logic a bit.

above, we match on..

```rust
        let nv_key_regex = match request.nv_key_regex.as_str() {
            "" => None,
            _ => Some(
                regex::bytes::RegexBuilder::new(&request.nv_key_regex)
                    // .. elided
                )
        };
```

..which we pass into this function and subsequently re-map `Option::None` back into an empty regex. it seems like we can both keep this diff smaller, and keep the underlying code the same, if we walk back the change to make these regex paremeters optional values.

what do you think @zbuc?

cratelyn (commented):
i have a question about the changes to the `watch_changes` signature below!

crates/cnidarium/src/rpc.rs
zbuc: I considered this however my understanding is that we cannot use optional parameters in protobuf as the Golang implementation doesn't handle them correctly. We previously used optional parameters all over, and had to excise them. We could get around this in a couple ways, either by using an enum, or a placeholder string value.

cratelyn: hm, i worry i might not have communicated this clearly. let me open a branch against this...

cratelyn: see #4364, this is what i was trying to get at. does that seem right to you? apologies if i am missing something here.

cratelyn: marking as resolved :white_check_mark:

crates/cnidarium/src/rpc.rs
```diff
@@ -121,17 +122,29 @@ impl QueryService for Server {
 
         const MAX_REGEX_LEN: usize = 1024;
```
cratelyn, line 123: > I am not sure if our current size limit (1024) is reasonable:

the description in #4360 questioned this limit, should we still bump this to something like..

```suggestion
        const MAX_REGEX_LEN: usize = 4096;
```

cratelyn (approved):
approving, modulo a question about the regex size limit mentioned in the original issue. nice! :sparkles:

crates/cnidarium/src/rpc.rs
cratelyn: (_i see this was already there, so this is not blocking_)

crates/cnidarium/src/rpc.rs
zbuc: In practice I think it might be okay as-is. The common regexes you'll want to use are simple and short prefix-style: `^dex/ra/` `^staking/` etc. and I only encountered this because I was "holding it wrong" and supplied a wildcard character match unnecessarily.

If we bump against it in the future we can increase it.

cratelyn: sounds good! marking as resolved now that we have closed this loop ➰

Issue #4356, opened 2024-05-08
auction: implement a value breaker


hdevalence:
Value moves into the auction component when auctions are created or when positions are closed, and out when it creates positions or when auctions are claimed.

The auction component should have the same kind of VCB as the DEX (can copy the code over). This should be pretty simple to implement, but it will be quite complicated to add a migration for it later.

PR #4363, opened 2024-05-09
auction(vcb): add a value circuit breaker

erwanor:
## Describe your changes

This PR adds a value circuit breaker, modeled after the dex component's.

#4356 

  > This is very consensus breaking, but not state breaking because this state didn't exist and the auction is not previously deployed so there are no inflows to track.


hdevalence (approved):
This looks correct to me.

TalDerei (approved):
As a point of process, Erwan conducted an impromptu pairing session where we discussed the high level goal of this PR: **1.** from first principles, make sure we identify all the proper junction points in the flow tracking, and **2.** make sure we properly track the inflows and outflows of value in these junction points. We then discussed and sanity checked the value transfer to / from the auction / dex components.

erwanor:
We missed a juncture point which is when a user submitted action cause an auction to end early, in that case the component's value balance needs to be *credited*. x-ref #4388

PR #4364, opened 2024-05-09
cnidarium: 🪼 `watch_changes` regex parameters aren't options (review)

cratelyn:
see discussion from #4361, here:
https://github.com/penumbra-zone/penumbra/pull/4361#discussion_r1595614864

**NB: this is a PR based on `4360_regex_watch`, not against `main`.** because this is largely a subtractive diff when compared against that base branch, see also this diff when compared against main, here: https://github.com/penumbra-zone/penumbra/compare/main...kate/4360_regex_watch.review-24-05-09

```diff
diff --git a/crates/cnidarium/src/rpc.rs b/crates/cnidarium/src/rpc.rs
index 72bc796be..a8accb144 100644
--- a/crates/cnidarium/src/rpc.rs
+++ b/crates/cnidarium/src/rpc.rs
@@ -28,6 +28,7 @@ use crate::rpc::proto::v1::{
     WatchResponse,
 };
 use futures::{StreamExt, TryStreamExt};
+use regex::Regex;
 use tokio_stream::wrappers::ReceiverStream;
 use tonic::Status;
 use tracing::instrument;
@@ -127,7 +128,7 @@ impl QueryService for Server {
             .map_err(|e| Status::invalid_argument(format!("invalid key_regex: {}", e)))?;
 
         // Use the `bytes` regex to allow matching byte strings.
-        let nv_key_regex = regex::bytes::RegexBuilder::new(&request.key_regex)
+        let nv_key_regex = regex::bytes::RegexBuilder::new(&request.nv_key_regex)
             .size_limit(MAX_REGEX_LEN)
             .unicode(false)
             .build()
```

### DIFF
--- a/crates/cnidarium/src/rpc.rs
+++ b/crates/cnidarium/src/rpc.rs
@@ -122,29 +122,17 @@ impl QueryService for Server {
 
         const MAX_REGEX_LEN: usize = 1024;
 
-        let key_regex = match request.key_regex.as_str() {
-            "" => None,
-            _ => Some(
-                regex::RegexBuilder::new(&request.key_regex)
-                    .size_limit(MAX_REGEX_LEN)
-                    .build()
-                    .map_err(|e| Status::invalid_argument(format!("invalid key_regex: {}", e)))?,
-            ),
-        };
+        let key_regex = regex::RegexBuilder::new(&request.key_regex)
+            .size_limit(MAX_REGEX_LEN)
+            .build()
+            .map_err(|e| Status::invalid_argument(format!("invalid key_regex: {}", e)))?;
 
         // Use the `bytes` regex to allow matching byte strings.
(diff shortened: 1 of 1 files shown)


zbuc:
This seems to be the original behavior, where if you supply a regex for one of the storage types, you get all events for the other storage type unfiltered. This seemed undesirable to me when I was using the interface, typically if you're supplying a regex you want something to see something very specific so I made it exclude the other storage type. But I'm open to changing my mind on this.

cratelyn:
> This seems to be the original behavior, where if you supply a regex for one of the storage types, you get all events for the other storage type unfiltered. This seemed undesirable to me when I was using the interface, typically if you're supplying a regex you want to see something very specific so I made it exclude the other storage type. But I'm open to changing my mind on this.

that makes sense, i'll go ahead and close this out. thanks for elaborating, i'm glad a diff helped suss this out!

PR #4368, opened 2024-05-09
penumbra: use `core.auction.v1`

erwanor:
## Describe your changes

This tags a `v1` protobuf package for the auction component.

It includes a breaking changes to `penumbra.v1` since we use the new `v1` package to directly replace the previous references to the auction component in the app, view, transaction, and governance components. This is an exceptional deviation from process because this project was in active development with a single consumer.

  > This change is both proto-breaking, consensus-breaking, and technically state-breaking. It doesn't require a migration because we never deployed these changes to our testnet.


conorsch (approved):
LGTM. The buf-lint job is failing, which is expected: we're moving proto versions. But the change is narrowly scoped to affect only the auction interfaces, which is what we want. Other CI jobs must pass, though.

conorsch:
> Other CI jobs must pass, though.

Confirmed.

Issue #4325, opened 2024-05-06
Release Testnet 74, deimos-7, via chain upgrade


conorsch:
# Testnet Release

Testnet name: Testnet 74, `penumbra-testnet-deimos-7`, with chain upgrade
Release date: 2024-05-09
Testnet release manager: @conorsch

# Testnet Release Manager Checklist

We intend to ship a chain upgrade to the current testnet, including all the changes queued up in main. We must ensure that all necessary migrations have been written and tested.

Pre-release:

- [x] Review all merged PRs with tags [`consensus-breaking` and/or `state-breaking`](https://github.com/penumbra-zone/penumbra/pulls?q=is%3Apr+label%3Astate-breaking%2Cconsensus-breaking+created%3A%3E%3D2024-03-25)
- [x] Confirm migrations exist where required for breaking changes already merged
- [x] Identify specific test cases for before/after behavior, querying chain state, to ensure that migrations are effective
- [ ] (Stretch goal): Land automated upgrade testing as described in #4323 
- [x] Land 11th hour PRs:
  - [x] https://github.com/penumbra-zone/penumbra/pull/4368
  - [x] https://github.com/penumbra-zone/penumbra/pull/4367
  - [x] https://github.com/penumbra-zone/penumbra/pull/4363
  - [x] https://github.com/penumbra-zone/penumbra/pull/4291

On release day:

- [x] Draft an announcement for peer review to ensure major changes included are comprehensive.
- [x] Disable testnet deploy workflow, so that chain is not reset
- [x] Bump the version number and push its tag, via [cargo-release](https://crates.io/crates/cargo-release).
    - [x] Run `cargo release minor` for a new testnet, or `cargo release patch` for a bugfix. For the latter, make sure you're on a dedicated release branch.
    - [x] Push the commit and newly generated tag, e.g. `v0.51.0`, to the remote.
- [x] Manually trigger container-build workflow, bc deploy workflow is disabled
- [ ] Wait for the ["Release" workflow](https://github.com/penumbra-zone/penumbra/actions/workflows/release.yml) to complete: it'll take ~60m, most of which is the Windows builds.
- [ ] Edit the newly created (and published) release object, then click "Generate release notes." Cut and paste the generated text from the bottom to the top of the post, then save it.
- [x] Close faucet (chain halt will make it inoperative anyway)
- [ ] Run migrations on all validators
- [ ] Run migrations on all fullnodes
[...]

PR #4372, opened 2024-05-10
docs: refine migrate steps

conorsch:
## Describe your changes
Fixed indented list. Explicitly mention backup. Remove recommendation, even optionally, to use `pd export` on post-migration state, because `pd export` won't generate an archive that's suitable for use with `pd testnet join`. We'll do that manually as part of post-release follow-up.

  > docs-only!


conorsch:
Docs-only, and CI is passing, so self-merging for inclusion with release comms. Refs #4325.

Issue #4374, opened 2024-05-10
Release Testnet 75, deimos-8, via chain reset


conorsch:
# Testnet Release

Testnet name: Testnet 75, `penumbra-testnet-deimos-8`, via chain reset
Release date: 2024-05-10
Testnet release manager: @conorsch

# Testnet Release Manager Checklist

We intend to ship a reset, to recover from the mismanaged upgrade to 74. See #4325 and  #4373 for details.

Pre-release:

- [x] Fetch funding streams from previously active validators
- [x] Prepare release PR that makes the following changes:
  * bump chain id from `penumbra-testnet-deimos-7` -> `penumbra-testnet-deimos-8` 
  * reset `TOTAL_HALT_COUNT` to `0` - https://github.com/penumbra-zone/penumbra/pull/4376
  * increment `APP_VERSION` from `3` to `4` - https://github.com/penumbra-zone/penumbra/pull/4375
  * bump crate versions `v0.74.0` -> `0.75.0`

On release day:

- [x] Draft an announcement for peer review to ensure major changes included are comprehensive.
- [x] Enable testnet deploy workflow, so that chain is reset
- [x] Bump the version number and push its tag, via [cargo-release](https://crates.io/crates/cargo-release).
    - [x] Run `cargo release minor` for a new testnet, or `cargo release patch` for a bugfix. For the latter, make sure you're on a dedicated release branch.
    - [x] Push the commit and newly generated tag, e.g. `v0.51.0`, to the remote.
- [x] Wait for the ["Release" workflow](https://github.com/penumbra-zone/penumbra/actions/workflows/release.yml) to complete: it'll take ~60m, most of which is the Windows builds.
- [x] You must [manually review](https://docs.github.com/en/actions/managing-workflow-runs/reviewing-deployments) the waiting deployment
- [x] Delegate to the Penumbra Labs CI validators; use amounts of ~200k `penumbra` per validator.
- [x] Open a position to provide some initial liquidity: `pcli tx lp replicate xyk penumbra:test_usd 20000penumbra --current-price 20`
- [x] Edit the newly created (and published) release object, then click "Generate release notes." Cut and paste the generated text from the bottom to the top of the post, then save it.
- [x] Close faucet (chain halt will make it inoperative anyway)
- [x] Update Galileo deployment, [following docs](https://github.com/penumbra-zone/galileo).
- [x] Make the announcement to Discord! 🎉🎉🎉

Post-release cleanup tasks

[...]

PR #4375, opened 2024-05-10
chore: prepare testnet 75

conorsch:
## Describe your changes

Makes three substantive changes:

  * resets `TOTAL_HALT_COUNT` to 0
  * increments `APP_VERSION`
  * increments chain_id `penumbra-testnet-deimos-7` -> `penumbra-testnet-deimos-8`

There are multiple chain id directories included, pulled from the `release/v0.73.2` tag. Due to our multi-branch workflow, we never backported those testnet configs to the main branch, but doing so is necessary now that we're releasing from main again.

Refs #4374.

  > This PR is in prep for a full chain reset.


COMPATIBILITY.md
```diff
@@ -3,3 +3,4 @@
 | 1 (Testnet 70)            | v0.70.x                | v0.37.5  |   v1     |
 | 2 (Testnet 73)            | v0.73.x                | v0.37.5  |   v1     |
 | 3 (Testnet 74)            | v0.74.x                | v0.37.5  |   v1     |
+| 4 (Testnet 75)            | v0.75.x                | v0.37.5  |   v1     |
```
conorsch, line 6: Bumping crate version only in the COMPAT doc for now. Post-merge, we'll bump-and-tag directly on the main branch, to trigger the release.

cratelyn (approved):
> There are multiple chain id directories included, pulled from the release/v0.73.2 tag. Due to our multi-branch workflow, we never backported those testnet configs to the main branch, but doing so is necessary now that we're releasing from main again.

thanks for explaining the backing context. this looks great.

cratelyn:
one non-blocking question, since github review does not provide a nice way to hang comments on a parent directory:

for extra clarity, i might suggest that we leave e.g. an empty tombstone / stub directory, `testnets/074-deimos-7`. this could contain a small `README.md` containing a small memorandum about why that migration failed, noting that the jump from `73` to `75` was intentional and related to an operational issue?

conorsch:
> for extra clarity, i might suggest that we leave e.g. an empty tombstone / stub directory, testnets/074-deimos-7. this could contain a small README.md containing a small memorandum about why that migration failed, noting that the jump from 73 to 75 was intentional and related to an operational issue.

This is a good suggestion! There are similar events would could backfill for posterity. The testnet dirs themselves are machine-readable, and their contents are baked into binaries, so I'd prefer to keep READMEs out of band. But a file in `testnets/` recording the history is a gift to future maintainers, for sure.

cratelyn:
> > for extra clarity, i might suggest that we leave e.g. an empty tombstone / stub directory, testnets/074-deimos-7. this could contain a small README.md containing a small memorandum about why that migration failed, noting that the jump from 73 to 75 was intentional and related to an operational issue.
> 
> This is a good suggestion! There are similar events would could backfill for posterity. The testnet dirs themselves are machine-readable, and their contents are baked into binaries, so I'd prefer to keep READMEs out of band. But a file in `testnets/` recording the history is a gift to future maintainers, for sure.

ah-ha! that is very good to know, had not realized these directories are machine-readable. knowing that, i agree with that plan of action. thank you for elaborating!

PR #4380, opened 2024-05-10
Support nonverifiable storage in the `pcli query key` interface

### LINKED ISSUE #2647: Avoid generating a merkle proof if a `KeyValue` request doesn't ask for it
Once #2628 lands, we will be able to retrieve JMT values directly from storage, circumventing the need to traverse the tree entirely. However, whenever a `KeyValue` request is received, we default to generating a (non-)inclusion proof and dropping it after the fact if it wasn't needed.

To fully take advantage of segmenting values from the tree structure, we should make sure that we do not generate a (non-)inclusion proof when a `KeyValueRequest` specify that it doesn't need it. The solution should take advantage of `TreeReader::get_value_option`.

### LINKED ISSUE #4359: Support non-verifiable storage keys in the `pcli query key` interface
While writing testing instructions for the migration described in https://github.com/penumbra-zone/penumbra/pull/4341 I noticed that the current cnidarium query interfaces `pcli query key` does not support querying from non-verifiable storage. The streaming interface (`pcli query watch`) does support it with the `--nv-key-regex` parameter, however I was unable to get it to work; the command `cargo run --release --bin pcli -- query watch --nv-key-regex '^dex/ab/.*` showed many other keys passing by but maybe I formatted the regex wrong.

Supporting this behavior is extremely useful for debugging running chain state.

Follow-up work on improving the usability of this interface would be nice (currently it prints base64-encoded data only, it would be nice to have a pretty-print option), though that is lower priority.

The `pcli query key` command should probably:

* by default, query both non-verifiable storage and the JMT keys and print results from both for all queries
* optionally support `--jmt` and `--non-verifiable` boolean flags.

zbuc:
## Describe your changes

This PR adds support for querying nonverifiable storage in the `pcli query key` interface.

It also changes the API to only generate proofs if requested.

Closes #4359 and closes #2647

  > Only changes the key value query API


proto/penumbra/penumbra/cnidarium/v1/cnidarium.proto
```diff
@@ -21,10 +21,19 @@ service QueryService {
...
+    STORAGE_BACKEND_JMT = 1;
+    STORAGE_BACKEND_NONVERIFIABLE = 2;
+  }
+
   // If set, the key to fetch from storage.
   string key = 2;
   // whether to return a proof
   bool proof = 3;
+  // Which storage to query.
+  StorageBackend storage_backend = 4;
```
erwanor, line 36: The idea of letting the user specify which storage layer they want to target is interesting, but I think the api would be simpler by just adding another message:

```protobuf
message NonVerifiableKeyValueRequest {
    message Key {
        bytes inner = 1;
    }

    Key key = 1;
}
```

This has a couple advantages:
1. We don't have to make callers `base64` key depending on which backend they target
2. We don't need an idle `proof` field that is only relevant for one backend
3. It makes our protobuf state query interface coherent with how we structured `StateRead`

crates/wasm/publish/pnpm-lock.yaml
```diff
@@ -0,0 +1,477 @@
+lockfileVersion: '6.0'
+
```
erwanor, line 2: I think this was staged by accident?

erwanor (approved):
LGTM, but we have to remove the `pnpm-lock` file that sneaked in

crates/wasm/publish/pnpm-lock.yaml
zbuc: Yeah, I have no idea why that was generated

PR #4358, opened 2024-05-08
ci: 🚭 fix summonerd smoke tests

cratelyn:
#### describe your changes

this provides some small changes needed for running the summoner smoke test locally.

#### issue ticket number and link

fixes #4321.

fixes #4370.

#### checklist before requesting a review

- [x] if this code contains consensus-breaking changes, i have added the "consensus-breaking" label. otherwise, i declare my belief that there are not consensus-breaking changes, for the following reason:

  > this branch only changes testing infrastructure and development tooling.

PR #4382, opened 2024-05-12
ci: 🩹 add missing `"` to pid expansion

cratelyn:
addendum to #4358.

  > this only makes a change to ci scripts.


cratelyn:
noticed this in https://github.com/penumbra-zone/penumbra/actions/runs/9036580908/job/24833738240

```
./deployments/scripts/smoke-summoner.sh: exit trap: line 1: unexpected EOF while looking for matching `"'
```

conorsch (approved):
Shellcheck's happy, I'm happy.

PR #4385, opened 2024-05-14
pcli: auction end-all, withdraw-all

hdevalence:
## Describe your changes

Changes the `pcli tx auction` commands to support "end all", "withdraw all" rather than requiring specific auction IDs

  > Only client-side changes

### DIFF
--- a/crates/bin/pcli/src/command/tx/auction/dutch.rs
+++ b/crates/bin/pcli/src/command/tx/auction/dutch.rs
@@ -2,18 +2,16 @@ use std::path::Path;
 
 use crate::command::tx::FeeTier;
 use crate::App;
+use anyhow::Result;
 use anyhow::{anyhow, bail, Context};
 use clap::Subcommand;
 use comfy_table::presets;
 use dialoguer::Confirm;
 use penumbra_asset::{asset::Cache, Value};
-use penumbra_auction::auction::dutch::actions::ActionDutchAuctionWithdrawPlan;
 use penumbra_auction::auction::{dutch::DutchAuction, dutch::DutchAuctionDescription, AuctionId};
-use penumbra_dex::lp::position::Position;
 use penumbra_keys::keys::AddressIndex;
 use penumbra_num::Amount;
-use penumbra_proto::{view::v1::GasPricesRequest, DomainType, Name};
-use penumbra_view::SpendableNoteRecord;
+use penumbra_proto::{view::v1::GasPricesRequest, DomainType};
 use penumbra_view::ViewClient;
 use penumbra_wallet::plan::Planner;
 use rand::RngCore;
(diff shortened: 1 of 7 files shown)


hdevalence:
This doesn't work because there's not a way to track the "local" state of the auction NFT, which may lag the on-chain state. So once a GDA is completed, close-all doesn't work (because all the auctions are believed to be closed), but withdraw-all doesn't work (because the auction NFTs are not closed)

hdevalence:
Suggestion: add a `local_seq` field to https://buf.build/penumbra-zone/penumbra/docs/main:penumbra.view.v1#penumbra.view.v1.AuctionsResponse that gives the local sequence number

erwanor (approved):
I have tested it with my local wallet, but it would be cool to have someone else take it for a spin.

crates/view/src/planner.rs
```diff
@@ -222,9 +222,33 @@ impl<R: RngCore + CryptoRng> Planner<R> {
...
+            asset_id: auction.description.input.asset_id,
+        };
+        let reserves_output = Value {
+            amount: auction.state.output_reserves,
+            asset_id: auction.description.output_id,
+        };
+
+        let plan = ActionDutchAuctionWithdrawPlan {
+            auction_id,
+            seq: 2,
```
cratelyn, line 247: is this intentionally hardcoded to `2`? it would be nice to leave a comment explaining why this is `2`, if so

crates/bin/pcli/src/command/tx/auction/dutch.rs
```diff
@@ -366,6 +370,85 @@ impl DutchCmd {
...
+            }
+        })
+        .collect();
+    Ok(auctions)
+}
+/// Return all the auctions that need to be ended, based on our local view of the chain state.
+async fn auctions_to_end(view_client: &mut impl ViewClient, source: u32) -> Result<Vec<AuctionId>> {
+    let auctions = active_dutch_auction_states(view_client, source).await?;
+
+    let auction_ids = auctions
```
cratelyn, line 414: non-blocking nit: this doesn't need a `let` binding, since we immediately return it.

ditto for `auctions_to_withdraw`, below

cratelyn (commented):
i have one question about a hard-coded sequence number, but this looks good!

crates/view/src/planner.rs
erwanor:
```suggestion
            seq: 2, // 1 (closed) -> 2 (withdrawn)
```

good point, here the sequence number corresponds to a withdrawn auction. It's documented in [`DutchAuction`] but added another comment to re-anchor it

crates/bin/pcli/src/command/tx/auction/dutch.rs
erwanor: that's subjective, but i prefer an extra binding to make it easy to skim the code and know what it returns

cratelyn (approved):
very nice!

PR #4386, opened 2024-05-14
ibc: audit fixes

### LINKED ISSUE #4418: ibc: False negative for Timeouts on ordered channel
**Describe the bug**

When handling MsgTimeout on ordered channels, penumbra-ibc uses != on sequence numbers, so it will only accept proof that the most recent packet timed out.

**Expected behavior**
Any packet should be able to be proven that it has timed out.

### LINKED ISSUE #4419: ibc: Missing expiration check in UpgradeClient
**Describe the bug**
The handling of UpgradeClient by ibc-go checks both that the client is not frozen and not expired.
The handling of UpgradeClient by penumbra-ibc only checks that the client is not frozen.

Since penumbra-ibc and ibc-go disagree about which UpgradeClient messages are valid, an at-
tacking chain can make chains with different implementations disagree on its state by sending an
upgrade request while expired (e.g., by delaying sending UpdateClient messages and using a low
trusting period)

**Expected behavior**
`penumbra-ibc` should check that the trusted client state is nonexpired in ClientUpgradeProofVeri-
fier::verify_client_upgrade_proof

### LINKED ISSUE #4420: ibc: Upgrade-path length hardcoded to two
**Describe the bug**
Whereas ibc-go uses all but the last element of UpgradePath for verifying the client and consensus
state proofs when processing an UpgradeClient message, penumbra-ibc only uses the first ele-
ment of upgrade_path, which is equivalent to an assumption that the path length is always two.

If the upgrade path is ever longer than two, a malicious chain may be able to include different client
and consensus state proofs at different paths, upgrading its state to different values on penumbra-
ibc and ibc-go chains.

**Expected behavior**

`penumbra-ibc` should use all but the last element of upgrade_path instead of upgrade_path[0] in ClientUpgrade-
ProofVerifier::verify_client_upgrade_proof.

### LINKED ISSUE #4421: ibc: Handshake state machines too permissive when processing Acks
**Describe the bug**
Both ConnectionOpenAck and ChannelOpenAck permit the party processing the message to have
their channel in State::TryOpen in addition to State::Init.

It may be possible for a malicious chain to swap roles partway through the handshake (in the fol-
lowing, A is an attacker-controlled chain and B is a penumbra-ibc chain initiating a connection with
it):
• B sends ConnectionOpenInit to A; state is (Init, None).
• A sends ConnectionOpenTry to B; state is (Init, TryOpen).
• B sends ConnectionOpenAck to A; state would normally be in (Open, TryOpen), but A can
deviate here and store TryOpen, resulting in state (TryOpen, TryOpen).
• A deviates from the protocol and sends ConnectionOpenAck to B (instead of sending Con-
nectionOpenConfirm). If B is an ibc-go node, B rejects this since it is in TRYOPEN state and
not INIT state, but on penumbra-ibc, this check succeeds. At this point, to pass the con-
nection -erkle proof check, A needs to prove that their channel is in TryOpen state, which
they can do if they deviated above. At this point, the client and consensus-state proofs
that B is processing refer to A's representation of B's state. If this step succeeds, (A, B)
are in state (TryOpen, Open).
• B sends ConnectionOpenConfirm to A (at this point the roles have reversed), and the re-
sulting state is (Open, Open).
This may result in not all relevant proofs being checked in both directions or IDs stored in B being
chosen by A.

**Expected behavior**

`penumbra-ibc` should enforce that connection.state/channel.state are only in State::Init in MsgConnectionOpe-
nAck::try_execute/ChannelOpenAck::try_execute, respectively.

avahowell:
## Describe your changes

This PR contains a set of fixes for the higher sev issues from the recent audit of the penumbra-ibc codebase, that are specific to the state machine side of things (as opposed to `ibc-types`, there will be a follow up PR for ibc-types and a new release for the relevant fixes there).

closes #4421 
closes #4420 
closes #4419 
closes #4418

### DIFF
--- a/crates/core/component/ibc/src/component/msg_handler/connection_open_ack.rs
+++ b/crates/core/component/ibc/src/component/msg_handler/connection_open_ack.rs
@@ -230,11 +230,10 @@ async fn verify_previous_connection<S: StateRead>(
 
     // see
     // https://github.com/cosmos/ibc/blob/master/spec/core/ics-003-connection-semantics/README.md
+    //
     // for this validation logic
-    let state_is_consistent = connection.state_matches(&State::Init)
-        && connection.versions.contains(&msg.version)
-        || connection.state_matches(&State::TryOpen)
-            && connection.versions.get(0).eq(&Some(&msg.version));
+    let state_is_consistent =
+        connection.state_matches(&State::Init) && connection.versions.contains(&msg.version);
 
     if !state_is_consistent {
         anyhow::bail!("connection is not in the correct state");
--- a/crates/core/component/ibc/src/component/msg_handler/timeout.rs
+++ b/crates/core/component/ibc/src/component/msg_handler/timeout.rs
@@ -83,7 +83,7 @@ impl MsgHandler for MsgTimeout {
 
         if channel.ordering == ChannelOrder::Ordered {
             // ordered channel: check that packet has not been received
-            if self.next_seq_recv_on_b != self.packet.sequence {
+            if self.next_seq_recv_on_b > self.packet.sequence {
                 anyhow::bail!("packet sequence number does not match");
             }
 
--- a/crates/core/component/ibc/src/component/msg_handler/channel_open_ack.rs
+++ b/crates/core/component/ibc/src/component/msg_handler/channel_open_ack.rs
@@ -110,7 +110,7 @@ impl MsgHandler for MsgChannelOpenAck {
 }
 
 fn channel_state_is_correct(channel: &ChannelEnd) -> anyhow::Result<()> {
-    if channel.state == ChannelState::Init || channel.state == ChannelState::TryOpen {
+    if channel.state == ChannelState::Init {
         Ok(())
     } else {
         Err(anyhow::anyhow!("channel is not in the correct state"))
--- a/crates/core/component/ibc/src/component/msg_handler/upgrade_client.rs
+++ b/crates/core/component/ibc/src/component/msg_handler/upgrade_client.rs
@@ -58,7 +58,7 @@ impl MsgHandler for MsgUpgradeClient {
             .context("couldn't decode proof for upgraded client state")?;
 
         state
-            .verify_client_upgrade_proof(
+            .verify_client_upgrade_proof::<HI>(
                 &self.client_id,
                 &proof_client_state,
                 &proof_consensus_state,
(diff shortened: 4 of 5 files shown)


crates/core/component/ibc/src/component/msg_handler/connection_open_ack.rs
```diff
@@ -230,11 +230,10 @@ async fn verify_previous_connection<S: StateRead>(
...
     // see
     // https://github.com/cosmos/ibc/blob/master/spec/core/ics-003-connection-semantics/README.md
+    //
     // for this validation logic
-    let state_is_consistent = connection.state_matches(&State::Init)
-        && connection.versions.contains(&msg.version)
-        || connection.state_matches(&State::TryOpen)
-            && connection.versions.get(0).eq(&Some(&msg.version));
+    let state_is_consistent =
+        connection.state_matches(&State::Init) && connection.versions.contains(&msg.version);
```
cronokirby, line 236: https://github.com/cosmos/ibc/blob/main/spec/core/ics-003-connection-semantics/README.md#opening-handshake

So is the spec here wrong in allowing (Init, TryOpen)?

crates/core/component/ibc/src/component/msg_handler/timeout.rs
```diff
@@ -83,7 +83,7 @@ impl MsgHandler for MsgTimeout {
 
         if channel.ordering == ChannelOrder::Ordered {
             // ordered channel: check that packet has not been received
-            if self.next_seq_recv_on_b != self.packet.sequence {
+            if self.next_seq_recv_on_b > self.packet.sequence {
```
cronokirby, line 86: If you get a timeout with a packet sequence in the future you should still act on it, right? (which this code does)

cronokirby (approved):
Looks good, just have a few questions which could be good to answer here for posterity

conorsch:
We believe that this PR may resolve a lingering blocker for interchaintest, where timeouts were not being handled properly. Let's include this for Testnet 76 (#4402), so downstream can consume it, and press on with the testing integration. Before I mash the merge button, @avahowell could you respond to @cronokirby's comments above? It'd be helpful to have the narrative when we refer back to this changeset.

crates/core/component/ibc/src/component/msg_handler/connection_open_ack.rs
avahowell: yes, i believe so

crates/core/component/ibc/src/component/msg_handler/timeout.rs
avahowell: My understanding is that we should be acting on any packet that occured in the past and that has not been received yet, which is what ibc-go does and what this change is attempting to do.

PR #4387, opened 2024-05-15
ci: add `penumbra-mock-*` to rustdoc build list

erwanor:
## Describe your changes

This will let us browse them on `rustdoc.penumbra.zone`

  > ci change to doc build

### DIFF
--- a/deployments/scripts/rust-docs
+++ b/deployments/scripts/rust-docs
@@ -57,6 +57,8 @@ cargo +nightly doc --no-deps \
   -p penumbra-wallet \
   -p poseidon-paramgen \
   -p poseidon-permutation \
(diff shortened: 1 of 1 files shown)


cratelyn (approved):

deployments/scripts/rust-docs
```diff
@@ -57,6 +57,8 @@ cargo +nightly doc --no-deps \
   -p penumbra-wallet \
   -p poseidon-paramgen \
   -p poseidon-permutation \
+  -p penumbra-mock-consensus \
```
cratelyn, line 60: mock consensus is already in this list, above ☝️

cratelyn:
approving, but one of these is already in the rust docs!

Issue #4388, opened 2024-05-15
auction(dutch): credit VCB when ending an auction


erwanor:
**Issue**
Auctions can be ended by the protocol *and* users. The component VCB is only credited when a closure is issued by the protocol. Instead, it should credit the component balance in both cases.

**Context**
The auction component maintains a value circuit breaker (VCB) accounting for a nonnegative `Balance` of assets deployed in the component. 

This mechanism serves two purposes: fault detection and isolation. The VCB detect value balance irregularities by systematically crediting/debiting the component as assets flows in and out of it. If its invariant is breached (negative balance) the component halts. Halting on a serious invariant breach insulates the rest of the system from excess value leaking into it. x-ref #4363

**Remediation**

To remediate this, we need a combination of:
- a hotfix that patch the incorrect logic
- a fix that enshrine the failure conditions into unit tests
- a migration that heals the chain state

**Hotfix**
The fix for this bug is simple: we need to credit the component with the value it pulls from the DEX engine.

**Fix**
We have a unit test that reproduces the exact failure conditions for this bug (omitted here) and memorialize context around it, similar to 53 halts.

**Migration**
The migration for this bug is difficult: we need to reconstruct an accurate tally of the component's balance.

PR #4394, opened 2024-05-15
auction: credit VCB when ending an auction

erwanor:
## Describe your changes

- [x] hotfix for #4388
- [x] reproduction for #4388 
- [x] migration for #4388 

  > This is consensus-breaking and state-breaking. Migration WIP.

### DIFF
--- a/crates/test/tracing-subscriber/src/lib.rs
+++ b/crates/test/tracing-subscriber/src/lib.rs
@@ -15,14 +15,21 @@ pub fn set_tracing_subscriber() -> DefaultGuard {
     let filter = "info,penumbra_app=trace,penumbra_mock_consensus=trace";
     let filter = EnvFilter::try_from_default_env()
         .or_else(|_| EnvFilter::try_new(filter))
-        .expect("should have a valid filter directive")
-        // Without explicitly disabling the `r1cs` target, the ZK proof implementations
-        // will spend an enormous amount of CPU and memory building useless tracing output.
-        .add_directive(
-            "r1cs=off"
-                .parse()
-                .expect("rics=off is a valid filter directive"),
-        );
+        .expect("should have a valid filter directive");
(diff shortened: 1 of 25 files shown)


crates/test/tracing-subscriber/src/lib.rs
```diff
@@ -15,14 +15,21 @@ pub fn set_tracing_subscriber() -> DefaultGuard {
...
-                .expect("rics=off is a valid filter directive"),
-        );
+        .expect("should have a valid filter directive");
+    set_tracing_subscriber_with_env_filter(filter)
+}
+
+/// Install a tracing subscriber with a custom filter.
+//  NB: this is marked as "dead code" but it is used by integration tests.
+#[allow(dead_code)]
+pub fn set_tracing_subscriber_with_env_filter(filter: EnvFilter) -> DefaultGuard {
```
erwanor, line 25: This adds a method to specify a custom `EnvFilter`. I'm not totally sure this is the right api, or desired. I added this so that I could scope the tracing logs to the exact subset that's relevant to observe the bug that this test repros, cc @cratelyn

cratelyn: yes yes yes!! very much in favor of doing this, i'd pondered doing before as well. :pray:

erwanor:
I have tested this in two ways:
- a local devnet running `v0.75.0` with a collection of auctions running every block
- an application test that reproduce the bug (and attests that the fix works). 

Additionally, I ran the migration on a fork of the running testnet to observe if any deviations were recorded, but without resuming. I didn't see any deviations, that is to say the tallied auction VCB and the one onchain were equal.

TalDerei (approved):

erwanor:
This is generating conflicts to every other U1 migrations and has received a lot of offline attention so merging this optimistically, but async reviews are welcome.

PR #4398, opened 2024-05-15
dex: Generate candlestick data and provide RPCs for access

hdevalence:
## Describe your changes

This adds support for tracking candlestick data to the dex component via a new `Chandelier` trait.

- Whenever a position execution or swap execution occurs, methods in the `Chandelier` are called which will extract and accumulate data into an ordered `Vec` stored in the object store
- at the end of the block, the chandelier collapses the block's collected data points into a `CandlestickData` and saves it in non-verifiable storage under the key `dex/candlesticks/{pair}/{height:020}` so that they can be iterated over in order
- Two RPCs have been added, one for retrieving non-streaming historical data, and another which allows you to watch realtime candlestick data for a given trading pair

#4399

  > Only affects non-verifiable storage


crates/core/component/dex/src/component/candlestick/candlestick.rs
```diff
@@ -0,0 +1,56 @@
...
+    pub(crate) close: f64,
+    /// The highest observed price during the block execution.
+    pub(crate) high: f64,
+    /// The lowest observed price during the block execution.
+    pub(crate) low: f64,
+    /// The volume that traded "directly", during individual position executions.
+    pub(crate) direct_volume: f64,
+    /// The volume that traded as part of swaps, which could have traversed multiple routes.
+    pub(crate) swap_volume: f64,
+}
```
hdevalence, line 23: Does these need to be `pub(crate)`?

crates/core/component/dex/src/component/candlestick/chandelier.rs
```diff
@@ -0,0 +1,429 @@
...
+            // we can iterate through them to create the candlestick data.
+            let mut open = None;
+            let mut close = 0.0;
+            let mut low = f64::INFINITY;
+            let mut high = 0.0;
+            let mut swap_volume = 0.0;
+            let mut direct_volume = 0.0;
+
+            // Create summary data on a per-trading pair basis.
+            for execution in block_executions {
```
hdevalence, line 118: Q: do we know that every `block_executions` that appears here will be nonempty? In other words, do we have to worry about whether our sentinel values like `f64::INFINITY` would leak through into the data we produce?

My assumption is that we wouldn't end up with an empty `block_executions` but wanted to check.

crates/core/component/dex/src/component/position_manager.rs
```diff
@@ -390,6 +392,10 @@ pub trait PositionManager: StateWrite + PositionRead {
             }
         }
 
+        // Update the candlestick tracking
+        self.record_position_execution(&prev_state, &new_state, &context)
+            .await?;
+
```
hdevalence, line 398: Q: would it make sense to combine the candlestick recording with the event emission, either by moving this up to 377 where the event is emitted, or by moving the event emission _inside_ the candlestick call, so that we always record the candlestick data and the event together?

crates/core/component/dex/src/component/rpc.rs
```diff
@@ -193,6 +205,77 @@ impl QueryService for Server {
...
     }
 
+    #[instrument(skip(self, request))]
+    async fn candlestick_data(
+        &self,
+        request: tonic::Request<CandlestickDataRequest>,
+    ) -> Result<tonic::Response<CandlestickDataResponse>, Status> {
+        let state = self.storage.latest_snapshot();
+        let start_height = request.get_ref().start_height;
+        let limit = request.get_ref().limit as usize;
```
hdevalence, line 215: We should use a `std::cmp::max` to clamp this to a constant, say 20_000 (= 100k seconds, just longer than 1 day) so that an RPC caller can't DoS the node by forcing it to accumulate giant amounts of data in memory

crates/core/component/dex/src/component/candlestick/candlestick.rs
zbuc: Yes, otherwise the Chandelier can't instantiate one (unless we make a `new` method but it seems unneccessary)

crates/core/component/dex/src/component/candlestick/chandelier.rs
zbuc: I don't believe it should be possible but I can add some checks to the `record_position_execution` and `record_swap_execution` to make sure that none are included.

crates/core/component/dex/src/component/position_manager.rs
zbuc: I considered this however the event emission and candlestick creation are different concerns and the `position_execution` method is supposed to be the sole method tracking position executions and tying the various subsystems together so this seemed most design-legible to me.

crates/core/component/dex/src/component/candlestick/chandelier.rs
zbuc: https://github.com/penumbra-zone/penumbra/blob/d6299e9dd1519784953139a6a2c37512239fdfe4/crates/core/component/dex/src/component/position_manager.rs#L375-L376  -- should be impossible for empty position executions to be recorded

zbuc: It does look like it's possible for no-op swap executions to be recorded, so an additional check there would be useful.

crates/core/component/dex/src/component/candlestick/candlestick.rs
hdevalence: Sorry, what I meant was, why not make them just `pub`? There's no invariants we're maintaining right?

crates/core/component/dex/src/component/router/fill_route.rs
```diff
@@ -81,7 +81,7 @@ pub trait FillRoute: StateWrite + Sized {
 
 impl<S: StateWrite> FillRoute for S {}
 
-async fn fill_route_inner<S: StateWrite + Sized>(
+async fn fill_route_inner<S: StateWrite + Sized + Chandelier>(
```
erwanor, line 84: let's remove the bound on `Chandelier`

crates/core/component/dex/src/component/position_manager.rs
```diff
@@ -159,7 +161,7 @@ impl<T: StateRead + ?Sized> PositionRead for T {}
 
 /// Manages liquidity positions within the chain state.
 #[async_trait]
-pub trait PositionManager: StateWrite + PositionRead {
+pub trait PositionManager: Chandelier + StateWrite + PositionRead {
```
erwanor, line 164: ditto, the blanket impl lets us skip those bounds

crates/core/component/dex/src/component/dex.rs
```diff
@@ -45,7 +46,7 @@ impl Component for Dex {
     }
 
     #[instrument(name = "dex", skip(state, end_block))]
-    async fn end_block<S: StateWrite + 'static>(
+    async fn end_block<S: StateWrite + 'static + Chandelier>(
```
erwanor, line 49: same, can remove and simplify this bound

crates/core/component/dex/src/component/position_manager.rs
```diff
@@ -390,6 +392,10 @@ pub trait PositionManager: StateWrite + PositionRead {
             }
         }
 
+        // Update the candlestick tracking
+        self.record_position_execution(&prev_state, &new_state, &context)
+            .await?;
```
erwanor, line 397: Can we swallow the error here please, we don't want failure to record auxiliary data to snowball into a chain halt

crates/core/component/dex/src/component/router/route_and_fill.rs
```diff
@@ -109,12 +110,29 @@ pub trait HandleBatchSwaps: StateWrite + Sized {
...
+
+        // Update the candlestick tracking
+        if let Some(se) = swap_execution_1_for_2.clone() {
+            tracing::debug!("updating candlestick for 1=>2 swap");
+            Arc::get_mut(self)
+                .expect("expected state to have no other refs")
+                .record_swap_execution(&se)
+                .await?;
+        }
+        if let Some(se) = swap_execution_2_for_1.clone() {
```
erwanor, line 127: nit: we're passing by reference below, so cloning the execution trace is unnecessary i believe

crates/core/component/dex/src/component/rpc.rs
```diff
@@ -193,6 +205,77 @@ impl QueryService for Server {
...
+            .get_ref()
+            .pair
+            .clone()
+            .ok_or_else(|| Status::invalid_argument("missing trading_pair"))?
+            .try_into()
+            .map_err(|_| Status::invalid_argument("invalid trading_pair"))?;
+        let prefix = state_key::candlesticks::by_pair_and_height(&pair, start_height);
+        tracing::trace!(?prefix, "searching for candlesticks from starting height");
+        let candlesticks = state
+            .nonverifiable_prefix_raw(prefix.as_bytes())
```
erwanor, line 229: I think I saw a comment about this in discord, yes you want a range query `start_height..`

crates/core/component/dex/src/state_key.rs
```diff
@@ -27,6 +27,39 @@ pub fn all_positions() -> &'static str {
...
+        "dex/block_swap_executions"
+    }
+
+    pub fn in_progress(pair: &DirectedTradingPair) -> String {
+        let pair_bytes = hex::encode_upper(pair.encode_to_vec());
+        format!("dex/in_progress_candlestick/{pair_bytes}")
+    }
+
+    pub fn by_pair_and_height(pair: &DirectedTradingPair, height: u64) -> String {
+        let pair_bytes = hex::encode_upper(pair.encode_to_vec());
```
erwanor, line 53: would prefer if we encoded directed trading pairs consistently across the dex component state keys, `start` + `/` + `end` unless there's a specific reason to do it specially here

crates/core/component/dex/src/component/position_manager.rs
erwanor: We could map the error and use `tracing::warn!`

erwanor:
Some suggestions, will take it for a spin trying to generate price data

erwanor (approved):

Issue #4336, opened 2024-05-06
staking: emit ABCI events for validator state transitions


conorsch:
**Is your feature request related to a problem? Please describe.**
Currently, the most authoritative view of the active validator set is obtained by running `pcli query validator list --show-inactive`. We have partial, but crucially flawed, metrics related to validator states that should be removed.

**Describe the solution you'd like**
When a validator changes state, e.g. from Defined -> Active, or Active -> Jailed, or Active -> Tombstoned, an ABCI event should be emitted. This would allow external tooling like block explorers to consume these events.

**Describe alternatives you've considered**
We've tried to shore up the emitted metrics, but decided these aren't the right way to represent the validator state info. 

**Additional context**
See related historical discussion in https://github.com/penumbra-zone/penumbra/issues/3746, and recent PR in  EventTombstoneValidator - https://github.com/penumbra-zone/penumbra/pull/4277.

conorsch:
Talked through the shape of these events with @plaidfinch and @cratelyn today. We want:

* catch-all state transition event: `penumbra.core.component.stake.v1.EventValidatorStateChange`, with fields:
  * validator identity key
  * new state
  * make sure all possible state transitions are accounted for: defined, active, inactive, jailed, tombstoned, enabled, disabled. check genesis states too (which may be earlier than the transition logic)
* missed-block event: `penumbra.core.component.stake.v1.EventValidatorMissedBlock`, with fields:
  * validator identity key
  * block not signed
  * (this event is only emitted if validator was Active but did not sign)

PR #4401, opened 2024-05-16
fix(pd): remove validator state metrics

conorsch:
## Describe your changes
The validator state metrics are not accurate, and should be refactored to use ABCI event emission so that developers can leverage the state transitions in a sane way. In the meantime, let's remove the incorrect metrics entirely.

Refs #3746, #4336.

  > only affects metrics emission and display, no changes to consensus logic


cratelyn (approved):
nice! thanks :pray:

PR #4403, opened 2024-05-16
Chore: remove unused deps

zbuc:
## Describe your changes

Used `cargo +nightly udeps --all-targets --all-features` to detect unused dependencies.

  > Only affects cargo dependencies


cratelyn (approved):
TIL! thanks for doing this 🙂

PR #4404, opened 2024-05-16
Implement IBC-related APIs for Skip integration

### LINKED ISSUE #4391: Implement missing IBC APIs for Skip
From Skip:

> I took a look at which grpc endpoints are working that Skip API depends on, and most of the IBC ones seem to work. I found three that we depend on that would be benefitial if we could have them implemented.

```
cosmos.bank.v1beta1.Query/TotalSupply
https://github.com/cosmos/cosmos-sdk/blob/main/x/bank/proto/cosmos/bank/v1beta1/query.proto#L57

ibc.applications.transfer.v1.Query/DenomTraces
https://github.com/cosmos/ibc-go/blob/main/proto/ibc/applications/transfer/v1/query.proto#L16

ibc.core.client.v1.Query/ClientStatus
https://github.com/cosmos/ibc-go/blob/main/proto/ibc/core/client/v1/query.proto#L47
```

zbuc:
## Describe your changes

This changeset implements some APIs Skip indicated would be useful for their integration efforts.

Specifically:

* `cosmos.bank.v1beta1.Query/TotalSupply` -- since we aren't using the Cosmos SDK, this didn't make sense to implement directly -- instead we've implemented a similar streaming `penumbra.core.component.shielded_pool.v1.QueryService/TotalSupply` call. Please note that only IBC transferred assets will have correct amounts set; internally tracked assets currently return a 0 amount (Skip said this is fine for their uses for now)
* `ibc.applications.transfer.v1.Query/DenomTraces`
* `ibc.core.client.v1.Query/ClientStatus`

I am not happy with some of the code I had to write to massage data into the right types but didn't have time to figure out anything better. I manually tested all three RPC methods locally and they seem to work as expected.

Closes #4391 

  > Only creates new RPCs that return data from state


crates/core/component/shielded-pool/src/component/rpc/transfer_query.rs
```diff
@@ -0,0 +1,91 @@
...
+        _: tonic::Request<QueryTotalEscrowForDenomRequest>,
+    ) -> std::result::Result<tonic::Response<QueryTotalEscrowForDenomResponse>, tonic::Status> {
+        unimplemented!()
+    }
+
+    async fn escrow_address(
+        &self,
+        _: tonic::Request<QueryEscrowAddressRequest>,
+    ) -> std::result::Result<tonic::Response<QueryEscrowAddressResponse>, tonic::Status> {
+        unimplemented!()
```
avahowell, line 34: Noting for the future that I think we can implement this by querying the value balance key, even though we don't have escrow accounts per se.

avahowell (approved):
LGTM

conorsch:
There's a CI failure here. On first pass, the check-all-features job failed, after which @zbuc added the "server" feature flag on the shielded-pool crate's `ibc-proto` import. However, that change broke the wasm-compat job. The former is preferable to the latter if we must break something, but let's try to get it passing before merge.

conorsch:
Nice, the latest feature-tweak resolved. Merging to get this included in upgrade testing, pending resolution of #4430.

hdevalence:
image

We can't and shouldn't implement a custom RPC like this. The issue was to implement the RPC needed to interoperate with existing tooling, not to make up our own.

PR #4406, opened 2024-05-16
add sync exclude flag

### LINKED ISSUE #4405: "sync_exclude" flag for Spends and Outputs
We want to be able to mark Spends/Outputs as not to be included in CompactBlocks after submission for high-volume applications, to make the view server more efficient on sync. We can do this by adding a new flag to Spends and Outputs, `sync_exclude`

To do this we can:

- [ ] Add another map() to note_payloads in finalize_compact_block to filter out relevant payloads based on this flag
- [ ] Modify Spend action to add the exclude flag
- [ ] Modify  SctManager:StateWrite:nullify to accept an additional argument to exclude from pending_nullifiers

- Questions
	- Do we want to store the associated notes and nullifiers in the view service on submission to maintain a historical record on the originating instance? Or do we need to provide for this at all?
	- Do we want to provide a "full sync" option that includes the "sync_exclude" transactions? This would allow someone to opt-in to the less efficient but more complete behavior currently implemented in the view server.

aubrika:

Closes #4405 

  > Doesn't touch any verifiable storage or circuits, makes only additive protobuf changes


crates/core/component/compact-block/src/component/manager.rs
```diff
@@ -91,6 +91,8 @@ trait Inner: StateWrite {
         let note_payloads = self
             .pending_note_payloads()
             .into_iter()
+            // Filter out notes with the compact block exclusion flag set.
+            .filter(|(_pos, note, _source)| !note.sync_exclude)
```
hdevalence, line 95: This will cause an SCT divergence, the note commitments need to be included in the compact block regardless of whether the note payload is included.

crates/core/component/shielded-pool/src/note_payload.rs
```diff
@@ -14,6 +14,7 @@ pub struct NotePayload {
     pub note_commitment: note::StateCommitment,
     pub ephemeral_key: ka::Public,
     pub encrypted_note: NoteCiphertext,
+    pub sync_exclude: bool,
```
hdevalence, line 17: Why do we want this to be a property of the `NotePayload` rather than the `Output`?

proto/penumbra/penumbra/core/component/shielded_pool/v1/shielded_pool.proto
```diff
@@ -60,6 +60,8 @@ message NotePayload {
   // An encryption of the newly created note.
   // 176 = 80(address) + 16(amount) + 32(asset ID) + 32(rseed) + 16(MAC) bytes.
   NoteCiphertext encrypted_note = 3;
+  // Whether to exclude this spend from the CompactBlock.
+  bool sync_exclude = 4;
```
hdevalence, line 64: This shouldn't be a property of the `NotePayload`, which represents an encrypted note to be scanned, because the point of the field is to mark it as not-to-be-scanned. Instead, the field should go on the `OutputBody` proto message.

aubrika:
Closing this one as underspecified - the proposed changes have consequences in need of more thorough design work

TalDerei:
> Closing this one as underspecified - the proposed changes have consequences in need of more thorough design work

maybe this is something we can revisit post mainnet? client-sync V2 tracked here: https://hackmd.io/MTA83QjNQoqA8Ela0atQYg

aubrika:
> > Closing this one as underspecified - the proposed changes have consequences in need of more thorough design work
> 
> maybe this is something we can revisit post mainnet? client-sync V2 tracked here: https://hackmd.io/MTA83QjNQoqA8Ela0atQYg

I think we can, it's a performance enhancement & having zero fees for Spends would be good for tx sweep, but in terms of priority & added bug/testing surface, this is a changeset I don't feel great about firing from the hip in terms of assurance before V1

PR #4412, opened 2024-05-18
feat: add method to `asset::Id` that gives the base64 encoding of the inner bytes

### LINKED ISSUE #4192: Add method to `asset::Id` that gives the base64 encoding of the inner bytes

This appears in ProtoJSON and it'd be convenient to have direct access to it.

Add a method to `asset::Id` returning the base64 encoding of the `inner` bytes, as would appear with ProtoJSON serialization.

TropicalDog17:
## Describe your changes
add method to `asset::Id` that gives the base64 encoding of the inner bytes

Closes #4192

  > It's just a helper method


erwanor (approved):
Looks great, thanks!

PR #4415, opened 2024-05-20
ci: fix rust-docs build

conorsch:
## Describe your changes
Bumped the `time` crate in order to resolve a build error with recent rust nightlies on the `rust-docs` build. Also removed the `index.html` file generated for rust-docs from version control, since that file is built on the fly in CI post-merge, and so will always be out of date.

  > small version bump of a dependency, no changes to application logic


conorsch:
See example failing build here: https://github.com/penumbra-zone/penumbra/actions/runs/9158128504/job/25175841362?pr=4413 the `Docs lint` job should pass on this PR.

erwanor (approved):

PR #4423, opened 2024-05-20
view: add upsert behavior to record_block to respect key uniqueness constraints for duplicate commitments

### LINKED ISSUE #3945: view service should gracefully handle duplicate note commitment during sync
**Describe the bug**
If an attacker sends a duplicate note commitment to a target address, when the target tries to sync the view service will hang (PING/REJECT vector)

**To Reproduce**

You need to send two notes with the same (address, value, rseed) to the target.
1. Set the Rseed in `OutputPlan::new` to e.g. `Rseed([0u8; 32])`. Relevant line [here](https://github.com/penumbra-zone/penumbra/blob/main/crates/core/component/shielded-pool/src/output/plan.rs#L36).
2. Send the same value to the target address twice, e.g. make two transactions sending 1upenumbra to the address
3. As the target, sync

**Expected behavior**

View service syncs without hanging

**Actual behavior**

View service hangs due to unique constraint violation:

```
2024-03-04T23:34:11.859746Z ERROR penumbra_view::worker: view worker error e=UNIQUE constraint failed: spendable_notes.note_commitment

Caused by:
    Error code 1555: A PRIMARY KEY constraint failed
Scanning blocks from last sync height 96893 to latest height 97379
[0s] ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░       0/486     0/s ETA: 0s
^C
```

### LINKED ISSUE #4136: view service should gracefully handle duplicate swap commitment during sync
**Describe the bug**
This is closely related to #3945 
A duplicate swap commitment will cause a hang of the view service for the swap claimant, but can only be triggered by someone in possession of the outgoing viewing key of the target, so does not have the same implications as #3945 in terms of privacy. 

**To Reproduce**
1. Set Rseed (generated in `SwapPlaintext::new`) for two identical swaps to `Rseed([0u8; 32])` 
2. Sync as the target

**Expected behavior**
Sync occurs without issue

**Actual behavior**
Sync hangs:

```
2024-03-28T22:56:32.729108Z ERROR penumbra_view::worker: view worker error e=UNIQUE constraint failed: swaps.swap_commitment

Caused by:
    Error code 1555: A PRIMARY KEY constraint failed
2024-03-28T22:56:42.867326Z ERROR penumbra_view::worker: view worker error e=UNIQUE constraint failed: swaps.swap_commitment

Caused by:
    Error code 1555: A PRIMARY KEY constraint failed
```

aubrika:
## Describe your changes

Closes #3945 
Closes #4136 

  > This only changes the view server behavior in record_block, so only non-consensus storage.

### DIFF
--- a/crates/view/src/storage.rs
+++ b/crates/view/src/storage.rs
@@ -1241,7 +1241,12 @@ impl Storage {
         dbtx.execute(
             "INSERT INTO notes (note_commitment, address, amount, asset_id, rseed)
                 VALUES (?1, ?2, ?3, ?4, ?5)
-                ON CONFLICT DO NOTHING",
+                ON CONFLICT (note_commitment) 
+                DO UPDATE SET 
+                address = excluded.address, 
+                amount = excluded.amount, 
+                asset_id = excluded.asset_id, 
+                rseed = excluded.rseed",
             (note_commitment, address, amount, asset_id, rseed),
         )?;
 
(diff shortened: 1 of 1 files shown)


erwanor:
This makes sense to me, but would like to defer to @redshiftzero to make sure this address the ping/reject

redshiftzero (approved):
LGTM! For posterity: I verified interactively that this is resolved following the STR in #3945. 

One other thought: Since we've had this bug twice in the form of #1507 and #3945, I've filed a ticket (#4506) to write a regression test which would be wise to implement given the privacy impact.

PR #4424, opened 2024-05-20
tonic: 🍸 patch tonic dependencies

cratelyn:
fixes #4392.

in #4392, we found an issue with gRPC reflection when recent versions of `grpcurl` were used to list supported endpoints.

this was tracked down and fixed in #hyperium/tonic#1701, in collaboration with @conorsch. that pr targets the more recent 0.11 tonic release however, which we are not yet using.

eventually we should upgrade to tonic 0.11 (see #4400). as a short-term measure, but we want to unblock engineers from Skip working on an integration with Penumbra. this commit patches the `tonic` dependencies used to point to a temporary fork of the `tonic` repo, where i have backported that patch (and some ci-related fixes) to the 0.10 release.

see also:
* https://github.com/hyperium/tonic/pull/1701
* https://github.com/penumbra-zone/tonic/pull/1
* https://github.com/penumbra-zone/tonic/pull/2
* #4392
* #4400

---

checking that all transitive dependencies point to the patched version:

```
; cargo tree | grep tonic
│   └── tonic v0.10.2 (https://github.com/penumbra-zone/tonic.git?tag=v0.10.3-penumbra#db355dd7)
├── tonic v0.10.2 (https://github.com/penumbra-zone/tonic.git?tag=v0.10.3-penumbra#db355dd7) (*)
│   ├── tonic v0.10.2 (https://github.com/penumbra-zone/tonic.git?tag=v0.10.3-penumbra#db355dd7) (*)
│   │   │   │   ├── tonic v0.10.2 (https://github.com/penumbra-zone/tonic.git?tag=v0.10.3-penumbra#db355dd7) (*)
│   │   │   │   ├── tonic v0.10.2 (https://github.com/penumbra-zone/tonic.git?tag=v0.10.3-penumbra#db355dd7) (*)
│   │   │   │   │   ├── tonic v0.10.2 (https://github.com/penumbra-zone/tonic.git?tag=v0.10.3-penumbra#db355dd7) (*)
│   │   │   │   ├── tonic v0.10.2 (https://github.com/penumbra-zone/tonic.git?tag=v0.10.3-penumbra#db355dd7) (*)
│   │   │   ├── tonic v0.10.2 (https://github.com/penumbra-zone/tonic.git?tag=v0.10.3-penumbra#db355dd7) (*)
│   │   ├── tonic v0.10.2 (https://github.com/penumbra-zone/tonic.git?tag=v0.10.3-penumbra#db355dd7) (*)
[...]

Issue #4392, opened 2024-05-15
Unexpected HTTP response codes for unimplemented GRPC methods


hdevalence:
**Describe the bug**

Skip reported unexpected HTTP response codes for unimplemented GRPC methods:

> Another thing I noticed I think has to do with the problem I was running into with grpcurl. Instead of returning "Unimplemented" GRPC error for unimplemented queries, I get back an unknown error type
> 
> 2024/05/15 15:05:12 rpc error: code = Unknown desc = unexpected HTTP status code received from server: 405 (Method Not Allowed); malformed header: missing HTTP content-type
exit status 1
> 
> Skip API uses the unimplemented error type to detect feature support in certain cases. We may need to handle this other error type specially since it seems like we are getting it instead of unimplemented.

They also reported that `grpcurl` didn't work, potentially because of problems with reflection.

This was against `https://grpc.testnet.penumbra.zone`; it's possible the custom load balancer setup there interferes with HTTP headers, we've had that issue in the past, but this seems more likely to be a problem in `pd` itself.

---

### :link: related work
* https://github.com/hyperium/tonic/pull/1701
* https://github.com/penumbra-zone/tonic/pull/1
* https://github.com/penumbra-zone/tonic/pull/2

hdevalence:
Link to gist with pd logs: https://gist.github.com/hdevalence/3cc8c9cef56463dc08cf4e303f316ba4

```
$ grpcurl -plaintext -vv localhost:8080 list
Failed to list services: rpc error: code = Unknown desc = unexpected HTTP status code received from server: 405 (Method Not Allowed); malformed header: missing HTTP content-type
```

conorsch:
> They also reported that grpcurl didn't work

Using `grpcurl` with a known-good query works fine:

```
❯ grpcurl grpc.testnet.penumbra.zone:443 penumbra.core.app.v1.QueryService.AppParameters  | jq .appParameters.chainId
"penumbra-testnet-deimos-8"
```

> potentially because of problems with reflection.

Reflection is also supported, shown here via `grpcurl`:

```
❯ grpcurl grpc.testnet.penumbra.zone:443 list
grpc.reflection.v1alpha.ServerReflection
penumbra.cnidarium.v1.QueryService
penumbra.core.app.v1.QueryService
penumbra.core.component.auction.v1.QueryService
penumbra.core.component.community_pool.v1.QueryService
penumbra.core.component.compact_block.v1.QueryService
[...]

PR #4425, opened 2024-05-21
Include vendored proto definitions in the proto_descriptor file

### LINKED ISSUE #4422: Vendored gRPC services aren't present in reflection
While testing the new IBC interfaces added in #4391 I noticed that third-party gRPC services don't appear in gRPC reflection because the list of reflected services comes from the Penumbra `gen/proto_descriptor.bin.no_lfs` file, and since these services' Protobuf files are vendored they aren't present.

@conorsch suggested that we might be able to accomplish this if we can shim references to the vendored services in to the proto descriptor file via `tonic-reflection`.

zbuc:
## Describe your changes

This changes the proto build process to include cosmos/ibc-go proto files in the vendored proto file list used to generate the proto_descriptor file.

As-implemented, there are some duplicate paths that appear both as dependencies of the penumbra protos and in ibc-go. The ibc-go protos will be included first so the penumbra protos take priority. This shouldn't be an issue in the way we currently use the protos but this seems important to document for future reference.

Closes #4422

  > Just changes gRPC reflection returns


crates/bin/pclientd/src/lib.rs
```diff
@@ -338,6 +338,8 @@ impl Opt {
                     .add_service(tonic_web::enable(stake_query_proxy))
                     .add_service(tonic_web::enable(compact_block_query_proxy))
                     .add_service(tonic_web::enable(tendermint_proxy_proxy))
+                    // TODO: should we add the IBC services here as well? they will appear
+                    // in reflection but not be available.
```
conorsch, line 342: As a general rule, I'd say any services that _can_ be exposed in both `pd` and `pclientd`, meaning they'll work equally well in both contexts, probably should be. Otherwise we're starting down a path where there are odd discrepancies that are hard to document. Not worth blocking merge here though, and I appreciate you adding this TODO.

conorsch (approved):
This is a slick change, I'm bullish on merge. 

It's possible that judicious use of vendoring here may affect downstream consumers of our protos: I'm thinking in particular of `interchaintest` pulling in copies of `buf.build/penumbra-zone` and also its own imports of the cosmos deps. In order to evaluate behavior there, we need to press on with merging, pushing to buf, and re-pulling in those other repos, which is too long of a round trip to block merge on now, given the priority on unblocking the Skip IBC work.

conorsch:
Confirmed that the new IBC methods are exposed via reflection by testing locally:

```
❯ grpcurl_1.8.7 -plaintext localhost:8080 list
grpc.reflection.v1alpha.ServerReflection
ibc.core.channel.v1.Query
ibc.core.client.v1.Query
ibc.core.connection.v1.Query
penumbra.cnidarium.v1.QueryService
penumbra.core.app.v1.QueryService
penumbra.core.component.auction.v1.QueryService
penumbra.core.component.community_pool.v1.QueryService
penumbra.core.component.compact_block.v1.QueryService
penumbra.core.component.dex.v1.QueryService
penumbra.core.component.dex.v1.SimulationService
penumbra.core.component.fee.v1.QueryService
penumbra.core.component.governance.v1.QueryService
penumbra.core.component.sct.v1.QueryService
penumbra.core.component.shielded_pool.v1.QueryService
penumbra.core.component.stake.v1.QueryService
penumbra.custody.v1.CustodyService
penumbra.tools.summoning.v1.CeremonyCoordinatorService
penumbra.util.tendermint_proxy.v1.TendermintProxyService
penumbra.view.v1.ViewService
```

whereas they are not present on main:

```
❯ grpcurl_1.8.7 -plaintext localhost:8080 list
grpc.reflection.v1alpha.ServerReflection
penumbra.cnidarium.v1.QueryService
penumbra.core.app.v1.QueryService
penumbra.core.component.auction.v1.QueryService
penumbra.core.component.community_pool.v1.QueryService
penumbra.core.component.compact_block.v1.QueryService
penumbra.core.component.dex.v1.QueryService
penumbra.core.component.dex.v1.SimulationService
penumbra.core.component.fee.v1.QueryService
penumbra.core.component.governance.v1.QueryService
penumbra.core.component.sct.v1.QueryService
penumbra.core.component.shielded_pool.v1.QueryService
penumbra.core.component.stake.v1.QueryService
penumbra.custody.v1.CustodyService
penumbra.tools.summoning.v1.CeremonyCoordinatorService
penumbra.util.tendermint_proxy.v1.TendermintProxyService
penumbra.view.v1.ViewService
```

N.B. the older `1.8.7` version of `grpcurl` is necessary, pending resolution of #4392. Fortunately, we have #4424 queued up to resolve.

Issue #4400, opened 2024-05-16
chore: 🧃 upgrade to `tonic@0.11.0`


cratelyn:
relates to:
* #4392
* hyperium/tonic#1701
* https://github.com/informalsystems/tendermint-rs/pull/1422
* https://github.com/informalsystems/tendermint-rs/pull/1423
* https://github.com/informalsystems/tendermint-rs/pull/1424

we observed some issues related to gRPC reflection in #4392, and subsequently opened a patch upstream to add the v1 version of the server reflection facilities.

tonic has since released a [0.11.0](https://crates.io/crates/tonic/0.11.0) version. once that patch is released, we'll need to bump our dependency on tonic to pull in those changes. doing so will involve upgrading from 0.10.2 to 0.11.0, which we should ideally handle first.

conorsch:
Resolved in #4997, which shipped as part of v1.0.0.

Issue #4264, opened 2024-04-24
Enhance script to support mac binary installation for cometBFT


TalDerei:
We currently have a utility script designed to download and install a specific version of CometBFT for use with Penumbra, primarily in CI environments. This script is tailored to Linux systems and does not support macOS, especially the newer Darwin Arm64 architecture.

Enhance the existing script to automatically detect the operating system and architecture (x86-64 for Intel and Arm64 for M1/M2/M3 chips) of the user’s machine, and download the appropriate binary for macOS alongside the existing support for Linux.

The alternative is downloading the proper [cometBFT binary version](https://github.com/cometbft/cometbft/releases/tag/v0.37.5) for Arm64 from source.

PR #4426, opened 2024-05-21
test(pd): add integration test for grpc reflection

conorsch:
## Describe your changes

Refs #4392. We want to ensure that server reflection remains working, despite changes to the tonic dependencies (#4400) and proto compiling logic #4422. While the most effective test is exercising all these protos regularly, we currently lack solid coverage, so this superficial integration test is a sanity-check to save time versus building locally and inspecting the output manually.

  > tests-only

### DIFF
--- a/.github/workflows/smoke.yml
+++ b/.github/workflows/smoke.yml
@@ -28,6 +28,9 @@ jobs:
           sh -c "$(curl --location https://raw.githubusercontent.com/F1bonacc1/process-compose/main/scripts/get-pc.sh)" --
           -d -b ~/bin
 
+      - name: Install grpcurl
+        run: ./deployments/scripts/install-grpcurl
+
       - name: Run the smoke test suite
         run: |
           export PATH="$HOME/bin:$PATH"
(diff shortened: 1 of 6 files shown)


conorsch:
Draft because I cobbled this together as part of initial work on #4392, and it proved useful while reviewing #4424 & #4425. Still needs some attention in getting CI working, and also pulling in the FILE_DESCRIPTOR set from the proto crate. Will circle back on it opportunistically next time I'm waiting on a long build.

cratelyn (commented):
i am in deep, earnest support of this! :raised_hands: a tested feature is a non-regressing feature. thank you for starting this work @conorsch!

.github/workflows/smoke.yml
```diff
@@ -28,6 +28,9 @@ jobs:
           sh -c "$(curl --location https://raw.githubusercontent.com/F1bonacc1/process-compose/main/scripts/get-pc.sh)" --
           -d -b ~/bin
 
+      - name: Install grpcurl
+        run: ./deployments/scripts/install-grpcurl
+
```
conorsch, line 33: @cratelyn N.B. I'd much rather use something like nix flakes or devbox to manage these dependencies. Not going down that rabbit hole now, but flagging that it'd be less work overall to track tooling deps in a holistic way, cf. #4264.

cratelyn (approved):
:test_tube: :tada:

PR #4427, opened 2024-05-21
Update static sites

grod220:
## Describe your changes

Pulls latest node status + minifront app (commit e35c6f7) and updates them on PD. 

  - Only changing static sites


conorsch:
Confirmed locally that the node-status page is rendering fine:

node-status-pr-4427

That's in addition to the "200 OK" sanity check test we added back in https://github.com/penumbra-zone/penumbra/pull/4139. If there's anything else I should be looking for specifically, holler @grod220 , otherwise LGTM!

PR #4428, opened 2024-05-21
dex: deduct arb surplus from component VCB

### LINKED ISSUE #4113: Arb execution does not update value circuit breaker
**Describe the bug**

The arb execution doesn't update the value circuit breaker. Since it burns the tokens, they should be debited from the DEX balance.

erwanor:
## Describe your changes

Close #4113

  > This is consensus-breaking and state-breaking, but does not require any migration to heal the dex vcb


hdevalence (approved):
LGTM

PR #4439, opened 2024-05-22
pcli: improve error message for delegation failure to suggest pcli tx sweep

### LINKED ISSUE #4414: pcli ux: delegation failure should recommend pcli sweep
A commonly reported problem in the Discord is that during a delegation attempt via pcli, an error is encountered:

```
Error: can't plan delegation
Caused by: 
    failed to plan transaction after 100 iterations
```

The solution to this problem is nearly always "run `pcli tx sweep`". It'd be nice if the error message recommended just that, so folks could resolve on their own.

The error message returned by pcli in this failure mode should explicitly recommend running `pcli tx sweep`, then re-run the delegation command.

We could futz around with the limits. Switching to `thiserror` would make it a bit more straightforward to perform this type of error-handling, but an anyhow::Context would be more than sufficient to help out folks using the CLI tooling.

**Additional context**

aubrika:
Closes #4414 

## Describe your changes

Changes a pcli error message to be more helpful

  > Only changes pcli output text

### DIFF
--- a/crates/bin/pcli/src/command/tx.rs
+++ b/crates/bin/pcli/src/command/tx.rs
@@ -562,7 +562,7 @@ impl TxCmd {
                     .delegate(epoch, unbonded_amount, rate_data)
                     .plan(app.view(), AddressIndex::new(*source))
                     .await
-                    .context("can't plan delegation")?;
+                    .context("can't plan delegation, try running pcli tx sweep and try again")?;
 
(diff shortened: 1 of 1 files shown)


cratelyn (approved):
:broom: ["_let's see that dust fly with that broom_"](https://www.youtube.com/watch?v=9xAi80ShG0I) :musical_note:

PR #4441, opened 2024-05-23
asset: add `asset.v1.Metadata.priority_score`

hdevalence:
## Describe your changes

This is intended to help support clients that need to display lists of assets. An asset registry can assign rankings to tokens, and frontend software can consume them without needing to have its own opinions about what the ranking should be. For instance, a registry could have hardcoded "tiers" or priorities, or priorities assigned by market cap, or the user's balances, etc.

  > This adds a new optional field to the proto which is used on-chain but never populates it with a non-default value, so encodings should not change.


grod220 (approved):
LGTM

PR #4442, opened 2024-05-23
Correct display of downtime in `pcli validator query uptime`

plaidfinch:
The downtime penalty is 0.01%, not 1% as previously displayed. This corrects the erroneous calculation which previously displayed an incorrect figure by a 100x factor.

## Describe your changes

This adds a `/ 100.0` to correct the missing factor, and an explanatory comment.

  > This is a display fix for `pcli`.

### DIFF
--- a/crates/bin/pcli/src/command/query/validator.rs
+++ b/crates/bin/pcli/src/command/query/validator.rs
@@ -286,8 +286,9 @@ impl ValidatorCmd {
                     100.0 * params.stake_params.missed_blocks_maximum as f64 / window_len as f64;
                 let percent_downtime = 100.0 * missed_blocks as f64 / window_len as f64;
                 let percent_downtime_penalty =
-                    // Converting from basis points squared to percentage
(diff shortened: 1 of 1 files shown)


crates/bin/pcli/src/command/query/validator.rs
```diff
@@ -286,8 +286,9 @@ impl ValidatorCmd {
                     100.0 * params.stake_params.missed_blocks_maximum as f64 / window_len as f64;
                 let percent_downtime = 100.0 * missed_blocks as f64 / window_len as f64;
                 let percent_downtime_penalty =
-                    // Converting from basis points squared to percentage
-                    params.stake_params.slashing_penalty_downtime as f64 / 100.0 / 100.0;
+                    // Converting from basis points squared to percentage: bais point ^ 2  %-age
```
cratelyn, line 289:
```suggestion
                    // Converting from basis points squared to percentage: basis point ^ 2  %-age
```

erwanor (approved):

PR #4444, opened 2024-05-23
Remove bespoke TotalSupply query, implement cosmos bank RPC instead

zbuc:
## Describe your changes

This follows up on work in https://github.com/penumbra-zone/penumbra/pull/4404 by removing the bespoke TotalSupply implementation and implementing the Cosmos Bank `cosmos.bank.v1beta1.Query/TotalSupply` RPC instead.

  > Changes read-only RPCs only


deployments/scripts/protobuf-codegen
```diff
@@ -18,9 +18,10 @@ pushd "${repo_root}/proto/"
 # Make sure the lockfile is up to date.
 buf dep update penumbra
 
-# Pull our vendored IBC-go proto defs so we can get reflection for service definitions.
+# Pull our vendored cosmos/IBC proto defs so we can get reflection for service definitions.
 # The penumbra dependencies will override some of these.
-buf export buf.build/penumbra-zone/ibc-go-vendor --output rust-vendored/
+buf export buf.build/cosmos/cosmos-sdk:e7a85cef453e4b999ad9aff8714ae05f --output rust-vendored/
+buf export buf.build/cosmos/ibc --output rust-vendored/
```
conorsch, line 24: In https://github.com/penumbra-zone/penumbra/blob/a9d5c7b5fcb5188804695c79144d4afe660fee05/proto/penumbra/buf.yaml#L7 we have some version-locking of these same deps, as they are included in the penumbra protos. Can we add some version-pinning here? As written, it looks like the `buf.build/cosmos/ibc` dep will pull from latest every time this script is run, which may not be what we want: controlled updates for regenerating the protos makes the most sense.

Unfortunately, there's not a good semver system, so in the past I've been inspecting hashes manually to pin deps.

crates/core/component/shielded-pool/src/component/rpc/bank_query.rs
```diff
@@ -0,0 +1,198 @@
...
+                )
+                .collect::<Vec<ibc_proto::cosmos::base::v1beta1::Coin>>(),
+        }))
+    }
+
+    async fn params(
+        &self,
+        _: tonic::Request<QueryParamsRequest>,
+    ) -> std::result::Result<tonic::Response<QueryParamsResponse>, tonic::Status> {
+        unimplemented!()
```
cratelyn, line 132:
```suggestion
        Err(Status::unimplemented("params endpoint is not implemented"))
```

can we return an unimplemented `Status` rather than calling `unimplemented!()` here? this way these endpoints won't allow a malicious caller to cause a node to panic.

zbuc: @cratelyn good catch, I see some other uses of `unimplemented!()` elsewhere in the codebase, we should probably audit those and make sure they're not triggerable in the nodes

cratelyn: agreed! with #4446 filed and 8a0c1d9 in hand, i'll mark this as resolved. thank you!

crates/core/app/src/rpc.rs
```diff
@@ -109,6 +110,9 @@ pub fn router(
         .add_service(we(TransferQueryServer::new(ShieldedPoolServer::new(
             storage.clone(),
         ))))
+        .add_service(we(BankQueryServer::new(ShieldedPoolServer::new(
```
avahowell, line 113: it does feel weird to have a bank query service, since we do not have a cosmos bank module. this makes me worried that in the future an ecosystem user will attempt to make RPCs against them, expecting them to work like how the cosmos Bank module is specified.

crates/core/component/shielded-pool/src/component/rpc/bank_query.rs
```diff
@@ -0,0 +1,198 @@
...
+                )
+                .collect::<Vec<ibc_proto::cosmos::base::v1beta1::Coin>>(),
+        }))
+    }
+
+    async fn params(
+        &self,
+        _: tonic::Request<QueryParamsRequest>,
+    ) -> std::result::Result<tonic::Response<QueryParamsResponse>, tonic::Status> {
+        Err(tonic::Status::unimplemented("not implemented"))
```
avahowell, line 132: wondering if we should give a more descriptive error here, which indicates that these can *never* be implemented, since penumbra is a private chain

avahowell:
LGTM apart from the one comment about more descriptive unimplemented errors

zbuc:
@hdevalence can we get an explicit +1 from you on this?

hdevalence:
Memorializing some discussion in Discord, about whether to import all the protos or use the `cosmos_sdk_proto` crate: https://docs.rs/cosmos-sdk-proto/latest/cosmos_sdk_proto/cosmos/index.html

My preference would be to use the `cosmos_sdk_proto` crate to pull in the generated proto code, so that we don't have to maintain our own proto generation of all the Cosmos SDK protos. (Are we going to commit to maintaining them going forward? What level of effort do we want to put into that?)

@zbuc mentioned that this has the downside that the cosmos methods would be excluded from reflection. 

Thinking of @avahowell's comment here, https://github.com/penumbra-zone/penumbra/pull/4444#discussion_r1612145024 , I'm wondering if that's actually an upside -- if we don't add reflection, we're not advertising support, but it's still there if someone knows to make the RPC call.

However, I'm fine with either approach.

zbuc:
> (Are we going to commit to maintaining them going forward? What level of effort do we want to put into that?)

We already had _some_ of them vendored by virtue of being dependencies of Penumbra protos, and I'm not sure the effort of maintaining _some_ vs _all_ is considerably different (basically just keeping the commit hash in sync as @conorsch pointed out)

zbuc:
Re: necessity of reflection -- I started using `grpcurl` recently and find the reflection useful (it's impossible to use it to call against an endpoint if it's not available in reflection...)

The efficiency of being able to test these endpoints quickly with `grpcurl` is pretty great so I would prefer to expose them in reflection

hdevalence:
Sounds good! let's do this then.

conorsch:
> it's impossible to use it to call against an endpoint if it's not available in reflection...

Strictly speaking it's possible to pass proto files on the CLI directly, via `grpcurl -proto <FILE> ...`, but that is an absolutely nightmare of devex. It's also worth noting that reflection is how interfaces like https://grpcui.testnet-preview.penumbra.zone/ work (same maintainers as `grpcurl`): notice that grpcui preview currently has the new IBC RPCs from #4404, but gprcui testnet doesn't yet.

So, hard agree on continuing to support reflection.

conorsch:
Once merged, this closes #4391.

conorsch (approved):

Issue #4443, opened 2024-05-23
pd: rework the persistent halt mechanism


erwanor:
**Context** 
`pd` implements a “persistent halt” mechanism to shutdown the node on restart if some halting predicate is true (based on a counter, a flag, whatever). That is to prevent full nodes from making progress at an upgrade or emergency height.

**Problems**
There are two issues with the existing mechanism:
1. It leaves `pd` in a zombie state - more on that below.
2. It exits the process before `pd` has had time to post a `Commit` response to cometbft.

The first point is a problem because it makes validators attempt to make forward progress on the current consensus round, which defeats the purpose of having persistent halts in the first place.

The second point is also problematic because it makes the `cometbft` state lag behind the application state. This was previously glossed over because the full node in a zombie state was able to enter recovery mode and replay the missing block.

**More on zombie mode**
If the node restarts, `pd` spins up four tower services for each ABCI connections: `Info`, `Consensus`, `Mempool`, and `Snapshot` (unused). The `Consensus` and `Mempool` services error on instantiating an `App`. Crucially though, `pd` does not crash, the `Info` service is still up (bug 1).

As a result, when `cometbft` comes back online it will initiate a recovery which involves making an `Info` request to the application to learn about the latest height. It learns that it is behind and catch-up one block (bug 2).

After that, if the timeout period has elapsed AND the comet node is a proposer then it will try to advance the chain state by a block: first by locking the mempool, collecting transactions and asking the app to authorize it (via `PrepareProposal`). At this point `pd` crashes because the consensus service has been wounded down (this is the current “halt” mechanism).

Other proposer nodes will attempt to lock in a new round, gossiping nil prevotes, and advancing their signing state (unwanted but correct outcome of the network not halting completely).

**Desired behavior** 
There are two high-level things that should happen:
-  if `pd` halts, it should halt, it should not serve any requests.
- `cometbft` must receive `Commit` responses. 

**Proposed fix**
A proposed fix for zombie halts:
- reform the code to halt the node completely (#4436)

A proposed fix for allowing `Commit` responses to comet:
- at halt/upgrade boundaries: use a timeout to give time for `pd` to reply with a `Commit` abci response
- refuse to prepare proposals if the chain halt bit is on

PR #4445, opened 2024-05-23
pd(migrate): set halt flag to false

erwanor:
## Describe your changes

This sets the halt flag to `false` after running the `Testnet76` migration. 

  > Migration change


conorsch:
Nice catch. Noticed this on initial migration testing, but go sidetracked debugging #4432 / #4443.

conorsch (approved):

PR #4447, opened 2024-05-23
proto: 🔧 conversions for `tendermint-*` compatibility

cratelyn:
#### 💭 describe your changes

in order to integrate mock consensus tests with the view server, we will need
the ability to respond to requests made by the view server to the tendermint
proxy service.

currently conversions from `tendermint`, `tendermint-rpc`, and
`tendermint-proto` types into `penumbra-proto` types are defined inline inside
of the `TendermintProxyService` implementation.

to aid that journey, this branch takes that code and decomposes it into a
family of `From<T>` and `TryFrom<T, Error = E>` implementations. this is not
a comprehensive pass at a compatibility layer between `penumbra-proto` and the
`tendermint` crates, it is tightly focused on laying the groundwork to
mock this service in tests.

to prevent the additions of new dependencies from interfering with builds in
environments such as `wasm32-wasi` or `wasm32-unknown-unknown`, these
facilities are gated behind a `#[cfg(feature = "tendermint")]` feature flag.

see #3913.

  > this does not make behavior changes to any of the code written, it is a
  > noöp, like-for-like refactoring. this generalizes logic for future reuse.


crates/proto/src/protobuf/tendermint_compat.rs
```diff
@@ -0,0 +1,710 @@
...
+        tendermint::abci::EventAttribute {
+            key,
+            value,
+            index: _,
+        }: tendermint::abci::EventAttribute,
+    ) -> Self {
+        Self {
+            key: key.into_bytes(),
+            value: value.into_bytes(),
+            // TODO(kate): this was set to false previously, but it should probably use the
```
zbuc, line 72: good catch, this will be important if we want to test indexing via mocks

zbuc (approved):
LGTM, much cleaner 🥇

PR #4449, opened 2024-05-23
Define staking events

### LINKED ISSUE #3797: staking: inventorize useful staking events to index

### LINKED ISSUE #4336: staking: emit ABCI events for validator state transitions

plaidfinch:
## Describe your changes

Resolves #4336, emitting events for validator definitions and state transitions. Included events:

- validator state change
- validator voting power change
- validator bonding state change
- validator rate data change
- validator definition upload
- validator missed block
- slashing penalty applied
- delegate
- undelegate
- funding stream rewards

All of these are proto-based events, and emitted both at genesis, when relevant, and when changed later.

Emitting events **at** genesis is technically impossible because the ABCI interface does not accept events as a response to `InitChain`, so instead genesis events are emitted at the start of the first block.

Resolves #4336. Closes #3797. Closes #[4512](https://github.com/penumbra-zone/penumbra/issues/4512).

  > This adds new events, but does not affect consensus.


proto/penumbra/penumbra/core/component/stake/v1/stake.proto
```diff
@@ -345,3 +347,55 @@ message EventTombstoneValidator {
...
+}
+
+message EventValidatorBondingStateChange {
+  // The validator's identity key.
+  keys.v1.IdentityKey identity_key = 1;
+  // The new bonding state of the validator.
+  BondingState bonding_state = 2;
+}
+
+message EventRateDataChange {
```
cratelyn, line 372: :bike: :house: :paintbrush: with apologies, a bicycle shed

```suggestion
message EventValidatorRateDataChange {
```

should this also have the `EventValidator`, since it also holds an identity key and relates to the broader lifecycle of a validator? :woman_shrugging: i don't feel strongly about it, but it feels grouped with the rest of these.

crates/core/component/stake/src/component/validator_handler/validator_manager.rs
```diff
@@ -70,16 +70,16 @@ use {
 ///  │   ┌─────────────┐        ┌────────────┐     │              │◀──┘ │
 ///  └──▶│   Jailed    │◀───────│   Active   │◀───▶│   Inactive   │     │
 ///      └─────────────┘        └────────────┘     │              │◀────┘
-///             │                                  └──────────────┘      
-///             │                                          ▲             
-///             └──────────────────────────────────────────┘             
-///
```
cratelyn, line 76: thanks for doing this sort of thing while you are in the neighborhood. scout's rule :fleur_de_lis:

cratelyn (commented):
this looks like it is coming together well! pardon the drive-by comments on a draft, i have one question to ask before these protobuf definitions ossify :angel:

crates/core/component/stake/src/component/action_handler/undelegate.rs
```diff
@@ -84,7 +85,7 @@ impl ActionHandler for Undelegate {
         tracing::debug!(?self, "queuing undelegation for next epoch");
         state.push_undelegation(self.clone());
 
-        state.record(event::undelegate(self));
+        state.record_proto(event::undelegate(self));
```
zbuc, line 88: should we roll this into `push_undelegation`?

zbuc:
This LGTM, I made a couple code organization changes and tested it and confirmed the deferred events work properly.

I have not tested it between upgrades however.

hdevalence:
Is anything blocking merging this?

zbuc:
> Is anything blocking merging this?

Only review

crates/core/app/src/app/mod.rs
```diff
@@ -151,6 +152,19 @@ impl App {
...
+        let events = state_tx.apply().1;
+
+        // Commit the genesis events to the state.
+        let mut state_tx = self
+            .state
+            .try_begin_transaction()
+            .expect("state Arc should not be referenced elsewhere");
+        for (i, event) in events.into_iter().enumerate() {
+            state_tx.put_deferred_event(i, event);
+        }
```
hdevalence, line 167: On reflection I have more concerns that this may be the wrong approach and I think we probably shouldn't do it.

hdevalence (changes requested):
Concerns about data model (will elaborate inline)

crates/core/app/src/app/mod.rs
zbuc: Is there another way we should handle it?

hdevalence: @zbuc sorry, i got distracted doing the full writeup and was delayed. filled in the comment now

zbuc: So the idea is that we remove this specially cased event emission and then let the indexers use genesis JSON to bootstrap initial state? Sounds fine to me.

zbuc: @hdevalence indexers would also need to handle the genesis data checkpointed during upgrades.

aubrika (approved):
LGTM pending requested changes

zbuc (approved):

hdevalence (approved):

PR #4454, opened 2024-05-24
dex: add duration metric for SimulateTrade RPC

hdevalence:
This removes the "swap" RPC, which was not useful (processing swaps is ~free)

  > metrics changes


hdevalence:
This is based against 0.75.x to make it easier to review and merge, it should go into main eventually

cratelyn (approved):
nice 📊

cratelyn:
> This is based against 0.75.x to make it easier to review and merge, it should go into main eventually

this sounds good to me. can i ask that upon merging, you file an issue to track this, so following up on that does not fall through the cracks?

PR #4468, opened 2024-05-24
view: 🔎 add tracing telemetry to view server

cratelyn:
#### 💭 describe your changes

while debugging my work integrating mock consensus tests with the view server,
it was helpful to add additional instrumentation to some relevant parts of the
`penumbra-view` crate. this branch adds some useful spans and events.

#### 🔖 issue ticket number and link

this branch contains preliminary changes related to #3913. see also, #4447.

#### ✅ checklist before requesting a review

- [x] if this code contains consensus-breaking changes, i have added the
  "consensus-breaking" label. otherwise, i declare my belief that there are not
  consensus-breaking changes, for the following reason:

  > this only makes changes to tracing events, and makes a small syntactic
  > change. no behavior changes here should affect consensus.

PR #4471, opened 2024-05-24
app: 📺 test that `Planner` can spend notes via mock consensus

### LINKED ISSUE #3913: Design a way to use the Rust view server with mock consensus integration tests

cratelyn:
#### 💭 describe your changes

fixes #3913.

this is the exciting culmination of work in pr's #3996, #4349, #4435,
 #4447, #4460, #4468, #4469, and #4470.

this adds a mock consensus test that spawns a gRPC server, creates a
view server, and then plans a transaction using the `Planner`. the
effects of this are then observed using the `ViewClient`.

**NB:** this is a large branch, so i have broken this into various preliminary PR's. this should land after:
* #4447
* #4468

these are related and included in this branch, but could land in parallel:
* #4469
* #4470

  > this only changes test code.
  
#### :link: related work

* #3857
* #3996
* #4349
* #4435
* #4447
* #4460
* #4468
* #4469
* #4470


crates/core/app/tests/view_server_can_be_served_on_localhost.rs
```diff
@@ -0,0 +1,205 @@
...
+
+    // Jump ahead a few blocks.
+    test_node
+        .fast_forward(10)
+        .tap(|_| tracing::debug!("fast forwarding past genesis"))
+        .await?;
+
+    let grpc_url = "http://127.0.0.1:8080"
+        .parse::<url::Url>()?
+        .tap(|url| tracing::debug!(%url, "parsed grpc url"));
```
hdevalence, line 67: Can we avoid using a network socket here, and use an entirely in-memory GRPC service (the way `pcli` does a `BoxGrpcService`, for instance), so that we can run all of these tests concurrently without contending for the same global environment state?

cratelyn: i would very much like to do this too. i have pulled on this thread and found that it's going to be a slightly involved change. the assumption that the rpc service lives across a network connection is baked into a few different places in the `ViewServer` and its underlying `Worker`.

the notion of a remote url is threaded around in places like `record_block`, here: https://github.com/penumbra-zone/penumbra/blob/main/crates/view/src/storage.rs#L1366-L1373
..and indirectly via `tonic::transport::Channel` in a few other places.

i'll take a swing at this, and report back.

cratelyn: this is doable, and i have most of that change assembled now.

there is a final hurdle, `Sync` bounds, that i'll handle next. for now, the relevant incremental commits are:

* 4e43181ef - view: use type-erased BoxGrpcService (does not build)
* 9fea1cf12 - view: `ViewServer` holds a `Channel`
* 44eb082c0 - view: hoist url out of worker
* 4cb63624b - view: avoid passing url to `record_block()`

cratelyn: @hdevalence this has turned into a bit of a quagmire. `BoxGrpcService` isn't `Sync`, but `ViewService` has a `Sync` bound. i've tried various approaches, but am finding this to be an invasive change.

particularly with some other items on my docket related to the testnet faucet, i think a prudent course of action for us is to:
* revert 4e43181, and punt on using an in-memory service **for now.** the changes in 4cb6362, 44eb082, and 9fea1cf all are good stepping stones we can keep.
* file a ticket to track view server support for an in-memory grpc service in the view server.
* pivot to other high-priority mainnet tickets, and revisit the in-memory grpc service after that work winds down.

what do you think?

cratelyn: oh, and to disambiguate: this test _does_ use an in-memory **view** service for the view client, but we do not have a way for the view service to talk to an in-memory rpc endpoint (_is there a better word for the broader rpc router returned by `penumbra_app::rpc::router()` served by `pd`?_).

cratelyn: @hdevalence and i talked a bit directly and agreed that in-memory support for the `ViewServer`'s rpc endpoint can be punted on for now. i've dropped the two tail commits off this branch, and it should be ready for a review now.

i will mark this as resolved, now that i've filed #4517 to track this

cratelyn:
rebased on the latest main :recycle:

cratelyn:
...and ci passed! this should be ready for another look, when you have time @hdevalence :slightly_smiling_face:

hdevalence (approved):

PR #4472, opened 2024-05-24
Add ID filtering to AuctionsRequest; add local_seq to AuctionsResponse

jessepinho:
## Describe your changes

This PR makes two changes:
1. Add an `auction_ids_filter` to `AuctionsRequest`, so that clients can make requests (e.g., with `queryLatestState`) for just specific auctions rather than all auctions at once.
2. Add a `local_seq` property to `AuctionsResponse` to indicate the local view service's own knowledge of the sequence number for an auction. (See https://github.com/penumbra-zone/penumbra/pull/4385#issuecomment-2111256876.)

Relevant changes are to the protobuf [here](https://github.com/penumbra-zone/penumbra/pull/4472/files#diff-03b7341d5bf81ab9c8d8542d220d5ba4ae122bd3837531309742b685d7ec1619), plus a couple TODO items [here](https://github.com/penumbra-zone/penumbra/pull/4472/files#diff-9f4ca4cefeac93c9275a2f9d962d6670f6ab1c1381fd3d7c806f9316e3779ccbR986) and [here](https://github.com/penumbra-zone/penumbra/pull/4472/files#diff-36188e6ab5083e8be9d29039370cc523cfcfede683b223327b2d90e3f75d40c7R461).

https://github.com/penumbra-zone/web/issues/1059 and https://github.com/penumbra-zone/penumbra/pull/4385#issuecomment-2111256876

  > Only touches Auctions RPC methods.


proto/penumbra/penumbra/view/v1/view.proto
```diff
@@ -170,6 +174,16 @@ message AuctionsResponse {
...
+  // The sequence number of the auction state _as known to the local view
+  // service_. Note that the local view service may lag behind the fullnode. For
+  // example, if the chain hits an auction's `end_height`, but the user hasn't
+  // yet exchanged their sequence-0 (opened) auction NFT for a sequence-1
+  // (closed) auction NFT, the local view service will have a sequnce number of
+  // 0.
+  //
+  // Dutch auctions move from:
+  // 0 (opened) => 1 (closed) => n (withdrawn)
+  uint64 local_seq = 5;
```
jessepinho, line 186: `note_record` above uses `4`, which is why this field uses `5`

erwanor: oh strange, could you move it so that they're ordered 🙏

jessepinho: done

hdevalence (approved):
LGTM

TalDerei:
did async review post merge and LGTM!

PR #4473, opened 2024-05-24
Implement sliding window algorithm for FMD params

### LINKED ISSUE #1087: FMD: initial `Clue` parameter selection
Write some code to, based on current network conditions - i.e. rate of transactions per epoch - provide a reasonable choice of `Clue` parameters, i.e. as a user if we want a given FPR we should set clue parameters to X

See related comment here: https://github.com/penumbra-zone/penumbra/pull/1312#discussion_r954544436:

> But if transactions with the wrong FMD precision are rejected, the requirement that the precision be defined pre-signing is effectively placing a bound on how long a signer can take to authorize a transaction, so we should consider that timing when deciding how frequently to adjust the precision.

Sub items:
- [x] Provide functionality in the node side to update the FMD parameters
- [ ] Implement functionality to add a sliding window measurement of network activity, where network activity is in terms of number of outputs. It should take a number of blocks (maybe an epoch's worth to start?)
- [x] Describe the attacks we consider
- [ ] Come up with some reasonable initial choice that scales the precision based on network activity

cronokirby:
## Describe your changes

This adds a new algorithm for FMD parameters, which uses a sliding window approximation to target a specified number of detections over a given window. The idea is that if you set a window, of say, 1 day (in blocks), then the algorithm will then approximate the number of clues detected in the past day, and then use that to set the precision, so that the lowest precision such that at least the target number of detections will be generated, if that number of clues is seen over the upcoming window.

We could use a window of 1 day and 1000 detections or so as reasonable production parameters.

Closes #1087.

  > This adds new possibilities to the chain parameters in a backwards-compatible way, but which will cause a consensus divergence if the new parameters are sent via governance etc. without an upgrade


crates/core/component/shielded-pool/src/fmd.rs
```diff
@@ -59,14 +59,86 @@ impl Default for Parameters {
...
+        if self.window_blocks == 0 {
+            return (
+                old.clone(),
+                MetaParametersAlgorithmState::SlidingWindow {
+                    approximate_clue_count: 0,
+                },
+            );
+        }
+
+        let new_clues_in_block = clue_count_delta.1.saturating_sub(clue_count_delta.1);
```
erwanor, line 87: I think this needs to be amended, because this will always be equal to zero

erwanor: also, candidly, i'm working out what index 0 and index 1 correspond to, a short comment would be helpful there

erwanor: maybe destructuring the tuple in the method signature? `(old_clue_count, new_clue_count): (u64, u64)`

redshiftzero (commented):
Looks good! One suggested change inline that we should make prior to merge

cronokirby:
I fixed the comments above, and then added a fix in terms of proto naming to correct the fact that the window should be measured in terms of *update periods*, and not blocks. e.g. a window of "4" means you consider the previous 4 update checks to the parameters, not the previous 4 blocks. (with the default update frequency of 16, this would be 64 blocks)

redshiftzero (approved):
LGTM! my comments were addressed

Issue #4475, opened 2024-05-24
dex(rpc): implement candlestick streaming


erwanor:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

erwanor:
Sorry misclicked before writing the description, editing

PR #4476, opened 2024-05-24
dex: implement streaming candlesticks like compact blocks

hdevalence:
## Describe your changes

Fill in the streaming notification RPC for candlestick data.

This also rearranges some of the existing code to split out the component parts from the data modeling parts.

In the future, we could add more metrics or consider writing something generic for this pattern of "subscribe and pull out of state", though it might not make sense until we have a third API since this one is already different from the compact block RPC (where there's a catch up phase). But that could be a good way to do things like per-RPC connection counters. Or perhaps that should better not be done "inside" each streaming RPC but as a middleware that could be on any streaming RPC.

#4475

  > no changes to consensus logic, only mechanical refactoring and RPC cahnges


erwanor (approved):
LGTM, tested it locally.

PR #4481, opened 2024-05-25
circuit: swap balance randomization

TalDerei:
## Describe your changes
The swap balance commitment is derived using a new blinding factor.

References #4482

  > changes proving keys


proto/penumbra/penumbra/core/component/dex/v1/dex.proto
```diff
@@ -68,8 +68,10 @@ message SwapBody {
   // in the future we'll want separate commitments to each delta_j_i
   // anyways in order to prove consistency with flow encryption.
   asset.v1.BalanceCommitment fee_commitment = 4;
+  // A commitment to the value balance.
+  asset.v1.BalanceCommitment balance_commitment = 5;
```
erwanor, line 72: We can't change protobuf like that unfortunately, this is a very breaking change.

Here are a couple useful links about protobuf versioning:
- the most important one: [protos dos and donts](https://protobuf.dev/programming-guides/dos-donts/)
- https://github.com/uber/prototool/blob/dev/style/README.md#package-versioning
- https://buf.build/docs/breaking/rules

crates/core/component/dex/src/swap/proof.rs
```diff
@@ -133,8 +137,8 @@ impl ConstraintSynthesizer<Fq> for SwapCircuit {
         let balance_2 = BalanceVar::from_negative_value_var(swap_plaintext_var.delta_2_value());
         let balance_1_commit = balance_1.commit(transparent_blinding_var.clone())?;
         let balance_2_commit = balance_2.commit(transparent_blinding_var)?;
-        let transparent_balance_commitment = balance_1_commit + balance_2_commit;
-        let total_balance_commitment = transparent_balance_commitment + fee_commitment;
+        let opaque_commitment = fee_balance.commit(balance_blinding_var)?;
+        let total_balance_commitment = balance_1_commit + balance_2_commit + opaque_commitment;
```
erwanor, line 141: I'm not totally sure this solve the shared randomizer problem, and as I understand it the ECC recommendation is to 1/ use a balance-level blinding factor $\tilde v_i$ and to build a completely new generator $G_{\tilde v_i}$ *in addition* to $cv_f$ so that you cannot easily isolate either of the swap inputs e.g. $cv - cv_f = [-v_1]G_1 + [\tilde v_i]G_{\tilde v_i}$. But AFAICT, what we're doing here is to sample a random scalar $\tilde v_f$ and then never use it? So there's still a single randomizer for the entire balance commitment! Maybe I'm off but I would like to understand why!

proto/penumbra/penumbra/core/component/dex/v1/dex.proto
erwanor: The right way to do this is to add a new field, without changing the numbering so that it doesn't affect the binary encoding of the messages.

erwanor (changes requested):
Requesting change because this is protobuf breaking but doesn't need to be. I also have a question about whether this is the actual recommended remediation

crates/core/component/dex/src/swap/proof.rs
erwanor: @TalDerei pointed out to me offline that the whole point of this approach is to yeet $cv_f$ out of the total balance commitment

TalDerei:
currently stepping out, but just want to point out this is still marked as a draft and I have a local change set with some fixes that I'll push later.

crates/core/component/dex/src/swap/proof.rs
```diff
@@ -133,8 +137,8 @@ impl ConstraintSynthesizer<Fq> for SwapCircuit {
         let balance_2 = BalanceVar::from_negative_value_var(swap_plaintext_var.delta_2_value());
         let balance_1_commit = balance_1.commit(transparent_blinding_var.clone())?;
         let balance_2_commit = balance_2.commit(transparent_blinding_var)?;
-        let transparent_balance_commitment = balance_1_commit + balance_2_commit;
-        let total_balance_commitment = transparent_balance_commitment + fee_commitment;
+        let opaque_commitment = fee_balance.commit(balance_blinding_var)?;
```
TalDerei, line 140: the imbalance from `fee_balance` has already been captured and iterator consumed on constructing the `fee_commitment`.  calling `fee_balance.commit(balance_blinding_var)` with a new blinding factor $\widetilde{v_i}$ should only perform the blinding operation inside `commit`. 

$cv = [-v_1] G_1 + [-v_2] G_2 +[\widetilde{v_i}]G_{\widetilde{v_i}}$, where $\widetilde{v_i}$ is a private randomizer to make the commitment to the value balance hiding.

TalDerei:
balance contributions in https://github.com/penumbra-zone/penumbra/blob/swap-randomization/docs/protocol/src/transactions/actions.md needs to be modified accordingly

TalDerei:
currently the test suite is passing, but `swap_claim_flow` in the smoke tests are failing for some reason –– haven't investigated.

TalDerei:
@redshiftzero pointed out offline that that in the commitment scheme, there may have been a requirement for computational binding that the generators be independently chosen. In sticking with the recommended remediation, we'll define a new constant generator $G_{v_i}$.

TalDerei:
closing PR as scope has shifted to `V2`.

Issue #4483, opened 2024-05-26
SwapClaim circuit specification is incomplete


TalDerei:
The `SwapClaim` circuit implementation cannot be audited against its specification, because its specification is incomplete. Specifically, the specification needs to provide more detail on the fee consistency and output amounts integrity checks.

PR #4484, opened 2024-05-26
specification: clarify swap claim integrity checks

TalDerei:
## Describe your changes
Specifies components of the integrity checks in the swap claim.

References #4483

### DIFF
--- a/docs/protocol/src/dex/action/swap_claim.md
+++ b/docs/protocol/src/dex/action/swap_claim.md
@@ -148,7 +148,7 @@ The zk-SNARK also certifies that:
 
 ### [Fee Consistency Check](#fee-consistency-check)
 
-The zk-SNARK certifies that the public claim fee is equal to the value
+The zk-SNARK certifies that the public claim fee $v_f$ is equal to the value
 witnessed as part of the swap plaintext.
 
 ### [Height Consistency Check](#height-consistency-check)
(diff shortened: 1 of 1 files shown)


TalDerei:
there are a things to note here:

1. The swap claim specification references that the body of a SwapClaim has a "balance commitment, which commits to the value balance of the spent note". This is inconsistent with the [protobuf definition](https://buf.build/penumbra-zone/penumbra/docs/78be1d64b1cb484ba4bc666d54dc76c5:penumbra.core.component.dex.v1#penumbra.core.component.dex.v1.SwapClaimBody) of a `SwapClaimBody`. The swap claim circuit also doesn't perform any value balance commitment related checks. 

2. Regarding the fee consistency check: in constructing the `swapClaim` plan, it uses the `claimFee` from the `SwapPlaintext` as the `claimFee` field inside `SwapClaimProofPublic`. Then, in the fee consistency check during circuit constraint generation, it uses the same `SwapPlaintext` inside `SwapClaimProofPrivate`. It’s comparing the _same_ swap claim fee, meaning the fee consistency check isn’t actually doing anything useful here?

redshiftzero:
> Regarding the fee consistency check: in constructing the swapClaim plan, it uses the claimFee from the SwapPlaintext as the claimFee field inside SwapClaimProofPublic. Then, in the fee consistency check during circuit constraint generation, it uses the same SwapPlaintext inside SwapClaimProofPrivate. It’s comparing the swap claim fee from the same source, meaning the fee consistency check isn’t actually doing anything useful here from what I can tell?

This is true for an honest prover, but a malicious prover can construct a `SwapClaim` with an arbitrary public claim fee, however the ZKP demonstrates that it must match the claim fee from the `SwapPlaintext`

redshiftzero (approved):

TalDerei:
> This is true for an honest prover, but a malicious prover can construct a SwapClaim with an arbitrary public claim fee, however the ZKP demonstrates that it must match the claim fee from the SwapPlaintext

Good point, I should've been strictly thinking about this from the vantage point of a malicious prover.

PR #4485, opened 2024-05-26
specification: file organization in protocol spec

TalDerei:
## Describe your changes
Several parts of the protocol are not specified, while being referred to in other parts of the documentation. We should remove these empty patches in our protocol spec, and use this PR as a guiding reference of what sections were removed as we periodically fill them in over time. @conorsch can you help better format the docs?

This references components **X** and **Y** in the ECC audit log. Auxiliary to this is component **Z**, which should probably be (but currently is **not**) captured here since they're also spec-related changes.  

This also captures supplemental spec changes based on specific spec-related comments made in the audit. 


docs/protocol/src/shielded_pool/note_plaintexts.md
```diff
@@ -5,6 +5,6 @@ Plaintext notes contain:
 * the value to be transmitted which consists of an integer amount $v$ along with a scalar (32 bytes) $ID$ identifying the asset.
 * $rseed$, a 32-byte random value, which will later be used to derive the note blinding factor used for the
 note commitment and an ephemeral secret key.
-* the destination address, described in more detail in the [Addresses](../addresses_keys/addresses.md) section.
+* the destination address controlling the note, described in more detail in the [Addresses](../addresses_keys/addresses.md) section.
```
TalDerei, line 8: In the note plaintext [section](https://hackmd.io/v0qMQRKqSz-OK0G4Ml383Q#Note-plaintexts) of the audit, there was some confusion about `pk_d`. Comparing the [spec](https://protocol.penumbra.zone/main/shielded_pool/note_plaintexts.html) against the code reveals some incongruences:

- the `Address` struct defines `pk_d` as the public key used for the payment address, but the spec defines `pk_d` 
as the diversified transmission key. 
- the `Note` struct consists of value, rseed, address, and transmission_key but the [spec](https://protocol.penumbra.zone/main/shielded_pool/note_plaintexts.html) is missing the diversified transmission key.

docs/protocol/src/shielded_pool/note_ciphertexts.md
```diff
@@ -1,3 +1,17 @@
 # Note Ciphertexts
```
TalDerei, line 1: Note encryption was previously unspecified.

docs/protocol/src/addresses_keys/addresses.md
```diff
@@ -8,6 +8,14 @@ attaches a *detection key* to each address, allowing a user to outsource
 probabilistic transaction detection to a relatively untrusted third-party
 scanning service.
 
+## Privacy Implications
```
TalDerei, line 11: Added privacy implications associated with detection keys based on comments made in component **Y**.

docs/protocol/src/shielded_pool/note_commitments.md
```diff
@@ -14,7 +14,7 @@ The note commitment is generated using rate-5 Poseidon hashing with domain separ
 The note commitment is then constructed using the above domain separator and
 hashing together the above contents along with the note blinding factor $rcm$:
 
-`note_commitment = hash_5(ds, (rcm, v, ID, B_d, pk_d))`
+`note_commitment = hash_6(ds, (rcm, v, ID, B_d, pk_d, ck_d))`
```
TalDerei, line 17: https://github.com/penumbra-zone/penumbra/blob/main/crates/core/component/shielded-pool/src/note.rs#L355

docs/protocol/src/SUMMARY.md
```diff
@@ -17,7 +17,6 @@
     - [Group Hash](./crypto/decaf377/group_hash.md)
     - [Test Vectors](./crypto/decaf377/test_vectors.md)
   - [Randomizable Signatures](./crypto/decaf377-rdsa.md)
-  - [Key Agreement](./crypto/decaf377-ka.md)
```
TalDerei, line 20: we should actually keep this section and link to the [implementation](https://github.com/penumbra-zone/decaf377-ka)

TalDerei:
@redshiftzero this doesn't have an associated issue attached to it

TalDerei:
this is a staging area for spec-related changes, and we'll focus on addressing the rest of the spec changes later this week once phase 2 is unblocked.

docs/protocol/src/community_pool/action/deposit.md
```diff
@@ -1 +0,0 @@
-# CommunityPoolDeposit
```
redshiftzero, line 1: The audit (correctly) pointed out that we lack complete documentation of the protocol, and have these stub pages in the protocol spec. We _should_ remove pages that are no longer relevant, but I think in every case the stub pages are relevant, i.e. the `CommunityPoolDeposit` does exist as an action, and as such should be ideally documented in the protocol specification. So I think the action here is to fill these pages out (not in this PR because that's a significant chunk of work) instead of deleting them. One thing we could do is just make a list of pages that should be filled out in an issue.

TalDerei: sounds great, will revert the deleted pages.

TalDerei: references https://github.com/penumbra-zone/penumbra/issues/4516

redshiftzero (approved):
LGTM!

PR #4489, opened 2024-05-28
dex: 🪣 add two larger dex buckets

### LINKED ISSUE #4464: dex: 🎩 investigate search path duration of 100ms
**Describe the bug**

image

@hdevalence noticed that there is a conspicuous ceiling of 100ms in the dex component's search path duration metrics, reported in Grafana.

we should investigate this further.

cratelyn:
fixes #4464.

this adds two larger buckets to the dex component's histograms.

when our dashboards calculate quantiles, we observed signals that some operations were taking longer than 100ms. to help obtain more accurate performance data, we add a 1 second and 10 second bucket.

#### :heavy_check_mark: checklist before requesting a review

  > this only changes telemetry


cratelyn:
#4502 adjusts the buckets to a logarithmically spaced group of buckets. closing this.

PR #4491, opened 2024-05-28
metrics: add dex dashboard

conorsch:
## Describe your changes
 Adds a new hand-written dashboard to display DEX event durations. We're still massaging the bucketing logic, but getting this dashboard into version-control immediately, so we can iterate on it.

Refs #4464.

  > no app code changes, metrics display only

### DIFF
--- a/deployments/config/grafana/dashboards/DEX.json
+++ b/deployments/config/grafana/dashboards/DEX.json
@@ -1,0 +1,671 @@
+{
+  "annotations": {
+    "list": [
+      {
+        "builtIn": 1,
+        "datasource": {
+          "type": "grafana",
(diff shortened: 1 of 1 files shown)


cratelyn (approved):
ah-ha! that's where these live. thanks @conorsch!

conorsch:
The testnet deployment pins a tag, but I manually modified it to use a container built from this branch, so the new dashboard is live immediately. In strict adherence to process, we should cherry-pick this onto a release-branch to make sure it's included in the next point release, but I'm not going to do that now: instead, I'm going to focus on planning the next release, be that a point release or another chain upgrade in the near future.

PR #4495, opened 2024-05-29
chore: 📌 pin `rocksdb` crate dependency specifically

cratelyn:
#### 💭 describe your changes

RocksDB is used for the consensus-critical storage layer in cnidarium.

the `rocksdb` crate uses a dependency called `librocksdb-sys`, containing FFI
bindings to RocksDB's headers. that crate internally clones a git submodule of
the RocksDB repository, builds the project, and statically links against it.

to avoid regular maintenance upgrading crates in the future from introducing
unexpected consensus-breaking changes, explicitly pin this crate to a specific
major, minor, and patch version.

this will also help clear the way for operators and engineers maintaining the
protocol to improve build times by building RocksDB out-of-band, and then
statically linking against that local copy of `librocksdb.a`.

#### 🔍 checking my work:

to confirm how i found this commit hash, for posterity:

from the `Cargo.lock`:

```toml
[[package]]
name = "librocksdb-sys"
version = "0.11.0+8.1.1"
```

..and confirming the rocksdb commit manually:

```sh
; pwd
/path/to/rust-rocksdb
; git show --oneline --quiet
b539412 (HEAD, tag: v0.21.0) Release 0.21.0 (#777)
; git submodule status
6a436150417120a3f9732d65a2a5c2b8d19b60fc librocksdb-sys/rocksdb (v8.1.1)
2b63814b15a2aaae54b7943f0cd935892fae628f librocksdb-sys/snappy (1.1.9)
```

> this does not change the version of `rocksdb` being used (note the lack of
> any changes to the Cargo.lock); this only changes the semantic version
> constraints applied to that package, adding some comments flagging relevant
> commit hashes & version numbers.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -204,7 +204,6 @@ rand                             = { version = "0.8.5" }
 rand_chacha                      = { version = "0.3.1" }
 rand_core                        = { version = "0.6.4" }
 regex                            = { version = "1.8.1" }
-rocksdb                          = { version = "0.21.0" }
 serde                            = { version = "1.0.186" }
 serde_json                       = { version = "1.0.96" }
 serde_unit_struct                = { version = "0.1" }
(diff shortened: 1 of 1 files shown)


cratelyn:
https://github.com/penumbra-zone/penumbra/commit/b228e3716dea08bd96f7e3afa2170f2bf4f941bf#commitcomment-142517732

:link: linking for visibility, i owe a response to this!

PR #4496, opened 2024-05-29
chore: 🪨 improve incremental builds by linking to rocksdb

cratelyn:
first, the cool part: here are incremental build timings for `cargo build -p cnidarium` filtering for compilation units that took > 1.5 seconds to compile.

on my machine, this improved incremental build times by 72.92% :heart: 

**before** :weary: 

warm-build-timing-without-rocksdb-linking

**after** :heart_eyes: 

warm-build-timing-with-rocksdb-linking

#### 💭 background

`librocksdb-sys(build)` is a somewhat infamous source of lengthy build
times when engineers are working in the monorepo. this is a crate whose
build script, clones and builds the rocksdb database from source,
statically linking against the generated `librocksdb.a` file.

this interacts poorly with a Cargo workspace like ours, which contains a
number of different features. because features are additive, running
commands like `cargo build --workspace` and `cargo build --package penumbra-app`
can result in different feature sets being enabled for leaf
dependencies, mandating that the library be rebuilt (_which means
rebuilding librocksdb.a once more_).

### flake: 🔗 link against a precompiled rocksdb

this adds a `ROCKSDB_LIB_DIR` environment variable to the nix
development shell.

this environment variable points to the directory that
contains the `librocksdb.a` file, to be statically linked against the
`librocksdb-sys` crate.

a `nixpkgs.rocksdb` build input is added.

#### 📐 build timing methodology

this shell script was used to generate build timings. in order to see
how this behaved on an incremental build, the script will clear the
cache, build the workspace, and _then_ build a crate that depends on
`librocksdb-sys`.

```sh
 #!/usr/bin/env bash
 #
 # see https://doc.rust-lang.org/cargo/reference/timings.html for more info.
 set -euox pipefail

 # clean out any preëxisting build artifacts.
 cargo clean;

 # build once before we generate our report.
 cargo build --workspace

 # build a particular crate, recording the build timings.
 cargo build --package cnidarium --timings

 # hang on to the report generated.
 cp target/cargo-timings/cargo-timing.html .
```

### DIFF
--- a/docs/guide/src/dev/build.md
+++ b/docs/guide/src/dev/build.md
@@ -80,7 +80,7 @@ Then, build all the project binaries using `cargo`:
 cargo build --release
 ```
 
-### Linking Against RocksDB
+### Linking Against RocksDB (Optional)
 
 Development builds can avoid the cost of recompiling RocksDB for storage libraries in the Cargo
 workspace. This manifests as a `librocksdb-sys(build)` message when building or testing crates
--- a/docs/guide/src/dev/build.md
+++ b/docs/guide/src/dev/build.md
@@ -80,5 +80,58 @@ Then, build all the project binaries using `cargo`:
 cargo build --release
 ```
 
+### Linking Against RocksDB
+
+Development builds can avoid the cost of recompiling RocksDB for storage libraries in the Cargo
+workspace. This manifests as a `librocksdb-sys(build)` message when building or testing crates
+in the monorepo.
+
+#### Building `librocksdb.a` from source
+
+First, clone the rocksdb repository:
+
+```sh
+# Clone the repository, and enter that directory.
+git clone git@github.com:facebook/rocksdb.git && cd rocksdb
+
+# Checkout the version of rocksdb used in `librocksdb-sys`.
+git checkout 6a43615
+
+# Add an environment variable pointing to this repository:
+ROCKSDB_LIB_DIR=`pwd`
+
+# Compile the static `librocksdb.a` library to link against:
+make static_lib
+```
+
+#### Building `libsnappy.a` from source
+
+next, clone the snappy repository and follow the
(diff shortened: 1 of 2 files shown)


flake.nix
cratelyn, line 1: cc: @plaidfinch, a validator using nix.

i suspect that some additional effort would be needed to fetch the _same_ version of RocksDB used implicitly by `librocksdb-sys`, before this is truly production-ready, but this greatly improved incremental builds on my machine.

hopefully, this is a good starting point for your own infrastructure. :slightly_smiling_face:

cratelyn: and cross-referencing #4495, which codifies that version explicitly.

docs/guide/src/dev/build.md
cratelyn, line 1: i'd like to note that these docs are best effort, as a nix user.

i found that the version of rocksdb we use has aged enough to have run into problems when compiled with a recent version of clang/gcc, but this is the broad shape of how someone using a distro like Ubuntu would do this.

i'm eager to land the _flake_ changes, below, so if these are causing problems i'm happy to drop them from this changeset and help land these docs separately.

conorsch (approved):
LGTM, although I am not using the nix setup myself. As a general rule, I think we should be careful to document abstruse setup steps like these as optional. Down the road, it would be grand to have a reusable nix-ish environment that's usable both on workstations and in CI. Thanks for driving toward that!

cratelyn:
> LGTM, although I am not using the nix setup myself. As a general rule, I think we should be careful to document abstruse setup steps like these as optional. Down the road, it would be grand to have a reusable nix-ish environment that's usable both on workstations and in CI. Thanks for driving toward that!

192663f adds `(Optional)` to be a little more explicit about that. thanks for taking a look!

PR #4498, opened 2024-05-29
crypto: used fixed nonce for swap symmetric encryption

redshiftzero:
## Describe your changes

The ECC audit pointed out that the use of a swap commitment as a nonce is unnecessary, and we can safely used a fixed nonce instead. 

  > Clients perform swap encryption/decryption, so there are no chain breaking changes. On the client side, this PR modifies the decryption path to first attempt to decrypt using the fixed-nonce introduced in this PR, and then try the legacy method. This way, clients can still decrypt their old swaps after upgrade, and thus claim the swap outputs.

### DIFF
--- a/crates/core/component/dex/src/swap/ciphertext.rs
+++ b/crates/core/component/dex/src/swap/ciphertext.rs
@@ -15,17 +15,13 @@ impl SwapCiphertext {
         commitment: note::StateCommitment,
     ) -> Result<SwapPlaintext> {
         let payload_key = PayloadKey::derive_swap(ovk, commitment);
-        self.decrypt_with_payload_key(&payload_key, commitment)
+        self.decrypt_with_payload_key(&payload_key)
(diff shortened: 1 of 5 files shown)


hdevalence:
If we make this change, do we lose the ability to decrypt old swap plaintexts?

If so, I think that's fine, but we should just be aware of what we're breaking.

redshiftzero:
If shipped without an upgrade, yes we do.

redshiftzero:
Per an offline discussion, @hdevalence pointed out that we can trial decrypt the swap ciphertexts such that after upgrade, users can still decrypt their old swap ciphertexts and claim the swap outputs. On the client side, this PR modifies the decryption path to first attempt to decrypt using the fixed-nonce introduced in this PR, and then try the legacy method.

cronokirby (approved):

Issue #4494, opened 2024-05-28
pd: add support for running in paused 'archive node' mode


avahowell:
**Is your feature request related to a problem? Please describe.**
In the CometBFT ecosystem, an 'archive node' refers to a node that is running an from an old (pre-upgrade) genesis, used to serve Comet RPC requests for data during that epoch. Post upgrade, any state from a pre-upgrade epoch is not queryable by the CometBFT RPC. This presents a problem for use cases such as relayers, which use the Comet RPC in order to fetch headers to use to build client updates. As such, it is typical for CometBFT operators to run a pre-upgrade node in a 'frozen' state, where that node only exists to serve RPC requests for headers from the pre-upgrade epoch, known as an 'archive node'.

There is not currently a natively supported pathway in `pd` for running in 'archive' mode (not producing any blocks, but still remaining online so as not to halt CometBFT). 

**Describe the solution you'd like**
We should add a flag to `pd start`, `pd start --archive`, that configures `pd` so that it expects to run in a frozen 'archive' state.

`--archive` should:

- Ignore/override any upgrade parameters indicating that the node should halt due to an upgrade.
- Instruct `pd` to not respond to any consensus-related ABCI requests. This should be done in a way that avoids `pd` or CometBFT halting.
- Instruct `pd` to not respond to any mempool-related ABCI requests. This should be done in a way that avoids `pd` or CometBFT halting.

conorsch:
In support of the upcoming #4497, want a much slimmer implementation: 

* `pd migrate ready-to-start` which flips the halt bit off, permitting the node to start despite the halt bit

We already have migration code that does this: https://github.com/penumbra-zone/penumbra/blob/319a3d925e201be3d061536fc68951df714e1860/crates/bin/pd/src/migrate/reset_halt_bit.rs

conorsch:
Now that we have #4499, I think we can deprioritize this work to post-v1. Separately I'm working on reindexing an event db from historical backups. As long as that works end-to-end, I believe we can get by with the functionality we have today.

PR #4499, opened 2024-05-29
feat(pd): add --ready-to-start flag to pd migrate

conorsch:
## Describe your changes

Adds a new flag to `pd migrate --ready-to-start` that makes an in-place edit to local state, setting an enabled halt bit to false.

Refs #4494.

  > This code directly munges local state, via a migration, and therefore deserves careful scrutiny.

### DIFF
--- a/crates/bin/pd/src/main.rs
+++ b/crates/bin/pd/src/main.rs
@@ -12,7 +12,7 @@ use cnidarium::Storage;
 use metrics_exporter_prometheus::PrometheusBuilder;
 use pd::{
     cli::{Opt, RootCommand, TestnetCommand},
-    migrate::Migration::Testnet76,
+    migrate::Migration::{ReadyToStart, Testnet76},
     testnet::{
         config::{get_testnet_dir, parse_tm_address, url_has_necessary_parts},
         generate::TestnetConfig,
@@ -431,6 +431,7 @@ async fn main() -> anyhow::Result<()> {
             home,
             comet_home,
             force,
+            ready_to_start,
         } => {
             let (pd_home, comet_home) = match home {
                 Some(h) => (h, comet_home),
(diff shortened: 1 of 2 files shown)


conorsch:
Tested this on a local devnet, and it appears to work well: if the adjacent cometbft instance also sets `--p2p.pex false`, then the node remains at the pre-upgrade height, and I can query the CometBFT RPC without problems.

crates/bin/pd/src/main.rs
```diff
@@ -441,11 +442,22 @@ async fn main() -> anyhow::Result<()> {
...
+                    .instrument(pd_migrate_span)
+                    .await
+                    .context("failed to disable halt bit in local state")?;
+                exit(0)
+            }
+
+            let genesis_start = pd::migrate::last_block_timestamp(pd_home.clone()).await?;
+            tracing::info!(?genesis_start, "last block timestamp");
             Testnet76
                 .migrate(pd_home.clone(), comet_home, Some(genesis_start), force)
```
avahowell, line 462: noting that this command *also* applies the testnet76 migration instead of just flipping the ready to start bit. Would it be better to have a separate command that only flips the bit?

conorsch: No, we definitely want to avoid that exact circumstance. As written, if `--ready-to-start` is supplied, then _only_ the halt-bit is modified, and the normal migration is _not_ run. That's because it bails out early via `exit(0)` on L456.

avahowell (approved):

erwanor (approved):

crates/bin/pd/src/main.rs
erwanor: I think we can make the case for either, `pd migrate run` vs. `pd migrate ready` or something. Either way, this PR does the job so that we can do the genesis restart with the new halt bit so lgtm

cratelyn (approved):

PR #4500, opened 2024-05-29
crypto: add extra overflow check to ivk derivation

cronokirby:

  > changes proving keys


redshiftzero (commented):
this fix LGTM! good to add in the proving key updates and then merge when CI is green

redshiftzero (approved):

TalDerei:
if this did overflow and we didn't check it, could it create an invalid proof that would still be accepted by a verifier? I guess what are the broader consequences of not performing this constraint check?

cronokirby:
In and of itself it isn't harmful but it's better to nip it in the bud. The direct issue is that there are two valid ivk values you can provide, from the perspective of the circuit. This lets someone claim to derive two ivks from the same fvk. As far as I can tell this doesn't lead to something like double spend, but this shouldn't be a possibility.

Issue #4497, opened 2024-05-29
Release Testnet 77, via chain upgrade


conorsch:
# Testnet upgrade

Testnet chain id: `penumbra-testnet-deimos-8` (unchanged)
Release date: 2024-05-31
Testnet release manager: @conorsch

# Testnet Release Manager Checklist

Pre-release:

- [x] Review all merged PRs with tags [`consensus-breaking` and/or `state-breaking`](https://github.com/penumbra-zone/penumbra/pulls?q=is%3Apr+label%3Astate-breaking%2Cconsensus-breaking+created%3A%3E%3D2024-01-01)
- [x] Enumerate the specific PRs that should be included in the release:
  - [x] https://github.com/penumbra-zone/penumbra/pull/4499
  - [x] ECC audit remediation - https://github.com/penumbra-zone/penumbra/pull/4498
  - [x] ECC audit remediation - https://github.com/penumbra-zone/penumbra/pull/4500
- [x] Confirm migrations exist where required for breaking changes already merged
- [ ] Identify specific test cases for before/after behavior, querying chain state, to ensure that migrations are effective
- [x] Prepare `upgrade-plan` governance proposal
  - [x] Decide on block height at which to halt
  - [x] Double-check you've got bonded stake
  - [x] Submit proposal
  - [x] Vote for it, notify validators of same
  - [x] Confirm proposal passed (default is 4h voting period)

On release day:

- [x] Draft an announcement for peer review to ensure changes included are comprehensive.
- [x] Disable testnet deploy workflow, so that chain is not reset
- [x] Bump the version number and push its tag, via [cargo-release](https://crates.io/crates/cargo-release).
    - [x] Run `cargo release minor` for a new testnet, or `cargo release patch` for a bugfix. For the latter, make sure you're on a dedicated release branch.
    - [x] Push the commit and newly generated tag, e.g. `v0.51.0`, to the remote.
- [x] Manually trigger container-build workflow, bc deploy workflow is disabled
- [x] Wait for the ["Release" workflow](https://github.com/penumbra-zone/penumbra/actions/workflows/release.yml) to complete
- [x] Edit the newly created release object, and add a note summarizing the intent of the release
- [x] Close faucet (chain halt will make it inoperative anyway)
- [x] Run migrations on all validators
- [x] Run migrations on all fullnodes
- [x] Update Galileo deployment, [following docs](https://github.com/penumbra-zone/galileo)
[...]

PR #4503, opened 2024-05-30
pd: add migration for testnet77

conorsch:
## Describe your changes
Adds a new migration, which resets the halt bit. No other migration logic is required, as no state-breaking changes are slated for testnet 77. This way, we preserve `pd migrate` as the default invocation for node operators. Also changed here are the APP_VERSION and corresponding docs entries.

Refs #4497.

  > Updates migrations, in order to ensure previously merged consensus-breaking changes are handled correctly.

### DIFF
--- a/crates/bin/pd/src/migrate/testnet77.rs
+++ b/crates/bin/pd/src/migrate/testnet77.rs
@@ -1,0 +1,81 @@
+//! Contains functions related to the migration script of Testnet77.
+//! The Testnet 77 release included several consensus-breaking changes,
+//! but no state-breaking changes, so the migration is essentially a no-op,
+//! other than resetting the halt bit.
+use anyhow::Context;
+use cnidarium::Storage;
+use jmt::RootHash;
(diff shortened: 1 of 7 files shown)


crates/bin/pd/src/migrate/testnet77.rs
```diff
@@ -0,0 +1,78 @@
...
+    let root_hash = initial_state
+        .root_hash()
+        .await
+        .expect("chain state has a root hash");
+    let pre_upgrade_root_hash: RootHash = root_hash.into();
+    let pre_upgrade_height = initial_state
+        .get_block_height()
+        .await
+        .expect("chain state has a block height");
+    let post_upgrade_height = pre_upgrade_height.wrapping_add(1);
```
erwanor, line 37: I think we're missing creating the `StateDelta` and calling `delta.ready_to_start()` + commit in place, unless my eyes betray me

erwanor (approved):

PR #4511, opened 2024-05-30
dex: evict excess of liquidity positions

erwanor:
## Description

This PR:
- [x] break out `StateWriteExt` and trim the dex component's public interface
- [x] block-based tracking of "active" trading pairs (= has had an LP opened)
- [x] implement the end block eviction logic

Final step to implement #4077 

  > Consensus breaking

### DIFF
--- a/crates/core/component/dex/src/component/dex.rs
+++ b/crates/core/component/dex/src/component/dex.rs
@@ -1,17 +1,18 @@
-use std::{collections::BTreeMap, sync::Arc};
+use std::sync::Arc;
 
 use anyhow::Result;
 use async_trait::async_trait;
 use cnidarium::{StateRead, StateWrite};
 use cnidarium_component::Component;
+use penumbra_asset::asset;
 use penumbra_asset::{Value, STAKING_TOKEN_ASSET_ID};
 use penumbra_proto::{StateReadProto, StateWriteProto};
 use tendermint::v0_37::abci;
 use tracing::instrument;
 
 use crate::{
-    component::flow::SwapFlow, event, genesis, state_key, BatchSwapOutputData, DexParameters,
-    DirectedTradingPair, SwapExecution, TradingPair,
+    component::SwapDataRead, component::SwapDataWrite, event, genesis, state_key,
+    BatchSwapOutputData, DexParameters, DirectedTradingPair, SwapExecution, TradingPair,
 };
 
 use super::{
@@ -54,7 +55,6 @@ impl Component for Dex {
         // This has already happened in the action handlers for each `PositionOpen` action.
 
         // 2. For each batch swap during the block, calculate clearing prices and set in the JMT.
-
         let routing_params = state.routing_params().await.expect("dex params are set");
 
         for (trading_pair, swap_flows) in state.swap_flows() {
@@ -135,9 +135,26 @@ impl Component for Dex {
     }
 }
 
-/// Extension trait providing read access to dex data.
+/// Provides public read access to DEX data.
 #[async_trait]
 pub trait StateReadExt: StateRead {
+    /// Gets the DEX parameters from the state.
+    async fn get_dex_params(&self) -> Result<DexParameters> {
+        self.get(state_key::config::dex_params())
+            .await?
+            .ok_or_else(|| anyhow::anyhow!("Missing DexParameters"))
+    }
+
+    /// Uses the DEX parameters to construct a `RoutingParams` for use in execution or simulation.
+    async fn routing_params(&self) -> Result<RoutingParams> {
+        let dex_params = self.get_dex_params().await?;
+        Ok(RoutingParams {
+            max_hops: dex_params.max_hops as usize,
+            fixed_candidates: Arc::new(dex_params.fixed_candidates),
+            price_limit: None,
+        })
+    }
+
     async fn output_data(
         &self,
         height: u64,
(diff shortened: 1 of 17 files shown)


crates/core/component/dex/src/component/eviction_manager.rs
```diff
@@ -0,0 +1,112 @@
...
+                .await?;
+
+            let overlap = overhead_ab.intersection(&overhead_ba);
+
+            for id in overlap {
+                self.close_position_by_id(id).await?;
+            }
+        }
+        Ok(())
+    }
```
erwanor, line 109: I considered `JoinSet`ifying across trading pairs, but it feels like unlikely to be helpful because it would take a great deal of trading pairs for this code to be slow, and gas costs and ineluctable eviction would make it exorbitant to sustain.

crates/core/component/dex/src/state_key.rs
```diff
@@ -237,5 +246,11 @@ pub(crate) mod eviction_queue {
 
             full_key
         }
+
+        pub(crate) fn parse_id_from_key(key: Vec<u8>) -> anyhow::Result<[u8; 32]> {
+            ensure!(key.len() == 155, "key must be 155 bytes");
+            let k = &key[123..155];
+            Ok(k.try_into()?)
+        }
```
erwanor, line 254: Slightly gnarly, I'll admit. This save us 32 bytes of nonverifiable storage, which actually kind of adds up to `64Kb` per trading pair per block in the pathological case.

erwanor:
I have tested this manually on a devnet using a tweaked version of `tx replicate`. Reviews are welcome now, but I won't merge until I have an app test that spins up a chain with a low max position parameter.

crates/core/component/dex/src/component/dex.rs
```diff
@@ -113,7 +114,15 @@ impl Component for Dex {
...
         }
 
-        // 4. Close all positions queued for closure at the end of the block.
+        // 4. Inspect trading pairs that saw new position opened during this block, and
+        // evict their excess LPs if any are found.
+        Arc::get_mut(state)
+            .expect("state should be uniquely referenced after batch swaps complete")
+            .evict_positions()
+            .await
+            .expect("MERGEBLOCK(erwan): remove this");
```
erwanor, line 123: This `MERGEBLOCK` is going to go away, we will log the failure and carry on.

crates/core/component/dex/src/component/eviction_manager.rs
cratelyn: that makes sense to me! :+1:

crates/core/component/dex/src/component/swap_manager.rs
```diff
@@ -1,37 +1,105 @@
...
+        self.dex_vcb_credit(Value {
+            amount: swap_flow.0,
+            asset_id: trading_pair.asset_1,
+        })
+        .await?;
+        self.dex_vcb_credit(Value {
+            amount: swap_flow.1,
+            asset_id: trading_pair.asset_2,
+        })
+        .await?;
```
cratelyn, line 84: spot checked that these field accesses are correct :+1:

crates/core/component/dex/src/component/dex.rs
```diff
@@ -135,9 +144,26 @@ impl Component for Dex {
...
+
+    /// Uses the DEX parameters to construct a `RoutingParams` for use in execution or simulation.
+    async fn routing_params(&self) -> Result<RoutingParams> {
+        let dex_params = self.get_dex_params().await?;
+        Ok(RoutingParams {
+            max_hops: dex_params.max_hops as usize,
+            fixed_candidates: Arc::new(dex_params.fixed_candidates),
+            price_limit: None,
+        })
+    }
```
cratelyn, line 165: i have a commit here, that i would like to offer: 31d9f5d12

cratelyn (approved):
this looks good to me. nice work! i found a bug in our "recently accessed assets" logic, but i see that it comes from preexisting code, so i don't consider it blocking this change, despite it being a part of the diff.

the logic looks good, and the refactorings along the way have great side-effects. :slightly_smiling_face:

crates/core/component/dex/src/component/dex.rs
```diff
@@ -160,47 +186,67 @@ pub trait StateReadExt: StateRead {
...
+            return;
+        }
+
+        // If the asset is already in the fixed candidate list, don't insert it.
+        if fixed_candidates.contains(&asset_id) {
+            return;
+        }
+
+        assets.insert(asset_id);
+        self.object_put(state_key::recently_accessed_assets(), assets);
```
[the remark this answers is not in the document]
zbuc: The recently accessed asset list is _intra-block_ so it will get reset when the next block begins -- it might be okay as-is. Or we increase the `RECENTLY_ACCESSED_ASSET_LIMIT` (but that could blow up routing complexity) -- it doesn't seem like we'd want to evict an asset that was added to the recently access list at the beginning of the current block, but we also probably wouldn't want to exclude assets that were accessed near the end of the block after the list is full.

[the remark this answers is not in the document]
zbuc: also (as @erwanor points out to the shock of shitcoin developers everywhere) -- "there aren't that many good assets" -- the limit of 10 may not be hit regularly -- however I am realizing there is a potential abuse scenario where the list could be "polluted" with assets that are unlikely to be useful in routing (user-specific tokens, NFTs and the like) -- we likely want to filter those from the recently accessed list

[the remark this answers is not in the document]
erwanor: Discussed this offline, but that's the expected behavior for this light liquidity heuristic, it takes `RECENTLY_ACCESSED_ASSET_LIMIT` non-default assets and use them to augment the search space. We don't need to do more than that because it's not load bearing for path search, it's only meant to supplement it while incurring as small an overhead as possible.

crates/core/component/dex/src/state_key.rs
```diff
@@ -60,6 +60,14 @@ pub mod candlesticks {
     }
 }
 
+pub mod block_scoped {
+    pub mod active {
+        pub fn trading_pairs() -> &'static str {
+            "dex/block_scoped/active/trading_pairs"
+        }
+    }
+}
```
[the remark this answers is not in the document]
erwanor: That's right! This key tracks object storage which is ephemeral and not part of the verifiable chain state.

crates/core/component/dex/src/component/dex.rs
erwanor: Cherry-picked, thx!

zbuc (approved):

Issue #4446, opened 2024-05-23
Audit use of panicky-macros


zbuc:
In #4444 @cratelyn noted that the use of the `unimplemented!()` macro in RPCs can allow a malicious client to crash the node software.

I performed a brief search through the codebase and saw several other calls to `unimplemented` elsewhere. We should audit the codebase for panicky macros (`panic`, `unimplemented`, `todo`, etc.) and ensure they're unable to crash node software.

erwanor:
It should only interrupt the connection right? if it makes the full node panic we need to do this immediately

cratelyn:
i'm not sure if there is comet behavior involved when you say "_full node panic_" but yes, `unimplemented!`, `todo!`, and co. will cause the current thread to [panic](https://doc.rust-lang.org/stable/src/core/macros/mod.rs.html#817). if panics are exposed in any publicly accessible endpoints, they almost always end up being DDoS-shaped bugs.

zbuc:
I just tested this by adding a panic to the `batch_swap_output_data` RPC. Good news is that the entire node does not crash; I think Tonic handles recovery from the task's panic.

PR #4518, opened 2024-05-31
Exclude IBC assets from TotalSupply if their denom metadata is nonexistent

zbuc:
## Describe your changes

This changes the `TotalSupply` query to skip including assets when there is no denom metadata available, for example liquidity position NFTs. This is okay since these are mostly of interest to a single user rather than the market as a whole.

I validated that the error when querying on the current v0.77.0 testnet goes away.

```
$ grpcurl -plaintext -vv localhost:8080 cosmos.bank.v1beta1.Query/TotalSupply
```

  > Changes RPC output only


erwanor (approved):
This looks good to me, we can do a round of improvement later as part of #4446

PR #4521, opened 2024-06-01
dex: less awkward logging if no arbs are found

### LINKED ISSUE #4513: Tracing output about DEX arb is suggestive of a bug
**Describe the bug**

```
2024-05-31T07:14:54.747390Z  INFO abci:EndBlock{height=223048}:end_block{height=223048}:dex: penumbra_dex::component::dex: arbitrage successful! surplus=Value { amount: 0, asset_id: passet1984fctenw8m2fpl8a9wzguzp7j34d7vravryuhft808nyt9fdggqxmanqm }
```
This was not a successful arbitrage, and should not be reported as such. But more to the point, we should be sure that we're not committing any state changes as a result of it (i.e., the fix is not to skip the tracing message, but to ensure that we're discarding 0-valued arb).

erwanor:
## Describe your changes

Fix #4513, this change the arbitrage method signature to return `Result<Option<Value>>`, and avoid logging that an arbitrage run is "successful" if the state fork gets discarded. 

  > Mechanical refactor

### DIFF
--- a/crates/core/component/dex/src/component/dex.rs
+++ b/crates/core/component/dex/src/component/dex.rs
@@ -107,8 +107,11 @@ impl Component for Dex {
             .arbitrage(*STAKING_TOKEN_ASSET_ID, arb_routing_params)
             .await
         {
-            // The arb search completed, it may or may not have found surplus.
-            Ok(v) => tracing::info!(surplus = ?v, "arbitrage successful!"),
+            // The arb search completed successfully, and surfaced some surplus.
+            Ok(Some(v)) => tracing::info!(surplus = ?v, "arbitrage successful!"),
+            // The arb completed without errors, but resulted in no surplus, so
+            // the state fork was discarded.
+            Ok(None) => tracing::debug!("no arbitrage found"),
             // The arbitrage search should not error, but if it does, we should
             // simply not perform arbitrage, rather than halting the entire chain.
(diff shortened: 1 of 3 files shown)


crates/core/component/dex/src/component/dex.rs
```diff
@@ -107,8 +107,11 @@ impl Component for Dex {
             .arbitrage(*STAKING_TOKEN_ASSET_ID, arb_routing_params)
             .await
         {
-            // The arb search completed, it may or may not have found surplus.
-            Ok(v) => tracing::info!(surplus = ?v, "arbitrage successful!"),
+            // The arb search completed successfully, and led to state changes.
+            Ok(Some(v)) => tracing::info!(surplus = ?v, "arbitrage successful!"),
+            // The arb completed without errors, but resulted in no surplus, so their
+            // state was not altered.
+            Ok(None) => tracing::info!("no arbitrage found"),
```
hdevalence, line 114: Nit: I'd demote this to debug so that we only show the info if it actually happened, and omit it on an empty block.

hdevalence (approved):

crates/core/component/dex/src/component/dex.rs
erwanor: sounds good

Issue #4522, opened 2024-06-03
Add a timestamp endpoint to SCT QueryService


turbocrime:
**Is your feature request related to a problem? Please describe.**

The SCT service stores timestamps for each block height, but this information is not accessible to clients. Users would like to know things like transaction times, which could be identified by block timestamp. Graph axes such as candlestick graphs could be labelled with timestamps.

**Describe the solution you'd like**

Add a timestamp endpoint to the SCT QueryService that allows querying the timestamp of a specific block height

**Describe alternatives you've considered**

Include block height timestamps in other queries that mention block heights, such as the CandlestickData query, or a CompactBlock query

turbocrime:
i think a lower resolution such as epoch timestamp would also be fine, for tx logs or graph purposes

aubrika:
Is this blocking on the web side @turbocrime ? If it is a nice to have we won't have bandwidth proximally

turbocrime:
closed by #4523

PR #4523, opened 2024-06-03
sct(component): add `TimestampByHeight` to query service

turbocrime:
## Describe your changes

Adds a TimestampByHeightRequest to SCT QueryService.

There is no corresponding migration added, so any block heights lower than the height at which this is deployed will have missing data.

https://github.com/penumbra-zone/penumbra/issues/4522

  > This adds a pb API endpoint and should not affect consensus. New data is tracked in nonconsensus storage only.


turbocrime:
no tests - haven't looked at that yet. is this the correct format of the timestamp? are there other considerations to make?

crates/core/component/sct/src/component/rpc.rs
```diff
@@ -55,4 +56,24 @@ impl QueryService for Server {
...
+
+    #[instrument(skip(self, request))]
+    async fn timestamp_by_height(
+        &self,
+        request: tonic::Request<TimestampByHeightRequest>,
+    ) -> Result<tonic::Response<TimestampByHeightResponse>, Status> {
+        let state = self.storage.latest_snapshot();
+
+        let height = request.get_ref().height;
+        let block_time = state.get_block_timestamp().await.map_err(|e| {
```
erwanor, line 68: This will always return the timestamp for the latest block in the snapshot

erwanor: I don't think we store height-by-height timestamp, but I can take a look once I'm back at the computer

zbuc: Correct, we would need to add a new nonverifiable storage key containing timestamp-by-height

zbuc:
I added a modification to track historic timestamps. Unfortunately writing a migration for this isn't really straightforward, but it's probably fine that historic data won't exist as it's in non-verifiable storage.

aubrika:
Seems like as written, worst case scenario is that querying for a historical block timestamp without a migration to populate that data is just going to throw an error. @turbocrime is that acceptable for your intended use case - i.e. erroring when called on blocks prior to an upgrade, returning data afterwards?

turbocrime:
throwing is fine - it's a non-critical case and we have a fallback already

turbocrime (commented):
reviewed, looks good

zbuc (approved):

PR #4528, opened 2024-06-03
dex: 🌱 add a `SimulationsDisabled` stub for dex queries

cratelyn:
if the `--enable-expensive-rpc` is not given, cryptic errors can be found on the client side. this adds a stub service that can be used in lieu of the expensive service, returning a helpful error message.

  > only changes rpc services.

### DIFF
--- a/crates/core/app/src/rpc.rs
+++ b/crates/core/app/src/rpc.rs
@@ -23,7 +23,7 @@ use {
     },
     penumbra_auction::component::rpc::Server as AuctionServer,
     penumbra_compact_block::component::rpc::Server as CompactBlockServer,
-    penumbra_dex::component::rpc::Server as DexServer,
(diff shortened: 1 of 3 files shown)


zbuc (approved):

hdevalence:
This doesn't really solve the problem of mishandling missing GRPC services though right?

cratelyn:
> This doesn't really solve the problem of mishandling missing GRPC services though right?

no, it doesn't. there's a broader problem, but this does help provide a friendlier error message for the common case of a client querying the simulation endpoint when that is disabled. this took ~2 minutes and it addresses the most common manifestation of that problem, since the simulation endpoint is disabled by default.

hdevalence:
Yep, sounds like a good fix to the object level problem, I'm just wondering if we should track an issue to improve the error handling in general

PR #4535, opened 2024-06-04
dex: correct placement of arb metrics emission

hdevalence:
## Describe your changes

We noticed there were no metrics being emitted for the cost of the arb search when no arb was found.

This fixes that, and should be backported into the next point release.

  > Only changes metrics emission


erwanor (approved):

erwanor:
Thank you for fixing this

Issue #4306, opened 2024-05-02
fees: exercise non-zero gas fees in testnet


TalDerei:
Perform a governance parameter change proposal to modify the [genesis parameter](https://github.com/penumbra-zone/penumbra/blob/main/crates/core/component/fee/src/genesis.rs) for setting non-zero gas prices in the testnet.  This seems like a natural step forward following https://github.com/penumbra-zone/penumbra/issues/4154. 

cc @conorsch

hdevalence:
This can't happen without changes currently part of #4320, because there is a critical bug in the current fee component that makes it impossible to set the fees via governance like they are supposed to be.

conorsch:
Blocked until we get out the next chain upgrade, tracked in https://github.com/penumbra-zone/penumbra/issues/4325.

erwanor:
We discussed this synchronously, and the current plan is to do a parameter change proposal some time after we land #4402, but well ahead of a second upgrade that would include #4326.

TalDerei:
what's the current status of this?

erwanor:
We haven't turned them on, but we could do that later today or tomorrow once 77 is deployed

conorsch:
We discussed this in sprint-planning today, and want to give it a go. I'm happy to pair with @TalDerei on submitting a gov proposal, so more folks on the team are familiar with the workflows involved. Still TBD is what the structure of the param change proposal needs to be. Generating a template, I see relevant entries as:

```
[[changes]]
component = "feeParams"
key = "fixedGasPrices"
value = "{}"

[[changes]]
component = "feeParams"
key = "fixedAltGasPrices"
value = "[{\"assetId\":{\"inner\":\"HW2Eq3UZVSBttoUwUi/MUtE7rr2UU7/UH500byp7OAc=\"}},{\"assetId\":{\"inner\":\"nwPDkQq3OvLnBwGTD+nmv1Ifb2GEmFCgNHrU++9BsRE=\"}}]"
```

So what should we set `fixedGasPrices` to, exactly? We need to specify the amount of upenumbra, and also figure out how to structure that request in the param change proposal. We'll test on preview and/or devnets to figure out the right shape.

conorsch:
To figure out the shape of the param change, as it needs to be specified in the TOML proposal file, I created a local devnet with fees enabled (via `--gas-price-simple 5`, same as we use in smoke tests), and then polled the grpc endpoint to inspect how the fee params were specified:

```
[...]

PR #4539, opened 2024-06-04
feat(fee): add FromStr, Display impls for FeeTier

conorsch:
## Describe your changes
We recently enabled fees on Testnet 77 (#4306), and in the process broke a few things, such as Galileo. While working on adding FeeTier support to Galileo, Galileo's clap setup wanted Display and FromStr. Added those to satisfy the build, and then circled back to pcli to reuse the same impls, which allowed us to snip out some duplicative code.

  > no changes to app logic, just adding more impls

### DIFF
--- a/crates/bin/pcli/src/command/tx.rs
+++ b/crates/bin/pcli/src/command/tx.rs
@@ -28,6 +28,7 @@ use regex::Regex;
 use liquidity_position::PositionCmd;
 use penumbra_asset::{asset, asset::Metadata, Value, STAKING_TOKEN_ASSET_ID};
 use penumbra_dex::{lp::position, swap_claim::SwapClaimPlan};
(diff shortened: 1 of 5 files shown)


cratelyn (approved):
nice! thanks for additionally cleaning up the duplicate `FeeTier` in pcli while you did this. i appreciate the thought, and attention to detail. great work @conorsch!

zbuc:
Does this still select "low" by default if no tier is chosen?

cratelyn:
> Does this still select "low" by default if no tier is chosen?

https://github.com/penumbra-zone/penumbra/pull/4539/files#diff-87de50c1ecb12dbe4371567b45a2b49fa5b793bc4bcee7cab6776cc40ea5d61dR145-R149

yes! see here :point_up:

Issue #4545, opened 2024-06-04
Release 0.77.2, via point-release


conorsch:
# Tooling release

Testnet name: Testnet 77, `penumbra-testnet-deimos-8` (unchanged)
Release date: 2024-06-04/5
Testnet release manager: @conorsch

# Testnet Release Manager Checklist

On Testnet 77, we enabled fees #4306, and some things broke. Release 0.77.2 is mostly oriented around resolving breakages, but we'll ship some opportunistic improvements where possible, too.

# Testnet Release Manager Checklist

Pre-release:

- [ ] Enumerate the specific PRs that should be backported for the release:
  - [ ] https://github.com/penumbra-zone/penumbra/pull/4536
  - [ ] https://github.com/penumbra-zone/penumbra/pull/4535
  - [ ] https://github.com/penumbra-zone/penumbra/pull/4539
  - [ ] https://github.com/penumbra-zone/penumbra/pull/4551
  - [x] https://github.com/penumbra-zone/penumbra/issues/4540
  - [x] https://github.com/penumbra-zone/penumbra/issues/4541

On release day:

- [x] Draft an announcement for peer review to ensure changes included are comprehensive.
- [x] Enable testnet deploy workflow, so that point-release logic will run
- [x] Bump the version number and push its tag, via [cargo-release](https://crates.io/crates/cargo-release).
    - [x] Run `cargo release minor` for a new testnet, or `cargo release patch` for a bugfix. For the latter, make sure you're on a dedicated release branch.
    - [x] Push the commit and newly generated tag, e.g. `v0.51.0`, to the remote.
- [ ] Manually trigger container-build workflow, bc deploy workflow is disabled
- [x] Wait for the ["Release" workflow](https://github.com/penumbra-zone/penumbra/actions/workflows/release.yml) to complete
- [x] Edit the newly created release object, and add a note summarizing the intent of the release
- [x] Close faucet (chain halt will make it inoperative anyway)
- [x] Update Galileo deployment, [following docs](https://github.com/penumbra-zone/galileo)
- [x] Make the announcement to Discord! 🎉🎉🎉

Post-release cleanup tasks

- [x] Ensure faucet is open

conorsch:
Shipped: https://github.com/penumbra-zone/penumbra/releases/tag/v0.77.2

PR #4551, opened 2024-06-04
ci: support point-releases to solo node

### LINKED ISSUE #4459: ci: solo pd node reset on point release
Notes for later: the point-release CI job doesn't handle the solo pd nodes (#3336) gracefully: https://solo-pd.testnet.plinfra.net/ was reset and now must sync from scratch. Let's circle back and make that nicer.

_Originally posted by @conorsch in https://github.com/penumbra-zone/penumbra/issues/4448#issuecomment-2128144106_

conorsch:
## Describe your changes
The solo node, run outside the cluster, was being fully reset on point releases, due to the aggressive reset logic that was written during the days of frequent chain resets. We're performing point-releases more regularly now, so let's make sure to avoid resetting the solo node in those cases. Still unsupported is chain-migration logic specifically for the solo node, but that's fine: we're handling chain upgrades out of band for now anyway.

Closes #4459.

  > ci-only, no application changes

### DIFF
--- a/deployments/scripts/redeploy-ci-fullnode-via-remote
+++ b/deployments/scripts/redeploy-ci-fullnode-via-remote
@@ -23,69 +23,109 @@ else
     exit 2
 fi
 
-# Additional sanity-check to ensure we're running in the proper CI context.
-if ! getent passwd | grep -q "^penumbra:" ; then
-    >&2 echo "ERROR: 'penumbra' user not found."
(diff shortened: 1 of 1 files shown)


aubrika (approved):
LGTM!

conorsch:
I'd like this in for #4545 so I'll backport shortly.

PR #4554, opened 2024-06-04
Change Planner to use optional gas prices field and error if unset

aubrika:
Closes #4555 and fixes fee inclusion in validator definition upload, validator vote casting, and tx sweep.

This also introduces some protection for future planner invocations. The `Planner`'s `gas_prices` field has been changed to an `Option` type and if it's not set it will return an error during planning:

```
$ cargo run --bin pcli --release -- --home ~/.local/share/pcli-preview/ validator definition upload --file validator.toml
...
Error: planner instances must call set_gas_prices prior to planning
```

- [x] If this code contains consensus-breaking changes, I have added the "consensus-breaking" label. Otherwise, I declare my belief that there are not consensus-breaking changes, for the following reason:

  > should be limited to addressing client-side failures on tx submission, so shouldn't change consensus on which

Issue #4540, opened 2024-06-04
fees: validator voting broken


conorsch:
Since enabling fees in #4306, we've observed that validator voting is broken:

```
❯ pcli validator vote cast yes --on 9
Scanning blocks from last sync height 375858 to latest height 375858
[0s] ██████████████████████████████████████████████████       0/0       0/s ETA: 0s
building transaction [1 actions, 0 proofs]...
finished proving in 0.002 seconds [1 actions, 0 proofs, 297 bytes]
broadcasting transaction and awaiting confirmation...
Error: error broadcasting transaction

Caused by:
    status: Internal, message: "Error submitting transaction: code 1, log: failed to deliver transaction: check_stateful failed: fee must be greater than or equal to the transaction base price (supplied: 0, base: 1)", details: [], metadata: MetadataMap { headers: {} }
```

Let's fix and point-release.

conorsch:
Resolved by https://github.com/penumbra-zone/penumbra/pull/4554

PR #4559, opened 2024-06-04
backport fixes to planner gas prices & validator voting

conorsch:
## Describe your changes
Backports two PRs for 0.77.2 (#4545):

* https://github.com/penumbra-zone/penumbra/pull/4510
* https://github.com/penumbra-zone/penumbra/pull/4554

The planner changes in the latter are crucial to point-release. The former less so, but including those changes reduced conflicts on the cherry-pick, and the client-side improvement is a nice-to-have as we focus on resolving #4540.

  > client-side changes only, no changes to consensus logic


cronokirby (approved):
(saw the original pr)

PR #4562, opened 2024-06-04
ci: don't run long tests on docs-only PRs

conorsch:
## Describe your changes
I want to keep the docs up to date, and sometimes that requires fussy management of version numbers, particularly as we invest in multi-branch workflows to support point releases (e.g. #4545). If *only* the docs changed, skip the long-running rust tests, because we don't need to validate rust functionality on docs changes.

  > ci-only


cratelyn (approved):
😍 amazing! thank you for this!

Issue #3620, opened 2024-01-17
penumbra: review user supplier string fields in messages


erwanor:
**Is your feature request related to a problem? Please describe.**
In #2066, we discovered that validator definition string fields were not length-limited. We should review governance action handlers (among others) to make sure that we check those fields.

Assigning myself to keep track of this issue, however it's a good first issue and I would be happy to have a contributor take it on and play assist.

cronokirby:
Idea: create a SmallString<N> type which will automatically validate upon construction that the string is less than N characters. Then you just put it in the action struct, and it becomes impossible to not validate this in the resulting app logic.

hdevalence:
I think it would be simpler and more ergonomic to put the validation in the `TryFrom` domain type conversion rather than introduce our own custom string type that would have feature parity with the stdlib's string type. Otherwise we put ourselves through worse ergonomics forever.

zbuc:
> I think it would be simpler and more ergonomic to put the validation in the `TryFrom` domain type conversion rather than introduce our own custom string type that would have feature parity with the stdlib's string type. Otherwise we put ourselves through worse ergonomics forever.

One downside of this is duplication, for example `Validator` has three different `TryFrom` implementations:

`impl TryFrom<ValidatorToml> for Validator {`
`impl TryFrom<pb::Validator> for Validator {`
`impl TryFrom<&TestnetValidator> for Validator {`

PR #4567, opened 2024-06-06
Enforce string limits for deserialization

zbuc:
## Describe your changes

This follows the suggestion from @hdevalence in https://github.com/penumbra-zone/penumbra/issues/3620#issuecomment-2141013858 to add validation of user-supplied `String` lengths to the various `TryFrom` protobuf deserialization implementations.

The user-supplied `String` values concerned are found within the governance and staking crates, concerning governance proposals and validator definitions.

**This is consensus breaking** as historic data **may not comply with the new limitations**. Thus there may be unexpected panics when trying to retrieve and deserialize data from storage.

An additional migration has been included to truncate any existing data that does not adhere to the new limits.

The strings are configured to use **byte length limits** rather than **character length limits**. This seems favorable because values will be guaranteed to fit into fixed-size buffers of the byte length limit for the field. Code based on https://github.com/rust-lang/rust/issues/93743 was pulled in to perform the truncation at UTF-8 character boundaries.

Closes https://github.com/penumbra-zone/penumbra/issues/3620

### DIFF
--- a/crates/bin/pd/src/migrate.rs
+++ b/crates/bin/pd/src/migrate.rs
@@ -10,19 +10,24 @@ mod testnet72;
 mod testnet74;
 mod testnet76;
 mod testnet77;
+mod testnet78;
 
 use anyhow::{ensure, Context};
-use penumbra_governance::StateReadExt;
-use penumbra_sct::component::clock::EpochRead;
+use jmt::RootHash;
+use penumbra_app::app::StateReadExt as _;
+use penumbra_governance::{StateReadExt, StateWriteExt as _};
+use penumbra_sct::component::clock::{EpochManager as _, EpochRead};
 use std::path::{Path, PathBuf};
 use tracing::instrument;
 
-use cnidarium::Storage;
+use cnidarium::{StateDelta, Storage};
 use penumbra_app::SUBSTORE_PREFIXES;
 
 use flate2::write::GzEncoder;
 use flate2::Compression;
 use std::fs::File;
+
+use crate::testnet::generate::TestnetConfig;
 
 /// The kind of migration that should be performed.
 #[derive(Debug)]
@@ -47,6 +52,9 @@ pub enum Migration {
     /// Testnet-77 migration:
     /// - Reset the halt bit
     Testnet77,
+    /// Testnet-78 migration:
+    /// - Truncate various user-supplied `String` fields to a maximum length.
+    Testnet78,
 }
 
 impl Migration {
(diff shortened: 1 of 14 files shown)


crates/core/component/governance/src/proposal.rs
```diff
@@ -82,6 +82,17 @@ impl TryFrom<pb::Proposal> for Proposal {
...
 
     fn try_from(inner: pb::Proposal) -> Result<Proposal, Self::Error> {
+        // Validation (matches limits from `impl AppActionHandler for ProposalSubmit`):
+        // - Title has a max length of 80 chars
+        if inner.title.len() > 80 {
+            anyhow::bail!("proposal title field must be less than 80 characters");
+        }
+
+        // - Description has a max length of 10_000 chars
+        if inner.description.len() > 10_000 {
```
erwanor, line 92: This makes sense, CSDK has a similar limit.

crates/core/component/governance/src/proposal.rs
```diff
@@ -82,6 +82,17 @@ impl TryFrom<pb::Proposal> for Proposal {
     type Error = anyhow::Error;
 
     fn try_from(inner: pb::Proposal) -> Result<Proposal, Self::Error> {
+        // Validation (matches limits from `impl AppActionHandler for ProposalSubmit`):
+        // - Title has a max length of 80 chars
+        if inner.title.len() > 80 {
```
erwanor, line 87: CSDK limit is 255 characters. This works too imo.

crates/core/component/governance/src/proposal.rs
```diff
@@ -95,6 +106,11 @@ impl TryFrom<pb::Proposal> for Proposal {
                     commit: if signaling.commit.is_empty() {
                         None
                     } else {
+                        // Commit hash has max length of 64 chars:
+                        if signaling.commit.len() > 64 {
```
erwanor, line 110: How about we make this 255 so it can be used with IPFS links, fine to leave as is too

crates/core/component/governance/src/proposal.rs
```diff
@@ -123,9 +139,15 @@ impl TryFrom<pb::Proposal> for Proposal {
                 Payload::UpgradePlan(upgrade_plan) => ProposalPayload::UpgradePlan {
                     height: upgrade_plan.height,
                 },
-                Payload::FreezeIbcClient(freeze_ibc_client) => ProposalPayload::FreezeIbcClient {
-                    client_id: freeze_ibc_client.client_id,
-                },
+                Payload::FreezeIbcClient(freeze_ibc_client) => {
+                    // Validation: client ID has a max length of 128 chars
+                    if freeze_ibc_client.client_id.len() > 128 {
```
erwanor, line 144: Makes sense to me

crates/core/component/governance/src/proposal_state.rs
```diff
@@ -292,6 +292,11 @@ impl TryFrom<pb::ProposalOutcome> for Outcome<String> {
                     withdrawn,
                 }) => Outcome::Failed {
                     withdrawn: if let Some(pb::proposal_outcome::Withdrawn { reason }) = withdrawn {
+                        // Max reason length is 1kb
+                        if reason.len() > 1024 {
+                            anyhow::bail!("withdrawn reason is too long")
```
erwanor, line 297: Makes sense, i'd put the charlimit in the error message so this bubble up as a useful message

crates/core/component/stake/src/validator.rs
```diff
@@ -133,6 +133,22 @@ impl TryFrom<ValidatorToml> for Validator {
...
+
+        // - Name has a max length of 140 chars
+        if v.name.len() > 140 {
+            anyhow::bail!("validator name must be less than 140 characters");
+        }
+
+        // - Description has a max length of 280 chars
+        if v.description.len() > 280 {
+            anyhow::bail!("validator description must be less than 280 characters");
+        }
```
erwanor, line 150: ACK. The funding stream check is in the `FundingStreams` conversion impl.

crates/core/component/stake/src/validator/definition.rs
```diff
@@ -29,11 +29,27 @@ impl From<Definition> for pb::ValidatorDefinition {
...
+        // - Name has a max length of 140 chars
+        if validator.name.len() > 140 {
+            anyhow::bail!("validator name must be less than 140 characters");
+        }
+
+        // - Description has a max length of 280 chars
+        if validator.description.len() > 280 {
+            anyhow::bail!("validator description must be less than 280 characters");
+        }
+
```
erwanor, line 50: I don't think we need/want this because we construct a `Validator` down below which is where the validation occurs

crates/bin/pd/src/migrate.rs
```diff
@@ -72,37 +80,96 @@ impl Migration {
...
+            .await
+            .expect("chain state has a block height");
+        let post_upgrade_height = pre_upgrade_height.wrapping_add(1);
+
+        let migration_duration = match self {
+            Migration::SimpleMigration => simple::migrate(storage.clone()).await?,
+
+            Migration::Testnet78 => testnet78::migrate(storage.clone()).await?,
+            _ => unreachable!(),
+        };
```
erwanor, line 107: I don't think we should do this specific refactor, it doesn't address the module's main issues (brittleness) but it shuffles important code paths around, right after we tested them, but ahead of having integration tests ready (#4337).

In its current state, it makes genesis generation work against a stale snapshot which is a bug. It would be caught by integration tests but this is an example of how deliberate we have to be about reworking this, especially now.

erwanor (changes requested):
Relative to the proposed migration structure, the limits lgtm

crates/bin/pd/src/migrate.rs
zbuc: I'm mixed on whether to refactor right now or not. I hear you about the testing, however adding the new migration seems to require a careful copy-pasting of genesis handling code and manual testing anyways. I like your suggestion of moving everything into a `Trait` impl.

crates/core/component/stake/src/validator/definition.rs
zbuc: Since the `Definition` can get included on-chain (as part of `Action::ValidatorDefinition`) it seems important to apply a length limit to it.

crates/bin/pd/src/migrate.rs
zbuc: Reverted the refactoring for now

crates/core/component/governance/src/validator_vote/action.rs
```diff
@@ -4,7 +4,7 @@ use penumbra_stake::{GovernanceKey, IdentityKey};
 use penumbra_txhash::{EffectHash, EffectingData};
 use serde::{Deserialize, Serialize};
 
-use crate::vote::Vote;
+use crate::{vote::Vote, MAX_VALIDATOR_VOTE_REASON_LENGTH};
```
cronokirby, line 7: Maybe we should use this constant in proposal_state.rs too? If it requires changing visibility then that's not worth it, just wondering.

crates/bin/pd/src/migrate/testnet72.rs
```diff
@@ -1,5 +1,5 @@
 //! Contains functions related to the migration script of Testnet72
-
+#![allow(dead_code)]
```
cronokirby, line 2: Why was this added to the old migrations?

crates/bin/pd/src/migrate/testnet78.rs
```diff
@@ -0,0 +1,392 @@
...
+    if index >= s.len() {
+        s.len()
+    } else {
+        let lower_bound = index.saturating_sub(3);
+        let new_index = s.as_bytes()[lower_bound..=index]
+            .iter()
+            .rposition(|b| is_utf8_char_boundary(*b));
+
+        // SAFETY: we know that the character boundary will be within four bytes
+        unsafe { lower_bound + new_index.unwrap_unchecked() }
```
cronokirby, line 357:
```suggestion
        lower_bound + new_index.unwrap()
```

Why don't we just avoid the unsafe thing altogether? Seems like we'd prefer a panic in migration testing, if for some reason this code is mistaken (this does seem safe to me, but I don't see the benefit in using an unchecked method)

cronokirby (approved):

crates/bin/pd/src/migrate/testnet72.rs
zbuc: They are no longer callable via `pcli` (since they are only valid for the penumbra version contemporary to the migration script) but we wanted to keep the code around to serve as an example of how to write a migration.

PR #4570, opened 2024-06-06
crypto: check ck_d is consistent with d, pk_d provided on a note

### LINKED ISSUE #4569: check for clue key replacement in addresses on notes
There wasn't a specific finding flagged but the ECC auditors asked a good question about our notes: what happens if an adversary mixes $d, pk_d$ from one address and replaces the clue key $ck_d$?

Currently we wouldn't detect this happening. To avoid shenanigans, when scanning, we should check the $ck_d$ is consistent with that expected from the diversifier.

redshiftzero:

Closes #4569

  > This is only used by the view service


TalDerei (approved):
LGTM!

cronokirby (approved):

PR #4571, opened 2024-06-06
dex: 🔖 instrument search path relaxation

cratelyn:
#### 💭 describe your changes

this makes a few changes to the dex component's prometheus metrics.

see #4464 for previous context.

importantly, this adds a new metric that tracks the time spent processing a
distinct, individual path. this specific metric is configured to use a larger
set of buckets on the upper end, so that we can capture and observe latencies
observed running pd on a testnet while positions are being routed.

this is a sibling metric to the existing path search metric, which records the
time spent finding the final output path. as a part of this branch, we also
fix a small bug, which is that the duration is still recorded if a dead-end
is reached, and no path could be found.

finally, the configuration of the prometheus metrics builder is tweaked so that
we can add counters to this component in the future. additional counters tracking
things like: (1) paths that were ruled out for being too expensive, (2) dead-ends
reached, (3) errors encountered, etc. i would love for reviewers to offer insight
about other conditions worth counting.

* #4464

  > this only affects metrics emission.

### DIFF
--- a/crates/bin/pd/src/main.rs
+++ b/crates/bin/pd/src/main.rs
@@ -166,14 +166,12 @@ async fn main() -> anyhow::Result<()> {
             };
 
             // Configure a Prometheus recorder and exporter.
+            use penumbra_dex::component::metrics::PrometheusBuilderExt;
             let (recorder, exporter) = PrometheusBuilder::new()
                 .with_http_listener(metrics_bind)
                 // Set explicit buckets so that Prometheus endpoint emits true histograms, rather
                 // than the default distribution type summaries, for time-series data.
-                .set_buckets_for_metric(
-                    metrics_exporter_prometheus::Matcher::Prefix("penumbra_dex_".to_string()),
-                    penumbra_dex::component::metrics::DEX_BUCKETS,
-                )?
+                .set_buckets_for_dex_metrics()?
                 .build()
                 .map_err(|e| {
                     let msg = format!(
--- a/crates/core/component/dex/src/component/router/path_search.rs
+++ b/crates/core/component/dex/src/component/router/path_search.rs
@@ -6,6 +6,7 @@ use cnidarium::{StateDelta, StateRead};
 use futures::StreamExt;
 use penumbra_asset::asset;
 use penumbra_num::fixpoint::U128x128;
+use tap::Tap;
 use tokio::task::JoinSet;
 use tracing::{instrument, Instrument};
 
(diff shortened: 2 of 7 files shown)


crates/core/component/dex/src/component/router/path_search.rs
```diff
@@ -48,14 +54,14 @@ pub trait PathSearch: StateRead + Clone + 'static {
 
         let entry = cache.lock().0.remove(&dst);
         let Some(PathEntry { path, spill, .. }) = entry else {
+            record_duration();
             return Ok((None, None));
         };
```
cratelyn, line 59: :point_up: this looks like an arm we should still record the duration of. even if we didn't find a path to return, we still searched for one.

i'd like to put a counter here and in some other arms below, in a subsequent branch.

crates/bin/pd/src/main.rs
```diff
@@ -166,14 +166,12 @@ async fn main() -> anyhow::Result<()> {
...
+            use penumbra_dex::component::metrics::PrometheusBuilderExt;
             let (recorder, exporter) = PrometheusBuilder::new()
                 .with_http_listener(metrics_bind)
                 // Set explicit buckets so that Prometheus endpoint emits true histograms, rather
                 // than the default distribution type summaries, for time-series data.
-                .set_buckets_for_metric(
-                    metrics_exporter_prometheus::Matcher::Prefix("penumbra_dex_".to_string()),
-                    penumbra_dex::component::metrics::DEX_BUCKETS,
-                )?
+                .set_buckets_for_dex_metrics()?
```
cratelyn, line 174: see the relevant commit for more info, but this (a) cuts down on non-local reasoning, and (b) sets the stage for us to add counters and gauges to this component.

hdevalence:
I'm not sure why we want to change the bucket resolution for the metrics in question?

hdevalence (changes requested):
I don't think we should be tuning the bucket sizes like this for the reason I mentioned in my comment

cratelyn:
#### :balance_scale: making the case

i have force pushed to drop d159da248 from this branch. i still feel very strongly that we are opting _not_ to collect information about how our software behaves in production by refraining from adding larger buckets.

prometheus metrics are not the proper tool for measuring performance, they are for observability and alerting. if we want performance measurements, we should write benchmarks and use a library like [`criterion`](https://docs.rs/criterion/latest/criterion/). prometheus metrics are also not for measurements of our specific testnet deployment, they are an operational tool for any deployment, regardless of that being a VM or an on-prem device.

having a bucket counting observations greater than 5 seconds allows operators to configure alerts so that they can be informed when exceptional events occurring in their system. having such a bucket also does not imply that 5 seconds is an acceptable duration for path searches, it helps differentiate significant outages from minor performance hiccups. having such a bucket also allows operators to compare long-tail latencies between a VM and an on-prem device to confirm that the latter does not experience these same periods of elevated latency. having such a bucket is how we can identify the source of elevated path search latency.

we have observed elevated durations like this in the wild, but are not able to make any confident statements about whether those searches took 251ms, or 5 seconds. a 1 second and 5 second bucket answers that question without any significant downside.

---

### :scissors: the buckets are gone now

with that case made, i've removed the buckets.

this branch contains other changes like 0108a4bde that will facilitate the addition of other metrics to the dex, which will help us improve the dex component's telemetry regardless of our buckets' upper-bound.

i would appreciate if you could take another look, when able. thanks.

conorsch:
Thanks, @cratelyn, both for the diff itself and for the detailed discussion you've been adding to the PRs and tickets related to this issue. Extremely high-quality work.

Issue #4566, opened 2024-06-06
document historical reindexing process for abci events


conorsch:
We need figure out the process for creating an ABCI event database from scratch. The process should look something like:

1. Download historical archive of chain state
2. Set up new node to run in archive mode
3. Sync node up to upgrade boundary
4. Update versions, rinse, repeat, until current height

Let's do it manually for now, working from private backups. When I last tried this, getting cometbft to "replay" blocks to pd over ABCI was fairly simple: unfortunately, that replay process did _not_ result in ABCI events being emitted to a newly created sidecar indexing db. Why is that? Is there some other process we should use to handle this use case?

Closely related to #4525 & #4526.

hdevalence:
Syncing a new cometbft instance will solve the event emission problem.

conorsch:
Progress! Dumping some notes here so I can pick this back up. The process so far is a bit clunky, but it's definitely tractable. 

We want to reindex events on the `penumbra-testnet-deimos-8` chain. There have been several upgrades already, so we'll need to rewind several versions to do that. Let's go back to `0.75.1`. That version won't run out of the box with archived node state, because the halt-height will have been reached, causing pd to exit. So, branch from the `v0.75.1` tag, increment the (now-defunct) const `TOTAL_HALT_COUNT` by one, and build that version of pd, saving it as `pd-0.75.1-plus-halt`. Use that to set up the initial version of the nodes.

0. Restore node state from a private backup. This backup was created by tarring up the entire `~/.penumbra/testnet_data/node0` directory on a fullnode. (Related we should sanitize this archive and host it, see #4525).
1. Install `pd-0.75.1-plus-halt` to the system. Run it.
2. Configure CometBFT v0.37.5 with extra options: `--p2p.pex=false --p2p.seeds='' --moniker archive-node-1`. Make sure its CometBFT RPC is accessible to a second node, soon to be created.
3. Set up _another_ node with `pd-0.75.1-plus-halt` and cometbft v0.37.5. Join this node to the first node via `pd-0.75.1-plus-halt testnet join <url>`. Run its cometbft with extra options:: `--p2p.pex=false --p2p.seeds='' --moniker='reindexer-1' --p2p.persistent_peers='tcp://<node_1_id>@<node_1_ip>:<node_1_port>'`
4. Edit node2's cometbft config `[tx_index]` block and set `indexer="psql"` & `psql_conn="<POSTGRES_DB_URL>"`. Run it.

[...]

Issue #4526, opened 2024-06-03
ci: abci event databases do not persist


conorsch:
**Describe the bug**
The Postgres databases storing ABCI event info for frontend apps like the block-explorer are not persisting. This has resulted in data loss for events on the `penumbra-testnet-deimos-8` chain. Let's fix the persistence problem, then try to reindex historical events (#4525).

**Additional context**
Most likely the problem is a naive application of the CometBFT Postgres schema on application start: if the upstream schema declaration is not idempotent, e.g. via use of "IF NOT EXISTS" then re-applying it is a destructive action, essentially dropping existing tables. If that's the case, let's drop a flag on-disk in the PVC and only apply the schema if that flag is missing.

PR #4575, opened 2024-06-07
ci: fix persistence for indexing dbs

conorsch:
## Describe your changes

Due to an oversight in the volume-mount path, out postgres databases for indexing ABCI events via CometBFT's psql indexer setup were not persisting across node restarts. This was due to the fact that the upstream `postgres` container image declares its own VOLUME at `/var/lib/postgresql/data`, visible here:

    ❯ podman inspect docker.io/postgres:16-bookworm | jq '.[0].Config.Volumes'
    {
      "/var/lib/postgresql/data": {}
    }

which clobbered our manual mount at `/var/lib/postgresql`, a level up. We must override the volume mountpoint in order for the PVC to take precedence over the anonymous volume. Done by adding the fullpath to that exact mount, and then overriding PGDATA to use a subdir therein.

Also pins a stable version of the container image tag, so we don't get surprised by postgres jumps.

Refs #4526. Will make sure this applies cleanly automatically on preview post-merge, then will update the deployments for currently-running networks. Does not handle historical reingestion; that's tracked separately in #4566.

  > ci-only, only updates deployment logic, doesn't touch app code

### DIFF
--- a/deployments/charts/penumbra-node/templates/statefulset.yaml
+++ b/deployments/charts/penumbra-node/templates/statefulset.yaml
@@ -30,6 +30,12 @@ spec:
     {{- if eq .Values.cometbft.config.indexer "psql" }}
     - metadata:
         name: db
+        labels:
+          {{- include "penumbra-node.labels" . | nindent 10 }}
+          "app.kubernetes.io/component": fullnode
+          {{- if .Values.part_of }}
+          "app.kubernetes.io/part-of": {{ .Values.part_of }}
+          {{- end }}
       spec:
         accessModes: [ "ReadWriteOnce" ]
         resources:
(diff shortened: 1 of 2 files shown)


deployments/charts/penumbra-node/templates/statefulset.yaml
```diff
@@ -30,6 +30,12 @@ spec:
     {{- if eq .Values.cometbft.config.indexer "psql" }}
     - metadata:
         name: db
+        labels:
+          {{- include "penumbra-node.labels" . | nindent 10 }}
+          "app.kubernetes.io/component": fullnode
+          {{- if .Values.part_of }}
+          "app.kubernetes.io/part-of": {{ .Values.part_of }}
+          {{- end }}
```
conorsch, line 38: The labels here always should have been present, but weren't. This change won't apply cleanly to running indexers, but that's fine: I need to recreate them anyway after this logic lands. I'll handle that manually.

PR #4576, opened 2024-06-07
pcli: add options to pcli v address for displaying an FVK or generating a payment address from a provided FVK

aubrika:
Adding a couple utility flags to `pcli v address` to assist in FVK verification (i.e. Alice can send Bob an FVK, Bob can send a message in a transaction memo to an address associated with that FVK, Alice can successfully read that memo with her FVK and return the same message to Bob, confirming the round-trip communication)

  > only changes `pcli` output, doesn't submit any activity

### DIFF
--- a/crates/bin/pcli/src/command/view/address.rs
+++ b/crates/bin/pcli/src/command/view/address.rs
@@ -1,8 +1,9 @@
 use anyhow::Result;
 use base64::Engine;
 use rand_core::OsRng;
+use std::str::FromStr;
 
-use penumbra_keys::{Address, FullViewingKey};
+use penumbra_keys::{keys::AddressIndex, Address, FullViewingKey};
 
 #[derive(Debug, clap::Parser)]
 pub struct AddressCmd {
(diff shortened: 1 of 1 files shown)


cronokirby (approved):

conorsch (approved):
Fantastic, thanks!

crates/bin/pcli/src/command/view/address.rs
```diff
@@ -45,7 +52,22 @@ impl AddressCmd {
             } else if self.compat {
                 println!("{}", address.compat_encoding());
             } else {
-                println!("{}", address);
+                if self.fvk {
+                    eprintln!("🔥 CAUTION: POSSESSION OF THE FOLLOWING FULL VIEWING KEY WILL");
+                    eprintln!("🔥 PROVIDE VISIBILITY TO ALL ACTIVITY ON ITS ASSOCIATED ACCOUNTS.");
+                    eprintln!("🔥 DISTRIBUTE WITH CARE!");
+                    eprintln!("");
```
conorsch, line 59: I feel strongly that the `println` messages _before_ emitted the requested FVK value should use `eprintln`. Doing so will ensure that the CLI is composable, e.g. can pipe to other processes and have stdout distinct from stderr. Just pushed up a fix doing same.

PR #4581, opened 2024-06-10
pd: 💤 a sleep worker, tracking scheduler latency

cratelyn:
this introduces a submodule to `pd::metrics`, which observes tokio scheduler latency and records it for the Prometheus exporter.

the core idea in this worker is that it will invoke `tokio::time::sleep(..)` to put the task to sleep for one second. upon being woken up by the scheduler, it will check the *new* wall-clock time and calculate the actual amount of time it spent waiting. if this took longer than expected, it increments the `pd_async_sleep_drift_milliseconds` counter by the observed latency, measured in milliseconds.

as the module-level docs note, this is a very useful tool to identify when the runtime is being disrupted by blocking I/O or other expensive computation.

  > metrics emission

### DIFF
--- a/crates/bin/pd/src/lib.rs
+++ b/crates/bin/pd/src/lib.rs
@@ -5,7 +5,7 @@
 // Requires nightly.
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
 
-mod metrics;
+pub mod metrics;
 
 pub mod cli;
 pub mod migrate;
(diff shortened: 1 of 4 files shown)


conorsch:
Refs #4464, I presume. Very curious to see what this will yield, thanks for taking the time to put it together!

conorsch (approved):
Merging optimistically, will inspect results on preview

Issue #4577, opened 2024-06-07
pcli view db corrupted with invalid anchors


conorsch:
**Describe the bug**
Since launching ceremony 2.1, we've seen a marked upstick in reports of:

> `status: Internal, message: "Error submitting transaction: code 1, log: failed to deliver transaction: check_stateful failed: provided anchor d1b0a17ec504a6c6abd31a4a30aceb934e9d87eb43014ec97fe4a1d86ddc1909 is not a valid SCT root", details: [], metadata: MetadataMap { headers: {} }`

**To Reproduce**
Several team members confirm this is reproducible by simply looping over `pcli-tx-send`. For example, here's an attempt I just submitted:

<details>
<summary>
full terminal session from send-loop
</summary>

```
❯ for i in $(seq 10); do pcli tx send --to $(pcli view address 10 -e) --memo "testing looped pcli sends $i" 0.1penumbra ; done
Scanning blocks from last sync height 438649 to latest height 438649
[0s] ██████████████████████████████████████████████████       0/0       0/s ETA: 0s
building transaction [3 actions, 3 proofs]...
finished proving in 0.547 seconds [3 actions, 3 proofs, 2402 bytes]
broadcasting transaction and awaiting confirmation...
transaction broadcast successfully: 0837b70f240bf405cd299443af16fa12d2a805d6f0849d67b22caf9d45c2a02a
transaction confirmed and detected: 0837b70f240bf405cd299443af16fa12d2a805d6f0849d67b22caf9d45c2a02a @ height 438651
Scanning blocks from last sync height 438651 to latest height 438651
[0s] ██████████████████████████████████████████████████       0/0       0/s ETA: 0s
building transaction [3 actions, 3 proofs]...
finished proving in 0.540 seconds [3 actions, 3 proofs, 2402 bytes]
broadcasting transaction and awaiting confirmation...
transaction broadcast successfully: 3fd02e3404905ca43b0460a6b34ead4524ad20db6d50980e1b67d71f811c5dfe
transaction confirmed and detected: 3fd02e3404905ca43b0460a6b34ead4524ad20db6d50980e1b67d71f811c5dfe @ height 438652
Scanning blocks from last sync height 438652 to latest height 438652
[0s] ██████████████████████████████████████████████████       0/0       0/s ETA: 0s
building transaction [4 actions, 4 proofs]...
finished proving in 0.700 seconds [4 actions, 4 proofs, 2784 bytes]
[...]

PR #4586, opened 2024-06-11
compact-block: try to ensure response consistency

hdevalence:
## Describe your changes

Wraps the mpsc used to transmit compact blocks with a stateful wrapper that tracks the expected next block height and rejects the send if it mismatches. This should act as a guard against any problem where the node misses or duplicates blocks in the response stream, by converting those to internal server errors that cause the stream to terminate and the client to reopen the connection.

Related to #4577

  > rpc changes only


cronokirby (approved):
Looks good to me, also I think both our view service implementations will work with interruptions like this, so.

PR #4588, opened 2024-06-11
4587: Adjust balances to sum like notes on a per-account index basis

### LINKED ISSUE #4587: Fix `Balances` view service API to combine notes
**Describe the bug**
Currently the `Balances` API as-implemented will return an entry per-note, rather than summing all note values.

**Expected behavior**
Since the intent of the `Balances` API is to return _account balances_ and `NotesRequest` API already exists and _correctly_ exposes individual notes, the `Balances` API should sum all notes on a per-account basis and return only a single entry per account/asset ID pair.

**Screenshots**
https://gist.github.com/zbuc/0c7c75c188a6aa0f99815c596033aa79

**Additional Context**
`pcli` uses a different API to retrieve balances (`NotesRequest`). This may account for mismatches in balances seen between `pcli` and prax.

zbuc:
## Describe your changes
This fixes the issue described in #4587 where the `Balances` view service API was in an in-between state between providing per-note balances and per-account index balances. This fixes the API to properly provide per-account index/asset balances (summing all notes on a per account/asset basis).

The only places the `Balances` API are used are in prax and the Hummingbot integration. `pcli` uses a different API to retrieve balances.

Closes #4587 

  > only affects view service

### DIFF
--- a/crates/view/src/storage.rs
+++ b/crates/view/src/storage.rs
@@ -271,7 +271,8 @@ impl Storage {
 
             tracing::debug!(?query);
 
-            let mut entries = Vec::new();
+            // Combine notes of the same asset/address index together
+            let mut balances: BTreeMap<AddressIndex, BTreeMap<asset::Id, Amount>> = BTreeMap::new();
 
             for result in pool.get()?.prepare_cached(query)?.query_map([], |row| {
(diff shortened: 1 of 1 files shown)


cronokirby (approved):

erwanor:
cc @conorsch

Issue #4591, opened 2024-06-12
rustls 0.20 is unmaintained


dynst:
The upstream [Security Policy](https://github.com/rustls/rustls/blob/main/SECURITY.md) for rustls says 0.20.x isn't receiving any updates. The only user of rustls 0.20 is `penumbra-auto-https` (running `cargo tree --inverse` shows everything else depends, directly or indirectly, on 0.21.x).

conorsch:
Closed by https://github.com/penumbra-zone/penumbra/pull/4592. Confirmed on preview that i can use latest-main pcli against the grpc endpoint and the status page is still working well too.

PR #4592, opened 2024-06-12
update rustls 0.20 -> 0.21

dynst:
## Describe your changes

This updates penumbra-auto-https to use rustls 0.21.

#4591

rustls 0.20.x is no longer receiving updates upstream. Updating to rustls 0.21 also consolidates all the penumbra crates to use a single version of rustls, instead of a mix of 0.21 + one holdout on 0.20, compiling two different versions of the same crate.


conorsch (approved):
LGTM, thanks for taking the time. I'm particularly interested in out the auto-https logic behaves with the axum server bump: will monitor deployment to preview post-merge.

PR #4593, opened 2024-06-12
Use friendlier denom rather than asset IDs in error

### LINKED ISSUE #4557: pcli should use friendlier denoms rather than asset IDs when printing error messages
**Describe the bug**
In #4554 @conorsch noted that he had difficulty understanding an error message from the planner returned when there wasn't enough balance to pay fees.

Since validators will primarily have to use `pcli` to manage their validators, it seems worthwhile to use a friendlier display denom in the error message you get when you don't have enough funds to cover planning:

`Error: ran out of notes to spend while planning transaction, need 1 of asset passet1984fctenw8m2fpl8a9wzguzp7j34d7vravryuhft808nyt9fdggqxmanqm`

devanoneth:
## Describe your changes

This improves the error message when a user does not have the required notes to cover a spend.

Closes #4557

  > change only made to error strings

### DIFF
--- a/crates/view/src/planner.rs
+++ b/crates/view/src/planner.rs
@@ -576,19 +576,19 @@ impl<R: RngCore + CryptoRng> Planner<R> {
         }
 
         let mut iterations = 0usize;
+        let asset_cache = view.assets().await?;
 
         // Now iterate over the action list's imbalances to balance the transaction.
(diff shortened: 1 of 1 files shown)


zbuc (approved):
Thanks!

PR #4564, opened 2024-06-05
dex: fold `Position`s into the price indexes

erwanor:
## Describe your changes

This PR:
- fold `Position` records into the DEX price indexes
- remove extra hashing of positions (to compute `Id`s)
- adds a migration script to population the nonconsensus idx

## Checklist before requesting a review

- [x] If this code contains consensus-breaking changes, I have added the "consensus-breaking" label. Otherwise, I declare my belief that there are not consensus-breaking changes, for the following reason:

  > This is consensus breaking, AND requires a migration.

PR #4508, opened 2024-05-30
fix(pd): reset app height in testnet 77 migration

conorsch:
## Describe your changes
Encountered CometBFT failure while testing chain resumption after applying testnet77 migration. Paired with @erwanor to resolve.

## Issue ticket number and link
Refs #4497.

## Checklist before requesting a review

- [x] If this code contains consensus-breaking changes, I have added the "consensus-breaking" label. Otherwise, I declare my belief that there are not consensus-breaking changes, for the following reason:

  > fixes a migration for upcoming release, based on migration testing

Issue #4294, opened 2024-04-30
Refactor `pd` migration code to abstract away boilerplate


aubrika:
Currently the migration code features a lot of copy-pasted functions which are common to every migration, and for the purposes of ergonomics, readability and bug surface this is not ideal. This should be kept as fairly straightforward or simple refactor with a mind to facilitating future automated testing of migrations and enhancing testability.

conorsch:
We're not prioritizing this (see p-low tag), but as we continue to write migrations, there will be ample opportunity to make the code more DRY.

PR #4596, opened 2024-06-12
Minor fixes to testnet 78 migration

zbuc:
## Describe your changes

Configure migrations to run testnet 78 migration, add more logging, fix bug with validator migration

  > consensus breaking because it's a migration

### DIFF
--- a/crates/bin/pd/src/migrate/testnet78.rs
+++ b/crates/bin/pd/src/migrate/testnet78.rs
@@ -1,10 +1,13 @@
 //! Contains functions related to the migration script of Testnet78.
+use anyhow::Context;
 use cnidarium::{Snapshot, StateDelta, Storage};
 use futures::TryStreamExt as _;
 use jmt::RootHash;
 use penumbra_app::app::StateReadExt as _;
 use penumbra_governance::StateReadExt as _;
+use penumbra_governance::StateWriteExt;
 use penumbra_proto::{StateReadProto as _, StateWriteProto as _};
+use penumbra_sct::component::clock::EpochManager;
 use penumbra_sct::component::clock::EpochRead as _;
 use penumbra_stake::validator::Validator;
 use std::path::PathBuf;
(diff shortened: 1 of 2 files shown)


conorsch:
Nice, thanks for following up. I'll dig into migration testing for 78 today!

conorsch:
Looks like the revised migration _applies_ fine, but the chain cannot restart after the migration is run. A common mistake: I did the exact same thing when writing the Testnet77 migration (https://github.com/penumbra-zone/penumbra/pull/4503) and had to follow up with a narrow fix in https://github.com/penumbra-zone/penumbra/pull/4508. 

Unsurprisingly I see that the competing Testnet77 migration in https://github.com/penumbra-zone/penumbra/pull/4564 does _not_ make this same mistake, hat-tip to @erwanor on that. But let's fix it here, so we can proceed with migration testing, and sort out conflicts in #4654 separately.

conorsch (changes requested):
Needs to reset the halt bit: I'll append a commit and re-review after testing.

zbuc:
Oof, good catch. I had refactored the migration code to de-duplicate all the common operations in migrations but then [reverted it](https://github.com/penumbra-zone/penumbra/pull/4567/commits/5f9b10ba8a90208d28a478f08735aa86b2be34d2) and forgot to add those steps back to the testnet 78 migration.

conorsch:
> I had refactored the migration code to de-duplicate all the common operations in migrations but then [reverted it](https://github.com/penumbra-zone/penumbra/pull/4567/commits/5f9b10ba8a90208d28a478f08735aa86b2be34d2) and forgot to add those steps back to the testnet 78 migration.

Thanks for tackling this, even if it didn't land. I remain in favor of cleaning up the migration tooling (#4294), but don't mind writing more migrations as practice to inform what the shape of that cleanup should be.

conorsch (approved):

crates/bin/pd/src/migrate/testnet78.rs
```diff
@@ -161,10 +182,12 @@ async fn truncate_validator_fields(delta: &mut StateDelta<Snapshot>) -> anyhow::
 ///   * Governance Signaling Proposals:
 ///    - `commit hash` (255 bytes)
 async fn truncate_proposal_fields(delta: &mut StateDelta<Snapshot>) -> anyhow::Result<()> {
+    tracing::info!("truncating proposal fields");
     let next_proposal_id: u64 = delta.next_proposal_id().await?;
 
     // Range each proposal and truncate the fields.
     for proposal_id in 0..next_proposal_id {
+        tracing::info!("truncating proposal: {}", proposal_id);
```
erwanor, line 190: @zbuc doesn't this have the same issue of pulling data using stronger rules that existed when it was stored?

zbuc: Yes, will follow up in a new PR

zbuc: https://github.com/penumbra-zone/penumbra/pull/4606

PR #4594, opened 2024-06-12
pd: use 10MB as mempool limit in comet config template

dynst:
## Describe your changes

Lower the default value for `max_txs_bytes` from 1GiB to 10MiB. Upstream also substantially lowered this default recently. 
https://github.com/cometbft/cometbft/pull/2767

See also the [new `nop` mempool type](https://github.com/cometbft/cometbft/pull/1643) they added to deal with the mempool being a potential DoS vector a la [p2p-storms](https://github.com/notional-labs/placid).

## Checklist before requesting a review

- [X] If this code contains consensus-breaking changes, I have added the "consensus-breaking" label. Otherwise, I declare my belief that there are not consensus-breaking changes, for the following reason:

The mempool and its configuration are outside consensus.

Issue #830, opened 2022-05-18
Budget notes per block


plaidfinch:
When adding notes to the TCT, we now have a hard limit on the number of output notes that can be included in a single block: 4^8. Once the TCT is integrated as a consensus-critical part of the chain, this will cause a chain halt if violated.

We should reject transactions in `check_tx_stateful` if they would cause the block to go "overbudget" on outputs. Additionally, we should restrict the number of funding streams per validator to a hardcoded limit, say, 64, to prevent a DoS attack where someone could upload 10,000,000 funding streams for their validator and cause the chain to reject all other transactions. We should also probably place funding stream outputs _first_ in a block, so that they are always definitely included, then add other transactions as possible.

Once we integrate ABCI++ it will be possible to control the inclusion of transactions in a block more precisely, and we should use this capability to allow transactions to linger to the next block in the mempool. In the current solution above, high-contention blocks may end up in a situation where the block proposer could add too many transactions, then some of them would be rejected for bringing the block overbudget, and they would need to be re-submitted.

hdevalence:
We may not need to do this at all, depending on our block sizes.  Each output has a proof (~200bytes minimum), plus a note payload (~200 bytes), plus an encrypted memo (~600 bytes), giving about 1KB of data.  This isn't including any spends or other actions. 

If blocks are capped at, say, 1MB, then they won't contain more than ~1000 outputs each, well below the maximum.

plaidfinch:
Doesn't this mean then that we will still have to budget, but for a different kind of quota? I imagine this is something we can only do effectively with ABCI++?

hdevalence:
Yes, we need ABCI++ for this.

erwanor:
This came up in discussions about adding a mint queue, reminder to myself to link this issue when it gets specified.

hdevalence:
That issue was #3545 ; we can leave it unplanned for now. Other than the genesis case, the reasoning I posted above still applies.

PR #4597, opened 2024-06-12
pd: configure cometbft with `max_bytes` to 1MiB

### LINKED ISSUE #4595: Reduce BlockParams::max_bytes
https://github.com/cometbft/cometbft/pull/1518

CometBFT lowered the default value for BlockParams.MaxBytes from 21MiB to 4MiB, after last year's [ASA-2023-002](https://github.com/cometbft/cometbft/security/advisories/GHSA-hq58-p9mv-338c).

dynst:
## Describe your changes

Reduce `tendermint::block::Size::max_bytes` from 21MiB to 1MiB.

Closes #4595

Consensus parameters are not part of consensus but could affect network liveness.

### DIFF
--- a/crates/bin/pd/src/testnet/generate.rs
+++ b/crates/bin/pd/src/testnet/generate.rs
@@ -287,7 +287,7 @@ impl TestnetConfig {
             consensus_params: tendermint::consensus::Params {
                 abci: AbciParams::default(),
                 block: tendermint::block::Size {
-                    max_bytes: 22020096,
+                    max_bytes: 4194304,
                     max_gas: -1,
                     // minimum time increment between consecutive blocks
                     time_iota_ms: 500,
--- a/crates/bin/pd/src/testnet/generate.rs
+++ b/crates/bin/pd/src/testnet/generate.rs
@@ -287,7 +287,7 @@ impl TestnetConfig {
             consensus_params: tendermint::consensus::Params {
                 abci: AbciParams::default(),
                 block: tendermint::block::Size {
-                    max_bytes: 4194304,
+                    max_bytes: 1048576,
                     max_gas: -1,
                     // minimum time increment between consecutive blocks
                     time_iota_ms: 500,


crates/bin/pd/src/testnet/generate.rs
```diff
@@ -287,8 +287,8 @@ impl TestnetConfig {
             consensus_params: tendermint::consensus::Params {
                 abci: AbciParams::default(),
                 block: tendermint::block::Size {
-                    max_bytes: 22020096,
-                    max_gas: -1,
+                    max_bytes: 4194304,
+                    max_gas: 10_000_000,
```
erwanor, line 291: I don't think we want to set this, the penumbra sdk does not use the cometbft notion of gas at all, so i'm worried setting a limit here would clash with how people running the software expect comet to reap from the mempool.

erwanor (changes requested):
The block limit makes sense to me, I think that's a good catch, thanks a lot. I have questions about the gas limit. I would like to get @conorsch or @hdevalence's go ahead on this too.

crates/bin/pd/src/testnet/generate.rs
dynst: Yeah, although setting it to limitless (-1) also sort of sends the wrong message. Should it just to set to zero? That's a pretty clear self-documenting 'we don't use this' signal.

erwanor: I think it's best to leave it to `-1` unless we understand why it would be bad. As defined it reflects the reality that the application doesn't expect comet to suggest raw proposals based on gas metering. I think setting this to a concrete value will make comet actively use it in along with `max_bytes`. If that's the case, then a value of zero would make comet suggest empty raw proposals

dynst: Appreciate the feedback, got rid of the change to max_gas. Also, turns out [the official docs](https://docs.cometbft.com/v0.38/spec/abci/abci++_app_requirements#states-and-abci-connections) themselves recommend -1 when 'the concept of gas is meaningless'!

crates/bin/pd/src/testnet/generate.rs
```diff
@@ -287,7 +287,7 @@ impl TestnetConfig {
             consensus_params: tendermint::consensus::Params {
                 abci: AbciParams::default(),
                 block: tendermint::block::Size {
-                    max_bytes: 22020096,
+                    max_bytes: 4194304,
```
erwanor, line 290:
```suggestion
                    max_bytes: 1048576,
```

and this will lgtm!

erwanor (approved):

conorsch:
OK, gas changes removed, that's great. I'm definitely on board with fixing the block size, but @erwanor: can you comment as to why 1MB is the "correct" value here? I'd like to cite that in a comment somewhere. Does the 1MB block size impose restrictions on e.g. genesis size?

erwanor:
There's some context in #830, but the short answer is that there's a hard limit on the number of notes that can be recorded in a TCT block and setting a 1MiB limit gives ample room for transactions (e.g. opening 30 LPs replicating an XYK is about 30KB) while being comfortably below the 65k state commitment upper bound

conorsch:
See closely related https://github.com/penumbra-zone/penumbra/pull/4594

dynst:
For future edification, something that wasn't immediately obvious to me: `max_bytes` actually includes the overhead of validator commitments and submitted evidence of double-signing, not just transactions.

https://github.com/cometbft/cometbft/issues/2756#issuecomment-2047145981

> The maximum data (transaction) size of a block depends on the number of validators and evidence byte size, as defined via the [types.MaxDataBytes](https://github.com/cometbft/cometbft/blob/b08546f6f24beb3e155f80d0bdd615ae6ef514a9/types/block.go#L279) method.

> When considering the default maximum block size (4 MiB) and a set of 128 validators, the maximum size of transactions is 4179365 ~ 3.98 MiB. So we can probably consider 4 MiB for simplicity.

Issue #4601, opened 2024-06-12
Spurious error related to "note commitment missing"


cronokirby:
See: https://discord.com/channels/824484045370818580/981304070444814367/1248691717322838118

The basic failure mode involves doing transactions in pcli, such as contributing to the ceremony, and failing to send a transaction because of the error `Note commitment missing`. This seems spurious, in that it doesn't reproduce easily.

We should look into this more, as we've seen several reports about this.

Also, it seems like Galileo was having these errors back in November as well.

cratelyn:
this looks like the same error observed in https://github.com/penumbra-zone/penumbra/issues/4574#issuecomment-2155251849

zbuc:
This is reproducibly happening for me on the current testnet

cratelyn:
there are a collection of errors, all filed around the same time, that relate
to inconsistent behavior when viewing transactions. i took some time to gather
these, and write up my understanding of the current state of affairs:

#4452 reports that the view service's response streams terminate when an error
is encountered. #4465 reports that the view service misses some tx's when
scanning. #4577 describes the view service being corrupted with invalid
anchors.

#4584, #4574, and #4238 all describe issues with `pcli` failing to collect
various information. #4584 reports that pcli fails to
`query dex all-positions`, #4574 reports that pcli fails to `sweep`, and #4238
reports that pcli fails to detect delegator votes.

#4601 describes `"note commitment missing"` errors when making transactions
with pcli, which is the underlying issue behind #4574 in particular.

in response to this, we have already shipped a few changes to address this:
- #4453 avoids recording invalid `Metadata`.
- #4578 adds height checks to the view service.
- #4586 adds height checks to the compact-block server.

these have all shipped in v0.77.3. 

#4605 added test coverage for `pcli sweep`, and #4617 made additional changes
to facilitate exercising pcli in mock consensus tests. #4517 tracks an issue
that soft blocks us from adding more test coverage of the view service, which
caused us to disable that test coverage in #4633.

**at this point i believe that #4601 and #4574 can also be closed via #4578
and #4586.** these added defensive checks to the compact-block RPC and the
view service to help keep syncing deterministic and reliable, which is a
common thread between all of these issues. going forward, we can build out
more assurance of this system by addressing #4517 next.

PR #4578, opened 2024-06-08
View Service: Don't sync the same block twice

cronokirby:
This should fix an issue wherein a load balanced RPC can cause data corruption by delivering the same block twice in the stream.

## Issue ticket number and link

This should close #4577.

## Checklist before requesting a review

- [x] If this code contains consensus-breaking changes, I have added the "consensus-breaking" label. Otherwise, I declare my belief that there are not consensus-breaking changes, for the following reason:

  > Just a client change

Issue #4574, opened 2024-06-07
bug: tx sweep failing on already-spent nullifier


aubrika:
**Describe the bug**

It seems `tx sweep` is currently broken, as reported here in Discord:

https://discord.com/channels/824484045370818580/1172730682783563826/1248651765323137178

zbuc:
I am able to `tx sweep` (mostly) successfully, however @conorsch and I noted that there seems to possibly be intermittent view server corruption leading to errors like this and SCT hash mismatches.

I say (mostly) successfully because I did get a different error...

```
# truncated: 24 successful sweeps

building sweep 24 of 25
building transaction [9 actions, 9 proofs]...
finished proving in 3.136 seconds [9 actions, 9 proofs, 4405 bytes]
broadcasting transaction and awaiting confirmation...
transaction broadcast successfully: 52cf05c1d14fa7791f44f6e24596376262a884e6718b97220c2dbc019bb4c326
transaction confirmed and detected: 52cf05c1d14fa7791f44f6e24596376262a884e6718b97220c2dbc019bb4c326
building sweep 0 of 3
building transaction [11 actions, 11 proofs]...
finished proving in 3.587 seconds [11 actions, 11 proofs, 5459 bytes]
broadcasting transaction and awaiting confirmation...
transaction broadcast successfully: bb88032a4eed65564973dd9aa7956113bf49f6469ca25d781534517f5bfecb76
transaction confirmed and detected: bb88032a4eed65564973dd9aa7956113bf49f6469ca25d781534517f5bfecb76 @ height 434808
building sweep 1 of 3
building transaction [9 actions, 9 proofs]...
Error: status: InvalidArgument, message: "Note commitment missing", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc"} }
```

cratelyn:
x-ref #4601

cratelyn:
#4605 introduced test coverage for this. #4633 disabled that test, due to the issues outlined in #4517. once we have support for an in-memory view service in tests, we can re-enable that.

cratelyn:
for visibility, cross-posting this comment: https://github.com/penumbra-zone/penumbra/issues/4601#issuecomment-2186961948

aubrika:
Closed as complete, should be fixed by #4578

Issue #4517, opened 2024-05-31
view: 🥜 `ViewServer` support for an in-memory gRPC endpoint


cratelyn:
refs:
* #4471
* #3913

mock consensus tests that use a view server must bind an rpc endpoint to a local port. this is unfortunate because it requires acquisition of a global resource, and can over time cause test suites to be flaky / brittle.

this could be abated by adding support for an in-memory gRPC endpoint to the `ViewServer` and its underlying `Worker`.

see https://github.com/penumbra-zone/penumbra/tree/kate/view-server-tests-redux.box-grpc-endpoint-experiment as a starting sketch.

cratelyn:
i observed the test from #4471 flake in CI, after opening #4585. re-running the job worked, fixing this issue will prevent that flake from occurring going forward.

PR #4605, opened 2024-06-12
misc: 🍱 add a mock consensus test exercising `penumbra_wallet::plan::sweep()`

cratelyn:
#### 💭 describe your changes

this introduces a mock consensus test, showing that the `tx sweep` command
works as expected for a simple case, when a wallet contains `SWEEP_COUNT + 1`
notes.

#4601 more directly tracks the issue of the `note commitment missing` error,
which i believe affects more than just sweeping. similarly, #4586 and #4578
fixed the issue of seeing `not a valid SCT root` errors, which could also occur
when sweeping notes, because it is effectively performing the same action of
performing many transactions in a hot loop, sending funds to oneself.

while this might not "fix" #4574 strictly speaking, this does introduce some
foundational test coverage so that we can write more robust regression tests
against the core logic responsible for sweep planning.

* #4574
* #4586
* #4578
* #4577

  > this branch only introduces tests, and makes minor refactors in client-side
  > code.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4418,6 +4418,7 @@ dependencies = [
  "penumbra-transaction",
  "penumbra-txhash",
  "penumbra-view",
+ "penumbra-wallet",
  "prost",
  "rand",
  "rand_chacha",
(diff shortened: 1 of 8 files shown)


cratelyn:
test flaked, see #4517. i'll re-run tests when i'm back at the computer.

cronokirby (approved):

PR #4606, opened 2024-06-13
Migration 78: Avoid deserializing proto types to domain types in migration

zbuc:
## Describe your changes

Don't deserialize proto types to avoid applying new domain validations to old types in migration 78

https://github.com/penumbra-zone/penumbra/pull/4596#discussion_r1637191934


erwanor (approved):
As far as I can tell, this looks gtm, but the migration is big and has to reach deep into the state so i think we'll get more mileage with practical testing.

PR #4610, opened 2024-06-13
cometindex: 📇 add a minimal indexer prototype

### LINKED ISSUE #4604: Deliver minimal indexer

cratelyn:
## Describe your changes

this is a continuation of #4573.

Adds the skeleton of a minimal but extensible indexer. The indexer consumes events from a source (raw event) database and writes to a destination database with compiled / processed tables. The indexer is a library that other binaries import and slot their indexes into.

The `cometindex` library is intended to be generic over any cometbft events. The `pindexer` binary is for penumbra specifically and we can add all our own code to it. This structure mirrors the existing distinction between comet (generic) and the app. The code is intended to be shaped so that it's easy for users to plug in additional indexes. In particular, all of the "meat" of the binary is in the `cometindex::Indexer` type, which builds an indexer and handles all arg parsing. This way, the external interface of the program, how it's invoked, etc, is the same no matter what app views are present, and orchestration code like containers doesn't need to be concerned with them.

Before merging:

- [x] Rename the `Index` trait to `AppView` (it's a view of some application data)
- [x] Rename the `create_tables` method to `init_chain`
- [x] Add polling for ingesting / processing new events after the initial catch-up (500ms?)
- [x] Make sure the `pindexer` `AppView` traits are public so that `pindexer` can be pulled in as a library
- [x] Add an extension trait in the `pindexer` crate that adds an `Indexer::with_default_penumbra_app_views()` method that hooks in all the `AppView` impls, so we can add to it as we fill in impls

Followup:

- [ ] Change `init_chain` to take a `serde_json::Value` and add plumbing to the `Indexer` builder to take a `genesis.json` file in a standardized way so that every instance of the binary doesn't need to implement that -- ideally this is the exact `genesis.json` that was used by comet, and the plumbing in the `Indexer` builder pulls out the `app_state` key and saves it to be able to pass it into each `AppView`. Longer term we could explore other options but for now I think it's simplest if we just make it a required parameter every time.

Closes #4604 

  > New tool outside of consensus


cratelyn:
opened #4611 to track the follow-on work related to `init_chain()`.

cratelyn:
@hdevalence this is ready for review, when you are able to take a look

aubrika (approved):
LGTM!

Issue #4582, opened 2024-06-10
Release Testnet 78, via chain upgrade


conorsch:
# Testnet upgrade

Testnet chain id: `penumbra-testnet-deimos-8` (unchanged)
Release date: 2024-06-25 (?)
Testnet release manager: @conorsch

# Testnet Release Manager Checklist

It's been over a week since the chain upgrade from 76 -> 77 (#4497), so let's schedule another. We plan to ship the entirety of main, to flush out all pending changes. As usual, special attention must be paid to breaking changes, which require writing migrations.

Pre-release:

- [x] Review all merged PRs with tags [`consensus-breaking` and/or `state-breaking`](https://github.com/penumbra-zone/penumbra/pulls?q=is%3Apr+label%3Astate-breaking%2Cconsensus-breaking+created%3A%3E%3D2024-01-01)
- [ ] Enumerate the specific PRs that should be included in the release:
  - [ ] https://github.com/penumbra-zone/penumbra/pull/4567 (and follow ups)
  - [ ] https://github.com/penumbra-zone/penumbra/pull/4564
  - [x] https://github.com/penumbra-zone/penumbra/issues/4602
  - [ ] https://github.com/penumbra-zone/ibc-types/pull/88 
  - [ ] https://github.com/penumbra-zone/penumbra/pull/4661
  - [ ] PR to bump ibc-types version (TK)
  - [ ] Handle the CometBFT block size changes (either via announcement ask or automated config munging) that were made in https://github.com/penumbra-zone/penumbra/pull/4594
- [ ] Confirm migrations exist where required for breaking changes already merged
- [ ] Identify specific test cases for before/after behavior, querying chain state, to ensure that migrations are effective
- [ ] Prepare `upgrade-plan` governance proposal
  - [ ] Decide on block height at which to halt
  - [ ] Double-check you've got bonded stake
  - [ ] Submit proposal
  - [ ] Vote for it, notify validators of same
  - [ ] Confirm proposal passed (default is 24h voting period)

On release day:

- [ ] Draft an announcement for peer review to ensure changes included are comprehensive.
- [ ] Disable testnet deploy workflow, so that chain is not reset
- [ ] Bump the version number and push its tag, via [cargo-release](https://crates.io/crates/cargo-release).
    - [ ] Run `cargo release minor` for a new testnet, or `cargo release patch` for a bugfix. For the latter, make sure you're on a dedicated release branch.
[...]

Issue #3188, opened 2023-10-12
Support custom CometBFT indexing options


conorsch:
**Is your feature request related to a problem? Please describe.**
For use cases like a blocker explorer, we want to configure CometBFT to index events via postgres. Our config template contains appropriate lines:

https://github.com/penumbra-zone/penumbra/blob/7626552c93544e1143228162db27e68dcdd706b4/testnets/cometbft_config_template.toml#L459-L463

Unfortunately, trying to modify those lines via tendermint-config throws an error, because only the kv and null indexer types are supported. As a workaround, one can just `sed` the necessary lines in place, but that's gnarly, since a multi-line replacement is required.

Ideally we'd support something like `pd testnet join --cometbft-indexer psql --cometbft-psql-conn <URL>`. To do so, we must either 1) get upstream tendermint-config to support the psql indexer type; or 2) create a wrapper struct for the cometbft config in our code, store the modified options, and edit the generated TOML on the fly. 

As we work more with the block explorer, we'll develop opinions on how these options should work. For now, `sed` it is.

hdevalence:
This is distinct but related to the possibility of setting custom indexer filters in the `pd` application itself; internally we support regex filtering but this is not exposed in any config.

PR #4614, opened 2024-06-14
migration: update cometbft tx size for t78 migration

conorsch:
## Describe your changes

We've already updated the CometBFT config template, but running nodes on the current testnet will still be using the prior value, generated at time of join. Let's update the value as part of `pd migrate`, logging a warning if we fail to do so.

Refs #4594, #4582.

  > relevant to consensus-breaking logic and how it's handled via migrations

### DIFF
--- a/crates/bin/pd/src/migrate/testnet78.rs
+++ b/crates/bin/pd/src/migrate/testnet78.rs
@@ -1,4 +1,5 @@
 //! Contains functions related to the migration script of Testnet78.
+use anyhow::Context;
 use cnidarium::{Snapshot, StateDelta, StateWrite, Storage};
 use futures::TryStreamExt as _;
 use futures::{pin_mut, StreamExt};
@@ -16,7 +17,10 @@ use penumbra_proto::{StateReadProto, StateWriteProto};
 use penumbra_sct::component::clock::EpochManager;
 use penumbra_sct::component::clock::EpochRead;
 use penumbra_stake::validator::Validator;
+use std::fs::OpenOptions;
+use std::io::Write;
 use std::path::PathBuf;
+use tendermint_config::TendermintConfig;
 use tracing::instrument;
 
 use crate::testnet::generate::TestnetConfig;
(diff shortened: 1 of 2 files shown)


conorsch:
Encountered some errors while testing this. In short, the TendermintConfig struct is stricter about type validation than CometBFT is. For example, `seed_mode = "false"` is accepted by CometBFT, but the struct demands `seed_mode = false`. Worse, an indexer node will fail, due to the `psql` indexer not being supported by tendermint-rs (#3188). Let's log warnings in those instances, rather than allowing the error to percolate up to bailing out on the top-level `pd migrate` action. If that still proves problematic, we can fall back to asking node operators to make the change manually.

crates/bin/pd/src/migrate/testnet78.rs
```diff
@@ -520,8 +524,47 @@ fn truncate(s: &str, max_bytes: usize) -> &str {
     &s[..closest]
 }
 
-mod tests {
+/// Edit the node's CometBFT config file to set mempool.max_tx_bytes to the required value.
+/// This value will affect consensus, but the config setting is specified for CometBFT
```
dynst, line 528:
```suggestion
/// Edit the node's CometBFT config file to set mempool.max_txs_bytes to the required value.
/// This value won't affect consensus, but the config setting is specified for CometBFT
```

dynst, line 528:
```suggestion
/// Edit the node's CometBFT config file to set mempool.max_txs_bytes to the required value.
/// This value won't affect consensus, but the config setting is specified for CometBFT
```

conorsch:
I'm happy with these changes: the cometbft munging will bail out before overwriting if it cannot interpret the on-disk cometbft file, logging a warning. That works well enough with the variety of munged configs I tested against.

crates/bin/pd/src/migrate/testnet78.rs
```diff
@@ -536,8 +540,34 @@ fn truncate(s: &str, max_bytes: usize) -> &str {
...
+#[instrument]
+pub(crate) fn update_cometbft_mempool_settings(cometbft_home: PathBuf) -> anyhow::Result<()> {
+    let cometbft_config_path = cometbft_home.join("config").join("config.toml");
+    tracing::debug!(cometbft_config_path = %cometbft_config_path.display(), "opening cometbft config file");
+    let mut cometbft_config = TendermintConfig::load_toml_file(&cometbft_config_path)
+        .context("failed to load pre-migration cometbft config file")?;
+    // We expect the pre-migration value to be '1073741824', assuming it's unedited from `pd testnet join`.
+    // let pre_migration_value = cometbft_config.mempool.max_txs_bytes;
+    // The new value was updated in GH4594 and represents ~10MB or (2^20 * 10).
+    let post_migration_value = 10485760;
```
erwanor, line 555: There are two `max_tx_bytes` parameters that we want to change:
- the first one is a `mempool` config parameter that controls the maximum size of a transaction to be included in the mempool/gossiped to other nodes 
- the second is a consensus parameter that controls the maximum size of block data (including headers)

We want to change both values:
- the mempool value should be set to something around 30KB (x-ref #4629)
- the consensus value should be set to 1MIB (x-ref #block size #4597)

erwanor (changes requested):

crates/bin/pd/src/migrate/testnet78.rs
dynst: There are two different mempool configuration parameters that are easily mixed up.

- `max_txs_bytes` is what this PR is about
- `max_tx_bytes` is what #4629 is about, though it'd make sense to add it to this PR, too, after #4632 gets merged.

conorsch: The `max_tx_bytes` vs `max_txs_bytes` distinction is certainly ripe for confusion, but if we're changing two values in the CometBFT config for Testnet 78, then we should update both values as part of the migration flow. I'll make a change to this PR to accommodate.

conorsch:
This is ready for review again. After migration, the cometbft config's mempool block should look like:

```
[mempool]
recheck = true
broadcast = true
wal_dir = ""
size = 5000
max_txs_bytes = 10485760
cache_size = 10000
keep-invalid-txs-in-cache = false
max_tx_bytes = 30720
max_batch_bytes = 0
```

I made sure to test with a customized config that set `persistent_peers`, and those changes were preserved post-migration.

PR #4616, opened 2024-06-14
proto: fee view

TalDerei:


TalDerei:
the changes required here are more involved because of the deprecated `fee` field, so it will remain in draft.

aubrika:
Closed as currently out of scope

PR #4621, opened 2024-06-14
auction: dex-enabled check during LP allocation

### LINKED ISSUE #4599: auction: dex-enabled check  ahead of allocating an LP
We should remove the DEX-enabled check in the `ActionDutchAuctionSchedule` handler, and move it ahead of calling the LP allocation routine "`continue` if the dex is disabled":

(https://github.com/penumbra-zone/penumbra/blob/516b1a1e8b61d66beb72ce6b3e899f9320864c30/crates/core/component/auction/src/component/dutch_auction.rs#L240-L241)

That's it. ~~Eventually, we should have typed errors for the `PositionManager` methods, this will let us internalize the checks there and pick and choose which one are fatal at the callsite. That's not the case yet and we shouldn't do that now.~~

erwanor:
## Describe your changes

Close #4599 

  > Consensus breaking, does not require a migration.

### DIFF
--- a/crates/core/component/dex/src/component/position_manager.rs
+++ b/crates/core/component/dex/src/component/position_manager.rs
@@ -81,11 +81,11 @@ pub trait PositionRead: StateRead {
         self.get(&state_key::position_by_id(id)).await
     }
 
-    async fn check_position_by_id(&self, id: &position::Id) -> Result<bool> {
-        Ok(self
-            .get_raw(&state_key::position_by_id(id))
-            .await?
-            .is_some())
+    async fn check_position_by_id(&self, id: &position::Id) -> bool {
+        self.get_raw(&state_key::position_by_id(id))
+            .await
+            .expect("no deserialization errors")
+            .is_some()
     }
 
     async fn best_position(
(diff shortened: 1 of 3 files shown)


crates/core/component/dex/src/component/position_manager.rs
```diff
@@ -81,11 +81,11 @@ pub trait PositionRead: StateRead {
         self.get(&state_key::position_by_id(id)).await
     }
 
-    async fn check_position_by_id(&self, id: &position::Id) -> Result<bool> {
-        Ok(self
-            .get_raw(&state_key::position_by_id(id))
-            .await?
-            .is_some())
+    async fn check_position_by_id(&self, id: &position::Id) -> bool {
```
conorsch, line 84: Why the change to a raw bool return type? With this change, we must use `expect` which will panic rather than return an error which can be handled more gracefully by callers.

erwanor: This is infallible unless the application state is corrupted, in that case there are no way to recover, so we do want to stop everything. So in effect, the `Result` here is superfluous, on the other hand it has a real cost of making the API unwieldy.

zbuc (approved):

PR #4626, opened 2024-06-16
chore: update required version for buf tool

TalDerei:
## Describe your changes
References https://github.com/bufbuild/buf/releases/tag/v1.32.0


conorsch:
Thanks for submitting this, @TalDerei! I tacked on some changes to get this to green. Also refs #4404 which tinkered with the buf cli invocation in the scripts.

cc @cratelyn it'd be grand if we were tracking tooling drift like this in the nix config. I just checked, and naively adding `buf` to the devShell like so:

```diff
diff --git a/flake.nix b/flake.nix
index 2957ecbb3..45e16b96e 100644
--- a/flake.nix
+++ b/flake.nix
@@ -104,7 +104,7 @@
           devShells.default = craneLib.devShell {
             inherit LIBCLANG_PATH ROCKSDB_LIB_DIR;
             inputsFrom = [ penumbra ];
-            packages = [ cargo-watch cargo-nextest protobuf cometbft rocksdb ];
+            packages = [ cargo-watch cargo-nextest protobuf cometbft rocksdb buf ];
             shellHook = ''
               export LIBCLANG_PATH=${LIBCLANG_PATH}
               export RUST_SRC_PATH=${pkgs.rustPlatform.rustLibSrc} # Required for rust-analyzer
```

pulls in buf v1.31.0, i.e. the version before this breaking change. When I check in nix package search, looks like v1.32.0 is only available under "unstable," which is pretty convenient, given that this was a breaking change. If we _did_ want to add 1.32, curious to learn how that'd be specified in the in-repo flake. Not adding it for now, but eager to get the first-run experience up to snuff.

conorsch (approved):

Issue #4022, opened 2024-03-14
Update CometBFT to v0.38.x


conorsch:
We've been using the `v0.37.x` release series of CometBFT since 2023-10 (#4021). We don't yet have compatibility with CometBFT v0.38.x, which implements full support for ABCI++, and requires our app code to add support for VoteExtensions. This issue to track priority of that upgrade.

conorsch:
The current version of CometBFT required for Penumbra, `v0.37.x`, was recently downgraded to high-security updates only, as of 2024-10-01. By EOY it'll be fully EOL. For reference see this [calendar of schedule](https://github.com/cometbft/cometbft/discussions/590#discussioncomment-10669600):

![cometbft-eol-schedule-2024-10](https://github.com/user-attachments/assets/ec4f102f-6c76-48bb-967d-985ab0e9f68b)

We should investigate what's required for to change versions. For instance, do the code changes required require a chain upgrade? Or can existing nodes just upgrade-in-place?

dynst:
It looks like it does require a coordinated upgrade and they don't support a mixed v0.37 + v0.38 network.

https://github.com/cometbft/cometbft/blob/v1.x/UPGRADING.md#v0380

v0.37 is technically EOL but generously did receive that [v0.37.15](https://github.com/cometbft/cometbft/blob/v0.37.x/CHANGELOG.md#v03715) security update in February #5050, which might be its last?

PR #4627, opened 2024-06-17
app: drop stateful mempool

erwanor:
## Describe your changes

This is a draft PR that removes the stateful mempool from the application, replacing it with:
- a stateless mempool that gates inclusion on validity against the latest version of the app state
- a prepare proposal implementation that filters for valid transactions
- a process proposal implementation that validates block quality and lays the groundwork to do optimistic execution

  > Consensus breaking

### DIFF
--- a/crates/core/app/src/app/mod.rs
+++ b/crates/core/app/src/app/mod.rs
@@ -253,6 +253,9 @@ impl App {
         let mut bytes_tracker = 0usize;
 
         for evidence in proposal.misbehavior {
+            // This should be pretty cheap, we allow for `MAX_EVIDENCE_SIZE_BYTES` in total
+            // but a single evidence datum should be an order of magnitude smaller than that.
+            evidence_buffer.clear();
             let proto_evidence: tendermint_proto::v0_37::abci::Misbehavior = evidence.into();
             let evidence_size = match proto_evidence.encode(&mut evidence_buffer) {
                 Ok(_) => evidence_buffer.len(),
--- a/crates/core/app/src/app/mod.rs
+++ b/crates/core/app/src/app/mod.rs
@@ -51,6 +51,9 @@ pub mod state_key;
 /// The inter-block state being written to by the application.
 type InterBlockState = Arc<StateDelta<Snapshot>>;
 
+/// The maximum size of a CometBFT block payload.
+const MAX_BLOCK_PAYLOAD_SIZE_BYTES: u64 = (1 << 20) - 1;
+
 /// The Penumbra application, written as a bundle of [`Component`]s.
 ///
 /// The [`App`] is not a [`Component`], but
@@ -184,7 +187,7 @@ impl App {
         // This is a node controlled parameter that is different from the homonymous
         // mempool's `max_tx_bytes`. Comet will send us raw proposals that exceed that
         // limit, presuming that a subset of those transactions will be shed.
-        // More context in https://github.com/cometbft/cometbft/blob/v0.37.5/spec/abci/abci%2B%2B_app_requirements
+        // More context in https://github.com/cometbft/cometbft/blob/v0.37.5/spec/abci/abci%2B%2B_app_requirements.md
         let max_proposal_size_bytes = proposal.max_tx_bytes as u64;
         // Tracking the size of the proposal
         let mut proposal_size_bytes = 0u64;
@@ -229,12 +232,11 @@ impl App {
             proposal_hash = ?proposal.hash,
             "processing proposal"
         );
-        let max_proposal_size = 1 << 20 - 1;
         let mut total_proposal_size = 0u64;
         for tx in proposal.txs {
             let tx_size = tx.len() as u64;
             total_proposal_size = total_proposal_size.saturating_add(tx_size);
-            if total_proposal_size >= max_proposal_size {
+            if total_proposal_size >= MAX_BLOCK_PAYLOAD_SIZE_BYTES {
                 return response::ProcessProposal::Reject;
             }
 
--- a/crates/core/app/src/app/mod.rs
+++ b/crates/core/app/src/app/mod.rs
@@ -181,20 +181,26 @@ impl App {
             num_candidate_txs
         );
 
-        let mut proposal_size_bytes = 0u64;
+        // This is a node controlled parameter that is different from the homonymous
+        // mempool's `max_tx_bytes`. Comet will send us raw proposals that exceed that
+        // limit, presuming that a subset of those transactions will be shed.
+        // More context in https://github.com/cometbft/cometbft/blob/v0.37.5/spec/abci/abci%2B%2B_app_requirements
         let max_proposal_size_bytes = proposal.max_tx_bytes as u64;
+        // Tracking the size of the proposal
+        let mut proposal_size_bytes = 0u64;
 
-        // First, we filter the proposal to fit within the block limit.
         for tx in proposal.txs {
             let transaction_size = tx.len() as u64;
 
             // We compute the total proposal size if we were to include this transaction.
             let total_with_tx = proposal_size_bytes.saturating_add(transaction_size);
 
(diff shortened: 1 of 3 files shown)


crates/core/app/src/app/mod.rs
```diff
@@ -181,34 +181,35 @@ impl App {
             num_candidate_txs
         );
 
-        let mut proposal_size_bytes = 0u64;
+        // This is a node controlled parameter that is different from the homonymous
+        // mempool's `max_tx_bytes`. Comet will send us raw proposals that exceed that
+        // limit, presuming that a subset of those transactions will be shed.
+        // More context in https://github.com/cometbft/cometbft/blob/v0.37.5/spec/abci/abci%2B%2B_app_requirements
```
dynst, line 187:
```suggestion
        // More context in https://github.com/cometbft/cometbft/blob/v0.37.5/spec/abci/abci%2B%2B_app_requirements.md
```

The link to `abci++_app_requirements.md` added in #3643 was missing the `.md`.

erwanor: Thanks!

crates/core/app/src/app/mod.rs
```diff
@@ -51,6 +51,9 @@ pub mod state_key;
 /// The inter-block state being written to by the application.
 type InterBlockState = Arc<StateDelta<Snapshot>>;
 
+/// The maximum size of a CometBFT block payload.
+const MAX_BLOCK_PAYLOAD_SIZE_BYTES: u64 = (1 << 20) - 1;
```
dynst, line 55: What happens if a chain votes to raise block `max_bytes` to a value larger than 1MiB? This hardcoded value just doesn't allow larger blocks, right?

erwanor: That's right, we don't have a flow to change consensus parameters anyway

crates/core/app/src/app/mod.rs
```diff
@@ -181,34 +184,35 @@ impl App {
             num_candidate_txs
         );
 
-        let mut proposal_size_bytes = 0u64;
+        // This is a node controlled parameter that is different from the homonymous
+        // mempool's `max_tx_bytes`. Comet will send us raw proposals that exceed that
+        // limit, presuming that a subset of those transactions will be shed.
```
dynst, line 189:
```suggestion
        // This is a node controlled parameter that is different from the homonymous
        // mempool's `max_tx_bytes`. Comet will send us raw proposals that exceed this
        // limit, presuming that a subset of those transactions will be shed.
```

'that limit' sounded like it was referring to mempool's homonymous parameter again.

crates/core/app/src/app/mod.rs
```diff
@@ -51,6 +51,15 @@ pub mod state_key;
...
 type InterBlockState = Arc<StateDelta<Snapshot>>;
 
+/// The maximum size of a CometBFT block payload (1MB)
+const MAX_BLOCK_TXS_PAYLOAD_BYTES: usize = 1024 * 1024;
+
+/// The maximum size of a single individual transaction (30KB).
+const MAX_TRANSACTION_SIZE_BYTES: usize = 30 * 1024;
+
+/// The maximum size of the evidence portion of a block (30KB).
+const MAX_EVIDENCE_SIZE_BYTES: usize = 30 * 1024;
```
conorsch, line 61: Thank you for taking the time to constify these values, so it's clearer that they're constant for a reason.

crates/core/app/src/app/mod.rs
```diff
@@ -181,47 +190,104 @@ impl App {
             num_candidate_txs
         );
 
-        let mut proposal_size_bytes = 0u64;
+        // This is a node controlled parameter that is different from the homonymous
+        // mempool's `max_tx_bytes`. Comet will send us raw proposals that exceed this
+        // limit, presuming that a subset of those transactions will be shed.
+        // More context in https://github.com/cometbft/cometbft/blob/v0.37.5/spec/abci/abci%2B%2B_app_requirements.md
```
conorsch, line 196: Explicitly referring to the overloaded terminology here is greatly clarifying. :+1:

crates/core/app/src/server/consensus.rs
```diff
@@ -162,15 +162,22 @@ impl Consensus {
...
     }
 
     async fn process_proposal(
         &mut self,
         proposal: request::ProcessProposal,
     ) -> Result<response::ProcessProposal> {
         tracing::info!(height = ?proposal.height, proposer = ?proposal.proposer_address, hash = %proposal.hash, "processing proposal");
-        Ok(self.app.process_proposal(proposal).await)
+        // We process the propopsal in an isolated state fork. Eventually, we should cache this work and
+        // re-use it when processing a `FinalizeBlock` message (starting in `0.38.x`).
```
conorsch, line 178: Appreciate the forward-looking comment (https://github.com/penumbra-zone/penumbra/issues/4022).

conorsch (approved):

conorsch:
Really solid work here. Especially grateful for the careful commenting throughout, as this work builds on a lot of recent and rather subtle refinements to the constraints on message sizes acceptable to the app.

conorsch:
Not merging because review from @hdevalence is still outstanding. Once it's got final approval, @erwanor, let's use the new squash-with-commit-comments-preserved workflow we've been discussing to get it in!

crates/core/app/src/app/mod.rs
```diff
@@ -51,6 +51,15 @@ pub mod state_key;
 /// The inter-block state being written to by the application.
 type InterBlockState = Arc<StateDelta<Snapshot>>;
 
+/// The maximum size of a CometBFT block payload (1MB)
+const MAX_BLOCK_TXS_PAYLOAD_BYTES: usize = 1024 * 1024;
+
+/// The maximum size of a single individual transaction (30KB).
+const MAX_TRANSACTION_SIZE_BYTES: usize = 30 * 1024;
```
hdevalence, line 58: Did we ever choose these in a principled way? Should they be parameters rather than hardcoded constants?

crates/core/app/src/server/mempool.rs
```diff
@@ -12,21 +12,10 @@ use tracing::Instrument;
...
-/// then reject the second transaction against the forked state. When we learn a
-/// new state has been committed, we discard and recreate the ephemeral fork.
-///
-/// After switching to ABCI++, we can eliminate this mechanism and just build
-/// blocks we want.
+/// A mempool service that applies transaction checks against an isolated application fork.
 pub struct Mempool {
     queue: mpsc::Receiver<Message<Request, Response, tower::BoxError>>,
-    app: App,
+    snapshot: Snapshot,
```
hdevalence, line 18: Why do we want the mempool to hold on to a `snapshot`? Previously, it was stateful and then it held a forked `App` state that was periodically reset, but do we need any of that now? What if the mempool just has a `Storage` handle and pulls the latest snapshot on every request?

hdevalence (approved):
I think we should consider making the constants be parameters, but I'm good to merge either way.

crates/core/app/src/app/mod.rs
erwanor: This was picked with three things in mind:
- large enough to allow JIT liquidity for multiple pairs
- large enough to allow a ranged collection of positions
- large enough to allow multiple IBC client updates
- small enough to allow many of those actions to be performed in a single block even in the extreme case where each transaction is one of those things

Making them parameters make sense, I didn't do it because it would require more work with `penumbra-governance` and hooking them into https://docs.rs/tendermint/latest/tendermint/consensus/params/struct.Params.html

crates/core/app/src/app/mod.rs
dynst: This particular helpful comment about PrepareProposal's `max_tx_bytes` was actually added in a January PR. #3643

PR #4558, opened 2024-06-04
ibc: remediate remaining monorepo issues from zellic audit

avahowell:
Closes #4549
Closes #4552
Closes #4553 
Closes #4550 
Closes #4546

PR #4630, opened 2024-06-17
fix(pd): adjust chain id randomizer for devnets

### LINKED ISSUE #4552: ibc: Trailing hex digits in chain IDs may be treated as revisions
**Describe the bug**
If preserve_chain_id is false when running `pd testnet generate`, theg enerated testnet will have a name ending in 8 hex nybbles, which can be parsed as a decimal approximately 2.3% of the time, which may be interpreted as a revision.

This is the [pd executable](https://github.com/penumbra-zone/penumbra/blob/c462e9a6cc86a37d42a9d75f1ab9ceb0e3789f77/crates/bin/pd/src/main.rs#L426-L429)s generation of hex testnet names.

**Expected behavior**
This likely does not have any security impact, but it is a deviation from the specification and can potentially lead to an issue in the future.

We should either append a -0 to the chain ID or encode the random value as decimal (so that it is consistently treated like a revision).

conorsch:
## Describe your changes
We automatically append a hex randomizer to devnet/preview chain ids, to avoid collisions. Client tooling may expect that suffix to be a revision number, a behavior we want to avoid. Let's stick an 'x' in the randomizer to prevent this behavior.

Follow-up to #4558. Revisits and therefore closes #4552.

  > no changes to app logic, only affects generation of new test chains

### DIFF
--- a/crates/bin/pd/src/main.rs
+++ b/crates/bin/pd/src/main.rs
@@ -329,7 +329,9 @@ async fn main() -> anyhow::Result<()> {
                     let randomizer = OsRng.gen::<u32>();
                     let chain_id =
                         chain_id.unwrap_or_else(|| env!("PD_LATEST_TESTNET_NAME").to_string());
-                    format!("{}-{}", chain_id, randomizer)
+                    // We insert an 'x' in the randomized hex string to ensure it's not parsed as a
+                    // revision id.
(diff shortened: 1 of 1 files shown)


avahowell (approved):
LGTM

PR #4631, opened 2024-06-17
stake: tighten downtime jail parameters to 6h40m

dynst:
## Describe your changes

Allowing 9,500 out of 10,000 blocks to go unsigned is incredibly lax and lets a downed validator linger in the active set for over 13 hours.

Instead of using values from the Cosmos Hub, this PR adapts values from a DEX appchain, Osmosis. They're moving to 1.5-second blocks, and validators get jailed if less than 80% of blocks are signed in an 80,000 block window. Adapting from 1.5s to Penumbra's 5s blocks, equivalent parameters would be:

- `signed_blocks_window_len: 24000` (33h20m window)
- `missed_blocks_maximum: 4800` (6h40m downtime)

  > changes the default value for two chain parameters, which can only be changed with a proposal vote on a live chain

### DIFF
--- a/crates/core/component/stake/src/params.rs
+++ b/crates/core/component/stake/src/params.rs
@@ -71,9 +71,9 @@ impl Default for StakeParameters {
         Self {
             unbonding_delay: 719 * 3 + 1,
             active_validator_limit: 80,
-            // Copied from cosmos hub
-            signed_blocks_window_len: 10000,
-            missed_blocks_maximum: 9500,
+            // Adapted from Osmosis. Jail iff 6h40m downtime in the last 33 hours.
+            signed_blocks_window_len: 24000,
+            missed_blocks_maximum: 4800,
             // 1000 basis points = 10%
             slashing_penalty_misbehavior: 1000_0000,
             // 1 basis point = 0.01%


erwanor:
This is a good suggestion, but I don't think we should change the recommended defaults in the software without data to support that the defined settings are problematic in the first place.

In particular, it's not completely clear there the tangible downside (double the drag on storage for uptime tracking, more punitive uptime req) outweighs the upside (more reactive active set). At a high-level, 13 hours feels reasonable to start with considering this is experimental software. But chains running the penumbra software can always coordinate to raise their own uptime requirement, even without changing the inlined defaults in the software.

erwanor:
Closing for now, but thanks for your effort and the useful data point.

dynst:
> But chains running the penumbra software can always coordinate to raise their own uptime requirement,

Fair! On this point though, is resizing the block window dynamically with a parameter vote supported at all right now?

https://github.com/penumbra-zone/penumbra/blob/2034d4fb7597061a0303a2257dd279f2b40417d2/crates/core/component/stake/src/uptime.rs#L24C7-L26C19

If the current design is reading, deserializing, then reserializing and writing, the entire 10,000-block window to storage every block, instead of a linked-list-ish design, that does seem like a bigger window would mean much more drag on I/O.

Issue #4629, opened 2024-06-17
pd: pick a smaller value for `mempool.max_tx_bytes`


dynst:
In the cometbft config template, maximum individual transaction size ([`max_tx_bytes`](https://github.com/penumbra-zone/penumbra/blob/1a19087c2f006d4672ac8b7dd7cb17f78fad65b4/testnets/cometbft_config_template.toml#L307C1-L309C23)) should probably be smaller than the maximum block size (the consensus parameter) that just got lowered in #4597. What's a reasonable value?

erwanor:
We need empirical data to set a judicious default in the template, so we can revisit later, but a safe upper bound should be ~30KB. This gives ample room for a user to open a collection of replicating LPs and roll block-scoped orders.

PR #4632, opened 2024-06-17
template(comet): lower mempool.max_tx_bytes from 1MiB to 30KiB

dynst:
## Describe your changes

Set the individual transaction size limit to be smaller than the entire block size, but to a still lax, permissive value.

#4629

  > the mempool is outside consensus; and this only edits a template file


conorsch:
Merging to land the config change, will follow up with migration scripting in #4614.

conorsch (approved):

PR #4637, opened 2024-06-17
pd: review pass on config template defaults

erwanor:


conorsch:
So far so good. Thanks for taking the time to annotate the default values with descriptive comments. Standing by to provide final review, after which I'll incorporate any changed values into the migration script for #4614.

erwanor:
I will follow up with putting the template defaults into `const`s that can be pulled from the app crate.

PR #4638, opened 2024-06-17
ibc: fix timeouts: use client height and timestamp

### LINKED ISSUE #4634: ibc: packet timeouts elapse, but withdrawn funds aren't refunded
**Describe the bug**
We believe that the Penumbra IBC implementation is not honoring packet timeouts on chain A messages. 

**To Reproduce**

There are two ways to reproduce this behavior. The most comprehensive is to run ICT on the Penumbra integration branches and view the test failure: TK

A lighter approach is to test against the Testnet 77 (which has a Hermes relayer running on it):

1. Craft a withdrawal with an intentionally low timeout-timestamp: `pcli tx withdraw --channel 8 --timeout-timestamp 5 100gn --to osmo1dlp27fw9ecdk6rwlxq4pqtjycheundlt09h7u7`
2. Observe that balance is decremented by the withdrawn value, i.e. `100gn` in the command above.
3. Wait a few minutes, check balance again, confirm that the withdrawn value remains decremented, when it should be refunded.

**Expected behavior**
I expect that if I packet timeout is triggered on an IBC withdrawal out of Penumbra, then the withdrawn funds are refunded. That's not what we're seeing.

**Additional context**
This problem was reported by @jtieri of SL, in the course of writing Penumbra integration for interchaintest.

avahowell:
Previously, when processing timeouts, we used the `client_processed_height` and `client_processed_timestamp` data to determine if a timeout duration had elapsed:

https://github.com/penumbra-zone/penumbra/blob/c9d778c0ad573b0aabfe4f3d7334eaa9778185e5/crates/core/component/ibc/src/component/msg_handler/timeout.rs#L70

However, those refer to a height and timestamp *on penumbra* when any client update for that client had last been processed:

https://github.com/penumbra-zone/penumbra/blob/c9d778c0ad573b0aabfe4f3d7334eaa9778185e5/crates/core/component/ibc/src/component/client.rs#L168

This comparison is invalid, we should instead be checking against the timestamp and height of the last consensus state from the counterparty client. This PR does that. This was verified to fix the interchain test failures in #4634 .

Closes #4634

### DIFF
--- a/crates/core/component/ibc/src/component/msg_handler/timeout.rs
+++ b/crates/core/component/ibc/src/component/msg_handler/timeout.rs
@@ -66,12 +66,12 @@ impl MsgHandler for MsgTimeout {
             .get_verified_consensus_state(&client_state.latest_height(), &connection.client_id)
             .await?;
         let last_update_time = last_consensus_state.timestamp;
-        let last_update_height = client_state.latest_height();
+        let proof_update_height = self.proof_height_on_b;
 
         // check that timeout height or timeout timestamp has passed on the other end
         if !self
             .packet
-            .timed_out(&last_update_time.into(), last_update_height)
+            .timed_out(&last_update_time.into(), proof_update_height)
         {
             anyhow::bail!("packet has not timed out on the counterparty chain");
         }
--- a/crates/core/component/ibc/src/component/msg_handler/timeout.rs
+++ b/crates/core/component/ibc/src/component/msg_handler/timeout.rs
@@ -61,13 +61,18 @@ impl MsgHandler for MsgTimeout {
             .context("failed to get connection")?
             .ok_or_else(|| anyhow::anyhow!("connection not found for channel"))?;
 
-        let chain_ts = state
-            .get_client_update_time(&connection.client_id, &self.proof_height_on_b)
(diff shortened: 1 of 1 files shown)


conorsch:
Upon further discussion, we are confident that this logic is correct and should be merged. However, we suspect there may still be follow-up required: there are two types of relevant timeouts here, height timeouts and timestamp timeouts. The IBC withdrawal struct requires that clients specify both: https://buf.build/penumbra-zone/penumbra/docs/main:penumbra.core.component.ibc.v1#penumbra.core.component.ibc.v1.Ics20Withdrawal so we should have two test cases in ICT, ensuring that each timeout type is exercised. We've notified @jtieri of this and he's working on it now.

conorsch:
> However, we suspect there may still be follow-up required: there are two types of relevant timeouts here, height timeouts and timestamp timeouts.{

@avahowell  if you have more thoughts on this, please make sure it's tracked in an issue, so we can follow up completely. Any further changes to the IBC packet handling will be consensus-breaking.

conorsch (approved):
Merging. As stated in comments, we may have more related changes once we perform further testing via ICT.

PR #4639, opened 2024-06-17
metrics: add os-level cpu stats to pd

conorsch:
## Describe your changes
Pulls in the `metrics-process` crate [0] to bolt on CPU usage and other OS-level info to the metrics emitted by pd. Doing so will support a better first-run experience for node operators who lack a comprehensive pre-existing setup for monitoring node health, so they can get a sense of how much load pd is under.

[0] https://docs.rs/metrics-process/2.0.0/metrics_process/index.html

None, was discussed briefly during sprint-planning recently. 

  > metrics-only, doesn't change app logic

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -798,6 +798,26 @@ dependencies = [
 ]
 
 [[package]]
+name = "bindgen"
+version = "0.69.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a00dc851838a2120612785d195287475a3ac45514741da670b735818822129a0"
+dependencies = [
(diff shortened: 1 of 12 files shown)


conorsch:
Still need to cobble together some graphs to display this new info, should be straightforward. Draft in the meantime.

conorsch:
Ready for review. Added a new dashboard to focus on pd performance:

new-charts

I'm not happy about how to translate the `pd_process_cpu_seconds_total` metric into a meaningful value, but that needn't block merge: we'll do that in grafana after we have actual data to work with.

cratelyn (approved):
✨ very nice!

Issue #4328, opened 2024-05-06
Swap alt fee tokens into the native token prior to fee burn


hdevalence:
As described in #4320, the initial implementation of multi-asset fee support burns the alt fee tokens. This is incorrect, since the chain should only be minting/burning native assets. The correct behavior is that the fees paid in alt fee tokens should be converted to the native token.  However, since the DEX is batched, we need some plumbing to collect fee amounts, include them in batch swaps, and then handle the resulting fee payments. This overlaps with work we'd thought about (but haven't done, and deferred) as part of #3666 and we should think about them at the same time.

zbuc:
> However, since the DEX is batched, we need some plumbing to collect fee amounts, include them in batch swaps, and then handle the resulting fee payments.

Should we add an additional fee multiplier to fees paid in the non-native asset to account for the additional work of swapping?

hdevalence:
I don't think so, since that cost isn't really attributable to a psecific transaction. 

I'm also not sure it's necessary. In this V0 design, there's no connection between the different prices, all of which are fixed -- so the values of the different fee tokens could float around and cause pricing disparities. To address this, I think that the fixed fees for alternative gas tokens should be significantly higher (10x). This way, users can come into the chain and use it, but they're incentivized to use the native fee token which will protect their privacy.

hdevalence:
The issue here is primarily that we don't currently have any plumbing set up to move the values of the fees around and then swap them, and doing that will require some design work. Maybe we don't get to doing this pre-mainnet.

aubrika:
Sounds like we need to discuss & clarify which part of this ticket is in scope for V1 - @hdevalence when you say "maybe we don't get to doing this pre-mainnet", do you mean this ticket generally? Will burning the alt fee tokens cause issues for users?

TalDerei:
what does it mean if this ticket is tagged as `v1  and removed from the current sprint, specifically in the context of the upcoming code freeze?

PR #4643, opened 2024-06-19
dex: Swap alt fee tokens for native fee token

hdevalence:
## Describe your changes

Moves fee handling out of the Transaction's ActionHandler and into
the component, which accumulates fees (in any token), and adds hooks to the DEX
thatadd chain-submitted swaps for any non-native fee tokens into the native
token. These are then accumulated back into the fee component.

The base fee and tip are tracked separately through this whole process, in
order to support the intended behaviour (the base fee is burned to account for
the _chain_'s resource use, the tip is sent to the proposer to encourage them
to include transactions in fee priority order). However, tip handling is not
currently implemented, so both are burned.  Later, the funding/distribution
components should extract the tip and send it to the proposer's funding
streams. However, until a robust fee market develops, this form of proposer
incentivization can be deferred. The accounting will be there for it when it
arises.

#4328 

  > consensus-breaking but not state-breaking, no migrations necessary


hdevalence:
Testing:

- [ ] Exercise on a devnet and check that behavior works
- [ ] Add state machine tests that a (non-DEX, say ordinary transfer) tx that pays an alt fee token causes a swap to occur, and that the resulting data is as expected
- [ ] Add DEX VCB tests that check VCB works as expected when there are multiple swaps per block.

crates/core/component/fee/src/component.rs
```diff
@@ -1,21 +1,26 @@
...
+use penumbra_proto::state::StateWriteProto as _;
 use tendermint::abci;
 use tracing::instrument;
+
+pub use fee_pay::FeePay;
 pub use view::{StateReadExt, StateWriteExt};
 
 // Fee component
-pub struct Fee {}
+pub struct FeeComponent {}
```
hdevalence, line 20: renamedbecause unlike elsewhere we have an otherwise conflicting Fee struct

crates/core/component/dex/src/component/dex.rs
```diff
@@ -83,6 +133,57 @@ impl Component for Dex {
                 .expect("handling batch swaps is infaillible");
             metrics::histogram!(crate::component::metrics::DEX_BATCH_DURATION)
                 .record(batch_start.elapsed());
+
+            bsods.insert(trading_pair, bsod);
+        }
+
+        // F.1. Having performed all batch swaps, "claim" the base fees and tips.
+        // The VCB has already been debited through the BSOD.
```
erwanor, line 141: ACK - VCB debit.

crates/core/component/dex/src/component/dex.rs
```diff
@@ -83,6 +133,57 @@ impl Component for Dex {
...
+                    .expect("bsod should be present for chain-submitted swap");
+
+                let (base_input, tip_input) = if pair.asset_1() == *asset_id {
+                    ((*base_fee, 0u64.into()), (*tip, 0u64.into()))
+                } else {
+                    ((0u64.into(), *base_fee), (0u64.into(), *tip))
+                };
+
+                let base_output = bsod.pro_rata_outputs(base_input);
+                let tip_output = bsod.pro_rata_outputs(tip_input);
```
erwanor, line 165: This is scary because we don't have a delta vector type that enforce trading pair ordering. ACK though.

crates/core/component/dex/src/component/dex.rs
```diff
@@ -83,6 +133,57 @@ impl Component for Dex {
...
+        {
+            let state_ref =
+                Arc::get_mut(state).expect("should have unique ref after finishing batch swaps");
+            for (asset_id, (base_fee, tip)) in base_fees_and_tips.iter() {
+                if *asset_id == *STAKING_TOKEN_ASSET_ID {
+                    // In this case, there was nothing to swap, so there's nothing
+                    // to claim and we just accumulate the fee we took back into the fee component.
+                    state_ref.raw_accumulate_base_fee(Fee::from_staking_token_amount(*base_fee));
+                    state_ref.raw_accumulate_tip(Fee::from_staking_token_amount(*tip));
+                    continue;
```
erwanor, line 151: ACK: this is sound because we zeroed out the tip accumulator (regardless of whether it was a staking token or not)

crates/core/component/dex/src/component/dex.rs
```diff
@@ -83,6 +133,57 @@ impl Component for Dex {
...
+                    ?base_output,
+                    ?tip_output,
+                    "claiming chain-submitted swap for alt fee token"
+                );
+
+                // Obtain the base fee and tip amounts in the native token, discarding any unfilled amounts.
+                let (swapped_base, swapped_tip) = if pair.asset_1() == *asset_id {
+                    (base_output.0, tip_output.0)
+                } else {
+                    (tip_output.1, base_output.1)
```
erwanor, line 179: I don't understand why this isn't:

```rust
let (swapped_base, swapped_tip) = if pair.asset_1() == *asset_id {
    // `0: asset_id, 1: staking token`
    (base_output.1, tip_output.1)
} else {
    // `0: staking token, 1: asset_id`
    (base_output.0, tip_output.0)
}
```

erwanor: We want to record the staking token output right?

erwanor (commented):
Left a few comments and ACK, will test this later (pairing on testing would be cool if there are takers)

crates/core/component/dex/src/component/dex.rs
zbuc: It does seem worthwhile making it clear that the ordering matters here, i.e. the dex component _must_ execute after anything that accumulates fees, otherwise the fee swapping won't happen.

hdevalence: However, we are guaranteed about that because the stuff that pays fees is transaction processing and we are now in end_block. So the only guarantee we need is that dex end block occurs before fee end block

crates/core/component/dex/src/component/dex.rs
hdevalence: Yeah this seems bugged

zbuc:
What should happen if there isn't enough liquidity to perform a swap for the alternative fee token? Currently it will be burned, which seems fine. What about when the funding component is implemented; will tips be paid out in alternative assets if there isn't liquidity available to swap them?

erwanor:
I'm working on this, and will post updates on testing

erwanor:
Tested on a devnet earlier today, I was able to observe:
- alt asset fee burn against an empty dex
- base fee swap correctly calculated
- base fee + tip swap correctly calculated
- corresponding VCB debit/credits that match pre-swap / post-swap values

I am close to done with a SM test that prove that a multiswap block is accounted for correctly by the VCB mechanism. #4645 is a WIP with SM test of a spend transaction generating alt fees that get swapped.

erwanor (approved):
We are still working on assurance. However, for the sake of not blocking release readiness I am in favor of merging this as-is once it has collected an additional ACK. We commit to following-up with application tests shortly.

TalDerei:
> We are still working on assurance. However, for the sake of not blocking release readiness I am in favor of merging this as-is once it has collected an additional ACK. We commit to following-up with application tests shortly.

I haven't sufficiently tested to ACK, but I support optimistic merging if you have and I'll flag if I discover something.

crates/core/component/dex/src/component/dex.rs
```diff
@@ -83,6 +133,59 @@ impl Component for Dex {
...
+                    (base_output.1, tip_output.1)
+                } else {
+                    // and vice-versa. `R_1` contains native tokens.
+                    (base_output.0, tip_output.0)
+                };
+
+                // Finally, accumulate the swapped base fee and tip back into the fee component.
+                // (We already took all the fees out).
+                state_ref.raw_accumulate_base_fee(Fee::from_staking_token_amount(swapped_base));
+                state_ref.raw_accumulate_tip(Fee::from_staking_token_amount(swapped_tip));
```
cronokirby, line 187:
```suggestion
                // Obtain the base fee and tip amounts in the native token, discarding any unfilled amounts.
                let ((swapped_base, unswapped_base), (swapped_tip, unswapped_tip)) = if pair.asset_1() == *asset_id {
                    // If `asset_id` is `R_1` we want to pull the other leg of the pair.
                    ((base_output.1, base_output.0), (tip_output.1, tip_output.0))
                } else {
                    // and vice-versa. `R_1` contains native tokens.
                    ((base_output.0, base_output.1), (tip_output.0, tip_output.1))
                };

                // Finally, accumulate the swapped base fee and tip back into the fee component.
                // (We already took all the fees out).
                state_ref.raw_accumulate_base_fee(Fee::from_staking_token_amount(swapped_base));
                state_ref.raw_accumulate_tip(Fee::from_staking_token_amount(swapped_tip));
                state_ref.raw_accumulate_base_fee(Fee(Value { amount: unswapped_base, asset_id: *asset_id }));
                state_ref.raw_accumulate_tip(Fee(Value { amount: unswapped_tip, asset_id: *asset_id }));
```

What if we also accumulated the unswapped portion?

crates/core/component/fee/src/component.rs
```diff
@@ -35,12 +40,27 @@ impl Component for Fee {
...
+    #[instrument(name = "fee", skip(state, _end_block))]
     async fn end_block<S: StateWrite + 'static>(
-        _state: &mut Arc<S>,
+        state: &mut Arc<S>,
         _end_block: &abci::request::EndBlock,
     ) {
-        // TODO: update gas prices here eventually
+        let state_ref = Arc::get_mut(state).expect("unique ref in end_block");
+        // Grab the total fees and use them to emit an event.
+        let fees = state_ref.accumulated_base_fees_and_tips();
```
cronokirby, line 50: If we don't take the fees out of the state here, won't they be present in the next block, causing fees to just accumulate over time?

cronokirby (changes requested):
Looks good, main thing I'm unsure about is whether or not we correctly handle the accumulated state of fees at an end of a block:
- Do we want to keep the unswapped amounts over to the next block, to try swapping again?
- Do we need to clear the amounts, or is the intention to accumulate them forever, so that the event is the total fees since the start? (I don't think that's the case)

crates/core/component/fee/src/component.rs
erwanor: This is using ephemeral storage, so the tally is discarded after the block finish executing.

hdevalence:
On keeping the unwrapped fees: I don't think so. That case shouldn't happen, now or in the future:

* Now, when there is a governance-controlled set of a few alt fee tokens, we expect those tokens to be on pairs for which there is liquidity. And since the list is known, it creates its own incentive — someone can make a tiny position on each pair with a nearly infinite price knowing that it would act as a position of last resort for any trades.

* In a future where fee token eligibility is automatic, it would be based on liquidity heuristics anyways.

hdevalence:
On clearing the accumulated fees: this is supposed to act as a defense measure against value creation bugs, by "taking" the value out before swapping it.

crates/core/component/fee/src/component.rs
hdevalence: Also the fees have to be taken out to prevent duplication of value

crates/core/component/dex/src/component/dex.rs
hdevalence: No, the goal is to collapse all the user supplied fee amounts to the native token

cronokirby (approved):

erwanor:
This contains the VCB repro, app test to follow shortly.

PR #4644, opened 2024-06-19
ibc: delete packet commitments instead of writing empty values

avahowell:
The IBC spec calls for deletion of packet commitments after a successful acknowledgement, in order to avoid double-acknowledgements.

Previously, our implementation accomplished this deletion by writing an empty byte slice vec![] to the commitment path.

However, this will cause nonexistence proofs by exclusion (the standard NX-proof type used in ics23) of adjacent keys to fail, since ics23 does not support existence proofs of empty values. We observed this when timeouts from osmosis testnet failed to verify the NX proofs from penumbra testnet.

This PR changes `delete_packet_commitment` to actually delete the commitment, and introduces a migration for testnet 78 that will delete all of the previously overwritten commitments.

### DIFF
--- a/crates/bin/pd/src/migrate/testnet78.rs
+++ b/crates/bin/pd/src/migrate/testnet78.rs
@@ -1,5 +1,6 @@
 //! Contains functions related to the migration script of Testnet78.
 use anyhow::Context;
+use cnidarium::StateRead;
 use cnidarium::{Snapshot, StateDelta, StateWrite, Storage};
 use futures::TryStreamExt as _;
 use futures::{pin_mut, StreamExt};
@@ -49,6 +50,7 @@ use crate::testnet::generate::TestnetConfig;
 ///   * Governance Signaling Proposals:
 ///    - `commit hash` (255 bytes)
 /// - Close and re-open all *open* positions so that they are re-indexed.
+/// - Update DEX parameters
 #[instrument]
 pub async fn migrate(
     storage: Storage,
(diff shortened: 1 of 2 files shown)


erwanor (approved):
ACK on the change, and the migration looks correct too.

To test this, we can:

1. Halt a full node running the current chain
2. Perform the migration (using `--force`)
3. Start the node again with peering disabled
4. Run `pcli query key ibc-data/receipts/ports/transfer/channels/channel-8/sequences/28`
5. Observe `Not found`

avahowell:
One thing that occurred to me is that perhaps we should make a design change to the storage API to prevent writing empty keys, since any empty key (in the ibc substore) will cause this problem

crates/bin/pd/src/migrate/testnet78.rs
```diff
@@ -151,6 +155,36 @@ pub async fn migrate(
...
+            Err(e) => {
+                tracing::error!(?e, "error reading packet commitment");
+                None
+            }
+        }
+    });
+    pin_mut!(stream_empty_commitments);
+
+    while let Some(key) = stream_empty_commitments.next().await {
+        delta.delete(key);
```
erwanor, line 182: We'll have to re-apply the prefix here

erwanor (changes requested):

erwanor:
```
Error: key not found! key=ibc-data/commitments/ports/transfer/channels/channel-7/sequences/7
```

We got it, while testing I found a couple issues with prefix caching, I will open issues shortly. This PR works though, so approving and merging on green CI.

erwanor (approved):
LGTM, tested the migration on current chain state.

PR #4646, opened 2024-06-20
cnidarium: fix outdated docs for StateRead::object_get()

dynst:
The documentation was contradicting itself since #2563.


erwanor (approved):
Thanks

PR #4651, opened 2024-06-21
staking: deprecate obsolete base reward rate field

erwanor:
## Describe your changes

This PR deprecate the `core.component.staking.v1.StakeParameters.base_reward_rate` field. This field is vestigial and unused. The actual chain base reward rate is computed downstream of the distributions component budget issuance for the epoch.

  > Technically could be client breaking but I don't think it is used anywhere


cratelyn (approved):
nice!

PR #4653, opened 2024-06-24
cnidarium: prefix queries over substores are hazardous

erwanor:
## Describe your changes

This PR contains a minimal reproduction for a bug in cnidarium's prefix query handling. It also contains a sketch for a fix that we can workshop. The bug was introduced in the original substore implementation PR (#3131).

## Minimal reproduction of the prefix range cache bug.

**Context**
`cnidarium`, our storage layer, supports prefix storage.
This allows users to configure independent storage units, each with
their own merkle tree, nonverifiable sidecar, and separate namespace.
Routing is done transparently without the user having to worry about
the details.

**Overview**
Prefix queries return tuples of (key, value)s, but instead of
returning the full key, they return the substore key. This is a layering
violation, and indeed causes a bug in the cache interleaving logic.

**Terminology**
- a `full_key`:  a key that contains a substore prefix, a delimiter, and a substore key.
- a `substore_key`: a key with a stripped prefix.

**Walkthrough**
`StateDelta` index changes using full keys, as it is not aware of the
particular substore configuration that it is working against, by design.
As part of the cache interleaving logic, the `StateDetla` will try look for
new writes or covering deletions. However, since the base prefix implementation
returns substore keys, the cache will build an incoherence range and panic (or miss data).

  > Consensus breaking in the sense that the chain won't halt if we hit this.

### DIFF
--- a/crates/bin/pd/src/migrate/testnet78.rs
+++ b/crates/bin/pd/src/migrate/testnet78.rs
@@ -169,9 +169,9 @@ async fn delete_empty_deleted_packet_commitments(
     pin_mut!(stream);
 
     while let Some(entry) = stream.next().await {
-        let (substore_key, value) = entry?;
+        let (key, value) = entry?;
         if value.is_empty() {
-            delta.delete(format!("ibc-data/{substore_key}"));
+            delta.delete(key);
         }
     }
 
--- a/crates/cnidarium/src/snapshot.rs
+++ b/crates/cnidarium/src/snapshot.rs
@@ -1,3 +1,4 @@
+use std::iter;
 use std::{any::Any, sync::Arc};
 
 use anyhow::Result;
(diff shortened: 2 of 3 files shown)


avahowell:
This should also change the migration in 78 for deleting empty ibc commitments right?

erwanor:
That's right, we should update it once this gets merged

aubrika:
> This should also change the migration in 78 for deleting empty ibc commitments right?

@cratelyn @erwanor let's include the migration fix in this PR

cratelyn:
added the `consensus-breaking` label, per the description.

cratelyn (commented):
:heavy_check_mark: this looks good. approved, pending https://github.com/penumbra-zone/penumbra/pull/4653#issuecomment-2187324830 being addressed

cratelyn (approved):

PR #4654, opened 2024-06-24
ibc: fix logging of timestamp timeouts

avahowell:


cratelyn (approved):

crates/core/component/ibc/src/component/packet.rs
```diff
@@ -161,7 +161,7 @@ pub trait SendPacketRead: StateRead {
         if packet.timeout_timestamp <= chain_ts.nanoseconds() {
             anyhow::bail!(
                 "timeout timestamp {} is less than the latest timestamp on the counterparty {}",
-                packet.timeout_height,
+                packet.timeout_timestamp,
                 chain_ts,
```
zbuc, line 165: can you change this to log `chain_ts.nanoseconds()` instead so the displayed types are the same?

zbuc (changes requested):

conorsch:
During interchaintest integration work, @zbuc noticed that the logging message was structured. Loosely related to #4638 & #4636.

zbuc (approved):

conorsch (approved):

avahowell:
I just noticed that the timestamp used here is wrong, it should be using the consensus state timestamp. will fix

conorsch:
Marking this as consensus-breaking after the fact, because the diff makes substantive changes to how timeouts are handled. Please remove the label if inappropriate.

PR #4665, opened 2024-06-25
Add transaction saving and broadcasting feature

### LINKED ISSUE #4660: Support storing signed messages from cli
In current flows with pcli, if the broadcasting the TX fails for any number of common networking issues then the user must sign again from scratch.

This is particularly frustrating when using threshold custody because then you must repeat the interactive steps.

The Cosmos SDK cli has `--offline` which assembles the signed message and 'tx broadcast' commands the enable creating a signed message and then later broadcasting it.

This would be helpful flow to have in pcli

cronokirby:
This adds:

  - `pcli tx --offline <FILE> <COMMAND>`
  - `pcli tx broadcast <FILE>`

The first will save the transaction to a file after signing, the second will broadcast the result of the first command.

Closes #4660.

  > Client side only

### DIFF
--- a/crates/bin/pcli/src/command.rs
+++ b/crates/bin/pcli/src/command.rs
@@ -6,7 +6,7 @@ pub use tx::TxCmd;
 pub use validator::ValidatorCmd;
 pub use view::ViewCmd;
 
-use self::ceremony::CeremonyCmd;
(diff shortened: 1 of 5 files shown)


aubrika (approved):
LGTM!

PR #4666, opened 2024-06-25
cp: add `initial_balance` to content message

erwanor:
## Describe your changes

This adds an `initial_balance` value to the genesis content message for the CP component.

  > Consensus breaking


crates/core/component/community-pool/src/genesis.rs
```diff
@@ -35,3 +43,15 @@ impl TryFrom<pb::GenesisContent> for Content {
 impl DomainType for Content {
     type Proto = pb::GenesisContent;
 }
+
+impl Default for Content {
+    fn default() -> Self {
+        Content {
+            community_pool_params: CommunityPoolParameters::default(),
+            initial_balance: Value {
+                amount: 0u128.into(),
```
erwanor, line 52: default value is zero

crates/core/component/community-pool/src/component.rs
```diff
@@ -28,6 +28,7 @@ impl Component for CommunityPool {
         match app_state {
             Some(genesis) => {
                 state.put_community_pool_params(genesis.community_pool_params.clone());
+                state.community_pool_deposit(genesis.initial_balance).await;
```
erwanor, line 31: init chain impl

conorsch:
There's no cli integration, but with the logic added, manual munging of a genesis file would be honored. We've now got a failing test, though:

```
 thread 'app_can_propose_community_pool_spends' panicked at crates/core/app/tests/app_can_propose_community_pool_spends.rs:305:5:
assertion `left == right` failed: the community pool should be empty at the beginning of the chain
  left: {passet1984fctenw8m2fpl8a9wzguzp7j34d7vravryuhft808nyt9fdggqxmanqm: 0}
 right: {}
 ```
 
 which looks rather superficial to fix: the definition of "empty" is should be summing the values of any assets and asserting zero, not assuming no assets are in there.

hdevalence (approved):

PR #3748, opened 2024-02-06
Make all crates use workspace dependencies when shared

cronokirby:
Whenever a crate is used more than once, this switches to a workspace dependency.

This makes upgrading versions much easier.

PR #4677, opened 2024-06-27
chore(Cargo.toml): 🪄 remove explicit `default-features = true`

cratelyn:
in #3748, we migrated to using cargo's workspace dependencies feature. this has been excellent, workspace dependencies rock. 🎸

one wrinkle in the [script] used to migrate all of our constituent crates' manifests over was that it provided an explicit `default-features` attribute, even when `True`.

default features are, by their nature, enabled by default. a dependency need only provide this attribute when _disabling_ default features, in order to e.g. explicitly provide a list of `#[cfg(..)]` features to conditionally compile into a library/binary.

via the magic of vim macros and the quickfix list, this commit mechanically removes any occurrences of `default-features = true` in our dependency tree.

[script]: https://github.com/penumbra-zone/penumbra/pull/3748#issuecomment-1930641300

  > this is a noöp cosmetic change to the package manifests

### DIFF
--- a/crates/bench/Cargo.toml
+++ b/crates/bench/Cargo.toml
@@ -62,12 +62,12 @@ ark-serialize = {workspace = true}
 ark-snark = {workspace = true}
 ark-std = {workspace = true, default-features = false}
 bech32 = {workspace = true}
-decaf377 = {workspace = true, features = ["r1cs"], default-features = true}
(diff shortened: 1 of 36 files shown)


cratelyn:
not going to tug on those ci failures right now. i _may_ pick this up later when i have spare cycles.

PR #4681, opened 2024-06-28
fix: token symbol in docs

vacekj:
## Describe your changes

Changes the token symbol from the old PEN to UM.

There might be more cases of this in the docs, so keeping as a draft until I've found all of them.


vacekj:
Closing to keep the symbol in the docs deployment-neutral.

PR #4684, opened 2024-06-28
pd: rename "testnet" subcommand to "network"

conorsch:
## Describe your changes
Mostly a cosmetic change, since it's a rename. Requires lots of carry-over throughout the docs. Kept the internal APIs the same for the most part. This is a breaking change from the context of node operators, though, because now pd will look in a different location for info.

N/A

  > This is a breaking change for the CLI, for sure, although I tried hard not to touch critical app logic. Careful attention should be paid to e.g. `pd network generate` and make sure the functionality remains the same.


erwanor (approved):
LGTM, no protocol changes

PR #4689, opened 2024-07-02
Add delegator vote to TxPlannerReq

grod220:
Web consumer uses TransactionPlannerRequest to initiate planning between frontend<->extension. To support delegator voting, we'd need to add to this protobuf definition. 

This reflects the way the core repo assembles a plan: https://github.com/penumbra-zone/penumbra/blob/main/crates/bin/pcli/src/command/tx.rs#L899-L938

@plaidfinch, thoughts?


grod220:
@plaidfinch the web stack appears to be the only consumer of `TransactionPlannerRequest` (given the frontend and extension are separate systems). Pcli doesn't issue this request, but directly interacts with the planner for Delegator Voting.

PR #4690, opened 2024-07-02
pd(generate): use constants from `penumbra_app`

erwanor:
## Describe your changes

`pd network generate` and the application use consensus parameters defined in two different places which is confusing and error-prone.

This PR makes `pd network generate` import constants from the app.

> This only affects new runs of `pd network generate`.


conorsch:
Refs #4627, #4614.

conorsch (approved):
Great find, @erwanor! Adding the constants was clutch, and using them where they're needed is even better.

PR #4692, opened 2024-07-02
feat: add devnet docker compose

vacekj:
## Describe your changes

- Clones the testnet docker compose file for use in launching a local devnet. Changes the `join` command to `generate` to -generate a local devnet config. 
- Specifies the platform property in the `pd` and `pd-init` services to be `linux/amd64` to be able to run compose on ARM machines (Apple Silicon).


cratelyn (approved):
nice! thanks for taking the time to do this 🎁

PR #4694, opened 2024-07-03
view: set `MAX_CB_SIZE_BYTES` to 12MB

erwanor:
## Describe your changes

This PR sets the max decoding size of the compact block client to 12MB.


erwanor:
Note that there's no need to set a max *encoding* size since it defaults to `usize::MAX`.

conorsch:
Rebased to include changes from #4690, #4689, and #4693. Testing now.

conorsch (approved):
Tested interactively against a devnet host and it worked well.

Issue #3997, opened 2024-03-11
spec: document privacy invariants


redshiftzero:
In #3867 we documented in the protocol spec the integrity invariants of each action containing a proof. We should do the same from a privacy standpoint:
- [x] Spend
- [x] Output
- [x] Swap
- [x] SwapClaim
- [x] DelegatorVote
- [x] UndelegateClaim

Issue #3867, opened 2024-02-22
document the critical properties / invariants of each proof statement and why they are upheld 


redshiftzero:
We should add to the protocol docs a description of the invariants of each proof statement. From discussion with @hdevalence:
> an example high-level property/invariant would be “notes cannot be spent without spend authority” and the explanation would cite each of the proof statements and how they fit together

After discussion with @cronokirby, we are going to:
- [x] Define the invariants with respect to the integrity of the state of the system and the allowable transitions (e.g. a transition should not create/destroy funds) for `Spend`s / `Output`s
- [x] Ensure the description of the proof statements in the `Spend` / `Output` section of the protocol docs are accurate
- [x] Write down why each invariant is upheld by the `Spend` / `Output` proof statements or another part of the protocol
- [x] Perform the same steps for the rest of the actions

redshiftzero:
This is completed, the remaining follow ups are #3997 and #3998

PR #4695, opened 2024-07-04
Create Privacy section in the guide

vacekj:
Starting this as a place to document the privacy features and invariants of the protocol. An example would be that trades are private as long as there are at least two in a batch.

This should serve as a one-stop point for _regular users_ (not developers, node operators or other technical users of the protocol) to learn about what actions are private, and under what condition.

### DIFF
--- a/docs/guide/src/privacy.md
+++ b/docs/guide/src/privacy.md
@@ -1,20 +1,31 @@
 # Privacy
 
-This is meant to describe the privacy properties of the protocol in detail,
-as well as document any invariants that must hold for privacy to be preserved, and other notes on privacy that might be useful to users.
+This document describes the privacy features of the Penumbra protocol, as well as conditions that must be true for privacy to be preserved.
 
 ## Transfers
 
-TODO
+Transfers do not reveal the asset, amount, sender or recipient identity. Transfers are also unlinkable to each other.
 
 ## Swaps
 
-Trades are private as long as there are at least two trades in a batch.
+Initiating a swap does not reveal the identity of the swapper or the pre-paid claim fee, but it does reveal the assets and amounts in the swap.
+
+Claiming a swap does not reveal the amounts, asset types or the identity of the claimant.
+
+### Conditions
+
+The above holds true as long as there are at least two swaps in a batch.
 
 ## Governance Voting
 
-TODO
+During a vote, the voting power (amount and asset type of the staked note that's used for voting),
+the vote itself, as well as the proposal being voted on are **revealed** during a delegator vote.
+However, the address of the voter is hidden, which renders the information revealed anonymous.  
 
 ## Staking
+
+TODO
+
+## IBC
 
 TODO
--- a/docs/guide/src/privacy.md
+++ b/docs/guide/src/privacy.md
@@ -1,0 +1,20 @@
+# Privacy
+
+This is meant to describe the privacy properties of the protocol in detail,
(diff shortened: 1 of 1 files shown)


TalDerei:
reference https://github.com/penumbra-zone/penumbra/issues/3867 / https://github.com/penumbra-zone/penumbra/issues/3997 in the protocol docs

vacekj:
@TalDerei and others to whom it may concern - I've described the transfers and swaps privacy properties in this PR, since that's what I believe most users will care about. I've dropped a TODO in the other sections on staking etc. One other area which I think we should document prior to it being enabled are IBC transfers - I'm not sure what the privacy properties are there, maybe @grod220 you can help in this regard since we talked about the private IBC inbound transfers. 

I would propose we merge this now (if all the info and wording is correct) and continue filling this in as we go.

docs/guide/src/privacy.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Transfers
+
+Transfers do not reveal the asset, amount, sender or recipient identity. Transfers are also unlinkable to each other.
+
+## Swaps
+
+Initiating a swap does not reveal the identity of the swapper or the pre-paid claim fee, but it does reveal the assets and amounts in the swap.
+
+Claiming a swap does not reveal the amounts, asset types or the identity of the claimant.
```
redshiftzero, line 13: While the two notes that get minted in a successful `SwapClaim` are private (including amounts/asset types), the `SwapClaim` action does reveal the _trading pair_ due to the claimant having to provide the public `BatchSwapOutputData` which has the trading pair as a field. 

The shielded output notes are not linkable to the `BatchSwapOutputData` though, so an observer of the chain just learns an anonymous someone minted shielded outputs of a swap of assets A/B from a swap that was executed in block X.

docs/guide/src/privacy.md
```diff
@@ -0,0 +1,31 @@
...
+
+### Conditions
+
+The above holds true as long as there are at least two swaps in a batch.
+
+## Governance Voting
+
+During a vote, the voting power (amount and asset type of the staked note that's used for voting),
```
redshiftzero, line 21: You could optionally add that the identity of the validator that the voter has delegated to is revealed in a delegator vote. It's implicit given you mention that the asset type of the staked note is revealed, but it's possibly more understandable for end users as the identity of their validator.

docs/guide/src/privacy.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Swaps
+
+Initiating a swap does not reveal the identity of the swapper or the pre-paid claim fee, but it does reveal the assets and amounts in the swap.
+
+Claiming a swap does not reveal the amounts, asset types or the identity of the claimant.
+
+### Conditions
+
+The above holds true as long as there are at least two swaps in a batch.
```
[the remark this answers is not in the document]
vacekj: Thank you, fixed.

docs/guide/src/privacy.md
vacekj: Thank you - I've tried to communicate this better in the updated version.

docs/guide/src/privacy.md
vacekj: Good point - added that in with a note. Feel free to reword or Resolve if good.

redshiftzero (approved):

PR #4698, opened 2024-07-05
penumbra: prepare release `0.79-alpha.6`

erwanor:
## Describe your changes

This prepare release `0.79-alpha.6` including the simtrade rpc changes.

  > Releng


conorsch:
Closing this in favor of tag-and-pushing directly on main, so that the release commit and associated tag are both generated via the `cargo-release` tooling.

PR #4707, opened 2024-07-14
Drive `pindexer` forward with staking example

hdevalence:
## Describe your changes

This PR drives the `pindexer` development forward with worked examples of staking-related data -- using https://penumbra.today as an example use case. With these changes, it should be possible for all of the data on that site to be rendered directly out of Postgres, without any RPC calls.

We should figure out how to test this -- it should be amenable to testing since we have a good source of event data (mainnet) and expectations about what it should look like -- but this is less critical at this instant.

  > Only changes to code outside `pindexer` are convenience methods.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5809,10 +5809,15 @@ version = "0.79.0"
 dependencies = [
  "anyhow",
  "cometindex",
+ "penumbra-app",
+ "penumbra-asset",
+ "penumbra-num",
  "penumbra-proto",
  "penumbra-shielded-pool",
+ "penumbra-stake",
  "serde_json",
  "tokio",
+ "tracing",
 ]
 
 [[package]]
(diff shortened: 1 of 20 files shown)


hdevalence:
Example invocation:
```
cargo run --release --bin pindexer -- -s "postgresql://localhost:5432/penumbra_mainnet_raw?sslmode=disable" -d "postgresql://localhost:5432/penumbra_mainnet_dst_a?sslmode=disable" -g ~/.penumbra/network_data/node0/cometbft/config/genesis.json
```
Here `penumbra_mainnet_raw` is a DB of raw event data from comet (see README.md in cometindex), `penumbra_mainnet_dst_a` is a DB of compiled data.

Since the raw event data is never touched and is in a separate database, you can reindex at any point by just dropping the dst database. The genesis file is needed to initialize `AppView`s (though not all `AppView`s make use of this).

PR #4708, opened 2024-07-14
Governance support for `pindexer`

plaidfinch:
## Describe your changes

This adds a governance module to `pindexer`, along with tiny changes to `pd`'s governance events to fully support it.

  > The only changes to `pd` are changes to events emitted, which do not break consensus.


crates/bin/pindexer/src/governance/schema.sql
```diff
@@ -0,0 +1,75 @@
+CREATE TYPE proposal_stage AS ENUM ('VOTING', 'FINISHED');
+CREATE TYPE proposal_status AS ENUM ('VOTING', 'WITHDRAWN', 'FINISHED', 'CLAIMED');
+CREATE TYPE payload_type AS ENUM ('SIGNALING', 'EMERGENCY', 'PARAMETER_CHANGE', 'COMMUNITY_POOL_SPEND', 'UPGRADE_PLAN', 'FREEZE_IBC_CLIENT', 'UNFREEZE_IBC_CLIENT');
+CREATE TYPE vote_type AS ENUM ('ABSTAIN', 'YES', 'NO');
+CREATE TYPE proposal_outcome AS ENUM ('PASSED', 'FAILED', 'SLASHED');
```
hdevalence, line 5: I don't think it's a good idea to define a new serialization format here, since every consumer of these tables will need to read it. Instead, we should use the serialization format we already have -- all these enums already have JSON representations, and we should use that. Example from the validator set `AppView`:

```
penumbra_mainnet_dst_a=# SELECT DISTINCT validator_state FROM stake_validator_set;
             validator_state
------------------------------------------
 {"state":"VALIDATOR_STATE_ENUM_ACTIVE"}
 {"state":"VALIDATOR_STATE_ENUM_JAILED"}
 {"state":"VALIDATOR_STATE_ENUM_DEFINED"}
(3 rows)
```

crates/bin/pindexer/src/governance/schema.sql
```diff
@@ -0,0 +1,75 @@
...
+),
+CONSTRAINT check_stage_status_consistency CHECK (
+    (stage = 'VOTING' AND status IN ('VOTING', 'WITHDRAWN')) OR
+    (stage = 'FINISHED' AND status IN ('FINISHED', 'CLAIMED'))
+),
+CONSTRAINT check_outcome_withdrawal_consistency CHECK (
+    (is_withdrawn = FALSE AND outcome = 'PASSED') OR
+    (is_withdrawn = TRUE AND (outcome != 'PASSED' OR outcome IS NULL))
+)
+);
```
hdevalence, line 44: It's not clear what the benefit of these constraints are. We're not in a situation where the SQL database is managing the canonical application state and enforcing logic. Instead, the SQL database is compiling data that already exists. It has to accept what comes.

crates/bin/pindexer/src/governance/schema.sql
```diff
@@ -0,0 +1,75 @@
+CREATE TYPE proposal_stage AS ENUM ('VOTING', 'FINISHED');
+CREATE TYPE proposal_status AS ENUM ('VOTING', 'WITHDRAWN', 'FINISHED', 'CLAIMED');
+CREATE TYPE payload_type AS ENUM ('SIGNALING', 'EMERGENCY', 'PARAMETER_CHANGE', 'COMMUNITY_POOL_SPEND', 'UPGRADE_PLAN', 'FREEZE_IBC_CLIENT', 'UNFREEZE_IBC_CLIENT');
+CREATE TYPE vote_type AS ENUM ('ABSTAIN', 'YES', 'NO');
+CREATE TYPE proposal_outcome AS ENUM ('PASSED', 'FAILED', 'SLASHED');
+
+CREATE TABLE governance_proposals (
```
hdevalence, line 7: This table needs to have a name that prefixes the module path, like the other tables, so that it's more composable with other `AppView` implementations.

crates/bin/pindexer/src/governance/schema.sql
```diff
@@ -0,0 +1,75 @@
...
+CONSTRAINT check_voting_power CHECK (voting_power >= 0),
+CONSTRAINT check_block_height CHECK (block_height >= 0)
+);
+
+CREATE TABLE current_block_height (
+height BIGINT NOT NULL
+);
+
+INSERT INTO current_block_height (height) VALUES (0)
+ON CONFLICT (height) DO NOTHING;
```
hdevalence, line 75: Initialization is the responsibility of `AppView::init_chain`, not the SQL schema

crates/core/component/governance/src/action_handler/validator_vote.rs
```diff
@@ -96,13 +96,8 @@ impl ActionHandler for ValidatorVote {
...
                 tracing::debug!(proposal = %proposal, "emergency pass condition met, trying to enact proposal");
-                // Try to enact the proposal based on its payload
-                match state.enact_proposal(*proposal, &proposal_payload).await? {
-                    Ok(_) => tracing::debug!(proposal = %proposal, "emergency proposal enacted"),
-                    Err(error) => {
-                        tracing::error!(proposal = %proposal, %error, "error enacting emergency proposal")
-                    }
-                }
+                // Enact the proposal based on its payload
+                state.enact_proposal(*proposal, &proposal_payload).await?;
```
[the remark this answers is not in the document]
hdevalence: We should not make any app-level changes to logic, even if we believe they're not consensus breaking. This PR should just add the indexer and augment any missing data from events.

crates/core/component/governance/src/component/view.rs
```diff
@@ -422,10 +422,24 @@ pub trait StateReadExt: StateRead + penumbra_stake::StateReadExt {
...
+        .await
+        .map(Option::unwrap_or_default)
+    }
+
+    /// Get all the active validator voting power for the proposal.
+    async fn all_validator_voting_power_at_proposal_start(
+        &self,
+        proposal_id: u64,
     ) -> Result<BTreeMap<IdentityKey, u64>> {
         let mut powers = BTreeMap::new();
```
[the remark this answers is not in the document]
hdevalence: Let's not do any refactoring of consensus-critical code.

hdevalence (changes requested):
There are two high-level changes that need to be made before this can be merged:

- [ ] There should be no changes, even refactorings, to consensus-critical code along the way to adding indexer support. At most, new data can be emitted in events. We don't have bandwidth to carefully review changes to consensus-critical logic along the way to adding this logic.

- [ ] The data produced by the `AppView` implementation is expected to be consumed by third-party clients. Thus, the `AppView` impl should not be defining its own data formats ad-hoc. It should be using the canonical data formats we already have, which can be processed using standardized tooling from the Buf registry.

crates/bin/pindexer/src/governance/schema.sql
hdevalence: Let's say a downstream consumer is trying to use the data produced by this `AppView`. How are they supposed to know about any of these enum variants? They have no source of truth and cannot expect to automatically deserialize data into whatever language they're using. Instead, using TEXT-encoded JSON fixes this problem.

crates/bin/pindexer/src/governance/schema.sql
hdevalence: NACK on having these constraints

crates/bin/pindexer/src/governance/schema.sql
```diff
@@ -0,0 +1,75 @@
...
+    (stage = 'VOTING' AND status IN ('VOTING', 'WITHDRAWN')) OR
+    (stage = 'FINISHED' AND status IN ('FINISHED', 'CLAIMED'))
+),
+CONSTRAINT check_outcome_withdrawal_consistency CHECK (
+    (is_withdrawn = FALSE AND outcome = 'PASSED') OR
+    (is_withdrawn = TRUE AND (outcome != 'PASSED' OR outcome IS NULL))
+)
+);
+
+CREATE TABLE validator_votes (
```
hdevalence, line 46: Table name needs to be fixed

crates/bin/pindexer/src/governance/schema.sql
```diff
@@ -0,0 +1,75 @@
...
+validator_identity_key TEXT NOT NULL,
+vote vote_type NOT NULL,
+voting_power BIGINT NOT NULL,
+block_height BIGINT NOT NULL,
+FOREIGN KEY (proposal_id) REFERENCES governance_proposals(proposal_id),
+CONSTRAINT check_voting_power CHECK (voting_power >= 0),
+CONSTRAINT check_block_height CHECK (block_height >= 0)
+);
+
+CREATE TABLE current_block_height (
```
hdevalence, line 70: Table name needs to be fixed, but it's also unclear why the governance AppView needs this information at all.

crates/core/component/governance/src/action_handler/delegator_vote.rs
```diff
@@ -91,7 +91,7 @@ impl ActionHandler for DelegatorVote {
             .cast_delegator_vote(*proposal, identity_key, *vote, nullifier, *unbonded_amount)
             .await?;
 
-        state.record_proto(event::delegator_vote(self));
+        state.record_proto(event::delegator_vote(self, &identity_key));
```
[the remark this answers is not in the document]
hdevalence: ACK

crates/core/component/governance/src/component.rs
```diff
@@ -128,23 +128,15 @@ pub async fn enact_all_passed_proposals<S: StateWrite>(mut state: S) -> Result<(
...
-                        match state.enact_proposal(proposal_id, &payload).await? {
-                            Ok(()) => {
-                                tracing::info!(proposal = %proposal_id, "proposal passed and enacted successfully");
-                            }
-                            Err(error) => {
-                                tracing::warn!(proposal = %proposal_id, %error, "proposal passed but failed to enact");
-                            }
-                        };
-
+                        state.enact_proposal(proposal_id, &payload).await?;
```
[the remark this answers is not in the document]
hdevalence: NACK - no changes to consensus logic

crates/core/component/governance/src/component.rs
```diff
@@ -128,23 +128,15 @@ pub async fn enact_all_passed_proposals<S: StateWrite>(mut state: S) -> Result<(
...
                         let proposal =
                             state
                                 .proposal_definition(proposal_id)
                                 .await?
                                 .ok_or_else(|| {
                                     anyhow::anyhow!("proposal {} does not exist", proposal_id)
                                 })?;
-                        state.record_proto(event::enact_proposal(&proposal));
+                        state.record_proto(event::proposal_passed(&proposal));
```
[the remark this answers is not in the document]
hdevalence: ACK

crates/core/component/governance/src/component/view.rs
```diff
@@ -156,7 +156,7 @@ pub trait StateReadExt: StateRead + penumbra_stake::StateReadExt {
     /// Get the total voting power across all validators.
     async fn total_voting_power_at_proposal_start(&self, proposal_id: u64) -> Result<u64> {
         Ok(self
-            .validator_voting_power_at_proposal_start(proposal_id)
+            .all_validator_voting_power_at_proposal_start(proposal_id)
```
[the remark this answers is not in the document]
hdevalence: NACK - no changes to consensus logic

crates/core/component/governance/src/component/view.rs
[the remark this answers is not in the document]
hdevalence: NACK - no changes to consensus logic

crates/core/component/governance/src/component/view.rs
```diff
@@ -851,12 +865,7 @@ pub trait StateWriteExt: StateWrite + penumbra_ibc::component::ConnectionStateWr
     }
 
     #[instrument(skip(self))]
-    async fn enact_proposal(
-        &mut self,
-        proposal_id: u64,
-        payload: &ProposalPayload,
-    ) -> Result<Result<()>> // inner error from proposal execution
-    {
+    async fn enact_proposal(&mut self, proposal_id: u64, payload: &ProposalPayload) -> Result<()> {
```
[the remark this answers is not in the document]
hdevalence: NACK - no changes to consensus logic

proto/penumbra/penumbra/core/component/governance/v1/governance.proto
```diff
@@ -532,6 +532,8 @@ message Ratio {
 message EventDelegatorVote {
   // The delegator vote.
   DelegatorVote vote = 1;
+  // The corresponding validator's identity key.
+  keys.v1.IdentityKey validator_identity_key = 2;
```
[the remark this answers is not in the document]
hdevalence: ACK

proto/penumbra/penumbra/core/component/governance/v1/governance.proto
```diff
@@ -542,6 +544,8 @@ message EventProposalDepositClaim {
 message EventValidatorVote {
   // The validator vote.
   ValidatorVote vote = 1;
+  // The validator's voting power at the time of the proposal's start.
+  uint64 voting_power = 2;
```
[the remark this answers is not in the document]
hdevalence: ACK

proto/penumbra/penumbra/core/component/governance/v1/governance.proto
```diff
@@ -554,8 +558,8 @@ message EventProposalSubmit {
   ProposalSubmit submit = 1;
 }
 
-message EventEnactProposal {
-  // The enacted proposal.
+message EventProposalPassed {
+  // The passed proposal.
```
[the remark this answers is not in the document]
hdevalence: ACK - this is reasonable, i think

crates/bin/pindexer/src/governance/schema.sql
hdevalence: IMO it's much cleaner if this is not part of a separate .sql file and instead just goes in the `AppView` source, since it prevents this kind of misunderstanding

hdevalence:
Running this, I get
```
Error: error returned from database: bind message supplies 0 parameters, but prepared statement "sqlx_s_2" requires 1

Caused by:
    bind message supplies 0 parameters, but prepared statement "sqlx_s_2" requires 1
```

zbuc:
LGTM. @plaidfinch is this ready to merge?

zbuc (approved):

conorsch (changes requested):
One more thing: CI is currently failing on the `buf lint` action, complaining about lack of prefixes on the enum variant names:

```
❯ buf lint proto/
proto/penumbra/penumbra/core/component/governance/v1/governance.proto:327:3:Enum value name "SIGNALING" should be prefixed with "PROPOSAL_KIND_".
proto/penumbra/penumbra/core/component/governance/v1/governance.proto:327:3:Enum zero value name "SIGNALING" should be suffixed with "_UNSPECIFIED".
proto/penumbra/penumbra/core/component/governance/v1/governance.proto:328:3:Enum value name "EMERGENCY" should be prefixed with "PROPOSAL_KIND_".
proto/penumbra/penumbra/core/component/governance/v1/governance.proto:329:3:Enum value name "PARAMETER_CHANGE" should be prefixed with "PROPOSAL_KIND_".
proto/penumbra/penumbra/core/component/governance/v1/governance.proto:330:3:Enum value name "COMMUNITY_POOL_SPEND" should be prefixed with "PROPOSAL_KIND_".
proto/penumbra/penumbra/core/component/governance/v1/governance.proto:331:3:Enum value name "UPGRADE_PLAN" should be prefixed with "PROPOSAL_KIND_".
proto/penumbra/penumbra/core/component/governance/v1/governance.proto:332:3:Enum value name "FREEZE_IBC_CLIENT" should be prefixed with "PROPOSAL_KIND_".
proto/penumbra/penumbra/core/component/governance/v1/governance.proto:333:3:Enum value name "UNFREEZE_IBC_CLIENT" should be prefixed with "PROPOSAL_KIND_"
```

In all other instances of use enums in protos, we adhere to that rule, so we should here, too. I expect this'll be a rather superficial change, given all the substantive review that's occurred recently. After that, LGTM.

crates/bin/pindexer/src/stake/validator_set.rs
```diff
@@ -28,7 +28,7 @@ impl AppView for ValidatorSet {
             // hence TEXT fields
             "CREATE TABLE stake_validator_set (
                 id SERIAL PRIMARY KEY,
-                ik BYTEA NOT NULL,
+                ik TEXT NOT NULL,
```
cronokirby, line 31: Do we have to change this?

crates/bin/pvm/Cargo.toml
```diff
@@ -0,0 +1,219 @@
+[package]
+name = "pvm"
+version = {workspace = true}
+edition = {workspace = true}
```
cronokirby, line 4: I think this file was added by mistake

cronokirby (changes requested):
I think there are some accidental changes in here, also I'd rather we not change the existing validator table format unecessarily; using bytes for the identity key seems fine to me.

I also wonder, since this component depends on that one, if ordering is handled correctly between them.

More generally, I think we should go ahead and take the event generation changes, and make a separate PR for those that we can go ahead and review more thoroughly and merge, so that those are squared away; unless those events expect more churn.

crates/bin/pindexer/src/block.rs
```diff
@@ -38,7 +38,7 @@ CREATE TABLE IF NOT EXISTS block_details (
         event: &ContextualizedEvent,
     ) -> Result<(), anyhow::Error> {
         let pe = pb::EventBlockRoot::from_event(event.as_ref())?;
-        let timestamp = pe.timestamp.expect("Block has no timestamp");
+        let timestamp = pe.timestamp.unwrap_or_default();
```
cronokirby, line 41: The solution to this is to use the version of pd on the main for indexing events, we don't want to silently have January 1970 timestamps, we want to fail early when we're using the wrong version of pd because other kinds of event omission are worse.

hdevalence:
I changed the existing table on purpose as part of the PR

crates/bin/pindexer/src/block.rs
hdevalence: No, this was on purpose in order to make it work.

crates/bin/pindexer/src/stake/validator_set.rs
hdevalence: Yes, the developer experience of using byte identity keys turns out to suck, so I fixed it

hdevalence:
This code should have been merged several days ago when @plaidfinch and I pushed it up to the finish line and it was reviewed and approved. It needs to go in as-is, not be rewritten again.

cronokirby (approved):

conorsch (approved):
Tested locally in dev env, works well. My earlier comments about the proto namespaces have been addressed. I see no reason not to ship this. Separately let's follow up with pindexer-focused docs that stipulate the version-matching and genesis-preservation reqs.

hdevalence (approved):

PR #4709, opened 2024-07-15
ci: support macos in "install-cometbft" script

### LINKED ISSUE #4264: Enhance script to support mac binary installation for cometBFT

conorsch:
## Describe your changes
Flagged by @TalDerei as a nice-to-have during a recent pairing session. Closes #4264.

  > dev-env only, no app changes

### DIFF
--- a/deployments/scripts/install-cometbft
+++ b/deployments/scripts/install-cometbft
@@ -7,26 +7,38 @@ set -euo pipefail
 # Sane defaults
 COMETBFT_VERSION="${COMETBFT_VERSION:-0.37.5}"
 
+# Determine platform, so we can support both macOS and Linux.
+uname_kernel="$(uname -s)"
+uname_machine="$(uname -m)"
+if [[ "$uname_kernel" = "Linux" && "$uname_machine" = "x86_64" ]] ; then
+    platform_string="linux_amd64"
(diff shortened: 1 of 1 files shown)


conorsch:
@aubrika Do me a favor and run this locally—I added a change to automatically add `cometbft` to `/usr/local/bin/cometbft`, prompting via `sudo` for creds to do so. Give that a shot and see how it treats ya: after running this script, `which cometbft` should show `/usr/local/bin/cometbft`. 

The correct long-term solution here is to just use nix to manage the dependencies, so we can more easily pull in external tools like postgres too, but I'll get to that later.

conorsch:
Closing this as moot, given the automated tooling in #4726. Later on I'll follow up to clean out the deprecated shell scripts, and updated the CI environment to reuse the nix config.

PR #4712, opened 2024-07-16
Make pindexer more easily embeddable as a library

cronokirby:
## Describe your changes

We don't expose cometindex as a library, but we do pindexer, and this adds a few tweaks to make pindexer possible to run inside an external piece of Rust code. This is useful, because it allows people to colocate their indexer and their block explorer, for example. 

  > Doesn't touch the core protocol.

### DIFF
--- a/crates/util/cometindex/src/indexer.rs
+++ b/crates/util/cometindex/src/indexer.rs
@@ -1,7 +1,6 @@
 use std::pin::Pin;
 
 use anyhow::{Context as _, Result};
-use clap::Parser;
 use futures::{Stream, StreamExt, TryStreamExt};
 use sqlx::{postgres::PgPoolOptions, PgPool};
 use tap::{Tap, TapFallible, TapOptional};
@@ -16,9 +15,9 @@ pub struct Indexer {
 }
 
 impl Indexer {
-    pub fn new() -> Self {
+    pub fn new(opts: Options) -> Self {
         Self {
-            opts: Options::parse(),
+            opts,
             indexes: Vec::new(),
         }
     }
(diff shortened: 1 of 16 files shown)


crates/util/cometindex/src/indexer.rs
```diff
@@ -16,9 +15,9 @@ pub struct Indexer {
 }
 
 impl Indexer {
-    pub fn new() -> Self {
+    pub fn new(opts: Options) -> Self {
         Self {
-            opts: Options::parse(),
+            opts,
```
hdevalence, line 20: I'm not sure this is desirable -- the original API was structured that way so that downstream consumers didn't have control over the CLI. Everyone gets the same binary, the only difference is which `AppView`s are hooked into it.

cronokirby: If you want to hook in different AppViews you have to consumer pindexer as a library, in which case having control over your CLI seems much better to me, and doesn't cause odd spooky behavior.

conorsch: I don't fully understand the disagreement about exposing CLI opts: in the interest of making `pindexer` more accessible, e.g. via importing into other rust code, I'm biased toward making this change and dealing with the consequences later.

conorsch (approved):
Approving, to unblock ongoing `pindexer` workstreams. There is still a lingering disagreement about how to handle CLI args wrt to downstream consumers. Let's figure that out as we go, and not block here on it!

crates/util/cometindex/src/indexer.rs
hdevalence: Yes, this change should be rolled back, it's an explicit design goal to not allow control over the CLI, so that every Penumbra indexer has the same interface.

PR #4715, opened 2024-07-16
fix: update FMD parameters when needed in compact block

### LINKED ISSUE #4714: Updated FMD parameters not provided in compact block
**Describe the bug**

When using an algorithm that dynamically changes the FMD parameters, the updated FMD parameters are not provided in the compact block. 

**To Reproduce**
1. Generate configs following the dev net instructions here: https://guide.penumbra.zone/dev/devnet-quickstart.html#generating-configs
2. Edit the generated `genesis.json` to ensure you have an allocation to a test wallet and replace the existing key for `fmdMetaParams` with:

```
"fmdMetaParams": {
            "fmdGracePeriodBlocks": "360",
            "slidingWindow": {
                "windowUpdatePeriods": 48,
                "targetedDetectionsPerWindow": 500
            }
        }
```

4. Follow the rest of the devnet set up guide beginning with https://guide.penumbra.zone/dev/devnet-quickstart.html#running-pd
5. Run through the first FMD parameter change
6. Attempt to send a tx

**Expected behavior**
You should see the tx succeed

**Actual behavior**
The tx will fail due to the wrong FMD parameters being used:
```
status: Internal, message: "Error submitting transaction: code 1, log: failed to deliver transaction: check_stateful failed: consensus rule violated: invalid clue precision", details: [], metadata: MetadataMap { headers: {} }
```

**Additional context**
Inspecting the pcli view database indicates that the FMD parameters are not being updated (you can verify this periodically through the devnet run after attempting a tx and using `select hex(v) from kv where k='fmd_params';`).

Looking at the logic in `finalize_compact_block` [here](https://github.com/penumbra-zone/penumbra/blob/main/crates/core/component/compact-block/src/component/manager.rs#L70-L78):

```rust
        let fmd_parameters = if height == 0 {
            Some(
                self.get_current_fmd_parameters()
                    .await
                    .context("could not get FMD parameters")?,
            )
        } else {
            None
        };
```

the FMD parameters are only being provided at block height 0

redshiftzero:
## Describe your changes

This fixes the compact block generation such that updated FMD parameters are provided. This is required to use an algorithm in the future that dynamically updates the FMD parameters (see #1087)

Closes #4714

### DIFF
--- a/crates/core/component/compact-block/src/component/manager.rs
+++ b/crates/core/component/compact-block/src/component/manager.rs
@@ -75,7 +75,10 @@ trait Inner: StateWrite {
             .get_current_fmd_parameters()
             .await
             .context("could not get FMD parameters")?;
-        let fmd_parameters = if previous_fmd_params != current_fmd_params || height == 0 {
+        let fmd_parameters = if (previous_fmd_params != current_fmd_params
+            && current_fmd_params.as_of_block_height == height)
+            || height == 0
+        {
             Some(current_fmd_params)
         } else {
             None
--- a/crates/core/component/compact-block/src/component/manager.rs
+++ b/crates/core/component/compact-block/src/component/manager.rs
@@ -67,12 +67,16 @@ trait Inner: StateWrite {
             (None, Vec::new())
         };
 
-        let fmd_parameters = if height == 0 {
-            Some(
-                self.get_current_fmd_parameters()
-                    .await
-                    .context("could not get FMD parameters")?,
-            )
+        let previous_fmd_params = self
+            .get_previous_fmd_parameters()
+            .await
+            .context("could not get FMD parameters")?;
+        let current_fmd_params = self
+            .get_current_fmd_parameters()
+            .await
+            .context("could not get FMD parameters")?;
+        let fmd_parameters = if previous_fmd_params != current_fmd_params {
+            Some(current_fmd_params)
         } else {
             None
         };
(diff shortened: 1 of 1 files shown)


crates/core/component/compact-block/src/component/manager.rs
```diff
@@ -67,12 +67,16 @@ trait Inner: StateWrite {
...
-            )
+        let previous_fmd_params = self
+            .get_previous_fmd_parameters()
+            .await
+            .context("could not get FMD parameters")?;
+        let current_fmd_params = self
+            .get_current_fmd_parameters()
+            .await
+            .context("could not get FMD parameters")?;
+        let fmd_parameters = if previous_fmd_params != current_fmd_params {
```
cronokirby, line 78:
```suggestion
        let fmd_parameters = if previous_fmd_params != current_fmd_params && current_fmd_params.as_of_block_height == height {
```

cronokirby: e.g. if the params are:
```
0 0 0 0 0 0 1 1 1 1 1 1
```
we only want to send over the new params in the compact block where they changed

cronokirby (approved):

conorsch:
> This fixes the compact block generation

which means this is a consensus-breaking change, no? There's no rationale in the PR text explaining why it _isn't_ a consensus-breaking change, so I'm adding the tag. Please remove if I've misunderstood.

cronokirby:
It breaks consensus only for a dynamic FMD algorithm

hdevalence:
This would need to be added in a chain upgrade, then, and there are currently no plans for chain upgrades, pending emergence of a community-driven upgrade process. So we'd need to icebox it for now.

conorsch (changes requested):
Requesting changes to block merge, pending chain-upgrade plans.

PR #4718, opened 2024-07-17
deps: update to ics23 0.11.3

hdevalence:
## Describe your changes

Updates to `ics23` `0.11.3`

  > No effect on consensus state as no IBC transactions have occurred and no changes in library other than bug fixes.


zbuc:
Are transitive dependencies such as `ibc-types` on `ics23` a concern here?

hdevalence:
There's only one version of the `ics23` crate in the workspace, and we can see from the lock file that it's set to the version we want. Cargo unifies compatible dependencies. This can also be verified using `cargo tree -i ics23`.

conorsch (approved):

faddat:
>  If this code contains consensus-breaking changes, I have added the "consensus-breaking" label. Otherwise, I declare my belief that there are not consensus-breaking changes, for the following reason:

awesome

Issue #2920, opened 2023-08-11
Completely remove all use of `unwrap` in `pd`; audit every use of `expect`


plaidfinch:
Bugs lurk where implicit panics can happen, see https://github.com/penumbra-zone/penumbra/issues/2919 for one recent example. We should never be using `unwrap` because it does not produce meaningful error messages or document the invariants that are being violated when the panic occurs.

We should audit every occurrence of `unwrap` and `expect` to ensure that we have fully documented the invariants that we believe hold and will never be violated. All `unwrap`s should be converted to `expect`s with good error messages. Any case where it is possible to return an `Err` variant instead, we should do that instead.

To aid in this task, I suggest that we add `#![deny(clippy::unwrap_used)]` at the crate level to anything in consensus (that is, `pd` and its dependencies). This will allow us to triage all uses of `unwrap` by following the clippy lints.

zbuc:
Removing the existing use of `unwrap` is fairly straightforward but auditing the use of `expect` afterwards requires a lot of specific context and reasoning to determine those invariants, as well as thinking through scenarios where they might not hold due to behavior of chain participants.

Here are some places that had `expect` statements that concerned me:

* `Unit::parse_value` -- there are some `anyhow::bail` calls here in methods that return `Result` types -- should they bubble up instead?
* `Builder::build` -- when parsing display regexes we can panic because of an invalid regex. Are these display regexes controlled by chain participants/could it cause a chain halt?

(will update as I go)

zbuc:
Closing this as complete, every consensus-critical and user-facing binary crate has been modified to have the clippy lint forbidding `unwrap` usage. 

Auditing the usage of `expect` was split into a separate task, #3050

PR #4719, opened 2024-07-17
shielded-pool: remove `unwrap`

dynst:
## Describe your changes

Remove `unwrap()` calls added in #4404 that made `#![deny(clippy::unwrap_used)]` unhappy.

See also #2920.

  > no functionality change


cronokirby:
These unwraps are gated behind `is_err`, so no risk there. In the interest of not touching potentially consensus breaking code, we should keep the expects in there for now.

Issue #4675, opened 2024-06-27
pindexer: ⏰ add `AppView` mapping height to timestamps


cratelyn:
other indexes include a mapping from block height to timestamp. add an app view that does _just_ that, as a stepping stone towards the views for block contents (#4674) and trading volume (#4673). 

https://github.com/penumbra-zone/dex-explorer/issues/25

cratelyn:
see #4522 / #4523 for an RPC that returns consensus timestamps based on height.

* #4522
* #4523

PR #4722, opened 2024-07-18
feat: add timestamp to all event protos

vacekj:
## Describe your changes

this makes a non-consensus-breaking change, adding a timestamp field to the EventBlockRoot proto.
This will be used to keep track of the timestamp of the block for use in pindexer.

I tested this by running a local devnet with the changes to core. Pd runs fine and the timestamps are correctly propagated to the raw events db.

prerequisite for https://github.com/penumbra-zone/penumbra/issues/4675

- only adds a field
- adds this field to an Event proto


hdevalence (approved):
This looks good, but can we add the timestamp to the `EventEpochRoot` and `EventAnchor` as well? Even if the data is duplicated across multiple events, I think it will be useful for downstream data consumers to be able to access it directly.

For instance, the difference between the time a transaction was accepted into a block and the timestamp of the anchor it proves against is an interesting measurement / information leak -- it gives info about the combined latency of proving + tx submission -- and it'd be convenient to have timestamps on every anchor.

cronokirby (approved):
Awesome, let's merge this!

PR #4725, opened 2024-07-18
Added CMake command to dev docs for linking against `librocksdb` 

ejmg:
In the current docs for linking against `libsrockdb`, building `libsnappy.a` will fail because of an error thrown while building its third party benchmarks which are enabled by default.

## Describe your changes
This PR adds the CMake CLI option to the docs so that snappy's optional benchmarks are disabled by default.

NA

This is a one line change to developer docs for linking against a crate dependency.

### DIFF
--- a/docs/guide/src/dev/build.md
+++ b/docs/guide/src/dev/build.md
@@ -121,7 +121,7 @@ git submodule update --init
 
 # Build snappy using cmake.
 mkdir build
-cd build && cmake .. && make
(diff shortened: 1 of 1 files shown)


cronokirby (approved):

conorsch:
Thanks for taking the time, @ejmg! Just-merged is a more comprehensive dev env in #4726, which should do the needful on setting up local tooling, regardless of your workstation platform.

PR #4728, opened 2024-07-21
fixing broken pathway

Budalebah:
fixing broken pathway for cometbft releases


cronokirby (changes requested):
This PR should be amended to only change install.md

Budalebah:
Okey let me fix

PR #4730, opened 2024-07-22
fixing broken pathway

Budalebah:
fixing broken pathway


conorsch:
Hello, @Budalebah, thanks for your interest in the project. I don't understand what this PR is trying to accomplish. I double-checked on the live https://guide.penumbra.zone site, and the CometBFT download URLs are working correctly. We have some CI checks that try to catch broken URLs in the docs, and one of those checks is failing on this PR, indicating that the change presented here is actually a regression. Please feel free to re-open if you believe I've misunderstood something.

Budalebah:
Hey, I wasnt about Urls, it was about the link about mentioned here. It has typos so directing to non working page

PR #4734, opened 2024-07-22
(backport) feat: add timestamp to all event protos (#4722)

cronokirby:
this makes a non-consensus-breaking change, adding a timestamp field to the EventBlockRoot proto.
This will be used to keep track of the timestamp of the block for use in pindexer.

I tested this by running a local devnet with the changes to core. Pd runs fine and the timestamps are correctly propagated to the raw events db.

prerequisite for https://github.com/penumbra-zone/penumbra/issues/4675

- only adds a field
- adds this field to an Event proto


conorsch:
Refs #4722.

conorsch (approved):

PR #4753, opened 2024-07-24
pindexer: implement dex value circuit breaker indexing

cronokirby:
## Describe your changes

This implements a basic dex indexing component for the value circuit breaker; more importantly, this adds some basic scaffolding that other dex related indexing can latch onto.

  > client code only, no event changes

### DIFF
--- a/crates/bin/pindexer/src/indexer_ext.rs
+++ b/crates/bin/pindexer/src/indexer_ext.rs
@@ -7,7 +7,6 @@ impl IndexerExt for cometindex::Indexer {
         self.with_index(crate::shielded_pool::fmd::ClueSet {})
             .with_index(crate::stake::ValidatorSet {})
             .with_index(crate::stake::Slashings {})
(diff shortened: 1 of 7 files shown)


crates/bin/pindexer/src/indexer_ext.rs
```diff
@@ -7,8 +7,8 @@ impl IndexerExt for cometindex::Indexer {
         self.with_index(crate::shielded_pool::fmd::ClueSet {})
             .with_index(crate::stake::ValidatorSet {})
             .with_index(crate::stake::Slashings {})
-            .with_index(crate::stake::MissedBlocks {})
```
conorsch, line 10: Why remove `MissedBlocks` here?

cronokirby: We don't consume this in our dashboards right now, and this slows down indexing a lot, since this is a frequent event.

conorsch (approved):

PR #4755, opened 2024-07-24
Add a Position to EventPositionOpen

cronokirby:
This is necessary because at no other point in the related chain of events do we ever see the full position, and we need to see the full position to get information such as the price.

We could simply add in the price, but it's more conservative to simply emit the entire position, in case there's anything else we might've forgotten about.

  > This adds new data to events.


aubrika (approved):
Looks purely additive to me - approved

PR #4756, opened 2024-07-24
ci: add pindexer to penumbra container

conorsch:
## Describe your changes
Adding pindexer to the default container build, to make the built binary accessible in testing and deploy environments.

N/A

  > only affects ci builds of containers; no changes to application logic


aubrika (approved):
👍🏻

PR #4757, opened 2024-07-24
Tweak block indexing to use height as the primary key

cronokirby:
Every block has a unique height, we almost always want to join with other tables based on block height, the serial ID has no meaning: ergo, we should get rid of the id, and just use height as the primary key.

There was no index on height either, which would have made joining on it slow.

Also removed unwraps and expects and as in favor of actual errors.

  > Indexer changes


aubrika (approved):
good change  👍🏻

PR #4758, opened 2024-07-25
Fix `cometindex` destination db and table initialization logic

ejmg:
## Describe your changes
The original logic for `Indexer::run` had a initialization check that created the watermark table and destination tables. The problem with this check is that there isn't a guarantee that the destination tables were initialized just because a watermark table is detected. This PR moves the initialization of destination tables outside of the check.

Only the crate `cometindex` is effected by this PR.

### DIFF
--- a/crates/util/cometindex/src/indexer.rs
+++ b/crates/util/cometindex/src/indexer.rs
@@ -88,27 +88,33 @@ impl Indexer {
         .await?;
 
         if !dst_db_initialized {
-            tracing::info!("no watermark found, initializing with genesis data");
+            tracing::info!("no watermark found, creating index_watermark and initializing dst_db with genesis data");
 
             // Create the table if it doesn't exist
             sqlx::query("CREATE TABLE index_watermark (events_rowid BIGINT NOT NULL)")
                 .execute(&dst_db)
                 .await?;
-
-            // Load the genesis JSON to be used populating initial tables
-            let genesis_content: serde_json::Value = serde_json::from_str(
-                &std::fs::read_to_string(genesis_json)
-                    .context("error reading provided genesis.json file")?,
-            )
-            .context("error parsing provided genesis.json file")?;
-            let app_state = genesis_content
-                .get("app_state")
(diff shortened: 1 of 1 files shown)


cronokirby:
I think the initialization logic we want can just be "is the database completely empty or not", having intermediate states beyond this is complicated to reason about and we can just not reason about it for now by completely wiping the database when we want to run a different indexer, which makes this particular situation not a thing we have to deal with.

ejmg:
Fair, but the current initialization logic still allows for a custom AppView to fail when ran because of how the check is structured. The watermark table can be defined without having the destination tables initialized and that will always throw an unnecessary error. This prevents that specific case regardless of what else downstream consumers might do with respect to initializing their indexers.

cronokirby:
Yes, but you *want* it to fail, because you don't want to start running an indexing component that's missed a bunch of history, because:
- missing history is in and of itself, not good
- components can enter weird states if they're run on partial history (for example, validator voting power could end up negative or something)

aubrika:
```there isn't a guarantee that the destination tables were initialized just because a watermark table is detected```

the above is worth fixing

```I think the initialization logic we want can just be "is the database completely empty or not", ```

and I agree that wiping the database completely & restarting from zero if not completely empty is the preferable fix, for the reasons Lucas indicates

conorsch:
@ejmg I'm seeing consensus here that the failure at startup when db contents are not what's as expected is by design, and part of the contract. I recognize that we could make the db init a bit smarter, but I agree it's safest to require that indexer operators provision the database cleanly ahead of running the indexer against it. Ideally we'll improve that story over time, but for now, especially given the debugging in e.g. https://github.com/penumbra-zone/penumbra/issues/4759, it's the best use of our time to be explicit about the db initial state. 

There's some great suggestions over in https://github.com/penumbra-zone/penumbra/pull/4752, and I'll let @cronokirby chime in about how best to support that workflow.

conorsch:
>the failure at startup when db contents are not what's as expected is by design, and part of the contract. 

We made a similar call over in https://github.com/penumbra-zone/penumbra/pull/4764

ejmg:
Sounds good to me 👍

Issue #4759, opened 2024-07-25
bug: pindexer db init fails


conorsch:
**Describe the bug**
I'm unable to get `pindexer` running locally in the dev env.

**To Reproduce**
1. Use nix dev env https://guide.penumbra.zone/dev/dev-env
2. `cargo run --release --bin pd -- network unsafe-reset-all`
3. `just dev`
4. Observe failure in `pindexer` startup

**Expected behavior**
I expect that `pindexer` will start fine in the default dev env. Instead, it fails on startup, with error screenshotted below.

**Screenshots**
![pindexer-startup-error](https://github.com/user-attachments/assets/319cc82c-4974-4a44-b2ae-7265157d935c)

**Additional context**
 It's possible that the ad-hoc postgres grants in the `deployments/compose/process-compose-postgres.yml` are no longer correct, or the problem could be in the sql logic within `pindexer`. Whatever it is, I could use a hand sorting out the correct settings, so I can continue setting up pindexer in other contexts.

conorsch:
Unsurprisingly, @aubrika was a huge help here. While debugging locally, I stumbled upon the [sqlfluff] tool, which can be used to lint sql files and find parsing errors:

```
❯ sqlfluff lint --dialect postgres crates/bin/pindexer/src/dex/dex.sql
== [crates/bin/pindexer/src/dex/dex.sql] FAIL
L:   7 | P:   1 | LT05 | Line is too long (81 > 80). [layout.long_lines]
L:   9 | P:   1 | LT05 | Line is too long (82 > 80). [layout.long_lines]
L:  12 | P:   1 |  PRS | Line 12, Position 1: Found unparsable section: 'CREATE
                       | DOMAIN IF NOT EXISTS Amount AS NU...'
WARNING: Parsing errors found and dialect is set to 'postgres'. Have you configured your dialect correctly?
```

That's probably worth sticking in CI.

ejmg:
my PR for cometindex fixes some logic that is connected to this [type of bug which IMO falls from the tacit contract](https://github.com/penumbra-zone/penumbra/pull/4758/files#diff-2c5fcc4d4d58fd75279842103feee963bd9778d7f7908e8407ff7a9acbb13d31R101-R106) on `AppView` that an implementer's `init_chain` requires sanitary SQL that makes sure to check/tear down anything that is defined for that specific indexer.

conorsch:
[...]

PR #4760, opened 2024-07-25
pindexer: repair SQL syntax for create type/create domain

aubrika:
Fixes some SQL statements in the prior commit which were not valid postgres.

  > Only indexer changes.

### DIFF
--- a/crates/bin/pindexer/src/dex/dex.sql
+++ b/crates/bin/pindexer/src/dex/dex.sql
@@ -9,12 +9,16 @@
 -- that given an `penumbra_asset::asset::Id`, we always know exactly how to filter
 -- tables, rather than needing to do a join with another table.
 
-CREATE DOMAIN IF NOT EXISTS Amount AS NUMERIC(39, 0) NOT NULL;
 
+DROP TYPE IF EXISTS Value CASCADE;
+DROP DOMAIN IF EXISTS Amount;
+
+CREATE DOMAIN Amount AS NUMERIC(39, 0) NOT NULL;
 CREATE TYPE Value AS (
   amount Amount,
-  asset BYTEA NOT NULL
+  asset BYTEA
 );
+
 
 -- Keeps track of changes to the dex's value circuit breaker.
 CREATE TABLE IF NOT EXISTS dex_value_circuit_breaker_change (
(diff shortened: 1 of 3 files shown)


crates/bin/pindexer/src/dex/dex.sql
```diff
@@ -9,13 +9,17 @@
...
-CREATE DOMAIN IF NOT EXISTS Amount AS NUMERIC(39, 0) NOT NULL;
 
+DROP TYPE IF EXISTS Value CASCADE;
+DROP DOMAIN IF EXISTS Amount;
+
+CREATE DOMAIN Amount AS NUMERIC(39, 0) NOT NULL;
 CREATE TYPE Value AS (
   amount Amount,
-  asset BYTEA NOT NULL
+  asset BYTEA
```
aubrika, line 19: If the `NOT NULL` constraint intended here previously is critical, this could be refactored with something like `CREATE DOMAIN Asset AS BYTEA NOT NULL;` but that might cause translation issues & require a bit more work that this

conorsch:
This helps, but I'm still seeing errors when I attempt to spin up locally. Running this command should pass, for us to merge this PR:

```
❯ sqlfluff parse --dialect postgres crates/bin/pindexer/src/dex/dex.sql
[...snip...]
==== parsing violations ====
L:  34 | P:   1 |  PRS | Line 34, Position 1: Found unparsable section: 'CREATE TABLE IF
                       | NOT EXISTS trace_step (\n...'
WARNING: Parsing errors found and dialect is set to 'postgres'. Have you configured your dialect correctly?
❯ echo $?
1
```

conorsch:
Much better! The `sqlfluff parse <file>` command passes now, but I'm still seeing errors on `pindexer` startup. This time it's:

```
Error: error returned from database: cannot insert multiple commands into a prepared statement
```

To test, I'm using the STR in #4759.

conorsch:
That did it! :muscle:

conorsch (dismissed):

conorsch:
Drat, I spoke too soon:

```
Error: error returned from database: duplicate key value violates unique constraint "block_details_pkey"                                                                                   ```

cronokirby:
That seems to point to a weirder thing in cometindex because that should only happen if you ingest two block root events with the same height

cronokirby:
https://github.com/penumbra-zone/penumbra/pull/4762 fixes a bug in cometindex which leads to this

conorsch:
> https://github.com/penumbra-zone/penumbra/pull/4762 fixes a bug in cometindex which leads to this

Yup, testing locally with that fix cherry-picked in resolves, thanks. :+1:

conorsch (approved):

PR #4762, opened 2024-07-25
Fix: index events according to their local row id

cronokirby:
We use this row id as a watermark, and update it after every ingestion of an event. Previously, we were iterating over things in an undefined order, and effectively sometimes ingesting events out of order, causing bad things to happen to the watermark, leading to much FUN.

  > indexing


vacekj (approved):

conorsch (approved):
Tested interactively in https://github.com/penumbra-zone/penumbra/pull/4760#issuecomment-2250879907

PR #4763, opened 2024-07-25
Update to latest frontends

grod220:
Been a few weeks since an update with the bundled frontends. Updating to latest main from `penumbra-zone/web`.


conorsch:
This is great, however we also want to include changes in https://github.com/penumbra-zone/web/pull/1533, so we'll need another shortly.

Issue #4761, opened 2024-07-25
`pindexer::block` panics on timestamp 


ejmg:
**Describe the bug**
As it stands, the custom app view introduced by #4713 for tracking block timestamps throws a hard error whenever it tries to extract a timestamp from a `EventBlockRoot`.
 
**To Reproduce**
1. Stand up a raw indexer db and destination db for pindexer and run the binary as the README instructs
2. You will immediately hit an application error

**Expected behavior**
It should correctly extract and index timestamp events for `EventBlockRoot` events without immediately killing pindexer.
**Screenshots**
basic stack trace with `RUST_LOG=debug`
```
     Running `target/debug/pindexer -s 'postgresql://localhost:5432/raw_79_1?sslmode=disable' -d 'postgresql://localhost:5432/compiled_79_1?sslmode=disable' -g genesis.json`
2024-07-25T02:23:54.653952Z  INFO cometindex::indexer: self.opts=Options { src_database_url: "postgresql://localhost:5432/raw_79_1?sslmode=disable", dst_database_url: "postgresql://localhost:5432/compiled_79_1?sslmode=disable", chain_id: None, poll_ms: 500ms, genesis_json: "genesis.json" }
2024-07-25T02:23:54.666144Z DEBUG sqlx::query: summary="SET SESSION CHARACTERISTICS AS …" db.statement="\n\nSET\n  SESSION CHARACTERISTICS AS TRANSACTION READ ONLY;\n" rows_affected=0 rows_returned=0 elapsed=994.125µs
2024-07-25T02:23:54.682616Z DEBUG sqlx::query: summary="SELECT EXISTS ( SELECT …" db.statement="\n\nSELECT\n  EXISTS (\n    SELECT\n    FROM\n      information_schema.tables\n    WHERE\n      table_name = 'index_watermark'\n  )\n" rows_affected=0 rows_returned=1 elapsed=7.050917ms
2024-07-25T02:23:54.682672Z  INFO cometindex::indexer: skipping genesis initialization
2024-07-25T02:23:54.689818Z DEBUG sqlx::query: summary="SELECT events_rowid FROM index_watermark" db.statement="" rows_affected=0 rows_returned=1 elapsed=1.210667ms
2024-07-25T02:23:54.689880Z DEBUG cometindex::indexer: fetched index watermark row_id=37835
[...]

PR #4764, opened 2024-07-25
fix: don't crash pindexer on blocks without a timestamp

vacekj:
## Describe your changes

this sets timestamp to default for blocks without one to

fixes https://github.com/penumbra-zone/penumbra/issues/4761

indexer changes only


cronokirby:
Crashing is good, it indicates that you have the wrong version of pd, before the event was added.

hdevalence:
No, it's not, this fixes a problem that messes up the software in prod.

hdevalence (approved):

PR #4766, opened 2024-07-25
Bug: correctly handle denoms with `/` in pcli tx lp

cronokirby:
The parsing looked for the first occurrence of the slash for

nX@nY/fee

but if X or Y contains a slash, this will be incorrect.

This fixes this somewhat, by looking for the last slash, and only considering that a fee if it ends with bps.

This will not correctly handle a denom containing both a slash and ending with bps, but beyond that we'll need to change the format probably.

  > client side change only, for pcli


aubrika (approved):
Seems good enough for now, we can address the denom incompatibility issue if/when it arises 🤷🏻‍♀️

PR #4769, opened 2024-07-25
Augment missing events for governance

cronokirby:
This takes the event related changes from https://github.com/penumbra-zone/penumbra/pull/4708 and pulls them out for independent review and merging.

  > event emission only.


cronokirby:
c.f. https://github.com/penumbra-zone/penumbra/pull/4708#discussion_r1677997075 for discussion on the lint failure

aubrika (approved):
LGTM 👍🏻

aubrika (changes requested):
One hesitation - are we sure renaming the EventEnactProposal proto isn't going to break anything? Seems better not to do this if we can avoid it, although semantically the change does seem like an improvement

hdevalence:
> One hesitation - are we sure renaming the EventEnactProposal proto isn't going to break anything? Seems better not to do this if we can avoid it, although semantically the change does seem like an improvement

Yes, it breaks anything that was using it, which at the time was only the code in that PR

PR #4771, opened 2024-07-25
feat: lp indexing

### LINKED ISSUE #4739: pindexer: dex: implement /api/lp related events and tables
We should implement feature parity with the following api:
https://github.com/penumbra-zone/dex-explorer/tree/2b70ae60529284309f8748f6ca6a154a2bab82a9/src/pages/api/lp

vacekj:
## Describe your changes

implements lp positions and execution indexing

TODO from reviews:
- [x] We want to store the trading pair likely as a new derived type (tuple) of (BYTEA, BYTEA)
- [x] We definitely need a table storing the current state of any given position, including permanent information like the trading pair, price, fees.
- [x] We need to include the reserve amounts in the lp_updates table, to have a history of the reserves over time.
- [x] Price info, for both updates and current state

fixes #4739 

indexer-only changes

### DIFF
--- a/crates/bin/pindexer/src/block.rs
+++ b/crates/bin/pindexer/src/block.rs
@@ -1,5 +1,3 @@
-use crate::PD_COMPAT;
-use anyhow::anyhow;
 use cometindex::{async_trait, sqlx, AppView, ContextualizedEvent, PgTransaction};
 use penumbra_proto::{core::component::sct::v1 as pb, event::ProtoEvent};
 use sqlx::{types::chrono::DateTime, PgPool};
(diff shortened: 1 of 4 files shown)


cronokirby:
This is a good start, but I think we want to compute some more derived information for use in the explorer, for example the price, so that the explorer can easily sort and fetch by price.

cronokirby (changes requested):
Let's clean up the accidental formatting changes and make sure that we have all the information we need to implement the APIs.

conorsch:
>This is a good start, but I think we want to compute some more derived information for use in the explorer

Downgrading to "draft", mostly to signal that we don't intend to block on upcoming point-release until this lands. Feel free to promote back to ready for review whenever, @vacekj!

cronokirby:
A few more things I noticed:
- We want to store the trading pair likely as a new derived type (tuple) of `(BYTEA, BYTEA)`
- We definitely need a table storing the current state of any given position, including permanent information like the trading pair, price, fees.
- We need to include the reserve amounts in the lp_updates table, to have a history of the reserves over time.

vacekj:
@cronokirby this leads me to believe we then have all we need in the data? Can we proceed to review and merge?

cronokirby:
I think we want:
- to store the price directly so that we can easily replicate the search by price queries
- to actually materialize each LP rather than computing it by taking the last seen update, because we need extra information like the price, and can avoid storing information that doesn't change in every update

but more broadly with this one we can actually test that it works by modifying the dex explorer raw indexing database queries to use it, and we should do that.

cronokirby (commented):
It would also be nice if we didn't create a diff over the parts of the database we haven't changed, we can format in a separate PR

cronokirby (approved):
Looks good to me now, did some brief testing as well

PR #4772, opened 2024-07-25
fix(pindexer): correct encoding of Amount into Postgres

cronokirby:
I tried and tried to actually use Numeric directly in a smart way, instead the hack is to just cast text into it in the SQL query. Oh well.

  > pindexer bug fix

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -762,17 +762,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cf9ff0bbfd639f15c74af777d81383cf53efb7c93613f6cab67c6c11e05bbf8b"
 
 [[package]]
-name = "bigdecimal"
-version = "0.3.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a6773ddc0eafc0e509fb60e48dff7f450f8e674a0686ae8605e8d9901bd5eefa"
-dependencies = [
- "num-bigint",
- "num-integer",
- "num-traits",
-]
-
-[[package]]
 name = "bincode"
 version = "1.3.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


conorsch:
Context: while ingesting events from a devnet, we encountered this error in pindexer:

```
Error: error returned from database: operator does not exist: - bytea

Caused by:
    operator does not exist: - bytea
```

which we tracked down to https://github.com/penumbra-zone/penumbra/blob/3b21ac8649e52bf3eaaa8a2ee5396aa88ab6a631/crates/bin/pindexer/src/dex/mod.rs#L81, specifically mismatched types.

conorsch:
Even with this place in place, when reindexing a devnet db, I encounter a new (but different) error:

```
2024-07-25T23:24:35.437139Z DEBUG index_event{height=18957 name="penumbra.core.component.dex.v1.EventArbExecution"}: sqlx::query: summary="INSERT INTO trace_step VALUES …" db.statement="\n\nINSERT INTO\n  trace_step\nVALUES\n  (DEFAULT, (CAST($1 AS Amount), $2)) RETURNING id;\n" rows_affected=0 rows_returned=1 elapsed=9.95282ms
2024-07-25T23:24:35.437480Z DEBUG sqlx::query: summary="SELECT events.rowid, events.type, blocks.height …" db.statement="\n\nSELECT\n  events.rowid,\n  events.type,\n  blocks.height AS block_height,\n  tx_results.tx_hash,\n  jsonb_object_agg(attributes.key, attributes.value) AS attrs\nFROM\n  events\n  LEFT JOIN attributes ON events.rowid = attributes.event_id\n  JOIN blocks ON events.block_id = blocks.rowid\n  LEFT JOIN tx_results ON events.tx_id = tx_results.rowid\nWHERE\n  events.rowid > $1\nGROUP BY\n  events.rowid,\n  events.type,\n  blocks.height,\n  tx_results.tx_hash\nORDER BY\n  events.rowid ASC\n" rows_affected=0 rows_returned=1 elapsed=189.203787ms
Error: error occurred while decoding column 0: mismatched types; Rust type `i64` (as SQL type `INT8`) is not compatible with SQL type `INT4`

Caused by:
    mismatched types; Rust type `i64` (as SQL type `INT8`) is not compatible with SQL type `INT4`
```

which suggests we've still got some typecasting to sort out.

conorsch (approved):

PR #4774, opened 2024-07-26
Governance indexing

cronokirby:
This should supercede https://github.com/penumbra-zone/penumbra/pull/4708, now that the event changes have landed.

I took the structure in there, and munged it a bit to fit the current framework we use in the reset of the indexer code.

  > pindexer only

### DIFF
--- a/crates/bin/pindexer/src/governance/governance.sql
+++ b/crates/bin/pindexer/src/governance/governance.sql
@@ -1,0 +1,55 @@
+CREATE TABLE IF NOT EXISTS governance_proposals (
+  proposal_id INTEGER PRIMARY KEY,
+  title TEXT NOT NULL,
+  description TEXT NOT NULL,
+  kind JSONB NOT NULL,
+  payload JSONB,
+  start_block_height BIGINT NOT NULL,
+  end_block_height BIGINT NOT NULL,
+  state JSONB NOT NULL,
+  proposal_deposit_amount BIGINT NOT NULL,
+  withdrawn BOOLEAN DEFAULT FALSE,
+  withdrawal_reason TEXT
+);
+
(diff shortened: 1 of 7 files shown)


hdevalence:
Closing in favor of the original work in #4708 which powers the governance frontend.

crates/bin/pindexer/src/governance/mod.rs
```diff
@@ -0,0 +1,411 @@
...
+    core::component::governance::v1::{self as pb},
+    event::ProtoEvent,
+};
+use penumbra_stake::IdentityKey;
+use sqlx::{PgPool, Postgres, Transaction};
+
+/// One of the possible events that we care about.
+#[derive(Clone, Debug)]
+enum Event {
+    ProposalSubmit {
```
hdevalence, line 20: We don't need to define an event enum here. Processing the events, as in the original code, works fine.

crates/bin/pindexer/src/governance/governance.sql
```diff
@@ -0,0 +1,55 @@
...
+  block_height BIGINT NOT NULL,
+  FOREIGN KEY (proposal_id) REFERENCES governance_proposals(proposal_id)
+);
+
+CREATE INDEX ON governance_delegator_votes (proposal_id);
+CREATE INDEX ON governance_delegator_votes (identity_key);
+CREATE INDEX ON governance_delegator_votes (vote);
+CREATE INDEX ON governance_delegator_votes (voting_power);
+CREATE INDEX ON governance_delegator_votes (block_height);
+
```
hdevalence, line 55: Not doing this was already discussed in the original PR: https://github.com/penumbra-zone/penumbra/pull/4708#discussion_r1678000354

PR #4776, opened 2024-07-29
feat: swaps indexing

### LINKED ISSUE #4744: pindexer: dex: implement `/api/swaps` related events and tables
We need to get to parity with:
https://github.com/penumbra-zone/dex-explorer/tree/2b70ae60529284309f8748f6ca6a154a2bab82a9/src/pages/api/swaps

vacekj:
## Describe your changes
implements indexing of swaps to satisfy the `/api/swaps` endpoint

- [x] There's also an event for Swap, so we should name the event in the code to BatchSwap to avoid confusion.
- [x] We probably want to record some more information in the swap table, like having the trading pair there, the total amount of inputs and outputs for each asset, the unfilled amounts, et.c

fixes #4744

  Indexer-only changes

### DIFF
--- a/crates/bin/pindexer/src/dex/dex.sql
+++ b/crates/bin/pindexer/src/dex/dex.sql
@@ -117,7 +117,7 @@ CREATE TABLE IF NOT EXISTS dex_lp_execution (
 );
 
 --- Represents instances where swap executions happened.
-CREATE TABLE IF NOT EXISTS swap
+CREATE TABLE IF NOT EXISTS dex_swap
 (
     height      BIGINT PRIMARY KEY,
     input       Value,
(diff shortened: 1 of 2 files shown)


cronokirby (changes requested):
We probably want to record some more information in the swap table, like having the trading pair their, the total amount of inputs and outputs for each asset, the unfilled amounts, et.c

cronokirby:
There's also an event for `Swap`, so we should name the event in the code to `BatchSwap` to avoid confusion.

crates/bin/pindexer/src/dex/dex.sql
```diff
@@ -113,3 +115,21 @@ CREATE TABLE IF NOT EXISTS dex_lp_execution (
   -- The end asset for this execution.
   context_end BYTEA NOT NULL
 );
+
+--- Represents instances where swap executions happened.
+CREATE TABLE IF NOT EXISTS swap
+(
+    height      BIGINT PRIMARY KEY,
+    input       Value,
+    output      Value,
```
cronokirby, line 124: I think these are redundant if we have the delta amounts

vacekj: removed.

zmanian:
I believe this PR would unlock using the native indexer to get things like 24 hour swap volumes which would be really helpful to go to market efforts.

How can I support getting it merged?

cronokirby (approved):

PR #4782, opened 2024-08-01
build(pd): fail if asset zipfiles aren't complete

### LINKED ISSUE #4703: Confusing error / DoS bug when pd is built without git lfs
**Describe the bug**

When pd is built without git lfs, the embedded mini front archive is not a valid zip file and so the node panics when trying to serve from it. The error message gives no indication that it's related to the build config.

I think we could define the DoS part as ok (don't build pd like that) and just improve the error message to start, then possibly circle back later

conorsch:
## Describe your changes
Updates the build.rs file for pd to error out if the frontend asset bundles are incomplete, by checking for a small filesize. That isn't a super durable method, but it's a cheap check and easy to update in the future.

Closes #4703.

  > changes to build logic, does not affect app consensus logic
  
## Testing and review

CI currently checks the happy-path, meaning it ensures that LFS is used for building Penumbra artifacts. If you're reviewing a Penumbra PR, your local dev env likely has git-lfs properly configured, too. To test the unhappy-path, try saving this Dockerfile locally:

```dockerfile
FROM docker.io/rust:1-slim-bookworm
RUN apt-get update && apt-get install -y git curl clang libssl-dev pkg-config
WORKDIR /app
RUN rustc --version
RUN git clone https://github.com/penumbra-zone/penumbra && \
        cd /app/penumbra && \
        git checkout 4703-pd-detect-non-lfs-zip-assets
WORKDIR /app/penumbra
RUN cargo build --release
```

and building it with `docker build -t tmp-4703-review .`. You should see an error message like:

```
error: failed to run custom build command for `pd v0.79.3 (/app/penumbra/crates/bin/pd)`

Caused by:
  process didn't exit successfully: `/app/penumbra/target/release/build/pd-fbb290712e0b8aed/build-script-build` (exit status: 1)
  --- stderr
  Error: asset zip file /app/penumbra/crates/bin/pd/../../../assets/minifront.zip is smaller than 500 bytes; install git-lfs, run 'git lfs pull', and
 retry the build
warning: build failed, waiting for other jobs to finish...
Error: building at STEP "RUN cargo build --release": while running runtime: exit status 101
```

### DIFF
--- a/crates/bin/pd/build.rs
+++ b/crates/bin/pd/build.rs
@@ -1,9 +1,47 @@
+use std::io::Read;
 use std::path::Path;
 
 use anyhow::Context;
 
 fn main() -> anyhow::Result<()> {
+    check_frontend_asset_zipfiles()?;
     setup_testnet_config()?;
+    Ok(())
+}
+
+// Check that the zip files for bundled frontend code are functional.
+// If git-lfs is not configured on the build host, the zip files will
+// be plaintext lfs pointer files.
+fn check_frontend_asset_zipfiles() -> anyhow::Result<()> {
+    // Declare a minimum filesize, below which we'll assume the zip file is
+    // actually a git-lfs pointer.
+    const MINIMUM_FILESIZE_BYTES: usize = 500;
+    // Build paths to the zip files in the local build env.
+    let zipfiles = vec![
+        Path::new(env!("CARGO_MANIFEST_DIR")).join("../../../assets/minifront.zip"),
+        Path::new(env!("CARGO_MANIFEST_DIR")).join("../../../assets/node-status.zip"),
+    ];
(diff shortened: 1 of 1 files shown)


cronokirby (approved):
Looks good to me; maybe we should add a subsequent PR to replicate the proof parameter trick of reading the LFS file in build.rs so that it "just works".

conorsch:
> maybe we should add a subsequent PR to replicate the proof parameter trick of reading the LFS file 

Yeah, I almost went this route, then backed off when I was tempted to generalize that logic rather than just copy/paste it between build.rs files. Good point. Overall, I think we'll be forced to solve issues like that comprehensively when we try to publish all the protocol repo crates to crates.io.

PR #4787, opened 2024-08-02
fix(ibc)!: handle ibc withdrawals correctly

conorsch:
## Describe your changes
This commit adds a `use_compat_address` bool to withdrawals, to allow issuing compat address withdrawals. It also implements correct error handling on withdrawal attemtps when counterparty chains return an error.

  > This change does affect consensus, and should only be released after considerations have been made for compatibility.


crates/core/component/shielded-pool/src/component/transfer.rs
```diff
@@ -540,14 +541,30 @@ impl AppHandlerExecute for Ics20Transfer {
...
+        mut state: S,
+        msg: &MsgAcknowledgement,
+    ) -> Result<()> {
+        let ack: TokenTransferAcknowledgement =
+            serde_json::from_slice(msg.acknowledgement.as_slice())?;
+        if !ack.is_successful() {
+            // in the case where a counterparty chain acknowledges a packet with an error,
+            // for example due to a middleware processing issue or other behavior,
+            // the funds should be unescrowed back to the packet sender.
+            timeout_packet_inner(&mut state, &msg.packet)
```
zmanian, line 561: So the logic in the ics 20 specification for handling error acknowledgements and for handling timeouts appears identical. 

I do wonder if it would be more maintainable to make the code a bit more like spec.

Otherwise looks good

zmanian (approved):

crates/core/component/shielded-pool/src/component/transfer.rs
hdevalence: @avahowell can comment more here but I think the goal was to have a minimal diff that used the existing (audited) code as much as possible

avahowell: Yes, I checked and verified that the behavior should be identical, then kept the diff minimal in order to minimize risk and maximize readability

cronokirby (approved):
I went through all the files and this seems to correctly achieve that:

1. that all outgoing ICS20 packets use the `penumbracompat1...` Bech32 encoding,
2. that packet acknowledgements with errors are treated as timeouts.

One point of uncertainty I had which I managed to clarify away was whether or not we would be able to correctly handle `FungibleTokenPacketData` containing a `penumbracompat1` address.
We can, because our handling of it uses the [`FromStr`](https://rustdoc.penumbra.zone/main/penumbra_keys/address/struct.Address.html#impl-FromStr-for-Address) method for `Address`, which will handle both `penumbracompat1` and `penumbra1`.

crates/bin/pd/src/migrate/mainnet1.rs
```diff
@@ -0,0 +1,175 @@
+//! Migration for shipping consensus-breaking IBC changes, fixing
+//! how withdrawals from Penumbra to Noble are handled, and ensures that IBC
+//! error messages from counterparty chains are processed.
+//!
+//! There's no state-altering logic in this migration: the migration simply
+//! flips the halt bit off, permitting the chain to restart post-upgrade.
```
hdevalence, line 6: This comment seems no longer accurate

crates/bin/pd/src/migrate.rs
```diff
@@ -52,6 +53,9 @@ pub enum Migration {
     /// - Truncate various user-supplied `String` fields to a maximum length.
     /// - Populate the DEX NV price idnexes with position data
     Testnet78,
+    /// Mainnet-1 migration:
+    /// - Reset the halt bit
```
hdevalence, line 57: This comment seems no longer accurate

hdevalence (approved):
LGTM

PR #4789, opened 2024-08-03
refactor(governance): update StateWriteExt::signal_upgrade comment

syvb:
## Describe your changes

`StateWriteExt::signal_upgrade` is infallible and doesn't await on anything, so its function signature can be simplified to not be async and not return a `Result<()>` that's always `Ok(())`.

Also fixes a comment referring to `TOTAL_HALT_COUNT`, which is now a boolean and not a count.

  > No change to behavior


cronokirby:
The comment change is good; for the other change, right now it can be done, but I like the ability to add fallibility and asyncness to the inner logic without needing to change the callsites. For example, if we need to read some state, for example.

syvb:
@cronokirby I updated the PR to just update the comment.

cronokirby (approved):

PR #4796, opened 2024-08-07
build: fix nix-build via allowing asset files

conorsch:
## Describe your changes
Adds a comprehensive allow-list of file extensions to the nix build context. Includes zip files, which will fail a build as of #4782, and other flat file inputs like CSV and JSON for testnet generation. Also adds the `pindexer` binary to the nix-build outputs.

Refs #4782. Doesn't close it, because it'd be nice to tack on a CI job that guards against breakage. Will follow up with that separately.

  > build logic only, no changes to app code
  
## Testing and review

1. Make sure you've got nix installed (see [dev env docs](https://guide.penumbra.zone/dev/dev-env) if you need help)
2. Run `nix build` in the repo root
3. Ensure no errors, exits 0.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -64,8 +64,13 @@
                 sha256 = "${penumbraRelease.sha256}";
               };
               filter = path: type:
-                # Retain proving and verification parameters, and no-lfs marker file ...
-                (builtins.match ".*\.(no_lfs|param|bin)$" path != null) ||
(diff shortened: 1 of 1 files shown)


mitchmindtree:
Can confirm `main` doesn't work for me, but this branch does!

Tested with `nix shell github:penumbra-zone/penumbra/4790-fix-nix-build` on nixos.

PR #4797, opened 2024-08-08
Basic IBC Handshake Test

### LINKED ISSUE #3758: tests: ❓ implement an IBC handshake using mock consensus engine
stub issue.

see #3588.

once we have the basics of our consensus engine working, we should proceed to provide facilities to test IBC functionality. let's start by writing a test that an IBC handshake works properly between two chains.

zbuc:
## Describe your changes

This adds a basic IBC handshake test, using the existing mock client. Some opportunistic refactoring is also included.

This PR adds a new `tests::common::ibc_tests` module, which contains a `MockRelayer` that can be extended later on.

Follow-up tasks should be basic transfer testing, transfer timeout testing, and testing with malformed requests.

While debugging this test, bugs were found in the various IBC query APIs, specifically that the `proof_height` was consistently being returned one lower than the height whose header would contain the app_hash necessary for validating the proof. The Go relayer is unaffected because it uses the ABCI RPC query interface instead, and Hermes uses the affected APIs but discards the affected `proof_height` fields and uses its own internal mechanisms for height tracking instead. Fixes were included for the affected APIs.

## Issue

Closes #3758

  > There are changes to TendermintProxy and IBC RPC responses however anyone using the affected RPCs would quickly run into the issues we saw in testing that revealed the bugs in the RPC responses, and the chain would properly block any requests made based on the incorrect response values. The changeset also affects the test code heavily however there should be nothing that affects consensus or state.


crates/core/app/tests/common/ibc_tests/mod.rs
```diff
@@ -0,0 +1,193 @@
...
+pub use relayer::MockRelayer;
+
+mod node;
+pub use node::TestNodeWithIBC;
+use serde::Deserialize;
+use tendermint::{consensus::params::AbciParams, public_key::Algorithm, Genesis};
+
+/// Collection of all keypairs required for a Penumbra validator.
+/// Used to generate a stable identity for a [`NetworkValidator`].
+/// TODO: copied this from pd crate
```
conorsch, line 25: N.B. I've long wanted to pull out most of the network code from pd into a new crate `penumbra-network`, and in the process clean up the interfaces for wrangling the validator keyrings.

crates/bin/pd/src/network/config.rs
```diff
@@ -175,6 +175,48 @@ pub struct ValidatorKeys {
 }
 
 impl ValidatorKeys {
+    /// Use a hard-coded seed to generate a new set of validator keys.
+    pub fn from_seed(seed: [u8; 32]) -> Self {
```
conorsch, line 179: Very cool to have this utility!

crates/bin/pd/src/network/config.rs
```diff
@@ -198,7 +240,7 @@ impl ValidatorKeys {
         let validator_cons_pk = validator_cons_sk.public_key();
 
         // generate P2P auth key for tendermint.
-        let node_key_sk = ed25519_consensus::SigningKey::new(OsRng);
+        let node_key_sk = ed25519_consensus::SigningKey::from(seed.0);
```
conorsch, line 243: I assume the hardcoding of a static seed is leftover from debugging—for the actual test suite, we should probably leave the keygen to be random as usual. Fine for now, because I know there'll be a lot of work coming on top of this. Mentioning it now in case you disagree with the end goal.

crates/core/component/ibc/src/component/rpc/connection_query.rs
```diff
@@ -58,16 +59,19 @@ impl<HI: HostInterface + Send + Sync + 'static> ConnectionQuery for IbcQuery<HI>
...
+                proof: proof.encode_to_vec(),
+                proof_height: Some(ibc_proto::ibc::core::client::v1::Height {
+                    revision_height: snapshot.get_block_height().await.map_err(|e| {
+                        tonic::Status::aborted(format!("couldn't decode height: {e}"))
+                    })? + 1,
+                    revision_number: HI::get_revision_number(&snapshot).await.map_err(|e| {
+                        tonic::Status::aborted(format!("couldn't decode height: {e}"))
+                    })?,
+                }),
+            };
```
conorsch, line 74: Regarding the `get_block_height + 1` logic throughout, that unblocked so much of these state machine tests, we should be documenting those sharp edges more comprehensively. Not blocking merge, but IMO, explaining the difference in height queries in the ibc module docs makes the most sense: it'll be immediately obvious to any developers looking at that code that the various notions of height are important and distinct.

conorsch (approved):
Excellent work, dude. So glad to see this work coming together, and some query bugs squashed along the way. Please keep the in-line comments coming, because we discovered a lot of subtlety the past few days, and you're in a great position to reduce surprises for future maintainers by documenting your travails.

Issue #4792, opened 2024-08-06
Discovering the position id of a new LP position created with pcli is high friction


zmanian:
When creating an lp position with `pcli tx lp order buy`, you get a transaction_hex back but a more typical workflow would be to either monitor the position with `pcli query dex position` or with the dex explorer. 

To do that step, you need your position id. My current workflow is to call `pcli view balance` and look all my open position_nfts to find the one that was created.

**Describe the solution you'd like**

It would be useful to either have a position specific view that would basically show richer information about all open positions and/or it would be useful to have the `pcli tx lp order buy/sell` commands return a position_id.

hdevalence:
One very low effort improvement is that the command should print out the position ID before submitting the TX

cronokirby:
Resolved in https://github.com/penumbra-zone/penumbra/pull/4798, for now

PR #4798, opened 2024-08-09
Print the position id when generating a position

zmanian:
Resolves issue #4792 

Prints the position Id when creating a new liquidity position

### DIFF
--- a/crates/bin/pcli/src/command/tx.rs
+++ b/crates/bin/pcli/src/command/tx.rs
@@ -957,7 +957,7 @@ impl TxCmd {
                 let position = order.as_position(&asset_cache, OsRng)?;
                 tracing::info!(?position);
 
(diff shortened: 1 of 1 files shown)


crates/bin/pcli/src/command/tx.rs
```diff
@@ -957,6 +957,8 @@ impl TxCmd {
                 let position = order.as_position(&asset_cache, OsRng)?;
                 tracing::info!(?position);
 
+                println!("Position id:{}", position.id().to_string());
```
hdevalence, line 960:
```suggestion
                println!("Position id: {}", position.id());
```
Add a space and omit redundant `to_string`

hdevalence (approved):

cronokirby (approved):

PR #4799, opened 2024-08-09
pcli: print position ID when creating limit orders

hdevalence:
## Describe your changes

Changes pcli to print out a position ID when creating limit orders, making it easier for a user to crossreference with a DEX explorer.

No issue

  > only trivial changes to pcli


hdevalence:
Duplicate of #4798

PR #4801, opened 2024-08-11
pcli: add un-advertised support for importing Prax registry data

hdevalence:
This makes a minimal set of changes to allow `pcli` to make use of the Prax registry. In order to not pre-empt more comprehensive planning about how such an integration should work, this PR just changes `pcli` so that if the registry is placed in a `registry.json` file in the `pcli` home directory, it will be imported into the view database on startup.

  > client-side changes only

### DIFF
--- a/crates/core/asset/src/asset/denom_metadata.rs
+++ b/crates/core/asset/src/asset/denom_metadata.rs
@@ -30,6 +30,7 @@ pub struct Metadata {
 }
 
 // These are constructed by the asset registry.
+#[derive(Debug)]
 pub(super) struct Inner {
     // The Penumbra asset ID
     id: Id,
--- a/crates/core/asset/src/asset/denom_metadata.rs
+++ b/crates/core/asset/src/asset/denom_metadata.rs
@@ -317,16 +317,19 @@ impl Metadata {
         if amount == 0u64.into() {
             return self.default_unit();
         }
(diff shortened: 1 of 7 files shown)


crates/core/asset/src/asset/denom_metadata.rs
```diff
@@ -317,16 +318,19 @@ impl Metadata {
...
+            if unit_amount <= amount && unit.exponent >= selected_exponent {
+                selected_index = unit_index;
+                selected_exponent = unit.exponent;
             }
         }
-        self.base_unit()
+        return Unit {
+            unit_index: selected_index,
+            inner: self.inner.clone(),
+        };
```
hdevalence, line 333: This bugfix needs a unit test.

It can be triggered by having a Metadata object where the list of units is not sorted. The previous code is incorrect on, for instance, the USDC Metadata currently in the Prax registry. That JSON object could be copied into a unit test, which can create a value like "1.234 USDC", parse it, and check that displaying it round trips (because the correct unit is selected).

hdevalence:
This doesn't fix problems with argument parsing, a lot of pcli arguments specify denoms and have the string parsing done inside clap, which has to be stateless. Full support will need to change this to take string args and then parse after the app is loaded and the registry can be pulled from the view database. But that can be future work

PR #4806, opened 2024-08-13
modify arkworks and decaf377 suite

TalDerei:


cronokirby:
For posterity what's the point of increasing the iteration count?

TalDerei:
> For posterity what's the point of increasing the iteration count?

less variation in the runtimes

conorsch (approved):

PR #4810, opened 2024-08-15
Cleanup tendermint proxy code, fix proxy timestamp handling

zbuc:
## Describe your changes

This PR cleans up the tendermint proxy code by removing the usage of tonic error types and fixes a bug in the proxy timestamp handling.

  > The timestamp handling change does not affect consensus because it only affects values returned to clients via the tendermint proxy, which is mostly unused. Any usage of these incorrect timestamps would be noticed as frequent and random failures which haven't been reported.


conorsch (approved):
LGTM!

PR #4811, opened 2024-08-15
Add `new_with_penumbra_prefixes` to the `TempStorageExt`

zbuc:
## Describe your changes

This adds a `new_with_penumbra_prefixes` function to the `TempStorageExt` trait, so that a `TempStorage` can be instantiated with the same storage prefixes that a real chain would use. It was added to the extension trait to avoid polluting the implementation inside the cnidarium crate with penumbra-specific details.

All existing tests have been changed to use the correct prefixes.

  > Only affects tests


conorsch (approved):
Changes are confined to test modules, so no concerns about contaminating types for external consumers. :+1:

Issue #4815, opened 2024-08-15
view service doesn't seem to populate auction related views sufficiently


cronokirby:
**Describe the bug**
If you query auction information view pcli, on a testnet with the hardcoded gm and penumbra assets, performing an auction from penumbra to gm will production auction views that are missing metadata about gm in the response, despite gm obviously being displayed correctly in other transaction views.

This seems to be related to how the transaction perspective is populated.
We probably forgot to add in related information when we were doing an auction pass.

**Additional context**

Uncovered in https://github.com/penumbra-zone/penumbra/pull/4814.

PR #4816, opened 2024-08-15
view: populate asset ids from dutch auction schedule / withdraw

cronokirby:
## Describe your changes

This populates the transaction perspective using the asset ids present in an auction, which improves display of auctions in transactions.

See #4815.

One thing missing to close the issue is that we don't correlate in the reserve information, but I'm not sure of a straightforward small change to do that.

  > view code only


conorsch (approved):
LGTM, and further action is tracked in https://github.com/penumbra-zone/penumbra/issues/4815.

Issue #3759, opened 2024-02-06
tests: ⏰ expose time facilities in mock consensus engine


cratelyn:
stub issue.

see #3588.

provide interfaces/facilities to move the "time" used to generate timestamps in the mock consensus engine.

PR #4818, opened 2024-08-15
Add ability to set fixed time and block duration callback in test node

zbuc:
## Describe your changes

This enhances test code reproducibility by allowing a fixed timestamp and block timing to be set rather than using system time.

#3759

  > only affects tests


aubrika (approved):
LGTM!

PR #4828, opened 2024-08-22
fix(tests): bind to unique ports in ibc tests

conorsch:
## Describe your changes
 Follow-up to #4747. Noticed the new tests were flaking across multiple runs. Turns out they were fighting for port binds, as described already in #4517. Let's just use a unique port pair for the MockRelayers and call it a day for now.

Refs #4517.

  > tests only

### DIFF
--- a/crates/core/app/tests/common/ibc_tests/node.rs
+++ b/crates/core/app/tests/common/ibc_tests/node.rs
@@ -96,7 +96,9 @@ impl TestNodeWithIBC {
             "b" => 1,
             _ => unreachable!("update this hack"),
         };
-        let grpc_url = format!("http://127.0.0.1:808{}", index) // see #4517
+        // We use a non-standard port range, to avoid conflicting with other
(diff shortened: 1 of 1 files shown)


conorsch:
Here's an example of a failing test run: https://github.com/penumbra-zone/penumbra/actions/runs/10515320975/job/29135123028 with error text:

```
--- STDERR:              penumbra-app::ibc_handshake ibc_handshake ---
thread 'ibc_handshake' panicked at crates/core/app/tests/common/ibc_tests/node.rs:120:47:
grpc server returned an error: Os { code: 98, kind: AddrInUse, message: "Address already in use" }
Error: could not connect to grpc server

Caused by:
    0: transport error
    1: error trying to connect: tcp connect error: Connection refused (os error 111)
    2: tcp connect error: Connection refused (os error 111)
    3: Connection refused (os error 111)
 ```

zbuc (approved):

PR #4830, opened 2024-08-23
Perform channel handshake in IBC state machine tests

zbuc:
## Describe your changes

This extends the IBC state machine tests to perform a channel handshake after the successful connection handshake. This is the final prerequisite to performing IBC transfers between chains in tests.

  > tests only


conorsch (approved):
Very cleanly done. Onward!

PR #4841, opened 2024-08-28
chore: update chain id for testnet 80

conorsch:
## Describe your changes

Contains a few small edits:

  * documents historical testnets
  * updates base allocations with new addresses
  * modifies testnet generation logic to match
  * generates new Testnet 80 config
  * updates dev env to use gas prices by default

Used these to generate local devnets while testing out the new allocation settings. Doing all this in preparation for launching a new public testnet.

N/A

  > no application code changes, only affects testnet fixtures


aubrika (approved):
LGTM 😎

Issue #4831, opened 2024-08-23
pcli: migrate balance function


aubrika:
We should change pcli to have a new tx subcommand, `pcli tx migrate`, which takes the _destination FVK_ as a parameter and then constructs a transaction that:

- [x] has the memo field “migrate from `penumbrafullviewingkey1…` to `penumbrafullviewingkey1…`”
- [x] spends every single note from every account
- [ ] replicates the balances per account in the wallet of the provided destination FVK (minus any transaction fees)
- [x] pays the fees for this transaction from the lowest-numbered account with sufficient funds
- [x] include help text in clap explaining that this is for migrating custody mechanisms

cronokirby:
Not quite done yet because we're not replicating the account structure

conorsch:
> replicates the balances per account in the wallet of the provided destination FVK (minus any transaction fees)

Ah, I didn't notice this was in the spec: thought the changes in #4842 were sufficient to close.

PR #4844, opened 2024-08-29
pmonitor: create client activity monitor

### LINKED ISSUE #4832: client: account activity monitor
We need a utility that can monitor the activity of one or more accounts, given only a full viewing key for each. 

The input is a JSON file containing `FullViewingKey`s, and the tool needs to be able to check that, for each account of each FVK, the current UM-equivalent balance of the account is greater than or equal to the genesis balance.

This will require:

- [ ] one-off scanning of the compact block with each FVK in order to compute the genesis allocations as they were in the genesis file
- [x] building a directory tree that can be pcli/pclientd home directories that have view databases in them and have the current state

In the (near) future the tool will also need to handle key rotations (transfer of balances from one account to another), so this work should be done to accommodate the work done to that end in #4831.

redshiftzero:
## Describe your changes

This PR adds a tool `pmonitor` that will enable one to monitor the activity of a list of `FullViewingKey`s

Plan:
- [x] Make a new tool `pmonitor` (because we need to use pclientd or pcli which assumes a single view key)
- [x] Load a JSON file containing a array of `FullViewingKey`s
- [x] Build a directory tree that has pcli/pclientd home directories that have view databases and have the current state
- [x] One off scanning of the compact block with each FVK to determine the genesis allocation for that FVK
- [x] For each FVK, check the current UM-equivalent balance for the balance is greater than or equal to the genesis balance
- [x] Print each FVKs, their genesis balance, and their current UM-equivalent balance
- [x] Handle migrated accounts 

Will close #4832

## Test

1. Point to the desired RPC and provide a JSON array of FVKs: `cargo run --bin pmonitor init --fvks test.json --grpc-url https://yourtestnetrpchere.example.com`

Expected state: you should have in your `zone.penumbra.pmonitor` home directory (stored next to wherever your pcli default homedir is, so on my Macbook it's `"/Users/redshiftzero/Library/Application Support/zone.penumbra.pmonitor"`) a:
- config file 
- for each FVK, a wallet_{uuid} directory 

2. Run `cargo run --bin pmonitor audit`:
```
Syncing wallet: /Users/redshiftzero/Library/Application Support/zone.penumbra.pmonitor/wallet_3
Scanning blocks from last sync height 286483 to latest height 286483
Wallet synced successfully
FVK: penumbrafullviewingkey1[...snip...]
Genesis UM-equivalent balance: 0
Current UM-equivalent balance: 2999999066
✅ Expected balance! Balance is greater than or equal to the genesis balance
...
```

This will first sync all the wallets, then check the balances. Subsequent executions of the `audit` command will start with the state from the previous run.

*Note:* To reset after testing:
`cargo run --bin pmonitor reset`

  > This is a client-side tool only


conorsch:
We just shipped the pcli-migrate-balance functionality (#4831), which enables full-wallet transfers, based on FVK. The `pmonitor` tool should recognize a migrated balance, which can be identified by a hardcoded memo: https://github.com/penumbra-zone/penumbra/blob/v0.80.4/crates/bin/pcli/src/command/migrate.rs#L105

aubrika:
Sounds like this will be wrapped up or close this week, and since this is a read-only op, whoever reviews this should be sure to test against both testnet and mainnet. I can help generate a list of FVKs and checking the implementation should be straightforward from there.

redshiftzero:
status: this now has support for migrated accounts, but needs testing before marking ready for review

conorsch:
I'll take a first pass at review here, confirming that the tool 1) correctly identifies when accounts have been spent out of; and 2) supports the pcli-migrate-balance flow.

crates/view/src/storage/schema.sql
```diff
@@ -54,7 +54,8 @@ CREATE TABLE tx (
     tx_hash                 BLOB PRIMARY KEY NOT NULL,
     tx_bytes                BLOB NOT NULL,
     block_height            BIGINT NOT NULL,
-    return_address          BLOB
+    return_address          BLOB,
+    memo_text               TEXT
```
conorsch, line 58: Schema change! This breaks existing clients. Testing locally against testnet, I do indeed see the error message:

> `Error: can't load view database created by client version 0.80.5 using client version 0.80.5: they have different schemata, so you need to reset your view database and resynchronize by running pcli view reset`

That's probably tolerable for a point-release, but we should be careful about downstream consequences. I'm thinking in particular of relayers like `hermes`, which will likely fail when their deps are upgraded to a more recent view crate. I see we already merged a schema change in https://github.com/penumbra-zone/penumbra/pull/4801 so we'll need to accommodate for the change already anyway. Relevant for rebasing the hermes fork, cc @zbuc.

crates/bin/pmonitor/src/main.rs
```diff
@@ -0,0 +1,465 @@
...
+                    println!(
+                        "Current UM-equivalent balance: {:?}",
+                        total_um_equivalent_amount
+                    );
+
+                    // Let the user know if the balance is unexpected or not
+                    if total_um_equivalent_amount < genesis_um_equivalent_amount {
+                        println!("✘ Unexpected balance! Balance is less than the genesis balance");
+                    } else {
+                        println!("✅ Expected balance! Balance is greater than or equal to the genesis balance");
```
conorsch, line 452: Love the emojis, makes the wall of text a bit easier to scan visually. I only tested with 10 FVKs, though, and most of the relevant content was off the screen. Maybe `tracing` statements would be more appropriate here.

conorsch (changes requested):
In terms of functionality, the tool appears to work well: in all the testing I've done, it correctly notates when balance is absent. Still more interactive testing to perform, but the implementation seems sound!

UX-wise, however, the tool obscures detection of misbehavior: any indication of a misbehavior is likely lost to scrollback in the default stdout, where there are more than, say, 5 FVKs to scan. Worse, the tool exits 0 regardless of whether misbehavior was detected. Suggestions to resolve:

1. emit a "summary" message at the end of the `audit` run, notating explicitly if any accounts were found to have funds missing;
2. designate a specific exit code, e.g. `10`, for signaling that misbehavior was detected; that will allow the tool to be run in headless environments on a schedule, without resorting to string parsing;
3. (optional) convert the per-account messaging to tracing-subscriber statements, subject to RUST_LOG tweaking, to enable focusing on the summary message.

I left some in-line comments throughout, and also pushed up a script I've been using locally to test on a devnet. The biggest complaint I have about the technical bits are that the tool assumes it can `cargo run --bin pcli`, which I'd prefer to address before merge.

conorsch:
Rebased on latest main, and added a handful of commits. Logging is now more verbose, but still TK are summary message and custom exit codes. During testing, I noticed that the tool will report an error if an account was migrated: we should refactor the audit logic to continue searching the migrated account, without reporting an error, or else explicitly state that a new migration was detected and therefore the tool should be rerun.

conorsch:
Looking great. I added some tests, and made some changes to implement the requested features, namely 1) summary message at end of run, and 2) exiting non-zero on failure. 

## Example output from passing `audit`
pmonitor-pass

## Example output from failing `audit`
pmonitor-fail

Also tacked on some module docs to explain how it's intended to be used. Haven't added any content to the [guide](https://github.com/penumbra-zone/guide), but we can do that in a follow up if it's useful.

conorsch (approved):

redshiftzero:
thank you for adding these changes @conorsch, I've reviewed them (as of eec9d2507b169ae5a60f39d29165d608fbfaf345) and they LGTM

PR #4854, opened 2024-09-13
cometindex: speedup by committing event changes in batches of 1000

cronokirby:
Instead of creating one transaction for each event we need to index, we instead only close this transaction every 1000 events (or when when we've caught up to the database).

This gives about a 5x performance in catch up speed.

  > indexing only

### DIFF
--- a/crates/util/cometindex/src/indexer.rs
+++ b/crates/util/cometindex/src/indexer.rs
@@ -152,6 +152,7 @@ impl Indexer {
         let mut relevant_events = 0usize;
 
         let mut es = read_events(&src_db, watermark);
+        let mut dbtx = dst_db.begin().await?;
         while let Some(event) = es.next().await.transpose()? {
             if scanned_events % 1000 == 0 {
                 tracing::info!(scanned_events, relevant_events);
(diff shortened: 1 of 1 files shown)


crates/util/cometindex/src/indexer.rs
```diff
@@ -179,7 +180,6 @@ impl Indexer {
             relevant_events += 1;
 
             // Otherwise we have something to process. Make a dbtx
```
zbuc, line 182: nit: comment is now out of date

crates/util/cometindex/src/indexer.rs
```diff
@@ -188,8 +188,15 @@ impl Indexer {
             }
             // Mark that we got to at least this event
             update_watermark(&mut dbtx, event.local_rowid).await?;
-            dbtx.commit().await?;
+            // Only commit in batches of <= 1000 events, for about a 5x performance increase when
+            // catching up.
+            if scanned_events % 1000 == 0 {
```
zbuc, line 193: it seems possible to miss this if the while loop short circuits from the check at line 172, resulting in more than 1000 events being committed at once

cronokirby: Switched to using the relevant_events as the metric, so that you commit every time you've processed 1000 relevant events, avoiding this issue.

conorsch:
Requested changes have been made, so I'm approving and merging.

PR #4860, opened 2024-09-17
governance: add `overflow-checks` crate-wide

redshiftzero:
## Describe your changes

Adds a guard against underflow/overflows for the `penumbra-governance` crate

  > Provided that there are not underflows/overflows, this should not be consensus-breaking

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -85,6 +85,12 @@ x86_64-unknown-linux-gnu = "buildjet-32vcpu-ubuntu-2204"
 [profile.dist]
 inherits = "release"
 
+[profile.dev.package.penumbra-governance]
+overflow-checks = true
+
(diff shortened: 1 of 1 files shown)


cronokirby (approved):

conorsch:
> Provided that there are not underflows/overflows, this should not be consensus-breaking

What kind of additional assurance can we perform to ensure that there are not underflows or overflows in the wild? Would full confidence require syncing a node built with the `Cargo.toml` from this PR against e.g. mainnet data?

conorsch (changes requested):
Requesting changes to block merge, since technically this is a consensus-breaking diff. Let's return to this diff once there are plans for a chain upgrade, at which time it would be acceptable to ship breaking changes.

PR #4865, opened 2024-09-21
feat: Noble forwarding address registration in pcli

### LINKED ISSUE #4857: Noble-Penumbra auto-deposit prototyping
Noble Express is shipping signerless autoforwarding soon, and we can potentially capitalize on this functionality to enable signerless autoforwarding to Penumbra.

The existing forwarding docs are here: 

https://docs.noble.xyz/modules/forwarding

And the described functionality will be extended to accept `nobled tx forwarding register-account-signerlessly` , which option is already live on their testnet.

The current forwarding protos are found here: https://github.com/noble-assets/forwarding/blob/v1.x/proto/noble/forwarding/v1/tx.proto (these will change with their 0.50 release in November)

On our side, we will at the very least need to set up IBC channels for testing, and we should do that first to see if it works for us out of the box - if we need additional changes, we can track them here as we discover them.

zbuc:
## Describe your changes

This adds support for registering Noble forwarding addresses, for example:

`pcli tx register-forwarding-account --noble-node http://noble-testnet-grpc.polkachu.com:21590 --channel channel-216 --address-or-index 2`

Note that your address will need to be funded prior to being registered; the [faucet](https://faucet.circle.com/) can be used along with the `pcli view noble-address ADDRESS_OR_INDEX --channel channel-216` command to get test funds.

Closes #4857

  > pcli only


cronokirby:
Do we want to do the auto-registration flow as a part of this PR, or in a subsequent one?

cronokirby (approved):

zbuc:
> Do we want to do the auto-registration flow as a part of this PR, or in a subsequent one?

IMO a subsequent PR. I think what is implemented in this PR is useful for testing and maybe some particular use cases, but we should make the default be to use the automatically cycling addresses to be implemented in a follow-up PR.

PR #4866, opened 2024-09-23
pindexer supply with destruction

cronokirby:
## Describe your changes

This adjusts the total supply indexer to account for whether or not value is locked in the dex, the auction component, or locked away after fees and arbitrage.

This matters because a significant amount of the native token has been locked in arbitrage, so this affects the end result by about 50% (in terms of net new supply)

Merge https://github.com/penumbra-zone/penumbra/pull/4863 first.

  > indexing only.


aubrika (approved):
👍🏻

PR #4870, opened 2024-09-30
Pindexer auction

cronokirby:
## Describe your changes

This adds an auction indexing component to pindexer.

Also, I added a utility function to the auction component to get the initial state of an auction from its description; this is useful so that the indexer always has some state of an auction.
I did not make the consensus code actually *use* this new function, but a subsequent refactor PR to use it would be nice to avoid code duplication, but in the interest of not adding new review burden, this PR does not do that.

Instead, reviewers, please check that I only moved code around, and that I only made more things visible in the crate.

  > indexing changes + above comments on the auction component.


cronokirby:
Closing in light of https://github.com/penumbra-zone/penumbra/issues/4883

Issue #4871, opened 2024-09-30
insights dashboard


aubrika:
Following design discussion, we will need:

- [x] an additional indexer query to filter ICS20 IBC events and read the relevant data
- [x] a minimal frontend displaying:
- - [x] Total supply of UM + Market Cap (using the native price of UM:USDC)
- - [x]  Number of unique addresses that have deposited into the protocol (addresses on counterparty Cosmos chains, not on Penumbra)
- - [x] Cumulative and Total Value Shielded, broken out by asset, with number of unique depositor addresses for that specific asset (e.g., how many addresses deposited USDC? how many deposited SHITMOS?)
- [x] container deployment(s) for a prod-facing indexer and the minimal frontend

PR #4874, opened 2024-10-01
feat: events for ibc transfers

cronokirby:
## Describe your changes

This adds specific events related to fungible token transfers, recording them as they happen in the shielded pool.

First of all, this is a much more legible and readily consumable event around actual IBC events we care about, which is reason enough to have this, imo, but also, there's currently a flaw in the whole event system in that we don't have access, through events only, to the actual acknowledgement data for a packet.

In particular, we can't tell using the raw ibc events if a packet was acked successfully, finalizing a transfer, or unsuccessfully, causing the transfer to be refunded.
Knowing which of the two matters a lot, and will cause queries like "how much of this asset has been locked in the shielded pool" to not return the right result by quite a bit.

  > Event addition only


aubrika (approved):
LGTM

cronokirby:
Note: there is a bug in this and the related PR for refunds, where the receiver address for refunds is populated incorrectly.

conorsch:
Refs https://github.com/penumbra-zone/penumbra/issues/4871

conorsch (approved):

PR #4875, opened 2024-10-01
Transfer events backport

cronokirby:
A backport of #4874 to v0.79.X, for reindexing purposes.


aubrika (approved):
LGTM

conorsch:
#4874 has been merged into main; @cronokirby can I trouble you to cherry-pick that squashed commit onto this branch? I tried myself, and wasn't sure about the conflict resolution, so I defer to you with your familiarity of the patch.

conorsch (changes requested):
requesting cherry-pick of recently merged commit

cronokirby:
The commit will have to be different because the IBC code changed in the meantime to have refunds for both transfers and error acks, so the logic of where you put the event emission is different; I think squashing the commits on this branch achieves the same effect as cherry picking and then having to modify the commit on main, idk

conorsch:
Thanks, I accept that interpretation. I guess what I _really_ care about is: do you believe this PR, #4875, is right now 100% suitable for squash-merge as into backport branch? We can discuss in detail on Monday, just wanted to confirm it's good to go, given both the revisions and the need for conflict resolution when re-generating the patch.

cronokirby (commented):
I checked the tricky bit again @conorsch, this has everything we need.

conorsch (approved):
let's do it :+1:

PR #4878, opened 2024-10-02
Implement auto-cycling of Noble deposit addresses

### LINKED ISSUE #4873: Automatically rotate Noble deposit addresses
Building on #4857, now that we have a basic implementation of Noble forwarding address registration, they should be rotated automatically on use.

image

We can keep this simple and use the existing work from #4857 to implement speculative registration:

To figure out the next sequence number, assuming the view service doesn't have it available, submit a register tx starting at index 0 until you get a "account doesn't exist" error, this is the next one in the sequence.

To poll: submit the register tx repeatedly until you get the "successfully registered" message instead of the "doesn't exist" error.

For now, the sequence number should be stored in the view service. We can figure out how to expose it to Prax later.

zbuc:
## Describe your changes

This introduces changes to rotate Noble addresses based on an incrementing sequence number, as well as renames the `pcli tx register-forwarding-account` command as `pcli view noble-address` to provide a better experience:

```
$ pcli view noble-address 0 --channel channel-221 --noble-node http://noble-testnet-grpc.polkachu.com:21590

next one-time use Noble forwarding address for account 0 is: noble1...

please deposit funds to this address...

awaiting deposit...


💫💫💫💫💫💫💫💫💫💫💫💫💫💫💫💫💫💫💫💫💫💫💫💫💫💫💫💫💫💫💫💫💫💫💫💫

registered Noble forwarding account with address noble1... to forward to Penumbra address penumbra1......

your funds should show up in your Penumbra account shortly
```

Closes #4873 

  > client-only

### DIFF
--- a/crates/bin/pcli/src/command/view/noble_address.rs
+++ b/crates/bin/pcli/src/command/view/noble_address.rs
@@ -1,19 +1,33 @@
 use anyhow::Result;
-use rand_core::OsRng;
+use futures::future::{BoxFuture, FutureExt};
+use penumbra_proto::{
+    cosmos::tx::v1beta1::{
+        mode_info::{Single, Sum},
+        service_client::ServiceClient as CosmosServiceClient,
+        AuthInfo as CosmosAuthInfo, BroadcastTxRequest as CosmosBroadcastTxRequest,
+        Fee as CosmosFee, ModeInfo, SignerInfo as CosmosSignerInfo, Tx as CosmosTx,
+        TxBody as CosmosTxBody,
+    },
+    noble::forwarding::v1::{ForwardingPubKey, MsgRegisterAccount},
+    Message, Name as _,
+};
+use std::time::Duration;
 
-use penumbra_sdk_keys::{Address, FullViewingKey};
+use penumbra_sd_keys::{address::NobleForwardingAddress, keys::AddressIndex, Address, FullViewingKey};
+use tonic::transport::{Channel, ClientTlsConfig};
+use url::Url;
 
 #[derive(Debug, clap::Parser)]
 pub struct NobleAddressCmd {
-    /// The address to provide information about
+    /// The account that should receive forwarded funds.
     #[clap(default_value = "0")]
-    address_or_index: String,
-    /// Generate an ephemeral address instead of an indexed one.
-    #[clap(short, long)]
-    ephemeral: bool,
+    account: u32,
     /// The Noble IBC channel to use for forwarding.
     #[clap(long)]
     channel: String,
+    /// The Noble node to submit the forwarding registration transaction to.
+    #[clap(long)]
+    noble_node: Url,
 }
 
 impl NobleAddressCmd {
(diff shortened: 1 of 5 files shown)


crates/bin/pcli/src/command/view/noble_address.rs
```diff
@@ -22,31 +36,243 @@ impl NobleAddressCmd {
...
+                    channel,
+                    fvk,
+                    account,
+                )
+                .await
+            }
+            NobleRegistrationResponse::Success => {
+                // This means the midpoint had a deposit in it waiting for registration.
+                // This will "flush" this unregistered address, however the user still wants a new one, so return the midpoint + 1.
+                Ok(mid + 1)
```
grod220, line 118: What about these cases:
- Both sequence 23 and sequence 24 has funds in it. Binary search ends up hitting 23 first with a registration call. It succeeds in flushing the funds and then this function returns 24. We display that address to the user, but when we poll, we successfully flush that one too. Then the user attempts to deposit into that address. Will that deposit bounce as it's already registered? 
- What if we flush the max sequence number. Would this return max + 1?

crates/bin/pcli/src/command/view/noble_address.rs
```diff
@@ -22,31 +36,243 @@ impl NobleAddressCmd {
...
+            }
+            NobleRegistrationResponse::AlreadyRegistered => {
+                if left == mid || right == mid {
+                    // We've iterated as far as we can, the next sequence number
+                    // after the midpoint should be the next available sequence number.
+                    return Ok(mid + 1);
+                }
```
grod220, line 125: What if all sequence numbers have been registered? Will this return max + 1?

crates/bin/pcli/src/command/view/noble_address.rs
hdevalence: 1. The program should check the address before it is displayed to ensure it's fresh (could have been used by another instance on another machine)

2. Yes it should wrap around after 2^16 deposits

grod220: Re point 2:

With a binary search, how there could be a notion of wrapping around? All sequence numbers will have been exhausted. Or are you suggesting once all are registered, we just offer up sequence 0 to re-use?

crates/bin/pcli/src/command/view/noble_address.rs
```diff
@@ -22,31 +36,243 @@ impl NobleAddressCmd {
...
+    let code = r
+        .tx_response
+        .ok_or_else(|| anyhow::anyhow!("no tx response"))?
+        .code;
+
+    match code {
+        9 => Ok(NobleRegistrationResponse::NeedsDeposit),
+        0 => Ok(NobleRegistrationResponse::Success),
+        19 => Ok(NobleRegistrationResponse::AlreadyRegistered),
+        _ => Err(anyhow::anyhow!("unknown response from Noble")),
```
grod220, line 238: On web, we don't get the `code 19`, but below when submitting a tx with the same sequence fyi. Looks like a [cometbft mempool err](https://github.com/cometbft/cometbft/blob/4eb7c058dd99c092b932c79d7cae33010f57cad9/mempool/errors.go#L12).

```
{"code":-32603,"message":"Internal error","data":"tx already exists in cache"}
```

crates/bin/pcli/src/command/view/noble_address.rs
```diff
@@ -22,31 +36,243 @@ impl NobleAddressCmd {
...
+) -> Result<u16> {
+    // perform binary search to find the first unused noble sequence number
+    // search space (sequence number) is 2 bytes wide
+    let left = 0u16;
+    let right = 0xffffu16;
+    let mid = (left + right) / 2u16;
+
+    // attempt to register midpoint
+    _get_next_noble_sequence(left, right, mid, noble_node, channel, fvk, account).await
+}
```
grod220, line 76: Here is a web implementation of binary search, inspired by this, but handles edge cases a bit differently: https://github.com/prax-wallet/prax/pull/215/files#diff-c569e62334fe6b19c6a1434837ca111daa9f4eb8287e7ab0c4925564c2ff44c2

crates/bin/pcli/src/command/view/noble_address.rs
zbuc: Wrap-around is hard to do without state tracking, so I think using a random sequence number (and potentially notifying the user that it's being re-used) is a good idea

conorsch:
The Noble testnet has changed since this PR was first authored. We believe the encoding error has been resolved, but there's a temporary blocker in that the current Noble testnet is charging gas fees for registration transactions. The Noble team is aware of this regression and intends to address it, so that registration txs are gas-less again. Once that's done, we're unblocked to test again.

@zbuc Can you write a detailed test plan in the meantime? Then we can functionally verify this PR is working against latest Noble testnet prior to merge.

PR #4879, opened 2024-10-04
automated generation of random txplans for effecthash testing

### LINKED ISSUE #4613: add `EffectHash` test vectors
We should add test vectors for the `EffectHash` mechanism for transaction signing, i.e. we should generate a bunch of random `TransactionPlan`s, compute their `EffectHash`, and add them as test vectors. This will guard against additions of proto fields from breaking the effect hash as well as aid in development for others implementing transaction signing.

avahowell:
Closes #4613


redshiftzero:
This is ready for review: we now have the test utility generating all possible actions in a `TransactionPlan`, and to prevent regressions we test the `EffectHash` computation based on those tx plans in a new test

redshiftzero (approved):

redshiftzero:
Merging this and adding #4947 in a followup

PR #4886, opened 2024-10-07
proto: add extension trait for easier parsing of domain types

cronokirby:
## Describe your changes

This allows creating domain types for each of the events, e.g. `EventBatchSwap`, and then doing `EventBatchSwap::try_from_event(&event)`, which makes the pindexer app views much more ergonomic, especially because we can avoid having to duplicate event parsing logic across views.

In subsequent PRs, we should create domain types for each component.

We can probably do this as needed, when we write app views touching particular components.

  > Just an internal refactor to events code, so doubly non breaking.

### DIFF
--- a/crates/proto/src/event.rs
+++ b/crates/proto/src/event.rs
@@ -1,4 +1,4 @@
-use crate::{Message, Name};
+use crate::{DomainType, Message, Name};
 use anyhow::{self, Context};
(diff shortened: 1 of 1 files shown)


cronokirby:
I also tested that, given this trait, you can write AppViews that make use of this parsing mechanism.

aubrika (approved):
LGTM

PR #4892, opened 2024-10-08
Allow building client crates without rocksdb

### LINKED ISSUE #4885: app crate doesn't keep component code behind feature flag
**Describe the bug**
Other components keep the state machine part of the code behind a feature flag. The app crate doesn't. This means that any downstream consumer of the rust view server has to pull in and build the entire full node state machine and rocksdb.

cronokirby:
This required:
- feature gating `cnidarium` in `penumbra-app`
- removing the stateless transaction check in the view crate
- removing the ceremony contribution command from `pcli`

Closes #4885

  > Only client code was changed, the non-client code that was changed was only changed to have the public API be reduced without certain default features enabled. PD should be completely unchanged by this fix.
  
## How to test this:

run `cargo build --bin pcli -p pcli` (and pclientd, and pindexer) and observe that rocksdb is not built.


conorsch:
Great work. I recorded some of the research findings in #4885. Testing locally, `cargo build -p pcli` generates a lot of warnings, so I'll stick with `cargo build --bin pcli` myself for the nonce.

conorsch (approved):

cronokirby:
I think a lot of the warnings might just be that certain package code paths become unused without certain features

PR #4894, opened 2024-10-10
feat(pindexer): add dex candlestick support to schema

### LINKED ISSUE #4869: Index candlestick events in pindexer
Post https://github.com/penumbra-zone/penumbra/issues/4868, we should format the candlestick events into the dex component of pindexer to make generating charts for the dex explorer easy.

We should look into and measure to see if on the fly windowing functions are sufficient to provide windows of 1m, 1h, 1d, etc. or if we instead need to memoize these at the pindexer level.

cronokirby:
Merge #4893 first!

## Describe your changes

Closes #4869.

This implementation isn't as fast as it could probably be, but more than enough to be real time, which is all that really matters.

(Maintaining the summary is annoying and kind of tricky, especially if you want to correctly handle a "frozen price" for pairs where no activity is happening)

  > indexing only

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5829,6 +5829,7 @@ dependencies = [
  "penumbra-proto",
  "penumbra-shielded-pool",
  "penumbra-stake",
+ "prost",
  "serde_json",
  "sqlx",
  "tokio",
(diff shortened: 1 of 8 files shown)


conorsch:
> Merge https://github.com/penumbra-zone/penumbra/pull/4893 first!

Done. Although I squash-merged that one, which is kind of a shame, because your commit messages were super descriptive. Anyway, I've rebased this work on top of latest main, including the changes from #4893.

cronokirby:
thanks!

conorsch (approved):

PR #4898, opened 2024-10-18
pindexer: indexing for insights dashboard

### LINKED ISSUE #4883: pindexer: schema refactor
We should bring the current indexer data model in line with the proposed vision, supporting one or more tables per AppView, which are closely associated with product UI elements. The current structure is highly agnostic about what kinds of queries someone might like to write, but we want to be more opinionated about pre-processing and projecting data, to enhance ergonomicity for external developers.

To this end, rather than bucketing raw event data based on event type and doing projections or views of that data via specific queries, each `AppView` should populate tables which correspond more directly to the way the data will be displayed on the frontend. This should ideally make the internals of the indexer more legible to new developers, while also making it easier to query for desired data.

We can further reflect the desired relations in the organization of the database itself, using the `database.schema.table` hierarchy provided by pgSQL. Discussion revealed that it may be somewhat tricky to decide where to delineate schemas, but even creating a single schema per `AppView` in `init_chain` would add a level of organization that could make the data structure easier to parse to someone seeing it for the first time. (I would call this a nice-to-have if it is particularly onerous, since throwing everything in a global schema is how it's already working, but it would feel very clean to have something like `insights.supply.total_unstaked` or `insights.ibc.external_supply` in terms of table namespacing and DB navigation)

The apps this data model should initially provide tables for are:

- [x] #4896
- [x] voting dashboard
- [x] dex explorer

These tables can be tacked on initially as additive changes to pindexer, so the changes can be merged in without breaking existing deployments, then followed up by pointing the apps at the new tables and deprecating the old.

This is a refactor rather than greenfield development so no new tooling decisions are expected.

### LINKED ISSUE #4896: pindexer: insights dashboard support
Following internal discussion, some design notes accepted for the insights dashboard:

- [ ] Pindexer should have a new AppView specifically for the insights dashboard data.

- [ ] One table for the token info, containing (at each height where the values change):
  - [ ] total supply
  - [ ] the total amount of tokens staked
  - [ ] the price
  - [ ] the market cap.

- [ ] Another table for the shielded pool, for each asset, at each time:

  - [ ] total value shielded
  - [ ] current value shielded
  - [ ] number of transparent addresses who have shielded

The intention is that the changes can then be calculated on demand by querying the current value, the value 24h ago, etc.

cronokirby:
Closes #4896.

This adds the requisite tables, with a component parametrized by the numeraire for price information. Price information uses candlestick data for generating the price, so the recently added event will need to be used there.

This also backfills in rich domain types into old crates as necessary, for convenience.

Also closes #4883, since at this point we've refactored out all requisite schemas.

  > indexing only

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5824,6 +5824,7 @@ dependencies = [
  "penumbra-auction",
  "penumbra-dex",
  "penumbra-fee",
+ "penumbra-funding",
  "penumbra-governance",
  "penumbra-keys",
  "penumbra-num",
--- a/crates/bin/pindexer/Cargo.toml
+++ b/crates/bin/pindexer/Cargo.toml
@@ -22,6 +22,7 @@ penumbra-app = {workspace = true, default-features = false}
 penumbra-auction = {workspace = true, default-features = false}
 penumbra-dex = {workspace = true, default-features = false}
 penumbra-fee = {workspace = true, default-features = false}
+penumbra-funding = {workspace = true, default-features = false}
 penumbra-keys = {workspace = true, default-features = false}
 penumbra-governance = {workspace = true, default-features = false}
 penumbra-num = {workspace = true, default-features = false}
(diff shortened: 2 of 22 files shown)


conorsch (approved):
I haven't functionally verified these changes by running a full reindex, but confirmed with @cronokirby that he has. After merge of this change, we're unblocked to start reindexing, and we'll plan to ship this pindexer update as part of the next point release.

PR #4905, opened 2024-10-25
ibc: handle height in gRPC request metadata if specified

noot:
## Describe your changes

Check if height is specified in the metadata for relevant IBC gRPC queries and get the respective snapshot if it is provided. If the height is not specified in the metadata, the latest snapshot is used, which is the existing behaviour.

Reasoning: since astria uses only gRPC queries for all the IBC queries in the hermes impl, but the gRPC proto messages don't contain the query height, we implemented a fix by putting the height in the metadata. Without this fix, our hermes impl would submit unexpected proofs on occasion, as the proofs would be off-by-one height than that was expected.

I opened a PR to update the protos, but the response by the maintainer of ibc-go suggests to me that they intend for having the height in the header be the actual fix (https://github.com/cosmos/ibc-go/pull/7303).

n/a

  > only affects the IBC gRPC server, and does not affect existing behaviour


crates/core/component/ibc/src/component/rpc/utils.rs
```diff
@@ -0,0 +1,160 @@
...
+#[instrument(skip_all, level = "debug")]
+fn determine_height_from_metadata(
+    metadata: &tonic::metadata::MetadataMap,
+) -> anyhow::Result<Height> {
+    match metadata.get("height") {
+        None => {
+            debug!("height header was missing; assuming a height of 0");
+            Ok(TheHeight::zero().into_inner())
+        }
+        Some(entry) => entry
```
erwanor, line 41: We can use the `ibc-types-core-client::Height` type and avoid the newtyping. With that said, I think we should just merge this, and then favor using the new messages once those become available: https://github.com/cosmos/ibc-go/pull/7303

erwanor:
Thanks a lot, learned a bunch from going down this rabbit hole. I am doing some testing and will merge EOD or early tomorrow.

erwanor (approved):
🙏

PR #4907, opened 2024-10-28
bug: pindexer: allow dex and auction supply to be temporarily negative

### LINKED ISSUE #4906: pindexer: crash on "out of range integral type conversion"
**Describe the bug**
A running instance of `pindexer` version `v0.80.7` recently crashed with error message:

```
Error: out of range integral type conversion attempted
```

After brief investigation, @cronokirby pointed to a possible underflow in how pindexer reconciles VCB events. Possible solutions are to enforce ordering in the emission of ABCI events, or just make pindexer more forgiving. We've already confirmed that total counts of the ins and outs of the dex in a pindexer database sum to 0, which is as expected, so any variation would happen with a single block and should be forgiven.

cronokirby:
Closes #4906.

## Describe your changes

This allows the dex and auction components of the old supply app view to become temporarily negative until the end of a block, so that the order of debits and credits within a block doesn't matter.

This should unblock the currently crashing pindexer on mainnet, and one can verify that the dex supply never goes below 0 empirically.

  > pindexer only.

### DIFF
--- a/crates/bin/pindexer/src/supply.rs
+++ b/crates/bin/pindexer/src/supply.rs
@@ -48,9 +48,9 @@ mod unstaked_supply {
         /// The supply that's not locked in any component.
         pub um: u64,
         /// The supply locked in the auction component.
-        pub auction: u64,
+        pub auction: i64,
         /// The supply locked in the dex component.
-        pub dex: u64,
+        pub dex: i64,
         /// The supply which has been (forever) locked away after arb.
         pub arb: u64,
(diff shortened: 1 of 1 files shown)


conorsch (approved):
Resolved the crash I observed earlier.

conorsch:
To test, I built a container from this PR, deployed it to the instance where the crash in #4906 was observed, and confirmed that `pindexer` was able to progress past the problematic block and continue to index events.

PR #4909, opened 2024-10-30
pindexer: insights: correctly handle Arb and Fees

cronokirby:
Previously, these were being added to the supply, instead of subtracted.

This causes the supply in the insights dashboard to be slightly inflated.

## Testing

To test, run pindexer and compare the following queries:
```
SELECT (
  SELECT SUM(um) as staked_um
    FROM (
      SELECT * FROM supply_validators
    ) validators
    LEFT JOIN LATERAL (
      SELECT um
      FROM supply_total_staked
      WHERE validator_id = id
      ORDER BY height DESC
      LIMIT 1
    ) staked_lateral ON TRUE)
+ um + auction + dex FROM supply_total_unstaked ORDER BY height DESC LIMIT 1;
```
```
SELECT total FROM insights_supply ORDER BY height DESC LIMIT 1;
```

  > Pindexer only

### DIFF
--- a/crates/bin/pindexer/src/insights/mod.rs
+++ b/crates/bin/pindexer/src/insights/mod.rs
@@ -422,7 +422,7 @@ impl AppView for Component {
                     self.price_numeraire,
                     Box::new(move |supply| {
                         Ok(Supply {
-                            total: supply.total + amount,
+                            total: supply.total - amount,
                             ..supply
                         })
                     }),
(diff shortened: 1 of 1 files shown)


cronokirby:
Not sufficient to fix the discrepancy ; further investigation required.

conorsch:
> Not sufficient to fix the discrepancy ; further investigation required.

Even though this doesn't resolve the problem fully, the code changes are solid and we want them, so I'm going to merge and include with the intent of shipping in the next point release (#4928). Will re-index against testnet and mainnet chains to examine behavior prior to tagging.

conorsch (approved):

PR #4910, opened 2024-10-30
Pindexer remove unused appviews

cronokirby:
## Describe your changes

This removes:
- The Dex AppView
- The ShieldedPool AppView

The Dex AppView shouldn't be needed, given that we have an app view specific to the dex explorer now.

The shielded pool app view was just an example, which can be removed given the many other examples now.

Other AppViews which should be removed in a later PR after making sure they're not used:
- the supply AppView
  - superceded by the insights app view, but is known to be used
- the ibc AppView
  - I think there's ongoing development using this, potentially?
  
## Testing

The other app views should continue to work.

  > indexing only


hdevalence:
SGTM

conorsch:
> the supply AppView, superceded by the insights app view, but is known to be used

Where? If we know for sure, some proactive outreach might be warranted. Also happy to just leave it in place if you are.

conorsch (approved):

cronokirby:
https://github.com/radiantcommons/penumbra-token-info uses it ; should be a very simple change

PR #4917, opened 2024-11-09
pindexer: flesh out dex-explorer indexer

### LINKED ISSUE #4914: Dex Public Trading Insights
For the purposes of the dex explorer, we want to flesh out our indexer a bit more.

We need, over the time frames:

- 1 minute
- 15 minutes
- 1 hour
- 4 hours
- 1 day
- 1 week
- 1 month

**Per Directed Pair**

- swap / direct volume
- liquidity (in terms of the target asset)
- price (in terms of the target asset)

and then the change over the window

**In Aggregate**

- total liquidity (currently)
- total swap / direct volume (over the window)
- number of active pairs (currently)
- top price mover (over the window)
- largest trading pair (over the window)
- number of "trades" (over the window)

cronokirby:
## Describe your changes

Closes #4914.

The internal architecture of the app view tries to make use of batch processing to the extent possible, which simplifies a lot of the logic.

The price charts remain unchanged, but I collapsed the two tables into one for performance and simplicity.

I also **did not** implement insertion of empty candles ; if there are gaps in the events, there will be gaps in the resulting database as well.

The main addition and where I spent most of my time on this is the addition of summaries of information over arbitrary windows. The idea behind the architecture here is that any time a change to liquidity, trade count, or a candle for a directed pair happens in a block, that block then gets a "snapshot" inserted, with the current price, liquidity, volume in that block, etc. At the end of this batch, the current summary is then updated, for each window, using those timed snapshots.  And then an aggregate summary, across all pairs, is created from these summaries, for each window.

In order to price values under a common denom, assets are filtered based on having a current USDC price, backed by enough liquidity (the denom and liquidity amount are parameters to the component).

For testing, I'd recommend trying to run the app view against mainnet and testnet, and checking some sanity items like the price not seeming crazy, and matching in the summary across all windows, etc.

I think for testing we'll notice potential issues relatively quickly when dogfooding the explorer.

  > indexing only

### DIFF
--- a/crates/bin/pindexer/src/dex_ex/schema.sql
+++ b/crates/bin/pindexer/src/dex_ex/schema.sql
@@ -1,19 +1,3 @@
-CREATE TABLE IF NOT EXISTS dex_ex_candlesticks (
-  id SERIAL PRIMARY KEY,
-  -- The price at the start of a window.
-  open FLOAT8 NOT NULL,
-  -- The price at the close of a window.
-  close FLOAT8 NOT NULL,
-  -- The highest price reached during a window.
-  high FLOAT8 NOT NULL,
-  -- The lowest price reached during a window.
-  low FLOAT8 NOT NULL,
-  -- The volume traded directly through position executions.
-  direct_volume FLOAT8 NOT NULL,
-  -- The volume that traded indirectly, possibly through several positions.
-  swap_volume FLOAT8 NOT NULL
-);
-
 -- Contains, for each directed asset pair and window type, candle sticks for each window.
 CREATE TABLE IF NOT EXISTS dex_ex_price_charts (
   -- We just want a simple primary key to have here.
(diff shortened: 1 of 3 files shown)


cronokirby:
image
image

crates/bin/pindexer/src/indexer_ext.rs
```diff
@@ -12,7 +12,14 @@ impl IndexerExt for cometindex::Indexer {
             .with_index(Box::new(crate::stake::DelegationTxs {}))
             .with_index(Box::new(crate::stake::UndelegationTxs {}))
             .with_index(Box::new(crate::governance::GovernanceProposals {}))
-            .with_index(Box::new(crate::dex_ex::Component::new()))
+            .with_index(Box::new(crate::dex_ex::Component::new(
+                penumbra_asset::asset::Id::from_str(
+                    // USDC
+                    "passet1w6e7fvgxsy6ccy3m8q0eqcuyw6mh3yzqu3uq9h58nu8m8mku359spvulf6",
```
conorsch, line 18: This asset id is only relevant to the chain `penumbra-1`, no? How will the indexing work on e.g. `penumbra-testnet-phobos-2`? Hard to evaluate the behavior without having already updated the `dex-explorer` code for the testnet deployment, which is still unplanned.

crates/bin/pindexer/src/dex_ex/schema.sql
```diff
@@ -1,19 +1,3 @@
-CREATE TABLE IF NOT EXISTS dex_ex_candlesticks (
```
conorsch, line 1: Breaking schema change for the [dex-explorer](https://github.com/penumbra-zone/dex-explorer), but worth it, because these changes are intended to ease dex explorer development. I'll file an issue over in that repo to track the required change.

conorsch (approved):
Overall looks good. The performance improvements are significant! I've got a small concern about chain-specific logic in the indexing code, in the form of hardcoded asset ids, but willing to follow up on that when we've updated the interfaces like dex-explorer to work with the new schema.

crates/bin/pindexer/src/dex_ex/schema.sql
conorsch: > I'll file an issue over in that repo to track the required change.

https://github.com/penumbra-zone/dex-explorer/issues/132

crates/bin/pindexer/src/indexer_ext.rs
cronokirby: The same issue is also present for the insights dashboard ; maybe we should add a command line argument to override USDC? (in both cases)

crates/bin/pindexer/src/dex_ex/schema.sql
cronokirby: As a further bit of context, the reason for the breaking change there is that it requires several more writes to move the candlesticks into a separate table, versus just writing them inline ; given the pre-alpha nature of the explorer a breaking change for simpler and more performant code forever is a good tradeoff

crates/bin/pindexer/src/indexer_ext.rs
conorsch: Not a bad idea, filed an issue in this repo https://github.com/penumbra-zone/penumbra/issues/4920

VanishMax (approved):
Can't wait to apply this to https://github.com/penumbra-zone/dex-explorer/pull/133!

@cronokirby I see that the header from the explore page can be simply taken from `dex_ex_aggregate_summary` table. What about the list of pairs and their stats?

image

hdevalence:
@VanishMax That should (initially, possibly permanently) be configuration data (JSON) loaded into the explorer. There can be one for testnet and one for mainnet. The trading interface needs to express opinions about the list, at minimum because it needs to choose which asset is treated as the numeraire (e.g., using USDC as asset 2), so it seems simplest to start with a list specified in config data (so that there can be one for a mainnet and one for a testnet deployment).

cronokirby:
From there, you can query the `dex_ex_pair_summary` table with a particular start asset, end asset, and window, which should give you all the stats you need.

PR #4923, opened 2024-11-14
UIP-5: outbound PFM support

avahowell:
## Describe your changes

Implementation of the candidate UIP-5: outbound PFM support (see https://forum.penumbra.zone/t/pre-uip-outbound-packet-forwarding-middleware-support/121)

Should be reviewed for correctness and adherence to the spec.

Note that this only includes the required protocol changes to allow clients to use the new memo field (a prerequisite for PFM support), and does not implement outbound packet forwarding directly (that is to be done in the client).


cronokirby:
This breaks consensus, right? I think we need to retarget this PR towards release/v0.81.0 instead

erwanor:
Yup, consensus breaking but not proto breaking

crates/core/component/shielded-pool/src/ics20_withdrawal.rs
```diff
@@ -118,6 +123,7 @@ impl From<Ics20Withdrawal> for pb::Ics20Withdrawal {
             timeout_time: w.timeout_time,
             source_channel: w.source_channel.to_string(),
             use_compat_address: w.use_compat_address,
+            ics20_memo: w.ics20_memo.to_string(),
```
erwanor, line 126: Do we need validation rules on memo size? I guess a user would pay gas for it, but this feels a bit scary, we do have them for governance proposals.

cronokirby: There was discussion in the UIP about this, but I think that because this goes out to other chains, we really can't anticipate potential use cases, and because gas is paid, this is fine, and I think any limit that would actually be meaningfully small might be too small for some wacky use case someone comes up with.

erwanor: Sounds reasonable 👍

erwanor (approved):
LGTM on code review

cronokirby (approved):

conorsch (changes requested):
PR should target branch `release/v0.81.x`; requesting changes to block merge into main.

erwanor:
Ugh, changing the base has picked up other commits making the diff monstruous. I will re-open this and target the 81 branch directly.

Issue #4921, opened 2024-11-13
update f4jumble to published crate


conorsch:
The `penumbra-keys` crate depends on `f4jumble` via the zcash repo as a git dependency: https://github.com/penumbra-zone/penumbra/blob/v0.80.8/crates/core/keys/Cargo.toml#L30

In order to publish the workspace crates to crates.io (e.g. #3010), we can [only depend on other crates on crates.io](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-other-registries):

> crates.io does not allow packages to be published with dependencies on code published outside of crates.io.

Therefore we should switch to the most recent version of `f4jumble` published to crates.io, which is `v0.1.0`: https://crates.io/crates/f4jumble/0.1.0

conorsch:
We want to upgrade `f4jumble` to `0.1.0`. To compare changes in the f4jumble source, we must compare the git rev version `2425a0869098e3b0588ccd73c42716bcf418612c` with that of the version published to crates.io as `0.1.0`. Here's a script I used to perform that comparison.

<details>
<summary>
bash script for comparing crate versions
</summary>

```
#!/bin/bash
# Quick script to compare two versions of the `f4jumble` dep,
# one from a specific git revision, the other from crates.io.
# Context for comparison is in
# https://github.com/penumbra-zone/penumbra/issues/4921
# Requires shell programs:
#
#   * curl
#   * diffoscope
#   * rsync
#   * tar
#
set -euo pipefail


# Create a local directory for unpacking the two archives,
# for comparison.
d="./workbench"
rm -rf "$d"
mkdir -p "$d"
cd "$d" || exit 1

# obtain the git rev version of the dep
git clone https://github.com/zcash/librustzcash
git -C librustzcash checkout 2425a0869098e3b0588ccd73c42716bcf418612c

parent_dir_for_git_rev_crate="f4jumble-0.1.0-git"
# structure the tar archive to have the same par dir name
# as the cargo version, to ease comparison.
parent_dir_for_git_rev_crate="f4jumble-0.1.0"
mkdir -p "$parent_dir_for_git_rev_crate"
rsync -a librustzcash/components/f4jumble/ "${parent_dir_for_git_rev_crate}/"
tar -cf f4jumble-git.tar "$parent_dir_for_git_rev_crate"
rm -r "$parent_dir_for_git_rev_crate"

# obtain the crates version:
# see here for discussion of this api call
[...]

PR #4926, opened 2024-11-15
chore: update f4jumble to 0.1.0

conorsch:
## Describe your changes

Updates the old git dependency to a suitable published version from crates.io. To review, see the diff review posted in #4921. I'd prefer that reviewers replicate that work, and confirm that 1) the procedure used for obtaining relevant sources was sound; and 2) the code changes in the dep introduced by this diff are acceptable.

Refs #4921.

  > The dep appears to function the same in pre and post version bump, as does our modified error handling of its return values. Reviewers should confirm those statements are true, and if so, then there's no affect on consensus.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2066,8 +2066,9 @@ dependencies = [
 
 [[package]]
 name = "f4jumble"
-version = "0.0.0"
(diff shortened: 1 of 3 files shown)


redshiftzero (approved):
I reviewed the diff between revision `2425a0869098e3b0588ccd73c42716bcf418612c` and 0.1.0 published on crates.io and agree there aren't changes to the `F4Jumble` transformation. LGTM

PR #4931, opened 2024-11-18
pindexer: dex-explorer: implement low high in summary

### LINKED ISSUE #4929: pindexer: dex-explorer: missing low and high in summary table
The summary table also needs to have the lowest and highest price over a given window, which it is currently missing.

cronokirby:
## Describe your changes

Closes #4929.

This was surprisingly simple, we were already doing an aggregate calculation over the window (summing values) so this was just a matter of adding a low and a high.

To test, index against main and check that nothing else has changed besides the addition of two new fields.

  > indexing


conorsch:
> To test, index against main and check that nothing else has changed besides the addition of two new fields.

Locally I've run `pindexer` against a cometbft db based on most recent HEAD of this repo, then again building pindexer from this PR branch, and both completed without errors. Ensuring the diff between the two dbs is as minimal as described is difficult, so I'm going to approve and keep moving: shortly, after landing a few more pindexer fixes, I'll update the db backing the testnet dex-explorer, which gives us more context to analyze the change here.

conorsch (approved):
Approving and merging, along with a batch of other pindexer improvements like #4909 & #4932.

PR #4932, opened 2024-11-18
pindexer: insights: correct calculation of total value shielded

### LINKED ISSUE #4925: Don't take into account refunds for "total value shielded"
It seems like on insights.penumbra.zone the "total value shielded" doesn't correspond exactly to deposit activity (e.g. for USDC). I think this might be because we take into account refunds here when we shouldn't.

cronokirby:
## Describe your changes

This corrects the calculation of the total value shielded. Previously, were were adding the absolute value of outbound transfers, as well as refunds to the total. Instead, we just need to consider deposits.

To test, we should look at current mainnet data, and make sure that the total for USDC is changing appropriately.

Closes #4925

### DIFF
--- a/crates/bin/pindexer/src/insights/mod.rs
+++ b/crates/bin/pindexer/src/insights/mod.rs
@@ -151,6 +151,7 @@ async fn asset_flow(
     asset_id: asset::Id,
     height: u64,
     flow: i128,
+    refund: bool,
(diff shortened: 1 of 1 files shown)


conorsch (approved):
Approving optimistically; along with another batch of pindexer improvements landing today, I'll reindex the insights db and we can observe the effect live.

Issue #4933, opened 2024-11-18
pindexer: swap_volume_over_window contains null values


conorsch:
**Describe the bug**
While testing recent `pindexer` bug fixes by rerunning `pindexer` from block zero, I encountered a failure on the `penumbra-1` chain:

```
2024-11-18T23:27:34.974561Z DEBUG sqlx::query: summary="INSERT INTO insights_supply(height, total, …" db.statement="\n\nINSERT INTO\n  insights_supply(\n    height,\n    total,\n    staked,\n    price,\n    price_numeraire_asset_id\n  )\nVALUES\n  ($1, $2, $3, $5, $4) ON CONFLICT (height) DO\nUPDATE\nSET\n  total = excluded.total,\n  staked = excluded.staked,\n  price = excluded.price,\n  price_numeraire_asset_id = excluded.price_numeraire_asset_id\n" rows_affected=0 rows_returned=0 elapsed=3.095199ms
2024-11-18T23:27:34.974774Z DEBUG sqlx::query: summary="SELECT events.rowid, events.type, events.height, …" db.statement="\n\nSELECT\n  events.rowid,\n  events.type,\n  events.height,\n  tx_results.tx_hash,\n  events.attrs\nFROM\n  (\n    SELECT\n      (\n        SELECT\n          height\n        FROM\n          blocks\n        WHERE\n   blocks.rowid = block_id\n      ) as height,\n      rowid,\n      type,\n      block_id,\n      tx_id,\n      jsonb_object_agg(attributes.key, attributes.value) AS attrs\n    FROM\n      events\n      LEFT JOIN attributes ON rowid = attributes.event_id\n    WHERE\n      block_id >= (\n        SELECT\n          rowid\n        FROM\n blocks\n        where\n          height = $1\n      )\n      AND block_id <= (\n        SELECT\n          rowid\n        FROM\n          blocks\n        where\n          height = $2\n      )\n    GROUP BY\n      rowid,\n      type,\n      block_id,\n      tx_id\n    ORDER BY\n      rowid ASC\n  ) events\n  LEFT JOIN LATERAL (\n    SELECT\n    *\n    FROM\n      tx_results\n    WHERE\n      tx_results.rowid = events.tx_id\n    LIMIT\n      1\n  ) tx_results ON TRUE\nORDER BY\n  events.rowid ASC\n" rows_affected=0 rows_returned=0 elapsed=830.672533ms
[...]

PR #4934, opened 2024-11-19
ci: fix container build for parameter-setup

conorsch:
## Describe your changes
Follow-up to #4930, in which we inadvertently broken container builds, due to a workspace dependency being introduced. We don't build container images on every PR push, so CI passed fine, only failing when trying to build/deploy post-merge. Tacked on a small change to address.

For testing, leave it to me: I'll re-deploy after this lands to get it going. 

Refs #4933.

  > build logic only, no changes to consensus

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -16,3 +16,5 @@
 !deployments/relayer/
 # summonerd tool, for orchestrating contributions
 !tools/summonerd/
+# parameter setup tool, for wrangling circuit configuration
+!tools/parameter-setup/


conorsch:
> For testing, leave it to me: I'll re-deploy after this lands to get it going.

Built a container from this branch and [confirmed build passing](https://github.com/penumbra-zone/penumbra/actions/runs/11903973386). Then I deployed a pindexer instance from that image tag and confirmed the problem described in #4933 was resolved. There's another error emitted now, but at least we've unblocked the image builds, so we're running the latest code again.

cronokirby (approved):

Issue #4899, opened 2024-10-21
breakage against Noble testnet v8.0.0-rc.2


conorsch:
## Describe the bug
An upcoming Noble chain upgrade to v8 is being prepared on the Noble testnet. For the Penumbra Labs testnet (https://testnet.plinfra.net), we've been running a version of Hermes that relays between `penumbra-testnet-phobos-2` and `grand-1`. On or around 2024-10-17, we started observing breakage when communicating with the Noble testnet node endpoint run by Polkachu: 
* grpc: `noble-testnet-grpc.polkachu.com:21590`
* rpc: https://noble-testnet-rpc.polkachu.com 

I confirmed out of band with Polkachu that this breakage corresponded to deployment of the [https://github.com/noble-assets/noble/releases/tag/v8.0.0-rc.2](v8.0.0-rc.2) tag to the testnet endpoint.

We first discovered this breakage when testing the behavior of the diff in https://github.com/penumbra-zone/penumbra/pull/4878. Similar breakage is also evident in the hermes relayer that PL is running.

## Example error messages

When running on the feature branch for #4878:

```
❯ cargo run -q --release --bin pcli --  --home ~/.local/share/pcli view noble-address --channel channel-221 --noble-node http://noble-testnet-grpc.polkachu.com:21590/
Error: status: Internal, message: "failed to decode Protobuf message: TxResponse.raw_log: BroadcastTxResponse.tx_response: invalid string value: data is not UTF-8 encoded", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "x-cosmos-block-height": "15378766"} }
 ```

When viewing the logs for the hermes relayer instance between testnets:

```
Oct 17 23:17:22 hermes hermes[1971991]: 2024-10-17T23:17:22.462927Z ERROR ThreadId(27) event_source.rpc{chain.id=grand-1}: failed to collect events: RPC error: serde parse error: invalid utf-8 sequence of 1 bytes from index 0 at line 1 column 312, retrying in 1.5s... height=15322789
Oct 17 23:17:24 hermes hermes[1971991]: 2024-10-17T23:17:24.305214Z ERROR ThreadId(27) event_source.rpc{chain.id=grand-1}: failed to collect events: RPC error: serde parse error: invalid utf-8 sequence of 1 bytes from index 0 at line 1 column 312, retrying in 2s... height=15322789
[...]

PR #4937, opened 2024-11-20
test: add Noble/Penumbra IBC smoke test script

zbuc:
## Describe your changes

This adds a test script that performs bidirectional IBC transfers between Penumbra and Noble.

Prerequisites:

* A Noble account with uusdc funds
* A Penumbra account with UM funds to pay fees
* A checkout of [the Noble SDK](https://github.com/noble-assets/noble/)
* A working `pcli` installation in the `$PATH`
* `jq`

This should be integrated into a GitHub workflow. It's not clear if it belongs in the existing smoketest workflow or a new workflow.

  > test only

### DIFF
--- a/deployments/scripts/noble_ibc_test.sh
+++ b/deployments/scripts/noble_ibc_test.sh
@@ -1,0 +1,177 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# This script automates the process of testing Noble IBC transfers
+
+# Constants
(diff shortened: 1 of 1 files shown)


conorsch:
Rebased this and ran `shellcheck` on it. While we're not using the script in CI, it will likely prove useful when evaluating fixes related to #4899, so I'm pulling it into tree.

conorsch (approved):

PR #4940, opened 2024-11-22
UIP-5: Outbound PFM support

erwanor:
## Describe your changes

Implementation of the candidate UIP-5: outbound PFM support (see https://forum.penumbra.zone/t/pre-uip-outbound-packet-forwarding-middleware-support/121)

Should be reviewed for correctness and adherence to the spec.

Note that this only includes the required protocol changes to allow clients to use the new memo field (a prerequisite for PFM support), and does not implement outbound packet forwarding directly (that is to be done in the client).

By @avahowell from https://github.com/penumbra-zone/penumbra/pull/4923


cronokirby (approved):

erwanor:
I will rebase and merge.

PR #4941, opened 2024-11-22
pindexer: implement symmetrization of charts

cronokirby:
Now charts implement price information from both directions of a pair.

This is needed for correct rendering on the dex explorer.

For testing, run a local copy of the dex explorer and notice that now both charts are better inverses of each other.

  > indexing

### DIFF
--- a/crates/bin/pindexer/src/dex_ex/mod.rs
+++ b/crates/bin/pindexer/src/dex_ex/mod.rs
@@ -645,7 +645,7 @@ impl Events {
     }
 
     fn with_candle(&mut self, pair: DirectedTradingPair, candle: Candle) {
-        // Popular both this pair and the flipped pair, and if the flipped pair
+        // Populate both this pair and the flipped pair, and if the flipped pair
         // is already populated, we need to mix the two candles together.
         let flip = pair.flip();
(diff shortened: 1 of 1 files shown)


crates/bin/pindexer/src/dex_ex/mod.rs
```diff
@@ -616,14 +645,20 @@ impl Events {
     }
 
     fn with_candle(&mut self, pair: DirectedTradingPair, candle: Candle) {
-        match self.candles.get_mut(&pair) {
-            None => {
-                self.candles.insert(pair, candle);
-            }
-            Some(current) => {
-                current.merge(&candle);
+        // Popular both this pair and the flipped pair, and if the flipped pair
```
vacekj, line 648:
```suggestion
        // Populate both this pair and the flipped pair, and if the flipped pair
```

vacekj (approved):
lgtm

hdevalence:
is this good to merge or does it require db changes? wondering if we could flush the pipeline and get it into the preview deployment

cronokirby:
This has no breaking change to the DB schema, it just changes the data inside of it, so I think we can deploy as is.

PR #4942, opened 2024-11-22
pd: implement no op migration to v0.81.0

cronokirby:
This implements a migration to version 0.81.0, which doesn't require any logic.

This will break PD, but targets the new release branch.


conorsch:
Tested locally by creating a devnet, then:

1. staking, bonding
2. proposing upgrade plan
3. voting
4. confirming halt
5. running `pd migrate` to apply the change: :heavy_check_mark: 
6. confirming chain restarts on new version of pd: :x: did not work

Step 6 failed because of the version mismatch between pd and local state: we must also increment `APP_VERSION` in the code so that the startup version check introduced in #4919 is satisfied. I pushed a commit to do just that.

Issue #4939, opened 2024-11-21
Add `TransactionView::summary()` method


hdevalence:
**Is your feature request related to a problem? Please describe.**

There are many applications where it would be useful to display a summary of a transaction's effects on the balances of various accounts, both pre- and post-confirmation:

- **Pre-confirmation**: in an approval dialog, for showing an exact preview of the effects of a transaction
- **Post-confirmation**: in a frontend, for showing a summary of what a transaction did

This logic should be implemented once, in the Rust code, so that it can be standardized across the Penumbra ecosystem.

**Describe the solution you'd like**

- [x] The `Balance` struct improperly has a derived Serde representation. This is a bug and a violation of our serialization policy. Verify that this was never used. Define a Protobuf message for the `Balance` struct and use that as the serialization format.

- [x] Define a new Protobuf message and corresponding domain type, `penumbra.core.transaction.v1.TransactionSummary`, something like (sketch):
```protobuf
message TransactionSummary {
    message Effects {
        core.keys.v1.AddressView address = 1;
        core.asset.v1.Balance balance = 2;
    }
    repeated Effects effects = 1;
}
```

- [x] Define a new method `TransactionView::summary(&self) -> TransactionSummary` that produces a `TransactionSummary`, iterating through each visible action and collecting the effects of the transaction. This should preserve the info available in the `AddressView`s about the meaning of the addresses.

erwanor:
This was done for `Output/Spend/Swap/SwapClaim` views in #4943. Adding the LP ones might enable some cool use cases, but afaict this is complete.

PR #4943, opened 2024-11-25
TransactionView: summary of transaction effects 

TalDerei:
## Describe your changes

References https://github.com/penumbra-zone/penumbra/issues/4939


crates/core/transaction/src/view.rs
```diff
@@ -94,6 +97,48 @@ impl TransactionView {
     pub fn action_views(&self) -> impl Iterator<Item = &ActionView> {
         self.body_view.action_views.iter()
     }
+
+    pub fn summary(&self) -> TransactionSummary {
+        let mut effects = Vec::new();
+
+        for action_view in &self.body_view.action_views {
```
TalDerei, line 104: comment: Iterates through the visible action views and adds the contributing balances to the transaction's effects. Sanity check the required and provided balances.

crates/core/transaction/src/view.rs
```diff
@@ -94,6 +97,48 @@ impl TransactionView {
...
+                        let address = AddressView::Opaque {
+                            address: note.address(),
+                        };
+
+                        effects.push(Effects { address, balance });
+                    }
+                    OutputView::Opaque { output: _ } => continue,
+                },
+                ActionView::Swap(_) => todo!(),
+                ActionView::SwapClaim(_) => todo!(),
```
TalDerei, line 135: comment: need to implement other action views

crates/core/asset/src/balance.rs
```diff
@@ -29,17 +28,50 @@ use decaf377::{r1cs::ElementVar, Fq, Fr};
...
+
+    fn try_from(v: pb::Balance) -> Result<Self, Self::Error> {
+        let mut balance_map = BTreeMap::new();
+
+        for imbalance_value in v.balance.into_iter().map(TryInto::try_into) {
+            let value: Value = imbalance_value?;
+            let amount = NonZeroU128::new(value.amount.into())
+                .ok_or_else(|| anyhow::anyhow!("amount must be non-zero"))?;
+
+            let imbalance = Imbalance::Provided(amount); // todo: fix this placeholder
```
TalDerei, line 58: comment: `Values` cannot be negative, a negative `Balance` is formed by negating a balance derived from a value. How do we calculate the Imbalance and differentiate between provided and required balances?

crates/core/asset/src/balance.rs
```diff
@@ -29,17 +28,50 @@ use decaf377::{r1cs::ElementVar, Fq, Fr};
...
+use penumbra_proto::{penumbra::core::asset::v1 as pb, DomainType};
 
 /// A `Balance` is a "vector of [`Value`]s", where some values may be required, while others may be
 /// provided. For a transaction to be valid, its balance must be zero.
-#[serde_as]
 #[derive(Clone, Eq, Default, Serialize, Deserialize)]
+#[serde(try_from = "pb::Balance", into = "pb::Balance")]
 pub struct Balance {
     negated: bool,
-    #[serde_as(as = "Vec<(_, _)>")]
```
TalDerei, line 39: confirmed the previous derived serde policy wasn't used.

crates/core/transaction/src/transaction.rs
```diff
@@ -7,14 +7,15 @@ use anyhow::{Context, Error};
 use ark_ff::Zero;
 use decaf377::Fr;
 use decaf377_rdsa::{Binding, Signature, VerificationKey, VerificationKeyBytes};
+use penumbra_asset::Balance;
 use penumbra_community_pool::{CommunityPoolDeposit, CommunityPoolOutput, CommunityPoolSpend};
 use penumbra_dex::{
     lp::action::{PositionClose, PositionOpen},
     swap::Swap,
 };
```
TalDerei, line 15: https://github.com/penumbra-zone/penumbra/blob/ff86c9a2fd502796156c244c66458fcada046603/crates/core/transaction/src/transaction.rs#L735-L739

does this lead to infinite recursion?

proto/penumbra/penumbra/core/asset/v1/asset.proto
```diff
@@ -92,6 +92,13 @@ message Value {
   AssetId asset_id = 2;
 }
 
+message Balance {
+  // Indicates if the balance is negated.
+  bool negated = 1;
+  // Represents the vector of 'Values' in the balance.
+  repeated Value balance = 2;
+}
```
hdevalence, line 100: xref https://github.com/penumbra-zone/penumbra/pull/4943/files#r1855748563

Since `Value`s can't be negative, I think we'll want to have a list of pairs, each with a negation flag, something like
```
message Balance {
  message SignedValue {
    Value value = 1;
    bool negated = 2;
  }
  repeated SignedValue values = 1;
}
```
The top-level bool in the Rust `Balance` struct is an optimization that makes sign changes less expensive (unclear this was really necessary perf-wise, but, we have it now), IMO it should be excluded from serialization by having the serialized form always be "normalized" with individual sign bools.

proto/penumbra/penumbra/core/transaction/v1/transaction.proto
```diff
@@ -54,6 +54,17 @@ message TransactionParameters {
   component.fee.v1.Fee fee = 3;
 }
 
+// Represents an individual effect of a transaction.
+message Effects {
+  keys.v1.AddressView address = 1;
+  asset.v1.Balance balance = 2;
+}
```
hdevalence, line 61: Let's move this into a sub-message so it's clear that it's logically scoped to be part of the TransactionSummary.

crates/core/transaction/src/transaction.rs
erwanor: Yup. good catch

crates/core/asset/src/balance.rs
erwanor: I think the modeling suggested by henry here: https://github.com/penumbra-zone/penumbra/pull/4943/files#r1855950445 takes care of that

proto/penumbra/penumbra/core/transaction/v1/transaction.proto
TalDerei: moved into a sub-message

```ts
// Represents a transaction summary containing multiple effects.
message TransactionSummary {
  // Represents an individual effect of a transaction.
  message Effects {
    keys.v1.AddressView address = 1;
    asset.v1.Balance balance = 2;
  }
  repeated Effects effects = 1;
}
```

proto/penumbra/penumbra/core/asset/v1/asset.proto
TalDerei: makes sense, refactored to normalize the top-level `Balance` and capture the sign information in the `SignedValue` pairs.

crates/core/asset/src/balance.rs
TalDerei: imbalance is now determined based on the `negated` flag for the individual `SignedValue` pairs.

crates/core/transaction/src/transaction.rs
TalDerei: fixed the domain type 

```ts
impl From<&Transaction> for pbt::Transaction {
    fn from(msg: &Transaction) -> Self {
        Transaction {
            transaction_body: msg.transaction_body.clone(),
            anchor: msg.anchor.clone(),
            binding_sig: msg.binding_sig.clone(),
        }
        .into()
    }
}
```

crates/core/asset/src/balance.rs
```diff
@@ -29,17 +28,50 @@ use decaf377::{r1cs::ElementVar, Fq, Fr};
...
+        Ok(Self {
+            negated: v.negated,
+            balance: balance_map,
+        })
+    }
+}
+
+impl From<Balance> for pb::Balance {
+    fn from(_v: Balance) -> Self {
+        todo!() // todo: implement fallible conversion
```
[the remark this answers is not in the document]
TalDerei: Implemented the domain type. Sanity check the `Amount` was computed properly, referencing https://github.com/penumbra-zone/penumbra/blob/main/crates/core/asset/src/balance.rs#L296-L301

crates/core/transaction/src/view.rs
TalDerei: emulated logic from https://github.com/penumbra-zone/penumbra/blob/main/crates/core/transaction/src/plan/action.rs#L201 and https://protocol.penumbra.zone/main/transactions/actions.html

proto/penumbra/penumbra/core/asset/v1/asset.proto
TalDerei: how should we handle cases where multiple entries share the same asset ID with [different](https://github.com/penumbra-zone/penumbra/pull/4943/files#diff-9064ed98f9c25749e6785cab41b153fbd2e3cf99ab342f16abdc83d16d1f0f5aR63-R68) negation flags, for instance:

```ts
        let proto_balance = pb::Balance {
            values: vec![
                pb::balance::SignedValue { 
                    value: Some(pb::Value { asset_id: Some((*STAKING_TOKEN_ASSET_ID).into()), amount: Some(Amount::from(100u128).into()) }), 
                    negated: false
                },
                pb::balance::SignedValue {
                    value: Some(pb::Value {
                        asset_id: Some((*STAKING_TOKEN_ASSET_ID).into()),
                        amount: Some(Amount::from(200u128).into()),
                    }),
                    negated: true,
                },
            ],
        };

        let balance = Balance::try_from(proto_balance).expect("fallible conversion");
```

The `BTreeMap` can't hold multiple imbalances for the same asset ID. Instead, should it be doing something more sophisticated and explicitly accumulating values during conversion?

TalDerei:
@hdevalence Is this sufficient for _pre-confirmation_ simulation, or would we also require a `TransactionPlan::summary()`? The current implementation of [TransactionView::summary(&self)](https://github.com/penumbra-zone/penumbra/pull/4943/files#diff-0dd701ecbaf794ba71840c63af273a0fbb9f18fcc6c4f4501c45adce537e688fR160) operates on a `TransactionView`, which relies on a fully-formed transaction (Tx + TxP = TxV).

crates/core/transaction/src/view.rs
```diff
@@ -94,6 +101,174 @@ impl TransactionView {
     pub fn action_views(&self) -> impl Iterator<Item = &ActionView> {
         self.body_view.action_views.iter()
     }
+
+    /// Acts as a higher-order translator that summarizes a TransactionSummary by consolidating
+    /// effects for each unique address.
+    pub fn accumulate_effects(summary: &TransactionSummary) -> TransactionSummary {
```
cronokirby, line 107: I think we should just have this be private, and call it internally

TalDerei: good catch — this was intended to be private

crates/core/transaction/src/view.rs
```diff
@@ -294,3 +466,458 @@ impl TryFrom<pbt::MemoPlaintextView> for MemoPlaintextView {
...
+                    .collect(),
+                transaction_parameters: plan.transaction_parameters.clone(),
+                detection_data: None,
+                memo_view: None,
+            },
+        };
+
+        let transaction_summary = TransactionView::summary(&transaction_view);
+
+        assert_eq!(transaction_summary.effects.len(), 2);
```
TalDerei, line 921: should add assertions about expected required and provided imbalances and sanity check final result.

crates/core/asset/src/balance.rs
```diff
@@ -29,15 +28,154 @@ use decaf377::{r1cs::ElementVar, Fq, Fr};
...
+                        }
+                        (Imbalance::Provided(existing_amount), true) => {
+                            match existing_amount.get().checked_sub(amount.get()) {
+                                Some(diff) if diff > 0 => {
+                                    *existing_amount = NonZeroU128::new(diff)
+                                        .ok_or_else(|| anyhow::anyhow!("Reduce provided amount"))?;
+                                }
+                                _ => {
+                                    *entry.get_mut() = Imbalance::Required(
+                                        NonZeroU128::new(amount.get() - existing_amount.get())
```
erwanor, line 124: Circling back to this later, I want to understand why this cannot be equal to zero (and thus, why it is infallible)

crates/core/asset/src/balance.rs
```diff
@@ -29,15 +28,154 @@ use decaf377::{r1cs::ElementVar, Fq, Fr};
...
+/// negated field is exlcuded during serialization. Rather, the
+/// sign information is captured in the `SignedValue` pairs.  
+///
+/// Since the underlying BTreeMap can't hold multiple imbalances for
+/// the same asset ID, we implement an ordering-agnostic accumulation
+/// scheme that explicitly combines imbalances.
+impl TryFrom<pb::Balance> for Balance {
+    type Error = anyhow::Error;
+
+    fn try_from(v: pb::Balance) -> Result<Self, Self::Error> {
```
erwanor, line 56: PR crux, reviewers should ACK.

crates/core/transaction/src/view.rs
```diff
@@ -94,6 +99,173 @@ impl TransactionView {
...
+                combined_proto.try_into().expect("Domain type conversion");
+
+            transaction_summary.effects.push(TransactionEffect {
+                address: (*address).clone(),
+                balance: combined_balance,
+            });
+        }
+
+        transaction_summary
+    }
```
cronokirby, line 154:
```suggestion
    fn accumulate_effects(summary: TransactionSummary) -> TransactionSummary {
        use std::collections::BTreeMap;

        let mut keyed_effects: BTreeMap<AddressView, Balance> = BTreeMap::new();
        for effect in summary.effects {
            *keyed_effects.entry(effect.address).or_default() += effect.balance;
        }
        TransactionSummary {
            effects: keyed_effects
                .into_iter()
                .map(|(address, balance)| TransactionEffect { address, balance })
                .collect(),
        }
    }
```

I think this whole method can lean on the existing sum method for Balance?

crates/core/transaction/src/view.rs
```diff
@@ -94,6 +99,173 @@ impl TransactionView {
...
+                    }
+                    SwapClaimView::Opaque { swap_claim: _ } => continue,
+                },
+                _ => {} // Fill in other action views as neccessary
+            }
+        }
+
+        let mut summary = TransactionSummary { effects };
+
+        Self::accumulate_summary(&mut summary)
```
cronokirby, line 267:
```suggestion
        let summary = TransactionSummary { effects };

        Self::accumulate_effects(summary)
```

(If you implement the suggestion above, this will need to change too)

crates/core/transaction/src/view.rs
TalDerei: this works well 👍

crates/core/keys/src/address/view.rs
```diff
@@ -11,7 +11,7 @@ use super::Address;
 ///
 /// This type allows working with addresses and address indexes without knowing
 /// the corresponding FVK.
-#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, PartialOrd, Ord)]
```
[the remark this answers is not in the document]
erwanor: Good idea, is it even necessary here? if so, then I think a comment like this one:
```suggestion
// Warning: This enum derives `PartialOrd` which means that re-arranging
// field-less variants is semver breaking. 
#derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, PartialOrd, Ord)]
```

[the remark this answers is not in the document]
TalDerei: It's necessary to satisfy other trait bounds in `accumulate_effects`. I added the proposed comment

crates/core/asset/src/balance.rs
cronokirby: I think it can be 0, you can send a proto with [-x UM, +x UM] and I think this method will panic?

crates/core/keys/src/address/view.rs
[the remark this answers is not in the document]
cronokirby: Yeah, also we want a canonical serialization for balances.

crates/core/asset/src/balance.rs
TalDerei: yup, the inverse case, [+x UM, -x UM] is also problematic. Explicitly handling the case where the difference is 0 and removing the entry from the map resolves this.

crates/core/keys/src/address/view.rs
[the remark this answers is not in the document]
TalDerei: added `PartialOrd` and `Ord` impls for `AddressView`

crates/core/transaction/src/view.rs
TalDerei: cc @erwanor @cronokirby

cronokirby (approved):

crates/core/asset/src/balance.rs
```diff
@@ -26,18 +26,86 @@ mod imbalance;
...
+        })
+    }
+}
+
+impl DomainType for Balance {
+    type Proto = pb::Balance;
+}
+
+/// Serialization should normalize the `Balance`, where the top-level
+/// negated field is exlcuded during serialization. Rather, the
```
erwanor, line 65:
```suggestion
/// negated field is excluded during serialization. Rather, the
```

crates/core/asset/src/balance.rs
```diff
@@ -26,18 +26,86 @@ mod imbalance;
...
+    fn from_signed_value(negated: bool, value: Value) -> Option<Self> {
+        let non_zero = NonZeroU128::try_from(value.amount.value()).ok()?;
+        Some(Self {
+            negated: false,
+            balance: BTreeMap::from([(
+                value.asset_id,
+                if negated {
+                    Imbalance::Required(non_zero)
+                } else {
+                    Imbalance::Provided(non_zero)
```
erwanor, line 53: ACK. LGTM

crates/core/asset/src/balance.rs
```diff
@@ -26,18 +26,86 @@ mod imbalance;
...
+/// Since the underlying BTreeMap can't hold multiple imbalances for
+/// the same asset ID, we implement an ordering-agnostic accumulation
+/// scheme that explicitly combines imbalances.
+impl TryFrom<pb::Balance> for Balance {
+    type Error = anyhow::Error;
+
+    fn try_from(balance: pb::Balance) -> Result<Self, Self::Error> {
+        let mut out = Self::default();
+        for v in balance.values {
+            let value = v.value.ok_or_else(|| anyhow!("missing value"))?;
```
erwanor, line 77: Should this default to zero? Zero values are omitted from proto messages

crates/core/asset/src/balance.rs
```diff
@@ -26,18 +26,86 @@ mod imbalance;
...
+
+impl From<Balance> for pb::Balance {
+    fn from(v: Balance) -> Self {
+        let values = v
+            .balance
+            .into_iter()
+            .map(|(id, imbalance)| {
+                // Decompose imbalance into it sign and magnitude, and convert
+                // magnitude into raw amount and determine negation based on the sign.
+                let (sign, magnitude) = if v.negated { -imbalance } else { imbalance }.into_inner();
```
erwanor, line 94: ACK, we apply the negated and collect the resulting sign. That makes sense to me

crates/core/asset/src/balance.rs
cronokirby: The current behavior is:

- Deserializing a proto with no values gives a balance of `{false, []}`
- Serializing this will return `[]`, i.e. no signed values.

This behavior seems sensible to me.

cronokirby: (and it roundtrips infinitely with no change, another sign of correctness)

erwanor: Oh NVM, NZU128

crates/core/transaction/src/view.rs
erwanor: Looks OK to me

crates/core/transaction/src/view.rs
```diff
@@ -94,6 +97,136 @@ impl TransactionView {
...
+                        let value_2 = Value {
+                            amount: swap_plaintext.delta_2_i,
+                            asset_id: swap_plaintext.trading_pair.asset_2(),
+                        };
+
+                        // Required imbalance (-)
+                        let mut balance = Balance::default();
+                        balance -= value_1;
+                        balance -= value_2;
+                        balance -= value_fee;
```
erwanor, line 185: ACK, effect is correct

crates/core/transaction/src/view.rs
```diff
@@ -94,6 +97,136 @@ impl TransactionView {
...
+
+                        // Provided imbalance (+)
+                        let mut balance = Balance::default();
+                        balance += value_fee;
+
+                        effects.push(TransactionEffect { address, balance });
+                    }
+                    SwapClaimView::Opaque { swap_claim: _ } => continue,
+                },
+                _ => {} // Fill in other action views as neccessary
```
erwanor, line 222: There might be a cool use case related with the LP actions, but we can just add them later.

erwanor (approved):

TalDerei:
change base to [release/v0.81.x](https://github.com/penumbra-zone/penumbra/tree/release/v0.81.x)?

PR #4946, opened 2024-12-01
Add pindexer support for position timelines in the DEX explorer

hdevalence:
## Describe your changes

These changes index the states of individual positions, allowing the dex explorer to render position timelines, allowing users to inspect their position history.

This also updates sqlx along the way.

Once these changes are merged, we'll need to trigger a reindex on the dev indexer.

  > indexer changes only

### DIFF
--- a/crates/bin/pindexer/src/dex_ex/schema.sql
+++ b/crates/bin/pindexer/src/dex_ex/schema.sql
@@ -82,3 +82,105 @@ CREATE TABLE IF NOT EXISTS dex_ex_metadata (
   -- The asset id to use for prices in places such as the aggregate summary.
   quote_asset_id BYTEA NOT NULL
 );
+
+CREATE TABLE IF NOT EXISTS dex_ex_position_state (
+  -- Call this rowid to distinguish it from the position ID.
+  rowid SERIAL PRIMARY KEY,
+  -- Immutable position data, defining the trading function.
+  position_id BYTEA NOT NULL UNIQUE,
+  asset_1 BYTEA NOT NULL,
+  asset_2 BYTEA NOT NULL,
+  p NUMERIC(39) NOT NULL,
+  q NUMERIC(39) NOT NULL,
+  close_on_fill BOOLEAN NOT NULL,
+  fee_bps INTEGER NOT NULL,
+  effective_price_1_to_2 FLOAT8 NOT NULL,
+  effective_price_2_to_1 FLOAT8 NOT NULL,
+  position_raw BYTEA NOT NULL,
+  -- The time and height at which the position was opened, and its initial reserves.
+  opening_time TIMESTAMPTZ NOT NULL,
+  opening_height INTEGER NOT NULL,
+  opening_tx BYTEA,
+  opening_reserves_rowid INTEGER NOT NULL,
+  -- The time and height at which the position was closed, if it was closed.
+  closing_time TIMESTAMPTZ,
+  closing_height INTEGER,
+  closing_tx BYTEA
+);
+
+CREATE INDEX ON dex_ex_position_state (position_id);
+CREATE INDEX ON dex_ex_position_state (opening_tx);
+
+CREATE TABLE IF NOT EXISTS dex_ex_position_reserves (
+  rowid SERIAL PRIMARY KEY,
+  position_id BYTEA NOT NULL,
+  height INTEGER NOT NULL,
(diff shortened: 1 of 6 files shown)


crates/core/asset/src/asset/id.rs
```diff
@@ -205,4 +207,39 @@ mod tests {
...
+            "cc0d3c9eef0c7ff4e225eca85a3094603691d289aeaf428ab0d87319ad93a302", // USDY
+            "a7a339f42e671b2db1de226d4483d3e63036661cad1554d75f5f76fe04ec1e00", // SHITMOS
+            "29ea9c2f3371f6a487e7e95c247041f4a356f983eb064e5d2b3bcf322ca96a10", // UM
+            "76b3e4b10681358c123b381f90638476b7789040e47802de879f0fb3eedc8d0b", // USDC
+            "2923a0a87b3a2421f165cc853dbf73a9bdafb5da0d948564b6059cb0217c4407", // OSMO
+            "07ef660132a4c3235fab272d43d9b9752a8337b2d108597abffaff5f246d0f0f", // ATOM
+            "5314b33eecfd5ca2e99c0b6d1e0ccafe3d2dd581c952d814fb64fdf51f85c411", // TIA
+            "516108d0d0bba3f76e1f982d0a7cde118833307b03c0cd4ccb94e882b53c1f0f", // WBTC
+            "414e723f74bd987c02ccbc997585ed52b196e2ffe75b3793aa68cc2996626910", // allBTC
+            "bf8b035dda339b6cda8f221e79773b0fd871f27a472920f84c4aa2b4f98a700d", // allUSDT
```
hdevalence, line 223: I wrote this test while inspecting the data, and figured it would be useful to have a record of.

hdevalence:
No idea what the crate features error is about and the CI job doesn't say what the problem is.

crates/bin/pindexer/src/dex_ex/schema.sql
```diff
@@ -82,3 +82,105 @@ CREATE TABLE IF NOT EXISTS dex_ex_metadata (
...
+  reserves_2 NUMERIC(39) NOT NULL
+);
+
+CREATE INDEX ON dex_ex_position_reserves (position_id, height, rowid);
+
+CREATE TABLE IF NOT EXISTS dex_ex_position_executions (
+  rowid SERIAL PRIMARY KEY,
+  position_id BYTEA NOT NULL,
+  height INTEGER NOT NULL,
+  time TIMESTAMPTZ NOT NULL,
```
cronokirby, line 129: these are redundant if we link to the reserves, right?

crates/bin/pindexer/src/dex_ex/schema.sql
```diff
@@ -82,3 +82,105 @@ CREATE TABLE IF NOT EXISTS dex_ex_metadata (
...
+);
+
+CREATE INDEX ON dex_ex_position_executions (height);
+CREATE INDEX ON dex_ex_position_executions (position_id, height, rowid);
+
+CREATE TABLE IF NOT EXISTS dex_ex_position_withdrawals (
+  rowid SERIAL PRIMARY KEY,
+  position_id BYTEA NOT NULL,
+  height INTEGER NOT NULL,
+  time TIMESTAMPTZ NOT NULL,
```
cronokirby, line 155: Also redundant if we reference the reserves

hdevalence:
@cronokirby FYI the reason I pulled newer sqlx was to get a conversion impl for the numeric types that didn't exist before (not otherwise attached to it but i think it would require more custom code to use the older version)

crates/bin/pindexer/src/dex_ex/schema.sql
hdevalence: Yeah, my thought was that they would be useful to have on hand when looking at an execution individually and there's not an issue with normalization for this application

crates/bin/pindexer/src/dex_ex/schema.sql
hdevalence: Same reasoning

crates/bin/pindexer/src/dex_ex/schema.sql
hdevalence: Either way. Not particularly attached but also don't see a compelling reason not to have the data there.

cronokirby (approved):

PR #4950, opened 2024-12-03
UIP-8: Transparent Addresses

hdevalence:
## Describe your changes

This commit experiments with a "truncated" address format, which aims at maximizing compatibility with other chains. This is kind of a janky hack, but may be worth doing anyways -- and could allow users to work around current and future Bech32m encoding compatibility issues (though our goal should be to ensure compatibility).

### Pros

- Uses Bech32, like other Cosmos chains
- Fits in 32 bytes -- looks exactly like a cosmos except for a different Bech32 HRP
- As an alternate encoding, requires no changes to other parts of Penumbra, can be parsed directly into an `Address`

### Cons

- A third address format adds UX complexity
- The truncated address points at a random account number
- Only one truncated address per IVK (reuse of the address causes linkability)
- Not compatible with FMD (clues created with the truncated address are not detectable)

### How it works

A Penumbra address has three components: the diversifier `d`, the transmission key `pk_d`, and the clue key `ck_d`. To truncate the address to a 32-byte encoding:
- The diversifier is required to be `[0; 16]`, omitted from the encoding, and filled in with the hardcoded value during decoding;
- The transmission key is encoded as 32 bytes;
- The clue key is stripped, and filled in on the decoding side with the identity element.

The fixed diversifier means that there's only one per IVK, and also that the truncated address points at a random sub-account: the fixed value `[0u8; 16]` is decrypted with the diversifier key to a random address index. Stripping the clue key and filling it in with a dummy value means that generated clues are not detectable, but should not have other effects on the protocol, and since FMD is currently unused this seems fine.

To allow the use of truncated addresses in IBC transfers, the `Ics20Withdrawal` action would need to be changed to have a `use_truncated_address` parameter.

No issue - starting point for forum discussion

- [ ] ~~I have added guiding text to explain how a reviewer should test these changes.~~ N/A - starting point for discussion 


hdevalence:
```
Truncated address: ptrunc16t45lct6x4w6an6hvv7zp9p49qx0teg78qyukcgfk76ch87leggq0wlmdc
Reconstructed address: penumbra169wv6gf7r2fp0p0z6dsdlr7ntjetp2v2s4q77l434vw2qsdg3lqzlpyy6krgueygk0z3m2c77galpszc6q053nd4tjyu03d6hawlw4992ahdasgazqy54k39g2ezu536etjzkr
Address index: AddressIndex { account: 1587124636, randomizer: "c4c0afca894fb03acc8465c9" }
Actual address for index: penumbra1yy6frkygnsmut4ad47zwcnkg35ywmld3pekg28qge0t3c4y8vfjhaavadu8czpmlksgp6jht3q42h45qjf6fvyhet20vdgvspd76gj2h8fhaqa8raewhy48zp4c36zgw0nqj37
```

crates/core/keys/src/keys/diversifier.rs
```diff
@@ -107,6 +107,15 @@ impl DiversifierKey {
         let cipher = Aes128::new(&key);
         cipher.decrypt_block(&mut block);
 
+        // Special case for the null ciphertext. This enables transparent addresses (which have
+        // a null diversifier) to use account 0.
+        if diversifier.0.iter().all(|&b| b == 0) {
```
hdevalence, line 112: nit: might be simpler to check equality with `[0u8; 16]`?

hdevalence:
Since UIP-7 changes the `Ics20Withdrawal` action to deprecate the old compat encoding and allow specifying a t-addr, I think we'll want to have some other changes:

- [ ] Add an `Address::encode_as_transparent_address() -> Option<String>` that checks that `d`, `ck_d` are zero and returns `Some(encoding)` if so or `None` if not; refactor the existing generation/encoding logic through this method
- [ ] Add the new flag to the ICS20 withdrawal action
  - I think we should keep the old flag in the domain types even if it's ignored (so that we don't have issues parsing old txs -- prost will drop unknown fields causing things to stop round tripping)
- [ ] Change existing IBC logic to ignore the compat flag and use the t-addr flag instead
- [ ] Add cli flags to pcli that use the new field

Then we can test logic by creating a new testnet with this branch, connect to noble testnet, and make testnet USDC transfers with t-addresses

crates/core/keys/src/address.rs
```diff
@@ -319,7 +342,29 @@ impl std::str::FromStr for Address {
...
+
+            let ck_id = fmd::ClueKey([0u8; 32]);
+
+            let address = Self::from_components(dzero, pk_dzero, ck_id)
+                .ok_or_else(|| anyhow::anyhow!("could not reconstruct transparent address"))?;
+
+            // Verify this is a valid transparent address, bailing if not
+            if address.encode_as_transparent_address().is_none() {
+                return Err(anyhow::anyhow!("invalid transparent address components"));
+            }
```
hdevalence, line 364: don't we know this will always be the case because we just constructed `d`, `ck_d`?

proto/penumbra/penumbra/core/component/ibc/v1/ibc.proto
```diff
@@ -54,6 +54,9 @@ message Ics20Withdrawal {
   // Whether to use a "compat" (bech32, non-m) address for the return address in the withdrawal,
   // for compatability with chains that expect to be able to parse the return address as bech32.
   bool use_compat_address = 8;
```
hdevalence, line 56: we could slap a protobuf deprecation marker on this

redshiftzero: oh cool

crates/core/keys/src/address.rs
redshiftzero: yep.. yep we do

crates/bin/pcli/src/command/tx.rs
```diff
@@ -264,10 +264,15 @@ pub enum TxCmd {
         /// The selected fee tier to multiply the fee amount by.
         #[clap(short, long, default_value_t)]
         fee_tier: FeeTier,
-        /// Whether to use a Bech32(non-m) address for the withdrawal.
+        /// Whether to use a Bech32(non-m) address for the withdrawal (deprecated).
         /// Required for some chains for a successful acknowledgement.
         #[clap(long)]
         use_compat_address: bool,
```
redshiftzero, line 270: do we want to keep the `use_compat_address` for any reason or good to remove it from pcli?

cronokirby:
Just an fyi, I think we want to be directing this towards release/0.81.x instead of main, right?

redshiftzero:
oops yeah, thanks!

crates/bin/pcli/src/command/tx.rs
hdevalence: I think we could drop it from pcli

erwanor: Do we want to have a `pcli view address --transparent-address` similar to `pcli view address --compat`?

This will be handy for testing roundtrips.

cronokirby:
Some more PR maintenance is that I think we want to just pick out the commits from https://github.com/penumbra-zone/penumbra/pull/4950/commits/37b15f48402387b9ff66fdeb5a1263659ba739ea onward, so that this PR only has the transparent address stuff, with the missing commits on 0.80.X to be handled subsequently.

redshiftzero:
I'm gonna rewrite the git history here and force push to clean up these errant commits from `main`

cronokirby (approved):

conorsch:
N.B. I just rebased `release/v0.81.x` on top of latest main, 707192c24f620e8c3f15413cbe294f96f0e7fc37, to ensure we have the fixes from 0.80.10 that ensure relayer stability.

erwanor (approved):

PR #4954, opened 2024-12-09
pd: use `commit_in_place` to write app safeguard

erwanor:
**Describe your changes**

This PR:
- fixes a bug in the app version safeguard logic that would cause discrepancies between JMT state versions and block height
- refactors some of the inner logic to avoid writing any state unless it is necessary (i.e, no safeguard was found)

**Testing this:**

To observe the state discrepancy and test remediation:

1. Checkout the commit that only contains the extra tracing statement: `git checkout a8c51f0cacb64304a6b1781226d6f74376800af7`
2. Create a local devnet
3. Restart the devnet N times
4. Observe that the state version and block height have offset=N

To test that this fixes it:
1. Run this patch
2. Restart the node
3. Observe that the state version stays stable.

  > Not consensus breaking.

### DIFF
--- a/crates/bin/pd/src/main.rs
+++ b/crates/bin/pd/src/main.rs
@@ -19,8 +19,8 @@ use pd::{
         join::network_join,
     },
 };
-use penumbra_app::app_version::assert_latest_app_version;
-use penumbra_app::SUBSTORE_PREFIXES;
+use penumbra_app::app_version::check_and_update_app_version;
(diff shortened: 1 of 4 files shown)


cronokirby (approved):

conorsch:
Refs #4919.

conorsch (approved):
Thanks for the test plan, made it simple to verify the fix. :pray:

PR #4955, opened 2024-12-09
ibc(rpc): always do snapshot selection for proof apis

erwanor:
## Describe your changes

This PR:
- instruments IBC gRPC spans on `DEBUG`
- prefer using the `HostInterface` trait to hardcoding height/revision for gRPC responses
- systematically defer to the query height hint (in the gRPC headers) for **every API that returns a proof** (extending #4905, cc @noot)

We will be able to rip out all of the header selection logic once `https://github.com/cosmos/ibc-go/pull/7303` is merged.

  > RPC

### DIFF
--- a/crates/core/component/ibc/src/component/rpc/connection_query.rs
+++ b/crates/core/component/ibc/src/component/rpc/connection_query.rs
@@ -103,7 +103,8 @@ impl<HI: HostInterface + Send + Sync + 'static> ConnectionQuery for IbcQuery<HI>
         let snapshot = self.storage.latest_snapshot();
         let height = HI::get_block_height(&snapshot)
             .await
-            .map_err(|e| tonic::Status::aborted(format!("couldn't decode height: {e}")))?;
+            .map_err(|e| tonic::Status::aborted(format!("couldn't decode height: {e}")))?
+            + 1;
 
         let connection_counter = snapshot
             .get_connection_counter()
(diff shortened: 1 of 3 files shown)


crates/core/component/ibc/src/component/rpc/connection_query.rs
```diff
@@ -96,12 +95,15 @@ impl<HI: HostInterface + Send + Sync + 'static> ConnectionQuery for IbcQuery<HI>
...
 
     /// Connections queries all the IBC connections of a chain.
+    #[tracing::instrument(skip(self), err, level = "debug")]
     async fn connections(
         &self,
         _request: tonic::Request<QueryConnectionsRequest>,
     ) -> std::result::Result<tonic::Response<QueryConnectionsResponse>, tonic::Status> {
         let snapshot = self.storage.latest_snapshot();
-        let height = snapshot.version() + 1;
+        let height = HI::get_block_height(&snapshot)
```
avahowell, line 104: Is there an off-by-one here?

erwanor: I think so, but I am not sure what `ibc-go` does. They field it with https://github.com/cosmos/cosmos-sdk/blob/332d0b106d1b98538187b24394b912616c97b185/types/context.go#L73 so I think our behavior of doing `proof_height + 1` is coherent. I will ask to double check. Fixed, and thanks for the catch.

erwanor: Confirmed that's correct: https://github.com/cosmos/ibc-go/blob/main/modules/core/client/query.go#L68

conorsch (approved):

PR #4960, opened 2024-12-14
ci: add cargo-release to deps

conorsch:
Adds two more packages to the nix dev again:

  * cargo-release
  * buf

Both of those tools are critical during release-engineering, so I'm adding them to provide a deterministic way to generate the necessary build inputs.

In this commit the nix env switches to the recently published stable release for 2024-11. That stable release pulls in `buf` version 1.47.0, which is sufficiently recent to satisfy the "at least 1.32" stipulation we've had in the docs [0] for a while now.

[0] https://guide.penumbra.zone/dev/protobuf#installing-buf

  > dev-only, does not change application source code. 
  
 ## Testing and review
 
 If you want to verify behavior locally, following the docs for setting up the nix dev env https://guide.penumbra.zone/dev/dev-env and then check versions:
 
 ```
 ❯ buf --version
1.47.0

❯ cargo release --version
cargo-release 0.25.10
```

You should see _exactly_ those versions. Additionally, you should be able to run `just protos` and confirm no files were modified.


erwanor (approved):

conorsch:
I regenerated the proto descriptors on main using the new codified version of `buf`. This opens the door to removing the slightly-too-forgiving check in CI where we ignore modifications to these files: https://github.com/penumbra-zone/penumbra/blob/c65480910f9543f6409f1df57419b999bf6fb0a1/.github/workflows/buf-pull-request.yml#L72-L91 but I'm not altering the CI workflows right now. Can circle back on that on a rainy day, in the context of porting more of the CI to reuse the same nix env that we maintain for local dev.

PR #4962, opened 2024-12-16
shielded-pool(ics20): process truncated addresses in packet data

erwanor:
## Describe your changes

This PR makes sure that truncated addresses are encoded correctly when ICS20 withdrawals are marshaled into fungible packet data.

In addition to that, it modifies pcli to (hazardously?) handcraft a return `Address` with an identity clue key and zero diversifier.

To test this PR, one can performa IBC roundtrip against a testnet chain with broken bech32m handling (e.g, `grand-1`).

#4950 

  > CB release branch.

### DIFF
--- a/crates/core/component/shielded-pool/src/ics20_withdrawal.rs
+++ b/crates/core/component/shielded-pool/src/ics20_withdrawal.rs
@@ -177,8 +177,6 @@ impl From<Ics20Withdrawal> for pb::FungibleTokenPacketData {
             ordinary_return_address
         };
 
-        tracing::warn!(?return_address, w.use_transparent_address, "return address");
-
         pb::FungibleTokenPacketData {
             amount: w.value().amount.to_string(),
             denom: w.denom.to_string(),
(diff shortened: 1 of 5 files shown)


cronokirby:
Why can't the pcli command just use the`FullViewingKey::transparent_address()" method to create the address?

erwanor:
Good idea!

crates/core/component/shielded-pool/src/ics20_withdrawal.rs
```diff
@@ -167,9 +167,14 @@ impl TryFrom<pb::Ics20Withdrawal> for Ics20Withdrawal {
...
-            true => w.return_address.compat_encoding(),
-            false => w.return_address.to_string(),
+        let ordinary_return_address = w.return_address.to_string();
+
+        let return_address = if w.use_transparent_address {
+            w.return_address
+                .encode_as_transparent_address()
+                .unwrap_or_else(|| ordinary_return_address)
+        } else {
+            ordinary_return_address
```
erwanor, line 177: Unfortunately, this removed compat support in the protocol.

PR #4964, opened 2024-12-16
pd(migrate): trace and check state version divergence

erwanor:
## Describe your changes

This PR:
- logs state divergences when processing `EndBlock` messages
- prevents `pd` operators from migrating corrupted state unless `--force` is on

- [x] I have added guiding text to explain how a reviewer should test these changes: simple control flow.


zbuc (approved):
LGTM. Warning if the chain height drifts is good and I understand the intent to avoid a chain halt.

PR #4963, opened 2024-12-16
upate tendermint to 0.40.0

SuperFluffy:
## Describe your changes

This patch started out as an attempt to penumbra to use crates.io cnidarium, but quickly snowballed from there to include several connected changs (primarily due to a hard coupling with prost and/or tonic):

1. cnidarium@v0.81.0
2. tendermint@v0.40.0
3. tonic@v0.12.3 (and related crates such as tonic_build)
4. prost@v0.13.4 (and related crates such as prost_build)
5. pbjson@v0.7.0 (and related crates such as pbjson_build)
6. axum@v0.7.9 (induced by the update to tonic)
7. axum-extra@v0.9.6
8. http@v1.2.0 (induced by axum and tonic)
9. http-body@v1.0.1
10. http-body-util@v0.1.2
11. ics23@v0.12.0
12. ibc-proto@v0.51.1
13. ibc-types@v0.15.0
14. tower-abci@v0.18.0
15. metrics@0.24.1 (and related utilities such as metrics-exporter-prometheus and metrics-tracing-context)

## Checklist before requesting a review

- [ ] I have added guiding text to explain how a reviewer should test these changes.

- [ ] If this code contains consensus-breaking changes, I have added the "consensus-breaking" label. Otherwise, I declare my belief that there are not consensus-breaking changes, for the following reason:

  > REPLACE THIS TEXT WITH RATIONALE (CAN BE BRIEF)

PR #4973, opened 2024-12-24
penumbra: pull in latest tendermint/ibc crates

erwanor:
Continuation of #4963, into a release branch `v0.82.x` before tagging a release candidate at that version and publishing the workspace using an `alpha` version.

This handles the domain type change for upgradeable channels (https://github.com/penumbra-zone/ibc-types/pull/84) smoothly. It makes sure to write default values to the new fields, which avoids wire protocol changes, and makes this PR non consensus/state breaking.


erwanor:
Merging optimistically so we can experiment with `v0.82.0-alpha.1`

PR #4974, opened 2024-12-25
Fix quotation mark

taozui472:


taozui472:
Hi, the unexpected marks might confuse readers and be distracted by them. Also some editors might not render correctly depending on the software or platform used, appearing as gibberish or causing formatting problems. Hope it helps. @hdevalence

hdevalence:
Thanks for the PR but we will keep the Unicode quote marks.

Issue #4978, opened 2025-01-10
Publish workspace to crates.io


conorsch:
## Overview
We want to publish the crates in the protocol repo workspace to https://crates.io, so that external developers can depend on Penumbra code when writing downstream tooling. To date, we've recommended that external developers use git dependendencies in Rust projects, e.g.

```
penumbra-transaction = { git = "https://github.com/penumbra-zone/penumbra", rev = "ac7abacc9bb09503d6fd6a396bc0b6850079084e", features = [
   download-proving-keys",
] }
```

but doing so is [not possible](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-other-registries) for projects already publishing their crates to crates.io. Two notable projects we intend to support are Astria (#3125) and Hermes. We're already maintaining a [fork of Hermes](https://github.com/penumbra-zone/hermes), and publishing the crates would unblock us from submitting the Penumbra integration for Hermes to upstream.

## Work to date

There's currently some WIP around the workspace consolidation happening in the following branches:

* https://github.com/penumbra-zone/penumbra/tree/release/v0.82.x
* https://github.com/penumbra-zone/penumbra/tree/erwan/prepare_crates_io

That work should be appended to, and organized in a feature branch that can be PR'd into main. In the process, we'll likely attempt to resolve both #4400 & #4682.

erwanor:
Thanks for writing this up and tracking it, fwiw:

>In the process, we'll likely attempt to resolve both https://github.com/penumbra-zone/penumbra/issues/4400 & https://github.com/penumbra-zone/penumbra/issues/4682.

I think this is already done in v0.82.x branch. `v1` reflection was only released as part of `tonic@0.12` iirc.

The release branch should have all of the ecosystem bumps (tonic, ibc-types, tendermint, tower, etc.)

erwanor:
Thanks for the write-up, my understanding is that this is caused by the `rustls` backend change to `aws-lc-rs` (and away from `ring`). I think this might *potentially* cause platform issues down the road, but we'll figure it out when we get there (and hopefully, the ecosystem will beat us to it).

1. `cargo add aws-lc-rs`
2. `let _ = aws_lc_rs::default_provider().install_default();` **per process** (so, e.g, one in `main.rs in the pcli bin)
3. Tracking the `tls_config` calls to make sure that they configured with the bundled certificates: `tls_config(ClientTlsConfig::new().with_webpki_roots())?`

[...]

PR #4979, opened 2025-01-10
test(pcli): sync wallet against testnet

conorsch:
## Describe your changes

This test is off-by-default, given that it talks to a remote endpoint, and is rather slow. It can be run directly via:

  cargo nextest run --release -p pcli --features integration-testnet sync_wallet_on_public_testnet

As usual, make sure to include the `--release` flag, otherwise it'll be much slower.

Related to changes described in #4978. Specifically, there were changes made in #4973 that broke ClientTls configs for tonic connections, but our test suite didn't catch that. 

## Testing and review

1. Check this branch out locally and run `cargo nextest run --release -p pcli --features integration-testnet sync_wallet_on_public_testnet`
2. Confirm that test passes!

It's also worth considering clicking this on in CI so we get assurance; I left it off by default but plan to ask for it to be used in subsequent testing towards #4978. Maybe we should just stick it in CI, at least temporarily?

  > Test-only, no changes to app code. Intended to reduce the odds of breaking changes being merged.

### DIFF
--- a/crates/bin/pcli/Cargo.toml
+++ b/crates/bin/pcli/Cargo.toml
@@ -15,6 +15,7 @@ dist = true
 [features]
 default = ["std", "parallel", "download-proving-keys"]
 download-proving-keys = ["penumbra-proof-params/download-proving-keys"]
+integration-testnet = []
 sct-divergence-check = ["penumbra-view/sct-divergence-check"]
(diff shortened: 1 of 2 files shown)


erwanor (approved):
LGTM

PR #4981, opened 2025-01-11
Pindexer metrics fix

cronokirby:
## Describe your changes

This fixes two bugs in pindexer:
1. Volumes in aggregate and per pair summary were far too high.
2. The different windows of the aggregate summary all reported the same value.

Bug 1. stemmed from multiplying by a price rather than dividing by it when mixing two directions of a candle.

Bug 2. was just a missing filter in an SQL query.

To test, run pindexer again, and do some sanity checks on the aggregate summary, and pair summaries for pairs like USDC / UM.

  > indexing only


erwanor (approved):
This makes sense to me, merging optimistically.

PR #4982, opened 2025-01-13
fix(pindexer): support big amounts

conorsch:
## Describe your changes

Use I256 values in pindexer.

N/A

## Testing and review

I've already validated that the patch works when indexing against testnet; now I'd like to have @cronokirby's eyes on it to confirm it's a sane approach. 

  > indexing-only


cronokirby (approved):
Okay, everything should have the same behavior

PR #4985, opened 2025-01-14
protos(view): add a `subaccount` filter to the `OwnedPositionIds` rpc

erwanor:
## Describe your changes

This PR:
- adds an `AddressIndex` field to the `OwnedPositionIds` rpc request and response
- regenerate the codegen protobufs

It doesn't seem important to do the extra work to support filtering in the rust view server. The natural consumers of that information are web interfaces. Market makers using the rust implementation would probably simply integrate it in their own OMS rather than relying on the view service RPC.

One question I have is whether the buf publish action will be able to pickup the change to the protos. 

  > view service changes


conorsch (changes requested):
These changes look solid. 

> One question I have is whether the buf publish action will be able to pickup the change to the protos.

Yes, these changes will get published to BSR automatically, _however_ this PR should target the `main` branch so that upon merge the workflow will run. Requesting changes:

1. Rebase PR against main
2. Regen protos after rebase
3. Re-target PR to main

Then we should be good to go. I'll handle the point release mechanics in #4988.

PR #4986, opened 2025-01-15
feat: publish workspace crates

conorsch:
## Describe your changes

Workspace refactor, preparing to publish the Penumbra crates. In doing so, renames all crates to have a prefix `penumbra-sdk-`, which is unique and controlled by PL. There's also a custom script that handles the publication action, which is non-standard.

The major obstacle is that the `penumbra-sdk-proof-params` crate contains binary keyfiles, which are managed in git-lfs. We cannot upload the raw keyfiles to crates.io, because that'd result in a 100MB crate. Instead, we use a custom script to revert the binary keyfiles to plaintext lfs pointers immediately prior to publishing to crates.io, which stays under the limit, and allows third-party tools to opt into downloading the key material via the `download-proving-keys` feature.

Given that `docs.rs` doesn't permit network access during crate build, we default to _warning_ about the missing keys during build of proof-params, recommending enabling the option. We could go further add fail the build, conditionally warning only in the [docsrs build context](https://docs.rs/about/builds). I chose to minimize special-casing as much as possible.

Towards #4978.

## Testing and review

1. Browse the recently published "alpha" crates: https://crates.io/search?q=penumbra-sdk
2. Try updating a Rust project that depends on Penumbra, as the [hermes fork](https://github.com/penumbra-zone/hermes) or the [reindexer](https://github.com/penumbra-zone/reindexer), to depend on the published crates. 
3. Confirm that docs are available: https://docs.rs/penumbra-sdk-app/latest/penumbra_sdk_app/

  > Careful attention was paid to make this a non-breaking change. All the crates have been renamed, but the code behaves the same. Testing should focus on making sure that key material for the proof params crate is fetched correctly, to avoid breakage.


crates/crypto/proof-params/build.rs
```diff
@@ -1,6 +1,12 @@
-#[cfg(feature = "bundled-proving-keys")]
+//! The Penumbra proving and verification key files are binary
+//! data that must be provided at build time, so that the key material
+//! can be injected into Rust types. The key material is too large, however,
+//! for uploading to crates.io (with the keys the crate weights ~100MB).
+//!
+//! Instead, we'll upload just git raw git-lfs pointer files when publishing to crates.io,
+//! then use the build.rs logic to fetch the assets ahead of compilation. Use the feature
+//! `download-proving-keys` to enable the auto-download behavior.
 use anyhow::Context;
```
conorsch, line 9: The changes to the proof-params `build.rs` file warrants close scrutiny.

deployments/scripts/publish-crates
```diff
@@ -0,0 +1,72 @@
...
+  >&2 echo "$s"
+  git diff | head -n 128
+  exit 1
+fi
+
+# Swap out the full binary contents for the raw git-lfs pointers; suggestion via
+# https://github.com/git-lfs/git-lfs/issues/951#issuecomment-581477084
+# This is a destructive action! Which is why we checked for dirty tree above.
+git read-tree HEAD && GIT_LFS_SKIP_SMUDGE=1 git checkout -f HEAD
+
```
conorsch, line 59: The `GIT_LFS_SKIP_SMUDGE=1` approach is slightly cursed: we're altering in-repo assets immediately prior to crate publication. This works™, though, so I'm suggesting it as the right-sized solution to get the workspace published.

conorsch (commented):
Added some comments in-line to guide review.

erwanor (approved):

PR #4987, opened 2025-01-15
pindexer: dex_ex: calculate USDC volume continuously

cronokirby:
This adds a field to the low level snapshot used to derive other metrics, which captures the current price of a given asset relative to USDC (configurable to an arbitrary indexing denom).

This then allows a pair summary to have a more accurate view of the volume in USDC terms, and for the aggregate summary to thus have a more accurate view.

As a side-effect, this also gets rid of liquidity minimums for considering denoms in the aggregate summary.

To test, run pindexer again, compare the aggregate summary, and then check that the converted volume is about the same.

  > indexing only


erwanor (approved):

conorsch:
> To test, run pindexer again, compare the aggregate summary, and then check that the converted volume is about the same.

I haven't done this, I'm merging optimistically, and will reindex when other pindexer changes (e.g. #4990) land.

PR #4990, opened 2025-01-16
Erwan/pindexer swap execs

erwanor:
## Describe your changes

This PR adds a `dex_ex_batch_swap_traces` table that tracks batch swap execution traces.

Notably, the schema contains a price float, but no input/output amount floats for now. This is good to have, but we can add it later and it's not immediately useful to the intended consumer of this data.

```sql
-- This table tracks individual execution traces for a directed batch swap.
CREATE TABLE IF NOT EXISTS dex_ex_batch_swap_traces (
  -- Primary key
  rowid SERIAL PRIMARY KEY,

  -- The height of the block the batch swap was included in.
  height INTEGER NOT NULL,
  -- The time the batch swap was included in a block.
  time TIMESTAMPTZ NOT NULL,

  -- The amount of asset 1 consumed by the micro execution in raw denom.
  input NUMERIC(39) NOT NULL,
  -- The amount of asset 2 produced by the micro execution in raw denom.
  output NUMERIC(39) NOT NULL,

  -- The amount of asset 1 consumed by the macro execution.
  batch_input NUMERIC(39) NOT NULL,
  -- The amount of asset 2 produced by the macro execution.
  batch_output NUMERIC(39) NOT NULL,
  -- The price (output/input) as a float
  price_float DOUBLE PRECISION NOT NULL,


  -- The directed start asset of the batch swap.
  asset_start BYTEA NOT NULL,
  -- The directed end asset of the batch swap.
  asset_end BYTEA NOT NULL,

  -- Each hop in the list contains an asset id.
  asset_hops BYTEA[] NOT NULL,
  -- Each hop in the list contains an amount.
  amount_hops NUMERIC(39)[] NOT NULL,
  -- Each hop in the list contains a position ID.
  position_id_hops BYTEA[] NOT NULL
);

CREATE INDEX ON dex_ex_batch_swap_traces (time, height);
CREATE INDEX ON dex_ex_batch_swap_traces (asset_start, asset_end);
```

  > pindexer changes

### DIFF
--- a/crates/bin/pindexer/src/dex_ex/mod.rs
+++ b/crates/bin/pindexer/src/dex_ex/mod.rs
@@ -5,7 +5,11 @@ use cometindex::{
     AppView, PgTransaction,
 };
 use penumbra_asset::asset;
-use penumbra_dex::lp::position::{Id as PositionId, Position};
+use penumbra_dex::{
+    event::EventBatchSwap,
+    lp::position::{Id as PositionId, Position},
+    SwapExecution,
+};
 use penumbra_dex::{
     event::{
         EventCandlestickData, EventPositionClose, EventPositionExecution, EventPositionOpen,
@@ -668,6 +672,7 @@ struct Events {
     position_executions: Vec<EventPositionExecution>,
     position_closes: Vec<EventPositionClose>,
     position_withdrawals: Vec<EventPositionWithdraw>,
+    batch_swaps: Vec<EventBatchSwap>,
     // Track transaction hashes by position ID
     position_open_txs: BTreeMap<PositionId, [u8; 32]>,
     position_close_txs: BTreeMap<PositionId, [u8; 32]>,
@@ -686,6 +691,7 @@ impl Events {
             position_executions: Vec::new(),
             position_closes: Vec::new(),
             position_withdrawals: Vec::new(),
+            batch_swaps: Vec::new(),
             position_open_txs: BTreeMap::new(),
             position_close_txs: BTreeMap::new(),
             position_withdrawal_txs: BTreeMap::new(),
@@ -842,6 +848,8 @@ impl Events {
                 if let Some(tx_hash) = event.tx_hash {
                     out.position_close_txs.insert(e.position_id, tx_hash);
                 }
+            } else if let Ok(e) = EventBatchSwap::try_from_event(&event.event) {
+                out.batch_swaps.push(e);
             }
         }
         Ok(out)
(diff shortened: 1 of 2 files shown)


conorsch:
To review, I've kicked off a reindex of a mainnet db using this binary. As long as that completes without error, I'll call it good enough, and get it in.

conorsch:
Encountered an error:

```
Error: error returned from database: syntax error at or near ")"

Caused by:
    syntax error at or near ")"
```

Investigating...

erwanor:
Thanks, downloading the mainnet db, it's taking a second

crates/bin/pindexer/src/dex_ex/mod.rs
```diff
@@ -982,6 +990,109 @@ impl Component {
...
+                "INSERT INTO dex_ex_batch_swap_traces (
+                height,
+                time,
+                input,
+                output,
+                batch_input,
+                batch_output,
+                asset_start,
+                asset_end,
+                price_float,
```
cronokirby, line 1043:
```suggestion
                price_float,
                asset_start,
                asset_end,
```
to match the order stuff gets passed in later.

cronokirby (changes requested):
I think there's currently a bug in the data insertion method.

crates/bin/pindexer/src/dex_ex/mod.rs
erwanor: Thanks!

crates/bin/pindexer/src/dex_ex/mod.rs
```diff
@@ -982,6 +990,107 @@ impl Component {
...
+                input,
+                output,
+                batch_input,
+                batch_output,
+                price_float,
+                asset_start,
+                asset_end,
+                asset_hops,
+                amount_hops,
+               position_id_hops,
```
cronokirby, line 1046:
```suggestion
               position_id_hops
```

I think this should fix the SQL error (no trailing commas allowed :()

conorsch:
Tested again locally on e079188a5ced677e98ddd298c9900417f940d74f (the most recent commit), and this time I was able to complete the reindex based on mainnet events. So LGTM, with the changes that have been added.

conorsch (approved):
Tested again locally on e079188a5ced677e98ddd298c9900417f940d74f (the most recent commit), and this time I was able to complete the reindex based on mainnet events. So LGTM, with the changes that have been added.

conorsch:
Intentionally not merging until @cronokirby has had a chance to look closely, and updates their requested-changes status.

cronokirby (approved):
All good for me!

PR #4992, opened 2025-01-17
protos(view): add a subaccount filter to the `OwnedPositionIds` rpc

conorsch:
## Describe your changes

This PR resubmits the unmerged changes from https://github.com/penumbra-zone/penumbra/pull/4985, targeting `main` branch. Below is the original submission text for 4985, which is still accurate here. 

This PR:
- adds an `AddressIndex` field to the `OwnedPositionIds` rpc request and response
- regenerate the codegen protobufs

It doesn't seem important to do the extra work to support filtering in the rust view server. The natural consumers of that information are web interfaces. Market makers using the rust implementation would probably simply integrate it in their own OMS rather than relying on the view service RPC.

One question I have is whether the buf publish action will be able to pickup the change to the protos. 

  > view service changes


conorsch:
This PR implements the changes requested in https://github.com/penumbra-zone/penumbra/pull/4985#pullrequestreview-2556903682.

conorsch:
Downgraded this PR to draft, so that https://github.com/penumbra-zone/penumbra/pull/4993 can land first. Once that's done, I'll need to rebase this and regenerate the proto files, to ensure no conflicts.

erwanor (approved):

crates/view/src/client.rs
```diff
@@ -709,6 +709,7 @@ where
                 tonic::Request::new(pb::OwnedPositionIdsRequest {
                     trading_pair: trading_pair.map(TryInto::try_into).transpose()?,
                     position_state: position_state.map(TryInto::try_into).transpose()?,
+                    ..Default::default()
```
cronokirby, line 712: bit of a nit, but I would prefer that we explicitly mark the keys that we're defaulting, so that we don't forgot to edit this file if we add a field that *does* need to be filled.

cronokirby (approved):
one non-blocking nit, otherwise looks good to me

PR #5006, opened 2025-01-25
penumbra: use `cnidarium@0.83`

### LINKED ISSUE #5004: bug: cnidarium metrics emission broken
**Describe the bug**
`pd` no longer emits most `cnidarium`-related metrics. 

**Expected behavior**
cnidarium-related metrics are still exposed by pd for scraping by prometheus.

**Screenshots**

Image

See the time series above. The drop-off in metrics coincides with the release of v1, in which we started depending on the out-of-tree [cnidarium crate](https://github.com/penumbra-zone/cnidarium). 

**Additional context**
Given the namespacing changes, we'll need to refine how pd access the metrics exposed by the now-external cnidarium crate.

erwanor:
## Describe your changes

Depends on https://github.com/penumbra-zone/cnidarium/pull/9, let's just merge it once the release gets tagged and published. CI won't pass until then.

Fix #5004 

Check `localhost:9000`

  > metrics bug fix

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1284,9 +1284,9 @@ dependencies = [
 
 [[package]]
 name = "cnidarium"
-version = "0.82.1"
+version = "0.83.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7f0060de634f1f7411ab7464f1021f48ffdeb72fd63cf19553035fa766563828"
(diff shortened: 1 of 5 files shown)


conorsch:
Tacked on a test to sanity-check the metrics emission, so we can catch this type of breakage in CI. Would have helped with the big version bumps recently. Confirmed that the new test fails on main, but passes here.

conorsch (approved):

PR #5008, opened 2025-01-28
view service: add LatestSwaps

TalDerei:
## Describe your changes

extends `ViewService` by adding a new `LatestSwaps` method, enabling clients to query recent swap activity. 

precursor to https://github.com/penumbra-zone/web/pull/2005

references https://github.com/penumbra-zone/dex-explorer/issues/320


erwanor (approved):

crates/view/src/service.rs
```diff
@@ -1860,6 +1862,14 @@ impl ViewService for ViewServer {
     ) -> Result<tonic::Response<Self::UnbondingTokensByAddressIndexStream>, tonic::Status> {
         unimplemented!("unbonding_tokens_by_address_index currently only implemented on web")
     }
+
+    #[instrument(skip_all, level = "trace")]
+    async fn latest_swaps(
+        &self,
+        _request: tonic::Request<pb::LatestSwapsRequest>,
+    ) -> Result<tonic::Response<Self::LatestSwapsStream>, tonic::Status> {
+        unimplemented!("latest_swaps currently only implemented on web")
```
VanishMax, line 1871: question: why is it so? I see many methods here are not implemented, is it deliberate?

TalDerei: we define the rpc service method in [.proto](https://github.com/penumbra-zone/penumbra/pull/5008/files#diff-03b7341d5bf81ab9c8d8542d220d5ba4ae122bd3837531309742b685d7ec1619), but we stub the service as unimplemented to ensure the rust build compiles, otherwise it'll complain. In this context, it means that the `LatestSwaps` view service method is not supported functionality in pcli, but is available in prax. 

happy to pair at some point to walk how protobuf + domain type generation + rpc services work in the monorepo.

PR #5017, opened 2025-01-28
ci: remove need refinment workflow

erwanor:
## Describe your changes
Turns out leaving this commented out causes the workflow to fail, so let's just remove it.

CI


conorsch:
Refs https://github.com/penumbra-zone/penumbra/pull/5001

conorsch (approved):

PR #5019, opened 2025-01-28
Update LICENSE-MIT

maximevtush:
Updated the copyright year from 2022 to 2025 in the LICENSE file

### DIFF
--- a/LICENSE-MIT
+++ b/LICENSE-MIT
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2022 Penumbra Labs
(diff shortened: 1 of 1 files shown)


conorsch:
Thanks for your interest in the project!

We don't need to update the year in the license files: that year denotes the year in which the software was first published, so it's accurate as-is. If we wanted to keep it up to date, we could use a date range, such as "2022 - 2025", but even that seems excessive. Also, the project is dual-licensed, meaning that any change made to the MIT license should be made to the Apache license, as well.

I'm closing this PR, but please get in touch if you'd like to dig into any open issues.

PR #5020, opened 2025-01-29
Implement parameters for liquidity tournament control (in funding)

### LINKED ISSUE #5013: funding: add LQT chain parameters
The funding component adds the following chain parameters:

```protobuf
// Distribution configuration data.
message FundingParameters {
  // The percentage of gauge votes that an asset must receive to receive any rewards
  // from the liquidity tournament. It is expressed in percentage, between 0 and 100.
  uint64 liquidity_tournament_gauge_threshold_percent = 3;
  // The maximum number of individual liquidity positions to consider when allocating
  // incentives by volume. This prevents DoS issues related to processing tiny rewards.
  uint64 liquidity_tournament_max_position_count = 4;
  // The maximum nunmber of delegator identities to process in a single tournament.
  uint64 liquidity_tournament_max_delegator_count = 5;
  // The share of rewards that goes to delegators.
  // It is expressed in percents, between 0 and 100 e.g, 52% is 52. 
  uint64 liquidity_tournament_delegator_share_percent = 6;
}
```

We should also add the logic that writes the parameters to the chain if that's not already the case.

cronokirby:
Closes #5013.

Diverges slightly from the ADR draft in that it creates a nested proto sub-message specific to the LQT. My rationale here was that this is nicer than prefixing every relevant field with "liquidity_tournament", both in the proto definitions, but also in the Rust code, because that way we can have a default for that entire set of parameters, and then the code can explicitly mark the absence of that set as to be interpreted as this default. Also works nicer for the future, in which we might be using the funding component for other things, each of which can cleanly have its own set of parameters.

I also added a little utility type for representing percentages. Not
strictly necessary, but I think it's a good move towards type safety.
The motivation here was all of the annoyance of dealing with keeping
track of what u64s were bps or bps^2 or yadda yadda yadda in other parts
of the codebase.

I've added some relevant unit tests.

  > adds new parameters that can be voted on, so consensus-breaking.


erwanor (approved):

erwanor (changes requested):
Can we add the logic to write the funding parameters in the state on a clean chain bootstrap

erwanor:
Other than that LGTM

crates/core/component/funding/src/component.rs
```diff
@@ -30,7 +30,9 @@ impl Component for Funding {
     #[instrument(name = "funding", skip(state, app_state))]
     async fn init_chain<S: StateWrite>(mut state: S, app_state: Option<&Self::AppState>) {
         match app_state {
-            None => { /* Checkpoint -- no-op */ }
+            None => {
+                state.put_funding_params(FundingParameters::default());
```
erwanor, line 34: Ah. I misspoke, and I understand your previous point better, I was thinking of the other variant `Some(app_state)`. Here this is indeed a no-op. The `None` case is only

Let me fix it

erwanor (approved):

PR #5033, opened 2025-01-30
Implement check and execute for LQT votes

### LINKED ISSUE #5032: funding: implement tally logic

cronokirby:
Closes #5032.

The check and execute logic for the action handler is still missing the nullifier check, but there's an obvious insertion point for adding that logic.

Testing deferred.

### DIFF
--- a/crates/core/component/funding/src/action_handler/liquidity_tournament/mod.rs
+++ b/crates/core/component/funding/src/action_handler/liquidity_tournament/mod.rs
@@ -1,15 +1,22 @@
-use anyhow::Context as _;
+use anyhow::{anyhow, Context as _};
 use async_trait::async_trait;
-use cnidarium::StateWrite;
-use penumbra_sdk_asset::asset::Denom;
+use cnidarium::{StateRead, StateWrite};
+use cnidarium_component::ActionHandler;
+use penumbra_sdk_asset::{asset::Denom, Value};
+use penumbra_sdk_governance::StateReadExt as _;
+use penumbra_sdk_num::Amount;
 use penumbra_sdk_proof_params::DELEGATOR_VOTE_PROOF_VERIFICATION_KEY;
+use penumbra_sdk_sct::component::clock::EpochRead as _;
+use penumbra_sdk_sct::epoch::Epoch;
+use penumbra_sdk_stake::component::validator_handler::ValidatorDataRead as _;
+use penumbra_sdk_tct::Position;
 use penumbra_sdk_txhash::TransactionContext;
 
+use crate::component::liquidity_tournament::votes::StateWriteExt as _;
 use crate::liquidity_tournament::{
     proof::LiquidityTournamentVoteProofPublic, ActionLiquidityTournamentVote,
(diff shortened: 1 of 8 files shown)


crates/core/component/funding/src/action_handler/liquidity_tournament/mod.rs
```diff
@@ -109,14 +112,23 @@ impl ActionHandler for ActionLiquidityTournamentVote {
...
+        // 2. We can tally, as long as the nullifier hasn't been used yet (this round).
+        let nullifier = self.body.nullifier;
+        let nullifier_exists = state.get_lqt_spent_nullifier(nullifier).await.is_some();
+        anyhow::ensure!(
+            !nullifier_exists,
+            "nullifier {} already voted in epoch {}",
+            self.body.nullifier,
+            current_epoch.index
+        );
+        state.put_lqt_spent_nullifier(current_epoch.index, nullifier, TransactionId([0u8; 32]));
```
erwanor, line 130: Yeah I think we can revisit in a follow-up PR.

crates/core/component/funding/src/component/state_key.rs
```diff
@@ -12,5 +12,59 @@ pub mod lqt {
...
+                let rest = &mut bytes;
+                let (bytes_prefix, rest) = rest.split_at_mut(PREFIX_LEN);
+                let (bytes_asset, rest) = rest.split_at_mut(ASSET_LEN);
+                let (bytes_power, bytes_voter) = rest.split_at_mut(POWER_LEN);
+
+                bytes_prefix.copy_from_slice(&prefix(epoch_index));
+                bytes_asset.copy_from_slice(&asset.to_bytes());
+                bytes_power.copy_from_slice(&((!power).to_be_bytes()));
+                bytes_voter.copy_from_slice(&voter.to_vec());
+
```
erwanor, line 65: 👍

crates/core/component/funding/src/liquidity_tournament/action/mod.rs
```diff
@@ -34,6 +37,17 @@ pub struct LiquidityTournamentVoteBody {
     pub rk: VerificationKey<SpendAuth>,
 }
 
+impl LiquidityTournamentVoteBody {
+    /// Get the asset id that should be incentivized.
+    ///
+    /// This will return None if the denom is not a base denom.
+    pub fn incentivized_id(&self) -> Option<asset::Id> {
+        REGISTRY
+            .parse_denom(&self.incentivized.denom)
```
erwanor, line 46: I presume this still work for unknown denoms right?

erwanor (approved):
ACK writing a zero-ed TxId in the NF set is fine, we can follow up with another PR

crates/core/component/funding/src/liquidity_tournament/action/mod.rs
cronokirby: https://rustdoc.penumbra.zone/main/src/penumbra_sdk_asset/asset/registry.rs.html#70

Yeah, the case in which it returns None is only when it's a known non-base denom.

PR #5036, opened 2025-01-31
Modify get/put current source to take in just a transaction ID

cronokirby:
Right now this is only ever a transaction id, and for LQT work, we need to be able to just use a transaction ID, so it's better to reify that, and change a few uses to wrap it in a source, versus having to unwrap it, introducing a spurious failure case

Smoke tests should be sufficient.

  > This state is always initialized before being read, so there's no issue in changing the type we store under this state key ; reviewers should double check this.

### DIFF
--- a/crates/core/app/src/action_handler/transaction.rs
+++ b/crates/core/app/src/action_handler/transaction.rs
@@ -4,7 +4,7 @@ use anyhow::Result;
 use async_trait::async_trait;
 use cnidarium::{StateRead, StateWrite};
 use penumbra_sdk_fee::component::FeePay as _;
-use penumbra_sdk_sct::{component::source::SourceContext, CommitmentSource};
+use penumbra_sdk_sct::component::source::SourceContext;
 use penumbra_sdk_shielded_pool::component::ClueManager;
 use penumbra_sdk_transaction::{gas::GasCost as _, Transaction};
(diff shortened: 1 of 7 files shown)


conorsch:
Needs more work for CI to pass. Should this change be targeting the LQT feature branch? It's currently targeting main, which doesn't seem like the worst idea, but we'd have to rebase.

cronokirby:
Yeah, targetting main was intentional ; would be good occasion to rebase protocol/lqt_branch, which we want to do from time to time anyways

hdevalence:
do we need to apply this to main? it shouldn't have consensus breaking effects but it also doesn't seem like there's a need to merge it in there, vs bundling it with known breaking changes ?

conorsch:
I agree this should target the LQT branch, out of conservativism. However, post-merge, we should indeed rebase.

erwanor (approved):

erwanor:
LGTM. Let's just bundle this with the protocol breaking changes. It shouldn't be CB, but it touches storage and there's no real upside in getting it into main.

conorsch:
> However, post-merge, we should indeed rebase.

Done. Latest sha1 on `protocol/lqt_branch` is `103b2a12c00ae60aba48f447b762108aa31ce390`.

PR #5038, opened 2025-01-31
Plumb transaction id into LQT vote action handlers

cronokirby:
## Describe your changes

This uses the newly added ambient tx id to populate the LQT nullifier key, making the logic in check and execute final.

This also removes the addition of the tx id to the TransactionContext. My understanding was that this was thought to be necessary to plumb the tx id for this action. Not particularly opposed to keeping that though.

Testing deferred.


TalDerei (approved):
the top-level action handler has access to the transaction id, which we set using `put_current_source`. If stateless checks require it, we can re-augment the transaction context with the id as needed.

erwanor (approved):

Issue #5039, opened 2025-02-01
distributions: implement RPC services


TalDerei:
sub distributions rpcs:

1. RPC that lets a client learn about the current pool size

```ts
service DistributionsService {
  rpc CurrentLqtPoolSize(CurrentLqtPoolSizeRequest) returns (CurrentLqtPoolSizeResponse);
}

// Request for retrieving the pool size of the current epoch from the chain state. 
message CurrentLqtPoolSizeRequest {}

// Response containing the pool size for the current epoch.
message CurrentLqtPoolSizeResponse {
  // The current epoch index.
  uint64 epoch_index = 1;

  // The total LQT pool size for the current epoch.
  core.num.v1.Amount pool_size = 2;
}
```

2. RPC that lets a client learn about a previous tournament pool size

```ts
service DistributionsService {
  ...
  rpc LqtPoolSizeByEpoch(LqtPoolSizeByEpochRequest) returns (LqtPoolSizeByEpochResponse);
}

// Request for retrieving the pool size at a specific epoch.
message LqtPoolSizeByEpochRequest {
  // The epoch for which we want to retrieve the pool size.
  uint64 epoch = 1;
}

// Response containing the pool size at a specific epoch.
message LqtPoolSizeByEpochResponse {
  // The epoch for which the pool size is returned.
  uint64 epoch_index = 1;
  
  // The total LQT pool size for the given epoch.
  core.num.v1.Amount pool_size = 2;
}
```

PR #5040, opened 2025-02-01
distributions: implement rpc services

TalDerei:
## Describe your changes

- Implements rpc services for the distributions component in `DistributionService`
- indexes distributions state key by epoch index and writes to NV storage

@cronokirby had some suggestions (which can be discussed in this PR) for replacing the historical epoch RPC, so I'm punting on the impl for now.

references https://github.com/penumbra-zone/penumbra/issues/5039

  > LQT branch


erwanor:
Let's do both RPCs so we can play with it in pcli, we can remove it once we ground the indexing story. The data is there, just expose it.

crates/core/component/distributions/src/component/state_key.rs
```diff
@@ -4,8 +4,8 @@ pub fn staking_token_issuance_for_epoch() -> &'static str {
 }
 
 // The amount of LQT rewards issued for this epoch.
-pub fn lqt_reward_issuance_for_epoch() -> &'static str {
-    "distributions/lqt_reward_issuance_for_epoch"
+pub fn lqt_reward_issuance_for_epoch(epoch_index: u64) -> String {
```
erwanor, line 7: Let's create an `lqt` module so that the call site is `lqt::v1::budget::for_epoch` and make it return a `Vec<u8>` since we know we want it to go into NV storage which has a byte oriented keyspace

crates/core/component/distributions/src/component/rpc.rs
```diff
@@ -54,4 +54,25 @@ impl DistributionsService for Server {
...
+        &self,
+        request: tonic::Request<pb::LqtPoolSizeByEpochRequest>,
+    ) -> Result<tonic::Response<pb::LqtPoolSizeByEpochResponse>, tonic::Status> {
+        // Retrieve latest state snapshot.
+        let state = self.storage.latest_snapshot();
+        let epoch_index = request.into_inner().epoch;
+        let amount = state
+            .get_lqt_reward_issuance_for_epoch(epoch_index)
+            .await
+            .expect(&format!(
```
erwanor, line 68: FWIW: should avoid panicking, even if the error is isolated from the rest of the node ops. Prefer `tonic` errors with the right status, here a not found is sufficient i think (or zero, doesn't matter).

crates/core/component/distributions/src/component/view.rs
```diff
@@ -21,8 +21,15 @@ pub trait StateReadExt: StateRead {
...
 
     // Get the total amount of LQT rewards issued for this epoch.
-    fn get_lqt_reward_issuance_for_epoch(&self) -> Option<Amount> {
-        self.object_get(&state_key::lqt_reward_issuance_for_epoch())
+    async fn get_lqt_reward_issuance_for_epoch(&self, epoch_index: u64) -> Option<Amount> {
+        let key = &state_key::lqt_reward_issuance_for_epoch(epoch_index);
+
+        self.nonverifiable_get(&key.as_bytes())
+            .await
+            .expect(&format!(
```
erwanor, line 29: FWIW: the usage is `expect(INVARIANT)` where `INVARIANT` is the reason why it is okay to panic.

erwanor (approved):

Issue #5010, opened 2025-01-28
Tracking issue: LQT


erwanor:
Draft ADR: Link to follow

This is a tracking issue for end-to-end support of LQT in the Penumbra stack. 

Issues:

The TODO below is turned into Github issues that contain a sketch of an implementation plan. Either copy-pasted/augmented from the ADR, or a few words about why it is not necessary.

Release process:

1. PRs are made against `protocol/lqt_branch` and require review.
2. PRs are squashed into `protocol/lqt_branch`

Tasks:

1. DEX component
    1. [x] #5015
    2. [x] #5028
    3. [x] #5067
    4. [x] #5051 
2. SCT component
    1. [x] #5011
3. Distributions component
    1. [x] #5012
    2. [x] #5025
    3. [x] #5039
    4. [x] #5041
4. Funding component
    1. [x] #5013
    2. [x] #5014
    4. [x] #5029
    7. [x] #5033 
    5. [x] #5032
    8. [x] #5035 
    9. [x] #5045 
    10. [x] https://github.com/penumbra-zone/penumbra/issues/5047
    12. [x] https://github.com/penumbra-zone/penumbra/issues/5053
    11. [x] #5046
    13. [x] #5066
    14. [x] #5058
4. Prax: https://github.com/penumbra-zone/web/issues/2022
5. Indexing
    1. [x] update reindexer logic for v1 compat with mainnet: https://github.com/penumbra-zone/reindexer/issues/30
    2. [x] update reindexer logic to understand LQT events https://github.com/penumbra-zone/reindexer/issues/32
    1. [x] pindexer: create an Lqt `AppView` https://github.com/penumbra-zone/penumbra/pull/5113
    2. [x] add new ABCI event emission for chain param modifications https://github.com/penumbra-zone/penumbra/pull/5110
    3. [x] pindexer: prepare Lqt stats
    4. [x] pindexer: prepare volume leaderboard
    5. [x] pindexer: prepare LQT LP leaderboard
    6. [x] pindexer: prepare asset leaderboard
7. Pcli    
    1. [x] #5068
    2. [x] #5069
8. misc
    1. [ ] land minifront changes that trigger reconnections
    2. [ ] export minifront bundle into pd on the `protocol/lqt_branch` work

TalDerei:
prax modeling to be tracked in https://github.com/penumbra-zone/web/issues/2022

PR #5043, opened 2025-02-02
community pool: use concurrent build

TalDerei:
## Describe your changes

uses concurrent build for community pool transactions

### DIFF
--- a/crates/core/app/src/action_handler/actions/submit.rs
+++ b/crates/core/app/src/action_handler/actions/submit.rs
@@ -346,18 +346,23 @@ async fn build_community_pool_transaction(
     transaction_plan: TransactionPlan,
 ) -> Result<Transaction> {
     let effect_hash = transaction_plan.effect_hash(&COMMUNITY_POOL_FULL_VIEWING_KEY)?;
-    transaction_plan.build(
-        &COMMUNITY_POOL_FULL_VIEWING_KEY,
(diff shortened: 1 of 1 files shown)


crates/core/app/src/action_handler/actions/submit.rs
```diff
@@ -346,18 +346,23 @@ async fn build_community_pool_transaction(
...
+            },
+            &AuthorizationData {
+                effect_hash: Some(effect_hash),
+                spend_auths: Default::default(),
+                delegator_vote_auths: Default::default(),
+            },
+        )
+        .await
+        .expect(
+            "community pool transactions are always well-formed and should never fail to build",
```
TalDerei, line 364: Is this true? also general question, why is this logic embedded in the top-level action handler?

conorsch: We can switch from `.build()` to `.build_concurrent()` without introducing an `expect()`. Let's be cautious about introducing changes to how errors are handled.

conorsch (changes requested):
Looks sound, although I'm tagging in @erwanor for confirmation on consensus implications. My asks are:

1. target the `protocol/lqt_support` feature branch, as described in #5010
2. don't introduce an `expect` if we can avoid it

In the meantime I'm tagging this as consensus-breaking out of caution.

Issue #5041, opened 2025-02-01
distributions: add distributions events


TalDerei:
Events: pool size increase

```ts
message EventLqtPoolSizeIncrease {
  uint64 epoch = 1;
  core.num.v1.Amount increase = 2;
  core.num.v1.Amount new_total = 3;
}
```

PR #5044, opened 2025-02-02
distributions: add distributions events

TalDerei:
## Describe your changes

add LQT pool size increase event to the distributions component, emitted at the end of each block. 

references https://github.com/penumbra-zone/penumbra/issues/5041

  > LQT branch


crates/core/component/distributions/src/component.rs
```diff
@@ -146,10 +147,35 @@ trait DistributionManager: StateWriteExt {
...
             ?new_issuance,
-            "computed new lqt reward issuance for epoch {}",
+            "computed new lqt reward issuance for currentepoch {}",
             current_epoch.index
         );
-        Ok(self.set_lqt_reward_issuance_for_epoch(current_epoch.index, new_issuance))
+
+        // Retrieve the previous cumulative LQT issuance total from NV storage.
+        let previous_issuance = self
+            .get_cummulative_lqt_reward_issuance(current_epoch.index - 1)
```
TalDerei, line 156: cumulative issuance of previous epoch

cronokirby:
I think we should go ahead and add an event type for this, since we'll be writing an app view for the events here at some point. For emission, c.f. https://github.com/penumbra-zone/penumbra/blob/155802109b157bcbf9d13b2beeda4772b7b9ec32/crates/core/component/dex/src/component/circuit_breaker/value.rs#L43 as an example

crates/core/component/distributions/src/component.rs
```diff
@@ -143,13 +144,39 @@ trait DistributionManager: StateWriteExt {
...
+            .await
+            .ok_or_else(|| {
+                tonic::Status::not_found(format!(
```
erwanor, line 160: This is not an RPC, so we shouldn't return an RPC error. In fact, we shouldn't return an error at all. We do not want to interrupt component logic. Just default to zero.

crates/core/component/distributions/src/component.rs
```diff
@@ -143,13 +144,39 @@ trait DistributionManager: StateWriteExt {
...
+
+        // Emit an event for LQT pool size increase.
+        self.record_proto(event::event_lqt_pool_size_increase(
+            current_epoch.index,
+            new_issuance,
+            cumulative_issuance,
+        ));
+
+        self.set_lqt_reward_issuance_for_epoch(current_epoch.index, new_issuance);
+        self.set_cumulative_lqt_reward_issuance(current_epoch.index, cumulative_issuance);
```
erwanor, line 177: This method only gets called at the end of every epoch, so something is wrong here right?

crates/core/component/distributions/src/component.rs
```diff
@@ -143,13 +144,39 @@ trait DistributionManager: StateWriteExt {
...
+                    "failed to retrieve cumulative LQT issuance for epoch {} from non-verifiable storage",
+                    current_epoch.index,
+                ))
+            })?;
+
+        // Compute cumulative LQT issuance up until current epoch.
+        let cumulative_issuance = previous_issuance + new_issuance;
+
+        // Emit an event for LQT pool size increase.
+        self.record_proto(event::event_lqt_pool_size_increase(
```
erwanor, line 170: We should emit an event every block, but we don't need to do anything else, the goal for this event is to be a convenient tick for observability, that's all we need imo.

erwanor (changes requested):
I think there are a few bugs, we should just emit an event every block and populate it using the reward rule for the budget. No need for extra state keys or tracking

crates/core/component/distributions/src/component.rs
TalDerei: well in the way I interpreted this event and it's associated fields, the logic is correct. I'll modify it to instead emit an event every block

crates/core/component/distributions/src/component.rs
TalDerei: resolving as outdated

erwanor (approved):
LGTM!

erwanor (approved):
Great

PR #5048, opened 2025-02-04
ci: slightly faster rust checks

conorsch:
## Describe your changes

A bit of spring cleaning for the CI suite. These changes aim at faster run times on PRs, mostly via reusing common tooling, and reducing setup time. These changes _don't_ affect which tests run, out of an abundance of caution,  because we have ongoing development in a feature branch right now (#5010). 

The longest-running CI check is currently the check-all-features job, which clocks in at ~14m with these changes. That's an improvement over the ~20m or so it's been taking on other PRs. For full details on the changes, see the commit messages; to summarize:

* smoke tests use the existing tooling env
* most checks can now be run locally via a justfile target, e.g. `just check`, to make CI failures more easily replicable locally
* test names have been reformatted for readability
* the noisy "release" checks are removed from PR runs

Overall the checks seem to complete >5m faster than before, which is good enough to merge as progress. I'm deferring more substantive changes like kicking long-running jobs to schedules, rather than on-every-PR, until we have a chance to observe these improvements over time, and until after the LQT push. 

### Example of CI job display _before_ changes

pz-gh-ci-before

### Example of CI job display _after_ changes

pz-gh-ci-after

N/A 

  > ci and test code only, application logic remains unchanged


conorsch:
The status checks being renamed will require an update to the branch-protection rules. The two (2) that were required were the buf checks. I can handle that in follow-up after merge.

cronokirby (approved):
LGTM

PR #5049, opened 2025-02-04
LQT: reimplementation with minimal in-memory state

cronokirby:
## Describe your changes

This leans much more heavily on the database for indexing and ranking,
such that the only bit of in-memory storage is just the number of assets
that can be considered at most, based on the threshold.

Downside is that more of the logic exists as the sum of the behavior across state keys,
and some tests were removed.

Adding some tests at the component level here which read and write against temp storage would be useful ; perhaps mocking the implementation of reward distribution or something.


crates/core/component/funding/src/component/liquidity_tournament/rewards/mod.rs
```diff
@@ -1,64 +1,118 @@
...
-    let mut tallies = Vec::new();
-    while let Some(x) = stream.next().await {
-        let (_, lp, volume) = x?;
-        let volume = volume.value();
-        total += volume;
-        tallies.push((Share::from(volume), lp));
+) -> anyhow::Result<u128> {
+    let mut out = 0u128;
+    let mut stream = state.ranked_voters(epoch, asset).take(
+        usize::try_from(params.max_delegators).expect("max delegators should fit in a usize"),
```
erwanor, line 32: nice, avoids platform specific behavior

crates/core/component/funding/src/component/liquidity_tournament/rewards/mod.rs
```diff
@@ -1,64 +1,118 @@
...
-    let mut stream = state.positions_by_volume_stream(epoch, asset)?.take(top_n);
-    let mut total = 0u128;
-    // This will end up containing the volume of each lp.
-    let mut tallies = Vec::new();
-    while let Some(x) = stream.next().await {
-        let (_, lp, volume) = x?;
-        let volume = volume.value();
-        total += volume;
-        tallies.push((Share::from(volume), lp));
+) -> anyhow::Result<u128> {
```
erwanor, line 29: I think we should use our `Amount` type everywhere, mostly a lesson learned from doing the 64 -> 128 bits migration

crates/core/component/funding/src/component/liquidity_tournament/rewards/mod.rs
```diff
@@ -1,64 +1,118 @@
...
+    state: &impl StateRead,
+    params: &LiquidityTournamentParameters,
+    epoch: u64,
+    asset: asset::Id,
+    total: u128,
+) -> impl Stream<Item = anyhow::Result<(Address, U128x128, TransactionId)>> + Send + 'static {
+    state
+        .ranked_voters(epoch, asset)
+        .take(usize::try_from(params.max_delegators).expect("max delegators should fit in a usize"))
+        .map_ok(move |((votes, voter), tx)| (voter, create_share(votes, total), tx))
```
erwanor, line 76: LGTM

crates/core/component/funding/src/component/liquidity_tournament/rewards/mod.rs
```diff
@@ -1,64 +1,118 @@
...
+    params: &LiquidityTournamentParameters,
+    epoch: u64,
+    asset: asset::Id,
+    total_volume: u128,
+) -> impl Stream<Item = anyhow::Result<(position::Id, U128x128)>> + Send + 'static {
+    state
+        .positions_by_volume_stream(epoch, asset)
+        .expect("should be able to create positions by volume stream")
+        .take(usize::try_from(params.max_positions).expect("max positions should fit in a usize"))
+        .map_ok(move |(_, lp, volume)| (lp, create_share(volume.value(), total_volume)))
```
erwanor, line 106: LGTM

crates/core/component/funding/src/component/state_key.rs
```diff
@@ -15,78 +15,247 @@ pub mod lqt {
...
+
+                        bytes_part0.copy_from_slice(PART0.as_bytes());
+                        bytes_epoch.copy_from_slice(format_epoch(epoch_index).as_bytes());
+                        bytes_part1.copy_from_slice(PART1.as_bytes());
+                        bytes_part2.copy_from_slice(PART2.as_bytes());
+
+                        bytes
+                    }
+
+                    const KEY_LEN: usize = PREFIX_LEN + POWER_LEN + ASSET_LEN;
```
erwanor, line 106: nit: `EPOCH_PREFIX_LEN` so we can read the schema in one-line. i think that's my main feedback. the code is more robust, if we have a single place to read the schema, it's perfect

crates/core/component/funding/src/component/liquidity_tournament/votes/mod.rs
```diff
@@ -2,43 +2,195 @@ use std::pin::Pin;
...
+use penumbra_sdk_sct::component::source::SourceContext as _;
+use penumbra_sdk_txhash::TransactionId;
 
 use crate::component::state_key;
 
-pub trait StateReadExt: StateRead {
-    fn vote_receipts(
+async fn get_u64(state: impl StateRead, key: &[u8]) -> u64 {
+    state
+        .nonverifiable_get_raw(key)
```
erwanor, line 15: Should use `StateReadProto` and `StateWriteProto` so everything is uniformly encoded using the overlay

crates/core/component/funding/src/component/liquidity_tournament/votes/mod.rs
```diff
@@ -2,43 +2,195 @@ use std::pin::Pin;
...
+        .await
+        .ok()
+        .and_then(|x| Some(u64::from_le_bytes(TryInto::<[u8; 8]>::try_into(x?).ok()?)))
+        .unwrap_or_default()
+}
+
+async fn put_u64(mut state: impl StateWrite, key: Vec<u8>, value: u64) {
```
erwanor, line 22: We can remove the get/put shim if we use `Amount + StateRead/WriteProtos`

crates/core/component/funding/src/component/liquidity_tournament/votes/mod.rs
```diff
@@ -2,43 +2,195 @@ use std::pin::Pin;
...
+    mut state: impl StateWrite,
+    epoch_index: u64,
+    asset: asset::Id,
+    power: u64,
+    voter: &Address,
+    tx: TransactionId,
+) {
+    state.nonverifiable_put_raw(
+        state_key::lqt::v1::votes::by_voter::ranked::key(epoch_index, asset, power, voter).to_vec(),
+        tx.0.to_vec(),
```
erwanor, line 107: Same thing here, we use the proto overlay to get consistent codecs across the protocol

crates/core/component/funding/src/component/liquidity_tournament/votes/mod.rs
```diff
@@ -2,43 +2,195 @@ use std::pin::Pin;
...
+
+fn delete_ranked_voter(
+    mut state: impl StateWrite,
+    epoch_index: u64,
+    asset: asset::Id,
+    power: u64,
+    voter: &Address,
+) {
+    state.nonverifiable_delete(
+        state_key::lqt::v1::votes::by_voter::ranked::key(epoch_index, asset, power, voter).to_vec(),
```
erwanor, line 119: One minor thing, but good to note: since we know we use NV storage, which has a byte oriented keyspace, we can make the state key return owned vectors or arrays.

crates/core/component/funding/src/component/liquidity_tournament/votes/mod.rs
```diff
@@ -2,43 +2,195 @@ use std::pin::Pin;
...
     }
+
+    fn ranked_voters(
+        &self,
+        epoch: u64,
+        asset: asset::Id,
+    ) -> Pin<Box<dyn Stream<Item = anyhow::Result<((u64, Address), TransactionId)>> + Send + 'static>>
+    {
+        self
+        .nonverifiable_prefix_raw(&state_key::lqt::v1::votes::by_voter::ranked::prefix(
```
erwanor, line 148: Proto overlay usage makes this method even more compact

crates/core/component/funding/src/component/liquidity_tournament/votes/mod.rs
```diff
@@ -2,43 +2,195 @@ use std::pin::Pin;
...
-            Vec::default(),
-        );
-        Ok(())
+    async fn tally(&mut self, epoch: u64, asset: asset::Id, power: u64, voter: &Address) {
+        let tx = self.get_current_source().expect("source should be set");
+        // Increment total.
+        {
+            let current = get_total(&self, epoch).await;
+            put_total(&mut *self, epoch, add_power(current, power)).await;
+        }
```
erwanor, line 176: LGTM

crates/core/component/funding/src/component/liquidity_tournament/votes/mod.rs
```diff
@@ -2,43 +2,195 @@ use std::pin::Pin;
...
+        // Adjust asset indices.
+        {
+            let current = get_asset_total(&self, epoch, asset).await;
+            delete_ranked_asset(&mut *self, epoch, current, asset);
+            let new = add_power(current, power);
+            put_asset_total(&mut *self, epoch, asset, new).await;
+            put_ranked_asset(&mut *self, epoch, new, asset);
```
erwanor, line 183: nit, take it or leave it: harmonize signatures, or is it on purpose to match the schema?

crates/core/component/funding/src/component/liquidity_tournament/votes/mod.rs
```diff
@@ -2,43 +2,195 @@ use std::pin::Pin;
...
+            delete_ranked_voter(&mut *self, epoch, asset, current, voter);
+            let new = add_power(current, power);
+            put_voter_total(&mut *self, epoch, asset, voter, new).await;
+            put_ranked_voter(&mut self, epoch, asset, new, voter, tx);
+        }
     }
 }
 
-impl<T: StateWrite + ?Sized> StateWriteExt for T {}
+impl<T: StateWrite> StateWriteExt for T {}
```
erwanor, line 196: Curious why restrict the bound here? the ?Sized is to work with trait objs iirc

crates/core/component/funding/src/component/liquidity_tournament/votes/mod.rs
cronokirby: I always forget that the primitive types implement the right traits 😅

crates/core/component/funding/src/component/liquidity_tournament/votes/mod.rs
cronokirby: Yeah the order is trying to match the schema

crates/core/component/funding/src/component/liquidity_tournament/votes/mod.rs
cronokirby: I ended up needing to add the sized constraint because the guts of the method now need borrows, iirc

erwanor (approved):

erwanor:
Send it!

PR #5050, opened 2025-02-04
chore: bump cometbft version to 0.37.15

conorsch:
## Describe your changes

The cometbft version bump is to include patches to security issues. See details in [0]. We already updated the project documentation in [1].

[0] https://github.com/cometbft/cometbft/blob/v0.37.15/CHANGELOG.md#v03715
[1] https://github.com/penumbra-zone/guide/pull/52

  > no changes to app code, bumping deps for related tooling

### DIFF
--- a/deployments/compose/docker-compose.dev.yml
+++ b/deployments/compose/docker-compose.dev.yml
@@ -57,7 +57,7 @@ services:
 
   # The CometBFT node
   cometbft-node0:
-    image: "docker.io/cometbft/cometbft:v0.37.9"
+    image: "docker.io/cometbft/cometbft:v0.37.15"
     ports:
       - "26656:26656"
       - "26657:26657"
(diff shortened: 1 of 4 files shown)


deployments/compose/docker-compose.dev.yml
```diff
@@ -57,7 +57,7 @@ services:
 
   # The CometBFT node
   cometbft-node0:
-    image: "docker.io/cometbft/cometbft:v0.37.9"
+    image: "docker.io/cometbft/cometbft:v0.37.15"
```
conorsch, line 60: Frankly I'm in favor of deleting the docker-compose files, because I don't believe they're used anywhere, but they may serve as a reference to someone trying to write their own, so leaving in-tree for now.

erwanor (approved):

PR #5055, opened 2025-02-05
bump tower@v0.5.2, tower-abci@v0.19.1, tower-actor@v0.2.0

SuperFluffy:
## Describe your changes

Bump dependencies tower@v0.5.2, tower-abci@v0.19.1, tower-actor@v0.2.0

Note that this is currently blocked on https://github.com/penumbra-zone/tower-actor/pull/5 being merged so that the `patch` section in the workspace `Cargo.toml` can be removed.

tower 0.4.13 still shows up in `Cargo.lock` as a dependency of axum-server@v0.7.1 and tonic@v0.12.3.

tower@0.5 is already a dependecy in tonic@master as of this PR.

Note for a penumbra-sdk-tower-trace release this consitutes a breaking change (because tower is part of its public API, for example through the trait bount `penumbra_sdk_tower_trace::trace:::InstrumentableService where Self::Service`).

  > Updating core dependencies to ensure matching semver.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1303,14 +1303,14 @@ dependencies = [
  "parking_lot",
  "pbjson",
  "pin-project",
- "prost 0.13.4",
+ "prost",
  "regex",
  "rocksdb",
  "serde",
  "sha2 0.10.8",
  "smallvec",
  "tempfile",
- "tendermint 0.40.1",
+ "tendermint",
  "tokio",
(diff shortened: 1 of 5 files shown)


SuperFluffy:
The error points to rand core, which this PR isn't actively touching. I have to take a step back but can look at this later (unless somebody else gets to it who has a better understanding of things).

Issue #1667, opened 2022-11-26
Tags are not pushed to Buf schema registry


hdevalence:
**Describe the bug**

Looking at https://buf.build/penumbra-zone/penumbra/tags/main, we just have a bunch of commit IDs and not any of our actual tagged versions.

I suspect this is because we're not pushing to buf on tags, only on deploys to main.

This is a problem because we would like API users to be able to pick a tagged version of the protobuf definitions to use against a tagged testnet.

hdevalence:
We should be sure to fix this before the next testnet release.

conorsch:
We suspect we just need to run the relevant github actions workflows on tag pushes. That should accommodate. I'll pick this one up.

conorsch:
Looks like #1690 didn't resolve this; I still don't see any named tags on the buf build page. Re-opening to track follow-up.

conorsch:
Looks like the buf-push action [does not support pushing named git tags](https://github.com/bufbuild/buf-push-action/issues/20) natively. In buf, a "tag" maps to the git _commit_ sha1 by default. Now that buf [supports multiple `--tag` options](https://github.com/bufbuild/buf/issues/575), we should be able to wire this up ourselves as manual workflow steps.

hdevalence:
@conorsch Do we still want to do this?

conorsch:
Yes, this would be hugely helpful for downstream implementers consuming the buf-built codegen: right now, the usual practice is "pull the latest" and try to time that action near release day. It'd be much saner if downstream could just say "I want v0.63.2" and pin that version explicitly. Looks like there's finally a pending PR to add this functionality to the GHA we use. I'll take a look and see if we can push tags from local CLI for a handful of our most recent tags.

conorsch:
I just pushed a few historical tags by checking them out in git and using `buf push --tag <tag>`. It didn't like when I tried point releases, though:

```
❯ git checkout v0.63.0
Previous HEAD position was 7b99503f3 chore: release version 0.62.0
HEAD is now at cff167a33 chore: release version 0.63.0

❯ buf push --tag v0.63.0
2b51f85bc0ed40c99c3e453a744c6cad

❯ git checkout v0.63.1
Previous HEAD position was cff167a33 chore: release version 0.63.0
HEAD is now at d24587cf2 chore: release version 0.63.1

❯ buf push --tag v0.63.1
already_exists: new commit identical to latest on branch

❯ git checkout v0.63.2
[...]

PR #5056, opened 2025-02-05
ci: make workflows runnable ad-hoc

conorsch:
## Describe your changes

We want to publish the protobuf changes in `protocol/lqt_support` (#5010) to unblock web integrations. In order to run the workflow ad-hoc on a branch, we need to permit `workflow_dispatch` runs. The `workflow_call` additions are not currently used, but have been in the past, and permit triggering via cross-repo API.

While authoring these changes, I noticed that the protobuf lint job is _not_ running against feature branches, meaning the PRs into `protocol/lqt_support` have skipped that check. We've been careful in authoring the proto changes, but we should still verify that check passes on the feature branch, prior to publishing.

Refs #5010. 

  > no code changes, only affects CI. will result in new protobuf changes to published to the repo.

### DIFF
--- a/.github/workflows/buf-pull-request.yml
+++ b/.github/workflows/buf-pull-request.yml
@@ -1,9 +1,10 @@
 name: protobuf
 on:
-  # Exclude feature branches, only run if the PR is targeting main.
+  # Run against all PRs, regardless of feature branch target.
   pull_request:
-    branches:
-      - "main"
+  # Also support ad-hoc calls for workflow.
+  workflow_call:
+  workflow_dispatch:
 jobs:
   # Ensure there are no breaking changes to the protocol specs,
   # by running the "buf lint" action against the changes in this PR.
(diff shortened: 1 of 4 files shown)


conorsch:
Sadly no, see #1667 for context. Will revisit that issue because it looks like upstream buf tooling may finally support meaningful tags.

conorsch:
> We've been careful in authoring the proto changes, but we should still verify that check passes on the feature branch, prior to publishing.

Good news, the buf lint action passes on the [latest LQT changes](https://github.com/penumbra-zone/penumbra/actions/runs/13163839445/job/36739057502). To test, I branched from `protocol/lqt_support`, rebased on main, and ran the workflow against that new branch. The rebasing was necessary, because the changes from this PR 5056 are not (yet) in the `protocol/lqt_support` branch, so the workflow cannot be run ad-hoc there:

ci-buf-attempt

Issue #5058, opened 2025-02-05
funding: implement query services


TalDerei:
implement funding rpc that allows the client to track whether a nullifier has already been used in the current epoch’s voting. 

```ts
service FundingService {
  // Checks if a particular nullifier has already been used in the current epoch.
  rpc LqtCurrentEpochVoted(LqtCurrentEpochVotedRequest) returns (LqtCurrentEpochVotedResponse);
}

// Request if a nullifier is already used in the current epoch.
message LqtCurrentEpochVotedRequest {
  // The nullifier to check.
  penumbra.core.sct.v1.Nullifier nullifier = 1;
}

// Response indicating whether the nullifier is used.
message LqtCurrentEpochVotedResponse {
  // Transaction Id where the vote occurred.
  txhash.v1.TransactionId tx_id = 1;
  // True if the nullifier has been used for voting in the current epoch.
  bool already_voted = 2;
}
```

PR #5061, opened 2025-02-05
funding: implement rpc services

TalDerei:
## Describe your changes

implement funding rpc that allows the client to track whether a nullifier has already been used in the current epoch’s voting period. This _isn't_ a bulk nullifier point query. 

references https://github.com/penumbra-zone/penumbra/issues/5058

  > LQT branch


proto/penumbra/penumbra/core/component/funding/v1/funding.proto
```diff
@@ -7,6 +7,27 @@ import "penumbra/core/component/shielded_pool/v1/shielded_pool.proto";
 import "penumbra/core/keys/v1/keys.proto";
 import "penumbra/core/num/v1/num.proto";
 import "penumbra/crypto/decaf377_rdsa/v1/decaf377_rdsa.proto";
+import "penumbra/core/txhash/v1/txhash.proto";
+
+// Query operations for the funding component.
+service FundingService {
+  // Checks if a particular nullifier has already been used in the current epoch.
+  rpc LqtCurrentEpochVoted(LqtCurrentEpochVotedRequest) returns (LqtCurrentEpochVotedResponse);
```
erwanor, line 15:
```suggestion
  rpc LqtCheckNullifier(LqtCheckNullifierRequest) returns (LqtCheckNullifierResponse);
```

proto/penumbra/penumbra/core/component/funding/v1/funding.proto
```diff
@@ -7,6 +7,27 @@ import "penumbra/core/component/shielded_pool/v1/shielded_pool.proto";
...
+}
+
+// Request if a nullifier is already used in the current epoch.
+message LqtCurrentEpochVotedRequest {
+  // The nullifier to check.
+  sct.v1.Nullifier nullifier = 1;
```
erwanor, line 21: I would add an `epoch_index`. If it's present, check the specific epoch. If it's absent, check the current one.

proto/penumbra/penumbra/core/component/funding/v1/funding.proto
```diff
@@ -7,6 +7,27 @@ import "penumbra/core/component/shielded_pool/v1/shielded_pool.proto";
...
+}
+
+// Response indicating whether the nullifier was already used.
+message LqtCurrentEpochVotedResponse {
+  // TransactionId where the vote occurred.
```
erwanor, line 26: I would add an `epoch_index` to make it easy for clients to anchor the response

proto/penumbra/penumbra/core/component/funding/v1/funding.proto
```diff
@@ -7,6 +7,27 @@ import "penumbra/core/component/shielded_pool/v1/shielded_pool.proto";
...
+  txhash.v1.TransactionId tx_id = 1;
+  // True if the nullifier has been used for voting in the current epoch.
+  bool already_voted = 2;
```
erwanor, line 29: Technically redundant, but probably makes the API a little more friendly. I would just add a comment that the tx_id is always present if this is true.

erwanor (approved):
suggestions

crates/core/component/funding/src/component/rpc.rs
```diff
@@ -0,0 +1,68 @@
...
+            let current_epoch = state.get_current_epoch().await.map_err(|e| {
+                tonic::Status::internal(format!("failed to retrieve current epoch: {e}"))
+            })?;
+            current_epoch.index
+        } else {
+            req_inner.epoch_index
+        };
+
+        // Perform a state nullifier lookup.
+        let maybe_tx_id = state.get_lqt_spent_nullifier(nullifier).await;
```
erwanor, line 50: Oh we're not using the `epoch_index`

TalDerei: oh ooops!

TalDerei: good catch

TalDerei: @erwanor actually the state key only operates on the current ambient epoch

erwanor: Can add a getter that takes both, it's going to be handy to test things in pcli too

crates/core/component/funding/src/component/liquidity_tournament/nullifier/mod.rs
```diff
@@ -29,6 +30,26 @@ pub trait NullifierRead: StateRead {
...
     }
+
+    /// Returns the `TransactionId` if the nullifier has been spent in a specific epoch;
+    /// Otherwise, returns None.
+    async fn get_lqt_spent_nullifier_by_epoch(
+        &self,
+        nullifier: Nullifier,
+        epoch_index: u64,
+    ) -> Option<TransactionId> {
+        let nullifier_key = &state_key::lqt::v1::nullifier::key(epoch_index, &nullifier);
```
TalDerei, line 41: could also query the current epoch first and check if epoch_index > current_epoch to help distinguish between non-existent epoch, but this might be an overkill. think it's fine

erwanor:
LGTM, merge on green CI!

PR #5062, opened 2025-02-06
view service: Implement LQT rpcs

TalDerei:
## Describe your changes

Implements `LqtVotingNotes` and `TournamentVotes` view service rpcs.

  > LQT branch


proto/penumbra/penumbra/view/v1/view.proto
```diff
@@ -152,6 +152,55 @@ service ViewService {
...
+}
+
+message LqtVotingNotesRequest {
+  // The epoch index to check for nullifier usage.
+  uint64 epoch_index = 1;
+  // The nullifier whose voting status is being queried.
+  core.component.sct.v1.Nullifier nullifier = 2;
+}
+
+message LqtVotingNotesResponse {
```
erwanor, line 171: That doesn't seem right, because this RPC is supposed to return SNRs that can be used to vote in the LQT

proto/penumbra/penumbra/view/v1/view.proto
```diff
@@ -152,6 +152,55 @@ service ViewService {
...
+    // The incentivized asset receiving this vote.
+    core.asset.v1.AssetId incentivized_asset = 1;
+    // The total amount of voting power allocated to this asset.
+    core.num.v1.Amount vote_power = 2;
+    // The total amount of rewards issued for votes on this asset.
+    core.asset.v1.Value reward = 3;
+  }
+  
+  // A list of votes cast for different incentivized assets.
+  repeated Vote votes = 1;
```
erwanor, line 203: One thing I forgot to add is the `TransactionId` for the vote

proto/penumbra/penumbra/view/v1/view.proto
```diff
@@ -152,6 +152,55 @@ service ViewService {
...
+}
+
+// Retrieves associated voting power and rewards and 
+message TournamentVotesResponse {
+  message Vote {
+    // The incentivized asset receiving this vote.
+    core.asset.v1.AssetId incentivized_asset = 1;
+    // The total amount of voting power allocated to this asset.
+    core.num.v1.Amount vote_power = 2;
+    // The total amount of rewards issued for votes on this asset.
```
erwanor, line 198: "Total amount" makes it sound like it is in relation to the full tournament, when it should be the "amount of voting power that this specific vote contributed"

proto/penumbra/penumbra/view/v1/view.proto
TalDerei: yeah that makes sense, thinking the request should take the `epoch_index` and `address_index`, and response should just be list of all eligible SNRs that _haven't_ been used to vote yet?

proto/penumbra/penumbra/view/v1/view.proto
TalDerei: modified the proto description

proto/penumbra/penumbra/view/v1/view.proto
TalDerei: added `TransactionId` to `TournamentVotesResponse`

proto/penumbra/penumbra/view/v1/view.proto
TalDerei: thinking about this more, what's the utility of returning SNRs to the frontend? why not just store them directly in to the `LqtVotingNotes` IndexedDB table? Our existing logic for querying [voting notes in wasm](https://github.com/penumbra-zone/web/blob/main/packages/wasm/crate/src/storage.rs#L263) + nullifier rpc should be sufficient to filter eligible voting notes and saving them to storage.

erwanor: I don't think we want to make assumptions about shared memory, for example, how would that work if the view server is remote (`pclientd`) or if we are using the rust view server like in pcli.

TalDerei: that's a good point thanks

proto/penumbra/penumbra/view/v1/view.proto
TalDerei: I think we also need the `AddressIndex` here

cronokirby (approved):

PR #5063, opened 2025-02-06
pindexer: add dex block summary

erwanor:
## Describe your changes
WIP pairing

https://github.com/penumbra-zone/dex-explorer/issues/338

### DIFF
--- a/crates/bin/pindexer/src/dex_ex/mod.rs
+++ b/crates/bin/pindexer/src/dex_ex/mod.rs
@@ -21,8 +21,8 @@ use penumbra_sdk_num::Amount;
 use penumbra_sdk_proto::event::EventDomainType;
 use penumbra_sdk_proto::DomainType;
 use penumbra_sdk_sct::event::EventBlockRoot;
+use sqlx::types::BigDecimal;
 use sqlx::{prelude::Type, Row};
-use sqlx::{types::BigDecimal};
 use std::collections::{BTreeMap, HashMap, HashSet};
 
 type DateTime = sqlx::types::chrono::DateTime<sqlx::types::chrono::Utc>;
@@ -1128,7 +1128,6 @@ impl Component {
 
         let mut batch_swap_summaries = Vec::<BatchSwapSummary>::new();
 
-        // Deal with the batch summaries:
         for event in &events.batch_swaps {
             let trading_pair = event.batch_swap_output_data.trading_pair;
 
--- a/crates/bin/pindexer/src/dex_ex/mod.rs
+++ b/crates/bin/pindexer/src/dex_ex/mod.rs
@@ -687,7 +687,10 @@ struct BatchSwapSummary {
 }
 
 // Add these serialization helper functions
-fn serialize_asset_id<S>(asset_id: &penumbra_sdk_asset::asset::Id, serializer: S) -> Result<S::Ok, S::Error>
+fn serialize_asset_id<S>(
+    asset_id: &penumbra_sdk_asset::asset::Id,
+    serializer: S,
+) -> Result<S::Ok, S::Error>
 where
     S: serde::Serializer,
 {
--- a/crates/bin/pindexer/src/dex_ex/mod.rs
+++ b/crates/bin/pindexer/src/dex_ex/mod.rs
@@ -24,6 +24,11 @@ use penumbra_sdk_sct::event::EventBlockRoot;
 use serde::{Deserialize, Serialize};
 use sqlx::types::BigDecimal;
 use sqlx::Row;
+use sqlx::{encode::IsNull, postgres::PgArgumentBuffer, Encode, Postgres};
+
+use sqlx::postgres::PgHasArrayType;
+use sqlx::postgres::PgTypeInfo;
+// use sqlx::types::ArrayType;
 use std::collections::{BTreeMap, HashMap, HashSet};
 
 type DateTime = sqlx::types::chrono::DateTime<sqlx::types::chrono::Utc>;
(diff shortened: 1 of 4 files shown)


crates/bin/pindexer/src/dex_ex/mod.rs
```diff
@@ -668,6 +672,40 @@ struct PairMetrics {
     liquidity_change: f64,
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize)]
+struct BatchSwapSummary {
+    #[serde(serialize_with = "serialize_asset_id")]
+    asset_start: penumbra_sdk_asset::asset::Id,
+    #[serde(serialize_with = "serialize_asset_id")]
+    asset_end: penumbra_sdk_asset::asset::Id,
```
erwanor, line 680: We could just use `Vec<u8>` and sqlx's `BigDecimal`, that way we do not need to marshal the struct to json.

crates/bin/pindexer/src/dex_ex/mod.rs
```diff
@@ -1081,6 +1133,109 @@ impl Component {
...
+            .sum::<usize>() as i32;
+        let num_txs = events.batch_swaps.len() as i32;
+
+        let mut batch_swap_summaries = Vec::<BatchSwapSummary>::new();
+
+        // Deal with the batch summaries:
+        for event in &events.batch_swaps {
+            let trading_pair = event.batch_swap_output_data.trading_pair;
+
+            if let Some(swap_1_2) = &event.swap_execution_1_for_2 {
```
erwanor, line 1160: These two `if let Some(swap) = ...` branches do the same thing, in both cases we want to record a new row so my suggestion would be to refactor to call a single method `record_swap_execution_summary(se: SwapExecution, ...)` which also handles writing to the DB.

conorsch:
@JasonMHasperhoven and I paired a bit to debug the failing smoke test. To investigate, I pulled down this branch, ran `just dev` to spin up the fullnode indexing config, and quickly identified the specific error that pindexer was showing:

pindexer-debug-2

That error, in text, is:

```
Error: error returned from database: column "batch_swaps" is of type batch_swap_summary[] but expression is of type jsonb
Caused by:
    column "batch_swaps" is of type batch_swap_summary[] but expression is of type jsonb
```

Clear enough: we've got to massage the types a bit. Typically we use JSON blobs to store Penumbra types in postgres via pindexer, but as the spec in https://github.com/penumbra-zone/dex-explorer/issues/338  indicates, we don't want to do that here. Instead, we want a full custom type inside the db. I took a stab at implementing `sqlx::Encode` for the `BatchSwapSummary` type, but wasn't able to finish it. The build-time error I'm getting now is:


```
error[E0277]: the trait bound `for<'a> &'a [BatchSwapSummary]: sqlx::Encode<'_, Postgres>` is not satisfied
    --> crates/bin/pindexer/src/dex_ex/mod.rs:1270:16
     |
1270 |         .bind(&batch_swap_summaries)
     |          ----  ^^^^^^^^^^^^^^^^^^^^ the trait `for<'a> sqlx::Encode<'_, Postgres>` is not implemented for `&'a [BatchSwapSummary]`, which is required by `&Vec<BatchSwapSummary>: sqlx::Encode<'_, Postgres>`
```

Subsequent work should pick up from there. Take the Encode impl with a grain of salt, @JasonMHasperhoven , because this code isn't actually working yet! But hopefully this is one step closer.

crates/bin/pindexer/src/dex_ex/mod.rs
```diff
@@ -706,6 +711,45 @@ where
...
+
+        // Encode each field
+        asset_start_bytes.encode_by_ref(buf);
+        asset_end_bytes.encode_by_ref(buf);
+        input_decimal.encode_by_ref(buf);
+        output_decimal.encode_by_ref(buf);
+
+        // Encode the primitive types directly
+        <i32 as sqlx::Encode<'_, Postgres>>::encode(self.num_swaps, buf);
+        <f64 as sqlx::Encode<'_, Postgres>>::encode(self.price_float, buf);
```
erwanor, line 740: Why not build the composite type using sqlx type, we can put whatever we want it in

crates/bin/pindexer/src/dex_ex/schema.sql
```diff
@@ -209,6 +209,48 @@ CREATE INDEX ON dex_ex_batch_swap_traces (time, height);
...
+    -- The price with `asset_end` as the quote asset.
+    price_float DOUBLE PRECISION
+);
+
+-- A summary of block data with a bias for DEX data.
+CREATE TABLE IF NOT EXISTS dex_ex_block_summary (
+    -- Primary key
+    rowid SERIAL PRIMARY KEY,
+    -- The height of the block.
+    height INTEGER NOT NULL,
```
erwanor, line 233: @cronokirby noticed that some inserts fail if we are adding a `UNIQUE` constraint on height, but my understanding is that this part of the pipeline should have exactly once semantics on delivery. is that right?

wondering if this is caused by the catchup logic: https://github.com/penumbra-zone/penumbra/blob/erwan/pindexer_block_summary/crates/util/cometindex/src/indexer/indexing_state.rs#L271 (`<=` vs. `<`). unsure.

erwanor:
AFAICT this is good to go, the integration test seems to timeout because of the commitment source addition.

JasonMHasperhoven (approved):

cronokirby (approved):

JasonMHasperhoven:
How do we fix the integration test here?

conorsch:
> How do we fix the integration test here?

The integration test communicates with the live endpoints for the testnet (#4983). Given the ongoing work towards #5010, the endpoint is no longer compatible with `main`, thus the failing test. We should probably disable it on main branch for now. I'll follow up in a separate PR.

crates/bin/pindexer/src/dex_ex/schema.sql
conorsch: DevOps hiring for RC

https://web3.career/staff-sre-engineer-platform-phantom/95791

Questions:

1. Given a greenfield project that needs hosting, how do you approach the problem? What do you focus on first? What information do you require? (Look for keywords like monitoring, network surface, observability, SLAs; explore from there.)
2. Sometimes on small teams there can be disagreements about architecture. Can you tell me about a time when you disagreed with a technical decision made by a colleague, and how you approached that situation?
3. Describe your preferred working style. Do you prefer to communicate via group chat in Slack or similar, via DMs, on Github issues, on video calls?
4. What types of work, whether backend systems administration or otherwise, do you find draining? What about rewarding?
5. If we were to move forward with this role,how could Radiant Commons support you? What does supportive work environment look like to you?


6. How do you prefer to run software if left to your own devices? Do you like container orchestration? 
7. How do you decide whether a service should be self-run, internally to the company, versus paid for as a Saas?
8. Can you share an "war stories" about running production infrastructure, incident response? 
9. When it comes to writing code vs automating infrastructure, which do you prefer? HOw are those responsibilities different, how related?
10. 








conor in 🌐 Antigonus in ~/src/penumbra on 🌱 main [$] via 🦀 v1.83.0 via ❄  impure (nix-shell-env)
❯ 1


After deploying `pindexer` at `v1.0.2`, the mainnet events pipeline stalled out:

```
2025-02-17T16:59:01.880880Z DEBUG sqlx::query: summary="SELECT total, staked, price …" db.statement="\n\nSELECT\n  total,\n  staked,\n  price\nFROM\n
 insights_supply\nORDER BY\n  HEIGHT DESC\nLIMIT\n  1\n" rows_affected=0 rows_returned=0 elapsed=246.583µs elapsed_secs=0.000246583
Error: error returned from database: duplicate key value violates unique constraint "block_details_pkey"

Caused by:
    duplicate key value violates unique constraint "block_details_pkey"
```

After switching to a build based on #5095, that error becomes:

```
2025-02-17T17:11:08.156215Z DEBUG sqlx::query: summary="INSERT INTO insights_supply(height, total, …" db.statement="\n\nINSERT INTO\n  insights_supply
(\n    height,\n    total,\n    staked,\n    price,\n    price_numeraire_asset_id\n  )\nVALUES\n  ($1, $2, $3, $5, $4) ON CONFLICT (height) DO\nUPDATE
\nSET\n  total = excluded.total,\n  staked = excluded.staked,\n  price = excluded.price,\n  price_numeraire_asset_id = excluded.price_numeraire_asset_
id\n" rows_affected=0 rows_returned=0 elapsed=545.631µs elapsed_secs=0.000545631
Error: error returned from database: type "batch_swap_summary" does not exist

Caused by:
    type "batch_swap_summary" does not exist
```

which looks like a genuine problem. I'm surprised that the `CREATE TYPE batch_swap_summary AS` declaration in the modified schema isn't sufficient.

PR #5065, opened 2025-02-06
lqt: emit events in dex/funding

erwanor:
## Describe your changes

This PR adds events for the LQT tournament

#5066 #5067 

  > LQT branch


proto/penumbra/penumbra/core/component/funding/v1/funding.proto
```diff
@@ -116,10 +117,49 @@ message ActionLiquidityTournamentVoteView {
...
+
+message EventLqtDelegatorReward {
+  // The epoch for which the reward was paid.
+  uint64 epoch_index = 1;
+  // The amount of reward, in staking tokens.
+  num.v1.Amount reward_amount = 2;
+  // The amount of reward, in delegation tokens.
+  asset.v1.Value delegation_tokens = 3;
+  // The recipient of the reward
+  keys.v1.Address address = 4;
```
TalDerei, line 136: just to clarify, appending the address is allowed here because the reward comes from the community pool which mints transparently?

proto/penumbra/penumbra/core/component/funding/v1/funding.proto
```diff
@@ -116,10 +117,49 @@ message ActionLiquidityTournamentVoteView {
...
+  dex.v1.PositionId position_id = 3;
+  // The incentivized asset.
+  asset.v1.AssetId incentivized_asset_id = 4;
+  // The total volume for the pair during the tournament, in staking tokens.
+  num.v1.Amount tournament_volume = 10;
+  // The cumulative volume for the LP, in staking tokens.
+  num.v1.Amount position_volume = 11;
+}
+
+message EventLqtVote {
```
TalDerei, line 156: what about transaction ID? or does this create a link in the events for the vote, associating it with the transaction hash that performed the voting in the [LiquidityTournamentReward](https://github.com/penumbra-zone/penumbra/compare/main...protocol/lqt_branch#diff-a58350bb3062b30c1ccfe47af108dc341e9fab8ef98b28fb1b45862f6595fb05R32-R37)?

cronokirby, line 156: I think we should add the recipient address and the transaction id. Tx IDs would be a good way to track "unique votes" roughly.

proto/penumbra/penumbra/core/component/dex/v1/dex.proto
```diff
@@ -632,6 +632,22 @@ message EventSwap {
   crypto.tct.v1.StateCommitment swap_commitment = 4;
 }
 
+// Tracks the LQT eligible volume for a position.
+// The volume corresponds to the outflow of staking tokens from
+// the position.
+message EventLqtPositionVolume {
```
cronokirby, line 638: Is there a good way to link this to the execution of the dex? Like it could be nice to put in a bsod reference or something which would let us know whether or not it ties in to block execution or arb.

cronokirby (approved):
Looks good, I think we could add a couple other bits of information to the events to ambient context. For the vote in particular this is a simple change. I'm less sure about the dex event though.

proto/penumbra/penumbra/core/component/funding/v1/funding.proto
erwanor: Good idea

proto/penumbra/penumbra/core/component/dex/v1/dex.proto
erwanor: That's a good idea, agreed. I think we will have to switch things around. We can do this on a second pass. One way to do this is to emit the event on `position_execution`. It will take some workshopping to do this without disrupting the method.

Issue #5069, opened 2025-02-06
pcli: `tx lp withdraw <position>`


erwanor:
We can withdraw a position multiple times using `pcli`.

Each time, the sequence number of the position increases.

We query the remote RPC to fetch the latest for the position so that we always compute an accurate balance commitment. This is important to do because the funding component will deposit rewards into that position, and our state might be stale.

PR #5070, opened 2025-02-06
pcli: support multiple lp withdrawals

erwanor:
## Describe your changes

Now we can use `pcli tx withdraw <id>` to withdraw a position multiple times, increasing its sequence. We fetch the reserves on every try so that the balance commitment is always accurate.

To view the output, use `pcli v balance --by-note` this is unfortunate but wontfix now.

#5069 

### DIFF
--- a/crates/view/src/service.rs
+++ b/crates/view/src/service.rs
@@ -747,7 +747,7 @@ impl ViewService for ViewServer {
                     tonic::Status::invalid_argument(format!("Could not parse pair: {e:#}"))
                 })?;
 
-            planner.position_withdraw(position_id, reserves, trading_pair);
+            planner.position_withdraw(position_id, reserves, trading_pair, 0);
         }
 
         // Insert any ICS20 withdrawals.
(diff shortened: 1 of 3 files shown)


crates/view/src/service.rs
```diff
@@ -747,7 +747,7 @@ impl ViewService for ViewServer {
                     tonic::Status::invalid_argument(format!("Could not parse pair: {e:#}"))
                 })?;
 
-            planner.position_withdraw(position_id, reserves, trading_pair);
+            planner.position_withdraw(position_id, reserves, trading_pair, 0);
```
cronokirby, line 750: Why is this being 0 ok? Wouldn't we want to follow the planning request?

erwanor: This is kind of a fake TPR implementation, unlike the one in Prax. It's not used by the regular pcli commands, I think an integration test uses it maybe. Making it work here is possible but low ROI, so I opted for not doing anything

cronokirby: Sounds good ; we could maybe add a comment to that effect

conorsch:
We're not currently exercising the multiple-withdraw behavior in the integration tests, but we'll be exercising these code paths a lot toward LQT support (#5010).

conorsch (approved):

PR #5071, opened 2025-02-07
LQT: fix start position <=> epoch comparison

cronokirby:
## Describe your changes

lqt: fix epoch <=> start position comparison

This allows planners to correctly use the first position in the epoch.
This works because the ZK proof verifies that the start position is strictly less.

c.f. https://github.com/penumbra-zone/penumbra/blob/7869803be0b844f09c8c5efeeaa92384b8a6ba35/crates/core/component/governance/src/delegator_vote/proof.rs#L234

Testing deferred.

### DIFF
--- a/crates/core/component/funding/src/action_handler/liquidity_tournament/mod.rs
+++ b/crates/core/component/funding/src/action_handler/liquidity_tournament/mod.rs
@@ -40,11 +40,18 @@ fn is_valid_denom(denom: &Denom) -> anyhow::Result<()> {
 }
 
 // Check that the start position is early enough to vote in the current epoch.
-async fn start_position_good_for_epoch(epoch: Epoch, start: Position) -> anyhow::Result<()> {
-    anyhow::ensure!(
-        epoch.index > u64::from(start.epoch()),
-        "position {start:?} is not before epoch {epoch:?}"
-    );
+async fn start_position_good_for_epoch(current: Epoch, start: Position) -> anyhow::Result<()> {
+    let ok = match (u64::from(start.epoch()), start.block(), start.commitment()) {
+        // Anything before now is definitely fine.
+        (e, _, _) if e < current.index => true,
+        // The ZK proof will check that note being spent is *strictly* earlier than this position,
+        // so we can allow the position to be the very first one in this epoch, and this will not
+        // allow anything in this epoch to be used *even* if the first created note this epoch
+        // is a delegated token.
+        (e, 0, 0) if e == current.index => true,
+        _ => false,
+    };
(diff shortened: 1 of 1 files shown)


cronokirby:
Maybe we should just restrict the check to be for (e, 0, 0) specifically, since that's the only thing I expect planners to produce

crates/core/component/funding/src/action_handler/liquidity_tournament/mod.rs
```diff
@@ -40,11 +40,18 @@ fn is_valid_denom(denom: &Denom) -> anyhow::Result<()> {
...
-    );
+async fn start_position_good_for_epoch(current: Epoch, start: Position) -> anyhow::Result<()> {
+    let ok = match (u64::from(start.epoch()), start.block(), start.commitment()) {
+        // Anything before now is definitely fine.
+        (e, _, _) if e < current.index => true,
+        // The ZK proof will check that note being spent is *strictly* earlier than this position,
+        // so we can allow the position to be the very first one in this epoch, and this will not
+        // allow anything in this epoch to be used *even* if the first created note this epoch
+        // is a delegated token.
+        (e, 0, 0) if e == current.index => true,
```
erwanor, line 51: Good catch, I wonder if this would make sense, either way can revisit, what matters is that the rule is correct

```rust

fn check_older_than_current_epoch(position: Position, current_epoch: &Epoch) -> anyhow::Result<()> {
    let e = u16::try_from(current_epoch.index)
        .map_err(|e| anyhow!("epoch ({current_epoch: ?} must fit into a u16: {e:?}"))?;

    // The precise epoch boundary for the current epoch
    let cutoff = Position::from((e, 0u16, 0u16));

    ensure!(
        position < cutoff,
        "Note position {position:?} is not older than cutoff {cutoff:?}",
    );
    Ok(())
}
```

erwanor: Maybe this is worse actually IDK. either way I feel good this is correct.

erwanor (approved):

crates/core/component/funding/src/action_handler/liquidity_tournament/mod.rs
cronokirby: This would actually be incorrect, because the circuit already checks that this position, which is not the position of the note, is *strictly greater* than the position of the note. Because of the double strict comparison, we need to instead make that a `<=` in your suggestion. (It's really annoying and tricky)

erwanor: oops right yup, was focusing on making the predicate take `Position`s

PR #5073, opened 2025-02-07
chore: bump app version 9 -> 10

conorsch:
## Describe your changes
Bumps the protocol version to denote the addition of LQT functionality. Also bumps the crate versions from 1.0.x -> 2.0.x, as an alpha. No tag has been created, to avoid triggering the formal release workflows, since we're still testing.

#5010.

## Testing and review

I'm submitting this PR ahead of a planned testnet upgrade, so we'll shortly learn how effective the overall upgrade process is. However, special attention should be given to the proposed versions strings. In particular, the transition from `0.81.x` to `1.0.x` maintained compatibility with APP_VERSION 9, and that may present difficulties during the upgrade for nodes on one or the other version.

  > explicitly denotes consensus-breaking changes, and therefore targets a protocol feature branch, rather than main

### DIFF
--- a/COMPATIBILITY.md
+++ b/COMPATIBILITY.md
@@ -10,3 +10,5 @@
 | 7 (Mainnet)               | v0.79.x                | v0.37.x  |   v1     |
 | 8 (Mainnet)               | v0.80.x                | v0.37.x  |   v1     |
 | 9 (Mainnet)               | v0.81.x                | v0.37.x  |   v1     |
+| 9 (Mainnet)               | v1.0.x                 | v0.37.x  |   v1     |
+| 10 (Mainnet)              | v2.0.x                 | v0.37.x  |   v1     |
--- a/crates/core/app/src/app_version/component.rs
+++ b/crates/core/app/src/app_version/component.rs
@@ -16,7 +16,8 @@ fn version_to_software_version(version: u64) -> &'static str {
         6 => "v0.77.x",
         7 => "v0.79.x",
         8 => "v0.80.x",
-        9 => "v0.81.x",
+        9 => "v1.0.x",
+        10 => "v2.0.x",
         _ => "unknown",
     }
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1319,7 +1319,7 @@ dependencies = [
 
 [[package]]
 name = "cnidarium-component"
-version = "1.0.1"
+version = "2.0.0-alpha.0"
 dependencies = [
  "anyhow",
  "async-trait",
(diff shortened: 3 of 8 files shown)


cronokirby (approved):

COMPATIBILITY.md
```diff
@@ -10,3 +10,5 @@
 | 7 (Mainnet)               | v0.79.x                | v0.37.x  |   v1     |
 | 8 (Mainnet)               | v0.80.x                | v0.37.x  |   v1     |
 | 9 (Mainnet)               | v0.81.x                | v0.37.x  |   v1     |
+| 9 (Mainnet)               | v1.0.x                 | v0.37.x  |   v1     |
+| 10 (Mainnet)              | v1.1.x                 | v0.37.x  |   v1     |
```
cronokirby, line 14: Wait, if 10 is breaking compared to 9, shouldn't it actually be version 2.0.x?

conorsch: Yes, we should be referring to the LQT changes at v2. I've updated the PR accordingly.

conorsch: To recap prior discussion, however, we're bumping the crate versions because of breaking changes in the Rust APIs, not because of the protocol version change, which is tracked separately. The compatibility doc was created to disambiguate here, but we should still write down a full explanation of versioning scheme and stick those in the dev docs.

crates/core/app/src/app_version/component.rs
```diff
@@ -16,7 +16,8 @@ fn version_to_software_version(version: u64) -> &'static str {
         6 => "v0.77.x",
         7 => "v0.79.x",
         8 => "v0.80.x",
-        9 => "v0.81.x",
+        9 => "v1.0.x",
+        10 => "v2.0.x",
```
conorsch, line 20: > the transition from `0.81.x` to `1.0.x` maintained compatibility with APP_VERSION 9, and that may present difficulties during the upgrade for nodes on one or the other version.

The clobbering of the `9` match is what I expect will introduce problems if used as-is on mainnet, where different versions are surely running. Must we extend the matching logic to accept multiple-choice? I don't see a way around that, unless we mandate upgrading 0.81.x -> 1.0.x -> 2.0.x.

erwanor: I thought UIP-6 only checked the `APP_VERSION`, can take a look

conorsch: On closer inspection, I agree: the crate version string is only ever used in user-facing messaging, so we should be good to go here.

PR #5074, opened 2025-02-07
pd: add no-op migration for APP_VERSION 10

conorsch:
## Describe your changes

Follow-up to #5073, which should have included a no-op migration.

Towards #5010. 

## Testing and review

This work will  be used to perform a chain upgrade on the PL testnet. Once that's done, we should expect CI to pass fully on this PR, in particular the "testnet-integration" suite.

  > provides migration logic for proposed protocol-breaking changes

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1319,7 +1319,7 @@ dependencies = [
 
 [[package]]
 name = "cnidarium-component"
-version = "2.0.0-alpha.0"
+version = "2.0.0-alpha.1"
 dependencies = [
  "anyhow",
  "async-trait",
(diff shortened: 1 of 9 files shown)


conorsch:
The no-op migration is necessary but not yet sufficient: I am indeed encountering the version mismatch logic anticipated in https://github.com/penumbra-zone/penumbra/pull/5073#discussion_r1946839568. Working with a pre-upgrade backup of node state to test changes, will append a patch here and mark RFR when done.

conorsch:
> The no-op migration is necessary but not yet sufficient

In the first draft of this PR, I'd neglected to update the default migration used by `pd migrate` to be the new `mainnet3` no-op migration. That's done now, and this patch was used to finalize the upgrade on the PL testnet. The testnet has resumed, as evidenced by the testnet-integration CI tests passing.

cronokirby (approved):

PR #5079, opened 2025-02-08
lqt: implement voting in pcli

cronokirby:
Stacked pr on #5077, please review and merge that first.

This implements all of the bits and bobs necessary to make `pcli tx lqt-vote` work.  I ran a devnet and tested that you can submit votes. The events for voting do get emitted, but there seems to be some issue with either the code or the setup I had in terms of rewards actually being distributed.

To test this pr, spinning up a devnet, or hitting a testnet post #5077 patch would work.

  > Not consensus breaking actually, relative to the protocol branch, or at least, it shouldn't be, since this only touches the client side of things.


erwanor (approved):
Thanks for your work on this, merging on green CI

PR #5080, opened 2025-02-09
tpr: extend to lqt actions

TalDerei:
## Describe your changes

extends `TransactionPlannerRequest` to support `ActionLiquidityTournamentVote` variant, which is necessary for wasm testing. 

  > LQT branch


proto/penumbra/penumbra/view/v1/view.proto
```diff
@@ -441,6 +443,19 @@ message TransactionPlannerRequest {
...
     repeated core.component.stake.v1.RateData rate_data = 5;
   }
+
+  message ActionLiquidityTournamentVote {
+    // The asset the user wants to vote for.
+    repeated core.asset.v1.Denom incentivized = 1;
+    // The address the user wants potential rewards to go to.
+    core.keys.v1.Address rewards_recipient = 2;
+    // The note containing the staked note used for voting.
+    SpendableNoteRecord staked_note = 3;
```
TalDerei, line 453: probably only need `Note` here, but taking an SNR object seems fine.

cronokirby: If we take a SpendableNoteRecord then we shouldn't take the position of the note as well, since the SNR contains the positiob

proto/penumbra/penumbra/view/v1/view.proto
```diff
@@ -441,6 +442,19 @@ message TransactionPlannerRequest {
...
+    // The position of the staked note.
+    uint64 staked_note_position = 4;
+    // The start position of the tournament.
+    uint64 start_position = 5;
```
cronokirby, line 456: I think we should take an epoch index here (same type, different field name) so that we match the plan. Also, any position should always be (epoch, 0, 0) so it might as well be the epoch.

proto/penumbra/penumbra/view/v1/view.proto
TalDerei: good catch, kept `SpendableNoteRecord` and removed note position

proto/penumbra/penumbra/view/v1/view.proto
TalDerei: modified field name to `epoch_index`

proto/penumbra/penumbra/view/v1/view.proto
TalDerei: should the request actually accept a `Vec< SpendableNoteRecord>`? what if the user's voting weight exceeds a single note and spans multiple notes?

cronokirby: yeah I think having a repeated here makes sense

TalDerei: modified https://github.com/penumbra-zone/penumbra/pull/5080/commits/13e82f25e92ce8f3fa63df2a91c32221f151be16

cronokirby (approved):

PR #5083, opened 2025-02-10
Add some more tracing to LQT code paths

cronokirby:
## Describe your changes

I found these debug traces very useful when messing around with stuff, and I think the LQT code paths were a bit light on tracing before.

Testing shouldn't be necessary, just adds tracing.

  > Just adds more logging


TalDerei (approved):

conorsch (approved):
Thank you :pray:

PR #5084, opened 2025-02-11
pcli: fix edge-case in undelegate-claim

erwanor:
## Describe your changes

This fixes an edge-case in `pcli tx undelegate-claim` that would compute an invalid penalty range for dated delegation tokens.

Instead of using the current epoch as the end of the penalty range, we compute a more precise bound using the block delay chain parameter. This fixes processing old delegation tokens bound to a validator that transitioned in and out of penalty states.

Later, we could expose an RPC method to simplify client-side logic. 

  > pcli


cronokirby (approved):

erwanor:
User confirmed this fixed their issue

PR #5085, opened 2025-02-12
LQT: adjust budget every block

cronokirby:
## Describe your changes

This adjusts the issuance budget every block, incrementing it by the issuance per block, every block, rather than clobbering the current value with `issuance_per_block * how_long_the_epoch_was` at the end of an epoch. This has the effects of:

- making the value always match that claimed by the events, simplifying indexing,
- seamlessly rolling over un-issued rewards to the next epoch,
- simplifying the code.

I also pushed a small change to use a proper Event struct.

Testing deferred.

  > This breaks consensus, but I think this can be applied to the testnet just by pushing the change to the majority validator, which will then issue the incorrect reward amount for exactly one epoch. This has no bearing on any eventual mainnet upgrade. We could avoid this by using a second state key, but putting in this scar tissue seems more performative than actually valuable, given that the upgrade isn't in a final state yet.

### DIFF
--- a/crates/core/component/distributions/src/component.rs
+++ b/crates/core/component/distributions/src/component.rs
@@ -14,7 +14,7 @@ use cnidarium::StateWrite;
 use cnidarium_component::Component;
 use penumbra_sdk_num::Amount;
 use penumbra_sdk_proto::StateWriteProto;
-use penumbra_sdk_sct::{component::clock::EpochRead, epoch::Epoch};
+use penumbra_sdk_sct::component::clock::EpochRead;
 use tendermint::v0_37::abci;
 use tracing::instrument;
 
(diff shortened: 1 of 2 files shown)


crates/core/component/distributions/src/component.rs
```diff
@@ -52,28 +52,26 @@ impl Component for Distributions {
...
-            .expect("block height should not be negative");
+        let epoch_index = current_epoch.index;
 
-        let new_issuance = state
+        let increase: Amount = state
             .get_distributions_params()
             .await
             .expect("distribution parameters should be available")
-            .staking_issuance_per_block as u128;
+            .liquidity_tournament_incentive_per_block
```
erwanor, line 61: 👍

erwanor (approved):

PR #5091, opened 2025-02-14
lqt: make transaction planner request take only a single denom

cronokirby:
## Describe your changes

Looking at the ongoing integration work in: https://github.com/penumbra-zone/web/pull/2041, the transaction planner request for lqt votes can only make use of a single incentivized denom anyways, which is what we want to support in the final ux, so having a repeated field here that subsequently gets truncated to the first provided value is superfluous. This change makes the field plain instead.

This can be tested inside of that PR.

  > Affects clients only.


TalDerei (approved):
thanks for noticing this, the use of `repeated` was an artifact of some misunderstanding, and the wasm planner already accounts for this by basically iterating over each delegation note and using the same incentivized asset that was passed in the request for each action it constructs.

erwanor (approved):

PR #5095, opened 2025-02-15
Some cometindex robustness improvements

cronokirby:
## Describe your changes

This tweaks the indexing logic in cometindex to add a bit of extra robustness against indexing the same block twice. The current logic is to generate batches of events, which are then sent over to logical threads for each app view, the idea being to only read each batch once, avoiding duplicate database fetching work from each thread, while allowing parallel processing. One potential race condition which might have been the cause of some oddities we've seen around duplicate processing is that each thread would process the entire batch as long as the next height it needed to index was somewhere inside of that batch. In practice, if the threads were in sync, this would be fine, but if for whatever reason they got desynced (e.g. one app view crashes because there's a bug or some weird data thing, then we patch the bug and restart pindexer, or cometindex dies when some indices have committed a batch, but not others, which can totally happen because some indices are faster than other), it's possible that this might lead to some threads processing some events twice. This adds some logic to truncate the events in each thread so that only the blocks that particular thread needs are indexed.

I also did a pass over on the rest of the indexing logic, and added a comment on a particularly tricky section, justifying its correctness.

CI should be sufficient to test this. We shouldn't observe any difference in behavior.

  > indexing only

### DIFF
--- a/crates/bin/pindexer/src/dex_ex/mod.rs
+++ b/crates/bin/pindexer/src/dex_ex/mod.rs
@@ -1425,7 +1425,7 @@ impl AppView for Component {
         let mut charts = HashMap::new();
         let mut snapshots = HashMap::new();
         let mut last_time = None;
-        for block in batch.by_height.iter() {
+        for block in batch.events_by_block() {
             let mut events = Events::extract(&block)?;
             let time = events
                 .time
--- a/crates/util/cometindex/src/index.rs
+++ b/crates/util/cometindex/src/index.rs
@@ -19,17 +19,60 @@ pub struct BlockEvents {
 
 #[derive(Clone, Debug)]
 pub struct EventBatch {
-    pub first_height: u64,
-    pub last_height: u64,
+    first_height: u64,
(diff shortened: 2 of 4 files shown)


cronokirby:
I also found a bug in the previous code causing events to be marked as being in the wrong block

conorsch:
Failing CI was due to testnet protocol mismatch; resolved by https://github.com/penumbra-zone/penumbra/pull/5093, so this must be rebased for CI to pass. Doing so now.

conorsch:
>  In practice, if the threads were in sync, this would be fine, but if for whatever reason they got desynced (e.g. one app view crashes because there's a bug or some weird data thing, then we patch the bug and restart pindexer, or cometindex dies when some indices have committed a batch, but not others, which can totally happen because some indices are faster than other),

Not only am I certain I've encountered this bug before, but I even saw it again this morning when updating the mainnet indexing pipeline. Big improvement here, thanks for taking the time!

conorsch (approved):

PR #5081, opened 2025-02-10
Record transactions into pindexer

JasonMHasperhoven:
## Checklist before requesting a review

- [ ] I have added guiding text to explain how a reviewer should test these changes.

- [ ] If this code contains consensus-breaking changes, I have added the "consensus-breaking" label. Otherwise, I declare my belief that there are not consensus-breaking changes, for the following reason:

  > REPLACE THIS TEXT WITH RATIONALE (CAN BE BRIEF)

PR #5097, opened 2025-02-18
cometindex: make transactions available in event batches

cronokirby:
The raw ABCI event database already has all the transaction data, and this PR amends our processing pipeline in cometindex to present blocks containing both a list of events, each of which may directly have all of this transaction data (along with a transaction hash), and a list of all the transactions themselves.

This also refactors things to be a bit more opaque, giving us more control over the internal representation of the batch of events.

The `ContextualizedEvent` struct is probably a little over-engineered, but avoids needlessly copying the transaction data around. Ultimately the performance of indexing is tied to data throughput, so we should be mindful of needless copying in the architecture.

This should help with the draft in #5081, avoiding the need for creating a bespoke event.

I tested this by setting up a little indexer just parsing out the tx data and serializing the JSON, and 

  > Indexing code changes only


erwanor (approved):
Nice

PR #5100, opened 2025-02-20
Store batch swap summaries as json

JasonMHasperhoven:
## Describe your changes

As discussed in the slack thread, parsing a custom type in typescript turns out to require a lot of hoops. This converts the batch_swap_summaries to a json so that it can be parsed easily.

Part of https://github.com/penumbra-zone/dex-explorer/issues/338

### DIFF
--- a/crates/bin/pindexer/Cargo.toml
+++ b/crates/bin/pindexer/Cargo.toml
@@ -45,7 +45,6 @@ tokio = {workspace = true, features = ["full"]}
 serde_json = {workspace = true}
 serde = { workspace = true, features = ["derive"] }
 sqlx = { workspace = true, features = ["chrono", "postgres"] }
-hex = "0.4"
 
 [dev-dependencies]
 url = {workspace = true}
--- a/crates/bin/pindexer/Cargo.toml
+++ b/crates/bin/pindexer/Cargo.toml
@@ -45,6 +45,7 @@ tokio = {workspace = true, features = ["full"]}
 serde_json = {workspace = true}
 serde = { workspace = true, features = ["derive"] }
 sqlx = { workspace = true, features = ["chrono", "postgres"] }
+hex = "0.4"
 
 [dev-dependencies]
 url = {workspace = true}
--- a/crates/bin/pindexer/src/dex_ex/mod.rs
+++ b/crates/bin/pindexer/src/dex_ex/mod.rs
@@ -677,22 +677,12 @@ struct PairMetrics {
 #[derive(Debug, Clone, Type, serde::Serialize)]
 #[sqlx(type_name = "batch_swap_summary")]
 struct BatchSwapSummary {
-    #[serde(serialize_with = "serialize_bytes")]
-    asset_start: Vec<u8>,
-    #[serde(serialize_with = "serialize_bytes")]
-    asset_end: Vec<u8>,
-    input: BigDecimal,
-    output: BigDecimal,
+    asset_start: String,
+    asset_end: String,
+    input: String,
+    output: String,
     num_swaps: i32,
     price_float: f64,
-}
-
-// Add this helper function to serialize byte vectors as hex strings
-fn serialize_bytes<S>(bytes: &Vec<u8>, serializer: S) -> Result<S::Ok, S::Error>
-where
-    S: serde::Serializer,
-{
-    serializer.serialize_str(&hex::encode(bytes))
 }
 
 #[derive(Debug)]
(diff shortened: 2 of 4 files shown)


conorsch:
Rebased to include recent merge of #5081. Might rebase once more, when #5101 lands, simply to ease my own testing here.

crates/bin/pindexer/Cargo.toml
```diff
@@ -45,6 +45,7 @@ tokio = {workspace = true, features = ["full"]}
 serde_json = {workspace = true}
 serde = { workspace = true, features = ["derive"] }
 sqlx = { workspace = true, features = ["chrono", "postgres"] }
+hex = "0.4"
```
conorsch, line 48: Nit: the `hex` dependency, if required, should be formatted as `hex = { workspace = true }`, so that there's only one version of `hex` throughout the repo to manage. However, I'm not convinced the dep addition is really necessary....

crates/bin/pindexer/src/dex_ex/mod.rs
```diff
@@ -673,13 +674,13 @@ struct PairMetrics {
...
-#[derive(Debug, Clone, Type)]
+#[derive(Debug, Clone, Type, serde::Serialize)]
 #[sqlx(type_name = "batch_swap_summary")]
 struct BatchSwapSummary {
-    asset_start: Vec<u8>,
-    asset_end: Vec<u8>,
-    input: BigDecimal,
-    output: BigDecimal,
+    asset_start: String,
+    asset_end: String,
```
conorsch, line 681: Encoding the assets as Strings discards meaningful typing: we should instead store `asset_start` and `asset_end` as `asset::Id`s, same as elsewhere in the pindexer dex code.

crates/bin/pindexer/src/dex_ex/mod.rs
```diff
@@ -673,13 +674,13 @@ struct PairMetrics {
...
+    input: String,
+    output: String,
```
conorsch, line 683: Similarly, casting to String for `input` and `output` discards meaningful typing. I think a better choice would be `Amount`, as done elsewhere in the code.

crates/bin/pindexer/src/dex_ex/mod.rs
```diff
@@ -1150,10 +1151,10 @@ impl Component {
                 let num_swaps = filtered_swaps.len() as i32;
 
                 batch_swap_summaries.push(BatchSwapSummary {
-                    asset_start: asset_start.to_bytes().to_vec(),
-                    asset_end: asset_end.to_bytes().to_vec(),
-                    input: BigDecimal::from(input.value()),
-                    output: BigDecimal::from(output.value()),
+                    asset_start: hex::encode(asset_start.to_bytes()),
+                    asset_end: hex::encode(asset_end.to_bytes()),
```
conorsch, line 1155: If you accept the suggestion to use domain types rather than Strings, then we no longer need the `hex::encode` call. If necessary, we can call `.encode_to_vec()` on domain types, but there should be sufficient serialization handlers already defined, so that you can just put the direct values in here: `asset_start: asset_start`, or just `asset_start,`.

conorsch (changes requested):
Left some suggestions based on visual review. I'll make sure to test this with a full reindex prior to merging.

JasonMHasperhoven:
From what I understand, the problem is that asset::Id needs to implement the sqlx::Encode trait to be used directly in SQL queries with SQLx, but it doesn't. And by looking at the code, the pattern used elsewhere is to convert the asset::Id to bytes. 

Please check if I'm doing this right @conorsch. I'm still very new to rust and these types as well. It's also new to me that we can store bytes into a json.

conorsch:
> From what I understand, the problem is that asset::Id needs to implement the sqlx::Encode trait to be used directly in SQL queries with SQLx

Frustrating! Doubly so because the requirement to implement `sqlx::Encode` is not actually necessary: that requirement is simply a holder from the previous implementation in https://github.com/penumbra-zone/penumbra/pull/5063/files#diff-b3385b96169ccb7d85884860d22206af9f9ce991bb889ad76b38a700e4ee3a10R675

Once we remove that annotation, and the related Type derivation, then we can smoothly use strong types like `asset::Id` and `Amount` directly within the `BatchSwapSummary` struct definition. 

> the pattern used elsewhere is to convert the asset::Id to bytes

In certain cases, yes, particularly when we `.bind()` values via the sqlx ORM handles. But for the present case, the `        .bind(serde_json::to_value(&batch_swap_summaries)?)` is sufficient: serde will handle converting the list of BatchSwapSummaries into a json representation, which satisfies the type the db is expecting.

conorsch (approved):

PR #5101, opened 2025-02-20
feat(pindexer): optional exit-on-catchup

conorsch:
## Describe your changes
For running batch jobs managing indexing pipelines, it'd be convenient to instruct pindexer to exit after it's finished processing the events that's available to it. This situation arises frequently in batch processing, where pindexer will be pointed at a source db that was restored from a db. Having pindexer exit 0 on success, else non-zero on error, will aid in building out test infrastructure.

N/A

  > only affects runtime options for `pindexer`, no changes to protocol or consensus logic

### DIFF
--- a/crates/util/cometindex/src/indexer.rs
+++ b/crates/util/cometindex/src/indexer.rs
@@ -6,29 +6,26 @@ use indexing_state::{Height, IndexingState};
 use std::sync::Arc;
 use tokio::{sync::mpsc, task::JoinSet};
 
+/// Attempt to catch up to the latest indexed block.
+///
+/// Returns whether or not we've caught up.
 #[tracing::instrument(skip_all)]
 async fn catchup(
     state: &IndexingState,
     indices: &[Arc<dyn AppView>],
     genesis: Arc<serde_json::Value>,
-    exit_on_catchup: bool,
-) -> anyhow::Result<()> {
+) -> anyhow::Result<bool> {
     if indices.len() <= 0 {
         tracing::info!(why = "no indices", "catchup completed");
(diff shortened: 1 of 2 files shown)


crates/util/cometindex/src/indexer.rs
```diff
@@ -21,8 +22,13 @@ async fn catchup(
     tracing::info!(?src_height, ?index_heights, "catchup status");
     let lowest_index_height = index_heights.values().copied().min().unwrap_or_default();
     if lowest_index_height >= src_height {
-        tracing::info!(why = "already caught up", "catchup completed");
-        return Ok(());
+        if exit_on_catchup {
+            tracing::info!("catchup completed, exiting as requested");
+            std::process::exit(0);
```
conorsch, line 27: It's a bit of a hack to `std::process::exit()` here. Perhaps we should instead return a custom Result::CatchUp that triggers an exit in the main loop? Didn't seem worthwhile for a quick feature add, but open to suggestions.

cronokirby: I wrote up a quick commit to avoid std::process::exit: https://github.com/penumbra-zone/penumbra/commit/da002eebf061f0d9f7f392ef3b13a08f85c802bd ; I think I'd like to see that included in this PR (or something equivalent). `Result<bool>`  seems fine enough to capture the possibilities there, and avoids the woes of direct exits, in case we want to run this code in a different context.

conorsch:
Thanks @cronokirby, that change is just what I was looking for!

cronokirby (approved):

PR #5102, opened 2025-02-21
feat: add summary to TransactionInfo

VanishMax:
Adds a `summary` field of type `TransactionSummary` to `TransactionInfo` proto.

Mostly needed for Web because Rust ecosystem can simply do `tx_info.view.summary()` but Web would have to re-implement this feature in TypeScript


cronokirby (approved):
This makes sense ; putting this in the view service avoids the need to duplicate the code, since we can re-use the Rust method in the WASM underlying the view service.

PR #5109, opened 2025-02-27
backport(v0.79.x): app: create event for when app parameters change

cronokirby:
## Describe your changes

Adds an event which fires whenever the app parameters change, to easily allow indexing components to track the current parameters as needed.

  > Adds a new event.


conorsch (changes requested):
There's a non-critical CI failure related to docs-lint, which I'd like to clean up. Crucially, all the rust checks and tests are passing, which is good. 

However, this PR is currently targeting `release/v0.79.x` which is the wrong branch: it should instead target `release/v0.79.x-contd`. That is confusing and incredibly hard to know. I'll add some historical info to the dev docs so that future maintainers have a point of reference when performing back port changes like this.

conorsch (approved):

PR #5113, opened 2025-02-28
LQT: Indexing V2

cronokirby:
## Describe your changes

This updates the lqt indexing to provide what should be more or less a backing of the UI we want to end up with.

Compared to other indexers, this innovates by:
- using a schema to have a nice namespace, rather than an ad-hoc table naming structure,
- trying to lean more on SQL instead of Rust.

I've found the latter to be quite beneficial compared to the dex work. It does require some more prowess in SQL, but it is at least 50% less tedious to update a view than it is to update the Rust code. One big advantage is that you don't have to write out all the boilerplate needed to read the backing tables into Rust, which saves a lot of code.

## Testing

Unfortunately, this won't work well on the testnet, because we don't have genesis params for the LQT.

On a devnet I ran this with a few interactions, getting the following results:
image
image
image

These all seem coherent at a basic level to me.

  > Indexing only


crates/bin/pindexer/src/lqt/schema.sql
```diff
@@ -1,32 +1,179 @@
...
+    power NUMERIC NOT NULL,
+    asset_id BYTEA NOT NULL,
+    address BYTEA NOT NULL
 );
 
-CREATE INDEX ON lqt_position_rewards (epoch);
+CREATE INDEX ON lqt._votes (epoch);
+CREATE INDEX ON lqt._votes (address);
+
+CREATE VIEW lqt.summary AS
```
erwanor, line 54: I think it would be helpful to have natural language summaries for each materialized views. The query is self-explanatory but the higher level purpose of the view, e.g, for the summary of lqt votes we need:
- to display how many reward were issued in total for each epoch
- to display the staking power mobilized for the epoch
- etc.

crates/bin/pindexer/src/lqt/schema.sql
```diff
@@ -1,32 +1,179 @@
...
+CREATE TABLE lqt._finished_epochs (
+    epoch INTEGER PRIMARY KEY
+);
+
+CREATE TABLE lqt._available_rewards (
+  epoch INTEGER PRIMARY KEY,  
+  amount NUMERIC NOT NULL
+);
+
+CREATE TABLE lqt._delegator_rewards (
```
erwanor, line 18: One latent question, why are we prefixing the tables with `_`, guess is that they are internals? but if they are used by the view, then they should probably not be right?

crates/core/component/dex/src/component/position_manager/volume_tracker.rs
```diff
@@ -41,24 +41,24 @@ pub(crate) trait PositionVolumeTracker: StateWrite {
...
-            .expect("the staking token is in the pair");
-
-        let new_balance = new_state
-            .reserves_for(*STAKING_TOKEN_ASSET_ID)
-            .expect("the staking token is in the pair");
-
         // We track the *outflow* of the staking token.
         // "How much inventory has left the position?"
-        let staking_token_outflow = prev_balance.saturating_sub(&new_balance);
+        let staking_token_outflow = flows.lambda_1;
```
erwanor, line 61: This is really confusing for me. I think it works because we flip the pair, so in the suspended `Flows` we have made sure that $\Lambda_1$ resolves to the staking token (unsure). But this seems more dangerous than just doing a dumb lookup and saturating sub. And I am still unclear on why we need to do that at all?

I think we should just not track the inflow, keep it tight here, and instead brainstorm a way to uniquely reference position executions so that we can refer to them in various events via some unique identifier.

crates/core/component/dex/src/lp/position.rs
```diff
@@ -214,6 +214,61 @@ impl Position {
...
     }
+
+    /// Compute the flows compared to a previous position
+    pub fn flows(&self, prev: &Self) -> Flows {
+        Flows::from_fee_and_reserves(self.phi.component.fee, &self.reserves, &prev.reserves)
+    }
+}
+
+#[derive(Clone, Copy, Debug)]
+pub struct Flows {
```
erwanor, line 225: I think this struct is a good idea, but fwiw its current shape seems informed by the previous implementation, but I think there's an opportunity to do something better here. The position execution recording in dex_ex has a context start/end, this makes it slightly confusing to use because the delta/lambda can be of totally different asset types than start/end (e.g, multi-hop path).

Instead, it would be useful to be able to address each position execution uniquely, and index these identifiers by context start/end, or undirected trading pair, or LQT eligible activity etc. 

Since this is already a useful refactor we can keep it like this, not blocking.

erwanor (changes requested):
ACK, all the important indexing looks right and good, looking forward to test it.

The change request is for:

- the volume tracker in the component code. I think we should keep it as it is, and not try to track the staking token inflow.

- anchoring the views with comments

crates/bin/pindexer/src/lqt/schema.sql
cronokirby: They are internals from the perspective of consumers of the AppView. Think of the AppView like a class / module, with the private members being the `_` tables, and the public members being the views. The public members can be constructed by reading the internal members, but outside consumers should only consume the interface via the public views.

cronokirby:
Re: volume tracker component. We could indeed not touch the original logic wrt the state update. We do need to still have the new logic wrt tracking the flow and flipping it, because the event needs to know the other bits of information, like where fees are flowing, how much volume came *in* etc. I think it makes sense to calculate the flows once once, then direct different parts of that struct to the bits of the event, and to update the state.

I think maybe a synthesis of the two approaches would be to add methods to the new `Flows` struct to reorient itself based on a desired asset, making that one first, if possible?

cronokirby:
c.f. https://github.com/penumbra-zone/penumbra/commit/8f2de6fecb42d8aec1121d83742d2b453bf6f46b for a proposed change to that effect @erwanor.

crates/core/component/dex/src/component/position_manager/volume_tracker.rs
cronokirby: I think having a better means of inspecting position executions for event emission would be great ; we do have an upstream requirement of needing this bit of info though, so I think for now we should keep it.

crates/bin/pindexer/src/lqt/schema.sql
cronokirby: I added in-schema comments, which can be queried from psql.

crates/core/component/dex/src/lp/position.rs
```diff
@@ -214,6 +214,61 @@ impl Position {
...
+            lambda_1,
+            lambda_2,
+            fee_1,
+            fee_2,
+        }
+    }
+
+    pub fn flip(&self) -> Self {
+        Self {
+            delta_1: self.delta_2,
```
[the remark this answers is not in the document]
cronokirby: Now Flows is adjusted to be relative to a *directed* trading pair, to avoid this issue.

erwanor (approved):
ACK, excited to test this, great work

PR #5123, opened 2025-03-08
pindexer: add context to indexing calls, to allow deferring work to after catchup

cronokirby:
## Describe your changes

Right now, every indexer gets called with a batch, with no good way of knowing whether or not more batches will follow. This creates unnecessary work. For example, the the dex explorer component, some tables exist only to provide a *current* summary of information. These only need to be computed once we've caught up to the tip of the chain. Any time we compute them before is wasted.

On the first 1_000_000 blocks, avoiding this summary computation saves 9 minutes, going from 21 minutes to 12.
The component (and others) could probably be re-architected in other ways to take advantage of this.

To test, run pindexer again, check that the results are the same, just computed a bit faster.

### DIFF
--- a/crates/bin/pindexer/src/block.rs
+++ b/crates/bin/pindexer/src/block.rs
@@ -1,4 +1,8 @@
-use cometindex::{async_trait, index::EventBatch, sqlx, AppView, PgTransaction};
+use cometindex::{
+    async_trait,
+    index::{EventBatch, EventBatchContext},
+    sqlx, AppView, PgTransaction,
+};
 use penumbra_sdk_proto::{core::component::sct::v1 as pb, event::ProtoEvent};
 use sqlx::types::chrono::DateTime;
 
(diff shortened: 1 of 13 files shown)


conorsch:
> On the first 1_000_000 blocks, avoiding this summary computation saves 9 minutes, going from 21 minutes to 12.

While significant, this speedup sounds modest to me, particularly because tx density will be higher later in the chain. Made sure to test. On tag `v1.2.0`:

```
2025-03-13T17:52:03.228664Z DEBUG run_pindexer{penumbra_environment=Mainnet}: pindexer_db_wrangler::pindexer: pindexer run finished duration=2h 24m 40s 843ms 45us 779ns per_block=2ms 193us 91ns
```

On feature branch for PR 5123:

```
2025-03-13T19:43:09.013256Z DEBUG run_pindexer{penumbra_environment=Mainnet}: pindexer_db_wrangler::pindexer: pindexer run finished duration=44m 29s 257ms 991us 733ns per_block=673us 948ns
```

That's `2h 24m` vs `44m`, a 3x speedup! Absolutely massive.

conorsch (approved):

PR #5128, opened 2025-03-13
pindexer: allow self-reset on schema changes

### LINKED ISSUE #5125: Allow pindexer to handle schema changes itself
Right now, when the pindexer schema changes, one needs to know about this inherently, manually wipe *the entire* database, and then reindex everything from scratch.

Wiping everything is often unnecessary, especially if a minor change in one schema causes other app views to need to be reindexed, unnecessarily, as well.

Having to manually keep track of whether schema changes happened is also error prone.

# Solution

App views should know what schema version they have, and we should record it in the state we keep track of for each app view. App views should also know how to reset themselves, so that when we detect that that version doesn't match what we currently have, the app view can safely delete all of the tables it might have created in the past.

cronokirby:
## Describe your changes

Closes #5125.

After this change, one final `dropdb penumbra` will be necessary, and then from now on, pindexer will be able to manage resetting itself, and do so in a minimally destructive way, reducing indexing times on changes. This works by having each 
index know it's version, with each breaking schema change inducing an increment to the version. Then, on startup, pindexer can figure out which particular app views need to be reset based on their version changes. App views with a version > 0 will need to implement a means of actually effecting this reset. This method should delete all tables any versions of the index might have created. In hindsight, every app view should have been putting things in a schema, like the proposed view in #5113 does. This makes the reset particularly easy: `DROP SCHEMA IF EXISTS lqt CASCADE;`. For other views, you'll need knowledge of any tables it might have created at the time of this PR.

This PR also allows app views to maintain a separate layer of views on top of their actual tables, which can be updated *without* needing indexing again. This is because we add a hook which always runs on pindexer startup, and by default does nothing. If an app view has "CREATE VIEW" statements, it can run these here, destroying any views it might have created previously.

### Compatibility

The extensions to the trait were designed such that they have sensible default implementations.

- By default, the new startup hook for views does nothing.
- The version defaults to `0`.
- The reset function will panic with a helpful error if it ends up being called, but it won't until the version is *manually* set to 1, indicating the first schema change post PR.

### Testing

I tested this on a merge branch of outstanding PRs, against a devnet. I tested that updating the version for an LQT app view caused that app view, and only that app view to reset, when re-running pindexer.

### DIFF
--- a/crates/util/cometindex/src/index.rs
+++ b/crates/util/cometindex/src/index.rs
@@ -157,6 +157,31 @@ impl EventBatchContext {
     }
 }
 
+/// Represents the version of an indexing view.
+///
+/// This allows tracking breaking change at the level of each individual view,
+/// rather than on the level of the database as a whole.
+///
+/// Versions can be compared to assess breaking changes:
+/// ```
+/// assert!(Version::with_major(3) > Version::with_major(2));
+/// ```
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, PartialOrd, Ord, sqlx::Type)]
+#[sqlx(transparent)]
+pub struct Version(Option<i64>);
+
+impl Version {
+    /// Construct a new version by specifying a "major" / "breaking" number.
+    pub fn with_major(v: u64) -> Self {
+        Self(Some(v.try_into().expect("version must fit into an i64")))
+    }
+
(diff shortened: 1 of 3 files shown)


conorsch (approved):
This looks gorgeous, eager to see it in practice. Will be clear about the need for one last drop in the release notes.

PR #5133, opened 2025-03-19
Remove rocksdb dependency from key crates

cronokirby:
## Describe your changes

Seems like we accidentally added a dependency on cnidarium, and thus the expensive and annoying to compile rocksdb crate in https://github.com/penumbra-zone/penumbra/commit/64c32ef4bf33c929560ae2fc46ac127030c01f68. This fixes that.

I also added a new smoke test to detect the presence of rocksdb in a few of our binary crates. This should help a lot in the future because we can easily see what diff caused the mess up.

The new smoke test should act as the test here.

  > Doesn't affect node software, and shouldn't be a breaking change in behavior for any crate.


cronokirby:
hmm, seems like we added in cnidarium because we moved the proto interface there ; I think to continue on this we'll need to feature gate the protos out of cnidarium...

conorsch:
Unfortunately we do actually need cnidarium, at least for the proto types. It's not a terrible idea to feature-gate proto types in the cnidarium interface, so we can be selective about the imports... but as things stand, I think we're stuck with the rocksdb dep for now.

cronokirby:
Closing for now, will open a new PR once appropriate changes land in cnidarium and we have a version to point to there.

PR #5134, opened 2025-03-20
Refactor `write_params` function signature for improved flexibility

tomasandroil:
Refactored the `write_params` function in `tools/parameter-setup/src/main.rs` to accept a generic `&Path` instead of `&PathBuf`. This change enhances flexibility and aligns with best practices, allowing the function to accept both `PathBuf` and `Path` references seamlessly.

### Changes:
- `write_params(target_dir: &PathBuf, ...)` → `write_params(target_dir: &Path, ...)`
- Updated the function call site accordingly.

No logic changes were introduced—only the parameter type was generalized.

  > This is a non-consensus-affecting refactor improving code flexibility.

### DIFF
--- a/tools/parameter-setup/src/main.rs
+++ b/tools/parameter-setup/src/main.rs
@@ -1,8 +1,8 @@
 #![deny(clippy::unwrap_used)]
-use std::path::PathBuf;
 use std::{
     env, fs,
     io::{BufWriter, Result},
+    path::PathBuf,
 };
 
 use ark_groth16::{ProvingKey, VerifyingKey};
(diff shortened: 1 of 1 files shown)


tools/parameter-setup/src/main.rs
```diff
@@ -1,8 +1,8 @@
 #![deny(clippy::unwrap_used)]
-use std::path::PathBuf;
 use std::{
     env, fs,
     io::{BufWriter, Result},
+    path::PathBuf,
```
dynst, line 5: This PR is fixing the [`ptr_arg`](https://rust-lang.github.io/rust-clippy/master/index.html#ptr_arg) clippy lint, right?

Missing an import of `path::Path` here (the build fails).

tomasandroil: Yes, I have added the missing import for `Path` in the `main.rs` file. I updated the `write_params` function to use `&Path` instead of `&PathBuf` as suggested, which improves flexibility and resolves the build failure.

PR #5137, opened 2025-03-21
Add an integrity check to cometindex / pindexer

cronokirby:
## Describe your changes

This adds a new command, reachable via `pindexer -s <DATABASE> integrity-check` which runs basic integrity checks on the raw database. Right now this can check that:
- there are no missing blocks,
- there are no missing events.

This also makes the indexing command `pindexer -s <DATABASE> index`, because I couldn't find a way to preserve the default command.

This can be run against known good database, and shouldn't fail.

  > indexing only

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6123,6 +6123,7 @@ dependencies = [
  "sqlx",
  "tokio",
  "tracing",
+ "tracing-subscriber 0.3.18",
  "url",
 ]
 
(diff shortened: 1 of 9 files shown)


conorsch:
> This also makes the indexing command pindexer -s <DATABASE> index, because I couldn't find a way to preserve the default command.

I'd prefer to avoid this breaking change. We should be able to gate on check-or-index in the main fn for pindexer. As long as the docstrings are clear about the intended effect, having a flag that alters runtime behavior is not terribly confusing.

cronokirby:
Cool, will change things to be a flag

cronokirby:
Ok, now the integrity checks are run by default, and a `--integrity-checks-only` flag will run only these.

conorsch:
Ran a reindex of mainnet data up to height `4136642`, took `23m 21s`.

conorsch (approved):

PR #5138, opened 2025-03-24
Bump pindexer lag allowance in integration test

cronokirby:
I observed some flakiness in the test because of this, where it was behind by 3 on one run, and fine on another. I think moving it to 4 is sensible. I also replaced the use of a vec with a more efficient range check.

Smoke tests should be sufficient here.

  > tests only


conorsch (approved):
The tests must pass. :metal:

PR #5140, opened 2025-03-26
Add ledger as a custody option to pcli

cronokirby:
## Describe your changes

This implements a new custody backend which communicates with a ledger device locally, using the app in https://github.com/Zondax/ledger-penumbra/. I tested that this works as you might expect, and it does, except for an update to the protos that affects the effect hash that needs to be merged into an alpha release in that repo.

I'm not super happy with the quality of the crates I used to communicate with ledger generically, but they get the job done.

You can use `pcli init ledger` to setup pcli with this backend, and then everything should "just work".
Some more polish could be needed in terms of providing informative error handling, but for now I think this code is 90% of what we need.

I also added a default implementation of `AuthorizationData` so that we can easily rebase the lqt branch on top of these changes without breaking it.

It's easiest to test with an actual device, I haven't tried using an emulator.

  > This should affect client code only

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -967,6 +967,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "block"
+version = "0.1.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0d8c1fef690941d3e7788d328517591fecc684c084084702d6ff1641e993699a"
+
+[[package]]
 name = "block-buffer"
 version = "0.9.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -995,6 +1001,36 @@ dependencies = [
  "futures-io",
  "futures-lite",
  "piper",
+]
+
+[[package]]
+name = "bluez-async"
+version = "0.7.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5ce7d4413c940e8e3cb6afc122d3f4a07096aca259d286781128683fc9f39d9b"
+dependencies = [
+ "async-trait",
+ "bitflags 2.6.0",
+ "bluez-generated",
+ "dbus",
+ "dbus-tokio",
+ "futures",
+ "itertools 0.10.5",
+ "log",
+ "serde",
+ "serde-xml-rs",
+ "thiserror 1.0.69",
+ "tokio",
+ "uuid",
+]
+
+[[package]]
+name = "bluez-generated"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4d1c659dbc82f0b8ca75606c91a371e763589b7f6acf36858eeed0c705afe367"
+dependencies = [
+ "dbus",
 ]
 
 [[package]]
@@ -1039,6 +1075,33 @@ dependencies = [
  "memchr",
  "regex-automata 0.4.9",
  "serde",
+]
+
+[[package]]
+name = "btleplug"
+version = "0.11.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ba345f9db94939c72959b2008abe1ffcdbcaa235243fd92ad12436532ff199cf"
+dependencies = [
+ "async-trait",
+ "bitflags 2.6.0",
+ "bluez-async",
+ "cocoa",
+ "dashmap",
+ "dbus",
+ "futures",
+ "jni",
+ "jni-utils",
+ "libc",
+ "log",
+ "objc",
+ "once_cell",
+ "static_assertions",
+ "thiserror 1.0.69",
+ "tokio",
+ "tokio-stream",
+ "uuid",
+ "windows 0.52.0",
 ]
 
 [[package]]
@@ -1117,6 +1180,12 @@ dependencies = [
  "libc",
  "shlex",
 ]
+
+[[package]]
+name = "cesu8"
+version = "1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6d43a04d8753f35258c91f8ec639f792891f748a1edbd759cf1dcea3382ad83c"
 
 [[package]]
 name = "cexpr"
@@ -1335,6 +1404,36 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "67ba02a97a2bd10f4b59b25c7973101c79642302776489e030cd13cdab09ed15"
 
 [[package]]
+name = "cocoa"
+version = "0.25.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f6140449f97a6e97f9511815c5632d84c8aacf8ac271ad77c559218161a1373c"
+dependencies = [
+ "bitflags 1.3.2",
+ "block",
+ "cocoa-foundation",
+ "core-foundation",
+ "core-graphics",
+ "foreign-types 0.5.0",
+ "libc",
+ "objc",
+]
+
+[[package]]
+name = "cocoa-foundation"
+version = "0.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8c6234cbb2e4c785b456c0644748b1ac416dd045799740356f8363dfe00c93f7"
+dependencies = [
+ "bitflags 1.3.2",
+ "block",
+ "core-foundation",
+ "core-graphics-types",
+ "libc",
+ "objc",
+]
+
+[[package]]
 name = "colored"
 version = "2.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1353,6 +1452,16 @@ dependencies = [
  "serde",
  "serde_json",
  "yansi",
+]
+
+[[package]]
+name = "combine"
+version = "4.6.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ba5a308b75df32fe02788e748662718f03fde005016435c444eea572398219fd"
+dependencies = [
+ "bytes",
+ "memchr",
 ]
 
 [[package]]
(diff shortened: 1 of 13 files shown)


cronokirby:
Seems like the CI environment is missing dbus related packages?

cronokirby:
> > It's easiest to test with an actual device
> 
> What can I expect to work? Is there no limitation on pcli functionality with soft-kms vs ledger backend? Can I use the ledger backend to send, swap, vote, and withdraw? Asking before testing with a hardware device, so I know what to expect.

With version 0.1.2 of the zondax app, you can expect initialization of the custody backend to work. You should expect signing to execute, but then for the transaction to be rejected by nodes because the signatures fail to verify.

For further testing, I removed the `encrypted_backref` from the `SpendBody` message, and signatures would then verify. This indicates the need for a version 0.1.3 of the app there.

cronokirby:
Okay, I've added feature gating for the ledger stuff in pcli ; our CI does still exercise all features though, so we'll need to look into that. I think it's good that we do so, because we want to make sure the code compiles at least, even if most users don't need it.

conorsch:
There's only a single CI job failing now, the one that requires nightly-rust for building the docs. I'm going to bite the bullet and wire up opt-in support for nightly in the nix flake, after which CI should be happy again.

cronokirby:
Note: should be possible to install both rust stable and nightly in the flake: https://github.com/oxalica/rust-overlay/blob/master/docs/reference.md

conorsch:
Looks like we've straightened out the CI, which is dope. Bear with me: I'd like to exercise the integration with a hardware token before finally approving. Since we want more changes to land upstream in the near future, we can wait for those changes to land before getting loud about the integration in release notes.

conorsch:
Took a stab at exercising the new ledger backend for pcli. I wasn't able to get it working, mostly due to struggles with the installing the app on the ledger device. In a nutshell, my question is: **How do I obtain v2 of the Ledger firmware?** A more detailed testing report follows, to aid future reviewers and/or maintainers.

First attempt, without opt-in feature flag flags, not recognizing the `ledger` backend:

```
❯ cargo run --release --bin pcli -- --home /tmp/ledger-test-1  init --grpc-url http://localhost:8080 ledger
    Finished `release` profile [optimized] target(s) in 0.32s
     Running `target/release/pcli --home /tmp/ledger-test-1 init --grpc-url 'http://localhost:8080' ledger`
error: Found argument 'ledger' which wasn't expected, or isn't valid in this context

USAGE:
    pcli init [OPTIONS] --grpc-url <GRPC_URL> <SUBCOMMAND>

For more information try --help
```

Enabling the feature-gated support with a Ledger device plugged in, but not yet unlocked, gives a very informative error message:

```
❯ cargo run --features ledger --release --bin pcli -- --home /tmp/ledger-test-1  init --grpc-url http://localhost:8080 ledger
    Finished `release` profile [optimized] target(s) in 0.32s
     Running `target/release/pcli --home /tmp/ledger-test-1 init --grpc-url 'http://localhost:8080' ledger`
Error: Status: Device locked
```

Using the feature flag with an unlocked Ledger complains about an unknown app:

```
❯ cargo run --features ledger --release --bin pcli -- --home /tmp/ledger-test-1  init --grpc-url http://localhost:8080 ledger
    Finished `release` profile [optimized] target(s) in 0.92s
     Running `target/release/pcli --home /tmp/ledger-test-1 init --grpc-url 'http://localhost:8080' ledger`
Error: unknown app: BOLOS
```

That's expected, given that I'm using a factory-reset Ledger device for testing. I tried following the [setup docs in the Zondax Penumbra repo](https://github.com/Zondax/ledger-penumbra/?tab=readme-ov-file), specifically by downloading the `installer_nanos_plus.sh` from the recent [v0.1.3 release](https://github.com/Zondax/ledger-penumbra/releases/tag/v0.1.3), but I'm struggling to use that script to install the Penumbra app:

```
❯ poetry run bash installer_nanos_plus.sh load
Generated random root public key : b'04a1197de9c1799eaf6c4b74ff90b50611c73516fd0ff2b21f3a453bcf09c962117f97f95cd3fd808cf2e9356abf8f5efde9364c773eadff6a1a5df1889261b62a'
Using test master key b'04a1197de9c1799eaf6c4b74ff90b50611c73516fd0ff2b21f3a453bcf09c962117f97f95cd3fd808cf2e9356abf8f5efde9364c773eadff6a1a5df1889261b62a'
Using ephemeral key b'04274f83ced6090400f9a8de26a49ecf802b091247e0257be5bc9e1b058c4950b81d1e4fe1513edab62d83be3a33afdf808ef5024801e9c95f393d2ce691c5ee58'
Broken certificate chain - loading from user key
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/home/conor/.cache/pypoetry/virtualenvs/ledger-testing-17Kjne5g-py3.12/lib/python3.12/site-packages/ledgerblue/loadApp.py", line 298, in <module>
    main(sys.argv[1:])
  File "/home/conor/.cache/pypoetry/virtualenvs/ledger-testing-17Kjne5g-py3.12/lib/python3.12/site-packages/ledgerblue/loadApp.py", line 257, in main
    loader.createApp(code_length, args.apiLevel, args.dataSize, paramsSize, args.appFlags, args.bootAddr|1)
  File "/home/conor/.cache/pypoetry/virtualenvs/ledger-testing-17Kjne5g-py3.12/lib/python3.12/site-packages/ledgerblue/hexLoader.py", line 374, in createApp
    self.exchange(self.cla, 0x00, 0x00, 0x00, data)
  File "/home/conor/.cache/pypoetry/virtualenvs/ledger-testing-17Kjne5g-py3.12/lib/python3.12/site-packages/ledgerblue/hexLoader.py", line 207, in exchange
    return self.scpUnwrap(bytes(self.card.exchange(apdu)))
                                ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/conor/.cache/pypoetry/virtualenvs/ledger-testing-17Kjne5g-py3.12/lib/python3.12/site-packages/ledgerblue/comm.py", line 163, in exchange
    raise CommException("Invalid status %04x (%s)" % (sw, possibleCause), sw, response)
ledgerblue.commException.CommException: Exception : Invalid status 511f (The OS version on your device does not seem compatible with the SDK version used to build the app)
```

There are a zillion ways to install Python deps; I opted for using a Nix flake to install `poetry`, then using `poetry` to install `ledgerblue` in an otherwise empty python project. I confirmed that the generated `poetry.lock` file pins `ledgerblue` version `0.1.54`:

```
[[package]]
name = "ledgerblue"
version = "0.1.54"
description = "Python library to communicate with Ledger devices"
optional = false
python-versions = ">=3"
files = [
    {file = "ledgerblue-0.1.54-py3-none-any.whl", hash = "sha256:df37562570ac4f9f38dc58f33af609c4689b1b7df889081889159d2a21d9c723"},
    {file = "ledgerblue-0.1.54.tar.gz", hash = "sha256:1e7f7d493e919c447a5c8ebeaea0373bdfda0bec216284cecdea07185fdf173e"},
]
```

which matches the most recently available version of ledgerblue [available on PyPI](https://pypi.org/project/ledgerblue/0.1.54/). The error message is quite clear: "The OS version on your device does not seem compatible with the SDK version used to build the app"; coupled with the statement in the [Zondax project readme](https://github.com/Zondax/ledger-penumbra/?tab=readme-ov-file) that "This project requires Ledger firmware 2.0" the solution is clear: I must figure out how to install Ledger firmware 2.0 on my device. I've updated the device several times via the official Ledger Live Desktop app, but no 2.0 option is offered, even if I opt-in to developer settings. I assume it's possible to build v2 of the firmware from source via the Zondax repo, but that option seems at odds with the statement, also in the Zondax README, that the installer script is all that's required. 

Part of my goal in offering to verify the new functionality here is to write a comprehensive test plan, which in turn gets us docs for free. But not quite there yet. Will keep working at it.

conorsch (changes requested):
> Looks like we've straightened out the CI

Partially true, but the container builds still fail, which isn't surfaced on PR CI. We can simply update the `cargo build --release` invocation to `cargo build --release --bins` to avoid the new usb deps when building `pcli`; that'll solve for the containerfile. But I haven't tested the implications for the cargo-dist release tooling, and I'll need to use a different repo to exercise those workflows. Punting on that for now, to ensure that near-term fixes can land on schedule.

cronokirby:
So, you actually only need version "1.2" of the firmware. I've been testing on version 1.3.1 and it works. I think calling that version "2" is a misnomer. Also, the error about the unknown app "BOLOS" is very misleading, and actually just indicates that you need to open the penumbra app first. I'll add a bit to the error message to make the clearer.

As far as installation, I've been using: https://hub.zondax.ch/penumbra. A bit worse for reproducibility though.

conorsch:
Rebased to resolve conflicts due to version drift (#5150).

conorsch:
Spent some more time with this changeset, and confirmed that it works as advertised. UX-wise, it's still a bit rough around the edges, but that's precisely why we're feature-gating and leaving it off by default. In my testing, preparing a single send transaction took about 90s, during which the Ledger displayed "Processing":

penumbra-ledger-1

On subsequent verification steps, lack of registry support means that raw asset ids are displayed, rather than the more friendly human-readable names:

penumbra-ledger-2

Similarly, the need to confirm each action within the plan was a bit off-putting:

penumbra-ledger-3

But this PR is not the place to resolve those quibbles, I merely document them here for those following along at home. 

Regarding the build environment changes, `cargo build --release` will now require the presence of USB packages; using `cargo build --release --bins`, as the containerfile now does, will avoid the USB requirement. Some docs updates to [the guide](https://github.com/penumbra-zone/guide) are warranted, but I'll add those later. In particular, I'll make sure to include notice in the release notes about the build env changes, because I expect it will break downstream builds for a few interested parties.

PR #5141, opened 2025-03-27
ci: add nightly rust to nix dev shell

conorsch:
## Describe your changes

Updates the nix flake setup to provide an optional devshell for nightly rust. This is mostly useful for the "rustdocs"" build, which requires use of rust nightly for the combined index landing page. Also updates the CI workflows to use the nix env for building rustdocs.

See related discussion in https://github.com/penumbra-zone/penumbra/pull/5140#issuecomment-2758892215. 

## Testing and review

Certainly we want CI to pass. As a bonus, it's worth confirming that executing the script `./deployments/script/rust-docs` works both inside and outside of a nix shell. 

  > CI/build logic only, no changes to application code

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -51,10 +51,11 @@
           craneLib = (crane.mkLib pkgs).overrideToolchain rustToolchain;
 
           # Add nightly Rust toolchain, required for building the rustdocs with combined index landing page.
-          nightlyRustToolchain = pkgs.rust-bin.nightly.latest.default.override {
+          # https://github.com/oxalica/rust-overlay/blob/master/README.md#cheat-sheet-common-usage-of-rust-bin
+          nightlyRustToolchain = pkgs.rust-bin.selectLatestNightlyWith(toolchain: toolchain.override {
             extensions = [ "rust-src" "rust-analyzer" ];
             targets = [ "wasm32-unknown-unknown" ];
-          };
+          });
           nightlyCraneLib = (crane.mkLib pkgs).overrideToolchain nightlyRustToolchain;
 
           # Important environment variables so that the build can find the necessary libraries
--- a/flake.nix
+++ b/flake.nix
@@ -52,7 +52,7 @@
 
           # Add nightly Rust toolchain, required for building the rustdocs with combined index landing page.
           # https://github.com/oxalica/rust-overlay/blob/master/README.md#cheat-sheet-common-usage-of-rust-bin
-          nightlyRustToolchain = pkgs.rust-bin.selectLatestNightlyWith(toolchain: toolchain.override {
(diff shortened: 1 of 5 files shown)


flake.nix
```diff
@@ -50,11 +50,54 @@
           rustToolchain = pkgs.rust-bin.fromRustupToolchainFile ./rust-toolchain.toml;
           craneLib = (crane.mkLib pkgs).overrideToolchain rustToolchain;
 
+          # Add nightly Rust toolchain, required for building the rustdocs with combined index landing page.
+          nightlyRustToolchain = pkgs.rust-bin.nightly.latest.default.override {
+            extensions = [ "rust-src" "rust-analyzer" ];
+            targets = [ "wasm32-unknown-unknown" ];
+          };
```
cronokirby, line 57:
```suggestion
		  # https://github.com/oxalica/rust-overlay/blob/master/README.md#cheat-sheet-common-usage-of-rust-bin
          nightlyRustToolchain = pkgs.rust-bin.selectLatestNightlyWith(toolchain: toolchain.override {
            extensions = [ "rust-src" "rust-analyzer" ];
            targets = [ "wasm32-unknown-unknown" ];
          });
```

we should use this method instead, which will search backwards to avoid flakiness when nightly isn't available on a particular day.

flake.nix
```diff
@@ -148,41 +191,26 @@
...
+              packages = commonDevPackages;
+              shellHook = ''
+                ${commonShellHook}
+                echo "Using stable Rust from rust-toolchain.toml"
+              '';
+            };
+
+            nightly = nightlyCraneLib.devShell {
+              inherit LIBCLANG_PATH ROCKSDB_LIB_DIR;
+              inputsFrom = [ penumbra ];
```
cronokirby, line 207: I'm a bit unsure about this, because I think this will pull in whatever inputs are necessary to build the penumbra package, including the stable toolchain. Probably fine for now though, to the extent we care about the bloat, we can just directly make a derivation for the docs in the future.

cronokirby (dismissed):
I think we should use the more stable method of selecting a nightly toolchain, otherwise LGTM

flake.nix
conorsch: The `inputsFrom` declaration on the nightly shell was absolutely required: the nightly build failed without it, complaining about openssl paths. It's possible that we could decouple the inputs further, but for now, I consider it necessary.

It might be useful to have a "lightweight" devshell that's only rust, without all the dev-env requirements like cometbft, postgresql, etc. But I don't see a need for that now, so let's punt.

conorsch:
> I think we should use the more stable method of selecting a nightly toolchain

Huge improvement, thanks for taking the time to rifle through the docs.

PR #5142, opened 2025-03-28
Refactor `server.rs`: cleanup imports, remove redundant clone, and clarify comments

torrpriius:
This PR improves the clarity and ergonomics of the `server.rs` module in the Penumbra ABCI app by applying the following changes:

- Removed unnecessary `self::` prefix in import paths for better readability.
- Removed a redundant `.clone()` call on `Info` instance since it's only consumed once.
- Replaced a misleading `Safety:` comment with a more accurate statement clarifying that all required components are passed to the builder.
- Removed unused `#[allow(...)]` directives in the `bounds_test` module.

### DIFF
--- a/crates/core/app/src/server.rs
+++ b/crates/core/app/src/server.rs
@@ -1,10 +1,11 @@
 //! Facilities related to the Penumbra app's ABCI server.
 
 use {
-    self::{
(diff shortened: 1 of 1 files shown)


erwanor:
Thanks, can reopen later!

Issue #5150, opened 2025-04-08
Release v1.4.0, via chain-upgrade


conorsch:
# Tooling Release

Today there will be patches released to critical IBC dependencies such as `tendermint-rs`, which provide security fixes. The `penumbra-1` chain has voted to halt at height `4378762` to coordinate an upgrade. This issue tracks preparation for `v1.4.0` which will include the necessary dependency bumps.

## Necessary changes

- [x] Prepare no-op migration against main: https://github.com/penumbra-zone/penumbra/pull/5151
- [x] Wait for publication of fix on tendermint-rs side: https://github.com/informalsystems/tendermint-rs/releases/tag/v0.40.3
- [x] Bump [ibc-types](https://github.com/penumbra-zone/ibc-types) to consume new tendermint-rs dep - https://github.com/penumbra-zone/ibc-types/pull/98
- [x] Release new version of ibc-types, publishing the patched version - https://crates.io/crates/ibc-types/0.15.1
- [x] Update protocol repo deps to depend on patched `tendermint-rs` - https://github.com/penumbra-zone/penumbra/pull/5155
- [x] Update protocol repo deps to depend on patched `ibc-types` - https://github.com/penumbra-zone/penumbra/pull/5155
- [x] Perform interactive QA on devnet
- [x] Cut final v1.4.0 release, announce

## Testing and assurance

We'll test these changes on a bespoke devnet. We've opted not to use the PL testnet because that's currently running LQT code (#5010).

## Compatibility
We're marking this a minor release, but critically a migration is required, given the chain-upgrade to coordinate the version bumps.

conorsch:
Release has shipped: https://github.com/penumbra-zone/penumbra/releases/tag/v1.4.0

PR #5151, opened 2025-04-08
chore: add migration for mainnet3

conorsch:
## Describe your changes
Updates the default `pd migrate` behavior to apply mainnet migration 3, a no-op migration to coordinate dependency updates related to IBC handling.

See #5150.

## Testing and review
The changes presented here are necessary, but not sufficient. Once the suitable fixes have been released, we'll need to add them here, and likely also in [ibc-types](https://github.com/penumbra-zone/ibc-types), and bump local references. Then we'll exercise the upgrade logic on a devnet created for this purpose.

  > This PR will affect consensus and therefore provides a migration to aid in applying the consensus-breaking change safely.

### DIFF
--- a/crates/bin/pd/src/migrate/mainnet3.rs
+++ b/crates/bin/pd/src/migrate/mainnet3.rs
@@ -42,7 +42,7 @@ pub async fn migrate(
     let (migration_duration, post_upgrade_root_hash) = {
         let start_time = std::time::SystemTime::now();
 
-        migrate_app_version(&mut delta, 9).await?;
+        migrate_app_version(&mut delta, 10).await?;
 
         // Reset the application height and halt flag.
         delta.ready_to_start();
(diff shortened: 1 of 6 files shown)


crates/bin/pd/src/migrate/mainnet3.rs
```diff
@@ -0,0 +1,103 @@
...
+        .get_block_height()
+        .await
+        .expect("chain state has a block height");
+    let post_upgrade_height = pre_upgrade_height.wrapping_add(1);
+
+    let mut delta = StateDelta::new(initial_state);
+    let (migration_duration, post_upgrade_root_hash) = {
+        let start_time = std::time::SystemTime::now();
+
+        migrate_app_version(&mut delta, 9).await?;
```
erwanor, line 45: should this be 9 or 10, i think 10 right?

conorsch: You're right, it should be 10. This is the part I always forget in writing no-op migrations. I've pushed another commit that changes L45 in `mainnet3.rs` to use 10, and also bumps the default app version for pd to be 10, too.

cronokirby (approved):

erwanor (approved):
Moving this to testing.

PR #5167, opened 2025-04-17
view service: rewards

TalDerei:
## Describe your changes

emulates (almost) the proto signature of our `tournamentVotes` view service implementation, but conceptually differs in how it's consumed. While `tournamentVotes` optionally takes an epoch index and retrieves the votes cast in the tournament for that specific epoch, `tournamentRewards` will instead retrieve a list of _all_ historical votes and associated rewards. 

  > LQT branch

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1319,7 +1319,7 @@ dependencies = [
 
 [[package]]
 name = "cnidarium-component"
-version = "2.0.0-alpha.7"
+version = "2.0.0-alpha.8"
 dependencies = [
  "anyhow",
  "async-trait",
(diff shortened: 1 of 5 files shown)


cronokirby:
Couldn't we use the existing RPC, with the "block_height" argument, which will return all the votes up to that height? If you set that to the current height, then you get the right result, right?

TalDerei:
> Couldn't we use the existing RPC, with the "block_height" argument, which will return all the votes up to that height? If you set that to the current height, then you get the right result, right?

we can certainly extend the implementation — afaict, it currently has a hardcoded path that expects an epoch index. Also, it should be updated to use a streaming rpc

TalDerei:
closing in favor of https://github.com/penumbra-zone/penumbra/pull/5168

PR #5168, opened 2025-04-17
view service: streaming support for tournament votes

TalDerei:
## Describe your changes

`tournamentVotes` should be a streaming rpc

simplification based on https://github.com/penumbra-zone/penumbra/pull/5167#issuecomment-2811561168

  > LQT branch


cronokirby (commented):
What if we also made it so that instead of returning a repeated `TournamentVotesResponse.Vote` we instead just returned one such vote, and inlined the fields into the response? i.e. the response just has one vote, and you stream the responses to have many of them

TalDerei:
> What if we also made it so that instead of returning a repeated `TournamentVotesResponse.Vote` we instead just returned one such vote, and inlined the fields into the response? i.e. the response just has one vote, and you stream the responses to have many of them

adjusted to match the streaming response format

TalDerei:
@cronokirby let’s hold off on the review for now. I’m not super convinced this is the optimal approach compared to having a dedicated rewards method. I’ll give it more thought tomorrow.

TalDerei:
@cronokirby @erwanor requesting another review, think it's good to go

proto/penumbra/penumbra/view/v1/view.proto
```diff
@@ -195,6 +195,8 @@ message TournamentVotesResponse {
   core.asset.v1.Value reward = 3;
   // The transaction ID for the vote. 
   core.txhash.v1.TransactionId transaction = 4;
+  // Epoch index the vote occured in.
+  uint64 epoch_index = 5;
```
TalDerei, line 199: comment: seems like useful information to retrieve when optionally querying by height, rather than relying on the SCT service.

TalDerei:
> If this shape works for the web contexts

It works, but not ideal. reverted back to streaming repeated `TournamentVotesResponse.Vote`

cronokirby (approved):

PR #5169, opened 2025-04-18
view service: store asset ID for voting rewards

TalDerei:
## Describe your changes

stores the asset ID of the reward, allowing consumers to derive the associated validator metadata at the end of the epoch for rewards that are distributed pro rata across delegations.

  > LQT branch


cronokirby (changes requested):
I don't think this is necessary, the reward is already a Value

PR #5170, opened 2025-04-22
fix IBC `packet_acknowledgements` query response

noot:
## Describe your changes

fix IBC `packet_acknowledgements` query response to utilize the `packet_commitment_sequences` in the request. this also fixes the behaviour of the method; previously, it was using `get_ack_sequence` which is used to store sequences for received `MsgAcknowledgement`s (which are for acks written on a counterparty), where we actually wanted to check for acks which were written to our own state (as we're using `snapshot.get_packet_acknowledgement`).

note that this has been tested on astria by ensuring that failure acks are handled properly (they weren't before), since hermes uses `query_packet_acknowledgements`.

n/a

  > this only affects RPC responses


conorsch:
Hey @noot, thanks for this fix. FYI it's available now as part of `v1.5.0`: https://github.com/penumbra-zone/penumbra/releases/tag/v1.5.0

PR #5173, opened 2025-04-22
pclientd: read key material interactively

### LINKED ISSUE #5171: pclientd: read key material from stdin
When initializing a new instance of `pclientd`, users must provide either a full seed phrase of a fullviewingkey on the command line. As a result, the key material will remain available in the system's shell history, which amounts to insecure handling of sensitive information.

`pclientd` should prompt for input when run interactively, or else read from stdin (to support running non-interactively). 

We could ignore this problem. 

**Additional context**
Back in #2693 we revised `pcli` to prompt for seed phrases on the command line interactively, rather than supporting reading the seed phrase as a CLI argument. We should update pclientd to do the same.

conorsch:
## Describe your changes
Updates `pclientd` to prompt for key material interactively, via a masked prompt, or else read from stdin, rather than accepting key material as CLI args.

Closes #5171.

## Testing and review

I updated the integration tests to exercise the CLI invocations, so CI passing is sufficient assurance her.

  > CLI-args only, no changes to app code

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4911,6 +4911,7 @@ dependencies = [
  "prost 0.13.4",
  "rand",
  "rand_core",
+ "rpassword",
  "rustls 0.23.21",
  "serde",
  "serde_json",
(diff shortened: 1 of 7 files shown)


cronokirby (approved):
Yeah I think having key material on the command line was a bad idea

PR #5174, opened 2025-04-22
pcli: read fvk interactively

conorsch:
## Describe your changes

Updates the `pcli init view-only` CLI behavior
to read the FVK from interactive input, or else stdin, but not accept a CLI arg that will be stored in shell history.

Refs #5171. Inspired by https://github.com/penumbra-zone/penumbra/pull/5173, so best review in tandem with that PR.

## Testing and review

It's worth considering whether we want to change the CLI behavior at all: it could break downstream client scripting. I argue that it's worthwhile, given the baseline security improvements, and also because it only affects creation of new wallets via pcli, and existing wallets will work just fine.

  > CLI args only, no changes to app code

### DIFF
--- a/crates/bin/pcli/src/command/init.rs
+++ b/crates/bin/pcli/src/command/init.rs
@@ -3,7 +3,7 @@ use std::{
     str::FromStr,
 };
 
-use anyhow::Result;
+use anyhow::{Context, Result};
 use camino::Utf8PathBuf;
 use penumbra_sdk_custody::threshold;
 #[cfg(feature = "ledger")]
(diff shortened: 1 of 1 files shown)


crates/bin/pcli/src/command/init.rs
```diff
@@ -102,6 +99,25 @@ pub enum SoftKmsInitCmd {
...
+    let mut password = String::new();
+    // The `rpassword` crate doesn't support reading from stdin, so we check
+    // for an interactive session. We must support non-interactive use cases,
+    // for integration with other tooling.
+    if std::io::stdin().is_terminal() {
+        password = rpassword::prompt_password(msg)?;
+    } else {
+        while let Ok(n_bytes) = std::io::stdin().lock().read_to_string(&mut password) {
+            if n_bytes == 0 {
+                break;
```
cronokirby, line 113: I think we want to continue here instead? otherwise the loop doesn't make much sense

cronokirby (approved):
This is a good idea!

PR #5181, opened 2025-04-28
LQT: Implement fixed duration

cronokirby:
## Describe your changes

This adds a new, optional, parameter to the LQT knobs which adds an end block to the rewards. The way the mechanism is supposed to work is that once that block is reached, no further rewards will be distributed.

This also adds an optimization in the base case where rewards are turned off, to not execute any of the LQT distribution logic at all. (If the amount of rewards to distribute is 0, you don't bother trying to fetch dex information and all that jazz).

For testing easiest way is probably setting up a devnet, and testing that LQT is only enabled for a fixed time with this change. On testnet this should be able to be applied transparently, and we could even try a temporary shutoff and startup again with some governance votes.

  > This is consensus breaking, in that if you exercised a governance vote with the new params, you would observe a divergence in nodes pre and post this patch.


crates/core/component/distributions/src/component.rs
```diff
@@ -54,12 +54,32 @@ impl Component for Distributions {
...
-            .expect("distribution parameters should be available")
-            .liquidity_tournament_incentive_per_block
-            .into();
+            .expect("distribution parameters should be available");
+        let increase = Amount::from(params.liquidity_tournament_incentive_per_block);
+        let current_block = u64::try_from(end_block.height).unwrap_or_default();
+        let end_block = params
+            .liquidity_tournament_end_block
+            .map_or(0u64, |b| b.get());
+        if current_block >= end_block {
```
erwanor, line 66: I think we need to add a predicate that `end_block != 0` right?

crates/core/component/distributions/src/component.rs
```diff
@@ -54,12 +54,32 @@ impl Component for Distributions {
...
+            // Doing this unconditionally for robustness. This should, in theory,
+            // only need to be an edge trigger though, like the event.
+            tracing::debug!(epoch_index, "zeroing out LQT reward issuance");
+            state.set_lqt_reward_issuance_for_epoch(epoch_index, 0u64.into());
```
erwanor, line 70: I think burning is fine, it's a potential sharp edge, but it reduces the amount of value flows.

erwanor (approved):
General ACK, two possible amendments:
- allowing open-ended seasons seems like a plus and avoids surprising behavior (omitting an expiration block/epoch bricks the season). I think that's the original intent and the control flow just needs to case `!= 0`. 
- the community pool *could* reappropriate funds. not doing so is a sharp-edge, but i do think we come out ahead with less value flows. we should however document it.

crates/core/component/distributions/src/component.rs
cronokirby: Good catch! The intent was end_block = None => no limit, to preserve current behavior.

crates/core/component/distributions/src/component.rs
cronokirby: This doesn't actually burn anything, because the rewards are only deducted from the community pool in the end epoch handler. (You should check this to confirm my understanding)

PR #5183, opened 2025-04-29
asset: fix ibc transfer path parsing

erwanor:
## Describe your changes

This fix a bug in the Cosmos query service that would prevent certain IBC assets from being returned by `ibc.applications.transfer.v1.Query/DenomTraces`

Querying the total supply:
```bash
echo "{}" | grpcurl -plaintext penumbra-1-skip-grpc.polkachu.com:21990 cosmos.bank.v1beta1.Query/TotalSupply | grep 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2
```

Returns:
```
      "denom": "transfer/channel-0/transfer/08-wasm-1369/0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2"
```

as expected.

But querying the denom traces:
```bash
 echo ‘{}’ | grpcurl -plaintext penumbra-1-skip-grpc.polkachu.com:21990 ibc.applications.transfer.v1.Query/DenomTraces | grep 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2
```

does not return anything (unexpected).

With this fix, it should return the denom trace as well.

  > RPC change

### DIFF
--- a/crates/core/asset/src/asset/denom_metadata.rs
+++ b/crates/core/asset/src/asset/denom_metadata.rs
@@ -9,7 +9,6 @@ use anyhow::{ensure, Context};
 use decaf377::Fq;
 use penumbra_sdk_num::Amount;
 use penumbra_sdk_proto::{penumbra::core::asset::v1 as pb, view::v1::AssetsResponse, DomainType};
-use regex::Regex;
 use serde::{Deserialize, Serialize};
 
 use crate::{
@@ -226,7 +225,6 @@ impl Inner {
     /// not include a unit for the base denomination.
     pub fn new(base_denom: String, mut units: Vec<BareDenomUnit>) -> Self {
         let id = Id(Fq::from_le_bytes_mod_order(
-            // XXX choice of hash function?
             blake2b_simd::Params::default()
                 .personal(b"Penumbra_AssetID")
                 .hash(base_denom.as_bytes())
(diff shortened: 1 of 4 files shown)


crates/core/asset/src/asset/denom_metadata.rs
```diff
@@ -585,6 +579,94 @@ impl Display for Unit {
...
 }
 
+pub mod parse {
+    use lazy_static::lazy_static;
+
+    lazy_static! {
+        static ref IBC_RE: regex::Regex = regex::Regex::new(
+            r"^(?<path>transfer/[a-z0-9][a-z0-9\-]{0,63}(?:/transfer/[a-z0-9][a-z0-9\-]{0,63})*)/(?<denom>[^/][A-Za-z0-9/._\-]*)$"
+        ).expect("regex compilation works");
+    }
```
cronokirby, line 589: nit: would be good to have a comment for the source of this regex, if this is coming from some upstream cometbft source

cronokirby (approved):
Generally looks good

crates/core/asset/src/asset/denom_metadata.rs
erwanor: unfortunately the source is me, i had to make it up

conorsch:
To verify, I built `pd` from `main` branch and bootstrapped a mainnet node:

```
❯ echo "{}" | grpcurl -plaintext localhost:8080 cosmos.bank.v1beta1.Query/TotalSupply | grep 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2
      "denom": "transfer/channel-0/transfer/08-wasm-1369/0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2",

❯  echo "{}" | grpcurl -plaintext localhost:8080 ibc.applications.transfer.v1.Query/DenomTraces | grep 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2
```

Comparing with the same node config, running with `pd` built from this branch:

```
# initial check same as before, no regression:
❯ echo "{}" | grpcurl -plaintext localhost:8080 cosmos.bank.v1beta1.Query/TotalSupply | grep 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2
      "denom": "transfer/channel-0/transfer/08-wasm-1369/0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2",
# populated denom trace output, this is the fix:      
❯ echo "{}" | grpcurl -plaintext localhost:8080 ibc.applications.transfer.v1.Query/DenomTraces | grep 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2
      "baseDenom": "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2"
```

It's not straightforward to add an integration test for this, since we need IBC'd assets on the target chain, and a local devnet config as used in CI won't have that. Might be worthwhile to tack onto the `testnet-integration` tests, but not blocking merge for that, since this fix, once released, will unblock external dev teams.

conorsch (approved):

PR #5191, opened 2025-04-30
Backport AppParameter event fix from 1.5. to v1.3.x

cronokirby:
C.f. changes in #5188


conorsch:
CI is failing, looking for a new proto that was merged in https://github.com/penumbra-zone/penumbra/pull/5179, but it's correct that's not present in this historical version, because we haven't backported it.

conorsch (approved):
Confirmed that the new branch `release/1.3.x` was created from the tag `v1.3.1` which is the most recent release on that series.

conorsch:
Post-merge this change should be release as `v1.3.2`.

PR #5197, opened 2025-05-12
pindexer: lqt: handle an edge case in streak calculation properly

cronokirby:
## Describe your changes

We were incorrectly selecting for the most recent non-zero gap,
by not handling the most recent epoch a given delegator has voted in.
In that case, there's no "end" to the gap, and we need to handle
the value being NULL.

The effect of this change is that streaks are now calculated correctly,
with the weird behavior of seeing a streak larger than the number
of epochs voted in not appearing.

Closes https://github.com/penumbra-zone/web/issues/2334

  > indexing only


conorsch:
Deployed to testnet infra, refs #5010.

conorsch (approved):

PR #5198, opened 2025-05-15
fix: add community-pool rpc support

### LINKED ISSUE #4381: `pcli query community-pool balance` doesn't work
**Describe the bug**

The command `pcli query community-pool balance` returns an error when you run it.

**To Reproduce**

```
$ pcli query community-pool balance
Scanning blocks from last sync height 4076 to latest height 4076
Error: cannot process Community Pool balance data

Caused by:
    status: Unknown, message: "grpc-status header missing, mapped from HTTP status code 405", details: [], metadata: MetadataMap { headers: {} }
```

**Expected behavior**

The community pool balance should be displayed.

**Additional context**

This is not specific to the Penumbra Labs CI node; I tested this with my own fullnode and the same error was returned.

conorsch:
## Describe your changes
We've been advertising this RPC in the protobuf docs, and even in the pcli subcommands, but it hasn't ever worked. To resolve, this PR:

* hooks up the missing CommunityPool RPC in `pd`
* exercises the already-existing `pcli q community-pool balance` logic via integration tests
* adds a fix for `pcli tx community-pool-deposit` to require a value, along with tests

Closes #4381. 

## Testing

I added tests to CI, so give them a visual review as a sanity-check, but green on CI is good enough to merge.

  > Wires up documented by missing RPC; no changes to consensus logic.

### DIFF
--- a/crates/core/component/community-pool/src/component/rpc.rs
+++ b/crates/core/component/community-pool/src/component/rpc.rs
@@ -1,0 +1,86 @@
+use penumbra_sdk_proto::core::{
+    asset::v1::{AssetId, Value},
+    component::community_pool::v1::{
+        query_service_server::QueryService, CommunityPoolAssetBalancesRequest,
+        CommunityPoolAssetBalancesResponse,
+    },
+};
+
+use async_stream::try_stream;
+use futures::{StreamExt, TryStreamExt};
+use std::pin::Pin;
+use tonic::Status;
+use tracing::instrument;
+
+use cnidarium::Storage;
+
+use super::StateReadExt;
+
+pub struct Server {
+    storage: Storage,
+}
+
+impl Server {
+    pub fn new(storage: Storage) -> Self {
+        Self { storage }
+    }
+}
+
+#[tonic::async_trait]
+impl QueryService for Server {
+    /// Stream of asset balance info within the CommunityPool, to satisfy the "repeated"
+    /// field in the protobuf spec.
+    type CommunityPoolAssetBalancesStream = Pin<
+        Box<
+            dyn futures::Stream<Item = Result<CommunityPoolAssetBalancesResponse, tonic::Status>>
+                + Send,
+        >,
+    >;
+
+    #[instrument(skip(self, request))]
+    async fn community_pool_asset_balances(
+        &self,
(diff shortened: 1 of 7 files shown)


crates/core/component/community-pool/src/component/rpc.rs
```diff
@@ -0,0 +1,86 @@
...
+    #[instrument(skip(self, request))]
+    async fn community_pool_asset_balances(
+        &self,
+        request: tonic::Request<CommunityPoolAssetBalancesRequest>,
+    ) -> Result<tonic::Response<Self::CommunityPoolAssetBalancesStream>, Status> {
+        let state = self.storage.latest_snapshot();
+        let request = request.into_inner();
+
+        // Asset IDs are optional in the req; if none set, return all balances.
+        let asset_ids: Vec<AssetId> = request.asset_ids;
```
cronokirby, line 49: For correctness I think we to use the actual domain type here, because that's the one where we actually control the equality function, and that way we don't run into weird things if for some reason the RPC returns an alt denom or something.

crates/core/component/community-pool/src/component/rpc.rs
```diff
@@ -0,0 +1,86 @@
...
+                "failed to find community pool balances",
+            ))
+        })?;
+
+        let s = try_stream! {
+            for asset_balance in asset_balances {
+                let v = Value {
+                    asset_id: Some(asset_balance.0.into()),
+                    amount: Some(asset_balance.1.into())
+                };
```
cronokirby, line 63: We can simplify this a bit by using the fact that the existing Value domain type has this conversion method: https://rustdoc.penumbra.zone/main/penumbra_sdk_asset/struct.Value.html#impl-From%3CValue%3E-for-Value

cronokirby (commented):
great addition ; we can do a slight refactor of the impl to be a bit more streamlined though (I think)

cronokirby (approved):

conorsch:
Thanks for the addition; https://github.com/penumbra-zone/penumbra/pull/5198/commits/5ccf5fe0e837f1361f1aee59e296fc02f6ad9423 looks mighty fine.

PR #5199, opened 2025-05-18
support for external circuit construction

TalDerei:
## Describe your changes

generating the public and private circuit inputs is currently tightly coupled with action construction, and that needs to be disaggregated. The motivation is twofold: (1) avoid generating the circuit inputs twice (once for delegated proving and once for local proof generation), and (2) because we'll be making HTTPs requests to Taceo's MPC network from TS, which requires the public inputs as part of the request payload. We define `ActionCircuit` enum to encapsulate that. 

additionally, implements helper methods that (1) read the circuit inputs and perform constraint synthesis only (decoupling it from proof generation), and (2) serialize the resulting witnesses and matrices into an amenable form. 

pairs with https://github.com/penumbra-zone/penumbra/pull/5209


TalDerei:
closing in favor of cherry picking commits in https://github.com/penumbra-zone/penumbra/pull/5225, rebase here was more tangled than expected

PR #5204, opened 2025-05-29
fix: lp position withdraw causes negative liquidity values

vacekj:

### DIFF
--- a/crates/bin/pindexer/src/dex_ex/mod.rs
+++ b/crates/bin/pindexer/src/dex_ex/mod.rs
@@ -833,16 +833,18 @@ impl Events {
                 out.positions.insert(e.position_id, e.position.clone());
                 out.position_opens.push(e);
             } else if let Ok(e) = EventPositionWithdraw::try_from_event(&event.event) {
-                // TODO: use close positions to track liquidity more precisely, in practic I (ck) expect few
-                // positions to close with being withdrawn.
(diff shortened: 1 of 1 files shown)


cronokirby (commented):
I'm not sure there's a difference in behavior here.

The relevant logic is:
```rust
            (true, None, new) => (-(new.r1.value() as f64), -(new.r2.value() as f64)),
            (_, None, new) => ((new.r1.value() as f64), (new.r2.value() as f64)),
            (_, Some(old), new) => (
                (new.r1.value() as f64) - (old.r1.value() as f64),
                (new.r2.value() as f64) - (old.r2.value() as f64),
            ),
```

So turning `(true, None, Some(reserves))` into `(false, Some(reserves), 0)` should result in the same values, i.e. `-reserves`

conorsch:
Indeed, I'm still seeing negative values after running this changeset against mainnet event data:

```
penumbra_raw=# SELECT um, auction, dex, arb, fees
       FROM supply_total_unstaked
       ORDER BY height DESC
       LIMIT 1;
       um       |   auction   |     dex      |     arb     |   fees
----------------+-------------+--------------+-------------+-----------
 80982158646814 | 24571209329 | -34236022151 | 62870499597 | 353350630
(1 row)
```

N.B. although the database name in the output above is `penumbra_raw`, it's really a pindexer database.

Issue #5205, opened 2025-05-30
Support decoded addresses in TransactionSummary


VanishMax:
In TransactionSummary computation logic, all addresses are not checked if owned by user, even though all info is available for it: https://github.com/penumbra-zone/penumbra/blob/main/crates/core/transaction/src/view.rs#L128

Web apps call this code in WebAssembly to calculate TransactionSummary and, it is safe to say, expect the AddressView proto to be decoded in the protocol.

PR #5206, opened 2025-06-02
feat: wip: use address views in tx summary

cronokirby:
Only on the spend and output actions for now. We'll need to think a bit more about how to handle swaps.
- balance attribution for swaps isn't clear to me,
- we either need to pass an FVK, or augment the swap view with a decoded claim address.

This doesn't solve #5205 yet, but I think this already provides value for spend and output actions, which are a lot of the actions, and so I think merging it in this form is a good idea while we figure out the other salient points.

For testing, looking at the resulting pcli changes should be enough.

  > view only

### DIFF
--- a/crates/core/transaction/src/view.rs
+++ b/crates/core/transaction/src/view.rs
@@ -125,9 +125,7 @@ impl TransactionView {
                         // Provided imbalance (+)
                         let balance = Balance::from(note.value.value());
 
-                        let address = AddressView::Opaque {
-                            address: note.address(),
-                        };
+                        let address = note.address.clone();
 
                         effects.push(TransactionEffect { address, balance });
                     }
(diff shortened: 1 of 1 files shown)


VanishMax (approved):

conorsch:
> For testing, looking at the resulting pcli changes should be enough.

What changes should one expect in pcli output with this diff? To test, I sent a small amount of testnet funds to an external wallet, noted the tx hash, then ran `pcli view tx <hash>` on both this branch, and a local branch rebased on top of `protocol/lqt_branch`, but there's no difference in the tx info that's displayed. 

I was careful to run `pcli view reset` in between `pcli view tx <hash>` runs, as the path in the pcli code made it look relevant to do so. But the resulting tx info dumps are identical. Should the pcli output indeed change based on this diff?

Issue #5214, opened 2025-06-12
Release v2.0.0, via major version bump


conorsch:
# Tooling Release

In order to ship consensus-breaking protocol changes for LQT, let's prepare a `v2.0.0` release, shipping the contents of the `protocol/lqt_branch` branch. Further reading on development efforts can be found at:

* [Forum post on Liquidity Tournament](https://forum.penumbra.zone/t/uip-liquidity-tournament/159)
* #5010
* [Forum post on UIP-9](https://forum.penumbra.zone/t/uip-encrypted-lp-metadata/185)

## Changes to include

- [ ] Everything on current main, including #5213 
- [ ] Everything in the `protocol/lqt_branch`
- [ ] Including UIP-9 https://github.com/penumbra-zone/penumbra/pull/5212

## Compatibility
As this is a consensus-breaking, we should be clear in all release communications about the necessary for a coordinated chain upgrade.
Careful attention should be given to the delta between most recent stable tag in the v1.x series and the upcoming v2.x series.

conorsch:
Shipped: https://github.com/penumbra-zone/penumbra/releases/tag/v2.0.0 and `penumbra-1` upgraded to v2 earlier today, via a vote.

PR #5217, opened 2025-06-12
Move LQT tournament work on top of main

cronokirby:
PRIVACY IS A CULT
PENUMBRA IS THE KEY
UM IS THE TOKEN

(please do not squash merge this PR, for the love of all that is holy and sacred in this world)


conorsch (approved):
Refs #5010. There are a few more small changes before final tag lands (#5214), but the mainnet vote has passed, indicating that mainnet will move to this version of the protocol, so pulling the changes in to main.

PR #5225, opened 2025-06-19
protocol: delegated proving support

TalDerei:
## Describe your changes

HackMD: https://hackmd.io/5ZBNL4_xSISFRP77h_5nTw

The generation of public and private circuit inputs is currently tightly coupled with action construction, and this logic needs to be disaggregated. This refactor is necessary to support secret sharing of witness data and to allow sending both witness shares and public inputs as part of the payload in HTTP requests to Taceo’s MPC network. We define an `ActionCircuit` enum to encapsulate that logic.

Additionally, implements helper methods that:
1. Extract circuit inputs and perform _only_ constraint synthesis without initiating proof generation,
2. Serialize the resulting witness data into an amenable form.

cc @cronokirby @erwanor 

-------------

pairs with https://github.com/penumbra-zone/web/pull/2470, https://github.com/prax-wallet/prax/pull/373

### DIFF
--- a/crates/core/transaction/src/constraints.rs
+++ b/crates/core/transaction/src/constraints.rs
@@ -12,10 +12,12 @@ use std::io::Write;
 use crate::action::ActionCircuit;
 
 pub fn generate_and_serialize_circuit(circuit: ActionCircuit) -> Result<(Vec<u8>, usize)> {
-    //  Generate constraints only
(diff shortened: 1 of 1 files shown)


TalDerei:
This was an experimental spike that proved successful. It will remain a draft until there's consensus on further direction of our proving pipeline.

TalDerei:
closing as unplanned

PR #5232, opened 2025-06-30
Skip arb executions for dex explorer price charts

cronokirby:
## Describe your changes

This also creates a backwards compatible mechanism to allow app views to automatically reset themselves if relevant options change. In this case, this allows this change to run on top of an existing indexed database, with the result (test this) of just resetting the dex explorer component, since the options affecting indexing will have changed.

I also added a safeguard to prevent accidentally clobbering over the relevant app view when the options change. Instead, an explicit opt in with `--new-options` is required to indicate that you intended to reset that app view with new options.

In particular, to test, running pindexer SHOULD FAIL without `--new-options`, and then SHOULD SUCCEED with `--new-options`, with the result of re-running the dex_ex app view, this time without arb executions in the price chart. (There's a new flag to enable them again)

  > indexing only

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6591,6 +6591,7 @@ version = "2.0.1"
 dependencies = [
  "anyhow",
  "assert_cmd",
+ "blake2b_simd 1.0.2",
  "chrono",
  "clap",
  "cometindex",
--- a/crates/bin/pindexer/Cargo.toml
+++ b/crates/bin/pindexer/Cargo.toml
@@ -20,6 +20,7 @@ network-integration = []
 
 [dependencies]
 anyhow = {workspace = true}
+blake2b_simd = {workspace = true}
 clap = {workspace = true}
 chrono = {workspace = true, features = ["serde"] }
 cometindex = {workspace = true}
--- a/crates/bin/pindexer/src/dex_ex/mod.rs
+++ b/crates/bin/pindexer/src/dex_ex/mod.rs
@@ -1481,7 +1481,18 @@ impl AppView for Component {
     }
 
     fn version(&self) -> Version {
-        Version::with_major(1)
+        let hash: [u8; 32] = blake2b_simd::Params::default()
+            .personal(b"option_hash")
+            .hash_length(32)
+            .to_state()
+            .update(&self.denom.to_bytes())
+            .update(&self.min_liquidity.to_le_bytes())
+            .update(&[u8::from(self.ignore_arb_executions)])
+            .finalize()
+            .as_bytes()
+            .try_into()
+            .expect("Impossible 000-001: expected 32 byte hash");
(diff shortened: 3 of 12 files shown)


conorsch:
> In particular, to test, running pindexer SHOULD FAIL without `--new-options`

Yes, this is true.

> then SHOULD SUCCEED with `--new-options`

No, this is false: pindexer now fails even when `--new-options` is provided. At first, I thought this was a regression that only occurred when pindexer's dest database is empty, because that's the situation most of my testing scripts provide—and that's why the smoke tests are failing on this PR—but I also verified that when when I import a recent pindexer dbdump and use that as a dest db, the error still occurs:

```
2025-07-01T20:36:11.091534Z  INFO sqlx::postgres::notice: relation "index_watermarks" already exists, skipping                                                
2025-07-01T20:36:11.094763Z  INFO sqlx::postgres::notice: column "option_hash" of relation "index_watermarks" already exists, skipping                        
2025-07-01T20:36:11.100592Z  INFO cometindex::indexer: resetting index name="dex_ex" old_version=Version { major: Some(1), option_hash: None } new_version=Version { major: Some(1), option_hash: Some([35, 5, 18, 164, 147, 134, 223, 247, 102, 41, 226, 188, 61, 22, 64, 33, 232, 13, 44, 2, 77, 226, 133, 177, 238, 37, 233, 18, 149, 244, 0, 9]) }                                                                                                                                    
2025-07-01T20:36:11.118165Z  INFO sqlx::postgres::notice: drop cascades to table dex_ex.candles                                                               
Error: error returned from database: type "version" does not exist                                                                                            
                                                                                                                                                              
Caused by:                                                                                                                                                    
    type "version" does not exist  
```

That specific error should be rather easy to patch. However, what's the goal of the `--new-options` flag? It seems designed to preserve db integrity for operators, so that appviews don't change in surprising ways. Is that about the gist of it? If so, I'm not sure it's worth trouble: it's actually a pretty cool feature of pindexer that it can update appviews retroactively when new versions are released. So, I suppose my suggestion is to drop the caution of the `--new-options` flag altogether, but open to further discussion.

cronokirby:
The point of the `--new-options` flag is not to prevent new _versions_ from automatically being applied, but to prevent new _command line options_ from automatically being applied. The intent is that if you run pindexer once with certain flags, that affect how the app views process data, and then run it again forgetting some flag, by default it will not reset the app view, but instead fail.

cronokirby:
I'm a bit puzzled by the smoke test failures. I ran pindexer locally with an empty destination database, and no --new-options flag, and it succeeded. Is there some kind of transient state with the smoke tests?

conorsch:
'twas but a trailing backslash!

conorsch (approved):
Looking good! I've tested extensively locally. I'll be clear about the new flag in the release notes, and ping a few known operators out of band.

PR #5233, opened 2025-07-02
 fix(docs): correctly determine version tag in notes workflow

ianlv:
## Describe your changes

 The notes.yml workflow was failing on push-triggered runs because it directly accessed
  github.event.inputs.image_tag, which is not defined in the push event context. This change introduces a
  step to properly determine the documentation version tag based on the event trigger, checking for
  workflow_call or workflow_dispatch inputs and falling back to github.ref_name. This ensures the workflow
  runs reliably regardless of how it is initiated.

### DIFF
--- a/.github/workflows/notes.yml
+++ b/.github/workflows/notes.yml
@@ -4,7 +4,17 @@ on:
     branches:
       - main
   workflow_dispatch:
+    inputs:
+      image_tag:
+        description: "Image tag to use for the docs subdirectory."
+        required: false
+        type: string
   workflow_call:
+    inputs:
+      image_tag:
+        description: "Image tag to use for the docs subdirectory."
+        required: false
+        type: string
 
 jobs:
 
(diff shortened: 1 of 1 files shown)


ianlv:
> > The notes.yml workflow was failing on push-triggered runs
> 
> Incorrect, the docs workflow has been passing just fine, and runs on every merge to main, which is classed as a "push" event to main.
> 
> > because it directly accessed github.event.inputs.image_tag, which is not defined in the push event context
> 
> While it's true that `github.event.inputs.image_tag` is not defined, all occurrences of that string are suffixed with `|| github.ref_name` as a fallback, to ensure that a valid value is always used.
> 
> I appreciate your interest in the project! I'm not going to merge this change, because I don't want to introduce new inputs on image tags in the docs builds. Zooming out, we should probably remove all mention of versioning from both the CI and the docs URLs, as was done for the guide in #4301. See related discussion in #6.
> 
> If you'd like to take a stab at adapting #4301 for the protocol docs, @ianlv, I'd be happy to review! You can also find me in the project Discord if you'd like to chat.

Thanks for your kindly help!

PR #5236, opened 2025-07-03
fix: update option hash in indexing state

cronokirby:
## Describe your changes

We were not writing the option hash when updating the indexing state, causing failures when we tried to run pindexer again without "new-options"

To test, one should be able to run pindexer several times with the same options without needing the flag, and querying index_watermarks should show the hash.

  > indexing


conorsch:
Follow-up to #5232. Supersedes #5235.

conorsch (approved):
Nice, this fixes the problem for me!

Issue #5237, opened 2025-07-03
Release v2.0.2, via point-release


conorsch:
# Tooling Release

In order to ship some minor improvements and bug fixes for event indexing, let's prepare a `v2.0.2` release, flushing out the current contents of the main branch.

## Changes to include

- [ ] Everything on current main
- [ ] New pindexer logic in https://github.com/penumbra-zone/penumbra/pull/5232
- [ ] Fix to new pindexer logic in https://github.com/penumbra-zone/penumbra/pull/5236
- [ ] dev: picturesque https://github.com/penumbra-zone/penumbra/pull/5228

## Compatibility
As this is a point-release, all changes must be fully compatible for all nodes and clients.
Careful attention should be given to the delta between most recent tag on the main branch:
https://github.com/penumbra-zone/penumbra/compare/v2.0.1..main

PR #5238, opened 2025-07-03
Erwan/subaccount migration

erwanor:

### DIFF
--- a/crates/bin/pcli/src/command/migrate.rs
+++ b/crates/bin/pcli/src/command/migrate.rs
@@ -339,7 +339,18 @@ impl MigrateCmd {
                 if *plan_only {
                     println!("{}", serde_json::to_string_pretty(&plan)?);
                 } else {
-                    app.build_and_submit_transaction(plan).await?;
+                    // Ask for confirmation
+                    print!("Send transaction? (Y/N): ");
+                    std::io::stdout().flush()?;
+
+                    let response: String = std::io::stdin().lock().read_line()?.unwrap_or_default();
+                    let trimmed = response.trim().to_lowercase();
+
+                    if trimmed == "y" || trimmed == "yes" {
+                        app.build_and_submit_transaction(plan).await?;
+                    } else {
+                        println!("Transaction cancelled.");
+                    }
                 }
 
                 Result::Ok(())
--- a/crates/bin/pcli/src/command/migrate.rs
+++ b/crates/bin/pcli/src/command/migrate.rs
@@ -1,9 +1,10 @@
 use crate::App;
 use anyhow::{anyhow, Context, Result};
+use futures::TryStreamExt;
 use penumbra_sdk_asset::{asset, Value, STAKING_TOKEN_ASSET_ID};
 use penumbra_sdk_keys::FullViewingKey;
 use penumbra_sdk_num::Amount;
-use penumbra_sdk_proto::view::v1::GasPricesRequest;
+use penumbra_sdk_proto::view::v1::{AssetsRequest, GasPricesRequest};
 use penumbra_sdk_view::ViewClient;
 use penumbra_sdk_wallet::plan::Planner;
 use rand_core::OsRng;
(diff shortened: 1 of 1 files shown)


crates/bin/pcli/src/command/migrate.rs
```diff
@@ -30,6 +51,23 @@ pub enum MigrateCmd {
...
+    SubaccountBalance {
+        /// Range of source subaccount indices to migrate from (e.g., 0..17)
+        #[clap(long, required = true, value_parser = parse_range)]
+        from: std::ops::Range<u32>,
+        /// Destination FullViewingKey to migrate to
+        #[clap(long, required = true)]
+        to: String,
+        /// Only print the transaction plan without executing it (for threshold signing)
+        #[clap(long)]
+        plan_only: bool,
```
conorsch, line 69: The `--plan-only` option is very cool :ok_hand: I'd like to see more of this, preferably with derived Display impls so we can reuse liberally.

crates/bin/pcli/src/command/migrate.rs
```diff
@@ -30,6 +51,23 @@ pub enum MigrateCmd {
...
+    /// A FullViewingKey must be provided for the destination wallet.
+    /// Gas fees will be paid from the source subaccount with the most fee token.
+    #[clap(name = "subaccount-balance")]
+    SubaccountBalance {
+        /// Range of source subaccount indices to migrate from (e.g., 0..17)
+        #[clap(long, required = true, value_parser = parse_range)]
+        from: std::ops::Range<u32>,
+        /// Destination FullViewingKey to migrate to
+        #[clap(long, required = true)]
+        to: String,
```
conorsch, line 66: The FVK should be _read_ from user input, not provided as a CLI arg, to avoid storing the FVK in shell history forever. We recently made similar changes to the pclientd and pcli inputs:

* https://github.com/penumbra-zone/penumbra/pull/5173
* https://github.com/penumbra-zone/penumbra/pull/5174

and IMO we should use a similar solution here.

crates/bin/pcli/src/command/migrate.rs
```diff
@@ -30,6 +51,23 @@ pub enum MigrateCmd {
...
+    /// Migrate balances from specified subaccounts to a destination wallet.
+    ///
+    /// All assets from the specified source subaccounts will be sent to the destination wallet.
+    /// A FullViewingKey must be provided for the destination wallet.
+    /// Gas fees will be paid from the source subaccount with the most fee token.
+    #[clap(name = "subaccount-balance")]
+    SubaccountBalance {
+        /// Range of source subaccount indices to migrate from (e.g., 0..17)
+        #[clap(long, required = true, value_parser = parse_range)]
+        from: std::ops::Range<u32>,
```
conorsch, line 63: Ideally I'd like to see support for arbitrary indices, rather than a range, which can be specified as optional repeating args with a min length of 1. That way, one could use `pcli migrate subaccount 4 62` and only the subaccounts 4 and 62 would be migrated, with nothing in between.

This is a nice-to-have, though, so let's wait to implement until we have a need for it. For the time being, range support is a marked improvement and a huge timesaver, especially for the threshold signing workflow.

crates/bin/pcli/src/command/migrate.rs
```diff
@@ -119,6 +157,202 @@ impl MigrateCmd {
...
+                        }
+                    }
+                }
+
+                if subaccount_values.is_empty() {
+                    anyhow::bail!("no notes found in the specified subaccount range");
+                }
+
+                // Find the subaccount with the most fee token to pay fees
+                let (&(fee_account, _), _) = subaccount_values
```
conorsch, line 243: Autodiscovery of an appropriate subaccount for being fees is slick; I would have hardcoded this as a CLI arg. Optionally, we could add a CLI arg here, but I love the default behavior doing the right thing.

conorsch (commented):
Left some comments from initial review. I'd like to see the FVK handled safely, everything else we can punt to follow-ups.

conorsch:
> I'd like to see the FVK handled safely, everything else we can punt to follow-ups

I hereby volunteer to add this functionality; we're not in a rush, so I'll circle back on this PR to add some cleanups before merging. I _don't_ plan to include this changeset in the next point release (#5237).

PR #5243, opened 2025-07-09
feat: pindexer: take in block time as an option for lqt

cronokirby:
This gives us a much more stable time, since we're not dependent on any kind of variation in actual block. This also works out fine, since as blocks do happen, we only project based on the remaining blocks in an epoch, so we're "off" by less and less as the epoch goes on. Also a much better experience, since now the time is guaranteed to go down monotonically.

Closes https://github.com/penumbra-zone/web/issues/2538.

We should test in the usual way.

  > indexing only.

### DIFF
--- a/crates/bin/pindexer/src/lqt/mod.rs
+++ b/crates/bin/pindexer/src/lqt/mod.rs
@@ -94,29 +94,13 @@ mod _params {
 }
 
 mod _meta {
-    use penumbra_sdk_sct::event::EventBlockRoot;
-
     use super::*;
-
-    const DEFAULT_BLOCK_TIME_SECS: f64 = 5.0;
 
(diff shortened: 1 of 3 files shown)


crates/bin/pindexer/src/lqt/mod.rs
```diff
@@ -94,29 +94,15 @@ mod _params {
 }
 
 mod _meta {
-    use penumbra_sdk_sct::event::EventBlockRoot;
-
     use super::*;
 
     const DEFAULT_BLOCK_TIME_SECS: f64 = 5.0;
```
TalDerei, line 99: remove

TalDerei (approved):
LGTM

PR #5244, opened 2025-07-09
feat(pcli): support lqt votes by subaccount

conorsch:
## Describe your changes
Previously the `pcli tx lqt-vote` command only inspected subaccount 0 when searching for delegation notes to use in voting. If a user stores their delegations in a different subaccounts, it wasn't possible to vote with them. Here we allow a user to opt in to voting in the LQT via a specific subaccount, provided as `--source n`, the same CLI used for `pcli tx send`.

n/a

## Testing and review

Visual review of the diff should be sufficient, as the logic is straightforward.

  > only changes client-side tooling, no changes to app consensus logic

### DIFF
--- a/crates/view/src/storage.rs
+++ b/crates/view/src/storage.rs
@@ -1034,10 +1034,8 @@ impl Storage {
             let mut results = Vec::new();
             for record in spendable_note_records {
                 // Skip notes that don't match the account index, if declared.
-                if let Some(address_index) = address_index {
-                    if record.address_index.account != address_index.account {
-                        continue;
-                    }
+                if matches!(address_index, Some(a) if a.account != record.address_index.account) {
+                      continue;
                 }
                 let asset_id = record.note.asset_id().to_bytes().to_vec();
                 let denom: String = dbtx.query_row_and_then(


cronokirby (approved):

conorsch:
Downgrading to draft so I can evaluate the multi-vote behavior via subaccounts on testnet.

conorsch:
Found a bug in the votable note lookups. At encouragement from @cronokirby, investigated the logic in `storage.rs` and found that if a subaccount is specified—which is usually isn't, given the default of None—the subaccount filter wasn't working, and no notes were returned. I don't quite understand what was incorrect about the sql implementation, but I modified the code to use Rust types, and now it behaves as expected.

crates/view/src/storage.rs
```diff
@@ -1036,6 +1033,12 @@ impl Storage {
             // do it this way; if it becomes slow, we can do it better
             let mut results = Vec::new();
             for record in spendable_note_records {
+                // Skip notes that don't match the account index, if declared.
+                if let Some(address_index) = address_index {
+                    if record.address_index.account != address_index.account {
+                        continue;
+                    }
+                }
```
cronokirby, line 1041:
```suggestion
                if matches!(address_index, Some(a) if a.account != record.address_index.account) {
                      continue;
                }
```

Potential improvement?

PR #5248, opened 2025-07-12
feat: make view service respect account ownership in positions

cronokirby:
Previously, we were ignoring the account index.

This modifies syncing to detect the account by looking at the note you get when you receive an LP NFT.

This is useful, because it allows segregating funds by account, instead of having to make a new wallet.

Smoke testing should be sufficient here.

  > client only


erwanor (approved):
Optimisitically, lgtm

PR #5249, opened 2025-07-14
fix(pindexer): resolve NaNs in dex_aggregate_summary

conorsch:
## Describe your changes
Evaluating this patch as a potential resolution for https://github.com/penumbra-zone/web/issues/2617 

It appears that the mainnet pindexer db started emitting `NaN` values in the `dex_aggregate_summary` table on 2025-07-13, which caused some application crashes. I made sure that the problem was not specific to pindexer `v2.0.3`, released last week, and was replicable on `v2.0.2`. It looks to my eye like there's a logic bug in pindexer relating to rolling time windows, that only now surfaced due to mainnet data conditions.

https://github.com/penumbra-zone/web/issues/2617

  > no, indexing only


conorsch:
Trialed the new logic on a local veil explore page, and it appeared to resolve the total trading volume stats:

dex-explore-comparison

Intriguingly the total tx count is 2 off, which might be a good lead.

conorsch:
Refs #5249; superseded by #5251.

Issue #5250, opened 2025-07-14
bug: pindexer mainnet data contains NaN values


conorsch:
**Describe the bug**
On 2025-07-13 the mainnet `pindexer` db started recording `NaN` values in Postgres. This shoudn't happen: the cometindex logic should be defensive about value ranges.

**To Reproduce**
1. Run pindexer (at least `v2.0.3` or `v2.0.2`, although presumably all versions are affected)

**Expected behavior**
There should be no `NaN` values in the database: such edge cases should be handled in the Rust code and resolved to usable values inside the db. 

**Screenshots**
Here's an example query that shows `NaN` files in a prod pindexer db:

```
pindexer=> SELECT the_window, encode(asset_start, 'hex') as asset_start, encode(asset_end, 'hex') as asset_end, direct_volume_indexing_denom_over_window, swap_volume_indexing_denom_over_window FROM dex_ex_pairs_summary WHERE direct_volume_indexing_denom_over_window = 'NaN' OR swap_volume_indexing_denom_over_window = 'NaN' ORDER BY the_window;
 the_window |                           asset_start                            |                            asset_end                             | direct_volume_indexing_denom_over_window | swap_volume_indexing_denom_over_window
------------+------------------------------------------------------------------+------------------------------------------------------------------+------------------------------------------+----------------------------------------
 1d         | 414e723f74bd987c02ccbc997585ed52b196e2ffe75b3793aa68cc2996626910 | 29ea9c2f3371f6a487e7e95c247041f4a356f983eb064e5d2b3bcf322ca96a10 |                                      NaN |                                    NaN
 1d         | 414e723f74bd987c02ccbc997585ed52b196e2ffe75b3793aa68cc2996626910 | 76b3e4b10681358c123b381f90638476b7789040e47802de879f0fb3eedc8d0b |                                      NaN |                                    NaN
 1mo        | 414e723f74bd987c02ccbc997585ed52b196e2ffe75b3793aa68cc2996626910 | 76b3e4b10681358c123b381f90638476b7789040e47802de879f0fb3eedc8d0b |                                      NaN |                                    NaN
 1mo        | 414e723f74bd987c02ccbc997585ed52b196e2ffe75b3793aa68cc2996626910 | 29ea9c2f3371f6a487e7e95c247041f4a356f983eb064e5d2b3bcf322ca96a10 |                                      NaN |                                    NaN
[...]

PR #5251, opened 2025-07-14
fix: pindexer: don't index candles with NaN

cronokirby:
## Describe your changes

This skips indexing candles with a price of NaN, at the source where we ingest point candles, representing a single trace input and output value. For some reason, the dex sometimes produces candles with 0 input and 0 output volume, resulting in a NaN, which then pollutes and corrupts other data. This fix addresses the source of the pollution in indexing.

We should investigate why these traces are being emitted.

To test, run pindexer again, observe that NaNs are out of the candles.

  > Indexing only.

### DIFF
--- a/crates/bin/pindexer/src/dex_ex/mod.rs
+++ b/crates/bin/pindexer/src/dex_ex/mod.rs
@@ -747,6 +747,10 @@ impl Events {
         let pair_2_1 = pair_1_2.flip();
         let input_amount = f64::from(input.amount);
         let output_amount = f64::from(output.amount);
+        if input_amount == 0.0 && output_amount == 0.0 {
+            tracing::warn!(?input, ?output, "ignoring trace with 0 input and output");
+            return;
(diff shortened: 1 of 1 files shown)


conorsch:
I think there's a side-effect here where counts for trades and volume get doubled; see side-by-side comparison with prod pindexer db and local instance backed by this PR:

dex-explore-compare-with-pindexer-5251

Intriguingly the per-pair daily volumes were already reported as 2x in pindexer as compared to the Veil list; as of this PR, though, the values are doubled again, to 4x the value displayed in the Veil table.

conorsch:
Given the diff the doubling doesn't seem like an error; if anything, perhaps the older counts were the inaccurate ones. I'm going to merge this and release it. As always, we'll continue to iterate on the indexing logic.

conorsch (approved):

conorsch:
Refs #5250.

Issue #5255, opened 2025-07-28
pd: simplify migration process


erwanor:
Every penumbra migration is the same one hundred lines of migration setup and cleanup boilerplate, wrapped around ten lines of important migration logic. The boilerplate encodes subtle folklore knowledge about how the upgrade process works. For example, committing storage should be done with `Storage::commit_in_place` to keep the JMT state versioning aligned on block height. There are many more e.g, halt bit flag reset, app height overwriting, and signing state watermark cleanup.

This folklore knowledge only exists in copy-paste lineage across the various migration scripts. A lot of engineering effort is burned on making sure that the boilerplate is authentic, rather than on the important migration logic payload. That is of course backwards.

**Migration flow**

There are a few migration scripts in `pd::migrate`, that always follow this pattern:

- setup: load storage, chain id extraction, root hash check
- migration: state transition business logic
- finalize: commit, app-height zero, generate checkpointed genesis, update watermark file

**`trait Migration`**

```rust
pub trait Migration {
    /// The unique identifier for this migration (e.g., "mainnet-4", "testnet-78")
    fn name(&self) -> &'static str;
    
    /// The target app version after migration
    fn target_app_version(&self) -> Option<u64> or u64 {
        None?
    }
    
    /// Prepare the migration, return pre-upgrade root hash and height
    async fn prepare(&self, pd_home: &PathBuf, comet_home: Option<&PathBuf>) 
        -> Result<(RootHash, Height)> { /* default impl */ }
    
    /// Execute the migration, return post-upgrade root hash and height
    async fn migrate(&self, pd_home: &PathBuf, comet_home: Option<&PathBuf>) 
        -> Result<(RootHash, Height)> { /* default impl calls migrate_inner */ }
    
    /// The actual migration logic - this is what implementations override
    async fn migrate_inner(&self, delta: &mut StateDelta<Snapshot>) 
        -> Result<()> { /* default: no-op */ }
    
    /// Complete the migration with genesis generation
    async fn complete(&self, pd_home: &PathBuf, comet_home: Option<&PathBuf>) -> Result<()> { /* default impl */ }
}
```

As a result, writing migrations becomes easy and accessible.

```rust
pub struct RescueIbcClient {
    pub consensus_state: Vec<u8>,
    pub target_height: u64,
}

impl Migration for RescueIbcClient {
    fn name(&self) -> &'static str { "mainnet-5" }
    fn target_app_version(&self) -> Option<u64> { Some(12) }
    
[...]

PR #5256, opened 2025-07-28
pd(migrate): first pass at simpler migrations

erwanor:
## Describe your changes

This is an early implementation of #5255 

Testing in progress but this should work ;)

  > Migration changes


crates/bin/pd/src/migrate/migrate2/framework.rs
```diff
@@ -0,0 +1,182 @@
...
+        if let Some(comet_home) = comet_home {
+            let genesis_path = pd_home.join("genesis.json");
+            crate::migrate::migrate_comet_data(comet_home.to_path_buf(), genesis_path).await?;
+        }
+
+        Ok(())
+    }
+
+    #[instrument(skip(self, pd_home, comet_home))]
+    async fn run(
```
cronokirby, line 141: I'm unsure if allowing this to be overriden makes sense. The other hooks, I can see a use case, but overriding this just seems like a footgun.

cronokirby (approved):
I like having default impls for the trait methods, but maybe we should move the "orchestration" method into a top level function that can't be overriden.

PR #5263, opened 2025-08-26
ibc: implement client recovery via hard-forks

erwanor:
## Describe your changes

This PR adds IBC client recovery via hard-forks.

When a Tendermint client expires, full node runners and validators can coordinate to fast-forward the client to a healthy state. This is done in two steps: First, a new healthy client is created which tracks the same counterparty chain on similar trust parameters. Then, a hard-fork is coordinated, and executed via `pd migrate`.

For example, in the case of client `0` expiring, a new client `1` is created. The hard-fork is executed via:

```
pd migrate ibc-recovery --old-client-id 07-tendermint-0 --new-client-id 07-tendermint-1
```

Note that relaying genesis restarts (context in [hermes docs](https://hermes.informal.systems/advanced/troubleshooting/genesis-restart.html)) still needs to occur.

When the upgrade uses UIP-4 and bumps `APP_VERSION` the migrate command is used as follow:

```
pd migrate ibc-recovery --old-client-id 07-tendermint-0 --new-client-id 07-tendermint-1 --target-app-version=12
```

**Implementation**

The implementation is done via an extension trait in the Penumbra IBC implementation. In particular, validation rules are directly inspired by [ADR-26](https://ibc.cosmos.network/architecture/adr-026-ibc-client-recovery-mechanisms/). Specifically, we check that the substitute is the same tendermint client as the subject, just unstuck.

**Alternatives considered**

This mechanism can be integrated as part of on-chain governance. However, although forcing social consensus is more onerous, it avoids increasing privilege load on the governance component. For that reason, it is avoided in favor of hard-forks.

  > consensus breaking when exercised.


conorsch:
Logic is sound. To test, I paired with @erwanor, and we both separately configured local Penumbra devnets to relay packets via hermes, forced client expiry on the Penumbra side, ran the migration, and confirmed we could round-trip funds along the original channel post-migration. It worked. Below I document the approximate steps taken to verify the functionality, for posterity.

<details>
<summary>
hermes devnet session notes
</summary>

```
# first, create a local devnet using the most recent penumbra stable version. use fast block times to accelerate state transitions.
cargo run --bin pd -- network geenrate --proposal-voting-blocks 100 --unbonding-delay 500 --epoch-duration 100 --chain-id "private-network"

# then run the local solo-validator network
just dev

# make sure you've got the latest hermes stable release locally
# i tested with: https://github.com/informalsystems/hermes/releases/tag/v1.13.2
# setup hermes keys and configs to point at the relevant chains.

# create short-lived client for stride on penumbra devnet
hermes create client --trusting-period 300s --reference-chain stride-1 --host-chain private-network-x85dc887f

# keep client active on penumbra
watch -n 30 -d 'hermes update client --host-chain private-network-x85dc887f --client 07-tendermint-0'

# create longer-lived client for penumbra on stride mainnet
hermes create client --trusting-period 2400s --reference-chain private-network-x85dc887f --host-chain stride-1

# keep client active on stride
watch -n 200 -d 'hermes update client --host-chain stride-1 --client 07-tendermint-173'

# create connection for the manually created clients
 hermes create connection --a-chain stride-1 --a-client 07-tendermint-173  --b-chain private-network-x85dc887f --b-client 07-tendermint-0
# which created on the penumbra side: connection-0
# and on the stride side: connection-180

# now we open a channel, referencing the manually created connection:
hermes create channel --a-chain stride-1 --a-port transfer --b-port transfer  --a-connection connection-180

# now we deposit a small bit of tokens from the stride network to penumbra devnet:
hermes tx ft-transfer --src-chain stride-1 --src-port transfer --src-channel channel-345 --amount 2000000 --dst-chain private-network-x85dc887f --denom ustrd --timeout-second
s 100 --receiver penumbra1flg69zy0fha79nrqjtyeyfzy2s26fnehtpkksfn72esrm2wnwyjj8wh0trawar782d77vd9e3vuuhkdtcnmnra9s848d5zz8fp6crrtf56a754p2d4je7g9sgzkftn58wsygcc

# then we withdraw a bit of those funds via pcli, to confirm round-tripping, i.e. bi-directional relaying is fully operational
# once confirmed, we cancel the "watch" command that's updating the client on the Penumbra chain.
# it's short lived, so it will expire within a few minutes. Use ibc-monitor to watch!
# After it's expired, try again to renew:
#
#   hermes update client --host-chain private-network-x85dc887f --client 07-tendermint-0
#
# \and confirm there's an error message, e.g.
#
#   ERROR foreign client error: client 07-tendermint-0 on chain id private-network-x85dc887f is expired: time elapsed since last client update: 389.389326966s
#
# then we can proceed with testing the migration logic, which requires that we first create a new client.
hermes create client --trusting-period 1d --reference-chain stride-1 --host-chain private-network-x85dc887f
# which created on the penumbra side: `07-tendermint-1`, which is the new client id required for the migration.

# then we vote for upgrade on the devnet, use the latest version of pd built from https://github.com/penumbra-zone/penumbra/pull/5263 to run the migration:

pd-next migrate ibc-recovery --old-client-id 07-tendermint-0 --new-client-id 07-tendermint-1 --target-app-version=12

# restart the solo validator and confirm it generates blocks. then use an archive node to perform hermes genesis-restart as described in
# https://hermes.informal.systems/advanced/troubleshooting/genesis-restart.html?highlight=genesis%20restart#updating-a-client-after-a-genesis-restart-without-ibc-upgrade-proposal
# resume relaying, and test a transfer.
# confirm that you can round-trip funds on the original channel post-migration!
```
</details>

I'm fine with this landing as-is. We'll need to add clear messaging in the release notes about usage and of course the any consensus-breaking changes.

PR #5265, opened 2025-08-28
pd: make it easy to do an upgrade hard-fork

erwanor:
## Describe your changes

This PR adds `pd migrate no-op [--target-app-version=<APP>]` which is tooling that can be used to produce a hard-fork upgrades. It uses the new migration framework and can serve as an example for prospective protocol engineers.

  > consensus breaking when exercised

### DIFF
--- a/crates/bin/pd/src/cli.rs
+++ b/crates/bin/pd/src/cli.rs
@@ -163,7 +163,13 @@ pub enum MigrateCommand {
         new_client_id: String,
         /// Optional app version to set during migration.
         #[clap(long, value_name = "VERSION")]
-        app_version: Option<u64>,
+        target_app_version: Option<u64>,
+    },
+    /// Perform a no-op migration that resets the halt bit and produces a new genesis.
+    NoOp {
+        /// Optional app version to set during migration.
+        #[clap(long, value_name = "VERSION")]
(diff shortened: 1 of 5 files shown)


crates/bin/pd/src/cli.rs
```diff
@@ -163,7 +163,13 @@ pub enum MigrateCommand {
...
         /// Optional app version to set during migration.
         #[clap(long, value_name = "VERSION")]
-        app_version: Option<u64>,
+        target_app_version: Option<u64>,
+    },
+    /// Perform a no-op migration that resets the halt bit and produces a new genesis.
+    NoOp {
+        /// Optional app version to set during migration.
+        #[clap(long, value_name = "VERSION")]
+        target_app_version: Option<u64>,
```
conorsch, line 172: I'm not sure I agree that the `--target-app-version` should be _optional_, as I struggle to identify a use case when a discrete target app version shouldn't be declared.

conorsch (approved):
Minor quibble about the optionality here—I slightly prefer forcing a specific app version to be set—but that's not worth blocking over, especially because this'll serve as a reference to future maintainers.

conorsch:
Food for thought: what's the difference between `pd migrate no-op` and `pd migrate ready-to-start`?

PR #5272, opened 2026-01-04
pd: implement mass pruning of chain state

erwanor:
**Describe your changes**

This PR includes a prototype of chain state pruning. It reduces mainnet storage requirements from ~160GB, at time of writing, to 16GB. 

The gist of this work ha been tested by synchronizing a pruned full node with mainnet, without any app hash divergence.

However, it is not yet ready for inclusion in the root tree. The crux of this code is in the process of being refactored, and will live in the [cnidarium](https://github.com/penumbra-zone/cnidarium) repository with the rest of the storage layer software. This PR does not include the **streaming implementation** of pruning, which exists but needs to be refined.

The PR does not compile because it effectively relies on a local patched version of cnidarium. I am sharing it to start getting feedback, and get the ball rolling.

The API to use pruning is `pd migrate prune`. The pruning code performs check and produce a report, making sure that the tree root hash has not changed and 

**Context**
The Jellyfish Merkle Tree (JMT) is a **versioned** sparse merkle tree data structure.
Every update to the tree bumps a monotonic version counter. The branching factor for
the JMT is 16. Each key insert is hashed into a 256 bits path key.
k := H(key) e.g, 0x0123456789abcdef, where each nibble is a conceptual level in the tree.
It is conceptual because the tree is compact so depth can be lower. Nodes are addressed by
their version and nibble path. The former allows storage and abstraction layers to improve
locality of caches and SSD layers.

By default, historical state is preserved which allows full nodes to answer historical
queries, and generate proofs at any version. In practice, there is no code using this
conceptual feature. It just takes up storage.

**Amplification**
Without pruning, stale nodes are preserved across writes. This allows a tree
to answer historical queries about arbitrary versions, at the cost of storage
and slower queries on cold rocksdb seeks.

The amplification from those extranodes is superlinear in the number of inserts.
The quick intuition is that each key update results in a new leaf and a stack of
ancestors. Without pruning, those ancestors nodes are never reclaimed. Therefore,
each update results in one stack of node per update.

How many nodes are in that stack? It varies, but it will be about the height of
the compact tree which scales with the logarithm of the # of keys in the tree.
that gives an amplification of N * log M, where N is the number of updates, and
M the number of keys. We don't account for overlapping ancestor nodes since it
asymptotically resolves to a logarithm factor anyway.

**Implementation**
JMT pruning is implemented as a `Migration`](super::Migration) that overwrites
default impls to no-op any genesis, or application state preparation. Since JMT
pruning MUST NOT be consensus breaking, it does not require a hard-fork or any
coordination. It is an entire local decision transparent to the rest of the
network.

Defining it as a local migration has two benefits, JMT pruning:
1. .. is just another offline state transition, and reuses existing tooling.
2. .. can be composed by other migration scripts, or used on its own.

  > not consensus breaking


crates/bin/pd/src/migrate/migrate2/prune_tree.rs
```diff
@@ -0,0 +1,480 @@
...
+    config: Arc<SubstoreConfig>,
+}
+
+impl PruningTreeStore {
+    fn new(db: Arc<DB>, config: Arc<SubstoreConfig>) -> Self {
+        Self { db, config }
+    }
+}
+
+impl TreeWriter for PruningTreeStore {
```
erwanor: All of this code should live in cnidarium

erwanor: Lifted into https://github.com/penumbra-zone/cnidarium/pull/12

conorsch:
> The crux of this code is in the process of being refactored, and will live in the [cnidarium](https://github.com/penumbra-zone/cnidarium) repository with the rest of the storage layer software.

Refs https://github.com/penumbra-zone/cnidarium/pull/12

